PR #54, opened 2013-08-21
Wordpress style summaries

rozza:
Allow full control of summaries which can be rendered as html rather
than text.  Using a `<!--more-->` html comment in your markdown / rst
you can indiciate where the summary should end and have the summary
converted to html.

### DIFF
--- a/hugolib/helpers.go
+++ b/hugolib/helpers.go
@@ -20,6 +20,7 @@ import (
 	"github.com/kr/pretty"
 	"html/template"
 	"os"
+	"os/exec"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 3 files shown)


mysphereseo:
When multiple posts are shown, it is easier for your readers to find what they want by only giving them a summary of each post rather than showing them the entire post. You can enable this post summary functionality by using the WordPress the_excerpt command. Here is an example blog with styled WordPress excerpts.
Get in touch at +91 999 800 4498
Inquire us your requirement at http://www.mysphereinfotech.com/getintouch.php
Website  -   http://www.mysphereinfotech.com/

PR #134, opened 2013-11-23
Example of a multilingual site.

egonelbre:
This is an example how I setup up a multilingual site. It still has some problems with custom selecting a layout file and url "/" not working. I couldn't use indexes for news item listing because the "indexes" don't allow front matter and it does automatic title pluralization.

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -18,6 +18,7 @@ import (
 	"github.com/spf13/cobra"
 	"net/http"
 	"strconv"
+	"strings"
 )
 
 var serverPort int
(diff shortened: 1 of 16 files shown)


simplyeffectiveweb:
I've just been playing with this today, and with a few tweaks English and Estonian pages work, based on clicking the preferred language.

However, one problem I can not get round is how to start with content. At present the 'start' page is blank (as no language and language folder is selected).

Other than .htaccess stuff once on a server, can anyone think of a simple way in Hugo to start with one or the other language?

egonelbre:
@simplyeffectiveweb For the page you want to be the first add `url: /`, alternatively add it as an [alias](http://gohugo.io/extras/aliases/) to the front matter. See the `index.en.md` in the example.

egonelbre:
@simplyeffectiveweb nevermind I forgot that it didn't work. Basically create an `index.html` with a redirect to your intended page, e.g.:

```html
<meta http-equiv="refresh" content="0; url=/home" />
```

egonelbre:
I've update the multilingual site https://github.com/egonelbre/hugo/tree/fix-multilingual it should work.

simplyeffectiveweb:
Ah yes, just tried the fix-multilingual and can confirm it works (and also the news works which it did not for me before).

Thanks for looking at that so quickly. It helps me a lot, and I think a lot of people will find this very useful.

Issue #211, opened 2014-02-28
Feature to have pagination stay within a section


abh:
I am using this in my "posts" template (yes, I think next/previous are backwards :-) )

        <ul class="pager">
          {{if .Next}}
            <li><a class="previous" href="{{.Next.Permalink}}">← {{.Next.Title}} </a></li>
          {{end}}
          {{if .Prev}}
            <li><a class="next" href="{{.Prev.Permalink}}">{{.Prev.Title}} →</a></li>
          {{end}}
        </ul>

I didn't expect it to include pages from outside the current section, but looking at the code I see there's just a concept of one list of pages.

It'd be nice with a way to have .Next and .Prev only iterate on items in the current section (I realize the follow-up question from someone else might be "in the current category" or other meta-data).

(I posted this [on the mailing list](https://groups.google.com/forum/#!topic/hugo-discuss/t4DsQ-VLQ0Y) earlier but I think it's more appropriate here now that I know better how the code works).

bjornerik:
This is solved in more than one way.

bryfry:
What are the ways! 

https://xkcd.com/979/

moorereason:
@bryfry,
Please use the discussion forums for this kind of request in the future.

You want `.NextInSection` and `.PrevInSection` from https://gohugo.io/templates/variables/.

bryfry:
Sorry, posted in humor and frustration.  Thanks for the link.

PR #266, opened 2014-04-24
Add in-section Next/Prev section pointers

johnsto:
Allows navigation to be added to pages that is restricted to the current section.

Templates can use `.PrevInSection` and `.NextInSection` in the same way they use `.Prev` and `.Next`.

Issue: #211

(Moved from #259)

### DIFF
--- a/docs/content/layout/variables.md
+++ b/docs/content/layout/variables.md
@@ -31,6 +31,8 @@ matter, content or derived from file location.
 **.TableOfContents** The rendered table of contents for this content<br>
 **.Prev** Pointer to the previous content (based on pub date)<br>
 **.Next** Pointer to the following content (based on pub date)<br>
+**.PrevInSection** Pointer to the previous content within the same section (based on pub date)<br>
+**.NextInSection** Pointer to the following content within the same section (based on pub date)<br>
 **.FuzzyWordCount** The approximate number of words in the content.<br>
 **.WordCount** The number of words in the content.<br>
 **.ReadingTime** The estimated time it takes to read the content in minutes.<br>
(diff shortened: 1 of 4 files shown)


bjornerik:
I'll reopen this as this is functionality that I miss, a very common use case and not really "added via b719ba7".

moorereason:
Thanks for reopening, @bjornerik.  This is an important missing feature.

@spf13 already declined to merge this PR, and I tend to agree with him that this PR is not the best (idiomatic?) long term solution (even though I've applied this PR locally).  [I posted some ideas](http://discuss.gohugo.io/t/proposals-for-next-prev-in-section/434) to the dev discussion forum last month, but I didn't get any feedback.  Perhaps we could find a better solution.

@spf13, care to comment here or in the discussion forum on what you'd like to see instead of this PR?

bjornerik:
Not sure what "idomatic" means in this context -- I understand the concern about polluting the structs with very specific fields, favoring more generic functions that can be mixed and matched.

But this is a very common use case, and people "guessing the API" without looking at the doc would probably try your implementation.

To quote John Carmack (the man behind Doom): 

"If you’re willing to restrict the flexibility of your approach, you can almost always do something better."

bjornerik:
@spf13 I have now started using this patch myself, and I hate having too many custom patches going on (the branching confuses me ...) ... The implementation looks clean enough -- I'm going to merge this if you don't shout in a day or so; the acceptance of silence :-)

spf13:
I'm fine with this overall. I agree that it's a common enough use case to call it out specifically. 

I don't love the current method names, but I can think of a better one.

We should make sure to document clearly that there is a general way to get Prev & Next and we've provided two shortcuts (global & section).

bjornerik:
The one who waits for something good ... Merged in a5f5106 - thanks!

PR #338, opened 2014-06-21
taxonomies -> Taxonomies

mhe:
Only after changing "taxonomies" to "Taxonomies" custom taxonomies appeared to work (hugo 0.11).

### DIFF
--- a/docs/content/taxonomies/usage.md
+++ b/docs/content/taxonomies/usage.md
@@ -22,7 +22,7 @@ Notice the format is **singular key** : *plural value*.
 ### config.yaml
 
     ---
-    taxonomies:
+    Taxonomies:
         tag: "tags"
         category: "categories"
         series: "series"


anthonyfok:
Sorry, this doesn't make sense: it is rather confusing or even disconcerting to me.  Who would have expected to use a capitalized `Taxonomies:` when all other top-level variables (?) use lowercase.

Thanks to @jaden's investigation, he found https://github.com/spf13/viper/blob/master/viper.go#L243 where normalizes the string to lowercase, probably to avoid this inconsistency.

Anyhow, to keep track, I am reopening this pull request until more is discovered about the history and developments since v0.11, so that perhaps we may fix the docs or the code accordingly.

anthonyfok:
It turns out it was only in Hugo v0.11 that we had to write "taxonomies" as "Taxonomies".
Here are the test results:

| Hugo Version | `indexes` | `Indexes` | `taxonomies` | `Taxonomies` |
|-----------|:----------:|:----------:|:-----------:|:------------------:|
| v0.10 | ✔ | ✔ | Not supported | Not supported |
| v0.11 | ✔ | ✔ | ✘ | ✔ |
| v0.12 | ✔ | ✔ | ✔ | ✔ |
| v0.13-DEV | ✔ | ✔ | ✔ | ✔ |

On August 5, 2014, @natefinch discovered this bug in Viper, filed Issue https://github.com/spf13/viper/issues/8: "Aliases don't work in config file", and submitted the fix in Pull Request https://github.com/spf13/viper/pull/9, which @spf13 promptly merged as https://github.com/spf13/viper/commit/2b24bea958e2d411fc25a82e44fbbcc3b6ed0441.

Hugo v0.12, which was released on September 1, 2014, is thus free from this bug.  Henceforth, we can safely use `[taxonomies]` in the site-wide `config.toml`.

The documentation will be updated accordingly.  :-)

jaden:
Thanks for researching this, it's much appreciated.

PR #475, opened 2014-09-03
Add support for excluding pages from the feed

tru:
This commit adds the possibility to exclude pages from the main RSS
feed.

It adds a new attribute on the page called excludefromfeed and in the
site generator it will just ignore those pages when generating the feed.


spf13:
CLA not signed. Closing.

PR #588, opened 2014-10-28
Add configuration option 'Title404'

halostatue:
Allow 'Title404' to be set for the title of the 404 page in config.*.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -43,8 +43,8 @@ import (
 var HugoCmd = &cobra.Command{
 	Use:   "hugo",
 	Short: "hugo builds your site",
-	Long: `hugo is the main command, used to build your Hugo site. 
-	
+	Long: `hugo is the main command, used to build your Hugo site.
+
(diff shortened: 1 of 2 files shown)


bep:
Currently you can do

```
{{ $.Scratch.Set "title" "You are in a maze of passages, all alike." }}
{{ partial "header.html" . }}
```

It still feel hackish, but it's better than nothing. I do not believe that this PR is ever gonna get merged in its current form; so maybe best to close it?

halostatue:
Let me see if the `{{ .Scratch… }}` solution works for me. If it does, I’ll close this because I have an alternative solution. If not, I would like to leave it open because I consider the fact that the 404 title is statically set to be a bug for which this is a (poor, but working) solution.

bep:
Second thought, leave it open. I agree that it's a bug.

bep:
Replaced by https://github.com/spf13/hugo/issues/1298

halostatue:
I have been doing some work to make this enabled in my template anyway.

PR #644, opened 2014-11-16
Various fixes in RSS feed generation

k4rtik:
- Prevent `.xml` generation for root section
- Remove redundant check for DisableRSS
- Fix permalinks for rel="alternate"
- Rename generated xml file to `<type>/index.xml`
- Add required description element in default template
- Make default RSS template validate on w3c (timezone format is still an issue)


lpar:
+1 for switching to Atom. I was astonished that you'd bothered to implement the obsolete RSS 2 spec.

PR #664, opened 2014-11-22
add support for deleting files from the publish directory during fsync

mohae:
add delete support for fsync. This addresses https://github.com/spf13/hugo/issues/379 by adding a boolean knob, `DeletePublishDir`, that sets `fsync's` Delete property. 

By default this flag is set to false, which preserves current behavior.  If the flag is set to true, orphaned files in the `PublishDir`, files that were not synced, are deleted.

This does not address stale content being received by the user due to caching issues between Hugo and the browser. Such caching issues are probably best addressed by supporting use of cache-busting.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -52,7 +52,7 @@ Complete documentation is available at http://gohugo.io`,
 
 var hugoCmdV *cobra.Command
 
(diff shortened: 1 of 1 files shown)


vuvgaG:
Has this option been disabled?

Issue #655, opened 2014-11-20
Add Image/Video Sitemaps


derekperkins:
I get a lot of traffic through my images, so it would be nice to have some media sitemap options.

Jos512:
Any update on including images in the sitemap? This is especially important for people who want to use a CDN. Since that would give duplicate content (CDN images vs. origin), a sitemap is a tool to communicate to Google what the preferred URL for the content is.

See [image sitemaps](https://support.google.com/webmasters/answer/178636?hl=en) and [Google webmaster blog](http://googlewebmastercentral.blogspot.nl/2012/04/1000-words-about-images.html) for more.

christophermancini:
I see that there is a Media.go object and the Image and Video structs are referenced there, but there is not docs on what, if anything they provide in terms of images and sitemaps.

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

thomasjsn:
This seems to be pretty easy now with page resources, I added this just before `</url>` in `layouts/_default/sitemap.xml`

```xml
  {{ range $i := .Resources.ByType "image" }}
  <image:image>
    <image:loc>{{ $i.Permalink }}</image:loc>
  </image:image>
  {{ end }}
```

PR #675, opened 2014-11-25
Added Images to sitemap.xml

derekperkins:
#655 This is dependent on image metadata being standardized as discussed at http://discuss.gohugo.io/t/hugo-seo-social-partials/353/17

### DIFF
--- a/docs/content/templates/sitemap.md
+++ b/docs/content/templates/sitemap.md
@@ -41,7 +41,12 @@ Protocol](http://www.sitemaps.org/protocol.html).
         <loc>{{ .Permalink }}</loc>
         <lastmod>{{ safeHtml ( .Date.Format "2006-01-02T15:04:05-07:00" ) }}</lastmod>{{ with .Sitemap.ChangeFreq }}
         <changefreq>{{ . }}</changefreq>{{ end }}{{ if ge .Sitemap.Priority 0.0 }}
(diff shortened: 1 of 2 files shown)


spf13:
@derekperkins I know you've been waiting a while on this one. Sorry it took me so long to get to it. It looks real nice. Merged as 3c147bd

derekperkins:
No problem. There have been plenty of other fish to fry. :)

spf13:
@derekperkins 

Now that this and the other PR have been merged can you add these new standards to the documentation?

We've gotten a bug report #972 about the sitemap being broken, but the problem isn't the template, but that the user had params.images as a string, here you expect it to have sub fields. 

We also should probably adjust the template to handle allowing users to just make params.images as a slice of strings.

bep:
Have a look at the template_embedded.go that also uses  .Params.images and expect it to be a slice of string.

I suggest a revert of the above commit -- and do a rethinking and doc job of the complete picture.

derekperkins:
There was a discussion started about standardizing metadata like images, but no conclusions were ever reached. I think we should make params.images something that we can expect. Somewhere I believe I submitted a PR that added Images as first class citizens to Node objects, or at least I thought about it. It's been a few months since I've done anything with Hugo.

bep:
Closing this for now. Open a new one when we figure out what we want.

PR #694, opened 2014-12-03
Adding top level .Hugo variable with Version, CommitID & BuildDate

derekperkins:
This originated with the discussion here: http://discuss.gohugo.io/t/template-variables-to-pull-hugo-version-information/419

As Hugo gains popularity and more themes are built, there is going to be needed information about the current Hugo environment to determine what the version supports. I attempted to follow the pattern in `hugolib/site.go` to add `hugolib/hugo.go`, to add a top level .Hugo variable that can be accessed from shortcodes, templates and partials.

As of right now, the version is a string constant at the top of the file that would need to be changed manually with each release, that shouldn't be too onerous. If we wanted to get fancier, maybe it could pull from the current commit / tag. I'm not sure the best way to go with that.

@spf13 I'm not quite sure how to expose that variable to the templating system, but I imagine it shouldn't be difficult.

The first thing I am using the version for is for a meta generator shortcode.
```
{{% hugo-generator %}}

// outputs
<meta name="generator" content="Hugo 0.13" />
```

I still have issues running all the tests on my Windows machine, which makes it a little harder to verify tests before pushing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 
 COMMIT_HASH=`git rev-parse --short HEAD 2>/dev/null`
 BUILD_DATE=`date +%FT%T%z`
-LDFLAGS=-ldflags "-X github.com/spf13/hugo/commands.commitHash ${COMMIT_HASH} -X github.com/spf13/hugo/commands.buildDate ${BUILD_DATE}"
+LDFLAGS=-ldflags "-X github.com/spf13/hugo/hugolib.CommitHash ${COMMIT_HASH} -X github.com/spf13/hugo/hugolib.BuildDate ${BUILD_DATE}"
 
 all: gitinfo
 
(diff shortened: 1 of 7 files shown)


anthonyfok:
Wow, you guys are awesome.  :-)  I was going to suggest this as a "new idea", but decided to search through the forum first, and lo and behold, this great idea has already been brought up, and with working implementation too!

Yes, something like `.Hugo.Version` would be great.  Besides the meta generator line, I was thinking that some kind of version check functionality would allow a theme to simultaneously support multiple versions of Hugo.  :-)

bjornerik:
Merged and fixed.

derekperkins:
@bjornerik What specifically wasn't initializing it properly?

bjornerik:
@derekperkins you could look at the code.

derekperkins:
@bjornerik I did look at the code and I didn't see anything jump out, which is why I asked.

PR #706, opened 2014-12-09
Apply a function over a sequence

halostatue:
> __Note__: The description below is no longer valid. See [this comment](https://github.com/spf13/hugo/pull/706#issuecomment-66401985) for a current correct description of this PR, which is ready but I have some suggested usability improvements.

@derekperkins, I was playing around with the code in #683 tonight and it’s
missing some functionality that I really want—and I mostly have it working, but
there are some incompatibilities with the `delimit` function as written.

Basically, I want to replace this:

```html
{{ with .Params.tags }}
<div class="tags-list">
  <span class="dark-red">Tags</span><span class="decorative-marker">//</span>
  {{ $len := len . }}
  {{ if eq $len 1 }}
    {{ partial "post/tag/link" (index . 0) }}
  {{ else }}
    {{ $last := sub $len 1 }}
    {{ range first $last . }}
      {{ partial "post/tag/link" . }},
    {{ end }}
    {{ partial "post/tag/link" (index . $last) }}
  {{ end }}
</div>
{{ end }}
```

with this:

```html
{{ with .Params.tags }}
<div class="tags-list">
  <span class="dark-red">Tags</span><span class="decorative-marker">//</span>
  {{ delimit (apply (sort .) "partial" "post/tag/link" ".") ", " }}
</div>
{{ end }}
```

This means:

1. Sort .Params.tags.
2. Loop over .Params.tags, replacing each tag with the output of `partial`
   with `"post/tag/link"` as the partial and the tag as the context (`.`).
3. Delimit the resulting objects with `", "`.

The result of `partial` is an HTML template, which is not trivially convertible
to a string, so I’m hitting panics in the delimit code because it tries to
`cast.StringE` each value. If that isn’t done, you can’t easily concatenate the
delimiters to HTML template…and you want it to be `safeHtml` in this case (but
I could probably do that with an appropriate `| safeHtml` at the end of the
`delimit`).

There’s a few things I don’t like about this particular syntax, but it offers
some intriguing capabilities if we can get it working (I *do* have `apply`
working in this branch, but not with your functions underneath).

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -258,11 +258,75 @@ Removes any trailing newline characters. Useful in a pipeline to remove newlines
 e.g., `{{chomp "<p>Blockhead</p>\n"` → `"<p>Blockhead</p>"`
 
 ### highlight
-Take a string of code and a language, uses Pygments to return the syntax
-highlighted code in HTML. Used in the [highlight
-shortcode](/extras/highlighting).
+Take a string of code and a language, uses Pygments to return the syntax highlighted code in HTML. Used in the [highlight shortcode](/extras/highlighting).
 
 ### ref, relref
 Looks up a content page by relative path or logical name to return the permalink (`ref`) or relative permalink (`relref`). Requires a Node or Page object (usually satisfied with `.`). Used in the [`ref` and `relref` shortcodes]({{% ref "extras/crossreferences.md" %}}).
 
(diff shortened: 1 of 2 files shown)


halostatue:
As soon as this is green, this is good to go.

bjornerik:
Merged in ae9c25a. This is very handy, looking forward to clean ups some of my templates with this.

Issue #716, opened 2014-12-15
A way to access content after <!--more--> marker


ghost:
Currently one can get either the summary (`.Summary`) or all the content (`.Content`), but there's no way to access only the content after the  marker, through a variable like `.MoreContent`, for example. In my case the summary and the rest are completely separate entities.

anthonyfok:
Hi @opennota,
Is there a reason why you closed this issue before it was fixed?  :-)

opennota:
@anthonyfok I don't need it anymore, and it doesn't look like anyone else is interested.

abirger:
@opennota, actually I would very much like this feature. I missed #716 and #718 in the first place, and I am glad that I've seen them now.

anthonyfok:
Let's reopen this issue then!  ;-)

bep:
+1

Dr-Terrible:
+1

bep:
@Dr-Terrible for the BF related issues, see https://github.com/russross/blackfriday/issues/166

Dr-Terrible:
@bep Thank you, your idea about separate entities makes more sense. I didn't think about the ToC either.

matthewdolman:
+1

I would like this.  In fact the more that we can define sections in our content the better.

matthewdolman:
If anyone is looking for a quick workaround for now you can do this:

{{ $body := replace .Content .Summary "" }}
{{ $body | safeHTML }}

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

f3l1x:
I like an idea. :+1: What can I do for it?

edio:
[...]

PR #718, opened 2014-12-15
Allow to access the rest of content after summary

opennota:
A quick and dirty PR for #716

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -56,19 +56,20 @@ type Page struct {
 	Tmpl            tpl.Template
 	Markup          string
 
-	extension           string
-	contentType         string
-	renderable          bool
-	layout              string
-	linkTitle           string
-	frontmatter         []byte
-	rawContent          []byte
-	contentShortCodes   map[string]string
-	plain               string // TODO should be []byte
-	plainWords          []string
-	plainInit           sync.Once
-	renderingConfig     *helpers.Blackfriday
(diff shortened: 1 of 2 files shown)


spf13:
@opennota 

I hate to ask this, but HEAD has drifted quite a bit since this PR. Now that v0.13 has shipped I'd love to merge this in. Would you be able to bring it up to date and repush?

anthonyfok:
Hi @opennota, would you like to follow up on what @spf13?  Because hugolib/page.go has seen some changes since you made your original Pull Request, it cannot yet be merged directly and need a little more work on your part.  Could you please do that so that @spf13 can merge it for v0.14?

Many thanks!

bep:
@opennota I think many would want this. Would be nice if you could revisit this.

opennota:
Rebased.

bep:
Great.

A couple of comments.

1) The above only supports "manually summaries". Users mix and match these and would expect to use the same template constructs for both.
2) There is still some "double work"  going on in here. Some of it may be out of scope of this PR (I would assume that, in most cases, one would use either `Content` or `ContentAfterSummary` and not both, so `Content` and `TableOfContents` could be methods that could be lazily evaulated) -- I would assume that  `ContentAfterSummary` is called several times, so cache the result, and assume that it will called from multiple go-routines at once. But I have not thought this all through, there may be better ways.

spf13:
CLA not signed. Closing.

PR #743, opened 2014-12-26
[Docs] Add a note about the plan to switch to Apache License

anthonyfok:
I propose adding the following note to `docs/content/meta/license.md`:

> *Note: There is a plan to switch to [Apache License, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0), for future Hugo releases.  This change would only apply forward.  Existing versions of Hugo will remain released under the SimPL-2.0 license.  For more information, see [the "Switching to Apache 2 license" thread in the public forum](http://discuss.gohugo.io/t/switching-to-apache-2-license/173).*

Is it appropriate?  Or is it premature?

### DIFF
--- a/docs/content/meta/license.md
+++ b/docs/content/meta/license.md
@@ -13,6 +13,8 @@ weight: 50
 
 Hugo is released under the Simple Public License.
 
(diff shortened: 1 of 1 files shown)


anthonyfok:
Since no one has raised any objections, I will assume general consent.  ;-)

anthonyfok:
Merged (cherry-picked) as 1348caa6 to master, and as 845c05b6 to v0.12.docs.

PR #755, opened 2014-12-29
StaticRelativePath inserts a relative path to static content

dblezek:
Useful if static content needs to be placed in a deep subdirectory.
Keeps the static directory clean.  Location of static content can be
configured from config.yaml rather than directory structure.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -136,6 +136,7 @@ func InitializeConfig() {
 	viper.SetDefault("FootnoteAnchorPrefix", "")
 	viper.SetDefault("FootnoteReturnLinkContents", "")
 	viper.SetDefault("NewContentEditor", "")
+	viper.SetDefault("StaticRelativePath", "")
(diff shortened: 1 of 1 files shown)


k4rtik:
Sounds useful, but would you like to add some tests as well so it's clearer?

dblezek:
@k4rtik, can you give some guidance on where to put the test?  The change inserts a directory between the publish directory and the static content, but it's only invoked at build time within hugo.  Is there a (system) test that builds a full site in temp directory and then tests it's validity?  This would be a sensible place to test this feature.  I didn't see any tests that exercise the main commands/hugo.go functionality.

k4rtik:
@dblezek I just manually tested out this change, this also affects the static assets from the theme and breaks the site completely. You will need to reconsider the change.

Agreed, I realized there is no system test for `hugo` command after you mentioned, but since this change introduces a new config parameter, you might want to document it with an example.

bjornerik:
I quess this is meant for people with full control of their assets (either writing their own theme, or not using one). Regular users will not notice.

PR #760, opened 2015-01-01
always use http scheme for localhost

### LINKED ISSUE #758: https protocol used with localhost
As @andrelima [mentioned here](https://github.com/spf13/hugo/issues/544#issuecomment-58817004):

`{{ .Site.BaseUrl }}` results in `https://localhost:1313/` if the baseurl defined in config uses https.

In this case it probably makes sense to override the protocol to http along with the host.

A work around (in some cases) is just to use relative paths instead of `{{ .Site.BaseUrl }}`.

I just upgraded hugo to latest (fb1b795d59b3c6d5d74df3a69b4026574b6dd5fa) and this issue remains.

nathany:
fixes #758

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -152,6 +152,7 @@ func fixUrl(s string) (string, error) {
 	if serverAppend {
 		if useLocalhost {
 			u.Host = fmt.Sprintf("localhost:%d", serverPort)
+			u.Scheme = "http"
 			return u.String(), nil
 		}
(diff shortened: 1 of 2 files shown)


bjornerik:
Thanks, merged in c6f4b09

dsimmons:
Thanks for this sir :+1:

nathany:
You're most welcome.

PR #765, opened 2015-01-04
Add operator argument to `where` template function

tatsushid:
It allows to use `where` template function like SQL `where` clause.
For example,

    {{ range where .Data.Pages "Type" "!=" "post" }}
        {{ .Content }}
    {{ end }}

Now these operators are implemented:

    =, ==, eq, !=, <>, ne, >=, ge, >, gt, <=, le, <, lt, in, not in

It also fixes `TestWhere` more readable

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -79,6 +79,25 @@ e.g.
        {{ .Content}}
     {{ end }}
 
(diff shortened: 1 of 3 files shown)


bjornerik:
Merged in 0d17ee7.

This looks extremely powerful and useful.

PR #767, opened 2015-01-04
Add config options to disable footnote anchor prefix and/or header ID suffix

jacobwgillespie:
The new config options are

* `DisableFootnoteAnchorPrefix` - bool - default: `false`
* `DisableHeaderIDSuffix` - bool - default: `false`

My site depends on anchor IDs being generated without the suffix, so I added this configuration option to disable the automatic suffix behavior.  I understand the risk of duplicate IDs, so that's why this is defaulting to `false`.

So, if the config file contains:

```toml
DisableHeaderIDSuffix = true
```

Then this:

```markdown
# My Header
```

will generate:

```html
<h1 id="my-header">My Header</h1>
```

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -123,6 +123,8 @@ func InitializeConfig() {
 	viper.SetDefault("BuildDrafts", false)
 	viper.SetDefault("BuildFuture", false)
 	viper.SetDefault("UglyUrls", false)
+	viper.SetDefault("DisableFootnoteAnchorPrefix", false)
+	viper.SetDefault("DisableHeaderIDSuffix", false)
 	viper.SetDefault("Verbose", false)
 	viper.SetDefault("CanonifyUrls", false)
 	viper.SetDefault("Indexes", map[string]string{"tag": "tags", "category": "categories"})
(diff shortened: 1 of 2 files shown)


bjornerik:
Not sure, but these looks like Blackfriday options - and should maybe be grouped with the others. Agree, @halostatue ?

See #744

halostatue:
Agree. Checking for `len(ctx.DocumentId) != 0` should probably be done once, though, not twice as it is now.

halostatue:
I’m also wondering if this is not really one meta-option that controls both of these. (That is, do we need two knobs or just one? I can’t see where you’d just turn off footnote anchor prefixes but not document anchor suffixes.)

jacobwgillespie:
Yeah, that makes more sense - I'm not sure the use case for turning off one or the other.  Let me adapt to be a Blackfriday option for both and I'll push.

jacobwgillespie:
Okay, I renamed it to Blackfriday: documentIDAnchor.  Not exactly happy with that name, but it's what I could come up with at the moment.

Also added some documentation.

jacobwgillespie:
I broke the tests - one sec...

jacobwgillespie:
Okay - renamed it to `plainIdAnchors`, default `false`, which seems more logical.  Should be ready for review.

Issue #766, opened 2015-01-04
.Site.Params empty in custom partial template


philgs:
I create a site and set the theme either through `config.toml` or command line to Hugo. I copy a partial from theme/layouts/partials/sidebar.html to layouts/partials/ and make some changes. As far as I can tell, .Sites.Params is empty.

Here's a specific example, using the "redlounge" theme. First, my `config.toml`:

```toml
title = "Phil's Not-Too-Awesome Website"
description = "A place for (links to) stuff by and about me"
sidebartagline = "This my tag line"
sidebartitle = "philgs"
baseurl = "http://philgs.me"
languageCode = "en-us"
useless1 = "waste"
useless2 = "waste"
useless3 = "waste"
```

`layouts\partials\sidebar.html` looks like this:
```html
<div class="sidebar pure-u-1 pure-u-md-1-4">
  <div class="header">
    {{ partial "sidebarheader.html" . }}

    {{ if .Site.Params.sidebartitle }}<h1 class="brand-title">{{ .Site.Params.sidebartitle }}</h1>{{ end }}
    {{ if .Site.Params.description }}<h2 class="brand-tagline">{{ .Site.Params.description }}</h2>{{ end }}

    <ul class="nav-list">
    {{ $count := 1 }}
    {{ range .Site.Params }}
      <li class="nav-item"><span class="nav-item-separator">{{ $count }}. </span>{{ . }}</a></li>
      {{ $count := add $count 1 }}
    {{ end }}
    </ul>

    <nav class="nav">
      <ul class="nav-list">
      {{ $count := 1 }}
        <li class="nav-item"><span class="nav-item-separator">//</span><a href="{{ .Site.BaseUrl }}">Home</a></li>
        {{ range .Site.Menus.main }}
          <li class="nav-item"><span class="nav-item-separator">{{ $count }}//</span><a href="{{ .Url }}">{{ .Name }}</a></li>
        {{ $count := add $count 1 }}
        {{ end }}
      </ul>
    </nav>

    {{ if .Site.Menus.social }}
    <div class="social-buttons">
      {{ range .Site.Menus.social }}
        {{ if .Url }}
        <a href="{{ .Url }}" target="_blank">{{ .Pre }}</a>
        {{ end }}
      {{ end }}
      
    </div>
    {{ end }}

  </div>
</div>
```

[...]

PR #769, opened 2015-01-05
Update documentation to address issue #766

philgs:
I was confused about how to set variables that are available through .Site.Params ([issue 766](https://github.com/spf13/hugo/issues/766)). Knowing the answer, I can see it now in the documentation, but obviously it wasn't clear to me from the outset. So I updated a couple of pages to try to make it more explicit.


bjornerik:
Merged and should be live soon.

PR #770, opened 2015-01-05
Add example code to Summaries

philgs:
I was initially confused about how to use summaries. The only example code I found in the docs was on the page for list nodes, but that uses `Render "summary"`, which is for views, not an article summary. I thought a little example here might clarify the issue for future users.


bjornerik:
Merged, thanks.

anthonyfok:
Great addition, @philgs!  Your example certainly helped me.  Many thanks!

PR #772, opened 2015-01-07
Fix `echoParam` template function to accept string key name

### LINKED ISSUE #771: echoParam doesn't work as advertised
From documentation:

```
If parameter is set, then echo it.

e.g. {{echoParam .Params "project_url" }}

```
But echoParam takes integer as second argument.

tatsushid:
This changes `echoParam` template function behavior to accept not only
an array or a slice and its index pair but also a map and its key pair.

This also changes the function that float and uint values are treated as
a valid result type of it.

Fix #771

### DIFF
--- a/tpl/template.go
+++ b/tpl/template.go
@@ -785,20 +785,36 @@ func IsSet(a interface{}, key interface{}) bool {
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


bjornerik:
El fantastico! 

Merged in d12f6cd

Issue #776, opened 2015-01-08
"hugo new" gives strange EOF error when an archetype file has no final EOL


ePirat:
When I do `hugo new foo.md` while in my site directory I get the following error:

```
ERROR: 2015/01/08 EOF
```

(One issue here is the error, the other one the not very helpful error message)
By far hugo from all static site tools I already used, has the most horrible learning curve. (Trying since 1 hour to get a super simple site set-up. It doesn't really help that the docs are just very repetitive and contain a lot of codesnippets with sligh variations all over the place, confusing new users completely instead of being helpful)

ePirat:
It seems the error was the default.md archetypes file of my theme. Removed it and it works fine now. (Maybe consider the new theme command not creating a archetypes file at all, or one that does not break things.)

anthonyfok:
Hi @ePirat,

Thank you for reporting this issue.  The solution is to end a final newline (or EOL) to the end of your default.md archetype file of your theme.  More discussion here:

> http://discuss.gohugo.io/t/archetypes-not-properly-working-in-0-12/544

1. Add warnings about this in the Hugo documentation, as several people have run into the same problem already.  (Users of editors like Vim, nano and gedit are immune to this because these editors enforce an EOL at the end of the file by default, but other editors like Emacs don't do that.)
2. Improve the error message.  It is difficult to determine what went wrong with just three characters "`EOF`"
3. Allow archetype files without the final EOL to compile anyway, but do give an appropriate and detailed warning.  (optional, to be discussed)

tatsushid:
This strange EOF seems to be caused at https://github.com/spf13/hugo/blob/master/parser/page.go#L207. I'll try to find out a way to avoid this issue.

bjornerik:
This should be fixed.

anthonyfok:
Many thanks to @tatsushid who fixed this issue with "Pull Request #785: Fix "hugo new" EOF error with an archetype file without the final EOL".

Issue #739, opened 2014-12-24
Doc to help users troubleshoot "Categories with accented characters" Unicode NFC/NFD mismatch on Mac OS X


anthonyfok:
See http://discuss.gohugo.io/t/categories-with-accented-characters/505 for more information.

PR #780, opened 2015-01-09
[Docs] New Troubleshooting section

anthonyfok:
With two entries of frequently encountered or obscured troubles so far:

- "Categories with accented characters" Unicode NFC/NFD mismatch
   on Mac OS X (See #739)
- `hugo new` aborts with cryptic EOF error (See #776)


anthonyfok:
Already committed.  Forgot to close it.  :-)

PR #783, opened 2015-01-09
note in archetype.md to alert that ending carriage return may be nessary to avoid EOF errors.

dkebler:
Already someone has added this to the troubleshooting section but users having this issue might not visit that page thus this quick note in archetype.md itself.  It would have saved me a bunch of time :-).  Hope you can update the docs site now and not wait until next release.


bjornerik:
Merged and live in a second. Thanks.

PR #785, opened 2015-01-10
Fix "hugo new" EOF error with an archetype file without the final EOL

### LINKED ISSUE #776: "hugo new" gives strange EOF error when an archetype file has no final EOL

tatsushid:
This rewrites `extractFrontMatterDelims` function to make it work with
an archetype file without the final EOL and adds more detailed error
messages and comments.

It also removes `matches` and `matches_quick` functions which aren't
called anywhere.

This would fix #776 but it doesn't warn when an archetype file without the final EOL is used. Which is good finishing with or without a warning? I think many text editors don't add the final EOL automatically so such a warning may be a little annoying.

### DIFF
--- a/hugolib/page_test.go
+++ b/hugolib/page_test.go
@@ -447,7 +447,7 @@ func TestDegenerateInvalidFrontMatterShortDelim(t *testing.T) {
 		r   string
 		err string
 	}{
-		{INVALID_FRONT_MATTER_SHORT_DELIM_ENDING, "Unable to read frontmatter at filepos 45: EOF"},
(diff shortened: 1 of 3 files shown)


bjornerik:
Merged in 78e9229 - works great!

PR #792, opened 2015-01-12
Fixed error in tutorial, added site variable docs

jaden:
While following the github pages tutorial I found some issues. These are
the commands I ran that worked.

Added site variables to the docs from the code.


anthonyfok:
Thank you so much for your contribution, @jaden!  Your corrections and additions have been committed, and is now live on http://gohugo.io/ .

PR #800, opened 2015-01-16
Missing . on footer partials

ptomasroos:


anthonyfok:
Thank you for your contribution, @ptomasroos!
Your correction has been committed as a4da526 to master, and as a72423f to v0.12.docs, and is now live on http://gohugo.io/templates/list/.

PR #801, opened 2015-01-16
Add trim and replace template functions

tomhelmer:
When working with muut i needed something that could transform "/article/photoswipe-gallery-hugo/" to "article-photoswipe-gallery-hugo". I could have made a custom function doing exactly this, but I thinj it's better solved with two more generic function used together:

     {{ $muutname := trim ( replace .RelPermalink "/" "-" ) "-" }}

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -276,6 +276,16 @@ Removes any trailing newline characters. Useful in a pipeline to remove newlines
 
 e.g., `{{chomp "<p>Blockhead</p>\n"` → `"<p>Blockhead</p>"`
 
+### trim
+Trim returns a slice of the string with all leading and trailing characters contained in cutset removed.
+
+e.g. `{{ trim "++Batman--" "+-" }}` → "Batman"
+
+### replace
+Replace all occurences of the search string with the replacement string.
+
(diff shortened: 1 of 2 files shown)


bjornerik:
Eventually people will ask to use this with `template.HTML` and not strings only.

See Chomp for an example.

Could you also `go fmt`  the source.

Please amend the commit and do a push -f.

tomhelmer:
Sure, I'm not sure what you mean with `template.HTML`, How would it it be used ?

When I look at chomp I figure the trim function would look like:

    func Trim(a interface{}, b interface{}) (string, error) {
    	a_str, err := cast.ToStringE(a)
    	if err != nil {
    		return "", err
    	}

    	b_str, err := cast.ToStringE(b)
    	if err != nil {
    		return "", err
    	}

      return strings.Trim(a_str, b_str), nil;
    }

bjornerik:
Yes - you are correct. The thing is, in a template, you're never really sure what the type is (int, int32, int64, template.HTML (a strongly typed ... string), string ...) so the  interface{} is the best argument here. If you look at the implementation of cast.ToString you will understand.

The usage is as before, like you say in the documentation.

Great addon btw, I need these.

bjornerik:
Merged.

PR #806, opened 2015-01-20
Fix two minor errors in two embedded templates

anthonyfok:
* template: _internal/_default/opengraph.html:39: unexpected EOF
 * template: _internal/_default/schema.html:15: unexpected {{end}}

Also change the DateTime inside these templates to ISO 8601 format,
and skip <meta itemprop="datePublished"> if `publishdate` is not set.

Perhaps it would be a good idea to expose `func (Time) IsZero`
to our templates?  :-)

Hi @derekperkins,

Could you please verify that the two embedded templates still work correctly with this pull request, to make sure I didn't goof anything up?  Many thanks!

Anthony


anthonyfok:
Merged as edcdb6f4.

derekperkins:
Looks good. Thanks for tweaking that.

anthonyfok:
Thank you @derekperkins!

PR #808, opened 2015-01-20
fix --uglyUrls params

xboston:


anthonyfok:
@xboston, Thank you for your careful proofreading and contribution!

Merged as 932214d7 (master) and edfaf07b (v0.12.docs).

Issue #476, opened 2014-09-04
Data models in Hugo similar to Jekyll and Mixture


isaaki:
Greetings,

Can't say enough good things about Hugo. Many thanks for all the work on it. 

I've yet only fiddled with Hugo but am currently considering it as my primary static site generator. Before doing so, I'm wondering about one core feature: data models. Essentially, is it possible to store and loop through objects stored in JSON/YAML files within views? 

Two existing examples of such a feature can be seen [here with Jekyll](http://jekyllrb.com/docs/datafiles/) and [here with Mixture](http://docs.mixture.io/models).

anthonyfok:
See also Pull Request #748, Feature: GetJson and GetJson in short codes or other layout files by @SchumacherFM

bep:
For docs, see

https://github.com/spf13/hugo/blob/master/docs/content/extras/datafiles.md

PR #748, opened 2014-12-27
Feature: GetJson and GetCsv in short codes or other layout files.

SchumacherFM:
I have finished the integration of getJson and getCsv into hugo as far as I think it fits.

I'm not a hugo (and Go) expert of where to put these functions and correct naming of functions and vars.

I would be glad if you can comment on spotted errors, bugs, bad design, etc. or even help me to improve the code more.

All tests and the new tests are passing. Goimports has reordered some import paths.

I have written a blog post about that new feature: [http://cyrillschumacher.com/2014/12/21/dynamic-pages-with-gohugo.io/](http://cyrillschumacher.com/2014/12/21/dynamic-pages-with-gohugo.io/)

There is not yet any official documentation because I want to get the code right. I think once this feature is merged as a hidden feature into hugo and some people tested it in the wild with no given complaints I will write the docs.

Looking forward for a successful merge :-)

PR #810, opened 2015-01-20
Suggestion for solution #476.

klakegg:
Reads files inside data/ and makes data available in .Site.Data. May need some refactoring.

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -30,6 +30,7 @@ import (
 	"github.com/spf13/cast"
 	"github.com/spf13/hugo/helpers"
 	"github.com/spf13/hugo/hugofs"
+	"github.com/spf13/hugo/parser"
 	"github.com/spf13/hugo/source"
 	"github.com/spf13/hugo/target"
 	"github.com/spf13/hugo/tpl"
@@ -77,6 +78,7 @@ type Site struct {
 	params      map[string]interface{}
 	draftCount  int
 	futureCount int
+	Data        map[string]interface{}
 }
 
 type targetList struct {
@@ -106,6 +108,7 @@ type SiteInfo struct {
 	Permalinks      PermalinkOverrides
 	Params          map[string]interface{}
 	BuildDrafts     bool
+	Data            *map[string]interface{}
 }
 
 // SiteSocial is a place to put social details on a site level. These are the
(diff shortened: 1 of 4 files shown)


bjornerik:
Hei og takk for sist!

No har ein også #748 -- men løysinga di er både enkel og fleksibel mht. format. Det blir kanskje heller eit spørsmål om "globale data" eller data som høyrer ei side til

Kan du fjerne den utkommenterte seksjonen og gjere ein "push -f" (overskrive innsjekkinga di).

To others: Sorry about the nynorsk language.

klakegg:
Verden er liten. ;)

Første innsjekk er overskrevet.

And to the english version...
I thing a "global data" solution is wanted in #476, and it's the most appealing solution to me. A "site data" solution could be to create a special directory inside the data directory.

hugolib/site.go
```diff
@@ -332,9 +373,10 @@ func (s *Site) initialize() (err error) {
 	}
 
 	staticDir := helpers.AbsPathify(viper.GetString("StaticDir") + "/")
+	dataDir := helpers.AbsPathify(viper.GetString("StaticDir") + "/")
```
bjornerik: -> DataDir

bjornerik:
This should work and is simple enough, @spf13 ?

It should support sub-folders, but that's simple to add.

spf13:
Looks really good. It's pretty much exactly like how I would do it (except in needs tests ;) ) . 

About sub directories, where would they appear?    .Site.Data[subdir][filename] seems like the obvious place to me.

bjornerik:
` .Site.Data[subdir][filename]` is clever -- then it's easy to iterate over all data files for a given subdir, but I suggest using path (i.e. Unix style slashes) and not filepath for the subdir key.

@klakegg do you have any time to spare to take a stab at adding support for subfolders? There are examples of using the `filepath.Walk` some other places in the Hugo code base. Note the ERROR logging on symbolic links -- this might be good to copy over, as this seems to confuse users.

Also some tests would be cool!

klakegg:
I think using Filesystem is better because of less duplication (symlinks) and included caching.

Files in data/ are imported before data in subfolders allowing overriding using subfolders.

bjornerik:
Filesystem vs Walk, I don't care.

But overriding in sub-folders will get confusing pretty fast.

* I think people would expect to be able to create more than one level of subdirs and address the data items in a way that matches that directory structure
* Ideally something like this: `/data/sub1/sub2/mydata.json` ==> `{{ sub1.sub2.mydata.mykey }}`

Any thoughts @spf13 ?

klakegg:
If /data/sub1/sub2.json sets key 'mydata.mykey' to 'foo', and you have a file /data/sub1/sub2/mydata.json setting 'mykey' to 'bar'. What to you expect .Site.Data.sub1.sub2.mydata to return?

Btw, now also with test. ;)

bjornerik:
In that case I would expect /sub2/mydata.json's value, but that isn't override as a feature -- I would even consider returning an error for these duplicates so the user can do some renaming.

bjornerik:
Oh, and the failing test is most likely due to Go maps' random iteration order.

klakegg:
The implementation is now changed to support any amount of subfolders. The test is still unstable (not changed) and will be updated if this is the solution we want.

bjornerik:
There is some spring code cleaning to be done here, but:

* It's simple enough after some limited tests, It does what I would expect.
* I might tone down the duplicate check to just log an ERROR and not STOP -- but that's details

So, @spf13 - to sum up (with my special $ syntax):

`/data/my.json$mkey:1234 => {{ .Data.my.mkey }} ==> 1234`
`/data/sub1/my.json$mkey:1234 => {{ .Data.sub1.my.mkey }} ==> 1234`
`/data/sub1.json$mkey:1234 => {{ .Data.sub1.my.mkey }} ==> 1234 ==> boom ERROR`

It wold need documentation and tests, but from a bird's perspective, it looks correct, right?

spf13:
@bjornerik I don't follow the special $ syntax. Does that just mean the files content after it?

I think we are heading in a very good direction. 

I think duplication should just log an ERROR.

bjornerik:
the $ syntax is just there to separate the file (my.json) with the key inside the file (mkey). It's just for this discussion and not part of any API.

spf13:
Ok. I think this looks really good. We should fix the tests, merge it, and add documentation.

bjornerik:
@klakegg could you fix/add some tests and doc?

bjornerik:
@klakegg bump. It isn't that busy in the government :-) If this is going into the 0.13 release it needs docs/tests -- and I'm too lazy/busy doing it myself.

bep:
OK, I need this one, so I'll finish it.

bep:
Replaced by #885

Issue #235, opened 2014-03-20
format string as date


deificx:
I've added an "updated" parameter to some of my posts, it would be nice to format them the same way as I format my date. 

And on that note, in posts it could be nice to write dates in ISO 8601 and have them formatted with day and month name. ie. {{ 2014-03-20 | format "Monday, Jan 2, 2006" }}

anthonyfok:
This issue was recently brought up on the discussion forum:

* Parsing dates in templates: http://discuss.gohugo.io/t/parsing-dates-in-templates/603

tatsushid:
Interesting. I'll try to write it later (tonight)

anthonyfok:
For reference: There has been quite a bit of refactoring since @spf13's reply last March, and the links are no longer accessible.  Here are the new links.

> Should be pretty straightforward to add. See the examples here…
> https://github.com/spf13/hugo/blob/master/template/bundle/template.go#L161

Permalink of the historic file: https://github.com/spf13/hugo/blob/64572d2d60a6bee0cc0b87dcf0a944d88375d964/template/bundle/template.go#L161

Steve was referring to the template `funcMap` defined in `NewTemplate()`.  The same but expanded `funcMap` is now in `init()` in `tpl/template.go`:

 * https://github.com/spf13/hugo/blob/master/tpl/template.go#L1245
    (though I am sure this link will be outdated very soon)

> You’ll probably need this function as well…
> https://github.com/spf13/hugo/blob/master/hugolib/metadata.go#L38

Permalink of the historic file: https://github.com/spf13/hugo/blob/2540d884d8d8566caa1cfb71a23f01a746133581/hugolib/metadata.go#L38

Steve was referring to `func stringToDate(s string) (time.Time, error)` contained therein.  This function has been renamed to `StringToDate()` and now resides in caste.go in "github.com/spf13/cast":

* https://github.com/spf13/cast/blob/master/caste.go#L302

anthonyfok:
@tatsushid wrote:
> Interesting. I'll try to write it later (tonight)

What a pleasant surprise!   :+1:

tatsushid:
Thanks for the reference. It's pretty easy work because of cast library, just I wrote

```go
func DateFormat(layout string, v interface{}) (string, error) {
        t, err := cast.ToTimeE(v)
        if err != nil {
                return "", err
        }
        return t.Format(layout), nil
}
```

If someone is interested in treating an int value as Unix time value and converting it into a string, I'll send PR to cast library

[...]

PR #814, opened 2015-01-21
Add dateFormat template function

tatsushid:
This is the implementation of the proposal at #235 and
http://discuss.gohugo.io/t/parsing-dates-in-templates/603/3


bjornerik:
Merged in 37490ee

anthonyfok:
Hi @tatsushid,
Please remember to send an request for a Slack invite here:
http://discuss.gohugo.io/t/how-about-an-irc-channel/416   :-)

Issue #470, opened 2014-09-03
Support for AsciiDoc


tbrooke:

ollin:
asciidoc +1

0xMF:
asciidoc +1

bjornerik:
asciidoc +1

voxadam:
Don't worry folks, Steve (aka @spf13) has stated publicly[1] that he's diligently working on a solution to bring asciidoc support to Hugo. 

[1] http://discuss.gohugo.io/t/contributing-golang-code/223/3

spf13:
I've made good progress on this. Love seeing the enthusiasm here.

0xMF:
@spf13 thanks! Liking Hugo thus far, making and using a theme seems easy and intuitive (so far) ;)

bguiz:
+1 for asciidoc
(yet another one)
Is this likely to come in in v0.13 or can we be waiting for this in v0.14?

spf13:
I really hope in v0.13 even if it's just experimental.

--
Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

anthonyfok:
Hi @tbrooke, @ejemba, @voxadam, @txchen, @ollin, @0xMF and @bguiz,

An experimental support for AsciiDoc has been added in Pull Request #826.
Since @VonC's excellent https://github.com/VonC/asciidocgo needs just a little bit more work before we could use it, PR #826 searches for the original [Asciidoctor](http://asciidoctor.org/) (written in Ruby) and pipes content through it, similar to how Hugo supports reStructuredText.

Caveat: Since Asciidoctor is not written in Go, it would be slow, but it works!  :-)

There may be some quirks and idiosyncrasies, so your feedback is very welcome and very much needed.  So, please try to build from source with #826 applied, and start using it in your projects, and give us feedback on what works and what don't!

Many thanks!

bguiz:
@anthonyfok Thanks for the great work. My trivial tests so far have been OK, but it looks like @bep has done a rather comprehensive one already: https://github.com/spf13/hugo/pull/826#issuecomment-71316363

@VonC I was going to say the same as what @bep said here: https://github.com/VonC/asciidocgo/issues/1
As the last commit was 7-8 months ago, it looks like it is pretty stagnant, but I do not want to jump to conclusions. Are you still working on this, or would you like it to be forked?

VonC:
[...]

PR #823, opened 2015-01-23
Experimental AsciiDoc support with external helpers

anthonyfok:
See #470

 * Based on existing support for reStructuredText files

 * Handles content files with extensions `.asciidoc` and `.ad`

 * Pipes content through `asciidoctor --safe -`.
   If `asciidoctor` is not installed, then `asciidoc --safe -`.

 * To make sure `asciidoctor` or `asciidoc` is found, after adding
   a piece of AsciiDoc content, run `hugo` with the `-v` flag
   and look for this message:

        INFO: 2015/01/23 Rendering with /usr/bin/asciidoctor ...

Caveats:

 * The final "Last updated" timestamp is currently not stripped.

 * When `hugo` is run with `-v`, you may see a lot of these messages

        INFO: 2015/01/23 Rendering with /usr/bin/asciidoctor ...

   if you have lots of `*.ad` or `*.asciidoc` files.

 * Some versions of `asciidoc` may have trouble with its safe mode.
   To test if you are affected, try this:

        $ echo "Hello" | asciidoc --safe -
        asciidoc: ERROR: unsafe: ifeval invalid
        asciidoc: FAILED: ifeval invalid safe document

   If so, I recommend that you install `asciidoctor` instead.

Feedback and patches welcome!

Ideally, we should be using https://github.com/VonC/asciidocgo,
@VonC's wonderful Go implementation of Asciidoctor.  However,
there is still a bit of work needed for asciidocgo to expose
its API so that Hugo can actually use it.

Until then, hope this "experimental AsciiDoc support through external
helpers" can serve as a stopgap solution for our community. :-)

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -180,6 +180,8 @@ func RenderBytesWithTOC(ctx RenderingContext) []byte {
 		return MarkdownRenderWithTOC(ctx)
 	case "markdown":
 		return MarkdownRenderWithTOC(ctx)
+	case "asciidoc":
+		return []byte(GetAsciidocContent(ctx.Content))
 	case "rst":
 		return []byte(GetRstContent(ctx.Content))
 	}
@@ -191,6 +193,8 @@ func RenderBytes(ctx RenderingContext) []byte {
 		return MarkdownRender(ctx)
 	case "markdown":
 		return MarkdownRender(ctx)
+	case "asciidoc":
+		return []byte(GetAsciidocContent(ctx.Content))
 	case "rst":
 		return []byte(GetRstContent(ctx.Content))
 	}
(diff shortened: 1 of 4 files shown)


bjornerik:
Cool.

But whats missing from asciidocgo that prevent us from using it?

bjornerik:
On an added note - I think we should wait and/or make it so we can use the Go implementation. One thing is to use a Python lib or code highlighting -- but dragging some slow Ruby into the mix for the vital rendering part ...

bjornerik:
Had a quick look at the go library -- and it looks very unfinished.

anthonyfok:
> Had a quick look at the go library -- and it looks very unfinished.

Indeed.  The fork https://github.com/ciarand/asciidocgo by @ciarand has gone a bit further, but still not in a readily usable state.  When a first look at it a couple months ago, it appears that @VonC has completed all the internals, but the actual external API is... not written (!) at the moment?  Looks very promising, but I don't think it would be ready for Hugo v0.13.  Maybe v0.14.

Yes, Asciidoctor will be quite a bit slower than the rest of Hugo, but I suppose it would satisfy the many AsciiDoc fans who want to start using Hugo with AsciiDoc today.  ;-)  We just need a BIG DISCLAIMER in the documentation that this AsciiDoc feature is experimental and s-l-o-w, and improvement in speed are expected in future versions once asciidocgo is ready, hopefully a few months later.  :-)

ciarand:
Yeah, unfortunately I haven't had (and probably won't have) time to fix it up
for at least the next 6 months.

anthonyfok:
Thank you for your input @ciarand, and for spearheading the effort to finish the last mile for asciidocgo.  Perhaps fellow Gophers would be interested in rising to the challenge to make asciidocgo complete?  :-)

By the way, apologies to fellow Hugo developers: I have mistakenly pushed this topic branch to the main hugo git repository, whereas I was supposed to push it to my fork and then make the pull request there.  (Oops!)  Sorry for my oversight.

anthonyfok:
Notes for possible solutions to the caveats:
* [Asciidoctor] Disabling Last Updated text in footer: http://mrhaki.blogspot.ca/2014/11/awesome-asciidoctor-disable-last.html
* [AsciiDoc] Re: One line input file fails to build with `--safe` (apparently due to unsafe ifeval) https://groups.google.com/forum/#!msg/asciidoc/yEr5PqHm4-o/GX9BIdgBuGQJ

bjornerik:
As to the topic branch: Please delete it on spf13, and push it to your own repo. That would also mean creating a new PR, but that's OK.

anthonyfok:
@bjornerik: Good idea, thanks!  I'll do that right away.

PR #825, opened 2015-01-24
Handle empty front matter in archetype.

trombonehero:
If an archetype has deliberately empty front matter (e.g., to suppress
generation of the 'draft' field or to force a particular front matter type
instead of the default TOML), we should handle it gracefully rather than
panic ("assignment to entry in nil map").


bjornerik:
Makes sense, merged in bd1b019

PR #828, opened 2015-01-24
Fix for page.GetParam() for JSON and TOML maps

anthonyfok:
Setting per-page Blackfriday angledQuotes did not work
with TOML or JSON front matter, but it does work with YAML.

It turns out that page.Params("blackfriday") returns
type map[interface{}]interface{} for YAML, but
type map[string]interface{} for JSON and TOML.

This patch updates page.GetParam() to catch the latter,
with an error message if page.GetParam() does not recognize
a type.  A test is also added.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"github.com/spf13/hugo/helpers"
 	"github.com/spf13/hugo/parser"
+	"reflect"
 
 	"github.com/spf13/cast"
 	"github.com/spf13/hugo/hugofs"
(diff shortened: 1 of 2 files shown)


bjornerik:
Merged in 803865f

Thanks for cleaning up my "mess"

anthonyfok:
@bjornerik, don't worry, I was actually trying to add a `fractions = false` option to scratch my own itch in #829, which I wouldn't have been able to do so easily without your pioneering work with `angledQuotes`.

Besides, who would have guessed that the JSON and TOML parsers return such different map structure from the YAML parser?

bjornerik:
Yea, well -- we have had the same issues in different wrapping before ...

PR #833, opened 2015-01-25
Add site-wide/per-page [blackfriday] `extensions` option

naoina:
Currently, Hugo cannot use other Blackfriday's extensions like `EXTENSION_HARD_LINE_BREAK` without code edit.

I have added feature of `extensions` option to use additional Blackfriday's extensions without code edit.

For example, if you want to use `EXTENSION_HARD_LINE_BREAK` extension, add following option to `config.toml`. (or to a front matter of a single page)

```
[blackfriday]
  extensions = ["hard_line_break"]
```

The specified extensions add to Hugo's default setting rather than overrides.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -136,7 +136,7 @@ func InitializeConfig() {
 	viper.SetDefault("FootnoteAnchorPrefix", "")
 	viper.SetDefault("FootnoteReturnLinkContents", "")
 	viper.SetDefault("NewContentEditor", "")
-	viper.SetDefault("Blackfriday", map[string]bool{"angledQuotes": false, "fractions": true, "plainIdAnchors": false})
+	viper.SetDefault("Blackfriday", new(helpers.Blackfriday))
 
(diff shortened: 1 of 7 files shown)


bjornerik:
Looks good, one objection:

There may be inconsistency in the naming used in Hugo, but `hard_line_break` (i.e. underscore) is not used.

I suggest going with `hardLineBreak` -- even if it's further from the Blackfriday name (which I believe they also regretted -- as it's not a Go thing with all upper case constants).

naoina:
That's true. I've fixed and pushed it.

bjornerik:
Merged in 407e80a

Thanks for this solid contribution.

anthonyfok:
Thank you so much, @naoina!  You make Hugo's support for Blackfriday configuration complete!

If you haven't yet, it would be wonderful if you could take some time to look at the thread at http://discuss.gohugo.io/t/switching-to-apache-2-license/173 and add your sign-off if you agree.  :-)

Thanks again!  ありがとうございます！

PR #835, opened 2015-01-25
Move s.prepUrl() before prepending BaseUrl.

trombonehero:
When we have an absolute menu url specified in the config file
(e.g., `menu: { main: { name: "News", url: "/news" } }`),
its menu entry is generated by prefixing it with the BaseUrl.
The result is then run through prepUrl(), which uses helpers.Urlize to
convert urls such as 'My First Link' to 'my-first-link'.

The behaviour is backwards: we do not want to run helpers.Urlize on the
BaseUrl, only on the absolute component. Currently, a BaseUrl such as
'http://my.edu/ENG101' will be converted to 'http://my.edu/eng101',
resulting in broken links in all of my menus.

This commit switches the URL prep and BaseUrl prepending actions around. I
would argue that these URLs shouldn't be run through prepUrl anyway
because the site developer has specified them explicitly in a config file
and might be surprised for, e.g., URLs to change case, but that's another
commit for another time.

### DIFF
--- a/hugolib/menu_test.go
+++ b/hugolib/menu_test.go
@@ -283,7 +283,7 @@ func doTestMenuWithUnicodeUrls(t *testing.T, canonifyUrls, uglyUrls bool) {
 	expectedBase := "/%D0%BD%D0%BE%D0%B2%D0%BE%D1%81%D1%82%D0%B8-%D0%BF%D1%80%D0%BE%D0%B5%D0%BA%D1%82%D0%B0"
 
 	if !canonifyUrls {
-		expectedBase = "/zoo" + expectedBase
+		expectedBase = "/Zoo" + expectedBase
 	}
 
 	var expected string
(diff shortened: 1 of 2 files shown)


bjornerik:
There are failing tests.

And I don't think this is relevant after #822 - the relative urls will stay relative until replaced by the "canocalizer" (if that option is enabled).

trombonehero:
Actually, I think that #822 exacerbated the problem: I have more broken links (where BaseUrl is made lower-case when it shouldn't be) after commit 1b42dc57 than before.

I see the test failures now: sorry, I didn't know how to run the test suite before. I'll investigate and see if I can come up with a better patch.

trombonehero:
Sorry, the test failures came from a typo: I meant to assign to `menuEntryUrl` but actually assigned to `menuEntry.Url`.

bjornerik:
I will look into it, but it sounds strange, as 8ee2630cda338cfc568ad8765c4f2d8084271c46 was mostly about keeping the menu urls relative.

bjornerik:
OK, I see it now. Thanks for this.

trombonehero:
No problem, thanks for not closing the pull request right away when I committed a typo that caused tests to fail!

bjornerik:
Merged in e72ce89

PR #847, opened 2015-01-27
Added ability to exec arbitrary commands from templates.

aliafshar:


spf13:
I'm still worried about this. I'm punting on this until after the v0.13 release is out. I'm thinking more of having a few dedicated exec based functions... like "grep"... but then concerned about how it would work with systems like windows which don't have them available.

yml:
@spf13 I understand the security concern but I think that a reasonable middleground would be to use a whitelist coming from the [datafiles](http://gohugo.io/extras/datafiles/) to restrict the binaries that can be called.

Does it sounds acceptable ?

bep:
@yml the datafiles can also be set in themes. So any white list would have to live in global config.

yml:
@bep good call. Would checking the `exec_whitelist` in the global config be enough to let this feature get in.

moorereason:
Closing this PR as we won't be merging, as is.  Lacks safeguards.  The security issues needs more thought and discussion.

PR #849, opened 2015-01-28
Version uses binary name instead of hugo

jaden:
On Windows the binary name is hugo.exe and running hugo version results in
this error:
    GetFileAttributesEx D:\Dev\Go\gopath\bin\hugo: The system cannot find the file specified.

This fixes that error and allows the binary name to be something other
than hugo on any OS.


bjornerik:
Merged.

PR #850, opened 2015-01-28
ignore root path, no need strip traling slash

fundon:
Sometimes I just want use that `{{ "/" | SafeUrl }}`, but in this case will generate `//` path, 
so we need ignore root path.

### DIFF
--- a/helpers/url.go
+++ b/helpers/url.go
@@ -15,11 +15,12 @@ package helpers
 
 import (
 	"fmt"
-	"github.com/PuerkitoBio/purell"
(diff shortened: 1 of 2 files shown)


bjornerik:
Merged.

bjornerik:
Oh, and thanks for the contribution.

The license change in Hugo is dragging out in time, so could you also sign off on

discuss.gohugo.io/t/switching-to-apache-2-license/173/32

fundon:
@bjornerik done.

PR #851, opened 2015-01-29
Use a regular expression in replaceShortcodeTokens

lotrfan:
The current shortcode implementation currently fails with things like

``` markdown
    [Article1]: {{% relref "article1.md" %}}
    [Article2]: {{% relref "article2.md" %}}

    This is a link to [Article2]. I might [want to use it more than
    once][Article2]. Here's a link to [Article1].
```

The expected result would be something like
``` html
This is a link to <a href="article2/">Article2</a>.
I might <a href="article2/">want to use it more than once</a>.
Here's a link to <a href="article1/">Article1</a>.
```

The current implementation, however, fails with an error along the lines of
`illegal state in content; shortcode token #2 is missing or out of order`.

With this implementation, all references to a given shortcode are substituted in any order,
which gives the expected result above.

### DIFF
--- a/hugolib/handler_page.go
+++ b/hugolib/handler_page.go
@@ -60,7 +60,7 @@ func (h markdownHandler) PageConvert(p *Page, t tpl.Template) HandledResult {
 	tmpContent, tmpTableOfContents := helpers.ExtractTOC(p.renderContent(helpers.RemoveSummaryDivider(p.rawContent)))
 
 	if len(p.contentShortCodes) > 0 {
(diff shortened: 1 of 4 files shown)


bep:
I will have a look at this, but could you squash your commit into one?

lotrfan:
I squashed into 2... the first is the main change; the second updates the tests to use `string`s instead of `[]byte`s, which is not really related to the first.

bep:
Merged in 35684e8

Thanks for this quality contribution. I had to check the performance implications of this (got a little skeptical when I saw "regexp", but it looks like the Go people have implemented it right - only a slight penalty for large data sets with lots of shortcodes).

PR #856, opened 2015-01-29
Fix for issue 839 and 490 on Windows

jaden:
The paths were seen as changed but not static because of the backslashes in
ev.Name. Once the backslashes were added, I discovered that the JSON
sent to livereload was invalid and failed to work because it had backslashes.

Hence the code to replace the backslashes from the path to make them work
in JSON and for the URL.

With this fix, changes to a stylesheet are shown on the page, and if it's a
single file that changed, it's reflected in the browser without reloading the whole
page.


helpers/path.go
```diff
@@ -174,9 +174,17 @@ func AbsPathify(inPath string) string {
 	return filepath.Clean(filepath.Join(viper.GetString("WorkingDir"), inPath))
 }
 
+func GetStaticDirPath() string {
+	return AbsPathify(viper.GetString("StaticDir"))
+}
+
+func GetThemesDirPath() string {
+	return AbsPathify("themes"+string(os.PathSeparator)+viper.GetString("theme")) + string(os.PathSeparator) + "static" + string(os.PathSeparator)
```
tatsushid: How about using `filepath.Join` (http://golang.org/pkg/path/filepath/#Join) instead of using raw `os.PathSeparator`? I think it makes the line more readable

jaden: Great idea. I'll update it.

jaden:
I still haven't quite figured out how to keep my dev branch in sync with upstream. Instead of merging, I think I should have done a rebase.

PR #858, opened 2015-01-29
Fix for issue 839 and 490 on Windows

jaden:
The paths were seen as changed but not static because of the backslashes in ev.Name. Once the backslashes were added, I discovered that the JSON sent to livereload was invalid and failed to work because it had backslashes.

Hence the code to replace the backslashes from the path to make them work in JSON and for the URL.

With this fix, changes to a stylesheet are shown on the page, and if it's a single file that changed, it's reflected in the browser without reloading the whole page.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -358,7 +358,7 @@ func NewWatcher(port int) error {
 						continue
 					}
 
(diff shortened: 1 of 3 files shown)


bep:
Merged in 523f38a

Thanks, it's hard to discover these issues from Linux ...

karland:
Works beautifully. Thanks.

PR #861, opened 2015-01-30
Use .Date.IsZero to skip unset date in embedded templates

anthonyfok:
Use `{{ if not .Date.IsZero }}` to print dates only when they are defined.  This is to avoid things like

    <lastBuildDate>Mon, 01 Jan 0001 00:00:00 +0000</lastBuildDate>

and

    <lastmod>0001-01-01T00:00:00+00:00</lastmod>

showing up in index.xml (RSS) and sitemap.xml.

Pipe dates with ±hh:mm time zone through `safeHtml` to prevent the `+` sign from turning into `&#43;`.

Also make some shuffling to avoid blank lines in the output.


bep:
Looks good, merged in 3d60955

PR #864, opened 2015-01-31
Print template parsing errors to aid troubleshooting

### LINKED ISSUE #316: Too little information about template problems
Today I was porting an existing theme over to use with Hugo, and when I was finished, or so I thought, and tried to compile I get an error stating `ERROR: 2014/06/02 Rendering error: html/template: "theme/_default/single.html" is an incomplete template`.

I am sorry, but this doesn't help me in any way; is this a default by the Golang templating? Because if not, I would love to know where exactly the error occured, so I could at least narrow it down.

anthonyfok:
Added a new Template.PrintErrors() function call,
used in hugolib/site.go#Process() so it does not clutter
up `go test -v ./...` results.

Special thanks to @tatsushid for mapping out the call trace
which makes it a lot easier to find the appropriate places
to place the Template.PrintErrors() call.

Fixes #316

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -269,6 +269,7 @@ func (s *Site) Process() (err error) {
 		return
 	}
 	s.prepTemplates()
+	s.Tmpl.PrintErrors()
 	s.timerStep("initialize & template prep")
 	if err = s.CreatePages(); err != nil {
 		return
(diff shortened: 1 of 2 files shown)


bep:
Merged in c507809

anthonyfok:
@bep Cool, many thanks!  :-)

PR #868, opened 2015-01-31
Fix defaults for Blackfriday

### LINKED ISSUE #866: Fix defaults for Blackfriday
A recent refactoring of the Blackfriday options seems to have messed up the default settings (fractions is turned off by default, should be on). Others?

naoina:
fixes #866

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -199,14 +199,20 @@ func (p *Page) getRenderingConfig() *helpers.Blackfriday {
 		pageParam := p.GetParam("blackfriday")
 		siteParam := viper.GetStringMap("blackfriday")
 
+		combinedParam := make(map[string]interface{})
+
+		for k, v := range siteParam {
+			combinedParam[k] = v
+		}
+
 		if pageParam != nil {
 			pageConfig := cast.ToStringMap(pageParam)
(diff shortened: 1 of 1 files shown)


bep:
It doesn't work.

I still have to add `fractions = true` to get fractions.

anthonyfok:
Hi @bep (oh, new GitHub handle!),

Actually, even before @naoina's refactoring, the old code

    viper.SetDefault("Blackfriday", map[string]bool{
        "angledQuotes": false,
        "fractions": true,
        "plainIdAnchors": false})

as found in eb686352b1785844778e1cba8c69b85079e3044f *did not* work for me either,
and that is why I had to push a follow-up commit f6c0b472e9f54fa3e0d2695a8cffe343f9112244 to workaround that, but I do feel that it was a kludge and not the real solution.

Apparently, for some reasons, `viper.SetDefault()` isn't working for "Blackfriday"... not working for maps, perhaps?

bep:
Aha, I see. OK. 

(yes, new GitHub handle - bep is my initials and my first choice for usernames. It was taken back then, it was taken now -- but was a dusty account with no activity, so I asked GitHub to delete that account).

PR #883, opened 2015-02-08
[commands/new.go] Update theme.toml etc.

anthonyfok:
- Add copyright years and author to the top of the file
- Write the current year from time.Now() to LICENSE.md
- Update theme.toml template to match that listed at https://github.com/spf13/hugoThemes/blob/master/README.md#themetoml
- Correct comment regarding `os.MkdirAll(p, 0777)`
- In createConfig(), split the `map[string]string` definition into multiple lines to facilitate future expansion.  Also add a trailing slash to sample "baseurl" definition.

----

### Note about the `theme.toml` template update:

Initially, I experimented with expanding the original `map[string]interface{}`:

```go
	in := map[string]interface{}{
		"name":        strings.Title(helpers.MakeTitle(filepath.Base(inpath))),
		"license":     "MIT",
		"licenselink": "",
		"description": "",
		"homepage":    "",
		"tags":        []string{"", ""},
		"features":    []string{"", ""},
		"author": map[string]interface{}{
			"name":     "",
			"homepage": "",
		},
		"original": map[string]interface{}{
			"author":   "",
			"homepage": "",
			"repo":     "",
		},
	}
```

However, since the map keys’ order isn't preserved, I thought it would be better to use `struct` instead.

```go
	in := struct {
		Name        string   `toml:"name"`
		License     string   `toml:"license"`
		LicenseLink string   `toml:"licenselink"`
		Description string   `toml:"description"`
		HomePage    string   `toml:"homepage"`
		Tags        []string `toml:"tags"`
		Features    []string `toml:"features"`
		Author      struct {
			Name     string `toml:"name"`
			HomePage string `toml:"homepage"`
		} `toml:"author"`
		Original struct {
			Name     string `toml:"name"`
			HomePage string `toml:"homepage"`
			Repo     string `toml:"repo"`
		} `toml:"original"`
	}{
		Name:     strings.Title(helpers.MakeTitle(filepath.Base(inpath))),
		License:  "MIT",
		Tags:     []string{"", ""},
		Features: []string{"", ""},
	}
```

Struct is less expensive than map, though the size of the `hugo` executable binary does increase by a little, around 10KB?

And, finally, to add a comment, I experimented with to this:

```go
	by = bytes.Replace(by, []byte("[original]"), []byte("# If porting an existing theme\n[original]"), 1)
```

But then I thought, as the `hugo` binary itself (as of v0.13-DEV) does not read any values from `theme.toml`, and since we do not allow `theme.yaml` or `theme.json`, why all the trouble?  Instead, the most efficient and flexible way to create `theme.toml` would be writing it out directly, the very same way we create `LICENSE.md`, and that is what you see in this Pull Request.  The `hugo` binary actually shrinks by 128 bytes (in Linux) after this!  :-)

But yes, we can always go back to using `struct` or `map` if necessary in a future release, hence the code blocks in this comment—for copy-and-paste.  ;-)

### DIFF
--- a/commands/new.go
+++ b/commands/new.go
@@ -1,3 +1,5 @@
+// Copyright © 2014-2015 Steve Francia <spf@spf13.com>.
+//
 // Licensed under the Simple Public License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -16,6 +18,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/hugo/create"
@@ -72,7 +75,7 @@ as you see fit.
 	Run: NewTheme,
 }
 
-//NewContent adds new content to a Hugo site.
+// NewContent adds new content to a Hugo site.
 func NewContent(cmd *cobra.Command, args []string) {
 	InitializeConfig()
 
@@ -135,7 +138,7 @@ func NewSite(cmd *cobra.Command, args []string) {
 	createConfig(createpath, configFormat)
 }
 
-//NewTheme creates a new Hugo theme.
+// NewTheme creates a new Hugo theme.
 func NewTheme(cmd *cobra.Command, args []string) {
 	InitializeConfig()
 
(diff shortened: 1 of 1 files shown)


anthonyfok:
Looks like there is no objection, so I went ahead and merged it.  :grin:

bep:
Clever strategy! -:)

PR #897, opened 2015-02-14
Fixed invalid link to partials in templates

ifraixedes:
I've found that those don't load the content in the website, but after I've seen that they are alias, however the alias work for "page variables" link but not for those.


spf13:
Merged as 366f991

PR #898, opened 2015-02-15
Add optional post formatting by using github.com/yosssi/gohtml.

pcdummy:
This pull adds the boolean option `PostFormat` to viper.

Pages will be formated by using [github.com/yosssi/gohtml](https://github.com/yosssi/gohtml) if you set `PostFormat` to `true`.

What do you think about it?
I also think about making this configureable per page, what do you mean?

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -42,6 +42,7 @@ import (
 	jww "github.com/spf13/jwalterweatherman"
 	"github.com/spf13/nitro"
 	"github.com/spf13/viper"
+	"github.com/yosssi/gohtml"
 )
 
 var _ = transform.AbsURL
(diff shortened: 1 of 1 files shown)


spf13:
It's an interesting idea. I think many people would probably prefer to actually shrink the HTML instead of prettifying it.

I would just make it a sitewide option. I'm a bit worried about how much this would add to rendering time and Hugo's binary size as well as it's completely superficial.

pcdummy:
Thanks for your response @spf13, you'r right its not that important for a bigger binary and the slowdown!

I'll think about minifing once i get the time to do so :) The goal was/is for me to remove empty lines,
shrinking will do that to. :)

Issue #648, opened 2014-11-17
Feature: update-timestamp command


jeffreyrogers:
I'm new to Hugo, but have really been enjoying how simple everything is. It seems to work much better with my development workflow than any other static site generator I've tried.

One feature I find myself wanting is the ability to update the timestamp on drafts. Often I'll start a new draft of something and won't finish it until much later, at which point the original timestamp is significantly out of date. Of course, I can edit them by hand, but I think a command like `hugo update-timestamp post/file-to-update.md` would be simpler/less error prone. Basically this command would just scan the metadata and replace the old date with a current one.

If this is something others feel would be worthwhile to have as well, then I can work on it. I haven't used Go before but it looks like it would be pretty easy to pick up and I think this change is relatively self-contained.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

PR #899, opened 2015-02-16
add publish command

mohae:
This is related to issue #648.

A `publish` command was added to hugo:

    hugo publish path\to\content

If the specified content is a draft, `draft` will be set to false. The `date` will be updated to time.Now() and uses RFC3339 format, consistent with the way the `new` command set's the `date. This does change the order of settings within the FrontMatter, the updated FrontMatter is written out in the same order that it exists on the file.

If the specified content is not a draft, nothing will be done.

Currently, it uses the path specified, e.g. `content/post/test.md`. If the path of the content, relative to the `content` directory is desired, please let me know.

No additional flags are supported.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -74,6 +74,7 @@ func AddCommands() {
 	HugoCmd.AddCommand(convertCmd)
 	HugoCmd.AddCommand(newCmd)
 	HugoCmd.AddCommand(listCmd)
+	HugoCmd.AddCommand(publishCmd)
 }
 
 //Initializes flags
--- a/commands/publish.go
+++ b/commands/publish.go
@@ -1,0 +1,148 @@
+// Copyright © 2013 Steve Francia <spf@spf13.com>.
+//
+// Licensed under the Simple Public License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://opensource.org/licenses/Simple-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


spf13:
I like the idea, but I think "publish" should actually publish it ... like move the files to a server. 

We have a different situation from systems like wordpress where removing the draft flag == publishing to the world. 

I'm not sure what verb we should use here, but we need something different here. I haven't tested it yet, but browsed the code which looks like a solid approach. 

Any ideas on a better action name?

bep:
I should be the opposite of draft;

finalize, conclude, complete

mohae:
Good point about the name of the command. I'm not sure what a more appropriate verb would be either.

mohae:
I haven't been able to come up with any verbs better than the ones that @bep suggested. But I started wondering if there should also be support for setting content as draft? 

Should this functionality be bikeshed a bit more on the forums?

Other notes:  
* There's a bug in one of my error outputs, it uses Println instead of Printf.
* When I change the command name I will also add code to handle the date setting not existing in the FrontMatter. It currently isn't handled and I'd feel better if it were to help make the code more bullet-proof. This shouldn't happen, but things often happen that shouldn't.

spf13:
I think there should only be a command to undraft a post ( at least to start ). 

It should set the pub date to now and make draft=false

mohae:
@spf13, sounds good.

I'm still not sure of the appropriate verb, but will probably end up picking one of @bep's suggestions.

If one were to look at this from an editorial process perspective, e.g. draft, approved, publish, then approved or some other verb may be appropriate. However, I'm not sure that that would make the command name more understandable to a user without requiring them to read additional documentation.

Any suggestions or recommendations would be appreciated.

mohae:
Closing because the branch reflects the command name, which is no longer valid. Being replaced by the finalizecmd branch.

mohae:
see https://github.com/spf13/hugo/pull/927

Issue #622, opened 2014-11-08
Static Local Website with Hugo (file:///)


JeremyBYU:
Hey everyone,
  This question will probably seem a little strange to most people because most people dont deploy websites by distributing a folder of html files.  However I sometimes do this at work (with a network file share)  instead of having to deal with a server.

  What I'm trying to do is create a website with hugo that will build all my html files for a simple website and be able to open the html files with a browser.   To accomplish this I set `baseurl = path_to_file_share`. It correctly links all resources in the html files (css, images, javascripts).  I open the index.html homepage file in the file share and it loads correctly.  However all the link to other posts are broken and are replaced with  `<a href="#ZgotmplZ">`

A little googling brought me to this [page](http://stackoverflow.com/questions/14765395/why-am-i-seeing-zgotmplz-in-my-go-html-template-output) which basically says that Go is considering the html to be unsafe and is escaping it.  I'm guessing it is the file protocol in the link that it is considering to be unsafe (e.g. file:///).  

I know that there are some ways in the Go templating language to disable this functionality (see link).  Is there any way we can put a flag in hugo (maybe the config.toml) that will inactivate this html escaping feature.

Thanks for your help 

Jeremy

rickb777:
This is a noble effort, but...

It would be good to fix the Hugo source code and submit a patch, instead of trying to fix links after they've been created.

anthonyfok:
Hi @JeremyBYU and @rickb777,

It turns out this issue is very similar to #347 where the user wanted to use `irc://` links that got turned into `#ZgotmplZ`.  This was fixed by the addition of the `safeUrl` template function in commit 724cc0ddff3427a37b1fa4367880fce23bb4f1f8 on 19 Jan 2015, now in v0.13-DEV.  This new feature is documented around http://gohugo.io/templates/functions/#toc_12 (scroll down a bit to see it).

So, @JeremyBYU, with Hugo v0.13-DEV, you may now use a `baseurl = "file:///path/to/your/website/"` with `uglyurls = true` and have the links work by modifying your templates to use, for example, `<a href="{{ .Permalink | safeUrl }}>` rather than `<a href="{{ .Permalink }}">`.

Please let us know how it works out for you!

rickb777:
This safeUrl solution seems like an ugly hack.

Why is it not possible to construct path-relative URLs so that safeUrl is not necessary?

anthonyfok:
Hi @rickb777,

[...]

Issue #157, opened 2013-12-18
Trouble with links when both HTTP and HTTPS versions of site available


bradberger:
Hi guys, first off Hugo is an amazing piece of software. Thanks so much for all the effort put into it.

Having one issue, though, when a HTTPS version of a site is available. The baseurl as set in config.yaml is rendered (perhaps) a bit too literally. I'm imaging many users will set it to something like "http://yoursite.example.com/" [as is shown in the documentation](http://hugo.spf13.com/overview/configuration).

If a site is using https, though, that breaks things in a browser, with an unsafe mixture of content. This only seems to occur with any `<script>` tags, not with `<a>` tags, `<link>` tags, etc. They all seem to be handled properly.

Why not just strip the http: or https: prefix from the baseurl, or at least encourage using a //www.example.com in the configuration. (FWIW, dropping the http: prefix in config.yaml works, but many may not know to do that.)

If this is something that's worth doing, I can probably submit a PR if desired, but wanted to talk it over first. Perhaps it's a non-issue, too, since the solution is so easy via the config file and very few blogs are actually going to have a valid certificate.

PR #902, opened 2015-02-16
Adapt to relative path bug fix in purell

### LINKED ISSUE #878: Possible bug with new version of purell
Hi,

I'm the author of the `purell` package used by hugo (https://github.com/PuerkitoBio/purell). There's a fix for a bug with relative paths in purell that apparently causes issues for hugo, so I thought I'd give you a chance to look at it before I merge it into master. The fix is in the `fix-relative-paths` branch.

To give an idea, this test currently fails in master:

* `foo/bar` with flag `FlagsAllGreedy` does not return `foo/bar`, but `/foo/bar` (rooted path)

The changes in the fix-relative-paths branch fix this case, but it came to my attention via @anthonyfok that this broke the hugo tests. It is possible that hugo somehow relies on the buggy behaviour of purell.

If you could run a few tests with the `fix-relative-paths` branch and confirm to me when all is good, we can synchronize the merge to master and reduce the chance of breaking stuff in the wild.

Thanks,
Martin

anthonyfok:
Temporary workaround for the bug fix and resulting behavioral change in purell.NormalizeURLString(): a leading '/' was inadvertently to relative links, but no longer, see #878.

I think the real solution is to allow Hugo to make relative URL with relative path, e.g. "../../post/hello-again/", as wished by users in issues #157, #622, etc., without forcing relative URLs to begin with '/'.  Once the fixes are in, let's remove this kludge and restore SanitizeUrl() to the way it was.

Fixes #878

### DIFF
--- a/helpers/url.go
+++ b/helpers/url.go
@@ -1,4 +1,4 @@
-// Copyright © 2013 Steve Francia <spf@spf13.com>.
+// Copyright © 2013-2015 Steve Francia <spf@spf13.com>.
 //
 // Licensed under the Simple Public License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


spf13:
Merged

PR #904, opened 2015-02-16
List of variables

karland:
Some variables are currently not documented and others are explained
across the document. So, I tried to pull an overview from the source.
Pls double check. I am not 100% sure, what the purpose of some variables
is or whether they are only relevant for previous versions. Thanks


anthonyfok:
This is great, @karland!  Thank you for your contribution!  Merged as 1827680beaf10c0710b3daed2cf1f6679c9a070a.

As you may already know, the new `hugo config` command is just in, which may help with checking the list of variables as well as showing their default values.  :-)

PR #907, opened 2015-02-17
Quote strings in `hugo config` output

anthonyfok:
Also, use ` = ` to separate keys and values if metaformatdata is "toml".

What do you think, @bep?  :-)


bep:
Nice! Merged in 9bf223e

Busy today?

anthonyfok:
Thank you @bep!

Yes, I am supposed to be busy today, though I still can't keep my hands off Hugo every now and then, hehe.  <grin>  Back to work with Gramps.  ;-)

PR #908, opened 2015-02-17
Suppress errors for symbolic links witch point to a file.

pcdummy:
This pull does not log the error `Symbolic links for directories not supported, skipping` for symbolic links which point to a file.

I'm not sure if the reordering of imports in `filesystem.go` my edtior does are correct, seems to work here.


spf13:
Merged as d2e022f2a7c3654cd1aaf24ea9134e61cb913cec.

PR #909, opened 2015-02-17
add note about running with --watch; remove some line breaks

riboflavin:
...loyment


spf13:
@riboflavin Merged as aeddaee! Welcome aboard as a contributor!

PR #921, opened 2015-02-23
Updated the 'examples' section of the docs.

atmd83:
The original gives usage as {{< highlight html >}}, when it should be {{% highlight html %}}

### DIFF
--- a/docs/content/extras/highlighting.md
+++ b/docs/content/extras/highlighting.md
@@ -49,7 +49,7 @@ If you want to highlight code, you need to either fence the code with ``` accord
 Not doing either will result in the text being rendered as HTML. This will prevent Pygments highlighting from working.
 
 ```
-{{</* highlight html */>}}
+{{% highlight html %}}
 <section id="main">
   <div>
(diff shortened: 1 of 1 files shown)


bep:
This is correct according to new 0.13 syntax. {{< (raw HTML, no Markdown) is almost always what you want for hightlight.

It is mentioned in the same document you edited.

atmd83:
ok. following the guide I could not get it to work. It only worked when switching to `%`. 
Do you mean by 'no markdown' that the syntax in the example should be used on a 'something.html' page rather then 'something.md' file?

bep:
No. No renaming of files.

That it doesn't work for you suggests that you use the previous version of Hugo (0.12) - 0.13 was released two days ago and I strongly suggest an upgrade.

See

https://github.com/spf13/hugo/releases

atmd83:
I see, that explains everything. I cloned hugo about a week about and only just started looking into the syntax highlighting, so I will have used the old version on the new docs. Will update the project.

Thanks

PR #927, opened 2015-02-25
add finalize command and tests to change draft status of specified content

mohae:
This is an update of the original publish command. 

The new command name is `finalize`, everything else remains the same.
A bug in the creation of an ERROR log entry was fixed too.

Finalize looks at the front matter of the specified content and checks to see if draft == true. If it isn't a draft, it logs an error and exits. Absence of the draft field will result in no changes too.

If it is a draft, the draft field's value gets set to `false` and the date field is updated with the current time and date in _RFC3339_ format. The content is then saved with the modified front matter.

The changes are done to be consistent with the OS's line endings and the front matter's format. The changes do not affect the order that the fields appear in the front matter as the parsed front matter is not written back to the file, only the changes to the values of the `date` and `draft` fields, should any changes be made.

This command requires the path of the content, including the `content` directory, no pre-pending is done to the passed path.


bep:
I know we had this discussion ... and I know I may have suggested this, but "finalize"? I suspect users might get confused by this (finalize as in, is it then locked for further editing?)

I suggest just name the command "undraft". Not a very good word, but I think it's much more clear to the common computer user (used to undo ...).

spf13:
More suggestions..

how about "bless", "approve", "accept", "ready"? Once we have a action we like I'm happy to merge and adjust... or if @mohae you want to adjust I'm fine with that as well. Overall I really like this and want it in ASAP.

mohae:
@spf13, sorry for not responding sooner.

I changed the command name and made this pull request to try and get additional feedback and get a conversation going about what it should be. I wasn't satisfied with "finalize" but it seemed to be the best of the prior round of suggestions. I would agree with @bep's skepticism about the name.

@spf13's new suggestions are certainly an improvement. I like "approve" or "accept" but am still concerned about how clear the intent would be to a user. I have no strong inclinations; I am happy to make the change to whatever is decided and either amend this pull request or create a new one that reflects the new command name.

I will make the necessary changes ASAP.

spf13:
Maybe undraft is the clearest. 

Steve Francia
spf13.com
@spf13

mohae:
I changed the finalize command to undraft as this seems to be the command name that best conveys what is happening. I'm not a fan of the command name, but c'est la vie.

Because of the commits that have occurred since the original pull request and my forgetting to delete the finalize command file, there are now 4 commits in this pull request. I tried to squash the commits, at least the draft and deletion of finalize command, but that did not work. 

There is probably a way to do so, but I do not have that knowledge. If there is a better way to handle this please let me know of direct me to a link where I can learn.

If this is too messy, I can just close this request and submit a new, clean pull request for the undraft command.

bep:
git rebase -i <the commit before you started this> is useful... then pick or fix.

bep:
`git rebase -i master` works great if you started your branch from master. Then s or f all but the first line ...

mohae:
bep, thanks for the advice. I had tried rebasing, but hadn't included enough in my prior attempt.

I seem to have made it worse with my subsequent attempt with git rebase -i commithash so I'm just going to close this request and create a new pull request. It's not something I'm happy about but, at this point, it's more expedient.

It seems I need to get a better grasp on rebase.

PR #940, opened 2015-03-02
Add getSql - SQL support for MySQL, MSSQL, PostgreSQL & SQLite3

SchumacherFM:
http://cyrillschumacher.com/2015/03/01/built-in-sql-support-for-gohugo.io/

Open question:
Use go build tags to distinguish between SQL drivers or include
always all SQL drivers? Including all slows down the build time
during development. Especially sqlite3 has cgo included. Are there
conflicts to other OS when cross-compiling?

When including all drivers then I can merge the content of folder
hugosql into folder tpl ... if desired.

If something with the code or documentation is wrong please comment :-)

Test coverage: 87.7% of statements


bep:
My 50 cents,

With Go being statically linked, I think this gets too complicated for what it gives.

Hugo is still static. It already supports data files in a wide range of formats. If I want data from a database, I have no problems using a third party program, or writing a script outside of Hugo, to dump it to file.

hugolib/shortcode.go
```diff
@@ -88,6 +89,14 @@ func (scp *ShortcodeWithPage) Get(key interface{}) interface{} {
 
 }
 
+// Scratch returns the writable context associated with this Node.
```
bep: What is this? And how is it related to doing database queries?

hugosql/sql.go
```diff
@@ -0,0 +1,358 @@
+package hugosql
+
+// sqlite3 is the default driver and always included.
```
bep: above it says that mysql is the default driver

ariejan:
The nice thing about _not_ using databases is that you can keep all your content under version control, which, besides getting a static site, is the point of tools like Hugo. Adding database support does not follow that opinion, IMHO.

SchumacherFM:
I'm happy closing this: :joy:

1. Introduces to many dependencies
2. I don't use this SQL feature by myself
3. No one asked about this. :no_mouth:

bep:
Good call. I guess you learned something building it.

SchumacherFM:
I learned a lot from building it! and I also appreciated your honest feedback!

PR #951, opened 2015-03-06
Prevent 404.html from prettifying into 404/index.html

### LINKED ISSUE #939: 404 page gets prettified
Looks like a repeat of #415. My `layouts/404.html` gets rendered to `public/404/index.html`. [Documentation](http://gohugo.io/templates/404/) says it should work. It was rendered correctly in Hugo 0.12, it broke in 0.13.

anthonyfok:
Restore @realchaseadams's commit 348e123
"Force `UglyUrls` option to force `404.html` file name"
which got lost after some refactoring (commit 8db3c0b).

Remove the equivalent "force `UglyUrls`" code for `sitemap.xml`
because the refactored code now calls `renderAndWriteXML()`
which uses `WriteDestFile()` which does not prettify a filename.

Fixes #939

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -1250,6 +1250,12 @@ func (s *Site) RenderHomePage() error {
 		}
 	}
 
+	// Force `UglyUrls` option to force `404.html` file name
+	if !s.PageTarget().(*target.PagePub).UglyUrls {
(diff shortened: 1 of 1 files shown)


bep:
Merged in c4c19ad 
Works great.

PR #966, opened 2015-03-09
Fix eq and ne tpl function issue

### LINKED ISSUE #961: eq has problems operating on range's index since 0.13
Create a new site, set theme, create three posts.
Add this simple template at `section/post.html`:
```
{{ $count := sub (len .Data.Pages) 1 }}
<h1>$count = {{ $count  }}</h1>
{{ range $index, $value := .Data.Pages }}
index = {{ $index }} and
(index == $count) is {{ eq $index $count }}<br>
{{ end }}
```
With hugo 0.12 I get this expected output:
```
$count = 2

index = 0 and (index == $count) is false
index = 1 and (index == $count) is false
index = 2 and (index == $count) is true
```

With hugo 0.13 and beyond, I get this unexpected output:
```
$count = 2

index = 0 and (index == $count) is false
index = 1 and (index == $count) is false
index = 2 and (index == $count) is false
```

Same thing happens if I replace `$count` in comparison with a fixed number. There seem to be something wrong with `$index` itself and the way `eq` uses it :(

tatsushid:
`eq` and `ne` template functions don't work as expected when those are
used with a raw number and a calculated value by add, sub etc. It's
caused by both numbers type differences. For example, `eq 5 (add 2 3)`
returns `false` because raw 5 is `int` while `add 2 3` returns 5 with
`int64`

This normalizes `int`, `uint` and `float` type values to `int64`,
`uint64` and `float64` before comparing them. Other type of value is
passed to comparing function without any changes.

Fix #961

### DIFF
--- a/tpl/template.go
+++ b/tpl/template.go
@@ -94,6 +94,21 @@ func New() Template {
 }
 
 func Eq(x, y interface{}) bool {
+	normalize := func(v interface{}) interface{} {
+		vv := reflect.ValueOf(v)
+		switch vv.Kind() {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			return vv.Int()
+		case reflect.Float32, reflect.Float64:
+			return vv.Float()
+		case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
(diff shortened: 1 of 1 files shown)


bep:
That was fast. Merged in 44cdb37

Issue #959, opened 2015-03-07
 BaseUrl vs BaseURL, DocumentId vs DocumentID etc.


bep:
The most important thing is being consequent, and that looks good - but Golint suggests using all caps for these abbreviations:

http://goreportcard.com/report/spf13/hugo

I guess the only reason to to lowercase these is in languages with automated getter/setter generation (read Java).

Fix it or let it be?

anthonyfok:
I think it would be fun and great to fix it.  :smile:

But let's hear from others too.  Perhaps @spf13 could chime in?  :wink:

spf13:
I would follow golint's suggestion on those. I think that the all caps as show reads best. How much would this impact users?

bep:
I would say "very little impact" on users, but I'm not totally sure.

But it's better to do it now than kick ourselves in the legs later on.

If @anthonyfok can do the work (suggest using an IDE with a rename function, like liteide) and do it in a pull request, then we can do some testing and see if we have to put in a temporary "alias or two".

anthonyfok:
Hi all,

This is the first commit that I came up with.  This should be rather safe, as I purposely avoided the command-line flags (cobra and pflag?) as well as struct fields like `BaseUrl` and `Url` that are used in Go HTML templates, to avoid breaking compatibility with existing scripts.

Creating aliases for template *functions* are as simple as adding an extra entry to the FuncMap, e.g. `GetJson` and `GetJSON` both work.  But what about template struct fields?

And, in addition to  `--baseUrl=http://example.com/', do we want to support `--baseURL` and `--baseurl` too?  Or should we make these long flag names case-insensitive?  (Or not?)

Many thanks!

spf13:
I think all flag names should be case insensitive personally.

anthonyfok:
Great minds think alike!  :wink:  <grin, duck, run>

Does that mean we need to patch https://github.com/spf13/pflag?
(I took a quick grep ("upper", "lower", "sensi") and didn't find anything in the upstream https://github.com/ogier/pflag either.)

spf13:
Correct. 

Steve Francia
spf13.com
@spf13

bep:
I have boldly merged the PR. Looked and tested OK by me.

As to the template funcs (getCsv, getCSV) I suggest add a "deprecated decorator func" that logs an error (so it shows up) that says: Fix this, this will be removed in the next version.

anthonyfok:
Just a quick status update:

[...]

PR #970, opened 2015-03-12
More initialism corrections (golint)

anthonyfok:
Deal with `Url -> URL` and `BaseUrl -> BaseURL` conversion
while preserving backward compatibility, and marking
relevant lines `deprecated`.

Also fix related initialisms in strings and comments.

Continued effort in fixing #959.

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -135,7 +135,7 @@ func serve(port int) {
 	}
 }
 
-// fixUrl massages the BaseUrl into a form needed for serving
+// fixURL massages the BaseURL into a form needed for serving
 // all pages correctly.
 func fixURL(s string) (string, error) {
 	useLocalhost := false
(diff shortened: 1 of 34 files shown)


bep:
See 6e30c10d09ec60e8df3b4c17e6ab7a5896245928

To communicate these changes to the users, I think it is better to create temporary methods on the structs for the renamed fields, like this:

```
// Url is deprecated. Will be removed in 0.15.
func (p *pager) Url() template.HTML {
	helpers.Deprecated("Paginator", "Url", "URL")
	return p.URL()
}
```

anthonyfok:
@bep Thank you for authoring `helpers.Deprecated()` and showing me how to use it.  It has been a great learning experience!

I need to leave the house in an hour though.  I hope to find time tonight (12 hours from now?) to study and incorporate your example to deprecated struct fields.  :-)

bep:
No rush. Note that, from the template's point of view, a method will look (almost) like a field, so the above should work for both cases.

anthonyfok:
Hi @bep!

You are brilliant!  And so is your `helpers.Deprecated()` function!  It is such a joy to use!  It makes the transition so much cleaner too!

I have just pushed 92ae298; please check and see if it is OK.

I will update the docs soon.  :-)

Thanks again!

bep:
Merged. Quality work!

I may be a person who thinks this is more important than the average developer Joe, but if you don't pull the tech debt by the horns, you get a legacy system pretty fast.

PR #978, opened 2015-03-16
Adds "getenv" as an alias to os.Getenv

### LINKED ISSUE #977: Add a function to access environment variables in templates
Allow a template like `{{ getenv "FOO" }}` to search the user's current environment and bring in the value for $FOO. If FOO is not set or exported, expand to an empty string.

Example:
```
$ export FOO=world ; hugo
```
With template:
```
Hello, {{ getenv "FOO" }}
```
Outputs:
```
Hello, world
```

quoha:
Fixes #977 by adding a wrapper for the os.Getenv function. The
wrapper takes as input a string that is assumed to be an exported
variable name.

### DIFF
--- a/tpl/template.go
+++ b/tpl/template.go
@@ -1341,6 +1341,7 @@ func init() {
 		"getCSV":      GetCSV,
 		"getCsv":      GetCSV,
 		"seq":         helpers.Seq,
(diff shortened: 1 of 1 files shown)


bep:
Merged in a044734

A one-liner in the doc would be nice.

PR #980, opened 2015-03-16
Adds documentation for the getenv template function

quoha:
Documents the `getenv` function added for #977.


mdhender:
Rebuilt as PR #1043

PR #990, opened 2015-03-21
Added 2 string template functions

fredxinfan:
Added the following string template functions in "tpl/template.go":

1. Substr
Slice a string based on a specified range. e.g {{ substr "abc" 1 2 }} will return "b".

2. Split
Split a string based on a specified delimiter. e.g {{ split "a,b,c"  "," }} will return [a b c].

This is my first commit to Hugo. I've tried to follow the contributor guide. However, please feel free to let me know if I've missed anything. Cheers

### DIFF
--- a/tpl/template_test.go
+++ b/tpl/template_test.go
@@ -276,7 +276,7 @@ func TestIn(t *testing.T) {
 	}
 }
 
-func TestSubstr(t *testing.T) {
+func TestSlice(t *testing.T) {
 	for i, this := range []struct {
 		v1     interface{}
 		v2     int
@@ -289,11 +289,11 @@ func TestSubstr(t *testing.T) {
 		{123, 1, 3, "23"},
 		{tstNoStringer{}, 0, 1, false},
 	} {
-		result, err := Substr(this.v1, this.v2, this.v3)
+		result, err := Slice(this.v1, this.v2, this.v3)
 
 		if b, ok := this.expect.(bool); ok && !b {
 			if err == nil {
-				t.Errorf("[%d] Substr didn't return an expected error", i)
+				t.Errorf("[%d] Slice didn't return an expected error", i)
 			}
 		} else {
 			if err != nil {
--- a/tpl/template_test.go
+++ b/tpl/template_test.go
@@ -276,6 +276,63 @@ func TestIn(t *testing.T) {
 	}
 }
 
+func TestSubstr(t *testing.T) {
+	for i, this := range []struct {
+		v1 interface{}
+		v2 int
+		v3 int
+		expect string
+	}{
+		{"abc", 1, 2, "b"},
+		{"abc", 1, 3, "bc"},
+		{"abc", 0, 1, "a"},
+	} {
+		result, err := Substr(this.v1, this.v2, this.v3)
+
+		if err != nil {
+			t.Errorf("[%d] failed: %s", i, err)
+			continue
+		}
+
+		if result != this.expect {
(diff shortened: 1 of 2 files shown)


bep:
This is good stuff.

Two comments:

1) The `str` argument in the two functions should be `interface{}`. See `Trim`. With the templates you never know what type you get, it could be `[]byte`, `template.HTML` ...
2) Could you squash your commits into one or two logical units (with proper commit messages). `rebase -i master` or similar is a good starting point.

tpl/template_test.go
```diff
@@ -276,6 +276,39 @@ func TestIn(t *testing.T) {
...
+	}
+}
+
+func TestSplit(t *testing.T) {
+	for i, this := range []struct {
+		v1 string
+		v2 string
+		expect []string
+	}{
+		{"a,b", ",", []string{"a", "b"}},
```
bep: You created a table-driven test, so you might as well add another line to this ... maybe another delim?

bep:
This is a good resource for how to write good commit messages:

http://chris.beams.io/posts/git-commit/

(and yes, I know -- the commit log in Hugo doesn't live up to this standard, always ... but it is allowed to try ...)

fredxinfan:
Hi @bep, thanks for all the advice. I've made some changes based on your feedback. Let me know if there are something more I need to improve.

Bookmarked the GIT article, very useful resource, appreciate it.

bep:
Merged in 04817c7

I changed your commit message, the most important being the imperative "add" instead of "added" in the subject line.

fredxinfan:
Ah ha, didn't get what dose the "imperative mood" mean when I read about it. Thanks for the change.

anthonyfok:
Hi @fredxinfan, thank you for your contribution, and @bep, thank you for your commit!

After some thoughts, I think the current `substr` is a misnomer.  For most programmers coming from other languages like C++, PHP, JavaScript, etc., when they see `substr()`, they would think of *start_pos* and *length*, so `substr("abc", 1, 2)` should return `bc`, not just `b`.

What is currently implemented is not a `substr`, but rather, a slice.  The line `return aStr[pos:length], nil` in `func Substr` reflects that confusion.  It is not `aStr[pos:length]`, but rather `aStr[start:end]`.

I think we should fix this before v0.14, otherwise it would *really* confuse users when v0.14 is released.  (Well, it already confused me... it took me days to figure out why `{{ substr "abc" 1 2 }}` would return `b`, when all my previous training says it should return `bc`.

Suggested actions:

 * Rename what is currently implemented as `substr` to `slice` or something similar?

 * Implement a `substr` function that matches Hugo users' expectation?

Thanks!

fredxinfan:
Hi @anthonyfok , I think what you suggested makes sense. The reason is good enough to me if someone already get confused by the function name, like you. The quickest way to resolve this is to rename ```substr``` to ```slice```. As for a more appropriate ```substr``` function, I'm not sure if it's necessary as they're really close enough to each other, but do let me know if you think it's important to have both. I will be happy to do it.

Cheers.

bep:
@anthonyfok - agree, please rename.

bep:
@fredxinfan see #991

bep:
Reopening this ...

I think that with "Slice" people will also expect to slice bread and not just strings. :-)

We could *maybe* make it more generic, but I think Slicestr may be the correct name?

anthonyfok:
@bep Good call about renaming `Slice` to `Slicestr`!

fredxinfan:
Yep. Sounds good :-)

PR #992, opened 2015-03-23
Implement substr template function

### LINKED ISSUE #991: Add Substr template func
See discussion in #990

anthonyfok:
Its behavior is similar to that in JavaScript
with special handling of negative length as found in in PHP.

Fixes #991


bep:
Merged in 5b0245c

Hmm... the int vararg makes for a better template api.

PR #1001, opened 2015-03-27
Cache pygments rendering between runs.

rubenv:
Brings my compile time down from 4 seconds to 24 milliseconds.
Instant livereload works again, as it should.


bep:
Merged in 15463d3

Some code intensive sites will breathe lighter after this ...

PR #1003, opened 2015-03-28
Update embedded rss.xml code in the docs

sainaen:
Nothing special. I just copied the code for `rss.xml` from `tpl/template_embedded.go` to the _RSS (feed) Templates_ page, because it was a bit outdated there.


bep:
Merged, thanks. Should be live soon.

PR #1004, opened 2015-03-28
Fix section name guessing on Windows

ntns:


bep:
Good catch.

Cherry-picked.

PR #1006, opened 2015-03-29
Updates documentation with instructions for a Mac

mdhender:
Adds step-by-step instructions for installing from both brew and
the official tarball. Not sure if steps for installing from source
belong here, so left them out.

Fixes 1005.


bep:
Merged and live soon.

PR #1018, opened 2015-03-31
Fix a bug in IsMenuCurrent when baseUrl is not at the root path

dannys42:
I'm having a problem on my site where my Site.baseUrl points does not point to the root path (eg. http://localhost/mysite/)

In this case, when IsMenuCurrent does not appear to work correctly.  I think menu.go:IsSameResource should really have something that can compare with a prepended Site.baseURL, similar to site.go:getMenusFromConfig.  However, I couldn't see an easy way to do that.  Instead, I opted to modify node.go:IsMenuCurrent to do a similar prepend.

However, I don't think this will catch other bugs in HasMenuCurrent or any other menu access points.

### DIFF
--- a/hugolib/node.go
+++ b/hugolib/node.go
@@ -18,6 +18,7 @@ import (
 	"html/template"
 	"sync"
 	"time"
+    "strings"
 )
 
 type Node struct {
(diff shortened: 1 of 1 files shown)


bep:
Good catch. Two comments:

* It would be nice to have a failing test that this commit fixes.
* Your commit message could need some polishing. See http://chris.beams.io/posts/git-commit/

spf13:
@dannys42 Thanks for the fix. I've merged it in as 2b91b480d01cfcd4b13922effb25d83bc5dd3fb2. @bep's suggestion to add a test would definitely be appreciated in a future PR.

bep:
The test in be325a30885cadaff6792c6aa79413fc1c20be5e covers this fix.

Issue #1009, opened 2015-03-29
Add command to dump template variables


bep:
See http://discuss.gohugo.io/t/the-never-ending-confusion-over-variables/804/10

* Not sure on what level this should be performed, site, page?
* Reflection should be able to read out the exported fields and funcs in a template context

mdhender:
I thought that we could write a for loop in the template to iterate through all the values in `.` and dump the values as needed. I added two functions in PR #1026 to help:
1. "inspect" is similar to `printf "%#v"`.
2. "typeOf" returns the type of a variable or value.

mdhender:
Future thoughts - given how fast Hugo is, would it be possible to have a template hiding on the back end that did a pretty print of the current environment and passed that through CSS to float a window that displayed the results in a navigable hierarchy?

bep:
Yes - that's a good idea ... but since the current context is perg page/node, it should be some kind of DebugWrapper around each html file.

PR #1026, opened 2015-04-01
Add template debugging functions

mdhender:
Template developers frequently need to show variable values while 
debugging templates. Issue #1009 requests better support for this.

The `printf "%#v"` idiom is powerful but requires some knowledge
of Hugo's internal representation of Go value types. Not all
template designers will have that knowledge. The two functions
`inspect` and `typeOf` provide a path towards easier to use and
understand printing of variables and values.

The `inspect` function replaces the `printf` call. This version
relies on the default `string` conversion for the result. This
provides a path towards pretty-printing variables and values.

Templates may use `typeOf` to determine the best formatting
for variables and values.

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -44,6 +44,90 @@ e.g.
 
     {{ if eq .Section "blog" }}current{{ end }}
 
+### inspect
+Return a string containing the default representation of a value.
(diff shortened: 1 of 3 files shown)


mdhender:
Rebuilt as PR #1042

PR #1027, opened 2015-04-01
Docs: Fix Url property in menu

JamesMcMahon:
The current docs don't build under Hugo v0.13. This is a simple fix that matches the property to what is detailed in the menu documentation page.

Specific error this fixes:
```
executing "partials/menu.html" at <.URL>: URL is not a field of struct type *hugolib.MenuEntry in partials/menu.html
```

I've only tested this in the v0.13 release version of Hugo, but I didn't see any breaking changes in master to change `Url` to `URL`.

### DIFF
--- a/docs/layouts/partials/menu.html
+++ b/docs/layouts/partials/menu.html
@@ -16,12 +16,12 @@
             </a>
                 <ul class="sub{{if $currentNode.HasMenuCurrent "main" . }} open{{end}}">
                     {{ range .Children }}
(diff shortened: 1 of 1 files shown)


bep:
The current docs is written for v0.14 and builds fine there.

For v0.13, see https://github.com/spf13/hugo/tree/v0.13.docs

JamesMcMahon:
My bad, I missed that.

Does that mean the next release is considered to have breaking changes?

bep:
That isn't breaking changes ... yet.

We have fixed a lot of Url => URL, Html => HTML etc., but will only print warnings in the next release.

JamesMcMahon:
Ah I see.

Sorry about the confusion. Thanks for the answer.

PR #1032, opened 2015-04-02
Log directories being watched when server is run

### LINKED ISSUE #1030: Log complete watch list
I love using hugo's `-w` option to watch files, but there are times when I'm editing things in `/static/` and would like to see changes reflected immediately. I know I could use something like `gulp-watch`, but could this be integrated into hugo? If this sounds like a good idea, I would be happy to submit a pull request. I was thinking something like: 

`hugo server --watch='static/css' --watch='config.toml'`

or

`hugo server --watch='static/css,config.toml'`

The first depends on the ability of the command line parser to support repeated options, but the second will break if (for some strange reason) a user has a comma in their filename. The `-w` flag would stay a boolean and watch the content directory so the common usage: 

`hugo server -w`

would continue to work as expected. However, I'm open to other ideas.

Thoughts?

dylanmackenzie:
Fixes #1030. Previously hugo only listed the content directory 
as being watched. Now we list all files being watched according to
`commands.getDirList()`. We also introduce a RemoveSubpaths
function and test in the helpers module to reduce noise in the command
line output by not showing subdirectories of ones already being watched.

For example, instead of:
`Watching for changes in $HOME/blog/content`
We get:
`Watching for changes in
$HOME/blog/{data,content,layouts,static,themes/my-theme}`

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -98,7 +98,14 @@ func server(cmd *cobra.Command, args []string) {
 
 	// Watch runs its own server as part of the routine
 	if serverWatch {
-		jww.FEEDBACK.Println("Watching for changes in", helpers.AbsPathify(viper.GetString("ContentDir")))
+		watched := getDirList()
+		workingDir := helpers.AbsPathify(viper.GetString("WorkingDir"))
+		for i, dir := range watched {
+			watched[i], _ = helpers.GetRelativePath(dir, workingDir)
+		}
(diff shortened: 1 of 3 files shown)


bep:
Good solution.

Merged in 99a18b2

bep:
@dylanmackenzie I know this PR is already merged, but we need you to sign the CLA to be able to liense relicense this under Apache 2. Or else we have to remove/rewrite this code.

bep:
OK, rewrite it is.

PR #1033, opened 2015-04-02
Added isso as Comments alternative

noqqe:
Hi,

I'm using isso with my hugo blog which works just fine.
I think this should be added to the docs :)

Thanks!


anthonyfok:
Thanks for your contribution, @noqqe!

Though I changed the link to http://posativ.org/isso/ instead as it seems to be more user-friendly than the GitHub link for newcomers to isso.

Merged and live on http://gohugo.io/extras/comments/.

Issue #852, opened 2015-01-29
Clarify use of `hugo server` in Quick Start Guide (was "Does hugo work for normal hosting sites?")


suntong:
Does hugo work for normal hosting sites, ie, those sites that offer static html, or dynamic php/mysql hosting? They normally will only give you (s)ftp access. hugo will not work for such home-page hosting sites right?

spf13:
Hugo works especially well on these hosted platforms.

suntong001:
Oh, thanks. In that case, please update either of the following articles, because I only have the impression that *"Hugo contains its own ... web server. Simply run hugo server and Hugo will ... run a server with your content"*.

Hugo Quickstart Guide  
http://gohugo.io/overview/quickstart/

Hugo Installation Guide
http://gohugo.io/overview/installing/

thanks

anthonyfok:
Hello @suntong001!

But it does mention on the very first page of the documentation, i.e. http://gohugo.io/overview/introduction/: 

> Technically speaking, Hugo is a static site generator.

That said, our documentation does always have room for improvement, and welcome any contributions that would improve its clarity and remove obstacles for our newcomers.  Would you like to kindly make the necessary edits and submit a Pull Request?  万分感谢！ :pray:

Welcome aboard!

bep:
More questions? http://discuss.gohugo.io/

anthonyfok:
In light of @suntong001's suggestion, as well as questions like the following popping up on our discussion forum:

 * http://discuss.gohugo.io/t/hugo-website-deployment/388
 * http://discuss.gohugo.io/t/after-using-hugo-server-do-i-need-to-run-hugo-before-deploying/601
 * http://discuss.gohugo.io/t/how-to-upload-my-hugo-blog-from-localhost-to-my-server/709

that are more or less related to URLs with `http://localhost:1313/` not working on a public web server, perhaps some kind of clarification in our docs may help remedy the situation.

For example, perhaps adding phrases like "development server" like on http://jekyllrb.com/docs/usage/ would help.  While `hugo server` is certainly capable on a production server, I would say that 99% of Hugo users would prefer to use Apache, Nginx or Lighttpd to serve their static website.

spf13:
I don't like the term "development server" as it connotes that it's not ready for production usage... as is the case with Jekyll. 

Hugo's server is more than powerful enough for production usage and I know of a lot of people using it as such... especially when using it with more powerful needs or when using it with Docker. 

[...]

PR #1038, opened 2015-04-04
Add a section comparing the 'build' and 'server'

mdhender:
One of the first things that new users have to understand is the
difference between Hugo as a web server and Hugo as a web site
generator. Issue #852 asked for documentation to make that clear.

This patch updates the overview page with that information. It will
seem repetitive to users that understand the difference. Weigh that
against the needs of those that don't.

Reference #852


bep:
Merged and live soon.

PR #1040, opened 2015-04-04
Update test logs for uniformity and consistency

### LINKED ISSUE #1028: Standardize on "got" in result log for test cases
Most test cases log errors using lower-case "got":
```
t.Errorf("[%d] got %v but expected %v"...
```
A few use upper-case "Got":
```
t.Errorf("[%d] Got %v but expected %v"...
```
Standardize on a format like "[%d] got X but expected Y".

mdhender:
Many minor fixes to make test logs more consistent and correct a
mispelling.

Standardize on "[%i] got X but expected Y" for log messages. Using
a consistent layout makes it easier to read the test results. This
was mostly changing "Got" to "got". Swapped the order of values on
several calls to bring them in line with the convention.

A few log messages had a sequence number added to identify the
exact scenario that failed. Otherwise, there would be no way to
ascertain which failed When there are many scenarios.

Correct spelling of "expected."

Fixes #1028

### DIFF
--- a/helpers/general_test.go
+++ b/helpers/general_test.go
@@ -27,7 +27,7 @@ func TestGuessType(t *testing.T) {
 	} {
 		result := GuessType(this.in)
 		if result != this.expect {
-			t.Errorf("[%d] GuessType guessed wrong, expected %s, got %s", i, this.expect, result)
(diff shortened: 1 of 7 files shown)


bep:
How on earth did I get to be an author on that commit?

PR #1041, opened 2015-04-04
Sort function names in templates documentation

### LINKED ISSUE #981: Arrange function names in templates/functions.md
It would be easier to find functions in the docs if they were sorted alphabetically.

mdhender:
The documentation on the functions is long.  It is easier to find
a function name when they are sorted within the sections.

Fixes #981


bep:
Merged in 6b50c8f

Only merged into master. Will have to handle 0.13 if we make changes that also apply to the production version.

PR #1042, opened 2015-04-04
Add template debugging functions for variables

mdhender:
Template developers frequently need to show variable values while
debugging templates. Issue #1009 requests better support for this.

There are already ways to display information about template
variables. This patch makes it easier for users that don't have
experience working with Go templates.

The `printf "%#v"` idiom is powerful but requires some knowledge
of Hugo's internal representation of Go value types. Not all
template designers will have that knowledge. The two functions
`inspect` and `typeOf` provide a path towards easier to use and
understand printing of variables and values.

The `inspect` function replaces the `printf` call. This version
relies on the default `string` conversion for the result. This
provides a path towards pretty-printing variables and values.

Templates may use `typeOf` to determine the best formatting
for variables and values. This replaces `printf %T`.

References #1009

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -26,16 +26,35 @@ and other basic tools; these are listed in the
 
 ## General
 
-### isset
-Return true if the parameter is set.
-Takes either a slice, array or channel and an index or a map and a key as input.
+### delimit
+Loops through any array, slice or map and returns a string of all the values separated by the delimiter. There is an optional third parameter that lets you choose a different delimiter to go between the last two values.
+Maps will be sorted by the keys, and only a slice of the values will be returned, keeping a consistent output order.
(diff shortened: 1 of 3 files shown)


bep:
This is a big maybe in my book. Both of these could be replaced with printf. I don't think we should duplicate existing functionality.

If some of that is hard to use, we should document it. I think people would be really happy to know that the powers of printf is available.

bep:
I think the correct way to do this is:

1) Add a new DebugEnabled flag (default false)
2) Add an internal partial that, if DebugEnabled, dumps the context into a HTML table

This way the users can chose where to put this info.

bep:
I'm closing this now. The thought behind this one is excellent -- but I'm not seeing anything here that isn't possible with `printf`.

Issue #1044, opened 2015-04-04
Add double curly brace {{ }} conflict override


bep:
See http://discuss.gohugo.io/t/double-curly-brace-conflict-override/979

alaska:
Seems like solving this would involve changing the delims in the template library as soon as the config file is loaded, but what about all of the templates bundled with Hugo?  You'd probably either
1. Want to change them to something else, like ```<< >>``` (blugh) or
2. Somehow specify a list of locations/files which use the alternate format.

Thoughts?

bep:
I added some thoughts on the PR on the same subject.

anthonyfok:
See Pull Request #1052 contributed by @danke-click (**marc** on the forum),
and Issue #1300 "Make go template delimiter configurable".

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

thevukaslt:
Hey, I run into same problem (Hugo & AngularJS) as well. But here is my idea, why not to implement Laravel solution:
> If you need to display a string that is wrapped in curly braces, you may escape the Blade behavior by prefixing your text with an @ symbol:
> `@{{ This will not be processed by Blade }}`

Under **[Displaying Raw Text With Curly Braces](https://laravel.com/docs/5.0/templates#other-blade-control-structures)** 

For me seems like very convenient yet clean solution. What do you think?

bep:
Replaced by #4513

PR #1052, opened 2015-04-06
Change delimiter enhancement for HUGO

danke-click:
*DelimiterLeft* and *DelimiterRight* can be set in config file. Default is "{{" and "}}".


bep:
Thanks for this. As you might notice, there are plenty test failures.

Some general comments:

I havent read too carefully through your changes, but:

1) Changes in 7 files tells me that this can be done much simpler. My initial thought when I looked at it was to hook into the AddInternalTemplate or something: https://github.com/spf13/hugo/blob/be2097e1ad789eca5d893805a059d94defbe5c48/tpl/template.go#L139
2) Your commit log looks messy and none of your commit messages ends with the magic `Fixes #1044` A good resource for quality commit messages: http://chris.beams.io/posts/git-commit/

helpers/content.go
```diff
@@ -367,3 +367,18 @@ func GetRstContent(content []byte) string {
...
+// Reads the Delimiter Settings from configuration
+// Default is DelimiterLeft = "{{" and DelimiterRight = "}}"
+var DLeft string
+var DRight string
+var ReplaceDelimiters bool = (DLeft != "{{")
+
+// Replaces the delimiters in hardcoded internal templates
+func SetDelims(snippet string) string {
+	if ReplaceDelimiters {
+		snippet = strings.Replace(snippet, "{{", DLeft, -1)
```
bep: strings.Replacer would be better in this case.

helpers/content.go
```diff
@@ -367,3 +367,18 @@ func GetRstContent(content []byte) string {
...
 }
+
+// Reads the Delimiter Settings from configuration
+// Default is DelimiterLeft = "{{" and DelimiterRight = "}}"
+var DLeft string
+var DRight string
+var ReplaceDelimiters bool = (DLeft != "{{")
+
+// Replaces the delimiters in hardcoded internal templates
+func SetDelims(snippet string) string {
```
bep: This func belongs in the tpl package and should be named more descriptive.

danke-click:
*Thank you very much* for your Feedback, your comments and all the additional information that you gave me. I really appreciate it! I also agree with your recommendations and already changed a lot of the code, but it will take more time to figure out how to deal with github in a good quality way ...

**Important:** When I analyzed the code I found several code blocks that generate pure HTML Code or files without using the template package. For example in *htmlredirect.go* to create aliases and there are other places as well.

For that reason I think we still need a helper function to correct the delimiters in that code. So I like to keep the function in the helper package, if that is okay? Otherwise I am totaly open to any suggestion to solve this problem.

I will come back, as soon I find out how to do the ```rebase``` thing :), 
thanks again, marc

bep:
1) None of your changes tells me that the SetDelim method will have to live outside the tpl package, and if so, it should have a different name (you lose the package context). No need to "fix" the templates in the tests. I would, however, expect to see some test cases that loop over both the regular case and some others to verify that the end result is the same.
2) For rebase, this resource may help: http://discuss.gohugo.io/t/howto-rebase-your-feature-branch/442/3

spf13:
@danke-click How is this coming along? Do you need any more help?

bep:
@danke-click any updates on this?

danke-click:
No. I've made all the changes, including all your recommendations a long
time ago.

Problem is, I don't have time to learn github, how to maintain a clear
commit log etc.  -  sorry, but I don't think I'll send any updates anymore.

From my point of view it is faar tooo complicated to deal with a system
like github if you don't use it regularly.



Am 24.07.2015 um 11:46 schrieb Bjørn Erik Pedersen:

bep:
Replaced by https://github.com/spf13/hugo/issues/1300

PR #1062, opened 2015-04-16
Add blackfriday/ExtensionsMask option.

glabra:
Added new config option blackfriday/extensionsmask to
disable unwanted default extensions.


bep:
Merged in 3596b1b

PR #1073, opened 2015-04-24
Use "hugo new" to indicate the command 

RickCogley:
(Dipping a toe in, and noticing there really are not that many weak areas in the docs. Great job. )

Noticed that content/types.md just had "new" instead of hugo new in backticks. 
I thought it is better to make it consistent with other areas with the same.

### DIFF
--- a/docs/content/content/types.md
+++ b/docs/content/content/types.md
@@ -11,7 +11,7 @@ weight: 40
 ---
 
 Hugo has full support for different types of content. A content type can have a
-unique set of meta data, template and can be automatically created by the ``hugo new``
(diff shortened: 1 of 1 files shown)


docs/content/content/types.md
```diff
@@ -11,7 +11,7 @@ weight: 40
 ---
 
 Hugo has full support for different types of content. A content type can have a
-unique set of meta data, template and can be automatically created by the new
```
RickCogley: Thought this felt a little ambiguous as just "new"

bep:
Merged and live soon.

PR #1075, opened 2015-04-25
Add ability to specify the name of the RSS file.

danapsimer:
I made this change to my fork in order to allow setting the name of the RSS file generated for the homepage.  I am migrating a site from Drupal to Hugo and drupal named the RSS feeds as /rss.xml.

However, this does not appear to be working.  I have compiled my changes and put the resulting executable in my /usr/local/bin and I have verified that it is being used by using "hugo version" but the generated site still has the RSS file in /index.xml.

Is there someplace else that needs to be changed to create this option?

Thanks!

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -1251,7 +1251,7 @@ func (s *Site) RenderHomePage() error {
 
 	if !viper.GetBool("DisableRSS") {
 		// XML Feed
-		n.URL = s.permalinkStr("index.xml")
+		n.URL = s.permalinkStr(viper.GetString("RSSUri"))
 		n.Title = ""
 		high := 50
 		if len(s.Pages) < high {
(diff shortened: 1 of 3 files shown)


danapsimer:
BTW, I will be adding tests for this code.  I am not really expecting that you merge it as it is now.

Thanks.

hugolib/site.go
```diff
@@ -1251,7 +1253,7 @@ func (s *Site) RenderHomePage() error {
 
 	if !viper.GetBool("DisableRSS") {
 		// XML Feed
-		n.URL = s.permalinkStr("index.xml")
+		n.URL = s.permalinkStr(s.Info.RSSUri)
```
bep: I would just use viper.GetString("RSSUri") directly here and below .

I would expect this to work.

danapsimer:
I am going to do this the way you suggest.

bep:
Reopen this. Two issues:

* The implementation only work for the top level RSS file (home page). Not for the other nodes (sections, tag listings).
* If you set the `RSSUri` to something other than `index.xml` the `RSSLink` returns wrong value.


ping @danapsimer

bep:
Reopened wrong issue.

PR #1077, opened 2015-04-26
Fix multilingual site

egonelbre:
Noticed that the site has gone out of sync and made some adjustments to it.


bep:
Great work! merged in 24d8ac1

PR #1078, opened 2015-04-26
added RSSUri to specify the name of the RSS file

danapsimer:
This is useful when migrating a site from another system like drupal that uses a different name for the RSS xml files.


bep:
@danapsimer there are some breaking test(s).

bep:
Merged infd08e16
This is very useful.

bep:
Reopen this. Two issues:

    The implementation only work for the top level RSS file (home page). Not for the other nodes (sections, tag listings).
    If you set the RSSUri to something other than index.xml the RSSLink returns wrong value.

ping @danapsimer

PR #1085, opened 2015-04-28
Go template primer - add index

RickCogley:
Related to @bjornerik 's answer in this discussion: http://discuss.gohugo.io/t/inserting-data-from-data-file-into-content-file-newbie-question/1002/3 ... I figured I'd make myself useful and add the reference to the index function, on the go template primer page. 

Also, I moved the reference links to the bottom. 

A general comment: as good as these docs are, the primer at this point makes some assumptions about audience knowledge, so some might find it lacking. Once I understand better, I might make some more clarifying edits.

### DIFF
--- a/docs/content/templates/go-templates.md
+++ b/docs/content/templates/go-templates.md
@@ -144,6 +144,10 @@ range.
         {{ $element }}
     {{ end }}
 
+The `index` function is a [Go][] built-in, and you can read about it [here][gostdlibpkgtexttemplate]. `index`:
+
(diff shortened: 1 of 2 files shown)


RickCogley:
Sorry I am not good at git. I intended to make a PR just for the last commit, because you had already merged my previous ones. This PR has three commits, though, and I don't understand why. Please advise if there is something I should do differently. For instance, is it better to make a new branch for every edit session?

bep:
It is best to create a new branch per edit session. Branches are cheap.

In this case just leave it be -- I can squash them together.

RickCogley:
Thanks @bep. That's what I'll do then!

bep:
Merged and live soone.

PR #1099, opened 2015-05-02
Make bind interface configurable, defaulting to 127.0.0.1

### LINKED ISSUE #1097: Web server listens on all interfaces and says it's on localhost
`hugo server -w` outputs, inter alia:

```
Web Server is available at http://localhost:1313/
```

In fact, it's actually listening on all interfaces. Really there should be an option/flag to pick an interface (which I don't see right now), or at least display the fact that it's actually open to the rest of the internet.

hut8:
Previously the server would always bind to all interfaces, which is weird since it's usually for testing. Resolves #1097 - also resolves some passive voice in help strings.

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -33,6 +33,7 @@ import (
 )
 
 var serverPort int
+var serverInterface string
 var serverWatch bool
 var serverAppend bool
 var disableLiveReload bool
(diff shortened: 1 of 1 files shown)


bep:
Thanks for this.

I'm not sure why I would want to do the below, but the log statement "Web Server is ..." is falsy below.

Could you fix that (amend the commit) -- and at the same time improve your commit message: split it in a title and body and end it with "Fixes #theid".


```
hugo server --baseUrl=bep2.local --bind=bep.local
...
Web Server is available at http://bep2.local:1313/
Press Ctrl+C to stop
```

hut8:
@bep Glad I can help! What is the solution to your conflict there? If "bind" is specified, it is likely to be "0.0.0.0" meaning "all interfaces." Of course the base URL should not be set to 0.0.0.0, right? But at the same time, we can't set the bind address to the base URL, since the base URL could contain components other than an address. So when there's a conflict between bind address and base URL, who "wins?"

bep:
The solution is to fix the log message. Should always use the value in bind.

hut8:
TIL that 0.0.0.0 can be used to connect to the local machine. Did not know that :) Updating patch now.

bep:
@hut8 I don't see any updated patch ...?

hut8:
Hang on, squashing commits, fixing commit messages and pushing...

bep:
OK, just ping me when done ...

hut8:
I forgot to reply to that before. I pushed the commits a while ago. Should be good now.
`u.Host = serverInterface` is the line I added to make the log message accurate.

bep:
Nah, my fault -- I looked at the changes, but didn't notice that part ... All good. Merged. Thanks.

PR #1117, opened 2015-05-09
Remove .html suffix from pagination section url.

calavera:
Fix the pagination urls when the section ends in `.html`, like `http://example.com/blog.html`

### DIFF
--- a/hugolib/pagination.go
+++ b/hugolib/pagination.go
@@ -23,6 +23,7 @@ import (
 	"math"
 	"path"
 	"reflect"
+	"strings"
 )
 
 type Pager struct {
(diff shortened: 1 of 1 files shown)


bep:
When does the section end in html?

bep:
Thanks for this, curious noone discovered this ... Your suffix-trimmer is slightly wrong, though, so I used `strings.TrimSuffix` instead.

See #1120

WalkerEpps:
Hi, guys
I'm having this ".html" suffix problem on pagination on Hugo "v0.15 BuildDate: 2015-11-25T15:34:54-02:00" .
I use OSX 10.11.4 and downloaded Hugo from release page.

PR #1124, opened 2015-05-10
Fix datePublished to publishdate

rauhs:


bep:
Merged in 84f4c21 and docs is live soon, thanks.

PR #1130, opened 2015-05-12
Edits on aliases, comments, theme customizing

RickCogley:
Fleshed out aliases section, loading the "redirect" keyword so that it's easier to find. Specifically added a "how aliases work" section.

Added Discourse to comments section.

Fleshed out themes/customizing, because it seems to be the source of a lot of questions on the forum.


bep:
Merged.

PR #1144, opened 2015-05-15
Edits to templates/404.md

RickCogley:
Added bit about how the 404.html page has to be set to load automatically - auto on Github but needs config on other web servers.

Also tweaked the text a little to emphasize it's a node type, and explain a little more about where the 404 template should be saved.


bep:
Merged and live soon. Very useful addons.

PR #1146, opened 2015-05-16
Edit docs extras syntax highlighting.md

RickCogley:
Added a bunch of clarifying narrative, looking at the discussion forum for what people are asking about, and what I myself was confused about.

* clearer separation of advantage of each style - server or client side - at the top
* inconsistent newlines (different column widths) so I just removed them from obvious paragraphs
* added that the highlight shortcode is not used for the client-side javascripts
* Hugo is taking crap for Pygments being slow, so tried to emphasize that's it's in Pygment's lap. I got your back, Hugo.
* On client-side added prism example
* More clarity on how the css and js needs to be added to your templates
* Explained how the client-side scripts work

### DIFF
--- a/docs/content/extras/highlighting.md
+++ b/docs/content/extras/highlighting.md
@@ -12,23 +12,19 @@ weight: 90
 toc: true
 ---
 
-Hugo provides the ability for you to highlight source code in two different
-ways &mdash; either pre-processed server side from your content, or to defer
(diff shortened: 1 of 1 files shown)


RickCogley:
Please also see my other PR https://github.com/spf13/hugo/pull/1130

bep:
Merged.

PR #1149, opened 2015-05-19
Clarifying Edits for templates/rss.md

RickCogley:
Added clarification for RSS doc page.

* a little formatting to make key words jump out
* emphasize that you can create your own
* add how to link to the feed in <head>
* add what .RSSlink does
* added point that a link to an RSS feed should be of type application/rss+xml


bep:
Merged.

Issue #1155, opened 2015-05-22
Add Taxonomy Term variables to the list of template variables


jramnani:
Taxonomy Term pages have specific variables in addition to "node" variables.  These should be documented in the template variables page so they are easier to find.

Issue #1156, opened 2015-05-22
Add doc for seq, slicestr and substr


bep:
Those 3 new template funcs are very useful and should be documented.

PR #1157, opened 2015-05-22
Docs: Add taxonomy term variables to the list of vars.

jramnani:
Taxonomy Term pages have variables in addition to those on "node"
pages.  Documenting these variables with all the other node variables
makes them easier to find.

Fixes: #1155


jramnani:
Fwiw, I'm not a fan of copy-pasta, so I'm looking for feedback if there's a better way to accomplish this goal.

bep:
Re copy-pasta, I don't think there are a current good way to avoid that. Shortcodes would be too complex.

bep:
Cherry-picked.

Good work -- if you want to do more good, have a look at #1156 -- it should be split in 3, and you may have to read the source code to document it.

PR #1159, opened 2015-05-23
Directory scanning command - getFileList 

rustyoz:
I have added a command to list files in a directory. 
In conjunction with a partial for a gallery plugin, this allows me to update a photo gallery by simply adding the file to the directory. 

the partial [gallery.html](https://github.com/rustyoz/russelloliver.com.au/blob/master/themes/ro/layouts/partials/gallery.html)
and is included at [index.html](https://github.com/rustyoz/russelloliver.com.au/blob/master/themes/ro/layouts/index.html#L44)

Signed-off-by: Russell Oliver <roliver8143@gmail.com>

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1223,6 +1223,7 @@ func init() {
 		"dateFormat":  DateFormat,
 		"getJSON":     GetJSON,
 		"getCSV":      GetCSV,
+		"getFileList": GetFileList,
 		"seq":         helpers.Seq,
 		"getenv":      func(varName string) string { return os.Getenv(varName) },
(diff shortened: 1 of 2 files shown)


bep:
A couple of comments:

1. You might as well name it ReadDir / readDir ... as that is what it does.
2. What are urlparts? Some doc and test cases would be good. I guess this is really a file path. So, maybe better with only one argument and filepath.Join it with the workingDir (some tests for input args that tries to list directories outside of that dir would also be good). Also take into account the Windows users who sends forward slashes.
3. See README / contribution guide lines for input on how to write a commit message

Issue #1147, opened 2015-05-17
RSSUri setting works only for main feed


bep:
Two issues:
-  The implementation only work for the top level RSS file (home page). Not for the other nodes (sections, tag listings).
-  If you set the RSSUri to something other than index.xml the RSSLink returns wrong value.

ping @danapsimer

danapsimer:
These are both good points, I will take a look at how to fix that.

Strubbl:
RSSLink still points to index.xml when setting RSSUri to something different. I am using Hugo Static Site Generator v0.15-DEV BuildDate: 2015-07-29T00:07:32+02:00

Is this issue closed by mistake?

nathany:
It's not clear to me how RSSUri should work with sections, tags, etc.

Meanwhile there are two other issues:

1. It's not documented at https://gohugo.io/overview/configuration/

2. Inside the xml, it has an incorrect (pretty) atom link. No file exists at this location:

```xml
<atom:link href="http://localhost:1313/articles/index.xml" rel="self" type="application/rss+xml"/>
```

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
This was fixed in Hugo 0.18.

PR #1162, opened 2015-05-24
Allow custom RSS Uri for section and taxonomy feeds

skybon:
Replaces the hardcoded `index.xml` address with RSSUri variable in site config. Refs #1147


bep:
Merged in 3cee9b7

PR #1169, opened 2015-05-26
Add `time.Time` type support to `where` tpl func

tatsushid:
`where` tpl function doesn't support `time.Time` type so if people want
to compare such values, it's required that these values are converted
into `int` and compare them.

This improves it. If `time.Time` values are passed to `where`, it
converts them into `int` internally, compares them and returns the
result.

See also
http://discuss.gohugo.io/t/future-posts-and-past-posts/1229/3

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -17,9 +17,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


bep:
That was quick. Merged in 601a2ce

PR #1173, opened 2015-05-27
Vamp.io site created with Hugo

tnolet:
Hi,

we're running http://vamp.io completely on hugo + github pages. Might be a nice example for people to look at. Source is here https://github.com/magneticio/vamp.io


bep:
I have no idea how this Pull Request happened. A bug in GitHub? I see no relecance to your site?

Could you open up a PR with change in https://github.com/spf13/hugo/tree/master/docs/content/showcase ?

PR #1175, opened 2015-05-28
Fix #1166 Add 1x retry with 2sec sleep time. (+Tests)

### LINKED ISSUE #1166: Add retry for dynamic content
Hi,

```getJson```  often get into an error that I believe is because HTTP server sometimes may return 404 or a wrong content, this is being parsed and the error is below: 
```
ERROR: 2015/05/25 Cannot read json from resource http://127.0.0.1:3000/123test with error message invalid character 'E' looking for beginning of value
ERROR: 2015/05/25 Error while rendering page 123test.md: reflect: call of reflect.Value.Type on zero Value
```
Is there a retry policy for the dynamic content ?

Thanks

### PR DESCRIPTION
**Pull request by SchumacherFM**
The retry gets triggered when the parsing of the content fails.
JSON unmarshal fails due to invalid JSON -> retry
CSV cannot find separator -> retry
CSV failed to parse -> retry

HTTP Response from host gets not checked nor there is a timeout
configured for the default http client.

If I should check for the returned HTTP status code ( != 200 ) then I need another client library ... which is one more dependency for hugo ... let me know.

### DIFF
--- a/tpl/template_resources.go
+++ b/tpl/template_resources.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/spf13/afero"
 	"github.com/spf13/hugo/helpers"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by bep** [comment]
Can you fix the commit message in line with the guide lines (and the +Tests is superflous, goes without saying ... "No tests" would be relevant if the commit doesn't contain any).

**Comment by SchumacherFM** [comment]
Done.

**Comment by bep** [comment]
Wouldn't it be simpler to just do a for loop around the httpClient.Get:

```
if res.StatusCode == 200 {
  break
} 
``

Maybe check for all status-codes in the 200-series (if >= 200 && < 300), but this looks too complicated to solve a simpler problem.`

SchumacherFM:
Yes I thought also this way ... but what if it's a 200er reponse with no valid file content?

Let us test (aka merge) this version and if it sucks I'll refactor it to the status codes.

bep:
Yea, well - I see the issue starter also talks about garbled content. Merged in cc5d63c

PR #1194, opened 2015-06-06
Fix substr tpl func's int type variant issue

### LINKED ISSUE #1190: Int Typecast
Might be related to #620. 

Can I even typecast ints in hugo? 

Running:
`substr .Permalink (add (len .Site.BaseURL) 3) (len .Permalink)`

produces the error:
wrong type for value; expected int; got int64

Because of the `(add (len .Site.BaseURL) 3)`

I am using 0.14 by the way

tatsushid:
`substr` template function takes one or two range arguments. Both
arguments must be int type values but if it is used with a calclation
function e.g. `add`, `len` etc, it causes a wrong type error.

This fixes the issue to allow the function to take other integer type
variant like `int64` etc.

This also includes a small fix on no range argument case.

Fix #1190


bep:
Merged in 51cabe6

PR #1200, opened 2015-06-10
Add `after` template function to complement `first`

ariejan:
The `first` template function allows the user to select the first X items of an array. I have a use case where I want those X first items, but I also want the rest of the items after X for separate processing. `after` adds this functionality.

```go
{{ range after 10 .Data.Pages }}
    {{ .Render "title" }}
{{ end }}
```
There is one open question and that's the one of naming things :smiling_imp: 
Is `after` properly named? I chose it over `remaining`, as the supplied limit would not make sense. Are there any other candidates that need to be considered? /cc @spf13 what's your opinion on this?

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -388,6 +388,42 @@ func First(limit interface{}, seq interface{}) (interface{}, error) {
 	return seqv.Slice(0, limitv).Interface(), nil
 }
 
+// Last is exposed to templates, to iterate over the last N items in a
+// rangeable list.
+func Last(limit interface{}, seq interface{}) (interface{}, error) {
+
+	if limit == nil || seq == nil {
+		return nil, errors.New("both limit and seq must be provided")
+	}
+
+	limitv, err := cast.ToIntE(limit)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if limitv < 1 {
+		return nil, errors.New("can't return negative/empty count of items from sequence")
+	}
+
+	seqv := reflect.ValueOf(seq)
+	seqv, isNil := indirect(seqv)
+	if isNil {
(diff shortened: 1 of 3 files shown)


tpl/template_funcs.go
```diff
@@ -177,7 +177,7 @@ func Substr(a interface{}, nums ...interface{}) (string, error) {
 	}
 
 	var start, length int
-	toInt := func (v interface{}, message string) (int, error) {
+	toInt := func(v interface{}, message string) (int, error) {
```
ariejan: Change made by `gofmt`.

tpl/template_funcs_test.go
```diff
@@ -4,13 +4,14 @@ import (
...
 	"fmt"
-	"github.com/stretchr/testify/assert"
 	"html/template"
 	"path"
 	"reflect"
 	"runtime"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
```
ariejan: Change made by `gofmt`.

ariejan:
@spf13 `gofmt` also made some minor tweaks to the code that snuck into my commits. Do you want me to pull them out before considering this for merging?

spf13:
no. It's all good. Go fmt is a good thing.

spf13:
I feel like we need a last as well with this change.

ariejan:
@spf13 seems like a good idea, while I'm at it. Give me few moments... :clock1130:

ariejan:
@spf13 pending tests, here's your `last` :shipit:

bep:
Merged in cc9536e Very useful. And with docs AND tests. Me like.

wayneashleyberry:
:heart:

PR #1201, opened 2015-06-10
Let travis test against go-tip

ariejan:
Just a quick PR to let Travis test against go-tip as well. Go tip is not supported by Hugo (I would assume), but it's a great way to know if there are any issues with upcoming releases of Go. Testing against Go tip has been marked as an allowed failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: go
 go:
   - 1.3
   - 1.4.2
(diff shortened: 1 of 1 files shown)


bep:
Travis syntax I didn't know about, cool.

Merged in 3ab00e6

And now I see my small menu test has started failing on Go 1.3.

PR #1204, opened 2015-06-12
Add ReadDir function to list local files

rustyoz:
Adds functionality wanted by @Lynxiayel in this community [post](http://discuss.gohugo.io/t/a-way-to-traverse-local-files/1249)

### DIFF
--- a/docs/content/extras/localfiles.md
+++ b/docs/content/extras/localfiles.md
@@ -1,0 +1,51 @@
+---
+aliases:
+- /doc/localfiles/
+date: 2015-06-12
+menu:
+  main:
+    parent: extras
+next: /community/mailing-list
+notoc: true
+prev: /extras/urls
+title: Traversing Local Files
+weight: 110
+---
+
+## Traversing Local Files
+
+Hugo includes a way to traverse local files. 
+This is done using the 'ReadDir' function. 
+
+## Using ReadDir
+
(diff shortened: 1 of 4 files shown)


spf13:
Can you add a test for this?

rustyoz:
I honestly don't know how I would do that pragmatically. 
The addition to the docs includes a shortcode that is used in the documentation that uses the command. The command is passed a string with a forward slash included and this works fine on Windows. 

At the moment all I can do is say that it has worked for me on both windows and linux with forward and backward slashes used in the path.

bep:
@rustyoz: This looks fine, but as it is now it is possible to navigate outside the WorkDir. This may sound innocent, and it mostly is, but one could imagine a theme author doing something stupid.

So some thightening of that issue would be nice.

rustyoz:
I have updated the commit to include a test for parent directory markers ".." which should prevent navigation outside of the working directory

rustyoz:
@spf13 @bep I would fix the merge error but I don't know what is causing it. any ideas?

bep:
@rustyoz I haven't forgotten this. I want this feature. Just want some other to chime on to evaluate if the ".." check is protection good enough.

@spf13

others?

rustyoz:
All good. It seems like the appveyor test failure was due to a missing newline of all things.

bep:
Merged in 81e69c4

This is very useful. And thanks for taking the time to write some solid documentation. It's boring, but needed.

PR #1216, opened 2015-06-19
Add support for baseof.ace templates in themes.

### LINKED ISSUE #1215: Unable to locate `baseof.ace` from theme
As shown in [my reduced test case](https://github.com/trombonehero/hugo-bugs/tree/master/baseoface), there is a problem when rendering ACE base templates in themes. In that test case, I have a `layouts/index.ace` that provides content to be rendered by the theme's `layouts/_defaults/baseof.ace` (whose full path is `themes/demotheme/layouts/_defaults/baseof.ace`). At `tpl/templates.go:307`, however, only four paths are checked for `baseof.ace`:

```go
pathsToCheck := []string{
	filepath.Join(templateDir, currBaseAceFilename),
	filepath.Join(templateDir, baseAceFilename),
	filepath.Join(absPath, "_default", currBaseAceFilename),
	filepath.Join(absPath, "_default", baseAceFilename)}
```

which translates to:

```
checking: '/Volumes/Dev/hugo-bugs/baseoface/layouts/index-baseof.ace'
checking: '/Volumes/Dev/hugo-bugs/baseoface/layouts/baseof.ace'
checking: '/Volumes/Dev/hugo-bugs/baseoface/layouts/_default/index-baseof.ace'
checking: '/Volumes/Dev/hugo-bugs/baseoface/layouts/_default/baseof.ace'
```

Notably, the theme directory is not checked. The theme directory *is* consulted when we're looking for templates in it via a call to `loadTemplates('/Volumes/Dev/hugo-bugs/baseoface/themes/demotheme/layouts', 'theme')`, but at that point it's too late: we've already failed to find the base template for my `index.ace`.

trombonehero:
When we find a template that requires a base template, we should also look
for that base template in the current theme.

Fixes #1215.

### DIFF
--- a/docs/content/templates/ace.md
+++ b/docs/content/templates/ace.md
@@ -35,6 +35,8 @@ In Hugo the base template will be chosen in the following order:
 2. <current-path>/baseof.ace
 3. _default/<template-name>-baseof.ace, e.g. list-baseof.ace.
 4. _default/baseof.ace	
+5. <themedir>/layouts/_default/<template-name>-baseof.ace
+6. <themedir>/layouts/_default/baseof.ace
 ```
 
 In the above, `current-path` is where the corresponding inner template lives, `list.ace`, `single.ace`, `index.ace` ...
--- a/helpers/path.go
+++ b/helpers/path.go
@@ -204,6 +204,15 @@ func GetStaticDirPath() string {
 	return AbsPathify(viper.GetString("StaticDir"))
 }
 
+// Get the root directory of the current theme, if there is one.
+// If there is no theme, returns the empty string.
+func GetThemeDir() string {
+	if ThemeSet() {
+		return AbsPathify(filepath.Join("themes", viper.GetString("theme")))
+	}
+	return ""
+}
+
 // GetThemeStaticDirPath returns the theme's static dir path if theme is set.
 // If theme is set and the static dir doesn't exist, an error is returned.
 func GetThemeStaticDirPath() (string, error) {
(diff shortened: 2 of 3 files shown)


trombonehero:
Is it a layering violation to query strings directly from viper? Is there a better way to find out what the current theme directory is?

bep:
@trombonehero I have closed my eyes to my own use of viper sometimes, but in this case I would use ´helpers.GetThemeDataDirPath()´

trombonehero:
Thanks, that is just the kind of helper function I was looking for.

I say "the kind of", however, because `helpers.GetThemesDirPath()` seems to return the current theme's `static` directory rather than the theme itself. Also, I'm not quite sure why there is so much duplication between `getThemeDirPath(string)` and `GetThemesDirPath()`... if I tried to refactor that a bit, would you mind taking a look? I imagine that would need to be a separate pull request.

bep:
Yea, you are right ... that method is really confusing and should be fixed (but look for usage first). Esp since there is also a GetThemeStaticDirPath ...

Just do the fixing in this PR.

trombonehero:
Update: in a local test that I'm doing, the updated code successfully finds the base template file, but Hugo renders an empty HTML file. Can you point me at a example of:

 * unit-testing these kinds of templates and/or
 * extracting "what went wrong" information to print to the command line?

It seems to me that Hugo should generate a warning if it generates a completely blank file.

bep:
If you are sure the "picking" of the template is correct, have a look at `TestAceTemplates` in `template_test.go`. But I guess there aren't any new tests you could add to this one.

The place where the templates gets chosen is a little bit too connected to the filesystem at the moment, and need some refactoring to be testable.

trombonehero:
It looks like the problem was in a mismatch between my built hugolib and hugo binary, or between my theme and my site, or some other mismatch that wasn't the fault of the code. :)

So: I believe that the code in this pull request is working correctly. You can test the functionality by cloning https://github.com/trombonehero/hugo-bugs and running `hugo --theme=demotheme` within
[the `baseoface` directory](https://github.com/trombonehero/hugo-bugs/tree/master/baseoface).

bep:
Merged.

PR #1225, opened 2015-06-25
Proper parsing structured array from yaml based FrontMatter

piotrkowalczuk:
I have faced exactly same problem like here: http://discuss.gohugo.io/t/use-an-array-of-structured-frontmatter-in-yaml/207

Since i use same markdown file as a input for pandoc converter i cant simply prefix structure by `params:`.


bep:
Have you tested this with TOML frontmatter?

piotrkowalczuk:
Localy it works. 

It looks like there is lack of automated tests for that.

bep:
Merged in 29e786a

PR #1238, opened 2015-06-29
Add nil comparison to where tpl function

### LINKED ISSUE #1232: Add query-support on non-existing page params in where template func
### Use Case:
There are about 200 pages in the website.
But some pages like "privacy", "about" and "terms and conditions" are appearing on the home page.
You dont want these pages on home page, so you declare a page specific param on these pages:

#### front matter:
+++
...
notOnHomePage = "true"
+++

#### and in index.html

```
...
{{range first 10 (where .Paginator.Pages ".Params.notOnHomePage" "!=" "true")}}
...
```

But now not even single post is shown on the homepage.

Original Discussion:
http://discuss.gohugo.io/t/how-to-filter-pages-on-homepage-based-on-page-params/1369

tatsushid:
`where` template function's internal condition check function always
returns `false` when a target value doesn't exist or it's nil value but
this behavior makes it difficult to filter values which doesn't have a
particular parameter.

To solve it, this adds nil value comparison to the function.
`where Values ".Param.key" nil` like clause can be used for the case
above.

Only "=", "==", "eq", "!=", "<>", "ne" operators are allowed to be used
with `nil`. If an other operator is passed with `nil`, the condition
check function returns `false` like before.

Fix #1232

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -536,17 +536,21 @@ func evaluateSubElem(obj reflect.Value, elemName string) (reflect.Value, error) 
 }
 
 func checkCondition(v, mv reflect.Value, op string) (bool, error) {
-	if !v.IsValid() || !mv.IsValid() {
-		return false, nil
+	v, vIsNil := indirect(v)
+	if !v.IsValid() {
+		vIsNil = true
 	}
-
-	var isNil bool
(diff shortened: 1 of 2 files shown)


bep:
Merged in dd732e8. That was quick! This is very useful.

PR #1241, opened 2015-07-02
Added missing "Social" section to SiteInfo

delputnam:
I could be wrong here, but it looks to me like .Site.Social.facebook is used in tpl/template_embedded.go, but the variable is never set. I've added a line to initializeSiteInfo to map the info from config into this variable.


bep:
Well spotted. Merged in 9b3d0cf

PR #1242, opened 2015-07-02
Add Page.RawContent() to access raw Markdown as a string.

maruel:
In particular, RawContent() excludes the metadata header.

This is necessary in the use case of embedding remarkjs.com slides, as it needs
the unprocessed Markdown content to generate the slides.

So a remarkjs based single.html will look like this:
```
<!DOCTYPE html>
<html>
<head>
  ...
  <script src="//gnab.github.io/remark/downloads/remark-latest.min.js"></script>
</head>
<body>
  <textarea id="source">
{{ .RawContent }}
  </textarea>
  <script>
    var slideshow = remark.create({});
  </script>
</body>
</html>
```

### DIFF
--- a/docs/content/templates/variables.md
+++ b/docs/content/templates/variables.md
@@ -49,6 +49,7 @@ matter, content or derived from file location.
 **.WordCount** The number of words in the content.<br>
 **.ReadingTime** The estimated time it takes to read the content in minutes.<br>
 **.Weight** Assigned weight (in the front matter) to this content, used in sorting.<br>
+**.RawContent** Raw Markdown content without the metadata header. Useful with [remarkjs.com](http://remarkjs.com)<br>
 **.IsNode** Always false for pages.<br>
 **.IsPage** Always true for page.<br>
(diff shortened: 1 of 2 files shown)


bep:
This looks fine ... but someone should add a line in the documentation about it ... and I'd rather that be you than me ...?

maruel:
Most Page methods aren't documented at http://gohugo.io/templates/variables/.
Compare with http://godoc.org/github.com/spf13/hugo/hugolib#Page;
Author(), Authors(), Extension(), Plain(), etc.

maruel:
To clarify my previous comment, I'll happily document if desired it but the precedent implies the reverse so that's why I hadn't in the first place.

bep:
Yea, well. Let us not learn by bad examples. 

You have a good use case for ´RawContent´. This use case is useful to others. And hence it should be documented.

maruel:
Added.

bep:
Merged in 24c8226

PR #1243, opened 2015-07-02
Add hasPrefix template function.

colinbate:
I found this function necessary when working on an archive page for a static site last year. I figured others might find it useful.


colinbate:
Looking at the new documentation, the GroupBy functionality should do what I need now, but I'll have to have a play around to see.

bep:
@colinbate did the GroupBy support your use case?

colinbate:
I haven't tried switching out to use GroupBy yet, although I suspect that will be useful. What I was able to use was the `where` template function (along with some metadata tweaks) to avoid using hasPrefix.

spf13:
Merged as 2e92f36

PR #1248, opened 2015-07-04
Format imports

squiidz:
Format imports


bep:
As you doesn't state what you used to format these -- and many/most of these files has been touched by `go fmt` recently -- I would say thanks, but this is better handled as an ongoing process.

PR #1255, opened 2015-07-08
Support GitHub-flavoured markdown code fences for highlighting #362

### LINKED ISSUE #362: Support GitHub-flavoured markdown code fences for highlighting
I think it would be much better if hugo supported
````markdown
```html
<h1>example</h1>
`` `
````
(The extra space is there to avoid GitHub parsing the inner block as markdown)
for highlighted code rendering.

The reason behind this is that a lot of Markdown editors support code fences, but not the `highlight` tag, so this makes hugo more consistent with other Markdown-renderering projects.

I don't think it would be too difficult for hugo to re-write the example above into the one below before passing it to pygments. What do you think?

```django
{{% highlight html %}}
<h1>example</h1>
{{% /highlight %}}
```

I'd be happy to try implementing this, although I've only got a little bit of experience with go.

bramp:
Fixes #362

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -148,6 +148,7 @@ func LoadDefaultSettings() {
 	viper.SetDefault("PygmentsStyle", "monokai")
 	viper.SetDefault("DefaultExtension", "html")
 	viper.SetDefault("PygmentsUseClasses", false)
+	viper.SetDefault("PygmentsCodeFences", false)
 	viper.SetDefault("DisableLiveReload", false)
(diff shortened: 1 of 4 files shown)


bep:
Oh, this is useful. I replaced all highlight shortcodes on one of my sites. Works great.

Merged. 

I polished your commit messages a little. Note the use of "Fixes #362" at the end of the body. This closes the issue (saves manual work for me).

bep:
svigerfar xcccxcgcgxgc

PR #1257, opened 2015-07-08
add a showcase of sa.muel.be

SamuelDebruyn:
I added my personal blog sa.muel.be as showcase. The source code used to generate hugo is available as well as the generated static files.


bep:
Merged and live soon.

PR #1261, opened 2015-07-09
Add arianv and learvnoverpass

Secretmapper:


bep:
As I see it, `https://github.com/Secretmapper/arianv` isn't a link to the Hugo source (it is a link to the generated HTML). It's fine if you do not want to show off the source, but then the source link should be removed.

Secretmapper:
Looks like I accidentally linked it to the gh-pages branch! Fixed now!

bep:
Merged and live soon.

PR #1265, opened 2015-07-10
Omit protocol from embedded font resources

EMarshal:
Fixes the mixed content errors and loads fonts when accessing
https://gohugo.io/


bep:
Good catch. Merged and live soon.

PR #1268, opened 2015-07-12
Checks to ensure theme directory, if specified, exists.

### LINKED ISSUE #1234: Do not start server on missing theme or any other reason no content can be generated
It may be easier to not start the server if there are no generated files.  (Ie. you must have at least one file in public/ (or it's equivalent) or the `hugo server` command with quit.

scottcwilson:
Fixes #1234

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -226,6 +226,11 @@ func InitializeConfig() {
 	}
 
 	if Theme != "" {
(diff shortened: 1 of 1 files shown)


bep:
Merged in 029cdb68f5d76efd697fb9561aec3bf0aa8fcefb

PR #1226, opened 2015-06-25
Allow page.HasMenuCurrent() and node.HasMenuCurrent() to proceed with…

vjeantet:
… multi-level nested menus

Currently HasMenuCurrent() only process the first 2 levels.

PR #1271, opened 2015-07-13
Allow page.HasMenuCurrent() and node.HasMenuCurrent() to proceed with multi-level nested menus

vjeantet:
new PR, with fixed tests (first PR was #1226)

The following test fixture was buggy i think, because when i look into the menuItem from 
```findTestMenuEntryByID(s, "grandparent", "grandparentId")``` I can see my home menu.
This test worked before because HasMenuCurrent returned after checking the first children, without checking children of children... and so on...

@bep ,  please can you review :)


bep:
Will have a look tomorrow.

bep:
Merged in 2c22a8b

PR #1273, opened 2015-07-14
Link to front matter was dead--now correct?

chreliot:
added leading backslash, changing [front matter](content/front-matter) to [front matter](/content/front-matter)


bep:
Merged and live soon.

PR #1278, opened 2015-07-17
Unclear RSS template docs

Clam-:
The documentation for the RSS templating is a little unclear.
http://gohugo.io/templates/rss/

Some users may attempt to look for a ```__internal``` directory rather than assume that's the aforementioned "Hugo own template."

Here's my suggestion.


bep:
Merged and live soon.

PR #1280, opened 2015-07-18
Adds a check for the setting of watch flag in config file.

### LINKED ISSUE #1074: Config variable «watch» in config.yaml doesn't work
I has this in my config.yaml:

```
# watch filesystem for changes and recreate as needed
watch: true
```
According to [Configuring Hugo](http://gohugo.io/overview/configuration/) documentation.
But, it didn't seem to work. I still need to use options `-w` or `-watch` in command line to watching files to work.

After commenting this in a [a forum thread](http://discuss.gohugo.io/t/watch-configuration-variable-doesnt-work/1092), I realized that maybe (and only maybe), considering that the documentation only says that:

```
# watch filesystem for changes and recreate as needed
watch:                      false
```

it's aimed for people that deploys their site with the hugo's built-in server (instead of nginx, apache...), and it's intended to **only** being set to `false`, **never** to `true`, just to deactivate the watch in production.

If that's the case, I think that the comment in that documentation:

```
# watch filesystem for changes and recreate as needed
```

Should be changed for something clearer, because now it'ss pretty confuse. Someone (that's me!) can guess that, if you set it to `true`, it will watch for changes without the need of explicitly using `-w` or `-watch` option in commandline.

Something like:

```
# disable watch of filesystem for changes (handy for production environments)
```

would be much clearer.

Just in case it is in fact a bug (and it should work also for activate `watch`):
I'm using archlinux and `hugo version`

>Hugo Static Site Generator v0.13 BuildDate: 2015-04-20T21:17:06+02:00

(installed from AUR package, [hugo](https://aur.archlinux.org/packages/hugo/)).

scottcwilson:
Fixes #1074

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -301,7 +301,7 @@ func build(watches ...bool) {
 	}
 	utils.StopOnErr(buildSite(BuildWatch || watch))
 
-	if BuildWatch {
+	if BuildWatch || viper.GetBool("watch") {
 		jww.FEEDBACK.Println("Watching for changes in", helpers.AbsPathify(viper.GetString("ContentDir")))
 		jww.FEEDBACK.Println("Press Ctrl+C to stop")
 		utils.CheckErr(NewWatcher(0))
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -301,7 +301,7 @@ func build(watches ...bool) {
 	}
 	utils.StopOnErr(buildSite(BuildWatch || watch))
 
-	if BuildWatch || viper.GetBool("watch") {
+	if BuildWatch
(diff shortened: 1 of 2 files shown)


scottcwilson:
Hmm... hang on ... this change seems to stop the webserver from starting up.

scottcwilson:
I believe this is correct now.

bep:
Please

1) Squash your commits into logical units (in this case: 1)
2) Fix your commit messages, see README

scottcwilson:
New Squashed PR: https://github.com/spf13/hugo/pull/1284

scottcwilson:
Closing this particular PR in favor of #1284

PR #1284, opened 2015-07-18
Add a check for the setting of watch flag in config file

### LINKED ISSUE #1074: Config variable «watch» in config.yaml doesn't work

scottcwilson:
Fixes #1074

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -67,6 +67,10 @@ func server(cmd *cobra.Command, args []string) {
 		viper.Set("DisableLiveReload", disableLiveReload)
 	}
 
+	if viper.GetBool("watch") {
+    	serverWatch = true
+    }
+
 	if serverWatch {
 		viper.Set("Watch", true)
 	}


bep:
Oh, and one last thing ... You do very good work, but remember to run "go fmt" on the code before commit.

I use LiteIde for Go coding, but I know other editors too have "go fmt on save".

scottcwilson:
Rats!  Sorry, I could have sworn I did.  I will double check the README before my next commit.  Thanks for your patience.

commands/server.go
```diff
@@ -67,6 +67,10 @@ func server(cmd *cobra.Command, args []string) {
 		viper.Set("DisableLiveReload", disableLiveReload)
 	}
 
+	if viper.GetBool("watch") {
+    	serverWatch = true
+    }
```
bep: If you can `go fmt` this and move it below the next conditional (of no practical difference, but as it is now it is slightly confusing to read).

scottcwilson:
Closing this PR; new one opened #1286.

bep:
Closing and open a new PR is good for the project, as it shows a lot of activity ... but:

I notice you do your changes on your own master branch.

* Please create a new branch for every new pull request (a Git Branch is cheap)
* When doing changes, just do a "commit --amend" or similar followed by "git push -f" (force push, nobody cares about the history of a temporary branch)

scottcwilson:
Got it. Thanks.

PR #1285, opened 2015-07-18
Fix periods in taxonomies create bad paths

### LINKED ISSUE #1188: Periods in taxonomies generates bad paths
See

http://discuss.gohugo.io/t/periods-in-taxonomies-are-generating-bad-paths/1258

ryanclarke:
Fixes #1188 

Finally found a fairly simple solution.


bep:
Merged in 18d69d7

I suspect a similar issue with sections ... and we should get some better tests for this. But that for another day. Thanks!

PR #1289, opened 2015-07-20
Let Amber template call Hugo's custom functions

tatsushid:
Amber doesn't share text/template `FuncMap` functions and has its own
function list. This allows Amber to call Hugo's custom functions.


bep:
Merged in 03a7016

PR #1294, opened 2015-07-20
Fix links to Ace and Amber pages

scottcwilson:
Links are incorrect in current version.  

http://gohugo.io/templates/ace/
refers to http://gohugo.io/templates/amber-templates  

and 

http://gohugo.io/templates/amber/
refers to http://gohugo.io/templates/ace-templates


bep:
Merged and live soon.

PR #1307, opened 2015-07-29
Asciidoc[tor]: use --no-header-footer option

Batmat:
This simplifies the retrieval of the HTML (no more need to extract the
part within body) and also removes the unwanted "Last Updated" part in
the article.

This is somehow a followup to @mojavelinux 's comment here: https://github.com/spf13/hugo/issues/470#issuecomment-107251499

Cheers


bep:
Nice. Merged in b23b546

PR #1311, opened 2015-08-01
Add map support to scratch

mjanda:
Hello,
go newbie here, so be kind and let me know if I'm doing something stupid there :)

I'll start by what I'm trying to solve. I'm using shortcodes to add references inline and automatically generate list of references at the end of the post. With current scratch, I was able to collect them and list them in the same order as they were mentioned. But I want them sorted and de-duplicated. And I don't think that's possible right now, so this is my solution.

I think something like that is relatively common use case and maybe it might be useful for other things.

### DIFF
--- a/hugolib/scratch.go
+++ b/hugolib/scratch.go
@@ -15,6 +15,7 @@ package hugolib
 
 import (
 	"github.com/spf13/hugo/helpers"
+	"sort"
 )
 
 // Scratch is a writable context used for stateful operations in Page/Node rendering.
@@ -50,6 +51,41 @@ func (c *Scratch) Set(key string, value interface{}) string {
 // Get returns a value previously set by Add or Set
 func (c *Scratch) Get(key string) interface{} {
 	return c.values[key]
+}
+
+// SetInMap stores a value to a map with the given key in the Node context.
+// This map can later be retrieved with GetSortedMapValues.
+func (c *Scratch) SetInMap(key string, mapKey string, value interface{}) string {
+	_, found := c.values[key]
+	if !found {
+		c.values[key] = make(map[string]interface{})
+	}
+
+	c.values[key].(map[string]interface{})[mapKey] = value
+	return ""
+}
+
+// GetSortedMapValues returns a sorted map previously filled with SetInMap
+func (c *Scratch) GetSortedMapValues(key string) interface{} {
+	if c.values[key] == nil {
+		return nil
+	}
+
+	unsortedMap := c.values[key].(map[string]interface{})
+
+	var keys []string
+	for mapKey, _ := range unsortedMap {
+		keys = append(keys, mapKey)
+	}
+
+	sort.Strings(keys)
+
+	sortedArray := make([]interface{}, len(unsortedMap))
+	for i, mapKey := range keys {
+		sortedArray[i] = unsortedMap[mapKey]
+	}
+
+	return sortedArray
 }
 
 func newScratch() *Scratch {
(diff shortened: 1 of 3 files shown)


hugolib/scratch.go
```diff
@@ -52,6 +53,39 @@ func (c *Scratch) Get(key string) interface{} {
...
+	}
+	
+	var keys []string
+	for mapKey, _ := range c.values[key].(map[string]interface{}) {
+		keys = append(keys, mapKey)
+	}
+
+	sort.Strings(keys)
+	
+	sortedArray := make(map[int]interface{}, len(c.values[key].(map[string]interface{})))
```
bep: In Go, since Go 1.1, you have no guarantee about the iteration order of maps. So what you want here is a slice.

hugolib/scratch.go
```diff
@@ -52,6 +53,39 @@ func (c *Scratch) Get(key string) interface{} {
 	return c.values[key]
 }
 
+// SetToMap stores a value to a map with the given key in the Node context.
+// This map can later be retrieved with GetSortedMap.
+func (c *Scratch) SetToMap(key string, mapKey string, value interface{}) string {
+	_, found := c.values[key]
+    if !found {
+		c.values[key] = make(map[string]interface{})
+	}
```
bep: Run go fmt on the code.

bep:
I thought Blackfriday/Markdown had good enough inline references?

I'm sure there are other general use cases for this, though.

bep:
Thinking about it, I don't think we can depend on the shortcodes being processed in order, either. It works now, but will break if we decide to do some lazy evaluation or process them in parallel.

mjanda:
Thank you, fixed.

Shortcode processing order isn't a problem **for me**, but I'm not sure what other possible uses do you imagine for this.

Regarding MD references, I'm not sure I'm able to get them to do what I want. I supply my shortcode with inline text, full citation and url and I'll get something like:

_In a groundbreaking paper by **Meyer-Rochow et al. (2003)**, blah blah blah..._

and at the end of the post, I'd get:
### References
* Meyer-Rochow, V. B., & Gal, J. (2003). _Pressures produced when penguins pooh - calculations on avian defaecation_. Polar Biology, 27(1), 56–58. doi:10.1007/s00300-003-0563-3

So I'm not dealing with links (they are optional) - links become unreachable, DOIs and citations will live forever :) But it's possible I'm missing something and markdown/blackfriday can actually do something like that.

bep:
OK, for that kind of references you'll not get much help from Markdown, I think.

It does look like you depend on a given order: To be able to print out all the references, all the references must be added first.

Let me think a little about this; at least now the code looks solid.

mjanda:
Well, my code for printing list of references is in template for single post and isn't part of content (can't forget to add this that way), so it's unlikely it would be affected by this.

hugolib/scratch.go
```diff
@@ -52,6 +53,39 @@ func (c *Scratch) Get(key string) interface{} {
...
+}
+
+// GetSortedMap returns a sorted map previously filled with SetToMap
+func (c *Scratch) GetSortedMap(key string) interface{} {
+	if c.values[key] == nil {
+		return nil
+	}
+
+	var keys []string
+	for mapKey, _ := range c.values[key].(map[string]interface{}) {
```
bep: Type assertions in Go are cheap, but doing it 3 times in a row screams for a variable. It's not a big thing, but I will probably change it at some point if you don't.

bep:
OK, that makes sense -- then we should be good to go ...

One thing:

`SetToMap`is a confusing method name (implies conversion between a Set and a Map). What about `SetInMap`?

* A couple of lines in the documentation would be nice
* Also, try to squash your commits into logical units

mjanda:
Ok, that squashing thing will probably take me some time to figure out. I'll take a look at it tomorrow.

bep:
I have one final naming objection:

`GetSortedMap` isn't really accurate when you return is a slice/array. People might expect to get a key/value map.

Maybe better just to spell it out long:

`GetSortedMapValues`

mjanda:
Ok, renamed and everything squashed into one commit.

bep:
Merged in 4bed696 ... very useful.

PR #1314, opened 2015-08-02
trim trailing spaces from YAML and TOML delimiters

jmhodges:
When someone hits space after typing "---" (or "+++") but before they hit return, hugo silently failed to parse the file. This corrects that.

Over half of my files from a jekyll site had this issue.


parser/parse_frontmatter_test.go
```diff
@@ -200,6 +200,10 @@ func TestPageHasFrontMatter(t *testing.T) {
 		{[]byte("---"), false},
 		{[]byte("---\n"), true},
 		{[]byte("---\n"), true},
+		{[]byte("--- \n"), true},
+		{[]byte("---  \n"), true},
+		{[]byte("--- \r\n"), true},
+		{[]byte("---  \r\n"), true},
```
tatsushid: I think it doesn't need to add `\r\n` case because at

```go
test.content = bytes.Replace(test.content, []byte("\n"), []byte(ending), -1)
```

line, `\n` characters in all test cases are replaced by `\n` or `\r\n` and both cases are tested.

parser/page.go
```diff
@@ -22,13 +23,9 @@ const (
...
 var (
-	delims = [][]byte{
-		[]byte(YAML_DELIM_UNIX),
-		[]byte(YAML_DELIM_DOS),
-		[]byte(TOML_DELIM_UNIX),
-		[]byte(TOML_DELIM_DOS),
-		[]byte(JSON_LEAD),
-	}
+	delims = regexp.MustCompile(
+		"^(" + regexp.QuoteMeta(YAML_DELIM) + "\\s*\n|" + regexp.QuoteMeta(TOML_DELIM) + "\\s*\n|" + regexp.QuoteMeta(JSON_LEAD) + ")",
```
tatsushid: It is better to use a raw string literal (string quoted by a back quote) as a regexp string. It allows simply writing `\s*\n|`, no need to escape `\s` by a back slash

parser/parse_frontmatter_test.go
jmhodges: Ah, shoot. I had seen that and thought I had removed all of these. Fixing.

jmhodges: Pushed. Thanks for the heads up.

jmhodges:
Fixed the review comments from @tatsushid

bep:
Thanks ... merged in 8d28686

PR #1315, opened 2015-08-02
add undocumented published setting to front matter

jmhodges:
A new "published" setting that is the opposite of "draft" is added and left intentionally undocumented.

This setting comes from jekyll and eases the transition to hugo greatly. We leave it undocumented so that folks don't rely on it, but also don't shoot themselves in the foot during a jekyll migration. The foot-shooting occurs if they have only a few documents that were drafts ("published: false") in
the jekyll version of their site and don't notice that those drafts were published in the migration to hugo.

I realize this will be a more controversial pull request. However, when combined with spf13/hugo#1314, and spf13/cast#17 (and maybe spf13/hugo#1313 if patched), the jekyll to hugo migration becomes very easy. With those changes, migrating simple websites (like https://www.somethingsimilar.com and many other jekyll installs) requires no editing of the original data at all. 

My jekyll to hugo migration became a copy and paste of my markdown files into the right directory plus adjusting some templates to use the right variables. It's nice.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -463,12 +463,15 @@ func (p *Page) RelPermalink() (string, error) {
 	return link.String(), nil
 }
 
+var ErrHasDraftAndPublished = errors.New("both draft and published parameters were found in page's frontmatter")
+
 func (p *Page) update(f interface{}) error {
 	if f == nil {
 		return fmt.Errorf("no metadata found")
 	}
 	m := f.(map[string]interface{})
 	var err error
+	draftSeen := 0
 	for k, v := range m {
 		loki := strings.ToLower(k)
 		switch loki {
@@ -508,8 +511,10 @@ func (p *Page) update(f interface{}) error {
 			}
 		case "draft":
 			p.Draft = cast.ToBool(v)
+			draftSeen++
 		case "published": // Intentionally undocumented
 			p.Draft = !cast.ToBool(v)
+			draftSeen++
 		case "layout":
 			p.layout = cast.ToString(v)
 		case "markup":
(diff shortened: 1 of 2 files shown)


bep:
SGTM - objections, @spf13 ?

spf13:
I think we should do all we can to make migration easier provided it doesn’t compromise Hugo’s ease of use. 

I like this idea a lot, but think we need to better handle the case where both published and draft are set.

jmhodges:
Pushed up an error check for that case. Was that what you were thinking about?

spf13:
Probably error and make sure that draft always takes precedence.

jmhodges:
Well, my latest change throws an error, but I assumed the code would treat the struct being updated as garbage on error. It is side-effect-y, so I guess I could do the footwork for not setting to Draft

bep:
What I believe @spf13 means, and it does make sense, is: If both draft and published are set, use the value in draft, but also LOG an ERROR.

spf13:
Yes. Thanks for clarifying. 


Best,
Steve

-- 
Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13




On Mon, Aug 3, 2015 at 11:37 AM -0700, "Bjørn Erik Pedersen" <notifications@github.com> wrote:












What I believe @spf13 means, and it does make sense, is: If both draft and published are set, use the value in draft, but also LOG an ERROR.

jmhodges:
Why log and not return an error?

jmhodges:
I've updated the code to log and only set draft if both published and draft are present. It still returns an error. If I misunderstood, please let me know.

jmhodges:
Ping? I'm still looking for some feedback.

I get a nice

    ERROR: 2015/08/30 both draft and published parameters were found in page's frontmatter

when running this code now.

bep:
Sorry, missed this one. This looks good. Merged in 11593b (I squashed your commits into one).

jmhodges:
Happy for the squash! Thanks.

PR #1322, opened 2015-08-03
Fix searching YAML/TOML delimiters in frontmatter

### LINKED ISSUE #1320: Ignore delimiter strings in values
If I use the following in my yaml:

    permalink: "/updates/2011/06/HTML5-Libraries---Late-June.html"
    ---

I get the following error:

    "Errors reading pages: Error:frontmatter delimiter must be followed by CR+LF or LF but those can't be found at filepos ..."

It's basically erroring at the '---' in the permalink, is there a way to escape the value?

tatsushid:
When a YAML/TOML's delimiter character sequence is included in a
frontmatter string, parser mistakes it as a delimiter. This fixes it by
checking a character right before the delimiter sequence is '\n' or it
is the beginning of the frontmatter.

Fix #1320

### DIFF
--- a/parser/page.go
+++ b/parser/page.go
@@ -207,7 +207,7 @@ func extractFrontMatterDelims(r *bufio.Reader, left, right []byte) (fm FrontMatt
 		switch c {
 		case left[len(left)-1]:
 			if sameDelim { // YAML, TOML case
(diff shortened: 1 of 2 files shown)


bep:
Merged in 98659bf

PR #1323, opened 2015-08-04
Add option to disable Blackfriday Smartypants

anthonyfok:
Can be used in site config or per page front matter:

```
[blackfriday]
smartypants = false
```

----

Hello,

Returning from a long hiatus from Hugo, I am delighted to see all the new nifty features in the new Hugo v0.14/0.15-DEV like the new `genautocomplete` and `gendoc`.

While trying to fine-tune the automatically generated command docs (gendoc) such as http://test.gohugo.io/commands/hugo_genautocomplete/, I tried to make the two hyphens in the `--completetionfile` long option display correctly like so:

> Add **--completionfile=**/path/to/file flag to set alternative file-path and name.

But all I could get was this, where the two hyphens got turned into an en dash:

> Add **&ndash;completionfile=**/path/to/file flag to set alternative file-path and name.

I was trying to avoid use of backticks (`) so the weird backticks won't show up when `hugo help genautocomplete` is run, even though I later changed my mind.

Seeing that the Kubernetes command-line documentation generated by cobra (e.g. http://kubernetes.io/v1.0/docs/user-guide/kubectl/kubectl_config.html) can handle the two hyphens correctly:

> The loading order follows these rules: 1. If the **--kubeconfig** flag is set, then only that file is loaded.

and seeing that GitHub Favored Markdown does not support smart dashes, I thought perhaps there may be cases where the ability to disable smart dashes or even SmartyPants altogether is necessary.

Hence this pull request to allow the end user to configure `blackfriday.HTML_USE_SMARTYPANTS` as necessary.

----

Meanwhile, I have also sent [a pull request to Blackfriday](https://github.com/russross/blackfriday/pull/190) to add `HTML_SMARTYPANTS_DASHES` to allow turning off just the smart dashes while keeping the handy smart quotes, etc.: https://github.com/russross/blackfriday/pull/190.  If that pull request is accepted, I will have another patch for Hugo to expose the new option.  :-)

Many thanks!

### DIFF
--- a/docs/content/overview/configuration.md
+++ b/docs/content/overview/configuration.md
@@ -166,6 +166,16 @@ But Hugo does expose some options---as listed in the table below, matched with t
 
 <tbody>
 <tr>
+<td><code><strong>smartypants</strong></code></td>
+<td><code>true</code></td>
+<td><code>HTML_USE_SMARTYPANTS</code></td>
+</tr>
+<tr>
+<td class="purpose-title">Purpose:</td>
+<td class="purpose-description" colspan="2">Enable enable smart punctuation substitutions.</td>
+</tr>
+
+<tr>
 <td><code><strong>angledQuotes</strong></code></td>
 <td><code>false</code></td>
 <td><code>HTML_SMARTYPANTS_ANGLED_QUOTES</code></td>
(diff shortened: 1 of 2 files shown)


bep:
Merged in 4a2eda4

And welcome back.

PR #1334, opened 2015-08-07
Improve sort template function

tatsushid:
This includes two commit to improve sort template function.

1. Make sort template function return an original array, slice or map's element type slice instead of `[]interface{}` type slice. If it sorts `map[string]int` type value, it returns `[]int` slice value.
2. Make sort template function accept field, method or map key chaining key string as its second argument like where template function doing. It allows to write a template code like
   
   ```
   {{ range sort .Data.Pages "Params.foo.bar" }}
       {{ .Content }}
   {{ end }}
   ```

   to sort pages by these front matter values


bep:
Very nice. Merged.

PR #1340, opened 2015-08-08
Use spf13/pflag's new SetAnnotation helper

anthonyfok:
And make our bash completion code more compact.

Saves 40 bytes in my locally compiled hugo binary!  ;-)


bep:
Merged in fee5519

40 bytes? Oh my! We should send out a press release!{

anthonyfok:
> 40 bytes? Oh my! We should send out a press release!{

:wink:

Thanks for the commit!

PR #1343, opened 2015-08-09
Implement `hugo --theme=[Tab][Tab]` bash completion

anthonyfok:
... using the new `BashCompSubdirsInDir` annotation in http://github.com/spf13/cobra.

Depends on Pull Requests spf13/cobra#126 (Add new BashCompSubdirsInDir annotation)
and spf13/hugo#1340 (Use spf13/pflag's new SetAnnotation helper).

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -112,6 +112,7 @@ func init() {
 	// for Bash autocomplete
 	validConfigFilenames := []string{"json", "js", "yaml", "yml", "toml", "tml"}
 	HugoCmd.PersistentFlags().SetAnnotation("config", cobra.BashCompFilenameExt, validConfigFilenames)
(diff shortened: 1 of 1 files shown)


bep:
Merged in 29c9a40

Now can I have the same in ZSH?

anthonyfok:
> Now can I have the same in ZSH?

Knowing that both you and Steve are avid zsh users, I have started experimenting with ZSH completion a couple weeks ago, but albeit very slowly as other things are keeping me busy.  I am still learning from https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org and similar links.  I think I will probably have a proof-of-concept hand-crafted zsh completion for Hugo ready before getting it implemented in cobra.

Don't hold your breath though.  I am very s-l-o-w, hehe.

PR #1350, opened 2015-08-13
Update press.md

thefln:
New article about Hugo on the Snipcart blog.


anthonyfok:
Great use case and write up, @thefln!  C'est magnifique !

Committed with slight modification ("August" → "Aug") as 35ff685c3099eb90e736d9bbbb65f83e07bebbff (master) and 35ff685c3099eb90e736d9bbbb65f83e07bebbff (v0.14.docs).

PR #1365, opened 2015-08-20
Fix -ldflags for Go 1.5

ascarter:
In Go 1.5, the following warning is given when running the Makefile:

    % make
    go build -ldflags "-X github.com/spf13/hugo/hugolib.CommitHash `git rev-parse --short HEAD 2>/dev/null` -X github.com/spf13/hugo/hugolib.BuildDate `date +%FT%T%z`" -o hugo main.go
    # command-line-arguments
    link: warning: option -X github.com/spf13/hugo/hugolib.CommitHash e791835 may not work in future releases; use -X github.com/spf13/hugo/hugolib.CommitHash=e791835
    link: warning: option -X github.com/spf13/hugo/hugolib.BuildDate 2015-08-20T13:45:25-0700 may not work in future releases; use -X github.com/spf13/hugo/hugolib.BuildDate=2015-08-20T13:45:25-0700

This pull request fixes the form of `ldflags` to be correct for Go 1.5. I believe this works fine on Go 1.4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 
 COMMIT_HASH=`git rev-parse --short HEAD 2>/dev/null`
 BUILD_DATE=`date +%FT%T%z`
-LDFLAGS=-ldflags "-X github.com/spf13/hugo/hugolib.CommitHash ${COMMIT_HASH} -X github.com/spf13/hugo/hugolib.BuildDate ${BUILD_DATE}"
+LDFLAGS=-ldflags "-X github.com/spf13/hugo/hugolib.CommitHash=${COMMIT_HASH} -X github.com/spf13/hugo/hugolib.BuildDate=${BUILD_DATE}"
 
 all: gitinfo
 


bep:
Merged in c2869ae

anthonyfok:
> This pull request fixes the form of ldflags to be correct for Go 1.5. I believe this works fine on Go 1.4

Maybe it is just me, but when I tried to run the Makefile on with Go 1.4.2, it gives an error:

```
$ make
go build -ldflags "-X github.com/spf13/hugo/hugolib.CommitHash=`git rev-parse --short HEAD 2>/dev/null` -X github.com/spf13/hugo/hugolib.BuildDate=`date +%FT%T%z`" -o hugo main.go
# command-line-arguments
usage: 6l [options] main.6
  -1	use alternate profiling code
  -8	assume 64-bit addresses
  -B info
    	define ELF NT_GNU_BUILD_ID note
  -C	check Go calls to C code
  -D addr
    	data address
  -E sym
    	entry symbol
  -I interp
    	set ELF interp
  -L dir
    	add dir to library path
  -H head
    	header type
  -K	add stack underflow checks
  -O	print pc-line tables
  -Q	debug byte-register code gen
  -R rnd
    	address rounding
  -S	check type signatures
  -T addr
    	text address
  -V	print version and exit
  -W	disassemble input
  -X name value
    	define string data
  -Z	clear stack frame on entry
  -a	disassemble output
  -c	dump call graph
  -d	disable dynamic executable
  -extld ld
    	linker to run in external mode
  -extldflags ldflags
    	flags for external linker
  -f	ignore version mismatch
  -g	disable go package data checks
  -installsuffix suffix
    	pkg directory suffix
  -k sym
    	set field tracking symbol
  -linkmode mode
    	set link mode (internal, external, auto)
  -n	dump symbol table
  -o outfile
    	set output file
  -r dir1:dir2:...
    	set ELF dynamic linker search path
  -race
    	enable race detector
  -s	disable symbol table
  -shared
    	generate shared object (implies -linkmode external)
  -tmpdir dir
    	leave temporary files in this directory
  -u	reject unsafe packages
  -v	print link trace
  -w	disable DWARF generation
Makefile:20: recipe for target 'gitinfo' failed
make: *** [gitinfo] Error 2
$
```

I thought of adding a `go version | cut -d' ' -f3` check in the Makefile to run different command accordingly, but then, maybe we should instead encourage everyone to upgrade to Go 1.5 which would eliminate the last remaining race condition in Hugo.  ;-)

ascarter:
@anthonyfok you are right - it doesn't work on 1.4.2. The only solution will be to either require Go 1.5 or to do as you suggest and wrap in a version check.

From `6l` help text:

  -X name value
        define string data

They have changed it in 1.5 in that `6l` is replaced by `link` which now takes:

  -X definition
    	add string value definition of the form importpath.name=value

I think it is up to hugo to decide if 1.5 is going to be required or not.

bep:
I think we say that for the provided make to work, 1.5 is required.

ascarter:
That seems logical - there should be very little reason to not move forward with 1.5. I recommend adjusting the README note that discusses setting the build version.

bep:
I think we're gonna blame the Go team for this one: They promised no breaking API changes,  but this one is so minor that we're not making any fuzz -- and we are not doing some extra work to be backward compatible here. We got enough work on the plate as it is.

PR #1378, opened 2015-08-24
Describe the use of custom archetypes in the docs

digitalcraftsman:
This PR adds a section about the creation and use of custom archetypes. Until now, only the default archetype is mentioned.

If this contributions will be accepted I would need to make an entry in this [forum thread](http://discuss.gohugo.io/t/switching-to-apache-2-license/173/52) so that Steve can switch to a different license, right?

### DIFF
--- a/docs/content/content/archetypes.md
+++ b/docs/content/content/archetypes.md
@@ -56,8 +56,35 @@ Hugo would create the file with the following contents:
 
 We see that the `title` and `date` variables have been added, in addition to the `tags` and `categories` variables which were carried over from `archetype/default.md`.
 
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
I saw a few moments ago that this repo has a branch called 'v0.14.docs ' but my PR's would be merged into the master branch. Is this okay or should create a new PR that can be merged into 'v0.14.docs '?

bep:
It is OK as is. For the docs that is for the release version, we cherry-pick the PR into both branches.

spf13:
Merged as d584ca59f03730431b71f6d1fa9620f7f7fc84fb

PR #1389, opened 2015-08-30
correct check of published boolean

jmhodges:
Deep apologies. I did not write tests before and the last commit that prevented Draft from being updated if both published and draft were set flipped the published boolean twice.

This now has tests.


bep:
Merged in 49fe04c. In the future, please start your commit subject with a capital letter.

Issue #557, opened 2014-10-12
Url is downcased


nono:
Hello,

on http://gohugo.io/content/front-matter/, it is said:

> url The full path to the content from the web root

But it seems it's not really the case. For example, if I put `url: "/Blog/2012/02/29/changement-de-moteur-pour-ce-blog"`, the generated file will be `public/blog/2012/02/29/changement-de-moteur-pour-ce-blog/index.html` (with a blog in downcase).

ThinkChaos:
Can confirm the issue.
hugo's server seems to be case insensitive, so there's no issue when you try locally, but GH pages is case sensitive (IMHO this should be default behavior for hugo), thus breaking the generated site.

PR #1392, opened 2015-09-01
add configuration variable: "disablePathToLower" See #557

coderzh:
…/url lowncase. See #557


bep:
Merged in 52d94fa67578f6b63035e73b236ca8abd40d0006

I modified the commit message slightly. Also, for future reference, it's best to create these pull requests from a separate branch and not from master.

PR #1399, opened 2015-09-03
add RuneSummary for Chinese and Japanese summary Fix #1377

### LINKED ISSUE #1377: Fix summary for Chinese and Japanese
I saw the code "var SummaryLength = 70" in the hugo/helpers/content.go, but in my case, the summary contains all the text in the markdown file except tags(I didn't use the SummaryDivider).

coderzh:
Now can use {{ .RuneSummary }} for Chinese and Japanese , or other languages. Fix #1377

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -168,6 +168,7 @@ func LoadDefaultSettings() {
 	viper.SetDefault("RSSUri", "index.xml")
 	viper.SetDefault("SectionPagesMenu", "")
 	viper.SetDefault("DisablePathToLower", false)
+	viper.SetDefault("HasCJKLanguage", false)
 }
 
 // InitializeConfig initializes a config file with sensible default configuration flags.
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -28,6 +28,7 @@ import (
 	"net/url"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -40,6 +41,10 @@ import (
 	"github.com/spf13/hugo/tpl"
 	jww "github.com/spf13/jwalterweatherman"
 	"github.com/spf13/viper"
+)
+
+var (
+	cjk = regexp.MustCompile(`\p{Han}|\p{Hangul}|\p{Hiragana}|\p{Katakana}`)
 )
 
 type Page struct {
@@ -67,7 +72,6 @@ type Page struct {
 	contentShortCodes   map[string]string
 	plain               string // TODO should be []byte
 	plainWords          []string
-	plainRuneCount      int
 	plainInit           sync.Once
 	plainSecondaryInit  sync.Once
 	renderingConfig     *helpers.Blackfriday
@@ -78,6 +82,7 @@ type Page struct {
 	Node
 	pageMenus     PageMenus
 	pageMenusInit sync.Once
+	isCJKLanguage bool
 }
 
 type Source struct {
@@ -111,30 +116,10 @@ func (p *Page) PlainWords() []string {
 	return p.plainWords
 }
 
-// RuneCount returns the rune count, excluding any whitespace, of the plain content.
-func (p *Page) RuneCount() int {
-	p.initPlainSecondary()
-	return p.plainRuneCount
-}
-
 func (p *Page) initPlain() {
 	p.plainInit.Do(func() {
 		p.plain = helpers.StripHTML(string(p.Content))
 		p.plainWords = strings.Fields(p.plain)
-		return
-	})
-}
-
-func (p *Page) initPlainSecondary() {
-	p.plainSecondaryInit.Do(func() {
-		p.initPlain()
-		runeCount := 0
-		for _, r := range p.plain {
-			if !helpers.IsWhitespace(r) {
-				runeCount++
-			}
-		}
-		p.plainRuneCount = runeCount
 		return
 	})
 }
@@ -218,7 +203,13 @@ func (p *Page) setSummary() {
 	} else {
 		// If hugo defines split:
 		// render, strip html, then split
-		summary, truncated := helpers.TruncateWordsToWholeSentence(p.PlainWords(), helpers.SummaryLength)
+		var summary string
+		var truncated bool
+		if p.isCJKLanguage {
+			summary, truncated = helpers.TruncateWordsByRune(p.PlainWords(), helpers.SummaryLength)
+		} else {
+			summary, truncated = helpers.TruncateWordsToWholeSentence(p.PlainWords(), helpers.SummaryLength)
+		}
 		p.Summary = template.HTML(summary)
 		p.Truncated = truncated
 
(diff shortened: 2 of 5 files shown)


bep:
Can you rebase this against master and squash the two commits into one?

coderzh:
Done. .WordCount and .Summary (70 words) works good for Chinese now.

bep:
Looks good, but could you:

1) Rebase against master so there is only one commit in it.
2) Run go fmt on your code
3) Add "Fixes #issueid" at the end or your commit message.

coderzh:
OK, I will do it tomorrow.

hugolib/page.go
```diff
@@ -362,7 +363,16 @@ func (p *Page) ReadFrom(buf io.Reader) (int64, error) {
...
 
 func (p *Page) analyzePage() {
-	p.WordCount = len(p.PlainWords())
+	p.WordCount = 0
+	for _, word := range p.PlainWords() {
+		runeCount := utf8.RuneCountInString(word)
+		if len(word) == runeCount {
+			p.WordCount++	
+		} else {
+			p.WordCount += runeCount
```
bep: This doesn't look like an universal truth. This may hold water for Chinese, Japanese etc., but will give wrong result for other UTF-8 rich languages. Also, I added the `RuneCount`method on Page for this purpose some time ago.

coderzh: @bep Yes, you are right. There is already a RuneCount method, so I reverted the change about WordCount. And I add RuneSummary method for those people who use Chinese or Janapese.
Ask a question, AppVeyor build failed randomly, do you know why?
Error message like that: 
Two or more menu items have the same name/identifier in "m1" Menu. Identified as "i2".

bep:
Sorry for dragging this one out, but:

* I would love to have as few methods in this area as possible.
* When I thought about the RuneCount vs. WordCount problem earlier, I found them incompatible -- so I added another method. **I would love to be proven wrong here so we can remove the RuneCount method.*
* As to Summary. I'm pretty sure that we can create one method that fits both Western languages and Japanese and Chinese. Having two methods is not optimal for theme authors etc.


/cc @RickCogley

coderzh:
Totally agree. But the reason why I add another RuneSummary method as RuneCount is that:

 * Count words by other language takes a litter more performance impact. With two methods, Western languages users can keep the same performance by the origin method.
 * Summary with fixed words count or make a whole sentence is not as much important as I think. Particular in Chinese language. The important thing is how long the characters takes, how much space it takes in my website. So Summary split the content by rune works for me.

But, I toally agree with you that keep as few method as possible. And I am looking forward to your implement like WordCount, ReadingTime and Summary that fits both Western languages and Japanese and Chinese.

bep:
Let me think a little about this.

spf13:
@bep What do you think about this one?

bep:
@spf13 I think:

1) Having double set of these funcs is bad for theme interop
2) Implementing something like isCJKLanguage() will probably kill performance.

So a configuration flag, maybe? Default to traditional word-counting, but use runes if this flag is set?

@tatsushid any thoughts?

tatsushid:
I think default configuration which is over-writable by frontmatter is reasonable. If someone want to make multi-language site, it should be benefit that someone don't need to modify a template, just changing the content metadata.

In Hugo, we can expect UTF-8 input so detecting including CJK is not so difficult, just checking a specific character code point range is included and less space characters rate of an article but still have the potential of mis-detecting therefore configuration value should be needed if we implement auto-detecting.

bep:
OK, so here is how we are going to do this:

* Introduce a global config flag: `isCJKLanguage`, default `false`.
* For `.Summary`: If `isCJKLanguage` is `true`, use the runes as basis for truncation, else keep as today.
* For `WordCount`:  If `isCJKLanguage` is `true`, use the runes as basis for calculation, else keep as today.
* Unexport `RuneCount`

The caching 

@coderzh is this something you could do?

/cc @RickCogley

coderzh:
@bep OK, I can do it.

spf13:
Is there a friendlier name than that or is CJKLanguage well know enough?

bep:
@spf13 it is a unfriendly term, but much better than isChineseOrJapaneseOrKoreanLanguage.

It is a new term for me (I learn a lot from Hugo), but a common enough term in  internationalization:

https://en.wikipedia.org/wiki/CJK_characters

coderzh:
Done:
 * Auto detect isCJKLanguage by regex
 * add `isCJKLanguage` frontmatter to force specify whether is CJKLanguage or not
 * For .Summary: If isCJKLanguage is true, use the runes as basis for truncation, else keep as today.
 * For WordCount: If isCJKLanguage is true, use the runes as basis for calculation, else keep as today.
 * Unexport RuneCount

coderzh:
I think frontmatter is better than global config about `isCJKLanguage`, because someone may write posts both in English and CJK Languague. And auto detecting is working most of time, event you write main English with a litter Chinese in one article.

bep:
@coderzh Two comments:

1) By global config flag, I meant global - as in `config.toml` / Viper. It may make sense to add an override to Page later, but we have to rethink that one -- I'm not a fan of adding attributes for every new flag.
2) Get rid of the Regexp check.

RickCogley:
My two yen: CJKLanguage is well known to those who deal with those
languages.



On Oct 7, 2015, at 00:35, Steve Francia <notifications@github.com> wrote:

Is there a friendlier name than that or is CJKLanguage well know enough?
On Tue, Oct 6, 2015 at 6:39 AM coderzh <notifications@github.com> wrote:

> @bep <https://github.com/bep> OK, I can do it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/spf13/hugo/pull/1399#issuecomment-145818475>.
>

—
Reply to this email directly or view it on GitHub
<https://github.com/spf13/hugo/pull/1399#issuecomment-145903027>.

coderzh:
@bep agree with most of your opinions. Well, there are two points need to confirm:

 1. Global config flag(config.yml) or FrontMatter or Both, or nether? I don't like adding any more config flag too. I have considered about it, seems that both global config and FrontMatter are helpful(FrontMatter can overwrite the global config). 
 2. Is auto-detecting needed? Or is there a better way to detect CJKLanguage than regex? For performance thinking, should remove auto-detecting, and then check CJKLanguage all by config flag?

spf13:
I think it should be...

1.  a global config saying "has CJKLanguage"

If this flag is enabled it would auto-detect CJK languages in the content
automatically.

2. A frontmatter setting saying explicitly to treat this content as CJKLang.

Thoughts?


--
Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

bep:
The suggestion from @spf13 works. My main concern is that for western-language sites it should behave as today. I don't want Hugo to scan every page for CJK characters that isn't there. (I have spent enough time with premature optimization to want to see that go to waste :-)).

coderzh:
Agree, I think it's a better solution than the others. But after I finished the code, I have done some performance testing, the result was very interesting. 

When the CJKLanguage flag turned on in global settings(means turn on auto-detecting CJKLang), the performance turned out to be a little better! (My Chinese blog)

Here are the performance testing details:

**Test 276 posts(my Chinese blog)** (ms)

CJKLanguage Detect On | Origin Hugo
:-- | :--
433 | 477
436 | 470
434 | 457
455 | 464
453 | 462

**Test 2751 posts** (ms)

CJKLanguage Detect On | Origin Hugo
:-- | :--
3334 | 3528
3353 | 3556
3426 | 3493
3413 | 3441
3378 | 3457

Then, I cloned spf13's blog (English Blog) and do the test again:
git clone https://github.com/spf13/spf13.com

**Test 149 posts(spf13.com)** (ms)

CJKLanguage Detect On | Origin Hugo
:-- | :--
1274 | 1312
1292 | 1301
1338 | 1230
1286 | 1264
1348 | 1288

(All tests on my Windows PC)

The result shows that, CJKLanguage auto-detecting nearly took no performance impact, and even better performance in Chinese posts.(Count and Truncation by runes maybe even quicker)

It will be easier to use with no CJKLanguage config flag, all is automation, any thoughts?

bep:
@coderzh the thing is: the potential penalty will be for non-CJK sites -- and I guess only noticeable for large sites / big content. But that is a common use case for Hugo (spf13's blog is small). If you don't see a penalty hit when comparing `if flag set` to `scan all the bytes in the document for cjk characters`, then you are testing with wrong data. Try Go´s benchmark tool (there are examples in Hugo) for before/after tests.

coderzh:
@bep I pushed the code followed spf13's suggestion.

I copied spf13's posts ten times,  nearly 2241 post files, and run with `hugo benchmark`  A/B testing again. The performance impact is also little. Maybe you can review my code, pull and then test again.

commands/hugo.go
```diff
@@ -168,6 +168,7 @@ func LoadDefaultSettings() {
 	viper.SetDefault("RSSUri", "index.xml")
 	viper.SetDefault("SectionPagesMenu", "")
 	viper.SetDefault("DisablePathToLower", false)
+	viper.SetDefault("IsCJKLanguage", false)
```
bep: Should be: "HasCJKLanguage" (according to spf13, and I agree)

coderzh: Already changed to "HasCJKLanguage"

bep:
Very good. Merged in 8233348

A line or two for the documentation would be welcomed.

PR #1402, opened 2015-09-04
Update organization.md

igregson:
Increases clarity on the different between `section` and `type`. 

The current `section` information here is wrong (sections can *not* be specified in front matter). This caused quite the headache. This change fixes this and also adds `type`, since it *can* be specified in front matter.

See here in forum: http://discuss.gohugo.io/t/type-vs-sections-whats-the-difference/178/5


spf13:
merged as d6f694552512f67dc7bc55afd660de017641fb58

PR #1403, opened 2015-09-04
Add a 404.html file if a new theme is created

### LINKED ISSUE #1372: Add 404 template by default if hugo new theme is used
If a user wants to create a new theme with Hugo's own command `hugo new theme <themeName>` a(n empty) default template for a 404 page should be added.

By running `hugo server` an empty `404.html` file will be added automatically to the `public` folder, from which Hugo serves the site. 

This addition just should remember theme creators to integrate a 404 page into their design.

digitalcraftsman:
This PR would close #1372


spf13:
Merged as fd7b315

PR #1408, opened 2015-09-06
Docs: Add new french mag article by @humboldtux to Press page

humboldtux:
An article i wrote for a french Linux magazine and published before summer.


spf13:
Merged as c7b701e

PR #1413, opened 2015-09-07
Added new PygmentsOptions setting. Fixes #1260

### LINKED ISSUE #1260: Default Pygments options
You can configure Pygments via PygmentsStyle and PygmentsUseClasses, but there are other options I want to set. For example, I may want to set linenos to some default for the entire blog. I don't want to set this in every shortcode block, and I'm not able to configure it for Github style code fences. As such I think there should be a way to set a default for the site.

My proposal is a new PygmentsOptions setting, that is a string such as `style=%s,linenos=true,startinline=true`. These options are always passed to pygments, in addition to any others that might be specified in the shortcode. To retain some backwards compatibility, we will allow PygmentsStyle and PygmentsUseClasses (if set) to override the specific options within the PygmentsOptions.

Finally, I noticed that currently the [code checks](https://github.com/spf13/hugo/blob/master/helpers/pygments.go#L156) if the options passed to pygment are one of "style, encoding, noclasses, hl_lines, linenos or classprefix". That actually limits the use of pygments, and I'm wondering why that restriction is there? If there is no objection, I would prefer to remove those checks, and instead depend on the return code of the pygment command.

bramp:
This is the global configuration string that is sent to pygments. Specific configurations can be overridden by the existing PygmentsStyle and PygmentsUseClasses options, as well as arguments to the Code Shortcode block.


bep:
Can you fix the commit messages?

bep:
@bramp can you follow the link to the CLA and sign it? (just push "Details")

bramp:
Sorry for my slow responses to fixing up some of these PRs. I would be happy to sign a CLA, but my employer needs to approve. I will follow up shortly.

bep:
Merged in c3931ef

PR #1417, opened 2015-09-10
Leave aliases unsanitized, allowing whitespace etc.

### LINKED ISSUE #701: Add support for alias with whitespace
possibly being a bit thick, but am having a bit of trouble making an alias for a page that had a space in it.

e.g. the old page was /wiki/Some Page

aliases = [ "/wiki/Some Page" ]

makes a page redirect for /wiki/Some-Page

aliases = [ "/wiki/Some+Page" ]

makes a page redirect for /wiki/SomePage

how could I create the right alias?

thanks!

anthonyfok:
so that user-created aliases like these:

    aliases = [ "/wiki/Some Page/", "/wiki/Some+Page/" ]

would work as expected.

Fixes #701: Add support for alias with whitespace

### DIFF
--- a/target/alias_test.go
+++ b/target/alias_test.go
@@ -16,11 +16,11 @@ func TestHTMLRedirectAlias(t *testing.T) {
 		{"", ""},
 		{"s", filepath.FromSlash("s/index.html")},
 		{"/", filepath.FromSlash("/index.html")},
-		{"alias 1", filepath.FromSlash("alias-1/index.html")},
-		{"alias 2/", filepath.FromSlash("alias-2/index.html")},
-		{"alias 3.html", "alias-3.html"},
+		{"alias 1", filepath.FromSlash("alias 1/index.html")},
+		{"alias 2/", filepath.FromSlash("alias 2/index.html")},
+		{"alias 3.html", "alias 3.html"},
 		{"alias4.html", "alias4.html"},
-		{"/alias 5.html", filepath.FromSlash("/alias-5.html")},
+		{"/alias 5.html", filepath.FromSlash("/alias 5.html")},
 		{"/трям.html", filepath.FromSlash("/трям.html")},
 	}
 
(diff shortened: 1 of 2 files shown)


anthonyfok:
This Pull Request, at least in this current early form, is more for facilitating discussions and testing for Issue #701 than for committing into `master`.  Even though this patch does not cause Issue #1418, I think we need some more checks of an alias path to prevent any potential security hole.

marekr:
What would happen on Windows if you try to create an alias with an question mark in its path which is fine on Linux but not on Windows. Would golang or hugo barf cleanly?

anthonyfok:
Even if you could add a question mark `?` to a filename, a webserver would treat whatever comes after `?` as query string and would not be able to find the static filename that you created.

In cases like this, some kind of rewrite rule is needed for the web server anyway.  For example, something like this worked for me (Apache 2.4):

```
RewriteEngine On
RewriteCond %{ENV:REDIRECT_STATUS} =""
RewriteCond %{QUERY_STRING} !=""
RewriteRule ^(.*)$ $1\%3F%{QUERY_STRING} [L]
```

Or, for this to work on Windows platform, it would be easier to just change the `?` to an `_`  (underscore), for example, as provided by the answer in http://stackoverflow.com/questions/4033546/apache-allow-filenames-with

(I simply changed `_` to `\%3F` to see if it works, and it does in Linux.)

For curiosity sake, what platform are you using?  And what kind of aliases with query string that you want to create?

marekr:
I don't want aliases with query strings, I'm simply just pointing out the Windows platform may have funny behavior if the bad characters aren't identified and a message thrown.

 I myself use Windows for testing but the final server runs Linux but it doesn't use apache. The current behavior of HTML file generation is actually really convenient for redirects rather than rules because of that.

anthonyfok:
Thanks for your advice, @marekr.

I have just made pull request #1427, which supersedes this one, with added validation.

PR #1427, opened 2015-09-13
Validate aliases to prevent directory traversal etc.

### LINKED ISSUE #701: Add support for alias with whitespace

### LINKED ISSUE #1418: Add validation for alias
Following some tests related to the discussions at Issue #701, I came to realize that Hugo would happily create aliases (redirections) outside of the `public/` directory if a content page contains aliases that traverse up parent directories.  For example:

    aliases = [ "/some-page/", "../some-page-2/", "../../some-page-3/" ]

After a Hugo run, `some-page-2/` would be created in the same parent directory as `public/`, whereas `some-page-3/index.html` would be created outside of the web site source directory altogether.

At least Hugo v0.13, v0.14, as well as the current HEAD are affected.  (I did not test v0.12.)

anthonyfok:
Add validation before creating aliases:

 * Prevent creating aliases outside webroot (public/ dir)
 * Skip empty "" alias
 * Skip "/" → "/index.html", which gets overwritten anyway
 * Refuse to create Windows-invalid filenames on Windows;
   warn on other platforms
 * In case of invalid aliases, after skipping them,
   return `err = nil` to prevent the error passing up
   all the way to `hugolib.Render()` and causing Hugo to abort.
 * Update alias tests.

Fixes #701: Add support for alias with whitespace
Fixes #1418: Add validation for alias

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -242,6 +242,7 @@ func (s *Site) Build() (err error) {
 		return
 	}
 	if err = s.Render(); err != nil {
+		// Better reporting when the template is missing (commit 2bbecc7b)
 		jww.ERROR.Printf("Error rendering site: %s\nAvailable templates:\n", err)
 		for _, template := range s.Tmpl.Templates() {
 			jww.ERROR.Printf("\t%s\n", template.Name())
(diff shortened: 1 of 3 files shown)


bep:
Merged in e71bef7

PR #1431, opened 2015-09-14
Add dynamic stylesheets based on the config

digitalcraftsman:


parser/css.go
```diff
@@ -0,0 +1,64 @@
...
+		return
+	}
+
+	err = t.Execute(f, nil)
+	if err != nil {
+		jww.WARN.Println("Can't interpret the CSS file: ", err)
+		return
+	}
+
+	f.Close()
```
bep: Do a `defer f.close()` right after you open it, to make sure it gets closed if you return earlier.

digitalcraftsman:
@bep Currently I've two problems to solve.

1. In [line 31](https://github.com/spf13/hugo/pull/1431/files#diff-092b76ef7e07d6f27aaa7bccc04a39b9R31) I need to pass config as second parameter. The user should be able to set something like custom colors in the under `params` in the `config.toml`. For example:

```toml
[params]
themeColor = "#abcdef"
```

Now I need to pass `params` as argument to function that executes the template. Inside the template I would like to call the variable `themeColor` like here:

#### style.css

```css
a {
    color: {{ .themeColor }}
}
```

In order to get the data of `params` I used `viper.Get("params")` but I always got an empty map.

2. Next, I need to find a spot where I can call my `ParseCSSFiles` function. Before this happens it's important that all static files are available in the `public` folder. Then the functions searches css files, parses it and saves the executed template back to the `public` folder. That ensures, that the original stylesheets aren't modified.

I thought about [this location](https://github.com/spf13/hugo/blob/f5308da32013279664379111d84fb8fa50d7e5f3/commands/hugo.go#L312) to call `ParseCSSFiles()`.

parser/css.go
digitalcraftsman: Done.

bep:
There are plenty of existing test cases that set ups a "mock" site with Viper config. Have a look in the /hugolib package.

digitalcraftsman:
Everything works fine now. Now I need to add tests. So, for `parse()` I could mock the `params` section but whats the best way to test `ParseCSSFiles()` and `filterFiles()`? Should I create a temporary file structure that contains all sorts of files?

digitalcraftsman:
I added documentation and a test function for `parse()`.

spf13:
I love this idea and the implementation is solid.

I feel that the CSS params should be kept separate from the site params. 

May be as simple as reading under the keyword `css`

eg.

 +```toml
 +[params.css]
 +    fontFamily = "Arial, Verdana, sans-serif"
 +    textColor  = "#6BC825"
 +```
 +
 +***

@digitalcraftsman thoughts?

digitalcraftsman:
@spf13 That's a good idea. I've rebased this change into the commit.

But the current state isn't mergable. You integrated in the latest release the option to render the pages (by default) to memory instead to the disk. My current approach tries to walk a directory on disk that doesn't exist.

Furthermore, in this case the `publishDir` will be set to `/` and the walker checks recursively each file from the root. Not a very good combination :wink: 

```go
if !renderToDisk {
	hugofs.DestinationFS = new(afero.MemMapFs)
	// Rendering to memoryFS, publish to Root regardless of publishDir.
	viper.Set("PublishDir", "/")
}
```

Does afero provide a similar function to do something similar?

digitalcraftsman:
@spf13 should I add an option that skips the default parsing of all CSS files? Some users don't need this or use this feature and it just drecreases the rendering performance.

spf13:
Anything in static shouldn't be touched. Things in /content are for
processing. We could give it a different extension which is probably the
best thing to do. Maybe styles.dcss, then add a handler for those file
types. If you do that you don't need to walk anything, all the plumbing is
already in place.

digitalcraftsman:
This would avoid a lot of problems, but do you think it is a good idea to mix content files with assets that usually belong to the static folder (even if they're not really static anymore)?

spf13:
Sometimes yes, sometimes no.

I think we need a better way of handling things than we have now. Our
current model is a bit overextended.

Read my reply to the 1.0 thread. I think folders representing the entirety
of a post and its media should be together.

Sites today are comprised of

Templates
Source content
Media
Static assets
Dynamic assets
Misc static files
Data files

And probably more I'm not considering.

Our model is really only initially designed for half of these with others
added on over time.

I've considered adding an assets folder at the root for dynamic assets and
rewriting the content handling to only handle media and content. I think
this is probably the cleanest approach, though it would be a bit breaking,
I don't think many use the current feature anyway and it would be simple to
just move assets to the new folder.

I do think this is probably the best approach to take.

I'm also curious how others (hexo, pelican) handle this.

digitalcraftsman:
I browsed the docs of other static site generators to find out how they solve this problem.


- Hexo  can be extended with CSS preprocessors via plugins that allows users easily to change global variables in a single place.
- Jekyll seems to process every file that contains a frontmatter like a template. I found [this issue](https://github.com/jekyll/jekyll/issues/3290) about exact this question and an [live example](https://github.com/skeeto/skeeto.github.com/blob/19fde92d32a7f645df3020706afb07d29631b36f/css/light.css)
- Pelican doesn't seem to offer such a feature
- Middleman is as extensible as Jekyll and could be capable to do this with plugins.

However, this feature could become very powerful in combination with CSS pre-processors like Less, Sass or Styles. 

As you suggested, we could create a new assests folder in the root directory and parse thoose files as templates that have access a limited variable scope in the config file, like the suggested `[params.css]` scope.

But where should these files be finally stored? In a next step, users could use CSS pre-processors to compile the stylesheets to regular CSS. We shouldn't mix the original stylesheets with those that were interpreted by Hugo.

bep:
Thinking a little bit general, what if we introduce an assets section in the configuration:

Pseudo-code:

```
asset:
   type: css
   input: ...
   filePattern: ...?
   actions: [template-parser, ...] (chained)
   destination: (default is static)

```

Someone have probably thought of a syntax for something similar out there ...

digitalcraftsman:
An assets section could make this process a bit more flexible. @bep, could you explain some variables in your pseudo-code? 

I tried to mock it a bit further in TOML.

```toml
[[assets]]
   # With type you think we could expand this to other types like JS files? 
   # If yes, the assets section itself should be an array
   type = "css"
   # Are these the input directories?
   input = ["path1", "path2"]
   # Do you think of an array of file extensions or about something
   # more complex like a regex pattern search?
   filePattern = ["scss", "less", "stylus"]
   # Internal actions (chained) that are built-in into Hugo?
   # This would be wisely thought if plugins are some day a part of workflows
   # Back to the present: which other actions do you have in mind?
   actions =  ["template-parser", "foo", "bar"]
   # Defaults to static. But could be a temporary folder for pre-processors that
   # are watched by Grunt/Gulp
   destination = "static"
```

bep:
I haven't thought this through to its fullest, but:

* Yes, assets should be a list (css for now, but js ... custom plugins in the future).
* Common thing is that they have a input and output
* On the edges (first and last in a chain) these will be a set of files from/to a directory. 
* Filepattern can be a regexp i guess, default = ".*" 
* To make it simple all paths are relative to workDir and one input and one output. If more paths are needed, define another asset.
* Action will first be a predefined set of actions (template processor), but we should define a good interface for this so we could expand it to user-defined actions (plugins) in the future.

spf13:
As I think about this more, I think the Hexo approach makes some sense and this should be part of the Asset solution. We need to implement that solution first.

fzyukio:
Any ETA when this is feature is available?
Currently I have to embed part of the CSS into HTML in order to use variables such as colours... And I hate this approach.

bep:
I think we're going to close this. A page with output format CSS should solve this. If that has issues, we should fix that.

anagio:
This would be a nice format to follow if we were able to configure asset types in the config
http://pre-commit.com/#new-hooks

The format of custom hooks is similar to middleman's external pipeline config.

PR #1432, opened 2015-09-15
Add base64Decode and base64Encode template functions

digitalcraftsman:


digitalcraftsman:
@bep You suggested to use `interface{}` as args type. Now I tried the following peace of code inside a template:

    {{ 22 | base64Decode}}

and got the following error message:

    executing "theme/_default/single.html" at <base64Decode>: error calling base64Decode: illegal base64 data at input byte 0

It seems that the integer isn't accepted as input regardless if the argument type is `interface{}`.

bep:
@digitalcraftsman was it a question in there somewhere?

digitalcraftsman:
Indeed, I've a small question. You suggested to use an `interface{}` as the [argument's type](https://github.com/spf13/hugo/pull/1432/files#diff-198cba9fba15ab440423625652e916baR1321) for the functions instead of a string. This change would make the template function more flexible if for example an integer is passed which could be casted to a string.

And that is exactly what I tried in a template. I passed an integer but the casting doesn't seem to work.

digitalcraftsman:
All commits have squashed into a single one. The pull request should be ready to be merged unless someone hsa something to complain.

bep:
Merged in 79f8bb625d2a7379d8eb0afa7e144444b4567e3d

Quality PR, with docs and all.

PR #1439, opened 2015-09-18
Add template func join [WIP]

digitalcraftsman:
This PR adds a template function that joins a slice of strings - a seperator can be defined. The only problem is that the function fails silently if the slice's values aren't strings (`[]string`).

TODO:
- Add documentation


bep:
What does this do that the delimit func doesn't do?

digitalcraftsman:
I've overseen `delimit` in the docs.

PR #1440, opened 2015-09-18
Add template funcs countwords and countrunes

digitalcraftsman:
Takes any type, tries to conver it into a string and returns the total number of words.

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -478,6 +478,26 @@ Converts all characters in string to uppercase.
 e.g. `{{upper "BatMan"}}` → "BATMAN"
 
 
+### countwords
+
+`countwords` tries to convert the passed content to a string and counts each word
+in it. The template functions works similar to [.WordCount]({{< relref "templates/variables.md#page-variables" >}}).
+
+```html
+{{ "Hugo is a static site generator." | countwords }}
+<!-- outputs a content length of 6 words.  -->
+```
+
+
+### countrunes
+
+Alternatively to counting all words , `countrunes` determines the number  of runes in the content and excludes any whitespace. This can become useful if you have to deal with
+CJK-like languages.
+
+```html
+{{ "Hello, 世界" | countrunes }}
+<!-- outputs a content length of 8 runes. -->
+```
 
 
 ## URLs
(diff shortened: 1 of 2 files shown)


docs/content/templates/functions.md
```diff
@@ -460,18 +524,25 @@ e.g.
 * `{{substr "BatMan" 0 -3}}` → "Bat"
 * `{{substr "BatMan" 3 3}}` → "Man"
 
+
+***
```
bep: What's all these asterisks?

digitalcraftsman:
I revisited the functions and aspects @bep mentioned in his comments. To support CJK languages I added the `countrunes` function.

Both `countwords` and `countrunes` behave like `.WordCount` and `.RuneCount`. In a nutshell, I extracted the code that both page variables use to determine the content length, wrapped them into template functions and tried different text snippets.

Usually all functions have a test case to check if they work properly. Since both template functions use the same code as `.WordCount` and `.RuneCount` I looked at their testing functions.

A [quick look](https://github.com/spf13/hugo/blob/49fe04c0bd8111bf686d9205d543f8651ea24cfc/hugolib/page_test.go#L585) at them showed that they used sample content defined in constants. I used them to test my template functions manually and compared the result with the exptected values in the existing test functions and got the same results. So, should I add testing functions anyway?

Finally, I removed the asterisks that were initially added as spacers between the functions to get a better overview.

digitalcraftsman:
Is this PR in a mergable state or is something missing?

bep:
It is mergable -- but I wonder: What is the use case for this? When do you need to count words?

digitalcraftsman:
I thought especially about external ressources that you can fetch with `getJSON`.

digitalcraftsman:
Is there any interest for such template functions or should I close this issue?

/cc @bep

spf13:
@digitalcraftsman @bep I'm fine with this code. I'm not sure what the valid use is for it though. If anyone has one, happy to merge. I'd be surprised if someone doesn't have one for it and it doesn't really add any weight to have this functionality.

bep:
I was going to say documentation, but I notice @digitalcraftsman has supplied perfect docs for this, so yes -- I don't need this, but I'm sure someone will. I'm fine with this.

spf13:
merged as dfa34af

PR #1446, opened 2015-09-20
one more deadlink -

vsopvsop:
http://www.sglyon.com/blog/2014/creating-the-site/


anthonyfok:
@vsopvsop, thank you again for reporting the dead links on Press page.
Clearing out dead links certainly improves the web browsing experience for new Hugo users and alike.

It turns out that Spencer Lyon has move his blog to a new domain, so I have updated the URL accordingly.  Ditto for Kenny Woo's article.  Jason Yee's... I can't find anything, looks like his site is down for good for the time being.

Thanks again!

Cheers,
Anthony

PR #1449, opened 2015-09-22
Change port number from 1414 to 1313 (default)

Spike35031:
Change the port number to access the site locally from 1414 to 1313 as per the default and the image.


anthonyfok:
@Spike35031, thanks for catching that discrepancy!

Merged as a43b037651328a94092bc054cbb5d0211dee7b1c (master)
and 98b5c3e2e11c880581207928b57f333bebe345b5 (v0.14.docs).

PR #1450, opened 2015-09-22
Add template func numFormat

### LINKED ISSUE #1444: Outputting numbers with thousands separators
Would it perhaps be possible to add thousand separators to Hugo?

We can already format dates, and I think it will be valuable if we can format numbers too. For example, instead of printing 2354098 formatting it to 2,354,098 makes it easier to read.

Also see the [forum topic](http://discuss.gohugo.io/t/outputting-numbers-with-thousand-separators/1740) where Bep already linked to a code snippet for this with a very flexible license. :)

digitalcraftsman:
numFormat can format floats and integers with custom thousands seperators and decimal seperators. Users can also the define the precision up to 9 digits after the comma.

Fixes #1444.

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1727,6 +1727,17 @@ func sha1(in interface{}) (string, error) {
 	return hex.EncodeToString(hash[:]), nil
 }
 
+// numFormat can format floats and integers with custom thousands seperators
+// and decimal seperators.
+func numFormat(format string, num interface{}) (string, error) {
+	switch conv := num.(type) {
+	case int, int8, int16, int32, int64, float32, float64:
+		return renderNumber(format, conv.(float64))
+	default:
+		return "", fmt.Errorf("%T is an invalid type and can't be formatted.", num)
+	}
+}
+
 func init() {
 	funcMap = template.FuncMap{
 		"absURL":       func(a string) template.HTML { return template.HTML(helpers.AbsURL(a)) },
@@ -1748,6 +1759,7 @@ func init() {
 		"eq":           eq,
 		"findRE":       findRE,
 		"first":        first,
+		"numFormat":    numFormat,
 		"ge":           ge,
 		"getCSV":       getCSV,
 		"getJSON":      getJSON,
--- a/tpl/template_funcs_test.go
+++ b/tpl/template_funcs_test.go
@@ -90,6 +90,7 @@ div: {{div 6 3}}
 emojify: {{ "I :heart: Hugo" | emojify }}
 eq: {{ if eq .Section "blog" }}current{{ end }}
 findRE: {{ findRE "[G|g]o" "Hugo is a static side generator written in Go." 1 }}
+numFormat: {{ numFormat "#,###" 12345.6789 }}
 hasPrefix 1: {{ hasPrefix "Hugo" "Hu" }}
 hasPrefix 2: {{ hasPrefix "Hugo" "Fu" }}
 in: {{ if in "this string contains a substring" "substring" }}Substring found!{{ end }}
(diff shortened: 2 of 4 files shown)


digitalcraftsman:
The template function is currently called `formatnum`. Does someone instead prefer a shorthand like `numf` similar to `Errorf` and `Printf` of the `fmt` package?

ariejan:
`Numf` sounds nice.

digitalcraftsman:
@ariejan That's already possible. I used this [github gist](https://gist.github.com/gorhill/5285193#file-render_number-go-L19-L26) for the implementation. The highlighted lines give you some examples for formating options. 

And regarding the name: it would be interesting what @bep, @spf13 or the others think about `formatnum` vs `numf`.

bep:
I'm not totally married to having to maintain this code - it is a much needed functionality, but it should be in or near Go's stdlib. In any case, it shouldn't live in tempalte_funcs.go, that is too crowded as is.

digitalcraftsman:
I can understand that it means extra work if a bug occurs. If you stick with this code, let it me know and I revisit the code.

Another option would be [go-humanize](https://github.com/dustin/go-humanize). It seems to be widely used (according to the repo stars) and has a good test coverage. On the one hand would this package give as even more options to format content. On the other hand does this mean to add another dependency.

bep:
My main concern is support. It's nice to just point the finger at a well documented project, in stead of having to dig into the code.

digitalcraftsman:
The package was just meant as an alternative. If you're focused on the script of the Gist regarding support, then you've just two options: either we dig into the code and maintain it if neccessary or we exclude the Gist as option entirely.

bep:
To conclude: The code looks clean enough, but we're not staffed to maintain/support it.

digitalcraftsman:
I've moved the helper functions to render the numbers into their own file.

tpl/template_funcs.go
```diff
@@ -1317,6 +1317,17 @@ func ModBool(a, b interface{}) (bool, error) {
...
 
+func FormatNum(format string, num interface{}) (string, error) {
+	switch conv := num.(type) {
+	case int:
+		return renderInteger(format, conv)
+	case float64:
+		return renderFloat(format, conv)
+	default:
+		return "", fmt.Errorf("Invalid type to format.")
+	}
```
moorereason: Couple questions:  shouldn't this switch handle all int and float types? And why have `renderInteger` when you just work with floats anyway?  Just use a single func and do the type casting here.

```go
switch conv := num.(type) {
case int, int8, int16, int32, int64, float32, float64:
    return renderNumber(format, conv.(float64))
default:
    return "", fmt.Errorf("Invalid type to format.")
}
```

tpl/template_funcs.go
```diff
@@ -1317,6 +1317,17 @@ func ModBool(a, b interface{}) (bool, error) {
 	return res == int64(0), nil
 }
 
+func FormatNum(format string, num interface{}) (string, error) {
```
moorereason: Make this a private func and add a comment describing it.

tpl/template_funcs_test.go
```diff
@@ -1581,3 +1581,30 @@ func TestSafeURL(t *testing.T) {
...
+		{"#,###.##", "12,345.68"},
+		{"#,###.", "12,346"},
+		{"#,###", "12345,679"},
+		{"#.###,######", "12.345,678900"},
+	} {
+		formNum, err := FormatNum(this.format, testNum)
+
+		if err != nil {
+			t.Errorf("Formating %v with [%s] as format caused an error: %s",
+				testNum, this.format, err)
```
moorereason: Add a `continue` after this line.

digitalcraftsman:
Closing this in favor of #2851

PR #1459, opened 2015-09-25
Add force flag to init new site inside non empty dir

ns3777k:
Add force flag

If flag is passed the site will be created inside non-empty folder
only if there are no existent folders or config with the same name
hugo creates.

Resolves: https://github.com/spf13/hugo/issues/1163

### DIFF
--- a/commands/new.go
+++ b/commands/new.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	"errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/hugo/create"
 	"github.com/spf13/hugo/helpers"
@@ -37,6 +38,7 @@ var contentFrontMatter string
 
 func init() {
 	newSiteCmd.Flags().StringVarP(&configFormat, "format", "f", "toml", "config & frontmatter format")
+	newSiteCmd.Flags().Bool("force", false, "Init inside non-empty directory")
 	newCmd.Flags().StringVarP(&configFormat, "format", "f", "toml", "frontmatter format")
 	newCmd.Flags().StringVarP(&contentType, "kind", "k", "", "Content type to create")
 	newCmd.AddCommand(newSiteCmd)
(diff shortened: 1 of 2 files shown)


bep:
What do you mean by "x,y,z"?

Also, could you get your commit message in line with the guidelines?

commands/new.go
```diff
@@ -117,13 +119,24 @@ func NewSite(cmd *cobra.Command, args []string) {
...
+		if isDir, _ := helpers.IsDir(createpath, hugofs.SourceFs); !isDir {
+			jww.FATAL.Fatalln(createpath, "already exists but not a directory")
+		}
+
+		if isEmpty, _ := helpers.IsEmpty(createpath, hugofs.SourceFs); !isEmpty && !forceNew {
 			jww.FATAL.Fatalln(createpath, "already exists and is not empty")
 		}
+
+		if forceNew {
+			if err = os.RemoveAll(createpath); err != nil {
```
bep: Remove all? No, this isn't what the user would expect.

ns3777k: Only if you pass *force* flag. Can make a make up

bep: The flag help text says "create even if destination is not empty"". It says nothing about deleting stuff. And rightfully so.

ns3777k: I edited flag help text & commit message.
The variables inside the method were called like "x", "y" and "z".
I renamed them to increase readability.

anthonyfok:
Hi @ns3777k,

I think even with the `--force` flag, it should not remove _any_ files or directories.
If it needs to overwrite any existing files, e.g. an existing `config.toml`, it should either abort or prompt the end user what to do.

Otherwise, this `--force` that you currently have is a "killer"---no, an "assassin" command that may seem innocuous to the end user but end up having his/her entire home directory deleted when he/she mistakenly type `hugo --force new site .`

bep:
Agree with @anthonyfok No deletions. Not even after prompting the user. I do not want to be responsible for wiping out peoples stuff.

ns3777k:
Misunderstood the issue. Got you. Get done on Sunday.

ns3777k:
Edited so there's no prompt or deletion.

spf13:
@ns3777k Thanks for this. The users will really benefit from it. It would be great to have a test accompanying it since it has gotten a bit complex.

ns3777k:
@spf13 Done.
Not sure If I chose the right way to make it testable so any comments are welcome :-)

spf13:
@ns3777k Please click on the "Details" link to the right of the license/cla line and click "approve" so we can merge this. The CLA isn't new, but this integration is. 

Thanks!

ns3777k:
@spf13 Approved. Still pending here though.

bep:
Merged in 5e97cf3020284859ddaba528801329a54cf95847

PR #1461, opened 2015-09-27
is it .md instead of .ad?

vsopvsop:


spf13:
Thanks! Merged as ead339a

bep:
Actually, `.ad` (Asciidoc) is supported, too, but md is less confusing in the doc.

PR #1463, opened 2015-09-28
Added dictionary function that could be used to pass maps into a template.

notzippy:
Allows templates to dynamically build maps.
Example usage: Creating and passing a map with the .Site information to a subtemplate while in a range on the parent

    {{$baseurl := .Site.BaseURL }}
    {{range .Site.Params.Bar}}
        {{partial "foo" (dict "foo" . "BaseURL" $baseurl)}}
    {{end}}

### DIFF
--- a/tpl/template_funcs_test.go
+++ b/tpl/template_funcs_test.go
@@ -324,6 +324,30 @@ func TestAfter(t *testing.T) {
 	}
 }
 
+func TestDictionary(t *testing.T) {
+	for i, this := range []struct {
+		v1     []interface{}
+		expecterr bool
+		expectedValue map[string] interface{}
+	}{
+		{[]interface{}{"a", "b"}, false, map[string]interface{}{"a":"b"}},
+		{[]interface{}{5, "b"}, true,nil},
+		{[]interface{}{"a", 12,"b",[]int{4}}, false,map[string]interface{}{"a":12,"b":[]int{4}}},
+		{[]interface{}{"a", "b", "c"}, true,nil},
+	} {
+		r,e := Dictionary(this.v1...)
+
+		if (this.expecterr && e==nil) || (!this.expecterr && e!=nil)  {
+			t.Errorf("[%d] got an unexpected error", i, e, this.expecterr)
+		} else if !this.expecterr {
+			if !reflect.DeepEqual(r, this.expectedValue) {
+				t.Errorf("[%d] got %v but expected %v", i, r, this.expectedValue)
+			}
+		}
+	}
+}
+
+
 func TestIn(t *testing.T) {
 	for i, this := range []struct {
 		v1     interface{}
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -50,6 +50,29 @@ e.g.
 
     // Outputs Tags: tag1, tag2 and tag3
 
+### dict
+Creates a dictionary (map[string, interface{}), expects parameters added in value:object fasion.
+Invalid combinations like keys that are not strings or uneven number of parameters, will result in an exception thrown
+Useful for passing maps to partials when adding to a template.
+
+e.g. Pass into "foo.html" a map with the keys "important, content" 
+
+    {{$important := .Site.Params.SomethingImportant }}
+    {{range .Site.Params.Bar}}
+        {{partial "foo" (dict "content" . "important" $important)}}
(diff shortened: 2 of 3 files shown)


bep:
Excellent idea. It would be nice with better test coverage and some lines for the documentation.

notzippy:
@bep 
The test covers the two exception cases (improper number of parameters, improper type of leading parameter), and two different valid cases. What additional tests should be performed ?

For documentation do you mean for the Dictionary function itself ? I did not see any docs for any of the template functions in that file so I did not include anything. If it needs to be added I can add some..

spf13:
Please add documentation to http://gohugo.io/templates/functions/ .

The source is at https://github.com/spf13/hugo/edit/master/docs/content/templates/functions.md

notzippy:
Updated pull request with a description of the function

bep:
@notzippy as I read the tests, it only checks for the presence or absence of an error. There are no assertions about correct output from the success cases.

notzippy:
@bep Good idea, I added an additional parameter for the expected result and checked it with a `reflect.DeepEqual` call on the result (when there was no error expected).

halostatue:
:+1:

tpl/template_funcs_test.go
```diff
@@ -324,6 +324,30 @@ func TestAfter(t *testing.T) {
...
+		{[]interface{}{5, "b"}, true,nil},
+		{[]interface{}{"a", 12,"b",[]int{1, 2, 4}}, false,map[string]interface{}{"a":12,"b":[]int{1, 2, 4}}},
+		{[]interface{}{"a", "b", "c"}, true,nil},
+	} {
+		r,e := Dictionary(this.v1...)
+
+		if (this.expecterr && e==nil) || (!this.expecterr && e!=nil)  {
+			t.Errorf("[%d] got an unexpected error", i, e, this.expecterr)
+		} else if !this.expecterr {
+			if !reflect.DeepEqual(r, this.expectedValue) {
```
bep: This will fail in some cases, as you cannot rely on the key order in Go maps. There are examples of ways to handle this in other tests, some less elegant than others. I believe I wrote an util func to do this, but cannot find it now.

notzippy: IMHO if the  `reflect.DeepEqual` function fails to match when returned order from a slice does not match another slice then that is the fault of `reflect.DeepEqual`. However I am not here to argue the semantics of how DeepEqual should work :-) So I just changed it to have a single array element.

bep: If you have issues with `reflect.DeepEqual`, take it up with the Go team. But believe me when I say this. I have been bitten by this myself.

notzippy: `reflect.DeepEqual` checks maps based on the keys https://golang.org/src/reflect/deepequal.go?s=3332:3371#L125 , perhaps in older versions of go you may have encountered this but it is not true anymore. I can reverse the order of the map and the tests still pass.

bep: The order of the keys is not guaranteed. This has been the situation since Go 1.1. It may work 999 out of 1000 times, maybe all the time on Windows, but some times on Linux ... my point is: Having shaky tests is a bad solution. Esp. when the workaround is known: Sort the map before doing the compare. 

I'm not debating this. Fix it if you want.

chipsenkbeil:
:+1:

tpl/template_funcs_test.go
notzippy: @bep This is the code in DeepEqual for comparing a map in deep equal it does *not* use the order of the keys to make the comparison, it iterates over the map keys and fetches the value per key. 

			for _, k := range v1.MapKeys() {
				if !deepValueEqual(v1.MapIndex(k), v2.MapIndex(k), visited, depth+1) {
					return false
				}
			}

bep: You are right. Thanks for sticking to your story. Sorry about that. If you haven't, could you "put your name" in this license thread:

http://discuss.gohugo.io/t/switching-to-apache-2-license/173

notzippy: Cool, all good man. Have signed the release form..

bep:
Merged in 3a27cef

I modified the commit message slightly, to fit in with our (pedantic) guidelines.

This will make many users happy. And I guess it will also close some issues ... have to track those down.

PR #1466, opened 2015-09-29
Add Eager reference to tools

zackbloom:


bep:
Thanks, but this looks like very general, and not Hugo specific. We haven't got enough room for everything remotely related.

PR #1467, opened 2015-10-01
Docs: Added RIDING BYTES to Showcase Page

ramonski:


spf13:
Merged as 7ecf2a55c1c9ddddbde35e0b2f8825d7634ef6d3. It would be great to have a "powered by Hugo" link in the footer and add a link to the source in the showcase.

PR #1476, opened 2015-10-06
Fix golint warnings for helpers/general.go

LK4D4:
It includes go-style of avoinding elses on after return and fix of
comment for NormalizeHugoFlags.

### DIFF
--- a/helpers/general.go
+++ b/helpers/general.go
@@ -430,7 +430,7 @@ func DoArithmetic(a, b interface{}, op rune) (interface{}, error) {
 	}
 }
 
-// NormalizeHugoFlagsFunc facilitates transitions of Hugo command-line flags,
+// NormalizeHugoFlags facilitates transitions of Hugo command-line flags,
 // e.g. --baseUrl to --baseURL, --uglyUrls to --uglyURLs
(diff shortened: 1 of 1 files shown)


bep:
In this case we're gonna keep the if/elses (golint is suggestions). Could you trim down this pull request to the error in the comment?

LK4D4:
@bep I wonder why you want to keep them?

bep:
In normal cases, I agree that multiple returns are better than nested if/elses. I have ran golint on Hugo several times in the past, and have skipped these same warnings before:

1) These nested, almost switch-like if-elses reads nicely as-is. It is a matter of taste.
2) Changing them just because golint says so is not a good reason enough. There is plenty of stuff to do in Hugo with more substance.
3) If the original author of this code had opened up this pull request, I would have merged it.

LK4D4:
@bep ok, your repo - your rules
even if I think that go-community style overpower any personal style. I think that it should be less place for personal styles.

bep:
Actually it's @spf13 s repo. I totally obey with the go vet rules. But I think you will find lots of golint warnings even in Go's stdlib.

bep:
Merged in 7b4e4769d070c3af97abaca9c1169f9e341be141

Thanks.

PR #1477, opened 2015-10-06
Use https for youtube shortcode

fmpwizard:
The Hugo site uses HTTPS but the video link was HTTP, which made chrome block the video


anthonyfok:
Hi @fmpwizard,

Thanks for your contribution.  Actually, the gohugo.io website is accessible at both http://gohugo.io/ and https://gohugo.io/, but indeed, as you have discovered, the embedded YouTube video fails to load when a user visit via HTTPS at https://gohugo.io/, and it seems YouTube favours SSL connection anyway, so yes, you patch looks good!

Merged as ddd8c77938e20ee4d43c98704197d9b63a903b32 (master) and e3d288982e52c24c2b49d90921d3baf0bf6dc360 (v0.14.docs)

fmpwizard:
Thanks!

PR #1479, opened 2015-10-08
V0.14.docs

thanhrossi:
Go lang for web ?


bep:
Not sure what the motivation with this is, but we are not merging in v014docs.

PR #1482, opened 2015-10-11
Remove '/' from .Hugo.Generator

luixxiul:
See: http://www.w3.org/TR/html5/syntax.html#start-tags

### DIFF
--- a/hugolib/hugo.go
+++ b/hugolib/hugo.go
@@ -26,6 +26,6 @@ func init() {
 		Version:    helpers.HugoVersion(),
 		CommitHash: CommitHash,
 		BuildDate:  BuildDate,
-		Generator:  template.HTML(fmt.Sprintf(`<meta name="generator" content="Hugo %s" />`, helpers.HugoVersion())),
+		Generator:  template.HTML(fmt.Sprintf(`<meta name="generator" content="Hugo %s">`, helpers.HugoVersion())),
 	}
 }


anthonyfok:
Hi @luixxiul,

According to http://www.w3.org/TR/html5/syntax.html#start-tags (emphasis mine):

> &#x36;. Then, if the element is one of the [void elements], or if the element is a [foreign element], ***then there may be a single "/" (U+002F) character***. This character has no effect on [void elements], but on [foreign elements] it marks the start tag as self-closing.

[void elements]: http://www.w3.org/TR/html5/syntax.html#void-elements
[foreign element]: http://www.w3.org/TR/html5/syntax.html#foreign-elements
[foreign elements]: http://www.w3.org/TR/html5/syntax.html#foreign-elements

So, while the "`/`" is unnecessary in `<meta name="generator" content="Hugo 0.14" />` for HTML’s "**HTML Syntax**", it is perfectly legal and allowed.

On the other hand, the "`/`" is mandatory for HTML5’s "**XHTML Syntax**", which I am sure some Hugo users use.  Furthermore, Hugo has the `blackfriday.HTML_USE_XHTML` flag enabled, so the void elements (or so-called "singleton tags" in Blackfriday source code) like `img`, `hr` and `br` are all closed with `/ >`, so it is not just `{{ .Hugo.Generator }}` that is using `/ >`.

That said, the HTML Syntax is more popular than the XHTML Syntax, so perhaps we should make it a user-configurable option, i.e. to enable XHTML-style close tag (`/ >`) or not (`>`).

bep:
>That said, the HTML Syntax is more popular than the XHTML Syntax, so perhaps we should make it a user-configurable option, i.e. to enable XHTML-style close tag (/ >) or not (>).

No. It is easy to work around this already if people really want an alternative.

luixxiul:
> No. It is easy to work around this already if people really want an alternative.

By postprocessing with tools like https://github.com/yoniholmes/grunt-text-replace ?

PR #1489, opened 2015-10-14
Create minimal default.md in archetype file fixes #315

### LINKED ISSUE #315: Cast error for `hugo new blabla/hello.md` with custom theme (due to empty archetype file)
Hello!

After creating a new theme and attempting to use Hugo's builder for content, I get the following error: `ERROR: 2014/06/01 Unable to Cast <nil> to map[string]interface{}`.

You can replicate the problem with Hugo v0.11 and the commands below.

    hugo new site test && cd test
    hugo new theme mytheme
    printf 'theme = "mytheme"\n' >> config.toml
    hugo new thispageworks.md
    hugo new this/fails.md

Am I doing anything wrong? I think this may be a bug.

lolindrath:
To fix the cast error, create a minimal default.md in the archetype
folder.

### DIFF
--- a/commands/new.go
+++ b/commands/new.go
@@ -163,7 +163,14 @@ func NewTheme(cmd *cobra.Command, args []string) {
 	touchFile(createpath, "layouts", "partials", "footer.html")
 
 	mkdir(createpath, "archetypes")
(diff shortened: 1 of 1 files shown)


spf13:
merged as fefd911

PR #1493, opened 2015-10-15
Fix #1406 by loading livereload.js from /

### LINKED ISSUE #1406: LiveReloadInject doesn't respect --appendPort=false
I'm running hugo on a remote server on port 3000, but nginx is proxying it as a virtual host. Live reload doesn't work because port 3000 is still included in the injected script tag.

Bug is here: https://github.com/spf13/hugo/blob/be45399cba0a47910db7a142e1574a9da1534713/transform/livereloadinject.go

Ideally this helper could be exported and used in the injector code:
https://github.com/spf13/hugo/blob/c00a1fcb7fc6a063725229c267a63b35dbf48d1a/commands/server.go#L156

kzys:
Instead of loading the file from http://localhost:port/, it can be
loaded from /.

### DIFF
--- a/transform/livereloadinject.go
+++ b/transform/livereloadinject.go
@@ -2,18 +2,13 @@ package transform
 
 import (
 	"bytes"
-	"github.com/spf13/viper"
 )
 
 func LiveReloadInject(ct contentTransformer) {
 	match := []byte("</body>")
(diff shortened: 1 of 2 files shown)


spf13:
merged as fdbfb4b

PR #1494, opened 2015-10-15
Disable LiveReload when Hugo is not running as a server

### LINKED ISSUE #1410: watch=true in config should not trigger livereload
I've built and deployed my site on a webserver. After that I got error message in console. It says that livereload.js is not available. 

image

Command ```hugo``` does not have option --disableLiveReload like ```hugo server```. Is there other way to exclude livereload code if it is not desirable at the moment?

kzys:
This change fixes #1410.


spf13:
Merged as 56417cd

PR #1498, opened 2015-10-16
The missing static directory shouldn't fail the build

### LINKED ISSUE #759: Unable to find Static Directory when copying assets from a theme
This is a little usability issue. Maybe it can be improved.

I removed some files from my `static` folder without thinking to leave a `.gitkeep` file. My [CI build](https://circleci.com/gh/go-fsnotify/fsnotify.org/6) logged this error message:

```
ERROR: 2015/01/01 Unable to find Static Directory: hyde in /home/ubuntu/fsnotify.org/static/
```

The result is it didn't copy over any of the assets from the hyde theme. In my case, this means those assets weren't deployed and missing from the server.

Two possibilities that could improve the situation:
* Just create the static directory if it's needed and missing.
* Make the ERROR into an actual error that does a log.Fatal / os.Exit(1) to fail the build.

kzys:
The lack of the directory would be worth to warn, since the directory is
created by "huge new site". But it doesn't have to fail the build.

This change fixes #759.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -328,12 +328,6 @@ func build(watches ...bool) {
 }
 
 func copyStatic() error {
-	staticDir := helpers.AbsPathify(viper.GetString("StaticDir")) + "/"
-	if _, err := os.Stat(staticDir); os.IsNotExist(err) {
-		jww.ERROR.Println("Unable to find Static Directory:", staticDir)
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


bep:
LGTM

spf13:
@kzys please just click on the CLA | not signed yet button to sign the CLA. I'm not sure why it's there twice, but you only need to do it once and it covers all your contributions.

bep:
Merged in 3a41254

PR #1501, opened 2015-10-17
Permalink on Node and Page should be of string

### LINKED ISSUE #1384: Permalink  on Node and Page should be of same type
Hello,

The `.Permalink` [variable](http://gohugo.io/templates/variables/) for pages and nodes returns the permanent link for the current webpage, but the `.Permalink` on pages is of type string while the same variable on nodes is of type template.HTML.

So when I have the following code in my default_head.html:

```
Page/Node permalink: {{ .Permalink }}
{{ if in .Permalink "categoryname" }}
    This is a match
{{ end }}
```

On my section page it prints:

```
Page/Node permalink: http://localhost:1313/categoryname/
```

On posts belonging to that section it prints:

```
Page/Node permalink: http://localhost:1313/categoryname/post-name/ This is a match
```

If it's not possible/feasible to change the behaviour of `.Permalink` on pages and nodes, perhaps a `.URL` variable can be added to pages? Then both a page and a node can have a string type variable (`.URL`) that returns the webpage's URLs (excluding the base URL).

(This issue is made after a discussion here: [Permalink node pages - if in issue](http://discuss.gohugo.io/t/permalink-node-pages-if-in-issue/1628).)

kzys:
This change fixes #1384.

### DIFF
--- a/hugolib/node.go
+++ b/hugolib/node.go
@@ -139,7 +139,7 @@ func (n *Node) RelRef(ref string) (string, error) {
 
 type URLPath struct {
 	URL       string
-	Permalink template.HTML
+	Permalink string
 	Slug      string
 	Section   string
 }
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -885,7 +885,7 @@ func (s *Site) RenderAliases() error {
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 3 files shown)


bep:
Merged in 99acbb2

PR #1505, opened 2015-10-19
Add internal Google Analytics template

### LINKED ISSUE #1424: Adding Google Analytics to the internal templates
Google Analytics is a widely used service in Hugo themes. This feature would be similar to the existing internal template for Disqus. Furthermore, theme creators just need to add a one-liner to their templates and they are ready to go.

```
{{ template "_internal/google-analytics.html" . }}
```

And the config file would allow the user to set their custom tracking code à la

```TOML
googleAnalytics = ""
```

digitalcraftsman:
This commits also adds the GoogleAnalytics
variable to insert the tracking code.

Fixes #1424.

### DIFF
--- a/docs/content/overview/introduction.md
+++ b/docs/content/overview/introduction.md
@@ -127,6 +127,7 @@ Hugo boasts the following features:
 ### Additional Features
 
   * Integrated [Disqus](https://disqus.com/) comment support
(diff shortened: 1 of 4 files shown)


spf13:
Merged as 9dbb6dc

Issue #1061, opened 2015-04-15
.Next vs .NextInSection goes in opposite directions


bep:
Which is confusing. 

I guess NextInSection does it right.

billychappell:
The reason they're different is due to .Next evaluating the same condition as .PrevInSection, while .NextInSection's matches .Prev in [hugolib/site.go](https://github.com/spf13/hugo/blob/master/hugolib/site.go); it looks like they were just mixed up.

Take a look:
(relevant functions can be found starting at line 382 and 839 in [hugolib/site.go](https://github.com/spf13/hugo/blob/master/hugolib/site.go))
```go
if i > 0 {
	wp.Page.NextInSection = s.Sections[k][i-1].Page
}
```
```go
if i > 0 {
	page.Prev = s.Pages[i-1]
}
```
and
```go
if i < len(s.Sections[k])-1 {
	wp.Page.PrevInSection = s.Sections[k][i+1].Page
}
```
```go
if i < len(s.Pages)-1 {
	page.Next = s.Pages[i+1]
}
```
I've created a fork and changed them to be consistent, then manually tested that they go in the same direction; you can compare it with the current release [here](https://github.com/spf13/hugo/compare/master...billychappell:master).

That's assuming .NextInSection/.PrevInSection were "correct," as bep stated. If you want, I can write some tests to ensure that we don't go through this again and that .Next/.NextInSection are linking to the last-written articles while .Prev/.PrevInSection go backwards in time before sending a pull request.

PS.  As a relatively inexperienced dev, this is my first time posting on a public Github repo; hopefully this helps! Let me know if I've done anything wrong or can do anything else to contribute.

bep:
Good analytics.

We should wait for @spf13 to get his saying before we do anything about this, but when the pages are sorted by data, it makes most sense that **next is newer**.

* But this is a breaking change in a very used feature
* But I guess we should do all the breaking changes when we can. If it's confusing for the current 2000 users it will be so for the future 20000.

billychappell:
@bep Thanks for the reply! I hadn't considered the level of impact it could have on users; I have one question while we wait for spf13:

With this type of breaking change, wouldn't it be better to add a new type, something like `LegacyMode  map[string]interface{}`, to the Site struct so that we can allow users to enable or disable deprecated features in versions immediately before and after such changes?

I'm imagining something like this:

```go
func (s *Site) setupPrevNext() {
	// New Site variables helps users transition
	// to breaking changes over multiple updates.
[...]

PR #1512, opened 2015-10-20
Added Testing for PrevInSection/NextInSection

billychappell:
See #1061


bep:
@billychappell I'm sorry that I have not followed up on these old pull requests. I appreciate the work, but I'm afraid that this code has drifted too much away from the current code base for us to use it. We would appreciate further work in this area, and I will promise faster response times.

PR #1520, opened 2015-10-24
Fix spell mistake in variables.md

arunk-s:


anthonyfok:
Good catch!  Thanks @arunk-s!

Merged as 277d0ae (master).

PR #1522, opened 2015-10-26
Set description in Page.Params map (Fix #1484)

### LINKED ISSUE #1484: Add description to Page.Params
Now it is set on Page.Description -- it would be useful if it was ALSO set in the Params map.

See #1462

arunk-s:
Set description in Page.Params map


bep:
Fixed in 9569d11

PR #1525, opened 2015-10-27
#1471 Added generated jekyll commands

### LINKED ISSUE #1471: Add doc for Jekyll import command
See https://github.com/spf13/hugo/pull/1469

/cc @anthonyfok

amasses:
Added generated docs for `hugo import jekyll` commands. Fixes #1471


bep:
Thanks, but that wasn't the documentation I was thinking about with that issue. When we do a new release, we will regenerate the complete commands-section.

PR #1530, opened 2015-10-28
Update link to S3 deploy instructions on Wercker

tdhopper:
Previous link goes to page with warning "This article is outdated"


anthonyfok:
Good idea, @tdhopper!  Thank you!

Merged as (08b3004) master and fc52361 (v0.14.docs).

PR #1545, opened 2015-11-03
Move blackfriday site-wide config loading to NewBlackFriday()

### LINKED ISSUE #1544: Markdownify gets context with empty config => ignores blackfriday variables
Meaning you can't disable fractions conversion etc...

mjanda:
Ok, I've tried to fix #1544

It's only partial solution - ```markdownify``` will still ignore page-level config.

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -24,7 +24,9 @@ import (
 	"unicode/utf8"
 
 	"github.com/miekg/mmark"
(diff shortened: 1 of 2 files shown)


bep:
>It's only partial solution - markdownify will still ignore page-level config.

It will always ignore page-level config.

bep:
Merged in 5838420

PR #1547, opened 2015-11-03
Doc fix: Replace SINGLE with PLURAL in taxonomy template file names

yvele:
Related to issue https://github.com/spf13/hugo/issues/1546


yvele:
This PR should be reverted (my mistake). See https://github.com/spf13/hugo/issues/1546#issuecomment-153753810

spf13:
WE can close this. We haven't merged it yet, so we don't need to revert it.

Issue #796, opened 2015-01-14
Add exec shortcode


bep:
Idea from this post:
http://discuss.gohugo.io/t/inline-ditaa-dot-or-plantuml-sources-in-hugo-posts/589/3

Similar to Highlight. but for any binary that writes to stdout.

Two arguments:
1. Binary
2. Args
- the inner content to stdin

I don't see any obvious issues why we could/should not add this? Security issues? It's up to the developer to use or abuse this?

bjornerik:
Come to think of it, it will give theme-authors the opening to do harmful stuff ... like "rm -r /".

But it's an interesting thought - if we could restrict it to a set of commands.

Just tested cat combined with highlight and some go files. Very useful.

halostatue:
I don’t think it gives theme authors anything—shortcodes aren’t rendered by templates, only content. Right?

anthonyfok:
Interesting, and very useful indeed!  :-)

Though, for security reasons, I think it should be a configurable option that is turned off by default.
It may also be helpful to print out the exec command for each website rebuild so that the end user is fully aware of what is going on.  And yes, restriction to a set of commands (and other needed sanitization) would be a good thing.

bjornerik:
@halostatue no, but shortcodes can be provided by themes. So a shortcode from a theme named `coolinnocentshortcode` is tempting to use.

But I think this is too useful turn down without some thinking.

bjornerik:
But it would be cool if we could make some of the obvious available as both template function and shortcode:

* cat
* tail
* curl
* ??

spf13:
Sorry I'm a bit late to this. When I saw this I immediately thought that it was a security issue as you pointed out in your follow up. 

I agree that the best approach would be to define a few commonly used commands.

cat & grep alone would be super powerful. Curl may be but could also pose a security risk. 

Another concern is that these won't work for Windows.

bjornerik:
Yea, thinking a little about this. I foresee a support nightmare. "I get 'broken pipe' with Photoshop".

I might maintain a private patch for myself, but for Hugo it's maybe better to find a Go implementation similar to cat etc.

My current use case is what Martin Fowler calls "[live code](http://martinfowler.com/articles/preparatory-refactoring-example.html)" in documentation - but much simpler than his setup, i. e. text extracts from source code on disc (I know about Gists).

rodlogic:
[...]

PR #1548, opened 2015-11-04
Added command exec shortcode. See #796

tubo028:
Currently what it can is receive 2 arguments (the command and options) and then insert stdout to the called point. Flowing inner text into stdin is not now.
Whitelist is configurable and only `cat` (for unix) and `type` (for windows) are available in default.

Now I am trying it on [my site](http://tubo28.me/). Just `cat` command reduced so many copy and paste on the source.
I am looking forwarding to being it official.


bep:
Can you fix the commit message to be in line with guidelines? (it's ok to amend your commit and do a push -f).

tubo028:
I done it, but my English ugly. Please point out them if unacceptable...

bep:
The English is just fine, and now the commit message looks good.

digitalcraftsman:
It would be great if you could add a few words about your new shortcodes in documentantions

moorereason:
I'm curious if we can write tests for this, as well.

tubo028:
You mean making a commit which adds description of the shortcode and some test codes?
I can write the documentation but I have not written any test codes, so I may be going to ask for help.

digitalcraftsman:
> You mean making a commit which adds description of the shortcode and some test codes?

That would be nice additions before we merging it.

We should at least add some test cases for the allowed default commands `cat` and `type`. We could test `cat` easily with a temporary file that is created during the test. The ouput of cat and the written content should be the same.

Does anyone has some good ideas for other test cases?

PR #1565, opened 2015-11-10
Fix GroupByParam to return original param string

### LINKED ISSUE #1564: GroupByParam makes the key lowercase
See 

http://discuss.gohugo.io/t/why-are-page-params-lowercase/2064/3

Not sure why.

@tatsushid ?

tatsushid:
Page.GroupByParam function internally uses Page.GetParam to get a
parameter value for a key of a page group but now Page.GetParam returns
a lowercase character string every time. It has no need to using
lowercase character string as a group key value and it confuse a
function user.

This fixes it to keep and return an original parameter string as a group
key value.

Fix #1564


bep:
Merged in 139be09

PR #1566, opened 2015-11-11
Add RSSLink to SiteInfo

moorereason:
It would be helpful to have be able to link to the main site RSS feed from any page/template.  This commit solves that problem.


bep:
Agree.

bep:
Merged in 3abb267

PR #1568, opened 2015-11-11
Adding article about moving to hugo

abhi1010:


anthonyfok:
Great article!  Thank you @abhi1010!

Committed as 8e68a64 (master) and f874b13 (v0.14.docs).

PR #1578, opened 2015-11-15
Return error if shortcode template.Tree is nil

### LINKED ISSUE #1575: Panic on invalid shortcode template
I'm running Hugo HEAD and Go 1.5.1.  Given the following shortcode `undef.html`:

```html+jinja
{{ $v = .Get 0 }}
```
And the following content:

```md
{{% undef test %}}
```

I get a panic:

```
sh$ hugo
ERROR: 2015/11/14 template: theme/shortcodes/crash1.html:1: undefined variable "$v"
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0x4bf7c3]

goroutine 54 [running]:
github.com/spf13/hugo/hugolib.isInnerShortcode(0xc82023c600, 0x0)
        /home/user/src/github.com/spf13/hugo/hugolib/shortcode.go:174 +0x133
github.com/spf13/hugo/hugolib.extractShortcode(0xc82004d8b0, 0xc82000d000, 0x7f2e85849800, 0xc820012a80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/user/src/github.com/spf13/hugo/hugolib/shortcode.go:346 +0xe8d
github.com/spf13/hugo/hugolib.extractShortcodes(0xc82000dc00, 0x3b6, 0xc82000d000, 0x7f2e85849800, 0xc820012a80, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/user/src/github.com/spf13/hugo/hugolib/shortcode.go:430 +0x84e
github.com/spf13/hugo/hugolib.extractAndRenderShortcodes(0xc82000dc00, 0x3b6, 0xc82000d000, 0x7f2e85849800, 0xc820012a80, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/user/src/github.com/spf13/hugo/hugolib/shortcode.go:253 +0x9d
github.com/spf13/hugo/hugolib.(*Page).ProcessShortcodes(0xc82000d000, 0x7f2e85849800, 0xc820012a80)
        /home/user/src/github.com/spf13/hugo/hugolib/page.go:884 +0x9c
github.com/spf13/hugo/hugolib.markdownHandler.PageConvert(0x0, 0x0, 0x0, 0xc82000d000, 0x7f4a53e64000, 0xc820012a00, 0x0, 0x0, 0x0, 0x0)
        /home/user/src/github.com/spf13/hugo/hugolib/handler_page.go:59 +0x7e
        ...
```

Upon looking into the code to try to find a fix, I considered adding the following at [shortcode.go(346)](https://github.com/spf13/hugo/blob/dcf1f96c2de5b3c0bcefd0741caf350318c4cbbd/hugolib/shortcode.go#L346):

```go
if tmpl.Tree == nil {
	return sc, fmt.Errorf("Template for shortcode '%s' failed to compile for page '%s'", sc.name, p.BaseFileName())
}
```

That fixes the panic, but the resulting page is render truncated at the point where the broken shortcode was referenced.  The same result can happen at line 343 if an undefined shortcode is referenced.  That's an acceptable result, I think.  However, the result is that the truncated `page.Content` is not rendered as Markdown and instead the raw Markdown syntax is output into the rendered HTML.  So, I stumbled onto another issue: that being the unfriendly way that the page renders if an erroneous shortcode reference is made.  I would rather see the broken shortcode reference result in no output and the remainder of the Content (or at least the truncated Content) be rendered as Markdown.

moorereason:
If a shortcode template fails to compile, the template will be non-nil, but
template.Tree will be nil which caused a panic.

Fixes #1575

### DIFF
--- a/hugolib/shortcode.go
+++ b/hugolib/shortcode.go
@@ -343,6 +343,9 @@ Loop:
 			if tmpl == nil {
 				return sc, fmt.Errorf("Unable to locate template for shortcode '%s' in page %s", sc.name, p.BaseFileName())
 			}
+			if tmpl.Tree == nil {
+				return sc, fmt.Errorf("Template for shortcode '%s' failed to compile for page '%s'", sc.name, p.BaseFileName())
+			}
 			isInner = isInnerShortcode(tmpl)
 
 		case tScParam:


bep:
Merged in 05c29ad

Issue #1576, opened 2015-11-15
Add built-in shortcodes for common usecases


digitalcraftsman:
Why shouldn't we add a number of shortcodes for the most popular usecases and services. New users could use them out of the box without the need of creating them manually.

Follow the [forum discussion](https://discuss.gohugo.io/t/shoud-we-add-internal-shortcodes-for-popular-services/2085) for more information.

Just a quick compilation of usecases / services (todo list):

- [x] Twitter
- [x] Speakerdeck presentations
- [x] YouTube videos
- [x] Vimeo videos
- [x] Github gists

derekperkins:
@digitalcraftsman - that's a good idea and not particularly difficult to implement. If you could figure out what the input options should be and sample output, that'd make it a piece of cake.

digitalcraftsman:
@derekperkins - below you find a list of the syntax that user would use to call the shortcodes. Do you think that something could be improved?

#### Speakerdeck
Required is the id of the slide. This id can be found on the right side of the presentation on Speakerdeck under 'Direct link'

    {{< speakerdeck "4e8126e72d853c0060001f97" >}}

#### YouTube
Just the id of the video is required:

    {{< vimeo "146022717" >}}

#### Vimeo
Just the id of the video is required:

    {{< vimeo "146022717" >}}

#### Gists
Gists would need the owner's username and its unique id:

    {{< gist "digitalcraftsman" "85f59771b5ae1e2091a8" >}}

bep:
A comment for the documentation: You only need quoted parameters if you have spaces or very special chars, so this is better:

```
{{< youtube w7Ft2ymGmfc >}}
```

digitalcraftsman:
Thanks for annotating this. I'll update the examples above.

digitalcraftsman:
All shortcodes are implemented (including Twitter) and the docs have been written.

derekperkins:
Awesome!

umtksa:
how about responsive vimeo and youtube
i have some custom shortcodes [here](https://github.com/umtksa/hugo-short-code)

PR #1594, opened 2015-11-20
Extend the built-in shortcodes

digitalcraftsman:
See #1576.

### DIFF
--- a/docs/content/extras/shortcodes.md
+++ b/docs/content/extras/shortcodes.md
@@ -171,9 +171,9 @@ You want to include a single tweet into your blog post? Everything you need is t
     
     https://twitter.com/spf13/status/666616452582129664
 
-Pass the tweet's author the id from the URL as parameters to the shortcode as shown below:
+Pass the tweet's id from the URL as parameter to the shortcode as shown below:
 
-    {{</* tweet spf13 666616452582129664 */>}}
+    {{</* tweet 666616452582129664 */>}}
 
 ***   
 
--- a/docs/content/extras/shortcodes.md
+++ b/docs/content/extras/shortcodes.md
@@ -9,7 +9,7 @@ next: /extras/pagination
 prev: /extras/permalinks
 title: Shortcodes
 weight: 80
-toc: true
+toc: false
 ---
 
 Hugo uses Markdown for its simple content format. However, there are a lot
@@ -63,7 +63,8 @@ The `<` character indicates that the shortcode's inner content doesn't need any 
 {{</* myshortcode */>}}<p>Hello <strong>World!</strong></p>{{</* /myshortcode */>}}
 ```
 
-## Hugo Shortcodes
+
+## Built-in Shortcodes
 
 Hugo ships with a set of predefined shortcodes.
 
@@ -101,6 +102,8 @@ closing shortcode.
       <span style="color: #f92672">&lt;/div&gt;</span>
     <span style="color: #f92672">&lt;/section&gt;</span>
 
+***
+
 ### figure
 `figure` is simply an extension of the image capabilities present with Markdown.
 `figure` provides the ability to add captions, CSS classes, alt text, links etc.
@@ -131,6 +134,8 @@ closing shortcode.
         </figcaption>
     </figure>
 
+***
+
 ### ref, relref
 
 These shortcodes will look up the pages by their relative path (e.g.,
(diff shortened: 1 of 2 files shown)


digitalcraftsman:
Should I squash everything into a single commit?

bep:
No. These are fine as separate commits.

digitalcraftsman:
I've merge them as part of Hugo v0.15 I could also update the release notes.

moorereason:
I've looked at the `youtube` shortcode, and I'm not a big fan of it.  `style` tags?  Shame.

I'm working on an enhancement.

moorereason:
I've submitted PR #1597 to add a shortcode `.ParamsType` function.  With that commit, I would propose updating these defaults to be more concise and flexible.  Here's how I would define the Youtube shortcode (line breaks for readability):

```django
{{ if eq .ParamsType "position" }}
<div class="{{ if len .Params | eq 2 }}{{ .Get 1 }}{{ else }}youtube-container{{ end }}">
  <iframe src="https://www.youtube.com/embed/{{ .Get 0 }}" allowfullscreen frameborder="0"></iframe>
 </div>
{{ else }}
<div class="{{ if .Get "class" }}{{ .Get "class" }}{{ else }}youtube-container{{ end }}">
  <iframe src="https://www.youtube.com/embed/{{ .Get "id" }}" allowfullscreen frameborder="0"></iframe>
 </div>
{{ end }}
```

Example references:
```
{{< youtube f8TkUMJtK5k >}}
{{< youtube f8TkUMJtK5k flex-video >}}
{{< youtube id="f8TkUMJtK5k" class="flex-video" }}
```

I like this approach because the user can control everything via SASS/CSS:

```sass
.flex-video {
  width: 560px;
  height: 315px;

  iframe {
    width: 100%;
    height: 100%;
  }
}
```

bep:
I had a look at this, too -- and having only inline style tags is not ideal. But it is the only way for a shortcode like this to be truly standalone.

But they should be fallback options, somehow.

digitalcraftsman:
@moorereason I know that inline style tags are an ugly way to add CSS code. But this is the only (as far as I know) for a shortcode to stay independent and to work out of the box. It was my intention that they take the full width of the layout and are responsive. With defaults, the user doesn't need to add any styling to achieve this.

I like your approach to give users more flexibility to style the elements of the shortcode.

moorereason:
I've updated PR #1597 to use a simple boolean, as @bep suggested.  Since it doesn't sound like I'm going to win the style argument, here's an updated example:

```django
{{ if .IsNamedParams }}
<div {{ if .Get "class" }}class="{{ .Get "class" }}"{{ else }}style="position: relative; padding-bottom: 56.25%; padding-top: 30px; height: 0; overflow: hidden;"{{ end }}>
  <iframe src="https://www.youtube.com/embed/{{ .Get "id" }}" {{ if not (.Get "class") }}style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;" {{ end }}allowfullscreen frameborder="0"></iframe>
 </div>
{{ else }}
<div {{ if len .Params | eq 2 }}class="{{ .Get 1 }}"{{ else }}style="position: relative; padding-bottom: 56.25%; padding-top: 30px; height: 0; overflow: hidden;"{{ end }}>
  <iframe src="https://www.youtube.com/embed/{{ .Get 0 }}" {{ if len .Params | eq 1 }}style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;" {{ end }}allowfullscreen frameborder="0"></iframe>
 </div>
{{ end }}
```
If you pass in a class name, we use that instead of the built-in style tags.  The logic on the `iframe` is that if you pass in a `class`, it's only for the out `div` container, and I assume that if you're using CSS classes, then you know how to select the `iframe` inside your container `div`.

bep:
That I like.

digitalcraftsman:
I've added the `IsNamedParams` property from @moorereason (see #1597) in the shortcodes.

bep:
Looks good to me. Any comments @moorereason ?

moorereason:
Does it matter we use `src="//hostname/..."` on some links and `src="https://hostname/..."` on others.  Why not remove the protocol from all links?  Isn't that the *proper* way to include remote resources?

On the twitter shortcode, I'd just send the tweet ID.  The username is unnecessary:

```
http://api.twitter.com/1/statuses/oembed.json?id={{ index .Params 0 }}
```

digitalcraftsman:
I've made every shortcode protocol independent by just leaving the `//`, except in the Twitter shortcode. This shortcode make use of `getJSON`. Without adding a specific protocol the function isn't able to request the data since an error message is returned: `unsupported protocol scheme ""`.

Now, the Twitter shortcode just requires the tweet's id.

moorereason:
LGTM :+1: 

PS - @digitalcraftsman, I think you need to sign the new CLA.  Hugo is now Apache 2.0 licensed!  Yea!

digitalcraftsman:
Done!

Nice to see that @spf13 was finally able to make transition.

bep:
As it is "years between Hugo releasees", we might take this discussion, too:

I notice you use the `GetJSON` to pull in tweets at build time. I guess this will load faster than using the embedded version?

```
<blockquote class="twitter-tweet" lang="en"><p lang="en" dir="ltr">.<a href="https://twitter.com/reneefrench">@reneefrench</a> found the original design by me and sketch by Renee for <a href="https://t.co/1PR6bX5cDd">https://t.co/1PR6bX5cDd</a> <a href="https://twitter.com/hashtag/golang?src=hash">#golang</a> <a href="https://t.co/qrVN7we8dP">pic.twitter.com/qrVN7we8dP</a></p>&mdash; Rob Pike (@rob_pike) <a href="https://twitter.com/rob_pike/status/667781817756356613">November 20, 2015</a></blockquote>
<script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>
```

bep:
Good work. Merged in 311307c

PR #1597, opened 2015-11-21
Add shortcode IsNamedParams property

moorereason:
It would be helpful to know whether a shortcode was called with positional or
named parameters.  This commit adds a `ParamsType` function that returns `name`
or `position`.

If this proposed change is acceptable, I can write tests for it.

### DIFF
--- a/hugolib/shortcode.go
+++ b/hugolib/shortcode.go
@@ -31,9 +31,10 @@ import (
 )
 
 type ShortcodeWithPage struct {
-	Params interface{}
-	Inner  template.HTML
-	Page   *Page
+	Params        interface{}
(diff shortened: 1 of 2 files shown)


bep:
I agree this is useful, but I would just make a `bool IsNamedParams` or `bool IsPosParams` -- I fail to see how we'll get a third option in there.

moorereason:
Reworked this PR to change to a boolean `IsNamedParams` property in `ShortcodeWithPage`.  Tests included.  Rebased commits to keep this clean.

bep:
Merged in fde47c5

Would be nice with 1 line in the documentation about this, see #1607

PR #1609, opened 2015-11-23
Add themesDir option to configuration

### LINKED ISSUE #1556: Support themesDir option
I would like to configure where my themes are located. Currently the "themes" dir name is hardcoded ([path.go#L213](https://github.com/spf13/hugo/blob/36adb5fb883bb7170d062334f8aaa64ed191be68/helpers/path.go#L213)) and not configurable. My proposal is to add a **themesDir** option (in addition to dataDir, staticDir, etc.):

``` yaml
themesDir: 'assets/docs/themes'
theme: 'myTheme'
```

In this case, hugo would use `asset/docs/themes/myTheme` as the theme folder.

digitalcraftsman:
themesDir can be used to change the default path of the themes folder.

Fixes #1556

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -147,6 +147,7 @@ func LoadDefaultSettings() {
 	viper.SetDefault("ArchetypeDir", "archetypes")
 	viper.SetDefault("PublishDir", "public")
 	viper.SetDefault("DataDir", "data")
+	viper.SetDefault("ThemesDir", "themes")
 	viper.SetDefault("DefaultLayout", "post")
 	viper.SetDefault("BuildDrafts", false)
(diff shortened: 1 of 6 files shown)


bep:
`viper.GetString("themesDir")` is repated a little too much, and should have its own method, I think.

digitalcraftsman:
Do you have an alternative name for [`GetThemeDir`](https://github.com/spf13/hugo/blob/f045d7a611b5c609d44e894b3c1165ce24b67776/helpers/path.go#L211)? This function would return an absolute path instead of a custom relative path.

spf13:
Merged as 4758732

PR #1620, opened 2015-11-25
Do not strip % characters out of URLs (Fixes #1292)

### LINKED ISSUE #1292: % characters are stripped out of URLs
The ```section``` variable passed into the pagination module is already URL encoded, 

https://github.com/spf13/hugo/blob/master/hugolib/pagination.go#L370

but the pagination URLs are encoded again using function ```URLizeAndPrep()``` in

https://github.com/spf13/hugo/blob/master/hugolib/pagination.go#L515

This yields incorrect URLs for pagination templates.

For example, a section name contains accent characters (añame) encoded as  ```a%C3%B1ame```, when processed by function ```URLizeAndPrep()```, will become ```ac3b1ame```, the ```%``` characters are stripped out.

And there may be other cases for non-English section names which are not handled correctly.

Removing the call to the ```URLizeAndPrep()``` function temporarily fix this issue for me, but the URL handling should be reviewed and should be consistent with URLs in Nodes and Pages.

tischda:
Do not strip % characters out of URLs (Fixes #1292)

### DIFF
--- a/helpers/path.go
+++ b/helpers/path.go
@@ -16,17 +16,18 @@ package helpers
 import (
 	"errors"
 	"fmt"
-	"github.com/spf13/afero"
-	jww "github.com/spf13/jwalterweatherman"
-	"github.com/spf13/viper"
-	"golang.org/x/text/transform"
-	"golang.org/x/text/unicode/norm"
 	"io"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 	"unicode"
+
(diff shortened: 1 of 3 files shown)


anthonyfok:
Hello @tischda, if you don't mind, could you please rebase your patch against the current `master` branch?  Many thanks!  :-)

spf13:
Merged as 40fccf2251a7187f002165653d42178e2f53a49b

PR #1626, opened 2015-11-26
Update press.md

ChristophFloat:
Include my Blogpost about using Hugo together with git to automatically generate the site when you push it into your Repo using a Git Hook (German).


anthonyfok:
Thanks @ChristophFloat!

Merged as 72a6697 (master) and 97c297a (v0.15.docs).

PR #1647, opened 2015-11-30
Added the command in the documentation.

horner:
It was confusing that the command was not in step 5, while it is in step 6.


anthonyfok:
Thank you for catching that, @horner!

Merged as  c83e10a (master) and 9677d3f (v0.15.docs).

PR #1650, opened 2015-11-30
Add Camunda blog and docs showcases

zeropaper:
We :heart: Hugo at Camunda.


anthonyfok:
Thank you for your contribution, @zeropaper!

But before I merge, may I ask you a favour?  Could you please recapture the thumbnail images, but this time, save them as PNG files instead?  I think these thumbnails would look much better as PNG files, and the file size would be about the same or even less because your website contains only a handful discreet colours, i.e. no gradients or photos.  :-)

Many thanks!

zeropaper:
Hi @anthonyfok
Sure!
The new pictures (and their respective paths in md) have been update.

Cheers and greetings from Berlin

anthonyfok:
Thanks for the quick reply!

I optimized the PNG files, squashed the two commits together,
and merged the result as d7ef272 (master) and 79ad055 (v0.15.docs).

 And your websites are now live on the Hugo [Site Showcase](http://gohugo.io/showcase/) page!

zeropaper:
Thanks a lot!

PR #1651, opened 2015-11-30
Use .Site.Pages in the documentation examples instead of .Site.Recent

gntech:
Replace the deprecated.Site.Recent variable with .Site.Pages in the documentation examples


anthonyfok:
@gntech, thank you for your careful proofreading!

Merged as b0b6a8c (master) and 7fbbe61 (v0.15.docs).

PR #1652, opened 2015-12-01
Change most global flags into local ones

### LINKED ISSUE #1624: CLI UX: Flags shouldn't be global
Command line flags should not be global. This is obvious when you look at for example the help for "gen man"; the options don't make sense with that command.

    $ hugo gen man --help
    This command automatically generates up-to-date man pages of Hugo's
    command-line interface.  By default, it creates the man page files
    in the "man" directory under the current directory.
    
    Usage:
      hugo gen man [flags]
    
    Global Flags:
      -b, --baseURL="": hostname (and path) to the root, e.g. http://spf13.com/
      -D, --buildDrafts[=false]: include content marked as draft
      -F, --buildFuture[=false]: include content with publishdate in the future
          --cacheDir="": filesystem path to cache directory. Defaults: $TMPDIR/hugo_cache/
          --canonifyURLs[=false]: if true, all relative URLs will be canonicalized using baseURL
          --config="": config file (default is path/config.yaml|json|toml)
      -d, --destination="": filesystem path to write files to
          --disableRSS[=false]: Do not build RSS files
          --disableSitemap[=false]: Do not build Sitemap file
          --editor="": edit new content with this editor, if provided
          --ignoreCache[=false]: Ignores the cache directory for reading but still writes to it
          --log[=false]: Enable Logging
          --logFile="": Log File path (if set, logging enabled automatically)
          --pluralizeListTitles[=true]: Pluralize titles in lists using inflect
          --preserveTaxonomyNames[=false]: Preserve taxonomy names as written ("Gérard Depardieu" vs "gerard-depardieu")
      -s, --source="": filesystem path to read files relative from
          --stepAnalysis[=false]: display memory and timing of different steps of the program
      -t, --theme="": theme to use (located in /themes/THEMENAME/)
          --uglyURLs[=false]: if true, use /filename.html instead of /filename/
      -v, --verbose[=false]: verbose output
          --verboseLog[=false]: verbose logging

anthonyfok:
This is to ensure that only the relevant command-line flags
for a certain Hugo subcommand is shown to the end user,
reducing clutter and improving user experience.

Fixes #1624 - CLI UX: Flags shouldn't be global

### DIFF
--- a/commands/benchmark.go
+++ b/commands/benchmark.go
@@ -28,23 +28,23 @@ var benchmarkCmd = &cobra.Command{
 	Short: "Benchmark hugo by building a site a number of times.",
 	Long: `Hugo can build a site many times over and analyze the running process
 creating a benchmark.`,
-	RunE: func(cmd *cobra.Command, args []string) error {
-		if err := InitializeConfig(); err != nil {
-			return err
-		}
-
-		return bench(cmd, args)
-	},
 }
 
 func init() {
+	initCoreCommonFlags(benchmarkCmd)
+
 	benchmarkCmd.Flags().StringVar(&cpuProfilefile, "cpuprofile", "", "path/filename for the CPU profile file")
 	benchmarkCmd.Flags().StringVar(&memProfilefile, "memprofile", "", "path/filename for the memory profile file")
 
 	benchmarkCmd.Flags().IntVarP(&benchmarkTimes, "count", "n", 13, "number of times to build the site")
+
+	benchmarkCmd.RunE = benchmark
 }
 
-func bench(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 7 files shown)


anthonyfok:
For some of the commands where the cmdName is passed along in `InitializeConfig(cmdName)`,
I had to move the RunE function definition out of the Command struct definition and set it in `init()` instead, like how it was done in server.go, to avoid an initialization loop during compile:

```
/home/foka/go/src/github.com/spf13/hugo/commands/list_config.go:58: initialization loop:
	/home/foka/go/src/github.com/spf13/hugo/commands/list_config.go:58 configCmd refers to
	/home/foka/go/src/github.com/spf13/hugo/commands/list_config.go:58 configCmd
```

PR #1656, opened 2015-12-01
Copy static dir files without theme's static dir

### LINKED ISSUE #1655: static folder content isn't copied over when used theme doesn't have one
I have created a new hugo site and I observed one behaviour with hugo v0.15.

1. I have a theme which doesn't have a static folder
2. The site uses that theme
3. Site has a static folder
4. `hugo` create public directory with the content except the content inside of "static" folder. It happens with `server` command too.

My current workaround is to create an empty "static" folder in the theme directory and then `hugo` copy the content inside of "static" into the "public" directory (or the specified output directory) as expected.

tatsushid:
Hugo command doesn't copy site's "static" directory files to "public"
directory if a specified theme doesn't have "static" directory because
theme's "static" directory checking always returns an error.

This fixes it by just showing a warning message about theme's "static"
directory absence and not skipping following processes.

Fix #1655

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -369,8 +369,7 @@ func copyStatic() error {
 
 	themeDir, err := helpers.GetThemeStaticDirPath()
 	if err != nil {
-		jww.ERROR.Println(err)
-		return nil
+		jww.WARN.Println(err)
 	}
 
 	// Copy the theme's static directory


anthonyfok:
Looks good to me!  Thank you @tatsushid!  Great work!

Merged as 6c81031 (master).

PR #1657, opened 2015-12-01
Make noTimes command flag work

### LINKED ISSUE #1583: noTimes flag is set but never used

tatsushid:
Hugo command provides "noTimes" flag but it doesn't affect synced files
timestamp regardless of whether the flag is specified or not.

This fixes it by adding a flag check and overwriting configuration if it
is specified.

Fix #1583


anthonyfok:
Thanks again @tatsushid for a great fix!

Merged as e989468 (master).

PR #1665, opened 2015-12-02
Remove transformation of menu URLs

gunnaraasen:
Fixes https://github.com/spf13/hugo/issues/1239. With this change, menu URLS will only be urlized and sanitized. It removes the URL prep step, which incorrectly transformed links in some cases. 

Would appreciate some feedback on `TestSectionPagesMenu` since I'm not quite sure if the changes preserve the intent of that test.

Also, I signed the CLA.

### DIFF
--- a/hugolib/menu_test.go
+++ b/hugolib/menu_test.go
@@ -47,7 +47,7 @@ const (
 	identifier="1"
 [[menu.tax]]
 	name = "Tax2"
-    url = "/two/key"
+    url = "/two/key/"
 	identifier="2"
 [[menu.tax]]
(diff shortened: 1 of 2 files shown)


gunnaraasen:
Not quite sure how this introduced 6 data races...

bep:
The data race (I think there is only one ...) was there before. I will have a look into that ... soon.

gunnaraasen:
Yeah, looks like the data races are in master too.

toddboom:
+1

bep:
Merged i d35d820

For future reference, notice the slightly modified commit message.

PR #1667, opened 2015-12-03
Render page prior to splitting on summary

OJ:
This is my first contribution to Hugo so I expect it to be strongly critiqued as I learn the ropes!

The issue that I found, which I briefly mentioned [here](https://discuss.gohugo.io/t/markdown-content-renders-as-regular-text-in-summary/1396/12), is that the summary generation doesn't work correctly in some cases when Markdown is used.

Hugo splits the page based on the summary marker, and then renders the header separately. If users make use of the references facility that comes with Markdown, then all the links in the reference list at the bottom of the page are no present when the markdown is rendered. Eg:
```
Long and fabulous [summary][] goes here!
.
.
<!--more-->
.
.
  [summary]: http://random.com/
```
With the links missing, the page ends up with Markdown creeping into the summary.

This PR contains an attempt at fixing this issue by rendering the page first. I don't think it's perfect, and so I would appreciate some commentary on what would be a better option. One thing that stands out for me is that the whitespace isn't trimmed due to blank lines being convered to `<p></p>`, but rather that cludge a fix together for that I thought I'd throw it open to commentary.

Thanks again for Hugo, I love it.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -196,8 +196,12 @@ func (p *Page) setSummary() {
 	// rendered and ready in p.contentShortcodes
 
 	if bytes.Contains(p.rawContent, helpers.SummaryDivider) {
-		sections := bytes.Split(p.rawContent, helpers.SummaryDivider)
-		header := sections[0]
+		// If markdown is in use, then we need to render the whole page
+		// prior to splitting because key bits of information (such as
+		// reference links) might be at the bottom of the file and will
+		// result in links not rendering correctly.
+		renderedPage := p.renderBytes(p.rawContent)
+		sections := bytes.Split(renderedPage, helpers.SummaryDivider)
 		p.Truncated = true
 		if len(sections[1]) < 20 {
 			// only whitespace?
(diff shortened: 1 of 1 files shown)


OJ:
I'm looking into fixing the build issues. Bear with me :)

OJ:
It looks like fixing this isn't going to be as simple as I thought. Forcing rendering before splitting wraps the `<!--more-->` tag in other markup in certain circumstances, and so splitting _after_ causes other kinds of issues. The few ideas I have come up with so far that work around this feel like kludges at best.

The cheap fix is to force the use of `[inline](links)`, but I am not a fan of those, and I don't really use them at all in my posts.

Keen to hear some ideas on how best to resolve this. I'll mull over some options as well. Thanks!

spf13:
One idea would be to replace the  with another string that may still be discoverable but would have different characteristics.

bling:
@OJ can you provide an example of where the more tag gets wrapped in other tags?  i merged your changes locally and it appears to work...but then again, i always have  split on its own line.

OJ:
@bling: sure mate. Basically anything that looks like this:

```
A paragraph goes like this.<!-- more -->Then another starts like this.
```
The unit tests for the project have such examples as well.

Cheers!

bling:
ah, i see what you mean.  i had my `<a>` tags styled with `display:block` so i didn't even notice (even with `<!--more-->` inline like what you mentioned).  it looks like [jekyll](https://github.com/jekyll/jekyll/blob/c63b51b6612a0bb867a69dff2aa38d66a08eb2a4/lib/jekyll/excerpt.rb#L123) does it by appending all the references to the summary after splitting.  seems like it would work here.

spf13:
I think we should follow the approach that @bling suggests here.

OJ:
+1.

I'll get onto this as soon as I've got the internet reconnected at home. I'll close this PR and open a new one. Thanks to all for the input!

PR #1679, opened 2015-12-06
Add Random function to template functions

EitZei:
Adds Shuffle function to shuffle sequence to random order.

The function can be used to generate random content to the sites. For example to display random quote or display list of random blog posts. 

It of course should be noted that in a context of STATIC site generator randomising is bit odd. However sites that are updated regularly or even automatically can achieve benefits from this. Also randomising works fine when used in a single-templates.

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -495,22 +495,12 @@ func After(index interface{}, seq interface{}) (interface{}, error) {
 	return seqv.Slice(indexv, seqv.Len()).Interface(), nil
 }
 
-// Random is exposed to templates, to iterate over N random items in a
-// rangeable list.
-func Random(count interface{}, seq interface{}) (interface{}, error) {
+// Shuffle is exposed to templates, to iterate over items in rangeable list in
+// a randomised order.
+func Shuffle(seq interface{}) (interface{}, error) {
 
-	if count == nil || seq == nil {
+	if seq == nil {
 		return nil, errors.New("both count and seq must be provided")
-	}
-
-	countv, err := cast.ToIntE(count)
-
-	if err != nil {
-		return nil, err
-	}
-
-	if countv < 1 {
-		return nil, errors.New("can't return negative/empty count of items from sequence")
 	}
 
 	seqv := reflect.ValueOf(seq)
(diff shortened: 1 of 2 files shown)


bep:
Looks good to me, and I agree about the usefulness.

But I would suggest to rename the function to `shuffle` -- which is what this does, isn't it?

(I will have a look into that failing test on Windows -- not related to this)

EitZei:
I agree that the ```shuffle``` would be more descriptive name. Maybe the count parameter should also be removed as the template library already contains functions for slicing sequences (```first```, ```after`` etc.). That way the function would better follow the single responsibility principle.

bep:
Yes, agree, this would be much cleaner without the count. The unix way.

tpl/template_funcs.go
```diff
@@ -1405,6 +1453,7 @@ func init() {
 		"first":        First,
 		"last":         Last,
 		"after":        After,
+		"random":		Random,
```
bep: run go fmt on the code.

tpl/template_funcs_test.go
```diff
@@ -328,6 +328,43 @@ func TestAfter(t *testing.T) {
...
+				t.Errorf("[%d] First didn't return an expected error", i)
+			}
+		} else {
+			resultsv := reflect.ValueOf(results)
+			if err != nil {
+				t.Errorf("[%d] failed: %s", i, err)
+				continue
+			}
+
+			if resultsv.Len() != this.expect {
```
bep: It would be nice to also test for the content of result. Maybe check that it contains all the items and maybe shuffle it a couple of times to see that the order changes.

spf13:
@EitZei any luck with the windows testing? I'd like to merge it if it's ready.

spf13:
the windows build issues were from somewhere else. I've merged this as 9008ac0. Thanks for your contribution @EitZei !

digitalcraftsman:
@EitZei, could open a new pull request and add a few words about `shuffle` in the documentation?

EitZei:
@digitalcraftsman Will do.

digitalcraftsman:
Thank you.

Issue #175, opened 2014-01-19
RSSLink values don't respect (non)uglyUrls


spf13:

glacjay:
When I set `uglyUrls` to `true`, `{{ .RSSlink }}` gives me `http://localhost:1313/.xml` instead of `http://localhost:1313/index.xml`. Is that caused by this bug?

glacjay:
@spf13 I think RSS links should not be uglified even when the user specify `uglyUrls = true`, is that correct?

spf13:
correct

--
Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

glacjay:
@spf13 So I think I encounter a bug, when I set `uglyUrls = true`, hugo gives me an incorrect RSS link `http://localhost:1313/.xml`.

Issue #1680, opened 2015-12-06
Fix test failure on Windows


bep:
```
_base_exists
--- PASS: TestDoNewSite_error_base_exists (0.00s)
=== RUN   TestDoNewSite_force_empty_dir
--- PASS: TestDoNewSite_force_empty_dir (0.00s)
=== RUN   TestDoNewSite_error_force_dir_inside_exists
--- FAIL: TestDoNewSite_error_force_dir_inside_exists (0.00s)
	assertions.go:203:                         	Error Trace:	new_test.go:68
			Error:		Expected value not to be nil.
		
=== RUN   TestDoNewSite_err
```

bep:
@spf13 I notice you have some non-Win-compatible commits in Hugo recently that is a suspect here.

anthonyfok:
See also https://github.com/spf13/hugo/pull/1669#issuecomment-161932753 and spf13/afero#26

anthonyfok:
Smart!  :+1:

spf13:
Resolved now

PR #1682, opened 2015-12-07
Fix /.xml RSSLink when uglyurls are enabled

jmcfarlane:
Prior to this commit the root url with uglyurls enabled is "/.xml".
This commit relates to #175.

### DIFF
--- a/helpers/url.go
+++ b/helpers/url.go
@@ -269,6 +269,10 @@ func Uglify(in string) string {
 		}
 		return in
 	}
+	// /.xml -> /index.xml
+	if name == "" {
+		return path.Dir(in) + "index" + ext
+	}
 	// /section/name.html -> /section/name.html
 	return path.Clean(in)
 }


jmcfarlane:
I'm not entirely sure what to think of the failing AppVeyor tests. Locally I ran `go test ./...` which passed. I can't imagine the diff here would result in the failures seen in the test results.

anthonyfok:
Don't worry, the AppVeyor test failure affects the current master too, and the problem is already known: see #1680 if you are interested.  :-)

anthonyfok:
Your patch looks good to me, but I will leave it to @ bep or @ spf13 for the code review as they are definitely more knowledgeable and qualified.  :-)

jmcfarlane:
Great, I know very little about the codebase so would benefit from any
feedback :)

spf13:
Merged as b1f2b43. Thanks for the fix @jmcfarlane !

Issue #1668, opened 2015-12-03
invalid Pygments option: lineanchors


wazoo:
When specifying the "lineanchors" option as specified here:

http://pygments.org/docs/formatters/

I get the following error message:

ERROR: 2015/12/03 invalid Pygments option: lineanchors

It looks like only some options are supported in the code for that module: https://github.com/spf13/hugo/blob/master/helpers/pygments.go#L135-L143

I can go ahead and add some more options since it looks like it just passes them through but I wanted to see if there were plans to already do that.

I would like to add the following opts:
- title
- prestyles
- anchorlinenos
- lineanchors
- linespans

robertbasic:
@wazoo is this now addressed? Can this issue be closed?

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

PR #1686, opened 2015-12-07
added more pygments opts

wazoo:
This is a pull request for  #1668, I ended up not using pygments (wanted to add a title and styling so I used shortcode) but I figured might as well share these changes anyway.  Also, I tried doing a rebase but it doesn't seem to have worked, not sure why but I figured since its one file its not that big of a deal.


spf13:
merged as 26d23f7. I'm not sure what you did, but I couldn't squash it either.

PR #1690, opened 2015-12-09
Remove Duplicate "Editor Plugins" Section

mholtrlc:
Editor Plugins was repeated a second time. with identical content


anthonyfok:
Thank you so much!
Merged as 1469ce0 (master) and a08ac13 (v0.15.docs).

PR #1693, opened 2015-12-09
0.15 Docs: remove references to dynamic CMS products

### LINKED ISSUE #1688: Incorrect information about ghost
'This means that, unlike systems like WordPress, Ghost and Drupal, which run on your web server expensively building a page every time a visitor requests one'

This section in the introduction of the docs is incorrect regarding ghost; all pages outside of the admin interface are static.

moorereason:
There's no need to call out other projects here, especially when we might
mischaracterize what they do.

Closes #1688


anthonyfok:
Merged as 4fef394 (v0.15.docs); see #1694.

PR #1694, opened 2015-12-09
Docs: remove references to dynamic CMS products

moorereason:
There's no need to call out other projects here, especially when we might
mischaracterize what they do.

See #1688


anthonyfok:
Agreed.  Better safe than sorry, and besides, I think Hugo has matured enough and gained enough reputation to stand on its own right.  :-)

anthonyfok:
Merged as 768370a (master) and 4fef394 (v0.15.docs).

PR #1699, opened 2015-12-10
Allow renaming of sitemap.xml

gcastel:
Hi, 

I developed this quick hack to enable the renaming of sitemap.xml.
(my use case : provide a main sitemap-index that imports hugo's sitemap)

I added the support of a "filename" property to the sitemap part of config.toml

Regards


spf13:
merged as 6cdb810. Thanks @gcastel

PR #1700, opened 2015-12-10
Added a missing closing tag

maxime-michel:


anthonyfok:
Good catch!  Thanks @maxime-michel!

Merged as 38c5db9 (master) and b49c3e1 (v0.15.docs).

PR #1704, opened 2015-12-12
Tools section: add Contentful converter and reference Jekyll import command

digitalcraftsman:


anthonyfok:
Thank you!

Merged as ebc187b and 7857601 (master), and as 52f5ade and 458499c (v0.15.docs).

PR #1712, opened 2015-12-17
Added async version of Google Analytics internal template

### LINKED ISSUE #1711: Hugo should include async version of Google Analytics
Relates to #1654 and #1505

Currently the GA template is added at the bottom of the body section instead of at the bottom of just before the `</head>` tag as [required by Google](https://support.google.com/analytics/answer/1008080)

This causes Google to be unable to verify ownership of the property in [Webmaster Central](https://www.google.com/webmasters/verification/home).

It should probably also use the **asynchronous** version from https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet

waddles:
fixes #1711 as per https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet

### DIFF
--- a/tpl/template_embedded.go
+++ b/tpl/template_embedded.go
@@ -233,4 +233,13 @@ ga('create', '{{ . }}', 'auto');
 ga('send', 'pageview');
 </script>
 {{ end }}`)
+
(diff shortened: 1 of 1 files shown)


spf13:
Merged as ff28120. Thanks!

PR #1713, opened 2015-12-17
Fix 3 data races in TestPageCache

moorereason:
This commit fixes 3 data races in TestPageCache:

- Pass "i" from the outer for-loop as a parameter to the goroutine
- Move l1.Lock to protect reading of o1
- Move l2.Lock to protect reading of o2

### DIFF
--- a/hugolib/pageCache_test.go
+++ b/hugolib/pageCache_test.go
@@ -15,10 +15,11 @@ package hugolib
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


anthonyfok:
Thank you Cameron!  It appears that Bjørn was working on the same issue and committed a similar fix in 0c01fef.  But thank you for your contributions as always!  :-)

bep:
Ah, sorry -- I didn't see this.

PR #1729, opened 2015-12-24
Improve Hugo examples

digitalcraftsman:


anthonyfok:
Hi @digitalcraftsman,

Is there any special reason why you change http://gohugo.io/overview/installing to https://gohugo.io/overview/installing?  I know that some people are pushing "HTTPS Everywhere", but I am not sure if we have reached that consensus here yet, especially when the Hugo documentation consistently uses http://gohugo.io/ (no HTTPS).

digitalcraftsman:
> I know that some people are pushing "HTTPS Everywhere", but I am not sure if we have reached that consensus here yet, especially when the Hugo documentation consistently uses http://gohugo.io/ (no HTTPS).

I've moved back to HTTP links. Since HTTPS support is given, I thought a switch would make no difference.

anthonyfok:
Thank you for changing your patch so promptly.

Personally, I have no problem one way or another, HTTP or HTTPS, but as far as I can tell, since @spf13 has consistently used http://gohugo.io/ in promoting Hugo, I infer that using plain HTTP for this address is Steve's preference, and I think we should respect that.

Furthermore, I do worry that HTTPS could be blocked in certain countries, so promoting the https://gohugo.io/ as the canonical URL may prevent some users from accessing our site.

That said, I have very glad that Steve did set up both HTTPS and HTTPS on the web server, so end users have the freedom to choose from the best of both worlds.  :-)

Of course, if, in the future, the Hugo community and especially Steve shall decide to start promoting https://gohugo.io/ as the canonical URL, I am perfectly fine with that, but when that happens, please change all instances in one go, i.e., in one single commit.  :-)

Merged as fc627d5 (master) and 344d33e (v0.15.docs).

(Oops, I just realized I did not need to push it to the v0.15.docs branch because it is not part of the docs.  Oh well, hehe.)

spf13:
HTTPS will use http2 now so we should use that when possible going forward.

PR #1734, opened 2015-12-29
Native Multilingual support in Hugo.

abourget:
The goal is to:
* have support to write content/post/whatever.en.md to /en/2015/12/22/whatever/index.html
* do the same with whatever.fr.md into /fr/2015/12/22
* enable or disable with the "Multilingual" switch in the config
* support having language switchers in templates, that know
  where the translated page is (with .Page.Translations)
  (when you're on /en/about/, there's a "Francais" link
   that points to /fr/a-propos)

Missing:
* example usage of the .Page.Translations feature.

* shortcode or function to do i18n of strings, without resorting to:
  {{ ( index $.Site.Data.translations $.Site.Params.locale ).topslogan }}

  * I'd like it to be based on data / JSON / YAML, but use a convention.

  * https://github.com/nicksnyder/go-i18n seems like a good candidate
    to do server-side translation if we want all the bells and whistles
    of a translation engine.. Has tooling to build translation workflows
    too.

* how to run it with two configs (one for each language), yet having
  something that writes the root "index.html" and redirect to something
  that either considers the users' browser, or has a fixed language
  redirection.

This is basic code to open up a discussion as to how best we could handle that.

I was thinking of exposing more `Translations` functionalities directly in the `.Site`, to crawl through one set of pages (instead of looping through all Pages, even though some are duplicated). Something like `BaseLanguagePages` ? and all those `Page` would have pointers to their translations.

We could run the generation once for each language, and each time generating a `/fr`, or `/en` subtree.. consistent in itself.. and knowing about "the other side" of translations (through .Page.Translations).  In that case, maybe `Site.Pages` could have only those of the current translation,
but `.Page.Translations` would know of the others.. that might simplify menu generation and all, as it
would never need to know about other translations, unless explicitly asked to.

So @spf13, what do you think of all that ?

### DIFF
--- a/docs/content/content/multilingual.md
+++ b/docs/content/content/multilingual.md
@@ -1,0 +1,155 @@
+---
+date: 2016-01-02T21:21:00Z
+menu:
+  main:
+    parent: content
+next: /content/example
+prev: /content/summaries
+title: Multilingual Mode
+weight: 68
+toc: true
+---
+
+Since version 0.16, Hugo supports a native Multilingual mode. You
+define the languages to render as such:
+
+```
+Multilingual:
+  en:
+    weight: 1
+    title: "My blog"
+    params:
+      linkedin: "english-link"
+  fr:
+    weight: 2
+
+    title: "Mon blog"
+    params:
+      linkedin: "lien-francais"
+    copyright: "Tout est miens"
+
+copyright: "Everything is mine"
+```
+
+Anything not defined in a `lang:` block will fall back to the global
+value for that key (like `copyright` in this example).
+
+With the config above, all content, sitemap, RSS feeds, paginations
+and taxonomy pages will be rendered under `/en` in English, and under
+`/fr` in French.
+
+
+### Translating your content
+
+Translated articles are picked up by the name of the content files.
+
+Example of translated articles:
+
+1. `/content/about.en.md`
+2. `/content/about.fr.md`
+
+You can also have:
+
+1. `/content/about.md`
+2. `/content/about.fr.md`
+
+in which case the config variable `DefaultContentLanguage` will be
+used to affect the default language `about.md`.  This way, you can
+slowly start to translate your current content without having to
+rename everything.
+
+If left unspecified, the value for `DefaultContentLanguage` defaults
+to `en`.
+
+By having the same _base file name_, the content pieces are linked
+together as translated pieces. Only the content pieces in the language
+defined by **.Site.CurrentLanguage** will be rendered in a run of
+`hugo`.  The translated content will be available in the
+`.Page.Translations` so you can create links to the corresponding
+translated pieces.
+
+
+### Language switching links
+
+A full example of the language-switching links would be:
+
+```
+        {{if .Site.Multilingual}}
+          {{if .IsPage}}
+            {{ range $txLang := .Site.Languages }}
+              {{if ne $lang $txLang}}
+                {{if isset $.Translations $txLang}}
+                  <a href="{{ (index $.Translations $txLang).Permalink }}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+                {{else}}
+                  <a href="/{{$txLang}}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+                {{end}}
+              {{end}}
+            {{end}}
+          {{end}}
+
+          {{if .IsNode}}
+            {{ range $txLang := .Site.Languages }}
+              {{if ne $lang $txLang}}
+                <a href="/{{$txLang}}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+              {{end}}
+            {{end}}
+          {{end}}
+        {{end}}
+```
+
+This makes use of the **.Site.Languages** variable to create links to
+the other available languages.  The order in which the languages are
+listed is defined by the `weight` attribute in each language under
+`Multilingual`.
+
+This will also require you to have some content in your `i18n/` files that would look like:
+
+```
+- id: language_switcher_en
+  translation: "English"
+- id: language_switcher_fr
+  translation: "Français"
+```
+
+and a copy of this in translations for each language.
+
+As you might notice, node pages link to the root of the other
+available translations (`/en`), as those pages do not necessarily have
+a translated counterpart.
+
+Taxonomies (tags, categories) are completely segregated between
+translations, will have their own tag clouds and list views.
+
+
+### Translations of strings
+
+Hugo uses [go-i18n](https://github.com/nicksnyder/go-i18n) to support
+string translations.  Follow the link to find tools to manage your
+translation workflows.
+
+Translations are collected from the `themes/[name]/i18n/` folder, in
+addition to the files present in `i18n/` at the root of your project.
+
+From within your templates, use the `i18n` function as such:
+
+```
+{{ i18n "home" }}
+```
+
+to use a definition like this one in `i18n/en-US.yaml`:
+
+```
+- id: home
+  translation: "Home"
+```
+
+
+### Multilingual Themes support
+
+To support Multilingual mode in your themes, you only need to make
+sure URLs defined manually (those not using `.Permalink` or `.URL`
+variables) in your templates are prefixed with `{{
+.Site.LanguagePrefix }}`. If `Multilingual` mode is enabled, the
+`LanguagePrefix` variable will be rendered as `/en` (based on
+`CurrentLanguage`). If not enabled, it will be an empty string, so it
+is harmless for non-multilingual sites.
--- a/docs/content/content/multilingual.md
+++ b/docs/content/content/multilingual.md
@@ -148,8 +148,12 @@ Hugo uses [go-i18n](https://github.com/nicksnyder/go-i18n) to support
 string translations.  Follow the link to find tools to manage your
 translation workflows.
 
-Translations are collected from the `themes/[name]/i18n/` folder, in
-addition to the files present in `i18n/` at the root of your project.
+Translations are collected from the `themes/[name]/i18n/` folder
+(built into the theme), as well as translations present in `i18n/` at
+the root of your project.  In the `i18n`, the translations will be
+merged and take precedence over what is in the theme folder.  Files in
+there follow RFC 5646 and should be named something like `en-US.yaml`,
+`fr.yaml`, etc..
 
 From within your templates, use the `i18n` function as such:
 
--- a/docs/content/content/multilingual.md
+++ b/docs/content/content/multilingual.md
@@ -10,8 +10,9 @@ weight: 68
 toc: true
 ---
 
-Since version 0.16, Hugo supports a native Multilingual mode. You
-define the languages to render as such:
+Since version 0.17, Hugo supports a native Multilingual mode. In your
+top-level `config.yaml` (or equivalent), you define the available
+languages in a `Multilingual` section such as:
 
 ```
 Multilingual:
@@ -32,7 +33,8 @@ copyright: "Everything is mine"
 ```
 
 Anything not defined in a `[lang]:` block will fall back to the global
-value for that key (like `copyright` in this example).
+value for that key (like `copyright` for the `en` lang in this
+example).
 
 With the config above, all content, sitemap, RSS feeds, paginations
 and taxonomy pages will be rendered under `/en` in English, and under
--- a/docs/content/content/multilingual.md
+++ b/docs/content/content/multilingual.md
@@ -71,30 +71,44 @@ translated pieces.
 
 ### Language switching links
 
-A full example of the language-switching links would be:
+Here is a simple example if all your pages are translated:
 
 ```
-        {{if .Site.Multilingual}}
-          {{if .IsPage}}
-            {{ range $txLang := .Site.Languages }}
-              {{if ne $lang $txLang}}
-                {{if isset $.Translations $txLang}}
-                  <a href="{{ (index $.Translations $txLang).Permalink }}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
-                {{else}}
-                  <a href="/{{$txLang}}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
-                {{end}}
-              {{end}}
-            {{end}}
-          {{end}}
+{{if .IsPage}}
+  {{ range $txLang := .Site.Languages }}
+    {{if isset $.Translations $txLang}}
+      <a href="{{ (index $.Translations $txLang).Permalink }}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+    {{end}}
+  {{end}}
+{{end}}
 
-          {{if .IsNode}}
-            {{ range $txLang := .Site.Languages }}
-              {{if ne $lang $txLang}}
-                <a href="/{{$txLang}}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
-              {{end}}
-            {{end}}
-          {{end}}
-        {{end}}
+{{if .IsNode}}
+  {{ range $txLang := .Site.Languages }}
+    <a href="/{{$txLang}}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+  {{end}}
+{{end}}
+```
+
+This is a more complete example. It handles missing translations and will support non-multilingual sites. Better for theme authors:
+
+```
+{{if .Site.Multilingual}}
+  {{if .IsPage}}
+    {{ range $txLang := .Site.Languages }}
+      {{if isset $.Translations $txLang}}
+        <a href="{{ (index $.Translations $txLang).Permalink }}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+      {{else}}
+        <a href="/{{$txLang}}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+      {{end}}
+    {{end}}
+  {{end}}
+
+  {{if .IsNode}}
+    {{ range $txLang := .Site.Languages }}
+      <a href="/{{$txLang}}">{{ i18n ( printf "language_switcher_%s" $txLang ) }}</a>
+    {{end}}
+  {{end}}
+{{end}}
 ```
 
 This makes use of the **.Site.Languages** variable to create links to
@@ -102,7 +116,8 @@ the other available languages.  The order in which the languages are
 listed is defined by the `weight` attribute in each language under
 `Multilingual`.
 
-This will also require you to have some content in your `i18n/` files that would look like:
+This will also require you to have some content in your `i18n/` files
+(see below) that would look like:
 
 ```
 - id: language_switcher_en
@@ -118,10 +133,10 @@ available translations (`/en`), as those pages do not necessarily have
 a translated counterpart.
 
 Taxonomies (tags, categories) are completely segregated between
-translations, will have their own tag clouds and list views.
+translations and will have their own tag clouds and list views.
 
 
-### Translations of strings
+### Translation of strings
 
 Hugo uses [go-i18n](https://github.com/nicksnyder/go-i18n) to support
 string translations.  Follow the link to find tools to manage your
@@ -150,6 +165,6 @@ To support Multilingual mode in your themes, you only need to make
 sure URLs defined manually (those not using `.Permalink` or `.URL`
 variables) in your templates are prefixed with `{{
 .Site.LanguagePrefix }}`. If `Multilingual` mode is enabled, the
-`LanguagePrefix` variable will be rendered as `/en` (based on
-`CurrentLanguage`). If not enabled, it will be an empty string, so it
-is harmless for non-multilingual sites.
+`LanguagePrefix` variable will equal `"/en"` (or whatever your
+`CurrentLanguage` is). If not enabled, it will be an empty string, so
+it is harmless for non-multilingual sites.
--- a/docs/content/content/multilingual.md
+++ b/docs/content/content/multilingual.md
@@ -31,12 +31,16 @@ Multilingual:
 copyright: "Everything is mine"
 ```
 
-Anything not defined in a `lang:` block will fall back to the global
+Anything not defined in a `[lang]:` block will fall back to the global
 value for that key (like `copyright` in this example).
 
 With the config above, all content, sitemap, RSS feeds, paginations
 and taxonomy pages will be rendered under `/en` in English, and under
 `/fr` in French.
+
+Only those keys are read under `Multilingual`: `weight`, `title`,
+`author`, `social`, `languageCode`, `copyright`, `disqusShortname`,
+`params` (which can contain a map of several other keys).
 
 
 ### Translating your content
@@ -168,3 +172,61 @@ variables) in your templates are prefixed with `{{
 `LanguagePrefix` variable will equal `"/en"` (or whatever your
 `CurrentLanguage` is). If not enabled, it will be an empty string, so
 it is harmless for non-multilingual sites.
+
+
+### Multilingual index.html and 404.html
+
+To redirect your users to their closest language, drop an `index.html`
+in `/static` of your site, with the following content (tailored to
+your needs) to redirect based on their browser's language:
+
+```
+<html><head>
+<meta http-equiv="refresh" content="1;url=/en" /><!-- just in case JS doesn't work -->
+<script>
+lang = window.navigator.language.substr(0, 2);
+if (lang == "fr") {
+    window.location = "/fr";
+} else {
+    window.location = "/en";
+}
+
+/* or simply:
+window.location = "/en";
+*/
+</script></head><body></body></html>
+```
+
+An even simpler version will always redirect your users to a given language:
+
+```
+<html><head>
+<meta http-equiv="refresh" content="0;url=/en" />
+</head><body></body></html>
+```
+
+You can do something similar with your `404.html` page, as you don't
+know the language of someone arriving at a non-existing page.  You
+could inspect the prefix of the navigator path in Javascript or use
+the browser's language detection like above.
+
+
+### Sitemaps
+
+As sitemaps are generated once per language and live in
+`[lang]/sitemap.xml`. Write this content in `static/sitemap.xml` to
+link all your sitemaps together:
+
+```
+<?xml version="1.0" encoding="UTF-8"?>
+<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+   <sitemap>
+      <loc>https://example.com/en/sitemap.xml</loc>
+   </sitemap>
+   <sitemap>
+      <loc>https://example.com/fr/sitemap.xml</loc>
+   </sitemap>
+</sitemapindex>
+```
+
+and explicitly list all the languages you want referenced.
--- a/hugolib/multilingual.go
+++ b/hugolib/multilingual.go
@@ -1,0 +1,48 @@
+package hugolib
+
+import (
+	"github.com/spf13/cast"
+	"github.com/spf13/viper"
+)
+
+type Multilingual struct {
+	enabled bool
+	config  *viper.Viper
+
+	Languages []string
+}
+
+func (ml *Multilingual) GetString(key string) string { return cast.ToString(ml.Get(key)) }
+func (ml *Multilingual) GetStringMap(key string) map[string]interface{} {
+	return cast.ToStringMap(ml.Get(key))
+}
+
+func (ml *Multilingual) GetStringMapString(key string) map[string]string {
+	return cast.ToStringMapString(ml.Get(key))
+}
+
+func (ml *Multilingual) Get(key string) interface{} {
+	if ml != nil && ml.config != nil && ml.config.IsSet(key) {
+		return ml.config.Get(key)
+	}
+	return viper.Get(key)
+}
+
+func (s *Site) SetMultilingualConfig(currentLang string, orderedLanguages []string, langConfigs map[string]interface{}) {
+	conf := viper.New()
+	for k, val := range cast.ToStringMap(langConfigs[currentLang]) {
+		conf.Set(k, val)
+	}
+	conf.Set("CurrentLanguage", currentLang)
+	ml := &Multilingual{
+		enabled:   len(langConfigs) > 0,
+		config:    conf,
+		Languages: orderedLanguages,
+	}
+	viper.Set("Multilingual", ml.enabled)
+	s.Multilingual = ml
+}
+
+func (s *Site) multilingualEnabled() bool {
+	return s.Multilingual != nil && s.Multilingual.enabled
+}
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -61,8 +61,10 @@ type Page struct {
 	PublishDate         time.Time
 	ExpiryDate          time.Time
 	Markup              string
+	Translations        Translations
 	extension           string
 	contentType         string
+	lang                string
 	renderable          bool
 	Layout              string
 	layoutsCalculated   []string
@@ -300,9 +302,11 @@ func (p *Page) getRenderingConfig() *helpers.Blackfriday {
 
 func newPage(filename string) *Page {
 	page := Page{contentType: "",
-		Source: Source{File: *source.NewFile(filename)},
-		Node:   Node{Keywords: []string{}, Sitemap: Sitemap{Priority: -1}},
-		Params: make(map[string]interface{})}
+		Source:       Source{File: *source.NewFile(filename)},
+		Node:         Node{Keywords: []string{}, Sitemap: Sitemap{Priority: -1}},
+		Params:       make(map[string]interface{}),
+		Translations: make(Translations),
+	}
 
 	jww.DEBUG.Println("Reading from", page.File.Path())
 	return &page
@@ -445,10 +449,12 @@ func (p *Page) permalink() (*url.URL, error) {
 		if len(pSlug) > 0 {
 			permalink = helpers.URLPrep(viper.GetBool("UglyURLs"), path.Join(dir, p.Slug+"."+p.Extension()))
 		} else {
-			_, t := filepath.Split(p.Source.LogicalName())
+			t := p.Source.TranslationBaseName()
 			permalink = helpers.URLPrep(viper.GetBool("UglyURLs"), path.Join(dir, helpers.ReplaceExtension(strings.TrimSpace(t), p.Extension())))
 		}
 	}
+
+	permalink = p.addMultilingualWebPrefix(permalink)
 
 	return helpers.MakePermalink(baseURL, permalink), nil
 }
@@ -458,6 +464,10 @@ func (p *Page) Extension() string {
 		return p.extension
 	}
 	return viper.GetString("DefaultExtension")
+}
+
+func (p *Page) Lang() string {
+	return p.lang
 }
 
 func (p *Page) LinkTitle() string {
@@ -699,28 +709,28 @@ func (p *Page) getParam(key string, stringToLower bool) interface{} {
 		return nil
 	}
 
-	switch v.(type) {
+	switch val := v.(type) {
 	case bool:
-		return v
-	case time.Time:
-		return v
+		return val
+	case string:
+		if stringToLower {
+			return strings.ToLower(val)
+		}
+		return val
 	case int64, int32, int16, int8, int:
 		return cast.ToInt(v)
 	case float64, float32:
 		return cast.ToFloat64(v)
+	case time.Time:
+		return val
+	case []string:
+		if stringToLower {
+			return helpers.SliceToLower(val)
+		}
+		return v
 	case map[string]interface{}: // JSON and TOML
 		return v
 	case map[interface{}]interface{}: // YAML
-		return v
-	case string:
-		if stringToLower {
-			return strings.ToLower(v.(string))
-		}
-		return v
-	case []string:
-		if stringToLower {
-			return helpers.SliceToLower(v.([]string))
-		}
 		return v
 	}
 
@@ -851,6 +861,7 @@ func (p *Page) parse(reader io.Reader) error {
 	p.renderable = psr.IsRenderable()
 	p.frontmatter = psr.FrontMatter()
 	p.rawContent = psr.Content()
+	p.lang = p.Source.File.Lang()
 
 	meta, err := psr.Metadata()
 	if meta != nil {
@@ -975,7 +986,6 @@ func (p *Page) FullFilePath() string {
 }
 
 func (p *Page) TargetPath() (outfile string) {
-
 	// Always use URL if it's specified
 	if len(strings.TrimSpace(p.URL)) > 2 {
 		outfile = strings.TrimSpace(p.URL)
@@ -997,6 +1007,7 @@ func (p *Page) TargetPath() (outfile string) {
 				outfile += "index.html"
 			}
 			outfile = filepath.FromSlash(outfile)
+			outfile = p.addMultilingualFilesystemPrefix(outfile)
 			return
 		}
 	}
@@ -1005,8 +1016,22 @@ func (p *Page) TargetPath() (outfile string) {
 		outfile = strings.TrimSpace(p.Slug) + "." + p.Extension()
 	} else {
 		// Fall back to filename
-		outfile = helpers.ReplaceExtension(p.Source.LogicalName(), p.Extension())
+		outfile = helpers.ReplaceExtension(p.Source.TranslationBaseName(), p.Extension())
 	}
 
-	return filepath.Join(strings.ToLower(helpers.MakePath(p.Source.Dir())), strings.TrimSpace(outfile))
+	return p.addMultilingualFilesystemPrefix(filepath.Join(strings.ToLower(helpers.MakePath(p.Source.Dir())), strings.TrimSpace(outfile)))
+}
+
+func (p *Page) addMultilingualWebPrefix(outfile string) string {
+	if p.Lang() == "" {
+		return outfile
+	}
+	return "/" + path.Join(p.Lang(), outfile)
+}
+
+func (p *Page) addMultilingualFilesystemPrefix(outfile string) string {
+	if p.Lang() == "" {
+		return outfile
+	}
+	return string(filepath.Separator) + filepath.Join(p.Lang(), outfile)
 }
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
@@ -28,8 +29,6 @@ import (
 	"time"
 
 	"sync/atomic"
-
-	"path"
 
 	"github.com/bep/inflect"
 	"github.com/fsnotify/fsnotify"
@@ -75,22 +74,24 @@ var distinctErrorLogger = helpers.NewDistinctErrorLogger()
 //
 // 5. The entire collection of files is written to disk.
 type Site struct {
-	Pages          Pages
-	Files          []*source.File
-	Tmpl           tpl.Template
-	Taxonomies     TaxonomyList
-	Source         source.Input
-	Sections       Taxonomy
-	Info           SiteInfo
-	Menus          Menus
-	timer          *nitro.B
-	targets        targetList
-	targetListInit sync.Once
-	RunMode        runmode
-	draftCount     int
-	futureCount    int
-	expiredCount   int
-	Data           map[string]interface{}
+	Pages           Pages
+	TranslatedPages Pages
+	Files           []*source.File
+	Tmpl            tpl.Template
+	Taxonomies      TaxonomyList
+	Source          source.Input
+	Sections        Taxonomy
+	Info            SiteInfo
+	Menus           Menus
+	timer           *nitro.B
+	targets         targetList
+	targetListInit  sync.Once
+	RunMode         runmode
+	Multilingual    *Multilingual
+	draftCount      int
+	futureCount     int
+	expiredCount    int
+	Data            map[string]interface{}
 }
 
 type targetList struct {
@@ -106,7 +107,8 @@ type SiteInfo struct {
 	Authors               AuthorList
 	Social                SiteSocial
 	Sections              Taxonomy
-	Pages                 *Pages
+	Pages                 *Pages // Includes only pages in this language
+	TranslatedPages       *Pages // Includes other translated pages, excluding those in this language.
 	Files                 *[]*source.File
 	Menus                 *Menus
 	Hugo                  *HugoInfo
@@ -125,6 +127,11 @@ type SiteInfo struct {
 	preserveTaxonomyNames bool
 	paginationPageCount   uint64
 	Data                  *map[string]interface{}
+
+	Multilingual    bool
+	CurrentLanguage string
+	LanguagePrefix  string
+	Languages       []string
 }
 
 // SiteSocial is a place to put social details on a site level. These are the
@@ -150,17 +157,17 @@ func (s *SiteInfo) GetParam(key string) interface{} {
 		return nil
 	}
 
-	switch v.(type) {
+	switch val := v.(type) {
 	case bool:
-		return cast.ToBool(v)
+		return val
 	case string:
-		return cast.ToString(v)
+		return val
 	case int64, int32, int16, int8, int:
 		return cast.ToInt(v)
 	case float64, float32:
 		return cast.ToFloat64(v)
 	case time.Time:
-		return cast.ToTime(v)
+		return val
 	case []string:
 		return v
 	}
@@ -538,6 +545,9 @@ func (s *Site) ReBuild(events []fsnotify.Event) error {
 
 	s.timerStep("read & convert pages from source")
 
+	// FIXME: does this go inside the next `if` statement ?
+	s.setupTranslations()
+
 	if len(sourceChanged) > 0 {
 		s.setupPrevNext()
 		if err = s.buildSiteMeta(); err != nil {
@@ -665,9 +675,9 @@ func (s *Site) readDataFromSourceFS() error {
 	dataSources = append(dataSources, &source.Filesystem{Base: s.absDataDir()})
 
 	// have to be last - duplicate keys in earlier entries will win
-	themeStaticDir, err := helpers.GetThemeDataDirPath()
+	themeDataDir, err := helpers.GetThemeDataDirPath()
 	if err == nil {
-		dataSources = append(dataSources, &source.Filesystem{Base: themeStaticDir})
+		dataSources = append(dataSources, &source.Filesystem{Base: themeDataDir})
 	}
 
 	err = s.loadData(dataSources)
@@ -688,9 +698,22 @@ func (s *Site) Process() (err error) {
 		return
 	}
 
+	i18nSources := []source.Input{&source.Filesystem{Base: s.absI18nDir()}}
+
+	themeI18nDir, err := helpers.GetThemeI18nDirPath()
+	if err == nil {
+		i18nSources = []source.Input{&source.Filesystem{Base: themeI18nDir}, i18nSources[0]}
+	}
+
+	if err = loadI18n(i18nSources, s.Multilingual.GetString("CurrentLanguage")); err != nil {
+		return
+	}
+	s.timerStep("load i18n")
+
 	if err = s.createPages(); err != nil {
 		return
 	}
+	s.setupTranslations()
 	s.setupPrevNext()
 	if err = s.buildSiteMeta(); err != nil {
 		return
@@ -709,6 +732,30 @@ func (s *Site) setupPrevNext() {
 			page.Prev = s.Pages[i-1]
 		}
 	}
+}
+
+func (s *Site) setupTranslations() {
+	if !s.multilingualEnabled() {
+		return
+	}
+
+	currentLang := s.Multilingual.GetString("CurrentLanguage")
+
+	allTranslations := pagesToTranslationsMap(s.Pages)
+	assignTranslationsToPages(allTranslations, s.Pages)
+
+	var currentLangPages []*Page
+	var otherTranslationsPages []*Page
+	for _, p := range s.Pages {
+		if p.Lang() == "" || strings.HasPrefix(currentLang, p.lang) {
+			currentLangPages = append(currentLangPages, p)
+		} else {
+			otherTranslationsPages = append(otherTranslationsPages, p)
+		}
+	}
+
+	s.TranslatedPages = otherTranslationsPages
+	s.Pages = currentLangPages
 }
 
 func (s *Site) Render() (err error) {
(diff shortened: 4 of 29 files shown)


hugolib/page.go
```diff
@@ -937,5 +953,12 @@ func (p *Page) TargetPath() (outfile string) {
...
 
-	return filepath.Join(strings.ToLower(helpers.MakePath(p.Source.Dir())), strings.TrimSpace(outfile))
+	return p.addLangPrefix(filepath.Join(strings.ToLower(helpers.MakePath(p.Source.Dir())), strings.TrimSpace(outfile)))
+}
+
+func (p *Page) addLangPrefix(outfile string) string {
+	if p.lang == "" {
+		return outfile
+	}
+	return p.lang + "/" + outfile
```
bep: I would use `filepath.Join` here -- to get the filepath separator correct

abourget: filepath uses an OS dependent path.. is that what we want here ? thought it was destined to build URLs.. I'll have to look into more details how URLs are generated.

What do you think of the general idea though ?

moorereason: On line 940, we take a url-style path and converts it to an OS dependent path with `filepath.FromSlash`.  `filepath.FromSlash` will convert `posts/index.html` to `posts\index.html` on Windows.

@bep is correct.  In this context, you want to use `filepath.Join`.

spf13: I like the general approach.

abourget:
@spf13 I was also looking at https://godoc.org/github.com/nicksnyder/go-i18n/i18n ...wouldn't it be awesome if, in addition to good support for i18n in terms of content, we had great translation capabilities?

We could load `i18n/{{$.Site.RenderLanguage}}.yaml` as an `go-i18n` translation on start, just like we load `.Site.Data`.. This would require `Multilingual` to be enabled.

I can see quite a nice mapping from the examples there to something Hugo-ish:
```
T("Hello {{.Person}}", map[string]interface{}{
    "Person": "Bob",
})
```
to something like `{{ i18n "Hello {{.Person}}" (dict "Person" "Bob") }}` or something like that.

Or `T("You have {{.Count}} unread emails.", 2)` -> `{{ i18n "You.. {{.Count}} unread mails" .Site.MailCountWhatever }}`

I'd add a native `i18n` function, or `T` (or both, one aliasing the other ?) to handle that.

What do you think ? What's your take on vendoring other libraries ?

spf13:
I like this idea a lot. I think it should be an independent commit/branch
though. Same goals, different functionality.

bep:
Also see https://github.com/golang/go/issues/12750

abourget:
@spf13  this PR is becoming more robust and more solidly tested.

It'd be great to merge in at some point.. can anyone review ? @bep or someone?

I've taken out the fixups and unrelated doc changes ... created other PRs for that.

abourget:
Oh I thought appveyor was broken for other reasons, but Multilingual mode breaks under Windows.. will fix and push again.

abourget:
@spf13 do you agree with the general approach ? Think it can be merged eventually ?

spf13:
Totally. Well done so far.

abourget:
ok, this should be ready. works on Windows / Linux.. anything missing that you see ?

spf13:
This is very VERY good. 

In reviewing I had a few ideas in ways we can further refine this so that it's a bit easier to use.

1. It would be nice to be able to migrate from a single language to a multilanguage without needing to go in and modify the filenames of content files. Would it make sense that the defaultLanguage is the file without a language extension and we tell it which one that is. 

2. Another idea is the concept of a base for the content. I think it would make a lot of sense to have translations still share a lot of the metadata with the original piece of content. 

3. I think it would make a lot of sense for taxonomies (and possibly categories) to have translations for them which would allow content to easily keep in sync.

I would think perhaps the best accomplish to do this would be the following approach.

Each piece of content has a base file. This is the one without a language extension. It would specify the metadata and content that each enabled language would use when rendering if the given metadata or content wasn't present for that specific language. For things like categories and taxonomy it would mean that you would only need to specify them in the base and apply the translation to the terms. This would keep the content nicely in sync. 

I believe that with translations that usually the case that a partially translated site is better than a partial site. By taking this approach any content that isn't yet translated will still be available, but in the base language.

While in my personal sites I have little experience with this and in particular the subtleties of the actual translation, I was responsible for the documentation of both MongoDB and Docker and we dealt with this a lot there. This follows more closely the approach we took with those sites. 

Alternatively if you wish to keep the content independent you could just not provide the base and this approach would still work. 

Does this make sense? I'm happy to merge the existing code and discuss this further as we look to implement this further.

anthonyfok:
Dear @abourget,

Thank you so much, really, from the bottom of my heart, for your bravery and devotion in tackling multilingual support for Hugo.  Indeed, after seeing the multilingual efforts at Drupal, especially how it has evolved in Drupal 7, and now in D8MI (Drupal 8 Multilingual Initiative), I came to the appreciation how immense and complicated this subject can be, and I had been shying away from it.  Thank you @abourget for stepping forward and facing the challenge head on, and thank you for your thoughtfulness and extensive work that have gone into it thus far.

There are several use cases that I am personally interested in about Hugo's multilingual initiative.  For example, Hugo's multilingual support should:

1. allow an extensive multilingual site with 35+ languages like http://www.debian.org/, with mechanism to allow translators to easily see changes between revisions of the source language (English).  _(www.debian.org has been using the venerable [Website Meta Language (WML)](http://thewml.org/) as its static site generator with CVS as its main website infrastructure since around 1997. It still works, but it is very slow, taking about an hour to regenerate the entire Debian website.)_

2. handle multilingual documentation, e.g. http://gohugo.io/overview/introduction/, with ease, with accommodation for incomplete/outdated translation as it often takes translators a long time to catch up.

3. handle bilingual or trilingual parish websites, for example, that are common in bilingual cities or countries like Hong Kong (Chinese and English) and Canada (English and French)

4. handle variants of the same language, e.g. the flexibility to allow one Chinese translation source but generating both zh-Hant and zh-Hans, or even zh-CN, zh-TW, zh-CN, zh-MO and zh-SG like on http://zh.wikipedia.org/, with automatic and user-customizable conversion tables/engines.  Or, alternatively, if the user would rather keep zh-Hant and zh-Hans totally separate translations like how it is on Apple, Google and Microsoft websites, they can do that too.

But indeed, instead of speculating, I should start actually trying out your excellent PR and try building test websites with the above scenarios and see how it works.  :grin:

I will go experiment and report back.

By the way, do you already have some example/test websites that can demonstrate the multilingual features that you have written thus far?  :-)

Many thanks again!

Anthony

spf13:
For what it's worth, for us to do what I suggested above we need to first transition into a new content data model where a single file no longer is tied to representing a page. I'm currently working on the selective reading (only reading files that have changed) into Hugo, but I think I'm going to tackle that next and will do it with multilingual in mind.

abourget:
quick note @spf13 .. the "DefaultContentLang" does your 1., if `about.md` stays `about.md`, it'll be considered `DefaultContentLang`, which by default is `en`. I think it's in the docs I put in there. I'll review the rest of your comments in a few moments...

abourget:
Yeah a simple site is mine: https://blog.abourget.net

Le jeu. 7 janv. 2016 13:50, Steve Francia <notifications@github.com> a
écrit :

abourget:
The core idea here is linking the translations using a base filename (`about.md` links to `about.fr.md`, same as `about.en.md` links to `about.de.md`).  This solves huge complexities you normally see (in terms of UI, management), with IDs everywhere, and translation tables.

Regarding history and diffing, Git is the best we could bring to the table. Collaboration through GitHub and organizations there is really all you need, especially on large sites.  Simple tooling could be built to diff the changes to the "base" language that occurred after the last translation occurred for each piece of content... so smart git-fu and we're done. Perhaps this could be added to hugo itself a bit later. All of that, thanks to filename-based linking.

The next thing most important I think is what we offer to the template designer. Right now,  a `Page` has a `Translations` as `map[string]*Page`, mapping to corresponding translations (based on base-filename). This way, designers/devs could get the base translation directly, fall back to it as required.

Now there are many ways to address i18n in a CMS. My references come from Typo3 couple of years ago, where they offered two superbly divergent i18n models:

* _tree separation_: you'd have a completely different tree for each language.. you'd manage page hierarchy separately.. so if you wanted a piece of content on a different language, you needed to add it to the tree in that language. More granular, but more job.
* merged tree: the page hierarchy would imply that each page has or should have a translation. It would fall back to the "default" language.

In both cases, the pieces of content could be translated and linked together.

This PR currently implements this behavior:  the `Pages` (used to render the lists, nodes, taxonomies, etc..) is a single-language list of content pages. It uses the `RenderLanguage` to filter out the translations.. The translations are however attached to the `Page` object themselves under the `.Translations` field, as cited before.  If you have `about.fr.md` (and no `about.md` nor `about.en.md`), when you run with `RenderLanguage: en`, you will simply not see that article.

Now this isn't implemented here, but we could have a top-level switch for two different behaviors:

1. content isolation (current behavior): allowing links between translated content pieces, but which would have a different tree, its own subset of tags and taxonomies..
2. use a reference language: use the `DefaultContentLang` if no translation exists for the current run's _RenderLanguage_. This way, the _Site.Pages_ would potentially have some pages in different languages (a partially translated site). To implement this, we'd probably want to have a `.Page.ReferencePage` or `.Page.ReferenceTranslation` which would be defined (and guaranteed to be defined ?) only in "reference language" mode. Template designers could check some `.Page.IsTranslated` or `TranslationPresent` or something that would tell if indeed the ReferenceTranslation is present.... I don't know.. something along those lines.

@anthonyfok regarding the `35+` translations, using this PR would simply mean you need 35 config files.

@spf13 regarding translation of the taxonomies.. I feel this should be solved using my other PR, i18n, which adds a sort of native place for translations.  On my site, right now, I'm using a hacky `data` file to do some translation of the languages (see https://github.com/abourget/blog/blob/master/data/i18n/langlinks.yml and its use here: https://github.com/abourget/blog/blob/d5f5b9a2ef756eb2cd26c0db52f6547e03e010fa/themes/polymer/layouts/partials/header.html#L200-L221 ).. but I feel it could become a convention in i18n directories.
Like the translation string name could be `taxonomies_tags_golang: "Golang"` ..

@anthonyfok regarding the `en-US` and other things like that:

* this PR will work if you specify `RenderLanguage: en-US` and you use something like `about.en-US.md` for your files.. what we do not have is a way to fallback to `about.en.md` .. I think this could be added later withoiut breaking things too much.
* the other PR implementing i18n with `go-i18n` *does* that.. it parses the languages and does some matching (that I haven't looked how it works yet).. but it _will_ return translations for `en-US` if you ask the language `en` for example. Maybe that could be reused to implement such a logic when deciding on which translations to show in which `config.en.yaml` rendition.

@anthonyfok also, regarding the zh.wikipedia.org .. this PR does not facilitate the generation of subdomains.  For that I thought maybe in the future we could support a translatable `BaseURL` .. we'd need to wire that in deeply, and sorta trigger off the `/en` prefix to URLs, but it could work. I'd see that as a refinement, as some other options for the future.

@spf13 regarding base metadata, if we had `.Page.ReferenceTranslation`.. that'd mean some template designers could decide to use common params or stuff exposed on that `Page`.. or not.. it would be an additional indirection if you want to add that layer of complexity.  Otherwise, would be kept simple.. with no sharing.  What do you think of that ?

feew.. does that make sense ? sorry for the wall of text! :)

anthonyfok:
I have just taken the first step to test your PR: by trying to add multilingualization (m17n) to gohugo.io documentation.

What I have done so far (very little):

1. Locally merge this PR for testing
2. Compile Hugo with `go get -v github.com/spf13/hugo`
3. Run `hugo server -s docs` and read http://localhost:1313/content/multilingual/ to figure out what I should do next.
4. Edit docs/config.toml to add:

    ```toml
    Multilingual = true
    RenderLanguage = "en"
    ```

    (I used `RenderLanguage = en` the first time and got `Error: While parsing config: Near line 6 (last key parsed 'RenderLanguage'): Expected value but found 'e' instead.`, so I went back and added quotation marks around `en`.  Yes, minor differences between TOML and YAML.)

5. Found this in the doc:

    > You will also need two config files (config.en.yaml and config.fr.yaml for example) and you will need to run hugo twice, to render each language’s HTML.

    and got somewhat confused.  Do I just run `hugo` twice?  Or do I need to do two separate `hugo` runs, as in `hugo --config config.en.yaml` and `hugo --config config.fr.yaml`?

6. Restart `hugo server -s docs` and tried to reload http://localhost:1313/content/multilingual/ in the browser, and got a `404 page not found` error.
7. Manually add `en/` to the URL, making it http://localhost:1313/en/content/multilingual/, and could read the page again.
8. Visit http://localhost:1313/.  Got a blank page.
9. Try http://localhost:1313/en/ and finally see the original home page.
10. Ran `hugo -s docs` and realized that /index.html is no longer generated, only /en/index.html.

Speak for myself as a newbie and a somewhat lazy user, some of expectation at this stage (after merely enabling multilingual mode) would be:

* The site should work as before, i.e. before the m17n migration, which means:
    - a working /index.html, be it a real file or an alias (redirection) to /en/index.html (optional, on by default)
    - /content/multilingual/ should not give 404 but be created as an automatic alias to /en/content/multilingual/ or whatever the DefaultContentLang is (optional, on by default)

* The user would want to run Hugo only once, not twice, or 35+ times, when all these languages reside in different sub-directories under the same domain name.  Besides, how is the user supposed to run `hugo server` multiple times in order to see the entire website in its full m17n glory?

That said, I think what you already have is wonderful and is a great step forward, but there are a few details that we need to "get it as right as we can (@spf13)" first before presenting it for the first time to our end users, "especially the interface, that's the hardest thing to change for users (@spf13)".

So, I think it is time that we start a new topic branch—let's call it `m17n`—so that we can merge all your m17n pull requests in one place, and all others can experiment and contribute to, so we can try to "get it as right as we can" before we push this `m17n` branch onto `master`.  What do you think?  :-)

bep:
>The user would want to run Hugo only once, not twice, or 35+ times, when all these languages reside

Totally agree. Also with the other stuff you are saying, and do a fantastic job with this thorough testing.

abourget:
Adressing a couple of things here @anthonyfok:

1. Regarding point _5._ (having to run `hugo` multiple once per language with `--config config.XX.yaml` files):

  * I totally agree.

  * I see different solutions to this:

    1. Have `hugo` detect the `config.XX.(yaml|toml|json)` configs, and when it detects many configs, run many times.  We'd have to merge the watchers so that `serve` continues to work. I don't know the implication of this in the code.

    1. Support a single `config.yaml` with many YAML documents inside (separated by `---`). It could even "infer" `Multilingual: true` when it detects such multi-lang documents (we could remove that key altogether).. From an interface perspective, I think I'd prefer this one, as it would be much easier to convert your existing things to Multiliingual.. and I don't see any other use of such multi-config documents.

      1. From this strategy, we could also infer that the first section defines the `DefaultContentLang` value, removing yet another key.  If we implement a `LanguageBasePath` (see below), it could have a default value of `""`, unless explicited changed, so transition from multilingual to non-multilingual would be even seamlesser (whatever that means).

      1. When multiple docs are detected, we could _require_ the `RenderLanguage` to be present, otherwise fail upfront.

  * If we could agree on a first direction, I'd like to get a shot at implementing this in this PR.

1. Regarding point _6._ and _7._, with the site moving to `en/` and getting a 404 after running with `Multilingual: true`:

  * Right now, everything gets prefixed with the current language (RenderLanguage).  I would be quite simple to have an override flag for that, where you'd select which base you want to add there, or none.

  * It wouldn't make sense for new sites to have a copy, or a list of aliases from one of the languages to their `/en` counterpart. So in the case of a migration, perhaps the site owner could set `LanguageBasePath` to `""` when we don't want `en/` appended.  For this example, it could conserve the original tree, whereas the translations would be mounted under `fr/`, etc..

    * This feature could _eventually_ be extended to support `LanguageBasePath: "https://fr.subdomain.com/blog"` or something like that, where `/blog` would be used to prefix files on disk as well as links, and the domain part could replace the `BaseURL` for those links. This might be complex though, as on disk, it would all be rendered in the same directories as the other languages.. and there would be risks of overwriting the same files per translations. (like `blog/tags/hello/index.html`)

    * Subdomains are a bit tricky, as they don't map properly to what's on disk, and it's simpler to render everything to the same `public/` directory.

  * Such a feature could also allow people to mount the english translation to `english/` yet keeping the terse `post.en.md` naming scheme.

1. Regarding point _123._ (empty root):

  * [This draft blog post](https://github.com/abourget/blog/blob/master/content/post/hugo-multilingual-site-support.en.md#sitemaps) shows how to solve the empty root and empty sitemap problem.

  * This solution uses static files, where you or your theme plainly dumps those static files at root. This means you can't use any programmatic/templateable way to generate them, based on `LinkLanguages` or `DefaultContentLang` or something.

  * Also, the current PR does not expose knowledge about the other available `config.XX.yaml` language configs. Maybe there could be a convention in the `layouts` directory, like `multilingual_root_index.html` and `multilingual_root_sitemap.xml` and those would populate the absolute root when in `Multilingual`. They'd be ignored otherwise..

anthonyfok:
> But I don't think a topic branch is a way to go at this point. This is a complex problem to solve, but it should be possible to get this PR up to a working state before involving other people.

Sorry, I wasn't clear in my previous comment.  What I was thinking, first and foremost, was actually how to get abourget/hugo#1 into the fold.  With the newly created `m17n` branch, which already has this PR #1734 merged, @abourget can make a pull request against the `m17n` branch.

Also, I was also thinking, because it is such a complex problem to solve, we might need several PRs together to get all the features in, and working in the `m17n` branch would facilitate that.  We could involve other people, but we don't have to.  We could keep `m17n` purely to keep track of @abourget's current and future m17n-related PRs.

(Though I was also hoping I could contribute a PR or two on top of @abourget's existing PR to address the minor issues that I pointed out above.  :wink:)

Nevertheless, whichever way we work doesn't really matter too much.  Branches in Git are cheap, and we can create and delete them any time we wish.  If you prefer working entirely within the `master` branch, that is fine with me too, though in that case I would like to have Hugo prints a warning saying that "the new Multilingual feature is currently experimental, and its interface may still change at any time without notice".

anthonyfok:
Hello @abourget,

Thank you for addressing my concerns in details.

1.  About point _5_ (having to run `hugo` multiple times, once per language with `--config config.XX.yaml` files):

    * Regarding your proposed solutions:

        1. **The "current solution" at http://gohugo.io/tutorials/create-a-multilingual-site/** was contributed by a smart Japanese Hugo contributor who ingeniously got a multilingual website to work when Hugo itself does not yet know m17n.  It is an awesome workaround for Hugo v0.15 and previous releases, but I am sure your new m17n approach will offer something even more integrated and convenient.

        2.  **A single config.yaml with many YAML documents inside (separated by `---`):** Sorry, this does not work for TOML and JSON config file, does it?

    I propose one of the following:

    1. Keep using `config.(yaml|toml|json)` as the main config file, while letting Hugo read sub-config files from:

        * `config.XX.(yaml|toml|json)`, or
        * `i18n/XX.(yaml|toml|json)`, or
        * `m17n/XX.(yaml|toml|json)`, or whatever subdirectory we decide on.

        Hugo could either read all sub-config files from within the `i18n/` or `m17n/` directory, or, perhaps better yet, to read only the sub-config files of the languages that are listed in the main config file, e.g., `SupportedLanguages = [ "en", "fr", "ja", "pt", "pt-BR", "zh-Hans", "zh-Hant" ]` (TOML), or, in YAML:

        ```yaml
        SupportedLanguages:
          - en
          - fr
          - ja
          - pt
          - pt-BR
          - zh-Hant
          - zh-Hant
        ```

    2. Have everything listed inside the main `config.(yaml|toml|json)`, but not as one with multiple sub-YAML documents, but rather, a special `[i18n]` or `[m17n]` section where multilingual configuration are listed?

    Or were you thinking that each language could have wildly different configurations, say, they could be using entirely different themes, so it is not just a simple matter of having basic language-specific configuration?

anthonyfok:
Regarding point _6._ and _7._, with the site moving to `en/` and getting a 404 after running with `Multilingual: true`:

No, no need to copy any files.  :wink:

I was thinking of something much simpler: [Aliases](http://gohugo.io/extras/aliases/), only that the aliases are to be automatically created if `RenderLanguage == DefaultContentLang`, without requiring the user to manually add `aliases` to the front matter of each Markdown file.  This way, with automatic aliases, Hugo would automatically create a `/content/multilingual/index.html` with the following content:

```html
<!DOCTYPE html>
<html>
<head><link rel="canonical" href="http://gohugo.io/en/content/multilingual/"/>
<meta http-equiv="content-type" content="text/html; charset=utf-8" />
<meta http-equiv="refresh" content="0;url=http://gohugo.io/en/content/multilingual/" />
</head>
</html>
```

anthonyfok:
Regarding point _8._ (empty root):

>  * [This draft blog post](https://github.com/abourget/blog/blob/master/content/post/hugo-multilingual-site-support.en.md#sitemaps) shows how to solve the empty root and empty sitemap problem.

Yes, awesome!  That is exactly the solution!  Please add that as an internal template to Hugo and have such `/index.html` automatically generated somehow.  Turnkey solutions that are ready to go right out of the box!  :wink:

abourget:
@anthonyfok regarding the splitting of TOML and JSON files.. we currently implement splitting of such documents according to these rules for our frontmatter: https://gohugo.io/content/front-matter/

"+++" splits TOML documents, "---" splits YAML documents, and JSON documents are delimited by "}" and "{" on their own lines.

Right now, `viper` loads the configuration, but we could pre-process the config file, and split according to the same rules before passing to `viper.ReadConfig(io.Reader)` ...

I'm just PRd [a small change](https://github.com/spf13/viper/pull/152) to `viper` to be able to re-run `buildSite` and swap configs for each of the languages one after the other..

Would be _wwaaayy_ simpler to keep a single config.. and not ask people to drop a bunch of files here and there.  And as I said, having a single config would make any future inheritance a lot clearer.  Even better, maybe a certain key could manage inheritance for subdocuments:

```
InheritBaseLanguageConfigKeys:
  - menu
  - taxonomies
  - params
```

that would only be read/used when parsing the sub-documents.. so you could define that in the top-document.. in that case sub-documents would have those keys applied to their language.  In each of the languages, you could also selectively inherit some keys from the base document, which would merge keys in addition to those defined in the base config.

Maybe there'd be a way to use such a pattern for front-matter on base language @spf13 ?

abourget:
There's a bunch of new `MergeConfig` things in `viper` now :)

abourget:
@anthonyfok I know the branch is now `m17n`, but I don't think this contraction is widely used in the industry,.. I see mostly `i18n` and `l10n`, and they have different meanings. i18n is for internationalization, meaning you're targeting audiences of different languages. l10n is for localization, and means more adapting to locales, so how people read numbers, how they format dates and all..

Having a `Multilingual` key seems fair when exposed to laymen, as they'll understand it enables many languages.. whereas `I18n: true` might be a little crude, and `Internationalization: true` is .. dunno, long, error-prone.. unclear for the layman ?

abourget:
btw, localization is totally unaddressed with this PR and the `go-i18n` one.. I opened an issue over there https://github.com/nicksnyder/go-i18n/issues/45 to try to see what we can do.

abourget:
folks, this PR https://github.com/spf13/hugo/pull/1764 implements the multi-lang in one run .. it works but is open to discussion.

You can run `hugo serve` and it will render all languages, and watchers will also trigger a re-run of all languages.. see the PR for other details.

abourget:
ok, I don't know where to put this anymore.. 

Basically, this implements the promises found in https://github.com/spf13/hugo/pull/1764 .. with the new single-config-file layout and fallback variables under `Multilingual:`.

It updates the docs, tests and has everything working.

Let me know what you think about it ..

gntech:
Hello, I am following this PR with great interest since I am planning to make my hugo site multilingual. Is there any way I can help out getting this forward? I am no go-expert but if there is need for testing or User feedback I can help out.

abourget:
I'm using this in prod.. it's ready I think. Perhaps needs review.

Le mer. 24 févr. 2016 12:41, Gustav Näslund <notifications@github.com> a
écrit :

gntech:
To just try it out I thought I would try to build the golangmonteral.org site from source. Shouldnt the following command install your version of hugo with your commits applied?

go get -v github.com/abourget/hugo

I tried it but couldnt build this site that should be set up to use multilingual, right?

https://github.com/gomtl/golangmontreal.org

I get the following errors:

» ~/go/bin/hugo
Started building site
ERROR: 2016/02/24 22:26:51 template.go:350: template: theme/index.html:11: function "i18n" not defined
ERROR: 2016/02/24 22:26:51 template.go:350: template: theme/partials/footer.html:18: function "i18n" not defined
ERROR: 2016/02/24 22:26:51 template.go:350: template: theme/partials/header.html:469: function "i18n" not defined
ERROR: 2016/02/24 22:26:51 template.go:350: template: theme/partials/homepage-latest-news.html:15: function "i18n" not defined
ERROR: 2016/02/24 22:26:51 template.go:350: template: theme/shortcodes/talks.html:13: function "i18n" not defined
ERROR: 2016/02/24 22:26:51 shortcode.go:266: Failed to handle template for shortcode 'talks' for page 'gomtl-01-talks.fr': Template failed to compile
ERROR: 2016/02/24 22:26:51 shortcode.go:266: Failed to handle template for shortcode 'talks' for page 'gomtl-01-talks': Template failed to compile
ERROR: 2016/02/24 22:26:51 template.go:119: html/template: "theme/partials/header.html" is an incomplete template in theme/partials/header.html
ERROR: 2016/02/24 22:26:51 template.go:119: template: theme/partials/content.html:2:17: executing "theme/partials/content.html" at <.Site.CurrentLanguag...>: CurrentLanguage is not a field of struct type *hugolib.SiteInfo in theme/partials/content.html
ERROR: 2016/02/24 22:26:51 template.go:119: html/template: "theme/partials/footer.html" is an incomplete template in theme/partials/footer.html
ERROR: 2016/02/24 22:26:51 general.go:222: Error while rendering taxonomy terms for tag: template: theme/_default/terms.html:2:23: executing "theme/_default/terms.html" at <.Site.LanguagePrefix>: LanguagePrefix is not a field of struct type *hugolib.SiteInfo

What am I doing wrong?

abourget:
hmm.. I didn't rewrite imports.. so you'd need to rather:

    go get github.com/spf13/hugo
    cd $GOPATH/src/github.com/spf13/hugo
    git remote add abourget git@github.com:abourget/hugo
    git checkout -b multilingual abourget/master
    go install -v

and try that.

gntech:
Thanks for pointing me in the right direction! :) Now I have done some initial testing but I have run into a small issue. Let me explain:

I have swedish, "sv", as default language on my site and I have added .en. to the pages I want to translate to english. For example I have `about.md` default to swedish and `about.en.md` containing the english version of the same page. Here comes the problem:
The swedish version is rendered correctly to `/sv/about/index.html` but the english version is rendered to `/en/about.en/index.html`. However the url to the english about page points to the supposedly correct adress `/en/about/index.html`.

Do you recognize this problem?

abourget:
hmm.. that,s curious.. indeed.. it should strip those `.en` extensions..
not sure what's going on here.. I'd need to be able to see and debug.. is
that public ?

gntech:
Its available in the "multilingual" branch here https://github.com/gntech/gntech.se

bep:
This doesn't merge as-is into master ... And the commit log is ... messy (not the football player). Could you do some squashing and merge fixing? Also: PRs should be made from branches and not from master.

gntech:
(Not counting the file rendering bug mentioned earlier)

User feedback:
I must say that this multilingual implemenatation is working very well for my needs as a user. It works just the way I would like it too. The way i18n support is handled is very intuitive. `{{ i18n "string" }}`

gntech:
@abourget I am thinking about this problem with the file rendering path. Is it possible that I have installed your version of hugo wrong in some way? Can I be missing some dependency? Please see if you can recreate the problem on your side using this demo https://github.com/gntech/hugo-multilingual-problem-demo

gntech:
@abourget I have found the bug that causes the pages to be rendered at the wrong path. Its in line 988 in hugolib/page.go

```
	if len(strings.TrimSpace(p.Slug)) > 0 {
		outfile = strings.TrimSpace(p.Slug) + "." + p.Extension()
	} else {
		// Fall back to filename
->		outfile = helpers.ReplaceExtension(p.Source.LogicalName(), p.Extension())
	}
```

It is a fallback to use the p.Source.LogicalName() as the target path if no slug is set in the frontmatter. And that LogicalName function returns the filename with just the markdown-extension stripped. It does not strip the language suffix.

I looked at the source to your blog and I think the reason why you havent seen the problem in production at your side is because you have set a slug in the frontmatter for your translated pages.

I do not have enough technical insight to the structure of Hugo to propose a solution to this problem by my own. However if I could get a hint at how to implement a solution I would gladely help out submitting a fix for this problem.

gntech:
I have an updated branch here with my little fix as well as a proposed fix for the merge conflict present between abourget/master and hugo/master

https://github.com/gntech/hugo/tree/multilingual

abourget:
Yes I remember leaving that one out. I also think the :filename permalink
config section would be invalidly rendered.

I didn't have time to check more. I'll try to review this week. Thanks for
the patch !

Le mer. 2 mars 2016 16:18, Gustav Näslund <notifications@github.com> a
écrit :

abourget:
@bep I've rebased everything on top of latest master and fixed conflicts and tests.

@gntech this rebase includes your change

@bep it's the first time I ever hear about not doing PRs from another `master`.. I'd be curious to know the rationale though, if you have any link.  If you prefer I close this PR and open another one, I _do_ have an `m17n` branch in my repo.. I kept master to keep the context in here..

abourget:
Things have moved much, maybe it _would_ be better to open a different PR, what do you think @spf13  ?

bep:
@abourget the reason for not doing PR / feature development on your main branch is that not-related changes tends to end up in the same change-set. You may say that "I do not do mistakes like that", but believe me, it happens to the best of us.

Jean85:
Great PR! I would love to use this, it will fit perfectly in my case. Is there a way to extract a Hugo bin from this PR? Unfortunately I'm no Go expert :(

abourget:
It's still pending review.. it's drifted a bit. @spf13 have you gotten a chance to go through it ?  If you say it looks good, I'll rebase it and we can do a final sweep before merging it.

fedelibre:
@Jean85 AFAICS this feature is in the master branch of @abourget hugo fork, so you should do:

    go get git@github.com:abourget/hugo.git
    ln -s $GOPATH/bin/hugo $HOME/bin/hugodev

(in case you have two hugo binaries)

I'm looking forward to test this feature. I have a project in my TODO list which will benefit a lot from this PR.

bep:
>It's still pending review.. it's drifted a bit. @spf13 have you gotten a chance to go through it ? If you say it looks good, I'll rebase it and we can do a final sweep before merging it.

I notice you shoot out to Steve with all your questions, so I haven't looked closely at this. This is a nice feature, I want it, many people want it, but it isn't going into master without my blessing (or departure).

abourget:
Wow, fine. I didn't know about the approval process.

How can I help move that forward ? Someone suggested not having that from a
master branch in my repo. Would that help ? I can rebase it right away so
it's easier for you to check it out. What do you think ?

anthonyfok:
Based on my limited experience, I will try to answer your questions.
Everyone, please correct me if I am wrong.  :-)

> Someone suggested not having that from a master branch in my repo. Would that help ?

It is recommended that you use a non-master branch for creating pull requests: That would make **your** life easier after your pull request is accepted and you try to merge it back to your `master` without rebasing or creating an ugly merge bubble.  That said, AFAIK, no, it creates no difference in the review process.

AFAIK, once you have opened a pull request, GitHub does not allow you to choose another branch, so please leave this as is.  (Don't open a new pull request just to change to a non-master branch.)

> I can rebase it right away so it's easier for you to check it out. What do you think ?

Yes, please do.  GitHub says "This branch has conflicts that must be resolved", so please do that when you have some free time.  Many thanks!

gntech:
@abourget  If you have the possiblity and time please look into rebasing this pull request against the lastest master, I think that would greatly enhance the reviewing process.

abourget:
Here folks, I've rebased on latest master and fixed merge issues.

There seems to be issues in `tpl/`'s tests:

```
--- FAIL: TestFuncsInTemplate (0.00s)
	template_funcs_test.go:201: Got
		"absURL: http://gohugo.io/\nabsURL: http://mysite.com/hugo/mystyle.css\nadd: 3\nbase64Decode 1: Hello world\nbase64Decode 2: 42\nbase64Encode: SGVsbG8gd29ybGQ=\nchomp: <p>Blockhead</p>\ndateFormat: Wednesday, Jan 21, 2015\ndelimit: A, B and C\ndiv: 2\nemojify: I ❤️  Hugo\neq: current\nhasPrefix 1: true\nhasPrefix 2: false\nin: Substring found!\njsonify: [\"A\",\"B\",\"C\"]\nlower: batman\nmarkdownify: <strong>BatMan</strong>\nmd5: b3029f756f98f79e7f1b7f1d1f0dd53b\nmod: 0\nmodBool: true\nmul: 6\nplainify: Hello world, gophers!\npluralize: cats\nreadDir: README.txt\nreadFile: Hugo Rocks!\nrelURL 1: http://gohugo.io/\nrelURL 2: /hugo/mystyle.css\nreplace: Batman and Catwoman\nsafeCSS: \nsafeHTML: Bat&Man\nsafeHTML: Bat&Man\nsafeJS: \nsafeURL: http://gohugo.io\nseq: [1 2 3]\nsha1: c8b5b0e33d408246e30f53e32b8f7627a7a649d4\nsingularize: cat\nslicestr: Bat\nslicestr: Man\nsort: [A B C]\nsub: 1\nsubstr: Bat\nsubstr: Man\ntitle: Bat Man\ntrim: Batman\nupper: BATMAN\nurlize: bat-man\n"
		Expected
		"absURL: http://gohugo.io/\nabsURL: http://mysite.com/hugo/mystyle.css\nadd: 3\nbase64Decode 1: Hello world\nbase64Decode 2: 42\nbase64Encode: SGVsbG8gd29ybGQ=\nchomp: <p>Blockhead</p>\ndateFormat: Wednesday, Jan 21, 2015\ndelimit: A, B and C\ndiv: 2\nemojify: I ❤️  Hugo\neq: current\nhasPrefix 1: true\nhasPrefix 2: false\nin: Substring found!\njsonify: [\"A\",\"B\",\"C\"]\nlower: batman\nmarkdownify: <strong>BatMan</strong>\nmd5: b3029f756f98f79e7f1b7f1d1f0dd53b\nmod: 0\nmodBool: true\nmul: 6\nplainify: Hello world, gophers!\npluralize: cats\nreadDir: README.txt\nreadFile: Hugo Rocks!\nrelURL 1: http://gohugo.io/\nrelURL 2: /hugo/mystyle.css\nreplace: Batman and Catwoman\nsafeCSS: Bat&amp;Man\nsafeHTML: Bat&Man\nsafeHTML: Bat&Man\nsafeJS: (1*2)\nsafeURL: http://gohugo.io\nseq: [1 2 3]\nsha1: c8b5b0e33d408246e30f53e32b8f7627a7a649d4\nsingularize: cat\nslicestr: Bat\nslicestr: Man\nsort: [A B C]\nsub: 1\nsubstr: Bat\nsubstr: Man\ntitle: Bat Man\ntrim: Batman\nupper: BATMAN\nurlize: bat-man\n"
```

the diff'd piece is:

```
safeCSS: Bat&amp;Man
safeHTML: Bat&Man
safeHTML: Bat&Man
safeJS: (1*2)
```

vs:

```
safeCSS: 
safeHTML: Bat&Man
safeHTML: Bat&Man
safeJS: 
```

Do you think it's related to this PR ? or perhaps a dependency is missing from my machine..

abourget:
Also, I've read the note about `MainSite` being used by `caddy`'s hugo module.. This PR renames `MainSite` to `MainSites` since there are many sites.. this will break caddy's module, unless it's updated alongside this PR.

moorereason:
@abourget, update your `github.com/spf13/cast` repo.

abourget:
ok, great :) all tests pass! :)

bep:
The Caddy-plugin should be fine.

bep:
@abourget sorry that this has dragged out ... the master branch is in a stable state now (as in: No more refactorings etc. from me). It would be great if you could do a rebase against master so me, @spf13 and others could take this for a spin, review it and get it into master. Many people want this. I want this.

bep:
@abourget any news?

abourget:
Just had a new child :) I'd like to wrap this up before 2 weeks. Do you think we can do that ? Is timing good ?

bep:
Congratulations on the new child, @abourget  -- as to timing, we are long due for a new release, and I have "declared code freeze", whatever that means. I was hoping we could pull this in as the "last thing", as I know many are hoping for it. 

I still think we can, as long we can test it good enough to feel confident it doesn't break other important things. By that I mean: We can live with a bug or two in the multilangual-feature itself (we just label it as experimental or something, and fix everything for Hugo 0.17), but major regressions are painful.

The main thing is to get this rebased as soon as possible ... like, when the kid is sleeping :-)

abourget:
ok, just rebased it.. tests have passed here.. let's see how that goes :)

my site (golangmontreal.org) renders well with it

jasongonzales23:
By an amazing coincidence I too need this very badly, like before the end of this week. Keep up the work guys!

bep:
>By an amazing coincidence I too need this very badly, like before the end of this week. Keep up the work guys!

By an amazing coincidence we need people to test this feature. If you can build Hugo from source, please take it for a spin, both on existing Hugo sites and with "multilanguage" site. And then please report back what works and not..

gntech:
I can report that my multilingual site www.gntech.se builds without errors using this latest code.

hugolib/multilingual.go
```diff
@@ -0,0 +1,48 @@
...
+func (ml *Multilingual) GetString(key string) string { return cast.ToString(ml.Get(key)) }
+func (ml *Multilingual) GetStringMap(key string) map[string]interface{} {
+	return cast.ToStringMap(ml.Get(key))
+}
+
+func (ml *Multilingual) GetStringMapString(key string) map[string]string {
+	return cast.ToStringMapString(ml.Get(key))
+}
+
+func (ml *Multilingual) Get(key string) interface{} {
```
bep: It isn't clear to me what is part of the "multilingual config" and not. I tried to port my own blog to be multilingual, and intuitively put paginatePath into the multilingual section expecting that to just work (I notice the viper.Get("...") so I understand why it doesn't). I think all config that is used during build of a site should be per language.

docs/content/content/multilingual.md
```diff
@@ -0,0 +1,155 @@
...
+together as translated pieces. Only the content pieces in the language
+defined by **.Site.CurrentLanguage** will be rendered in a run of
+`hugo`.  The translated content will be available in the
+`.Page.Translations` so you can create links to the corresponding
+translated pieces.
+
+
+### Language switching links
+
+A full example of the language-switching links would be:
```
bep: You lost me a little with this example, and if you lost me -- it will be too complex for the regular Hugo.

This can be made much simpler, and we don't have to export this many nuts and bolts (variables and funcs) to the users.

**For a given page or node I'm interested in created a list of available translations (excluding the page I'm currently on), if available.**

So in pseudo code:

```
for range .Translations
   a href= .Permalink  .Language  // the page/node should know which language it represents
```







```

File: hugolib/multilingual.go [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+	if ml != nil && ml.config != nil && ml.config.IsSet(key) {
+		return ml.config.Get(key)
+	}
+	return viper.Get(key)
+}
+
+func (s *Site) SetMultilingualConfig(currentLang string, orderedLanguages []string, langConfigs map[string]interface{}) {
```
bep: This should not be exported. I suggest a NewSite(config) method.

hugolib/site.go
```diff
@@ -105,7 +106,8 @@ type SiteInfo struct {
 	Authors               AuthorList
 	Social                SiteSocial
 	Sections              Taxonomy
-	Pages                 *Pages
+	Pages                 *Pages // Includes only pages in this language
+	TranslatedPages       *Pages // Includes other translated pages, excluding those in this language.
```
bep: Not sure how useful TranslatedPages are. I would prefer AllPages; then I could filter with a where clause if needed.

hugolib/multilingual.go
```diff
@@ -0,0 +1,48 @@
+package hugolib
+
+import (
+	"github.com/spf13/cast"
+	"github.com/spf13/viper"
+)
+
+type Multilingual struct {
```
bep: I guess this could be unexported.

hugolib/site.go
```diff
@@ -124,6 +126,11 @@ type SiteInfo struct {
 	preserveTaxonomyNames bool
 	paginationPageCount   uint64
 	Data                  *map[string]interface{}
+
+	Multilingual    bool
+	CurrentLanguage string
+	LanguagePrefix  string
+	Languages       []string
 }
```
bep: Consider if we need to export all of the above, in the light of the suggested API-changes above.

docs/content/content/multilingual.md
```diff
@@ -0,0 +1,155 @@
...
+    copyright: "Tout est miens"
+
+copyright: "Everything is mine"
+```
+
+Anything not defined in a `lang:` block will fall back to the global
+value for that key (like `copyright` in this example).
+
+With the config above, all content, sitemap, RSS feeds, paginations
+and taxonomy pages will be rendered under `/en` in English, and under
```
jasongonzales23: Just to be clear the above example should be contained in, for example, a config.toml file at the root of the site?

docs/content/content/multilingual.md
```diff
@@ -0,0 +1,155 @@
...
+          {{end}}
+        {{end}}
+```
+
+This makes use of the **.Site.Languages** variable to create links to
+the other available languages.  The order in which the languages are
+listed is defined by the `weight` attribute in each language under
+`Multilingual`.
+
+This will also require you to have some content in your `i18n/` files that would look like:
```
jasongonzales23: Having little exposure to i18n in go or Hugo, these instructions and example are too vague for me to understand. What are `i18n` files for? How are they different from the translations you defined above? As I understand it, I can translated markdown files for content, and I can define translations in my config.toml (or yml) file, yes? If I am correct about that what role do these `i18n` fiels play?

docs/content/content/multilingual.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+### Translations of strings
+
+Hugo uses [go-i18n](https://github.com/nicksnyder/go-i18n) to support
+string translations.  Follow the link to find tools to manage your
+translation workflows.
+
+Translations are collected from the `themes/[name]/i18n/` folder, in
+addition to the files present in `i18n/` at the root of your project.
```
jasongonzales23: Some direction as to why I would want to choose one location over the other would be helpful.

jasongonzales23:
I'm actually not able to get it to work yet. With a minimal setup I get `function "i18n" not defined`. I must be missing something.

jasongonzales23:
Must I do anything more than `hugo` to generate the static files with i18n?

jasongonzales23:
BTW, to get the imports I needed to:

go get github.com/spf13/hugo
cd $GOPATH/src/github.com/spf13/hugo
git remote add abourget git@github.com:abourget/hugo
git fetch --all
git checkout -b multilingual abourget/master
go install -v

jasongonzales23:
my .toml files don't appear to be supported, is that coming?

jasongonzales23:
A few notes from the day of trying to get my company multilingual site working with this fork:

- It isn't clear to me how to build site nav as I currently have, that is, configuring it from the site config.yml. Is it possible to pass the current language in somehow into config??

- Many times while updating code the watcher didn't recompile the translations, so it was necessary for me to restart `hugo server` in order to get the site to render properly.

Cheers for all your work!

bep:
@jasongonzales23 TOML files work fine and current language is a static thing only accessible from the templates. Let us keep discussions/comments in this tread as focused as possible, and try not to drift into detailed problems with one particular site.

jasongonzales23:
One more note. It seems that if I don't create *some* kind of content for the non-default language, no page is generated.

For example, I have a static page at `content/getstarted/sample.pdf`

And I want that replicated under the `/de/` branch of the site. But currently when I generate the static site, that static file is not created for the German site, nor is a `/getstarted` directory. I assumed that unless I specified translated content, that the translated site would default (in my case, to English) and still mirror the site under all languages. Did I misunderstand?

jasongonzales23:
@bep I should have been more precise about the nav not working. With this fork, I can't yet figure out how to get the nav selected states nor how to get the nav to point to the correct subdirectories. These seem like critical features for release. If we can help document that clearly, I think it will greatly help the product.

gntech:
When I converted my site into multilingual I found the process quite
straight forward actually. But it is possible that I don't use the most
advanced features of Hugo.

There is a variable called
.Site.LanguagePrefix that can be used in links in for exemple the nav. It
is documented in a section called Multilingual Themes support in the
documentation.

There is also a full example of how to implement a language switching link.

@bep <https://github.com/bep> I should have been more precise about the nav
not working. With this fork, I can't yet figure out how to get the nav
selected states nor how to get the nav to point to the correct
subdirectories. These seem like critical features for release. If we can
help document that clearly, I think it will greatly help the product.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/spf13/hugo/pull/1734#issuecomment-219619697>

jasongonzales23:
Hey everyone,

I realize my feedback sounds pretty nit-picky, but I should be clear that I think I am actually a good use case for this update. I am migrating a fairly mature existing site to be multilingual with this. The good news is that I am able to get the site to work, but have had to rewrite some code to do so.

But the real big bug that is troubling me now is with LiveReload. Here is what I am seeing:

1. fire up hugo server
2. make some change to CSS (for example)
3. watch page reload and become another language
4. try directly going to the language I want to see
5. still see the same undesired language no matter where I go
6. scratch head, restart hugo server
7. do this every time you make a CSS change

Hope this helps!

jasongonzales23:
Discovered another bug, I can only get my 404 page to render in one language.

abourget:
You could add the JS detection we propose to your 404 page as a workaround,
but if you think about it.. how could you know the language without web
server having a say ?  There's no page! We can't decide from there where to
go.. unless JS detection is used.

I'll review all of these things and hope I can address concerns.

martybr:
I've got a problem when rendering menu's. If I enable "SectionPagesMenu" my sections are added to the menu but with an URL without the language prefix. For example the URL for the "post" section just becomes "/post/" instead of "/en/post/". The menu items defined in the front matter work correctly, i.e.. have the language prefix in the URL.

Also, I would expect the "menu" tree to be defined in the Multilingual block so it is possible to create custom localized menu entries. Or am I missing something?

One last item: There is a great example of using "apply" to generate tags and categories URL's. But because variables in a partial are limited to the scope you have supplied it is not possible to add the $.Site.CurrentLanguage variable to the URL you generate. I've now solved it with the "range" command but I really liked the simplicity of the "apply" example. Perhaps I'm missing something to get this working.

bep:
@abourget what is the status of this PR? I can understand people having life outside of open source development, but this is dragging out. If we cannot close this PR in a week or two I suggest we just close it, so we don't getting people's hopes filled up for no reason.

hugolib/multilingual.go
abourget: Last time I thought of those things, I tried to limit the things that would be multi-lingual aware, knowing we could easily add some more, but it would be difficult to remove some once people depend on them.

I remember some parameters would break across language switching, and that production of certain links to other languages required the same config for all language to make sense. I'm not sure about `paginatePath`.. but we could add it if it's missing, useful and doesn't break cross-language links generation.

abourget:
@jasongonzales23 I'm not sure I understand the `content/getstarted/sample.pdf` file.. is that in your `content` directly or is it the rendered path ?  Wouldn't you put static assetes like a .pdf in `static/` ? In that case it would only be rendered at the root of your site, and not a copy in each language.  But I feel that's not what you meant.

hugolib/site.go
abourget: This was geared towards simplifying the code in templates to show links to other languages.. which I assume is going to be used this way most of the time.

It could be `.Translations` and we'd filter out.. 

There are also considerations about the way we structure a site.  There are those paradigms where each "default" language page has an equivalent, and in the absence of a translated page, the "default language" version is shown.

There is also a model where you can have a page in a translated language without one in the default language.  That would create a different navigation tree. In that model, you wouldn't want a fallback to a default language, because you'd be okay with having no translation for a given page.

So those models are not directly related to this comment, but they need to be kept in mind, so we provide the right data for those who want to build each models.  Now I think this would require a bit more reflection.

abourget:
Last time I checked, this PR was in good shape.. I just didn't have time to address the list of comments.. I think mostly they are about better documentation.. and I agree with that.  I'd like those who test-drive the PR to add to the docs directly.. as it will come from experience.. I'm not 100% sure I'd cover all the things merely by reviewing or addressing a few comments.

The other thing is about exported/unexported vars.  I don't have strong opinions about exported/unexported Go functions.. as I personnally use `hugo` as a library.. and I'd assume most people use the command-line program.

The only thing that I'd want to get right are the variables exposed to template developers.. like `.TranslatedPages` and `.AllPages` suggestion. Choosing the right name for what the current variables do, and perhaps leaving namespace room to augment the available translation helpers/vars..

I'd also need to rebase this PR.. Since 0.16 is released, can we merge it when I rebase it.. so everyone can start contributing to it ? in terms of docs, or tweaks.. I think we agree on the large chunks.. and haivng it in the master branch would allow everyone to start poking at it.

abourget:
Checking at the conflicts just now.. and merging it after this rebase would be great, so I don't have to deal with things like down the road:

```diff
-       switch v.(type) {
+       switch val := v.(type) {
        case bool:
++<<<<<<< 5388211c1158b81725af04152c6e73eddc1435a0
 +              return v
 +      case time.Time:
 +              return v
++=======
+               return val
```

abourget:
Ok, just rebased. Not much was in conflict.  Tests passed over here. We'll see how it goes.

@bep can you list the blockers to a merge ? I'll make sure I address them, and then let's merge and have the community jump in to fine tune with the different use cases.

I use this branch on 2 sites right now, and planning a third.. so I'll put my hands back into it soon. https://golangmontreal.org and https://blog.abourget.net/  you can check the source for both sites ( https://github.com/gomtl/golangmontreal.org and https://github.com/abourget/blog )

abourget:
I've tweaked a few of my previous comments, perhaps you want to read them on Github instead of by e-mail :)

jasongonzales23:
@abourget, thanks so much. I am using this branch in prod, so would very much see it merged in soon!

bep:
@abourget my main comment is still untouched:

```
You lost me a little with this example, and if you lost me -- it will be too complex for the regular Hugo.

This can be made much simpler, and we don't have to export this many nuts and bolts (variables and funcs) to the users.

For a given page or node I'm interested in created a list of available translations (excluding the page I'm currently on), if available.

So in pseudo code:

for range .Translations
   a href= .Permalink  .Language  // the page/node should know which language it represents
```

docs/content/content/multilingual.md
abourget: yeah I'll add a note on that..

docs/content/content/multilingual.md
abourget: yes, ok I agree.. I started at the beginning with `.Translations` excluding the current language, and I don't remember the reason why I added the current translations (I think it might have had to do with the fact that otherwise there's no place where you can access a complete picture of all the pages on a site.).

I agree with reverting that to the behavior you propose.  It will indeed simply the example (get rid of `{{ if ne $lang $txLang}}).... we'll still need `{{ if isset $.Translations $txLang }}` and the rest.. so it'll remove 4 lines in the example.

Do you see other ways we can shrink that ?

abourget: part of the complexity in here is to keep showing the links in other languages even if the pages there aren't translated.

perhaps we can show a simpler example (with no `{{ if .Site.Multilingual }}`) and a more complete example..  so we don't turn off new users. What do you think ?

abourget:
Ok @bep  I pushed fixes and simplifications.

hugolib/multilingual.go
bep: Sure, but the thing is, if you don't cover ALL the stuff that creates files on disk, you will end up with a lot of broken sites. I know it demands some work, but I thought that was this.

Which isn't a feature I'm merging.

hugolib/site.go
bep: I suggest removing Translagedpages.

docs/content/content/multilingual.md
```diff
@@ -0,0 +1,170 @@
...
+translated pieces.
+
+
+### Language switching links
+
+Here is a simple example if all your pages are translated:
+
+```
+{{if .IsPage}}
+  {{ range $txLang := .Site.Languages }}
```
bep: This and below isn't what I talked about in my previous commit, is it?

abourget: yes.. that's the simplest form I found and use.. if you have something better to propose.. I'm all ears.

hugolib/multilingual.go
abourget: We have to start somewhere.  I'm not pretending the thing is perfect.  But i'd like to let the community give it a try.

This would not break existing sites.. only potentially take more time when community members start building multilingual sites, as we sort out all the kinks. I know it's a big chunk.. but aren't we better to bring it in sooner in the release process, rather than close to a release date?

If I list in the documentation all the keys that are picked-up under `Multilingual`, would that be satisfying ?  I could try to list the keys that are explicitly _not_ taken into account.. if I can build such a list.  Otherwise, I'm a bit blocked.. as I'm not sure what else would need to be done.

hugolib/site.go
abourget: ok I found it.. `site_test.go` reveals the issue.  Basically, with content files like this (where `DefaultLang = "en"`:

    content/doc1.en.md
    content/doc1.fr.md
    content/doc2.fr.md

The `doc2.fr.md` file will be totally unreachable, as `.Pages` only lists the current language's pages (assuming we're rendering the `en` site).

In that case, we could have `AllPages` as you suggested.. where you'd filter out.. That would do the trick.

hugolib/multilingual.go
abourget: I've looked into it for `paginatePath`, and it would require quite a bit of rework to have it there... because where `viper.Get("paginatePath")` is used is often outside the scope of a `Site`..

For now, I'd tolerate the fact that this piece isn't translated nor translateable.  I'm pushing some improved docs regarding this.

hugolib/site.go
abourget: since we have `.Pages` already, and `setupTranslations` tweaks it to remove pages from the other language, we'd need to make `.AllPages` the main reference throughout the code in `site.go`, and in `setupTranslations`, filter out the non-current-language pages and put them in `.Pages`.

A little more involved..

abourget:
@jasongonzales23 the latest code should solve a bunch of your issues.. namely the LiveReloading not picking up translations (because of the way .Pages and .TranslatedPages worked).. Now .AllPages includes all translations and watch all those files for changes before rebuilding.

Everything is rebased on master now.

the1900:
I really want this feature.

When I didn't know about this issue. I've tried to make multilingual different way.

The section is the language type what I want to publish. `example.com/en/who-am-i` is in the en folder and `example.com/ko/who-am-i` is in the ko folder.
But this way has a many problem like index.html(homepage) for each languages or duplication Taxonomy (same content with tags but different languages)

dukebarman:
When merge? :)

spil-rkoopmans:
I really want this feature too. Is there an estimate when this will be merged? Will I be safe if I use @abourget his fork in production? I really need this for production sites or I am afraid I have to resort to something different

bep:
@spil-rkoopmans It will be merged when it is solid enough to merge.

docs/content/content/multilingual.md
bep: I have proposed a simpler form above.

spil-rkoopmans:
thank you @bep. I understand. I have decided to go with Hugo anyways using the method described in the v0.16 documentation. I really love this product and think you guys are doing a great job! We will see what happens in the future and adjust our project accordingly!

docs/content/content/multilingual.md
abourget: I do not see how the form I documented can be simplified. Your propositiin doesn't take into account the differences between Pages and Nodes and the potential different decisions a user might take.

 I see this as complete , unless we want to abstract those decisions into yet another concept different than Pages and Nodes, something that would unite them into a Translatable or something.

bep:
Merged into https://github.com/spf13/hugo/tree/multilingual

I will walk this one into master when I get some time to spare. Thanks for the great job.

abourget:
Continued here: https://github.com/spf13/hugo/pull/2303

PR #1739, opened 2016-01-01
Update press.md

ominty:
fix tables, add new entry


anthonyfok:
Thank you!

Note that the publish date of Armin Hanisch's blog article is actually 18 Aug 2015, not 01 Jan 2016.
Nevertheless, I have decided to keep it at the top of the list as you have done so users will see it a as a newly listed article, but will will move it down to its place probably a month later so the entire list is properly sorted reverse chronologically.

  Merged as b309fe8 (master) a03d022 (v0.15.docs).

Issue #1644, opened 2015-11-29
Partial Static Sync


spf13:
Today Hugo synchronizes the entire static directory each time a file change is detected. While it only copies the changed file(s), it still goes through the process of evaluating all files on both sides to determine which file needs to be synced. This isn't necessary as we already know what file changed. We should just sync that file.

anthonyfok:
Hello @ffrizzo,

Thank you for your contribution, @ffrizzo!  But I think I encountered a bug in your current implementation in #1671: the static directory in the theme is not synchronized.  It detects that I changed the style.css (which is inside the theme's static/ directory), but it is probably trying to copy it from the wrong place.

ffrizzo:
Hello @anthonyfok ,

I'll take a look on that and I'll submit a new pull request with the fix.

Skrylar:
On my box, it actually copies a bit more than just the changed file. Every single file is written to the output directory for every single run of hugo, with the exception of those in the static files directory.

spf13:
@Skrylar can you provide more information about that? Can you paste the output running with `--verbose`?

Skrylar:
@spf13 
```
INFO: 2016/02/21 00:08:56 hugo.go:439: Using config file: c:\Dev\Hugotest\config.toml
WARN: 2016/02/21 00:08:56 hugo.go:522: Unable to find Theme Static Directory:
INFO: 2016/02/21 00:08:56 hugo.go:542: c:\Dev\Hugotest\static\ is the only static directory available to sync from
INFO: 2016/02/21 00:08:56 hugo.go:580: syncing static files to c:\Dev\Hugotest\public\
INFO: 2016/02/21 00:08:56 site.go:1275: found taxonomies: map[string]string{"tag":"tags", "category":"categories"}
WARN: 2016/02/21 00:08:56 site.go:1995: Unable to locate layout for 404 page: [404.html]
WARN: 2016/02/21 00:08:56 site.go:1971: "404.html" is rendered empty
0 draft content
0 future content
10 pages created
0 non-page files copied
0 paginator pages created
0 tags created
0 categories created
in 31 ms
```

If I force the cache directory to be somewhere (ex. in the "cache" folder of the same directory), no files are left behind there. I'm not sure if that's intended or a symptom.

jdevoo:
[...]

Issue #1559, opened 2015-11-08
IgnoreFiles for static folder


fmpwizard:
As discussed here
http://discuss.gohugo.io/t/slow-to-work-with-large-number-of-static-files/2046

Update `IgnoreFiles` to also affect files in the `static` folder.

ffrizzo:
I'll work to fix that issue.
But I would like to discuss if is better create a separated method or I just use the hugo.copyStatic and checking if we need ignore file or not.

mpaluchowski:
Will this also cover the `/static` folder of themes?

SiamKreative:
From my understanding this has already been merged. However, I have the following in my **config.toml** `ignoreFiles = [ "\\.less$"]` and Hugo still detect changes. What am I doing wrong?

![image](https://cloud.githubusercontent.com/assets/1778633/26048000/8aca90dc-397f-11e7-9867-0a71993c83cf.png)

lwchkg:
I'm wondering what is meant exactly for `ignoreFiles`. If I specify `a.md` there, do I mean `content/a.md` or `a.md` in the root?

BTW, I'm considering whether we should split `ignoreFiles` into 3 parts: `ignoreContent`, `ignoreStaticFile` and `ignoreForWatch` (files in there are not watched for site rebuilds).

inwardmovement:
I can't manage to make Hugo ignore `static/css` and `static/js` folders with `ignoreFiles: ["static//css//.*.css$", "static//js//.*.js$"]`... the regex is right according to [regex101](https://regex101.com/r/4704Jp/1), is this a Hugo bug?

moorereason:
@inwardmovement, please use the [Discussion Forums](https://discourse.gohugo.io/).

paulswanson:
I would really like to some progress on this issue / request. Here's an example. I have SASS folder under theme, which is precisely where it should live, and I need it to be ignored from the build. However, the poorly documented "ignoreFiles" options only seems to target specific folders, again undocumented. It would be incredibly helpful, if this feature were simply broadened to accept any particular path.

paulswanson:
I'm happy to have a go at contributing a pull request for this, if there's any general will to make the ignoreFiles option a little more versatile.

pauby:
If this is still live, should it not be re-opened?

XhmikosR:
Yeah, this is still an issue.

pauby:
[...]

PR #1741, opened 2016-01-02
Copy static files changed

ffrizzo:
Fix watcher copy static theme files for
changed files

See #1644 #1559


spf13:
Merged as 8f6f131

PR #1745, opened 2016-01-03
Update doc to reflect default value of watch variable

k4rtik:
This occurred with the change at https://github.com/spf13/hugo/commit/97d068958048daf65d84b2a8357f6f6fa3b207d1


spf13:
merged as 0bb3b49

PR #1746, opened 2016-01-03
Small refactor: don't call `p.Permalink()` for all aliases

abourget:
call it only once and only if necessary.


spf13:
Merged as dcad066. Thanks @abourget !

PR #1747, opened 2016-01-03
Doc updates

abourget:
This one might need review.. I found out about `partials` not having access to `$`.. I would've liked it did.. but at least knowing it helps.


spf13:
Merged as bc48b46

PR #1748, opened 2016-01-03
Replace *bytes.Buffer with io.Writer in rendering

mvdan:
Found via github.com/mvdan/interfacer.

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -1682,14 +1682,14 @@ func (s *Site) renderAndWritePage(name string, dest string, d interface{}, layou
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


bep:
Merged in 7c5a1fd ... and thanks for the quality commit message.

mvdan:
Sarcasm? :)

bep:
No. I have fought pretty hard to get people, esp. @spf13, in line with our simple commit message guidelines ... Yours may be short (which in this case is good), but spot on correct.

mvdan:
Oh, hadn't noticed the commit message guidelines. Saw the usual fmt, rebase and test and assumed it was as usual.

My personal commit message policy is very similar to yours, so thankfully we're on the same page. Although I try to keep the bodies short whenever possible, unless they really must explain something.

PR #1751, opened 2016-01-04
Updated content in "What is Hugo?" session

emilylauren2490:


spf13:
I'm pretty sure Comodo Cloud doesn't support hosting websites? If I'm wrong, then please update your website with directions how and reopen.

PR #1756, opened 2016-01-05
Update press.md

ominty:
New entry: Harald Weidner in german at Cologne Linux Meeting 2015: Link to set of slides available on bottom of page


spf13:
Merged as 9d70e78

PR #1764, opened 2016-01-09
WIP: Implemented multi-site rendition in one `hugo` run.

abourget:
This is the first draft to a single-file multi-document YAML config.

You can run: `hugo -D serve` and it will rebuild all languages in a swift.

There's much room for optimization after this.  We could re-use the `Site` object, caching the files that were read.  We could simply call `Build()` again on the `Site`.. however that would need to be tested to ensure nothing leaks between runs..

This requires https://github.com/spf13/viper/pull/152 to be merged in
before compiling.

With this merged in, a file like:

```
---
theme: "polymer"

baseurl: "https://blog.abourget.net"
languageCode: "en-US"
title: "geek joy"

paginate: 10 # optional
disqusShortname: "blogabourget" # optional
copyright: "(ɔ) 2015 Copyleft. Alexandre Bourget."

Multilingual: true
RenderLanguage: en
DefaultContentLang: en
LinkLanguages:
 - en
 - fr

params:
  author: "Alexandre Bourget"
  profile: "images/profile.jpg" # optional
  cover: "images/cover-bw.jpg" # optional
  twitter: "bourgetalexndre" # optional
  github: "abourget" # optional
  linkedin: "alexandrebourget" # optional
  googleAnalyticsUserID: "UA-12745853-1"

permalinks:
  post: "/:year/:month/:day/:filename/"

MetaDataFormat: "yaml"

---

theme: "polymer"

baseurl: "https://blog.abourget.net"
languageCode: "fr-CA"
title: "geek joy"

paginate: 10 # optional
disqusShortname: "blogabourget" # optional
copyright: "(ɔ) 2015 Alexandre Bourget. Tous droits renversés."

Multilingual: true
RenderLanguage: fr
DefaultContentLang: en
LinkLanguages:
 - en
 - fr

params:
  author: "Alexandre Bourget"
  profile: "images/profile.jpg" # optional
  cover: "images/cover-bw.jpg" # optional
  twitter: "bourgetalexndre" # optional
  github: "abourget" # optional
  linkedin: "alexandrebourget" # optional
  googleAnalyticsUserID: "UA-12745853-1"

permalinks:
  post: "/:year/:month/:day/:filename/"

MetaDataFormat: "yaml"
---
```

works.

Now it shows needs for deduplication.. but that can be managed in a second time.

/cc @spf13 @anthonyfok

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -56,6 +56,7 @@ func readInMultilingualConfig(cfgFile, source string) error {
 		return fmt.Errorf("found 0 config blocks in %q", filename)
 	}
 
+	// Reset config to base language.
 	viper.SetDefaultConfig(translationsConfigs[0])
 
 	return nil
@@ -76,14 +77,5 @@ func viperSetAll(key string, value interface{}) {
 	viper.Set(key, value)
 	for _, conf := range translationsConfigs {
 		conf.Set(key, value)
-	}
-}
-
-// viperSetDefaultAll calls `viper.SetDefault` on all translations'
-// configs.
-func viperSetDefaultAll(key string, value interface{}) {
-	viper.SetDefault(key, value)
-	for _, conf := range translationsConfigs {
-		conf.SetDefault(key, value)
 	}
 }
--- a/commands/config.go
+++ b/commands/config.go
@@ -1,0 +1,89 @@
+package commands
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/spf13/viper"
+)
+
+var translationsConfigs []*viper.Viper
+
+func init() {
+	translationsConfigs = make([]*viper.Viper, 0)
+}
+
+// readInMultilingualConfig reads the configuration, and loads any translations
+// sub-documents in the main config file.
+func readInMultilingualConfig(cfgFile, source string) error {
+	viperSetConfigFile(cfgFile, source)
+
+	// TODO: ignore error in ReadInConfig, simply use that for
+	// discovery of the file.  Then split the config file in chunks,
+	// identify the chunks and feed to ReadConfig().  This could be
+	// replaced if `viper` exposed the File without reading it.
+	_ = viper.ReadInConfig()
+	filename := viper.ConfigFileUsed()
+	cnt, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+
+	// Initial support for YAML only:
+	for i, configDocument := range strings.Split(string(cnt), "\n---") {
+		if len(strings.TrimSpace(string(configDocument))) == 0 {
+			continue
+		}
+		viper.SetDefaultConfig(viper.New())
+
+		viperSetConfigFile(cfgFile, source)
+
+		err := viper.ReadConfig(bytes.NewBuffer([]byte(configDocument)))
+		if err != nil {
+			return fmt.Errorf("in config document %d: %s", i+1, err)
+		}
+
+		viper.RegisterAlias("indexes", "taxonomies")
+
+		LoadDefaultSettings()
+
+		translationsConfigs = append(translationsConfigs, viper.DefaultConfig())
+	}
+
+	if len(translationsConfigs) == 0 {
+		return fmt.Errorf("found 0 config blocks in %q", filename)
+	}
+
+	viper.SetDefaultConfig(translationsConfigs[0])
+
+	return nil
+}
+
+func viperSetConfigFile(cfgFile, source string) {
+	viper.SetConfigFile(CfgFile)
+	// See https://github.com/spf13/viper/issues/73#issuecomment-126970794
+	if Source == "" {
+		viper.AddConfigPath(".")
+	} else {
+		viper.AddConfigPath(Source)
+	}
+}
+
+// viperSetAll calls `viper.Set` on the all translations' configs.
+func viperSetAll(key string, value interface{}) {
+	viper.Set(key, value)
+	for _, conf := range translationsConfigs {
+		conf.Set(key, value)
+	}
+}
+
+// viperSetDefaultAll calls `viper.SetDefault` on all translations'
+// configs.
+func viperSetDefaultAll(key string, value interface{}) {
+	viper.SetDefault(key, value)
+	for _, conf := range translationsConfigs {
+		conf.SetDefault(key, value)
+	}
+}
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -519,8 +519,9 @@ func getDirList() []string {
 }
 
 func buildSite(watching ...bool) (err error) {
-	startTime := time.Now()
+	defer viper.SetDefaultConfig(translationsConfigs[0])
 
+	startTime := time.Now()
 	for _, langConfig := range translationsConfigs {
 		viper.SetDefaultConfig(langConfig)
 
@@ -534,7 +535,7 @@ func buildSite(watching ...bool) (err error) {
 		}
 		site.Stats()
 		if len(translationsConfigs) > 1 {
-			jww.FEEDBACK.Printf("translation %q rendered", viper.GetString("RenderLanguage"))
+			jww.FEEDBACK.Printf("translation %q rendered ", viper.GetString("RenderLanguage"))
 		}
 		jww.FEEDBACK.Printf("in %v ms\n", int(1000*time.Since(startTime).Seconds()))
 	}
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -252,14 +252,8 @@ func LoadDefaultSettings() {
 // A Hugo command that calls initCoreCommonFlags() can pass itself
 // as an argument to have its command-line flags processed here.
 func InitializeConfig(subCmdVs ...*cobra.Command) error {
-	viper.SetConfigFile(CfgFile)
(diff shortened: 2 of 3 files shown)


abourget:
btw, this will totally break TOML and JSON configurations.. we'd need to have explicit support for splitting TOML and JSON files according to our rules here https://gohugo.io/content/front-matter/

anthonyfok:
> btw, this will totally break TOML and JSON configurations.. we'd need to have explicit support for splitting TOML and JSON files according to our rules here https://gohugo.io/content/front-matter/

Sorry, I know you have put in a lot of time into this, but perhaps I wasn't clear enough in my previous comments...  Using YAML's `---` to split a config.yaml into 2, or 3, or 35 parts, that just doesn't sit well with me.  It seems somewhat a kludge, and unstructured...

And forcing TOML and JSON config files into the same mold when it is not in their DNA... (Oh no!)

Is there anyway we can have multilingual configuration in a nested structure within the main config file instead?  (Unfortunately, I don't know YAML well enough to write an example on-the-spot.)  :blush:

abourget:
What I would expect to unclutter would be:

```
---
theme: "polymer"

baseurl: "https://blog.abourget.net"
languageCode: "en-US"
title: "geek joy"

paginate: 10 # optional
disqusShortname: "blogabourget" # optional
copyright: "(ɔ) 2015 Copyleft. Alexandre Bourget."

Multilingual: true
RenderLanguage: en
DefaultContentLang: en
LinkLanguages:
 - en
 - fr

params:
  author: "Alexandre Bourget"
  profile: "images/profile.jpg" # optional
  cover: "images/cover-bw.jpg" # optional
  twitter: "bourgetalexndre" # optional
  github: "abourget" # optional
  linkedin: "alexandrebourget" # optional
  googleAnalyticsUserID: "UA-12745853-1"

permalinks:
  post: "/:year/:month/:day/:filename/"

MetaDataFormat: "yaml"

InheritBaseConfig:
 - theme
 - baseurl
 - title
 - paginate
 - disqusShortname
 - DefaultContentLang (this could be implicit though)
 - LinkLanguages (maybe also ?) 
 - params
 - permalinks
 - MetaDataFormat
---

RenderLanguage: fr
languageCode: "fr-CA"
copyright: "(ɔ) 2015 Alexandre Bourget. Tous droits renversés."

---

RenderLanguage: ru
languageCode: "ru-BMA"
copyright: "(ɔ) 2015 Alexandre Bourget. All russian rights reversed."

```

That makes more sense if in the same file.. as the flow of the file allows you to quickly view and tweak the site-wide configs.

I'm not against having a bunch of of `config.TX1.yaml`, `config.TX2.yaml` with a base `config.yaml` for example.. but putting that in sub-directories with i18n/data.. that wouldn't be pretty.. config lives in one location: the root dir.

* With such a setup, the list of LinkLanguages could be inferred from the `config.LANG.yaml`.. and even the RenderLanguage could be infered from there.  Now for inheritance, we'd still need something.. 

* Looking at the keys in my example config.. it seems that simply having a copy of the same keys, and overwriting certain other keys could bring us a long way.

spf13:
I think the issue here is we use the Config file today for two things. App
configuration and site configuration. Only the latter would need different
values per language. I suggest we split them up and have one site Config
file per language.

spf13:
And to be clear, still use the base/override approach with the site Config.

bep:
In my head your thinking is a little bit too constrained by today's situation (with the multilanguage tutorial with the many configs ... which makes sense as a workaround).

When building this as a feature you can ... build it properly.

@anthonyfok comes with maybe an extreme use case (debian.org 30+ langs), but in my head this is how it should be done:

* The language files themselves need to be split into files (this is how this is, as I see it)
* The language configuration per language should be minimal - and should be easy to model as a fairly simple `languageConfig` struct, with inheritance built in. One `config.toml` file is a good, simple mental model that I think we should keep.

abourget:
@spf​13 can you elaborate on the base/override model you had in mind ?

abourget:
and also, where you'd cut the line between app and site config ?

spf13:
I've given this more thought and agree with @bep. 

I think we should stick with the one config file we have (for now.. I'm still considering splitting it into two, but independent of this). 

There's a set of fields that are site specific parameters. Some are universal and others are created for a specific site.

For multilanguage we should have a specific dictionary in the config that conveys this.. I'll give an example

```yaml
---
baseurl: "http://yoursite.example.com/"
title: "Steve cooks food"
permalinks:
  post: /:year/:month/:title/
params:
  Subtitle: "baking all the things"
  AuthorName: "Steve Jones"

multilanguage:
  en:
    title: "Steve cooks food"
  es:
    title: "Steve prepara comida"
    params:
      Subtitle: "todas las cosas de la hornada"
      AuthorName: "Estaban Jones"
  ru:
    title: "Steve варится еда"
    params:
      Subtitle: "Выпечки все вещи"
```

For any of the language keys we would overlay them on top of the base definitions at the top.. So If I was requesting params.AuthorName while building the Russian content I would get "Steve Jones" as the base config provides that.  If I requested the same params.AuthorName while building the spanish content I would get "Estaban Jones" as that key is overwritten by the local "es".

Does that help make more sense?

Viper will get us part of the way there, but we will also have to rewrite some of the GetParams and similar functions to become "multilingual aware".

abourget:
Hmm Steve, it seems to me that those things belong to translations.. not to
site config. I understand that before i18n exists with translations
strings, it makes sense to have it in there.. but a theme could provide
translations for its base strings, and the way to override its values
should be in i18n/en-US.yaml and friends.

with the other go-i18n PR, you can override string by string..

What do you think?

Le lun. 11 janv. 2016 22:12, Steve Francia <notifications@github.com> a
écrit :

bep:
>Hmm Steve, it seems to me that those things belong to translations.. not to site config. 

Steve was demonstrating a concept, not which values should go where: Only one config file with the per language specific config in one struct -- and the TRANSLATIONS in a per-file. Please come up with a design, then we can all have a look at it and see if we all agree.

abourget:
ok, I see.. yeah... I'll try to come up with something concrete.. in reality, not much of the `viper.SetDefault`  in `LoadDefaultSettings()` need to be multilingual aware.. and this PR does most of the work for them..

I think it'd make sense to bring vars for multilingual under one key..

abourget:
What about something like:
```
---
theme: "polymer"

baseurl: "https://golangmontreal.org"
languageCode: "en-CA"
title: "Golang Montréal - #gomtl"

paginate: 10 # optional
copyright: "(ɔ) 2015 Copyleft. The #gomtl team."

MetaDataFormat: "yaml"

permalinks:
  post: "/:year/:month/:day/:filename/"

params:
  profile: "images/profile.jpg" # optional
  twitter: "GolangMontreal" # optional
  github: "gomtl" # optional
  googleAnalyticsUserID: "UA-XXXXXXXX-1"

DefaultContentLang: en

Multilingual:
  en:
    weight: 1
    params:
      author: "#gomtl team"

  fr:
    weight: 2
    params:
      author: "l'équipe #gomtl"
    languageCode: "fr-CA"
    copyright: "(ɔ) 2015 L'équipe #gomtl. Tous droits renversés."

---
```

This way we could get rid of `LinkLanguages` (to generate language switching links, could be programmatically determined with `weight`s)

No need for `RenderLanguage` .. would be taken from the run, once for each key under `Multilingual`.

`Multilingual: true` would be inferred by having anything under `Multilingual: map[string]interface{}`

We'd still keep `DefaultContentLang`, unless we add a key like `DefaultContentLanguage: true` under one of the `en:` or `fr:` keys.

Params could be merged from language, inheriting from site configs,.. on a key-by-key basis.  `copyright`, `title`, `languageCode`, `baseurl` would also be fall-back keys.. perhaps along some other keys I missed.

To keep things simple, I wouldn't put things like `permalinks` in the inheritance model, as it would complexify creation of cross-language links... A bunch of other keys are in the same situation, in which case we wouldn't have them inherited.

Does that make more sense ?  I think that'd be a lot better than what is in here.

abourget:
Also would have the side-effect of easing transition to multi-lingual for existing sites, merely adding:

```
DefaultContentLang: en
Multilingual:
  en:
    weight: 1
  fr:
    weight: 2
    params:
      author: "overrides"
    languageCode: "fr-CA"
    copyright: "(ɔ) 2015 Other override here.."
```

would create a multi-lingual site.

This alone wouldn't solve having aliases from the old URLs to the new ones though.

spf13:
Yes, I think this approach makes a lot more sense. I agree that things like
permalinks should be in the universal config.. not in a language specific
config. We should probably define these "universal" vs "language specific"
or "localizable" ahead of time.

abourget:
Ok great.. I'll go ahead and implement something like that.. Indeed listing the properties somewhere would be useful..

Since we have a m17n branch now, would you be okay with me merging the go-i18n branch within this one.. and refactor the whole thing.. right now I'm doing efforts to split the PRs, but my sites use both.. as you kind of need i18n for strings for any sane multilingual theme anyway :)

Example running site: https://golangmontreal.org and its source https://github.com/gomtl/golangmontreal.org

spf13:
I'm totally fine with this being a single pull request. We're definitely
giving feedback the whole way.

abourget:
I think we can follow up on https://github.com/spf13/hugo/pull/1734 .. there's only one branch about this, and everything is in there.

Issue #1772, opened 2016-01-13
On arm64, "hugo -s docs" fails with "function not implemented"


anthonyfok:
According to the Debian arm64 build daemon [build log](https://buildd.debian.org/status/fetch.php?pkg=hugo&arch=arm64&ver=0.15%2Bgit20160109.147.dd1d655-1&stamp=1452558812), a `hugo -s docs` would fail with "function not implemented" error:

```
# TODO: Add "hugo-docs" package in the future
obj-aarch64-linux-gnu/bin/hugo -s docs -d public/html --uglyURLs
2016/01/12 00:33:24 function not implemented
debian/rules:9: recipe for target 'override_dh_auto_build' failed
```

Initial code search on "function not implemented" led me to `golang.org/x/sys/unix`, Errno 38 or 0x26, i.e., `ENOSYS`.

I was able to reproduce this on Debian arm64 developer machine asachi.debian.org, and it is not one single "function not implemented" errors, but random amount of it, from 9 errors to 94812 errors.  I'll need more time to investigate.

Note that this problem does **not** affect `armhf` and `armel` platforms.

anthonyfok:
After adding

```go
log.SetFlags(log.LstdFlags | log.Lshortfile)
```

to Hugo’s main.go (credit: http://stackoverflow.com/questions/24809287/how-do-you-get-a-go-golang-program-to-print-the-line-number-of-the-error-it-ju), the source of the error became more apparent:

```
2016/01/13 04:17:32 viper.go:260: error: function not implemented
2016/01/13 04:17:32 viper.go:260: error: function not implemented
2016/01/13 04:17:32 viper.go:260: error: function not implemented
2016/01/13 04:17:33 viper.go:260: error: function not implemented
2016/01/13 04:17:33 viper.go:260: error: function not implemented
2016/01/13 04:17:33 viper.go:260: error: function not implemented
2016/01/13 04:17:33 viper.go:260: error: function not implemented
2016/01/13 04:17:33 viper.go:260: error: function not implemented
0 draft content
0 future content
165 pages created
0 non-page files copied
0 paginator pages created
41 tags created
0 groups created
in 1100 ms
```

[...]

PR #1773, opened 2016-01-13
Do not call watchConfig() when not in watch mode

anthonyfok:
See #1772 

----

@spf13, does this look okay to you?  :-)

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -87,7 +87,8 @@ func isUserError(err error) bool {
 	return userErrorRegexp.MatchString(err.Error())
 }
 
-//HugoCmd is Hugo's root command. Every other command attached to HugoCmd is a child command to it.
+// HugoCmd is Hugo's root command.
+// Every other command attached to HugoCmd is a child command to it.
 var HugoCmd = &cobra.Command{
 	Use:   "hugo",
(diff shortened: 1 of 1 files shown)


anthonyfok:
(Note to self: Next time, test before pushing a commit for the first time.  :wink:)

Finally got my PR to work as expected.  My initial push erroneously used `viper.GetBool("watch")` which was never set and thus simply disabled watchConfig() indiscriminately... Oops!

I then discovered that the config watch only worked the first time when I do `echo "" >> docs/config.toml`, and then it got stuck, hence I added one more line to fix the problem.  From the commit log:

> Also, force DisableLiveReload to true when running "hugo --watch" (build-only non-server mode) to prevent livereload.ForceRefresh(), which would end up blocking watchConfig() forever, from being called because livereload.Initialize() is never called in this case.
>
> This fixes the bug where "hugo --watch" could only reload config.toml once before it gets stuck for good at livereload.ForceRefresh().
>
> This is also consistent with Hugo's existing behaviour: Non-server "hugo --watch" has never injected livereload.js since the inception of the "watch" feature in Hugo v0.12.

spf13:
This looks great. You've fixed two bugs in one commit. Merged as e8eb618

PR #1776, opened 2016-01-13
add Hypothesis as an alternative commenting system

fiatjaf:


spf13:
I'm a bit confused about this addition. As I understand it, hypothesis is 100% on the client meaning that each visitor would need to install it. Yes it does work with a Hugo site, but it would work with any site and since there is no action for the Hugo user to take it seems odd to mention it.

fiatjaf:
It works like Disqus, the site can include the Hypothesis script and the user will see Hypothesis and be able to comment on the page without having to manually install it.

fiatjaf:
You should need morereason to close this PR.

PR #1783, opened 2016-01-16
Replace placeholder values with real years

pra85:


anthonyfok:
Thank you for your good will, but this is not the place to do it.  Note how it is the Appendix of the original Apache License, and notice how it says:

> To apply the Apache License to your work, attach the following boilerplate notice, with the fields enclosed by brackets [] replaced with your own identifying information.

So, the boilerplate should remain a boilerplate.  And, as a matter of fact, we have already applied the Apache License to our work by using the boilerplate as a start and adding it to the beginning of all *.go files in the Hugo project.

Nevertheless, we value your continued contribution.  Hope you enjoy Hugo!  See you around!

PR #1788, opened 2016-01-20
Security: Disallow listing of absolute paths

hagbarddenstore:
Removes the possibility of listing all directories in system folders like
/etc, /usr, etc.

This is potentially dangerous when Hugo is used in shared environments, as
reported by @fransr.

### DIFF
--- a/tpl/template_resources.go
+++ b/tpl/template_resources.go
@@ -276,6 +276,10 @@ func ReadDir(path string) []os.FileInfo {
 		jww.ERROR.Printf("Path %s contains parent directory marker", path)
 		return nil
 	}
+	if filepath.IsAbs(path) {
+		jww.ERROR.Printf("Path %s is an absolute path", path)
+		return nil
+	}
 
 	p = filepath.Clean(path)
 	p = filepath.Join(wd, p)


spf13:
I'm trying to understand this PR a bit better. The patch is intended to remove the ability for a template function to list an absolute path as it could lead to exposing information.

This seems to me to be the wrong way to go about things. If the user has access to that data already then couldn't they find another way of exposing it? 

Doesn't every shared PHP host have the same issue? All of PHP can access and expose absolute paths.

What are we losing taking away the access to absolute paths for all of the Hugo users who aren't in a shared environment?

My preferred solution would be to remove access to that information in the first place rather than trying to plug each possible access point.

@hagbarddenstore @fransr please help me understand this better.

moorereason:
This PR outlaws absolute file paths in templates.  :-1:  And what prevents someone from using `../../../etc`?

IMO, this patch is unnecessary.  I don't see a need to complicate Hugo for this scenario.  If there's a security problem, it's with the hosting provider's environment, not Hugo.  If they want to use Hugo as their security layer, they should fork Hugo.

fransr:
Thanks for the comments, I'll join in here to explain the issue. I agree on some points you're raising, I actually discussed that with @hagbarddenstore before he sent in the PR just because of this.

So, the issue was identified not in a shared environment, but in an application that was using Hugo as a Wiki-template engine (which probably was a bad idea from the start, as @moorereason points out). This meant that all users having access to the wiki were able to edit the Hugo-templates, which in return made it expose server directories due to the absence of checking if the directory was absolute or not).

The argument and the single reason that the PR was sent in was the lines just above the added code:
https://github.com/spf13/hugo/pull/1788/files#diff-f92113c4eec98f7b7251a4054f295045R275

As you see, it's checking if the path contains ".." to prevent parent dirs, which is completely bypassed by using an absolute location. Is there any reason why those lines of code are present? Would be great to know the reason behind it to understand the difference between allowing absolute and restricting relative parent directories.

And to be fair, my guess is that the implementation we identified wasn't really a proper way of using Hugo.

moorereason:
Thanks for the comments, @fransr.  That context makes a little more sense, but I still contend that it's the environment's responsibility to restrict access, not Hugo's.

spf13:
It's not a question of proper or not. It's open source, you can use it how
you please.

It's a question of modifying the code in a way that is restricting all
users to satisfy a very specific use. The great thing about open source is
that you can apply your own patch to your instance.

You do bring up a good point that we should review that entire function
more closely and ensure the behavior is as it should be. You are right that
it should be consistent.

PR #1799, opened 2016-01-30
Remove destination from config file docs

### LINKED ISSUE #1757: Remove "destination" from config file docs
The "Configuring Hugo" docs mistakenly list `destination` as a config file option.  The correct option is `publishdir`. It affects 0.15 and HEAD.  See [“destination” configuration parameter doesn’t work when cli argument does](https://discuss.gohugo.io/t/destination-configuration-parameter-doesnt-work-when-cli-argument-does/2394) discussion thread.

This issue only adds to the confusion from the disparity between the CLI flag and the config file.  The CLI uses `--destination`.  We should consider changing the flag to match the config file.

moorereason:
Applies to 0.15 and HEAD.

Fixes #1757

### DIFF
--- a/docs/content/overview/configuration.md
+++ b/docs/content/overview/configuration.md
@@ -89,8 +89,6 @@ Following is a list of Hugo-defined variables that you can configure and their c
     dataDir:                    "data"
     defaultExtension:           "html"
     defaultLayout:              "post"
-    # filesystem path to write files to
-    destination:                ""
     disableLiveReload:          false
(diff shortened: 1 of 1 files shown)


moorereason:
Cherry-picked fixes: https://github.com/spf13/hugo/commit/9b8d2ce52bd30ac891a537caaa9207f5105362b8, https://github.com/spf13/hugo/commit/4a76b2ec11c96e3060044acde04d621d764f3b70

Issue #1815, opened 2016-02-04
Access url slug in templates


christophermancini:
Is it possible to access the slug for a page without explicitly defining the slug in the frontmatter? I have been digging through the code but cannot quite tell for sure. It appears it can't be.

moorereason:
Try `.URLPath.Slug`

christophermancini:
Nope, still only accessible if the slug is defined in front matter. :(

christophermancini:
So, the closest I could get is `.Source.LogicalName` but then I need to chop off the `.md`. I may submit a PR to set the slug if one is not provided.

bep:
As slug is user-provided and there is logic to use it if set, It doesn't make sense to put something in it if ... not set.

What is your use case?

christophermancini:
http://rainbowsixbootcamp.com is built on Hugo. I created a page for each operator, weapon, weapon mod, etc. When I am trying to generate a list of related pages, I want to include an image and I don't want to have to code the path to the image for each file. So, I want to link to the image matching the content files slug.

E.g. The weapon skins, I have an icon file showing the skin texture, a 320x180 image and a 1600x900 image. There are ~140 skins and more created with each DLC drop. If the skin is called `[Acid Rain](http://rainbowsixbootcamp.com/advanced-training/gear/weapon-mod/acid-rain/)`, then my images are:

- http://rainbowsixbootcamp.com/images/gear/weapon-mod/skin/acid-rain-th.jpg
- http://rainbowsixbootcamp.com/images/gear/weapon-mod/skin/acid-rain-320x180.jpg
- http://rainbowsixbootcamp.com/images/gear/weapon-mod/skin/acid-rain-1600x900.jpg

If I have access to the slug without having to explicitly define it every time, then things are way easier.

bep:
> .Source.LogicalName but then I need to chop off the .md. I may submit a PR to set the slug if one is not provided.

.Source.BaseFileName is the LogicalName without the suffix.

christophermancini:
@bep Yea, I figured that out. Thanks.

robertbasic:
@christophermancini is this still an issue?

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

[...]

PR #1817, opened 2016-02-04
add accessor for the slug used for the page

christophermancini:
As mentioned in #1815 I want to be able to access the generated slug to simplify management of a pages digital assets. So I created a simple method to access either the user defined slug or the base filename since it is used for the slug otherwise. I guess the only potential issue with this, is that if the site uses ugly url's (with the .html), then this will exclude the .html from the "slug". Not necessarily sure this is a bad thing. 

Why I think their should be an accessor to the slug versus directly accessing .Source.BaseFileName from the template is because you need to be able to handle for when a slug is user defined.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -459,6 +459,14 @@ func (p *Page) Extension() string {
 	return viper.GetString("DefaultExtension")
 }
 
+// returns frontmatter slug or generates from filename
+func (p *Page) GetSlug() string {
+	if len(p.Slug) > 0 {
+		return p.Slug
+	}
+	return p.Source.BaseFileName()
+}
+
 func (p *Page) LinkTitle() string {
 	if len(p.linkTitle) > 0 {
 		return p.linkTitle


moorereason:
We already have `.Slug`.

moorereason:
I see what you're trying to do, but this PR feels like a hack.  :disappointed: 

I'd rather see us just set a default slug for each page based on the LogicalName.  If you define a slug in your front matter, we overwrite the default.  The main downside I can see is that such a change would probably break people's custom permalink `:slug` usages, since `:slug` currently uses the title if no slug is defined.

What do you think @bep?

bep:
This PR is a no-go. The Slug is defined as something defined by the user in front matter... and this is a getter that isn't really a getter. 

If something is added it should be named something else ... but we have WAY TOO many related methods already, and this is easily solved with an if/else with/else construction in the template.

christophermancini:
I realized after submitting this that it might not be the best approach because if the method of generating the slug should one not be provided change, this would be out of sync.

I am fine with coding it deeper into the system. The thing is, a slug is a part of every page / node in the site whether it is user defined or generated. By omitting it from the Page object if generated, you are requiring the template creator to know how the slug is generated which I feel provides a poor user experience.

If you will approve a PR that is completed with proper integration, I will do the work. Let me know your thoughts. @bep @moorereason

bep:
I'm closing this, please open another issue for whatever something else.

Issue #1654, opened 2015-12-01
Adding more documentation about Google Analytics template


digitalcraftsman:
I've recognized that the docs are missing an example of the built-in [Google Analytics template](https://github.com/spf13/hugo/pull/1505) **and** of the corresponding variable for the config file.

What would be a good place to insert them into the docs? Furthermore, the docs provide an [example](http://gohugo.io/templates/partials/) of a footer partial taken from the code of spf13.com.

Should I remove the GA part of this example?

/cc @anthonyfok @bep

cfjedimaster:
Dumb question - but where exactly *is* this documented? I've been looking for a while now and can't find it.

cfjedimaster:
Site variables (https://gohugo.io/templates/variables/), mentions .Site.GoogleAnalytics and says this:

"A string representing your tracking code for Google Analytics as defined in the site configuration."

Ok, so I go to configuration (https://gohugo.io/overview/configuration/) but I see nothing there.

digitalcraftsman:
Hello @cfjedimaster,

I've integrated this feature shortly before the release of Hugo v0.15. But this didn't included the documentation. Since a while I'm waiting for feedback to add it.

However, you can enable the internal Google Analytics template by adding the following line to your theme:

    {{ template "_internal/google_analytics.html" . }}

Now, move to the config file and assign your tracking code to the corresponding variable:

    googleAnalytics = ""

The template behaves similar to the one of Disqus. If `googleAnalytics` is not set or equals an empty string the internal template will not appear in your generated sites.

cfjedimaster:
Thanks. Should the bug be reopened though?

digitalcraftsman:
Which bug do you mean?

cfjedimaster:
Well, it says, "I've recognized that the docs are missing an example of the built-in Google Analytics template.", and I guess I saw this as related, but maybe a new bug should be filed for the variable?

digitalcraftsman:
Now, I understand. I've updated the first comment and mentioned the config variable explicitly.

> but maybe a new bug should be filed for the variable

Both things are closely related. Therefore, I would keep them together in a single issue.

cfjedimaster:
Cool, and thank you for the help.

On Fri, Jan 15, 2016 at 3:17 PM, digitalcraftsman <notifications@github.com>
wrote:

> Now, I understand. I've updated the first comment and mentioned the config
> variable explicitly.
>
> but maybe a new bug should be filed for the variable
>
[...]

PR #1829, opened 2016-02-06
Adds documentation for Google Analytics internal template

dherbst:
Adds googleanalytics tracking id line into the configuration documentation.

Adds a new page in extras, I named it "Analytics" in case there are additional analytics templates added in the future, they can be added to this page.  I added weight 15 so it would show up under Aliases and above Builders.

See #1654

### DIFF
--- a/docs/content/extras/aliases.md
+++ b/docs/content/extras/aliases.md
@@ -8,7 +8,7 @@ date: 2013-07-09
 menu:
   main:
     parent: extras
-next: /extras/builders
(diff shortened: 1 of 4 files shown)


digitalcraftsman:
Thanks for your effort to document this feature. Merged as 06e90b1237cc29fa17be96d13bf9d0428e65e2a8 (master) and 5abb8bb3a6cd1204d17bdb76f82967fcd5919522 (v0.15.docs)

PR #1831, opened 2016-02-07
adding shelan.org to showcase page

shelan:
Adding shelan.org blog space to showcase page.


digitalcraftsman:
Hello @shelan,

thank you for adding your website to the showcase of Hugo. I've merged your commit as d1b0290fef8be78a2eb9c8497e0312ccfc4adc82 (master) and 145be4ca38cd0f9ba24dd27edef704c1a4dfc98a (v0.15.docs).

Within the next minutes you should see your website in the showcase.

anthonyfok:
Hi @digitalcraftsman,

Thank you for handling the merging of Showcase-related pull requests!  However, please double check to see the submitted thumbnail images in the pull requests to Showcase is sized at 600px × 400px, like all the other existing `*-tn.png` or `*-tn.jpg` files.  If not, please ask the submitter to fix it before merging.  (Sorry, this requirement does not seem to be documented.)

At my current location with not a not-too-fast Internet connection, I can see the original shelan-tn.png taking 5 to 10 times longer to load than other thumbnails.  Also, due to its incorrect aspect ratio, the base of the image is not aligned, with the other two thumbnails on the same row, and will eventually create problems for the next row of thumbnails.

Also, very often, we will need to optimize the submitted thumbnail image.  Please see what I did in commit 8681a35.  I usually use pngquant to change the submitted full-color (24-bit?  32-bit?) PNG image to an indexed-color image.  At times, for images with lots of colors, I would convert it to JPEG, depending on the resulting visual quality and the resulting filesize.

In the case of shelan-tn.png, the original file was reduced from 334125 bytes to 26929 bytes.

Yes, the Showcase is growing rather quickly, and that is why we need to keep an eye on the thumbnail filesize, otherwise the Showcase page will soon become painfully slow to load.

Many thanks for your help!  :-)

Anthony

shelan:
Hi @anthonyfok 

I strongly agree with your suggestions. It was my mistake that I have added the unmodified version of the  thumbnail. I have created a new PR with a reduced and corrected thumbnail.

Thank a lot for pointing it out.

Best Regards,

Shelan

digitalcraftsman:
Hi @anthonyfok,

it makes indeed sense to compress / optimize the images. Thank you for sharing your workflow and tools for the compression. I took a note for the next time :smile: 

I think we should add a note about thumbnail requirements [here](http://gohugo.io/community/contributing/) and reference it at the bottom of the showcase page.

While editing the contribution page in docs: you added a few a days the file `CONTRIBUTING.md` in the root of the project (with bd0176a5487b0d8c368a14b3f283b9029f51d619). Would it make sense to include the contribution guidelines into the documentation?

anthonyfok:
Hi @shelan, worry not, it was my mistake that I had procrastinated in putting up the necessary documentation in the right place.

@digitalcraftsman, you are awesome!  I love your addition of Showcase submission guideline to the `CONTRIBUTING.md` file!  Thank you for taking the time and effort to write all that up!

digitalcraftsman:
@anthonyfok it would be great if we only need to compress the thumbnails in the long run before merging additions to the showcase.

PR #1833, opened 2016-02-09
Command Package test coverage increase

Skarlso:
Hi! 

I started to increase coverage in command package. Hugo.go's current coverage is at 49.5%. 

I would like to request a little review before I continue on to cover more parts of hugo.go which will be the other 40% or so. 

Thanks,
Gergely.


Skarlso:
Interesting. I did squash my commits, but it created a pull out of 9 commits instead of the squashed one.

commands/hugo_test.go
```diff
@@ -0,0 +1,317 @@
...
+	if actualUserError != expectedUserError {
+		t.Errorf("Actual (%v) returned command did not equal expected (%v) command.", actualUserError, expectedUserError)
+	}
+}
+
+func TestSystemErrorCommandNewFormattedSystemError(t *testing.T) {
+	expectedSystemError := commandError{"test:12", false}
+	actualSystemError := newSystemErrorF("test:%d", 12)
+
+	if actualSystemError != expectedSystemError {
```
Skarlso: I will re-write these to use assert.

PR #1823, opened 2016-02-05
Added content_test.go to increase test coverage.

Skarlso:
Proper pull request. For all the conversation please see pull request #1812.

PR #1812, opened 2016-02-04
Increased coverage on Helpers: content.go.

Skarlso:
Also, some minor tweaks to content.go. Removed two unused functions and did a little formatting here and there.

PR #1834, opened 2016-02-09
Increasing coverage for Command Package

Skarlso:
Begun to increase coverage in the command package. Starting with
hugo.go.
- Added hugo_test.go and started to increase coverage for Hugo command
- Two minor tweaks to hugo.go in comment and conditional structure
- Current coverage of hugo.go is at 40.1%
- Most of the configuration initialization has been covered sofar
- Larger parts uncovered remaining are: NewWatcher, getDirList and copyStatic

### DIFF
--- a/commands/hugo_test.go
+++ b/commands/hugo_test.go
@@ -1,0 +1,402 @@
+package commands
+
+import (
+	"fmt"
+	"io/ioutil"
+	"math"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/spf13/hugo/helpers"
+	"github.com/spf13/hugo/hugofs"
+	"github.com/spf13/hugo/hugolib"
+	jww "github.com/spf13/jwalterweatherman"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+)
+
+func clearConfigFile() {
+	viper.SetConfigFile("")
+	theme = ""
+	cfgFile = ""
+	destination = ""
+	source = ""
+	baseURL = ""
+}
+
+func TestUserErrorCommandNewUserError(t *testing.T) {
+	defer viper.Reset()
+	expectedUserError := commandError{"test\n", true}
+	actualUserError := newUserError("test")
+	assert.Equal(t, expectedUserError, actualUserError)
+}
+
+func TestSystemErrorCommandNewSystemError(t *testing.T) {
+	defer viper.Reset()
+	expectedSystemError := commandError{"test\n", false}
+	actualSystemError := newSystemError("test")
+	assert.Equal(t, expectedSystemError, actualSystemError)
+}
+
+func TestUserErrorCommandNewFormattedUserError(t *testing.T) {
+	defer viper.Reset()
+	expectedUserError := commandError{"test:%d 12\n", true}
+	actualUserError := newUserError("test:%d", 12)
+	assert.Equal(t, expectedUserError, actualUserError)
+}
+
+func TestSystemErrorCommandNewFormattedSystemError(t *testing.T) {
+	defer viper.Reset()
+	expectedSystemError := commandError{"test:12", false}
+	actualSystemError := newSystemErrorF("test:%d", 12)
+	assert.Equal(t, expectedSystemError, actualSystemError)
+	if actualSystemError != expectedSystemError {
+		t.Errorf("Actual (%v) returned command did not equal expected (%v) command.", actualSystemError, expectedSystemError)
+	}
+}
+
+func TestUserErrorReturnError(t *testing.T) {
+	defer viper.Reset()
+	cError := commandError{"test error", false}
+	expectedError := "test error"
+	actualError := cError.Error()
+	assert.Equal(t, expectedError, actualError)
+}
+
+func TestIsUserError(t *testing.T) {
+	defer viper.Reset()
+	var errors = []struct {
+		err error
+	}{
+		{fmt.Errorf("argument")},
+		{fmt.Errorf("flag")},
+		{fmt.Errorf("shorthand")},
+		{commandError{"test", true}},
+	}
+
+	for _, v := range errors {
+		if !isUserError(v.err) {
+			t.Errorf("Error (%v) should have been flagged as a user error.", v)
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


commands/hugo_test.go
```diff
@@ -0,0 +1,305 @@
...
+	file, err := hugofs.SourceFs.Create(configPath)
+	if err != nil {
+		t.Error(err.Error())
+	}
+	err = hugofs.SourceFs.Chmod(configPath, os.ModeTemporary|os.ModePerm)
+	if err != nil {
+		t.Error(err.Error())
+	}
+	basicConfig := `
+        contentdir = "` + string(paths[2]) + `"
```
Skarlso: I need to escape path here so that it doesn't cause any problem on a non-unix system.

Skarlso:
Damn.... This did not go so well.... I created random folders now so they didn't collide. But that only caused chaos apparently.

Skarlso:
Alright. Testing these would lead to race condition so I will test individual functions and operations, rather than the Watcher. The Watcher is basically two endless goroutines.

Skarlso:
I removed the watcher test. It's simply not possible to test the watch since it's basically and endless blocking routine. Anything I wrote or come up with, ends up being a race condition for viper.

I will rather try to test other parts of the watcher's functionality separatly.

commands/hugo_test.go
```diff
@@ -0,0 +1,314 @@
...
+	viper.SetConfigFile("")
+	cfgFile = ""
+	destination = ""
+	source = ""
+	baseURL = ""
+}
+
+func TestUserErrorCommandNewUserError(t *testing.T) {
+	viper.Reset()
+	defer viper.Reset()
```
bep: The viper.Reset()s aren't needed here.

commands/hugo_test.go
```diff
@@ -0,0 +1,314 @@
...
+	expectedError := "test error"
+	actualError := cError.Error()
+	assert.Equal(t, expectedError, actualError)
+}
+
+func TestIsUserError(t *testing.T) {
+	viper.Reset()
+	defer viper.Reset()
+	var errors = []struct {
+		err error
```
bep: I would add expected as a bool to this struct, then add that to the list below. Then you can add tests for errors that should not be flagged as user errors. Also include userError and serverError in the test table.

commands/hugo_test.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func TestInitializeConfigSetsWorkingDirToSourceIfSet(t *testing.T) {
+	viper.Reset()
+	defer viper.Reset()
+	defer hugoCmdV.ResetCommands()
+	defer hugoCmdV.ResetFlags()
+	defer clearConfigFile()
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	blogSuffic := strconv.Itoa(r.Intn(math.MaxInt8))
```
bep: suffix?

Skarlso: Yes. :-)

commands/hugo_test.go
Skarlso: Thanks @bep. I'm in Boston right now, so sorry for the stale PR. I'll be finishing it once I'm home.

Skarlso:
Back to business. I will update this ASAP.

Skarlso:
@bep Hi there. Sorry, for taking so long. Got bogged down with life.

So, I dug in again for a few hours. I was trying to write tests for the go routines and the NewWatcher stuff, but it gets pretty insane. I'd even go so far as to say, that it's not worth it. 

I'd say this is as far as it gets without becoming very messy an hard to maintain which is not in the spirit of this project. I would like to continue so, If you have some suggestions, I would very gladly hear them!

Otherwise, I'd think it's ready, apart maybe from some smaller fishes, some if conditions here and there which could still sort of be covered. What say you?
Thanks.

commands/hugo_test.go
```diff
@@ -0,0 +1,356 @@
...
+		{undraftCmd},
+		{importCmd},
+		{genCmd},
+	}
+
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	blogSuffix := strconv.Itoa(r.Intn(math.MaxInt8))
+	basepath := filepath.Join(os.TempDir(), "blog"+blogSuffix)
+	configPath := filepath.Join(basepath, "config.toml")
+	hugofs.Source().MkdirAll(basepath, os.ModePerm|os.ModeDir)
```
moorereason: Your tests are creating a lot of files on the local file system that are not cleaned up (`/tmp/blog*`).  Look at using `SetSource()` and `SetDestination()` from `hugofs` to use `afero.MemMapFs`.

Skarlso: Good stuff man! Thanks for the tip. I'm looking into that.

Skarlso: @moorereason actually, I remember trying that out. But it didn't work. However, I would like to think that I know this stuff a bit better now. :) I'll try again.

Skarlso:
The leaking is unbelievable though. My tests are NOT happy if I don't have these lines in front of all my tests.

```go
	defer viper.Reset()
	defer clearConfigFile()
	defer helpers.ResetConfigProvider()
	defer hugofs.InitDefaultFs()
	defer hugoCmdV.ResetCommands()
	defer hugoCmdV.ResetFlags()
```

That's pretty heavy. :)

Skarlso:
@bep and @moorereason. I removed much of the code, and added to use MemFs where possible. The theme I could only test with valid path, since that requires the theme directory to actually exist. But it works, and I'm testing version mismatch as well. I can still add in a couple of cases actually. Coverage is up to 43.4%. And I can make it still go up a bit. 

Also, I realised that I had some false positives back in the old days. Now those are actual positives.

Skarlso:
@bep -- You know what. I was going to let this go. But no. You are the reason I didn't come back for a long long time, and the fact that I switched countries. You are not a very nice person. You are arrogant, rude, and completely frustrating to talk to. So you know what? You don't have to put up with me any longer. I do not like working with you. And you don't like working with me either probably. Hence. Bye.

bep:
@Skarlso - lets make this clear

1) You don't know me, we never met
2) I have revisited the communication on your first contributions back in February, and I was nothing but polite and helpful, see #1812 and #1823
3) My main point about my remarks about the revival of this PR here and on the forum is this: You are now not a fresh Hugo contributor anymore, you are an experienced developer, and I expect you to try a little bit harder before you shout out for help. We are busy, I prefer to spend most of my time on what I love: Code. If you think that makes me "a not very nice person", that is totally up to you.

PR #1835, opened 2016-02-09
Fix git commit blog article link.

mortonfox:
The trailing forward slash produces a 404 error instead of redirecting to the correct URL.


digitalcraftsman:
Hello @mortonfox,

thank for very much for the correction of the link in the README :+1: 

Best regards,
Digitalcraftsman

PR #1838, opened 2016-02-11
Make the two common Taxonomy templates explicit.

ObviousDesign:
Hi there,
https://gohugo.io/taxonomies/overview/ introduces the definitions 'Term' as a Key within that taxonomy.

The filename Hugo accepts for Taxonomy terms can be confusing:
SINGULAR**.terms.**html -> Renders at /SINGULAR/
SINGULAR.html -> Renders at /SINGULAR**/TERM/**

Ideally the filenames Hugo accepts for templates are swapped:
SINGULAR.html -> Renders at /SINGULAR/
SINGULAR**.terms.**html -> Renders at /SINGULAR**/TERM/**

PR #1839, opened 2016-02-11
Make the two common Taxonomy templates explicit #2

ObviousDesign:
Included explanation for Taxonomy lists, next to Taxonomy term templates. Changed the title of the document to cover both.


anthonyfok:
Note: This PR is related to PR #1838 but not a duplicate despite the same commit message title.

PR #1849, opened 2016-02-13
Remove stray redis docs

derekperkins:
Somehow the "Little Redis Book" in Chinese got committed into `hugolib` a year ago. :)


digitalcraftsman:
This file is currently used in  a [benchmark](https://github.com/spf13/hugo/blob/8509727fe862d06b3371651454fcb1af54b4ab21/hugolib/benchmark_test.go#L23). It was added with e67db666c8e4d9c62ef3edae3fdd523566eba91c by @noahcampbell.

derekperkins:
Ok, that makes more sense. I did run tests and benchmarks and nothing failed.

Issue #448, opened 2014-08-29
Add ability to generate per-{year, month, day} archives


yacoob:
I'm in process of migrating a blogger based blog to hugo. Compare [blogger site](http://gcu.tactical-grace.net) to [hugo-backed site](http://ucg.tactical-grace.net/) (source [here](http://github.com/yacoob/gcu-hugo)).

One thing that I'll lose as a result of this migration is an archive of posts (like [this one](http://gcu.tactical-grace.net/search?updated-max=2014-05-04T17:47:00%2B01:00&max-results=4)). I don't really care all that much, and I'll most likely replace it with a list of _all_ posts, with a different layout, generated from `layouts/section/post.html`. I'd like to ask you, however, to consider adding a capability for hugo to generate time-based archives. Such archives would most likely be useful to people who _want_ to preserve their existing archive pages.

It'd generate content under `public/<section>/2014/index.html` for example, from `layouts/post/list.html`. It'd need to allow per-section and per-level (yearly, monthly, daily) control. It'd also need to play well with `permalinks` parameter from config.yaml - I don't really have a good idea how to address that.

Is such feature request a sensible one? :)

matrixik:
Just fox example from Wordpress:
https://wordpress.org/plugins/compact-archives/
http://www.wpbeginner.com/archives/

AriaFallah:
Any progress on this or workarounds?

vielmetti:
Any ideas on this one?

yacoob:
FWIW, I've stopped using Hugo for the site that I wanted this feature for, so I'm muting this bug - don't expect any answers here :>

mcliment:
I'm not a Hugo expert and my templates may have some newbie errors but I managed to make archives work using taxonomies.

Check out this repo: https://github.com/mcliment/f1blog-archived/ -there are taxonomies per year defined in `config.toml` and in every post there's the corresponding frontmatter information-.

The resulting site is here: http://f1blog.climens.net/ with most of the things I had in Wordpress.

smakinson:
Did this one get dropped for good?

[...]

PR #1850, opened 2016-02-13
Standardize author data

derekperkins:
Here's the first attempt at standardizing Hugo author details.
1. Data lives in `/data/authors` with one file per author. I thought about nesting it in `/data/site/authors` or `/data/hugo/authors`, but those seemed unnecessarily hidden.
2. Author front matter is an array to support multiple authors, though it will be up to themes to support multiple authors or not. There is a convenience `.Author` method to return a single author. 
   _if there are multiple authors and this is used, there is no guarantee which author will be returned_
3. I changed `.Site.Author` from `map[string]interface{}` to `Author`, though I don't know if that is in use outside of the embedded RSS feed
4. Social data should only be the username
5. For Author social details, I added a `.URL()` convenience method that returns canonical urls for major social networks
6. I added a params section to the author, though as of right now it is `map[string]string` instead of `map[string]interface{}`
7. I use snake_case in the data files but templates will access them with PascalCase
8. Here's a sample author file

```
given_name      = "Derek"
family_name     = "Perkins"
display_name    = "Derek Perkins"
thumbnail       = "img link"
image           = "img link"
short_bio       = "super short bio"
bio             = "Here's Derek's awesome bio"
email           = "derek@email.com"

[social]
    facebook    = "derekperkins"
    github      = "derekperkins"
    twitter     = "derek_perkins"
    googleplus  = "DerekPerkins1"

[params]
   random       = "whatever you want"
```

### DIFF
--- a/hugolib/author.go
+++ b/hugolib/author.go
@@ -13,6 +13,14 @@
 
 package hugolib
 
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cast"
+)
+
 // AuthorList is a list of all authors and their metadata.
 type AuthorList map[string]Author
 
@@ -43,3 +51,131 @@ type Author struct {
 // - linkedin
 // - skype
 type AuthorSocial map[string]string
+
+// URL is a convenience function that provides the correct canonical URL
+// for a specific social network given a username. If an unsupported network
+// is requested, only the username is returned
+func (as AuthorSocial) URL(key string) string {
+	switch key {
+	case "github":
+		return fmt.Sprintf("https://github.com/%s", as[key])
+	case "facebook":
+		return fmt.Sprintf("https://www.facebook.com/%s", as[key])
+	case "twitter":
+		return fmt.Sprintf("https://twitter.com/%s", as[key])
+	case "googleplus":
+		isNumeric := onlyNumbersRegExp.Match([]byte(as[key]))
+		if isNumeric {
+			return fmt.Sprintf("https://plus.google.com/%s", as[key])
+		}
+		return fmt.Sprintf("https://plus.google.com/+%s", as[key])
+	case "pinterest":
+		return fmt.Sprintf("https://www.pinterest.com/%s/", as[key])
+	case "instagram":
+		return fmt.Sprintf("https://www.instagram.com/%s/", as[key])
+	case "youtube":
+		return fmt.Sprintf("https://www.youtube.com/user/%s", as[key])
+	case "linkedin":
+		return fmt.Sprintf("https://www.linkedin.com/in/%s", as[key])
+	default:
+		return as[key]
+	}
+}
+
+func mapToAuthors(m map[string]interface{}) Authors {
+	authors := make(Authors, 0, len(m))
+	for authorID, data := range m {
+		authorMap, ok := data.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		a := mapToAuthor(authorID, authorMap)
+		if a.ID != "" {
+			authors = append(authors, a)
+		}
+	}
+	sort.Stable(authors)
+	return authors
+}
+
+func mapToAuthor(id string, m map[string]interface{}) Author {
+	if id == "" {
+		return Author{}
+	}
+
+	author := Author{ID: id}
+	for k, data := range m {
+		switch k {
+		case "givenName", "firstName":
+			author.GivenName = cast.ToString(data)
+			author.FirstName = author.GivenName
+		case "familyName", "lastName":
+			author.FamilyName = cast.ToString(data)
+			author.LastName = author.FamilyName
+		case "displayName":
+			author.DisplayName = cast.ToString(data)
+		case "thumbnail":
+			author.Thumbnail = cast.ToString(data)
+		case "image":
+			author.Image = cast.ToString(data)
+		case "shortBio":
+			author.ShortBio = cast.ToString(data)
+		case "bio":
+			author.Bio = cast.ToString(data)
+		case "email":
+			author.Email = cast.ToString(data)
+		case "weight":
+			author.Weight = cast.ToInt(data)
+		case "social":
+			author.Social = normalizeSocial(cast.ToStringMapString(data))
+		case "params":
+			author.Params = cast.ToStringMapString(data)
+		}
+	}
+
+	// set a reasonable default for DisplayName
+	if author.DisplayName == "" {
+		author.DisplayName = author.GivenName + " " + author.FamilyName
+	}
+
+	return author
+}
+
+// normalizeSocial makes a naive attempt to normalize social media usernames
+// and strips out extraneous characters or url info
+func normalizeSocial(m map[string]string) map[string]string {
+	for network, username := range m {
+		if !isSupportedSocialNetwork(network) {
+			continue
+		}
+
+		username = strings.TrimSpace(username)
+		username = strings.TrimSuffix(username, "/")
+		strs := strings.Split(username, "/")
+		username = strs[len(strs)-1]
+		username = strings.TrimPrefix(username, "@")
+		username = strings.TrimPrefix(username, "+")
+		m[network] = username
+	}
+	return m
+}
+
+func isSupportedSocialNetwork(network string) bool {
+	switch network {
+	case
+		"github",
+		"facebook",
+		"twitter",
+		"googleplus",
+		"pinterest",
+		"instagram",
+		"youtube",
+		"linkedin":
+		return true
+	}
+	return false
+}
+
+func (a Authors) Len() int           { return len(a) }
+func (a Authors) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a Authors) Less(i, j int) bool { return a[i].Weight < a[j].Weight }
--- a/hugolib/author.go
+++ b/hugolib/author.go
@@ -15,14 +15,35 @@ package hugolib
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 	"strings"
 
 	"github.com/spf13/cast"
 )
 
-// AuthorList is a list of all authors and their metadata.
-type AuthorList map[string]Author
+var (
+	onlyNumbersRegExp = regexp.MustCompile("^[0-9]*$")
+)
+
+// Authors is a list of all authors and their metadata.
+type Authors []Author
+
+// Get returns an author from an ID
+func (a Authors) Get(id string) Author {
+	for _, author := range a {
+		if author.ID == id {
+			return author
+		}
+	}
+	return Author{}
+}
+
+// Sort sorts the authors by weight
+func (a Authors) Sort() Authors {
+	sort.Stable(a)
+	return a
+}
 
 // Author contains details about the author of a page.
 type Author struct {
@@ -43,7 +64,7 @@ type Author struct {
 	languages   map[string]Author
 }
 
-// AuthorSocial is a place to put social details per author. These are the
+// AuthorSocial is a place to put social usernames per author. These are the
 // standard keys that themes will expect to have available, but can be
 // expanded to any others on a per site basis
 // - website
--- a/hugolib/author.go
+++ b/hugolib/author.go
@@ -26,15 +26,21 @@ type AuthorList map[string]Author
 
 // Author contains details about the author of a page.
 type Author struct {
-	GivenName   string
-	FamilyName  string
-	DisplayName string
-	Thumbnail   string
-	Image       string
-	ShortBio    string
-	LongBio     string
-	Email       string
-	Social      AuthorSocial
+	ID          string
+	GivenName   string            // givenName OR firstName
+	FirstName   string            // alias for GivenName
+	FamilyName  string            // familyName OR lastName
+	LastName    string            // alias for FamilyName
+	DisplayName string            // displayName
+	Thumbnail   string            // thumbnail
+	Image       string            // image
+	ShortBio    string            // shortBio
+	Bio         string            // bio
+	Email       string            // email
+	Social      AuthorSocial      // social
+	Params      map[string]string // params
+	Weight      int
+	languages   map[string]Author
 }
 
 // AuthorSocial is a place to put social details per author. These are the
@@ -130,6 +136,18 @@ func mapToAuthor(id string, m map[string]interface{}) Author {
 			author.Social = normalizeSocial(cast.ToStringMapString(data))
 		case "params":
 			author.Params = cast.ToStringMapString(data)
+		case "languages":
+			if author.languages == nil {
+				author.languages = make(map[string]Author)
+			}
+			langAuthorMap := cast.ToStringMap(data)
+			for lang, m := range langAuthorMap {
+				authorMap, ok := m.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				author.languages[lang] = mapToAuthor(id, authorMap)
+			}
 		}
 	}
 
@@ -139,6 +157,68 @@ func mapToAuthor(id string, m map[string]interface{}) Author {
 	}
 
 	return author
+}
+
+// languageOverride returns an author with details overridden by the provided language
+func (a Authors) languageOverride(lang string) Authors {
+	langAuthors := make(Authors, 0, len(a))
+	for _, author := range a {
+		langAuthors = append(langAuthors, author.languageOverride(lang))
+	}
+	return langAuthors
+}
+
+// languageOverride returns an author with details overridden by the provided language
+func (a Author) languageOverride(lang string) Author {
+	if a.languages == nil {
+		return a
+	}
+
+	langAuthor, ok := a.languages[lang]
+	if !ok {
+		return a
+	}
+
+	if langAuthor.GivenName != "" {
+		a.GivenName = langAuthor.GivenName
+	}
+	if langAuthor.FirstName != "" {
+		a.FirstName = langAuthor.FirstName
+	}
+	if langAuthor.FamilyName != "" {
+		a.FamilyName = langAuthor.FamilyName
+	}
+	if langAuthor.LastName != "" {
+		a.LastName = langAuthor.LastName
+	}
+	if langAuthor.DisplayName != "" {
+		a.DisplayName = langAuthor.DisplayName
+	}
+	if langAuthor.Thumbnail != "" {
+		a.Thumbnail = langAuthor.Thumbnail
+	}
+	if langAuthor.Image != "" {
+		a.Image = langAuthor.Image
+	}
+	if langAuthor.ShortBio != "" {
+		a.ShortBio = langAuthor.ShortBio
+	}
+	if langAuthor.Bio != "" {
+		a.Bio = langAuthor.Bio
+	}
+	if langAuthor.Email != "" {
+		a.Email = langAuthor.Email
+	}
+
+	for k, v := range langAuthor.Social {
+		a.Social[k] = v
+	}
+
+	for k, v := range langAuthor.Params {
+		a.Params[k] = v
+	}
+
+	return a
 }
 
 // normalizeSocial makes a naive attempt to normalize social media usernames
--- a/hugolib/author.go
+++ b/hugolib/author.go
@@ -132,20 +132,20 @@ func mapToAuthor(id string, m map[string]interface{}) Author {
 
 	author := Author{ID: id}
 	for k, data := range m {
-		switch k {
-		case "givenName", "firstName":
+		switch strings.ToLower(k) {
+		case "givenname", "firstname":
 			author.GivenName = cast.ToString(data)
 			author.FirstName = author.GivenName
-		case "familyName", "lastName":
+		case "familyname", "lastname":
 			author.FamilyName = cast.ToString(data)
 			author.LastName = author.FamilyName
-		case "displayName":
+		case "displayname":
 			author.DisplayName = cast.ToString(data)
 		case "thumbnail":
 			author.Thumbnail = cast.ToString(data)
 		case "image":
 			author.Image = cast.ToString(data)
-		case "shortBio":
+		case "shortbio":
 			author.ShortBio = cast.ToString(data)
 		case "bio":
 			author.Bio = cast.ToString(data)
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -409,12 +409,10 @@ func traverse(keys []string, m map[string]interface{}) interface{} {
 
 func (p *Page) Author() Author {
 	authors := p.Authors()
-
-	for _, author := range authors {
-		return author
+	if len(authors) > 0 {
+		return authors[0].languageOverride(p.Node.lang)
 	}
-
-	return authors[0].languageOverride(p.Node.lang)
+	return Author{}
 }
 
 // Authors returns all listed authors for a page in the order they
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -429,9 +429,10 @@ func (p *Page) Authors() Authors {
 			return Authors{a.languageOverride(p.Node.lang)}
 		}
 	}
-	authors := authorKeys.([]string)
-	if len(authors) < 1 || len(p.Site.Authors) < 1 {
-		return AuthorList{}
+
+	authorIDs, ok := p.Params["authors"].([]string)
+	if !ok || len(authorIDs) == 0 || len(p.Site.Authors) == 0 {
+		return Authors{}
 	}
 
 	authors := make([]Author, 0, len(authorIDs))
@@ -445,7 +446,8 @@ func (p *Page) Authors() Authors {
 			authors = append(authors, a.languageOverride(p.Node.lang))
 		}
 	}
-	return al
+
+	return authors
 }
 
 func (p *Page) UniqueID() string {
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -416,21 +416,32 @@ func (p *Page) Author() Author {
 	return Author{}
 }
 
-func (p *Page) Authors() AuthorList {
-	authorKeys, ok := p.Params["authors"]
-	if !ok {
-		return AuthorList{}
+// Authors returns all listed authors for a page in the order they
+// are defined in the front matter. It first checks for a single author
+// since that it the most common use case, then checks for multiple authors.
+func (p *Page) Authors() Authors {
+	authorID, ok := p.Params["author"].(string)
+
+	if ok && authorID != "" {
+		a := p.Site.Authors.Get(authorID)
+		if a.ID == authorID {
+			return Authors{a}
+		}
 	}
 	authors := authorKeys.([]string)
 	if len(authors) < 1 || len(p.Site.Authors) < 1 {
 		return AuthorList{}
 	}
 
-	al := make(AuthorList)
-	for _, author := range authors {
-		a, ok := p.Site.Authors[author]
-		if ok {
-			al[author] = a
+	authors := make([]Author, 0, len(authorIDs))
+	for _, authorID := range authorIDs {
+		if authorID == "" {
+			continue
+		}
+
+		a := p.Site.Authors.Get(authorID)
+		if a.ID == authorID {
+			authors = append(authors, a)
 		}
 	}
 	return al
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -413,7 +413,8 @@ func (p *Page) Author() Author {
 	for _, author := range authors {
 		return author
 	}
-	return Author{}
+
+	return authors[0].languageOverride(p.Node.lang)
 }
 
 // Authors returns all listed authors for a page in the order they
@@ -425,7 +426,7 @@ func (p *Page) Authors() Authors {
 	if ok && authorID != "" {
 		a := p.Site.Authors.Get(authorID)
 		if a.ID == authorID {
-			return Authors{a}
+			return Authors{a.languageOverride(p.Node.lang)}
 		}
 	}
 	authors := authorKeys.([]string)
@@ -441,7 +442,7 @@ func (p *Page) Authors() Authors {
 
 		a := p.Site.Authors.Get(authorID)
 		if a.ID == authorID {
-			authors = append(authors, a)
+			authors = append(authors, a.languageOverride(p.Node.lang))
 		}
 	}
 	return al
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -410,7 +410,7 @@ func traverse(keys []string, m map[string]interface{}) interface{} {
 func (p *Page) Author() Author {
 	authors := p.Authors()
 	if len(authors) > 0 {
-		return authors[0].languageOverride(p.Node.lang)
+		return authors[0].languageOverride(p.lang)
 	}
 	return Author{}
 }
@@ -424,7 +424,7 @@ func (p *Page) Authors() Authors {
 	if ok && authorID != "" {
 		a := p.Site.Authors.Get(authorID)
 		if a.ID == authorID {
-			return Authors{a.languageOverride(p.Node.lang)}
+			return Authors{a.languageOverride(p.lang)}
 		}
 	}
 
@@ -441,7 +441,7 @@ func (p *Page) Authors() Authors {
 
 		a := p.Site.Authors.Get(authorID)
 		if a.ID == authorID {
-			authors = append(authors, a.languageOverride(p.Node.lang))
+			authors = append(authors, a.languageOverride(p.lang))
 		}
 	}
 
--- a/hugolib/author_test.go
+++ b/hugolib/author_test.go
@@ -222,7 +222,7 @@ func TestPageAuthor(t *testing.T) {
 				Params: map[string]interface{}{
 					"author": "derek",
 				},
-				Node: Node{Site: defaultAuthorSiteInfo},
+				Site: defaultAuthorSiteInfo,
 			},
 			authorDerek,
 		},
@@ -232,7 +232,7 @@ func TestPageAuthor(t *testing.T) {
 				Params: map[string]interface{}{
 					"author": "tanner",
 				},
-				Node: Node{Site: defaultAuthorSiteInfo},
+				Site: defaultAuthorSiteInfo,
 			},
 			authorTanner,
 		},
@@ -242,7 +242,7 @@ func TestPageAuthor(t *testing.T) {
 				Params: map[string]interface{}{
 					"author": "abc",
 				},
-				Node: Node{Site: defaultAuthorSiteInfo},
+				Site: defaultAuthorSiteInfo,
 			},
 			Author{},
 		},
@@ -250,7 +250,7 @@ func TestPageAuthor(t *testing.T) {
 			"no ID",
 			&Page{
 				Params: map[string]interface{}{},
-				Node:   Node{Site: defaultAuthorSiteInfo},
+				Site: defaultAuthorSiteInfo,
 			},
 			Author{},
 		},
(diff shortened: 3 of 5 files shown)


derekperkins:
- [ ] Decide how to implement author archives
- [ ] Add documentation
- [ ] Add a default user when creating a new hugo site
- [ ] Where it makes sense, alter themes.hugo.io to use it

digitalcraftsman:
@derekperkins I like the approach goes - convention over configuration. But I think it will take a bit until theme creators migrate this standard (and their users as well).

derekperkins:
I just changed the directory to `_authors` and now I delete the `.Site.Data` entry so the only way to access the author data is through `.Page.Author` or `.Page.Authors`

rdwatters:
@derekperkins This is good stuff. The best part is the "params," which means it's extensible, which makes me happy. Here is an uber nitpicky suggestion, so take it however you'd like. 

If conventions are the ultimate goal, how about naming parameters in a more => less abstracted style:

```
name_given:
name_family:
name_middle:
<!--Think it needs display name?-->
image:
image_thumb:
bio_short:
bio_long:
email:
[social]
[params]
<!--Rather than keeping social as a fixed key-value of the author content type, could this be included as part of "params" instead?-->
```

Also, does the `_` in `_authors` signify a collection rather than a content directory? 

Thanks for all the work on this. The Hugo community is fantastic.

spf13:
I think social should benefit from the same treatment as author as loads of
themes reference it and it would be great to have a standard. Social should
be both available for the site as well as using he same standard for the
author.

derekperkins:
> I think social should benefit from the same treatment as author

@spf13 Are you saying that you prefer using a fixed struct for social as opposed to a `map[string]string`?

derekperkins:
@rdwatters 
> Think it needs display name?

Yes, I think it needs a display name similar to WordPress. Not every site wants to use real names. You do bring up a good point though. Maybe we should make a function for themes to use that does a few fallbacks, checking first for DisplayName and then falling back to Given + Family.

> Does the _ in _authors signify a collection rather than a content directory?

I switched to an underscore because that's how `_default` works. Once I load up the actual author struct, I delete `.Site.Data["_authors"]` so it isn't accessible there anyways. I'm not super passionate about it, it just seemed to make sense.

> Also, w/r/t snake vs pascal, have there been any conversations as to casing conventions for all of Hugo?

I personally prefer Pascal over snake casing. I thought I remembered seeing somewhere that data files only used lowercase, but I couldn't find a reference. I'm more than happy to switch to Pascal casing.

derekperkins:
@rdwatters & @spf13 - For social, I could potentially do a transparent map. We have a struct for the most common networks, but make all of them accessible through a function. When reading the data in, I could store any unmatched networks in that map and return them as if they were natively supported. That allows for more extensibility while still retaining a definite standard.

spf13:
> @spf13 Are you saying that you prefer using a fixed struct for social as opposed to a map[string]string?

It doesn't need to be a struct necessarily, but I'd like to have some definition around what Social means and make it universal and standard. Part of this would be making sure that the keys are consistent across all places social is used. 

To look at your example: You've gone ahead and provided usernames/tokens for each of these values while others may put full URLs in them. Templates/Themes would really struggle with rendering these in a standard way. I'm not sure what the right answer is, but whatever it is it needs to be consistent so that Themes can just be interchangeable. 
Additionally you've written out the key "googleplus" where others may just have used "google".

Perhaps the ideal approach is to allow the definition to be loose but the parsing and internal access to be rigid. 

This would mean that "google" and "googleplus" would both match the social network "google". The ingestion could examine the input and parse it out into separate url and token values for each social network. 

I would think a struct like:

```
type socialNetwork struct {
	name         string
	aliases       []string
	token         string
        urlPattern  text.template
}

type socialEntry struct {
	network     socialNetwork
	token         string
}

```

Or something similar for each social network would be defined in the code. Then inside Site and Author you would have a map of [socialnames]SocialEntry.

Not sure if any of that made sense, but the idea is similar to how we handle params.

derekperkins:
> You've gone ahead and provided usernames/tokens for each of these values while others may put full URLs in them.

I'm currently parsing out usernames even if someone enters the full url or eliminating the `@` if they use `@username` for just that reason and added the `URL()` function for theme authors.

> This would mean that "google" and "googleplus" would both match the social network "google".

Do you feel strongly about that? I'd rather just tell people what to put in their config.

I can move forward with the structs you suggested. Having `socialNetwork` like you have here would make it trivial to extend the built in list of supported networks from a theme or config file if someone wanted to use a less common network.

spf13:
I don't feel strongly about that at all. As long as the documentation is clear, I'm fine with declaring the keys to use.

ryanwalls:
Any updates?  I think this would be a nice addition.

hugolib/author.go
```diff
@@ -13,23 +13,30 @@
...
-	GivenName   string
-	FamilyName  string
-	DisplayName string
-	Thumbnail   string
-	Image       string
-	ShortBio    string
-	LongBio     string
-	Email       string
-	Social      AuthorSocial
+	GivenName   string            // given_name
```
bep: Not sure what the comments are meant to be here, but if you suggest some kind of "XML or JSON" naming here, you might as well add JSON tags here so it gets used when people do `{{ author | jsonify }}`

hugolib/author.go
```diff
@@ -43,3 +50,84 @@ type Author struct {
...
 // - skype
 type AuthorSocial map[string]string
+
+// URL is a convenience function that provides the correct canonical URL
+// for a specific social network given a username. If an unsupported network
+// is requested, only the username is returned
+func (as AuthorSocial) URL(key string) string {
+	switch key {
+	case "github":
+		return "https://github.com/" + as[key]
```
bep: A map with the different social providers would look simpler and cleaner than a switch.

hugolib/author.go
```diff
@@ -43,3 +50,84 @@ type Author struct {
...
+	}
+	return authors
+}
+
+func mapToAuthor(m map[string]interface{}) Author {
+	author := Author{}
+	for k, data := range m {
+		switch k {
+		case "given_name":
+			author.GivenName = cast.ToString(data)
```
bep: Ah, now I see the "given_name" ... Should it not be "givenName"? I'm pretty sure camel casing is some kind of Hugo standard.

hugolib/author.go
bep: We should also add a way for end users to add missing/ovverrided social providers to that map.

hugolib/author.go
derekperkins: I can't remember for sure, but I think it was an issue with the data files all being lowercase.

hugolib/author.go
derekperkins: Agreed, but I'm not sure the best way to go about adding the override.

hugolib/author.go
bep: Nah, that should not be an issue here ... I could see a casing issue with the root entity (_authors), but I believe we should stick with camels.

hugolib/author.go
bep: Never mind -- that can be handled in a separate issue. But we should provide that.

hugolib/author.go
bep: To make it clear: the JSON or YAML or TOML name for GivenName would be givenName.

derekperkins: Do you like namespacing as `_authors`? The `_` prefix is used elsewhere for hugo core items, so I extended it to here.

bep: Yes, I think that will work. It clearly marks this as "reserved".

dencold:
Is this going to be part of the 0.17 release? I don't see any activity here for over a month, so wanted to offer my help if there is anything else remaining (documentation?) for this to be part of the next release of hugo. I've been doing some similar work on my personal theme to add better support for Author metadata, so I'd really like this to be standardized for general users of Hugo.

Let me know how I can help! Great addition @derekperkins!

derekperkins:
Ok, I just switched to camelCase and rebased on top of the current master.

@dencold / @ryanwalls: I don't have time to create documentation, so if you could flesh that out, I don't see any reason this shouldn't be merged for 0.17.

digitalcraftsman:
Because this pull request is already open for a while I began to refactor the `URL` method as bep suggested and added a few tests. I could also add documentation for the current state of the pull request.

Temporary, I created a [branch](https://github.com/digitalcraftsman/hugo/commits/authors) on my own fork for the additions.

Depending on the state of this pull request, I could create a temporary branch in this repo (like multilingual) in which this pr and my additins will be merged.

A few things are still undiscussed:

- the generation of an author archive was suggested. Are there any ideas to approach this? Maybe we  should also have an eye on #448.

- add a default user when creating a new hugo site. Should we just prompt the user to fill out basic information like name, bio and email? Similar to:

```
$ hugo new site example.com
Do you want to create an author profile? [y/n] y
What is your first name?
John
What is your last name?
Doe
... 
```

derekperkins:
I think URL should remain a switch. Different networks are going to have different ways to handle them, which a map can't handle. I don't think the switch is hard to read. 

I think this PR should get merged as is, and we can discuss further implementation details like you've brought up in their own separate discussions.

derekperkins:
@digitalcraftsman Thanks for working on tests. 

I love the idea for prompting author creation on a new site.

digitalcraftsman:
> I think URL should remain a switch. Different networks are going to have different ways to handle them, which a map can't handle. I don't think the switch is hard to read.

Let the others decide. I don't mind both approaches.

Regarding the prompt: we should create an internal template for an author file. The user can either ignore the generation of an author (if it makes no sense) or he can fill out the minimum of information needed. In the last case we insert the given data into the template and save it as an author file.

Furthermore, I would like to introduce an `hugo new author` command that shows basically the same prompt. That users don't have to copy a author template from the docs.

digitalcraftsman:
I added some documentation to my branch for this feature.

dencold:
I wanted to chime in to say I really like the direction of this PR. Glad we are getting close to a standard for Author support in Hugo, going to help theme authors & downstream users a lot. I had a couple of thoughts here.

0. `URL` method: I agree with @bep and @digitalcraftsman. Using a `map` to resolve the canonical path seems much more idiomatic Go than the current `switch` approach. @derekperkins, could you elaborate on how you could see a social network somehow not fitting this approach? Maybe we're missing something here.
0. Author prompting: I *love* this idea. However, I would be more in favor of the `hugo new author` approach here. The `hugo new site` command does not currently have any prompting whatsoever, and I don't think we should introduce it now. If we do, we should get signoff from @spf13 since this is a changed behavior.

I'm happy to contribute documentation and anything else to get this feature greenlit for the next release. It's a nice step forward. Thanks again everyone!

digitalcraftsman:
@dencold I agree that it's better to use only a completely seperate command for the profile creation instead of modifying `hugo new site`.

derekperkins:
@dencold Both of these are valid Google Plus links for my profile. If you use a map with a simple `fmt.Sprintf()`, you wouldn't be able to intelligently add/subtract the `+`.

https://plus.google.com/115741277991776525046
https://plus.google.com/+DerekPerkins1

I wouldn't want to force the user to pick one or the other and I wouldn't want to force the user to include the `+`, the same way I want it to work whether or not a Twitter user includes `@` as a prefix to their username. 

Maybe that logic gets moved to parse-time and this can be a map, but there needs to be a switch statement somewhere that allows for intelligent handling of different social network specifics.

derekperkins:
@digitalcraftsman / @bep: We're embarking on a site redesign, so this has bumped up my priority list. What's the minimum that needs to happen before this gets merged? Hoping we can get it in sometime in the next day or two.

bep:
If you squash the commitrs, i will merge tomorrow.

bep:
Oh and the above is GitHub on mobile while drinking...

derekperkins:
Thanks! Squash complete.

derekperkins:
@digitalcraftsman Since you've already taken a look at this, I just refactored a bit before it gets merged. Here are the change highlights.

1. I made front matter more flexible. It checks for `author` to contain a single author id, or it checks for `authors` to contain an array of author ids.
1. Author() now returns the first author listed instead of a random one from a map
1. I removed the deprecated `Author` field from `Site`
1. Added a regexp for the Google+ case I mentioned above

derekperkins:
And I aliased `GivenName` to `FirstName` and `FamilyName` to `LastName`, both when reading data in from the author file and when being referenced as a field.

dencold:
Just wanted to note that @digitalcraftsman did some **fantastic** work on fleshing out the documentation on his branch. I think we should try and get that pulled in as well.

derekperkins:
Yeah, his docs are awesome. Once the core author code is merged, then he can make a separate PR with his docs and tests.

hugolib/page.go
```diff
@@ -168,30 +168,55 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
...
-	authors := authorKeys.([]string)
-	if !ok || len(authors) < 1 || len(p.Site.Authors) < 1 {
-		return AuthorList{}
+// getAuthors first checks for a single author in the params,
+// then checks for multiple authors. []Author is returned to preserve
+// the input order so Author() is deterministic even when there
+// are multiple authors
+func (p *Page) getAuthors() []Author {
+	authorKey, ok := p.Params["author"].(string)
+	if ok {
```
bep: This doesn't look correct.

hugolib/page.go
```diff
@@ -168,30 +168,55 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
...
-	authors := p.Authors()
+	authors := p.getAuthors()
+	if len(authors) == 0 {
+		return Author{}
+	}
+	return authors[0]
+}
+
+// Authors returns all listed authors for a page
+func (p *Page) Authors() AuthorList {
```
bep: The AuthorList here is a map and not a list and it should be a sorted slice with a well defined sort key (weight?)

hugolib/page.go
```diff
@@ -168,30 +168,55 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
 	return p.Site.Params[keyStr], nil
 }
 
+// Author returns the first listed author for a page
 func (p *Page) Author() Author {
-	authors := p.Authors()
+	authors := p.getAuthors()
```
bep: See comment below: If we want to pick 1 author as THE author, we must define what that is. I think the first one in a well defined sorted list (see below) would make sense.

bep: Author and Authors should also be added to Node. We will eventually end up with one struct with front matter support for all page types, but in the first step we should just return the site authors for the Node version of these.

derekperkins:
@bep I just pushed changes based on your comments.
 - changed `page.Authors()` to return `[]Author` ordered by the front matter
 - `page.Author()` has two possible return options.
  1. first check for `author` in the params as a single string - this always takes precedence since most people don't need multiple authors per post
  1. return the first listed from params["authors"], defined as a slice of strings 
 - added `node.Author()` and `node.Authors()`, sorted by ID as defined by the filename
 - added sort methods for `.Site.Authors`, a map
  1. `ByID` 
  1. `ByIDDesc` 
  1. `ByGivenName` 
  1. `ByGivenNameDesc` 
  1. `ByFirstName` 
  1. `ByFirstNameDesc` 
  1. `ByFamilyName` 
  1. `ByFamilyNameDesc` 
  1. `ByLastName` 
  1. `ByLastNameDesc` 
  1. `ByDisplayName` 
  1. `ByDisplayNameDesc`

bep:
>My only concern about adding author functions to Node at this point is that once Nodes gain their own front matter, the implementation will change, breaking anyone who expects all .Site.Authors to be returned.

This will give the users an option to be more specific for, say, section `blog`; not a breaking change if they choose not to use it. I would say that if we decide to add that API later, it would be a bigger change ... And the IsNode, IsPage dance is extremely confusing for the users.

hugolib/author.go
```diff
@@ -13,23 +13,51 @@
...
+
+var (
+	onlyNumbersRegExp = regexp.MustCompile("^[0-9]*$")
+)
+
 // AuthorList is a list of all authors and their metadata.
 type AuthorList map[string]Author
 
+// Authors returns a slice of authors sorted by ID
+func (al AuthorList) Authors() []Author {
```
bep: * All the sorting stuff was kind of more and slightly different from what I suggested. I suggest we remove all the different sort methods for now (too many and too much boilerplate), 
* Add `Weight` to Author. This is in line with the reset of Hugo. Less magical than changing the ID from `bep` to `111bep` to make sure I'm the most important.
* Add an `Authors` type to it with the `Less` func etc. that sorts by weight, id, ...? some sensible fallbacks.


The tests can wait until we get the `Multilingual` feature in (will be much easier then).

derekperkins: I can remove some of the sorting methods, maybe the Desc ones, but I think Authors are fundamentally different than other sorting options in Hugo. A menu or a taxonomy doesn't really have a logical way to be ordered, but it's very logical to order a list of people sorted by their names. I fully expect people to add `/authors/` pages that list all the authors on the site.

I agree that we don't want people changing their IDs, so weight is a good option, but by that same token, people aren't going to want to go in and change the weight of every single person each time they add an author to make sure that names sort alphabetically.

bep: Sure, but if the default sorts first by Weight, then by Name or whatever. I asked you to remove the sort methods because 1) They are very verbose and not implemented the way sorting is done elsewhere (on several levels) and 2) They have no tests. I don't want to add code that I know I'm gonna have to rework.

So they can be added later, the most important part being a sensible sort for the Author() func.

derekperkins: I can go ahead and do that. How would you like me to implement a sort if not using the stdlib interface?

bep: The default sort on `Authors`, i.e. only one sort, can be implemented as methods on `Authors`. The additional sorts, which is now out of scope, could be implemented similar to the `pageSorter`. As to the `Desc` funcs, it is cleaner in my eye to just chain the methods, like `ByName().Reverse()`. Also, I believe we use `sort.Stable` all (?) other places.

hugolib/author.go
```diff
@@ -43,3 +71,189 @@ type Author struct {
...
+			author.Bio = cast.ToString(data)
+		case "email":
+			author.Email = cast.ToString(data)
+		case "social":
+			author.Social = normalizeSocial(cast.ToStringMapString(data))
+		case "params":
+			author.Params = cast.ToStringMapString(data)
+		}
+	}
+	return author
```
moorereason: Prior to the return, can we add this?
```go
if author.DisplayName == "" {
	author.DisplayName = author.GivenName + " " + author.FamilyName
}

Comment (digitalcraftsman) [inline hugolib/author.go]: We should assume that the users want's to use a custom display name instead of a hardcoded scheme. Furthermore, other cultures use different naming conventions. The Chinese [switch the given name and family name].

Comment (moorereason) [inline hugolib/author.go]: I understand that.  I'm just suggesting that we set a reasonable default, which Derek mentioned previously.

Comment (digitalcraftsman) [inline hugolib/author.go]: Ok, that's another story :wink:

**Comment by bep** [comment]
We have a 0.17 train leaving the station on 20th (macOs Sierra) and should get this one in ... status?

**Comment by derekperkins** [comment]
@bep Is that now closer to what you are wanting?

**Comment by dencold** [comment]
this would be fantastic to get pulled into 0.17. we should grab the work @digitalcraftsman did on the documentation as well!

**Comment by derekperkins** [comment]
@dencold He's got that in a separate PR #2419

**Comment by bep** [comment]
@derekperkins doing that sort every time someone is asking for author(s) is not optimal, even if it is a small data set. That, and the `Sort` method on `AuthorList` that has other side effects than sorting (returns a slice), looks a little clumsy to me.

Here is what I suggest:

* Remove the `AuthorList` type and replace it with `type Authors []Authors`
* Move the Swap, Less methods by weight to that type.
* Do the sorting once (in `mapToAuthorListmapToAuthorList`)
* Add a `Get` method to that type to get a single author by id.

**Comment by dencold** [comment]
@derekperkins yep, I know @digitalcraftsman has a separate PR, but it isn't tagged with the 0.17 milestone, so just want to make sure it's pulled in.

Let me know if there's anything I can do to help contribute here!

**Comment by derekperkins** [comment]
@bep Those were all very good changes.

**Comment by bep** [comment]
Add a sort to the Page authors + rebase this PR against master and we are good to go.

**Comment by derekperkins** [comment]
> Add a sort to the Page authors

@bep Not quite sure what you meant by that. I added `Authors.Sort()` back in, so you can chain `Sort` after calling `Page.Authors`.

These commits are now rebased on master, so once you're good with the changes, I'll squash the commits.

File: docs/content/templates/variables.md [hunk 9]
```diff
@@ -168,8 +168,13 @@ Also available is `.Site` which has the following:
 **.Site.Files** All of the source files of the site.<br>
 **.Site.Menus** All of the menus in the site.<br>
 **.Site.Title** A string representing the title of the site.<br>
+<<<<<<< HEAD
 **.Site.Author** A map of the authors as defined in the site configuration.<br>
```
bep: Merge issue above.

hugolib/author.go
```diff
@@ -13,23 +13,57 @@
...
+	for _, author := range a {
+		if author.ID == id {
+			return author
+		}
+	}
+	return Author{}
+}
+
+// Sort sorts the authors by weight
+func (a Authors) Sort() Authors {
```
bep: This isn't needed. The authors provided by node and page should already be sorted.

hugolib/page.go
```diff
@@ -182,33 +182,41 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
...
+	authors := make([]Author, 0, len(authorIDs))
+	for _, authorID := range authorIDs {
+		a := p.Site.Authors.Get(authorID)
+		if a.ID == authorID {
+			authors = append(authors, a)
 		}
 	}
-	return al
+
+	return authors
```
bep: Sort this before returning.

derekperkins: That is purposefully not sorted, as stated in the function comment. The default is to return the authors in the order listed.

bep: Ah, sorry -- I read it too fast. OK, if you can fix the "merge comments" and squash I'll merge.

derekperkins: The only use case for `Authors.Sort()` is to chain onto this function, if you wanted authors sorted by weight after getting them from the page. It doesn't seem like a common use case, so we just need to decide whether or not to leave the `Sort` function in.

derekperkins:
Before merging this, I wanted to take one last look at @spf13's suggestion (see below) to use a struct for social instead of just a simple map. Once we want to make this extensible, will we regret having users inputting just a simple map of social values? I'll be the first to admit that the current implementation is a little janky, stripping tags trying to get it to just a username so it can rebuild the correct canonical path later. My thought was to make that a little more robust once we look at making it pluggable, so I just want to make sure we aren't hamstringing ourselves.

Appreciate feedback from @moorereason, @digitalcraftsman, @rdwatters and @dencold.

>To look at your example: You've gone ahead and provided usernames/tokens for each of these values while others may put full URLs in them. Templates/Themes would really struggle with rendering these in a standard way. I'm not sure what the right answer is, but whatever it is it needs to be consistent so that Themes can just be interchangeable. 
Additionally you've written out the key "googleplus" where others may just have used "google".

>Perhaps the ideal approach is to allow the definition to be loose but the parsing and internal access to be rigid.

>This would mean that "google" and "googleplus" would both match the social network "google". The ingestion could examine the input and parse it out into separate url and token values for each social network.

>I would think a struct like:

```
type socialNetwork struct {
    name         string
    aliases       []string
    token         string
    urlPattern  text.template
}

type socialEntry struct {
    network     socialNetwork
    token         string
}
```

>Or something similar for each social network would be defined in the code. Then inside Site and Author you would have a map of [socialnames]SocialEntry.

>Not sure if any of that made sense, but the idea is similar to how we handle params.

moorereason:
@derekperkins,

I agree with you (and Steve).  I'd much rather launch an extensible framework from the beginning if we can.  Standardized author data would be great for the community, but I don't think we should rush it out the door when we know it's half-baked.  If we want to create a standard, let's do it right the first time.  Let's *not* release a standard and then deprecate it on the next release.

Also, don't forget that this PR needs tests.

derekperkins:
I'm not suggesting that the current implementation is necessarily wrong or going to have to be deprecated. I think the plugability for social is going to happen on the network level.

```
// inside some /data/_socialNetworks/someUnsupportedNetwork.toml
    name         string
    aliases      []string
    token        string
    urlPattern   text.template
    

// author.toml
[social]
    facebook               = "derekperkins"
    someUnsupportedNetwork = "abc/646.@#%"
```

If that is all that adding new network support that isn't baked into Hugo requires, then the current config / api is sufficient. We can change or upgrade the internals without breaking users, and per the Hugo standard, we can make all built-ins overridable.

>Let's not release a standard and then deprecate it on the next release.

Agreed, but I think waiting until things are perfect is the wrong call.

moorereason:
@derekperkins, my apologies.  I haven't had time to test this PR, so I was making some assumptions based on how you described the situation.  If we can extend it like you say, then I'm good with it.

derekperkins:
@moorereason No worries, I apologize if I came across as defensive. I just wanted to step back and make sure that I hadn't lost sight of the forest for the trees and missed something obvious that we'd regret later.

derekperkins:
@bep Rebased, squashed and ready to go

bep:
I haven't read the latest discussion, but my main concern is the API towards the user. If the internals are a map, a struct or whatever ... we can fix that in the future.

digitalcraftsman:
The corresponding documentation has been merged.

dencold:
Hey all, sorry I'm a bit late to the punch here. I've been away traveling and haven't had time to monitor my github feed. Last bit of discussion seemed to be centered around @derekperkins suggestion to have social details as a struct instead of a simple map. I'm in favor of this but I think it warrants its own separate issue to provide the proper focus. My vote here would be to keep the current social map as is for 0.17 and then have a fully formed implementation for 0.18. Just my 2 cents.

Thanks to everyone for pulling this PR together. I'm really looking forward to updating my theme to use the new setup.

digitalcraftsman:
This means we finally postpone this feature, right? In that would also revert the corresponding documentation that I added in #2419.

rdwatters:
@derekperkins So sorry for the delay; I'm always traveling for work these days. Like always, my input will be limited to the non-go-programmer-and-rank-amateur-js-developer variety, but I'm curious about the value-add on this feature. These are my thoughts after just reading @digitalcraftsman's docs and #2419:

1. W/r/t using a `.toml` file in `_data`, what about sites that have 50 or even 100 authors? Is this easier to create and manage than using `content/authors` as a section, and then setting up an `author` archetype with whatever metadata/front matter that needs to be repeated? My workplace is not a typical use case, but the attributes I'd want on an author content type would be things like institution, CV, degrees, job title, etc and probably wouldn't include anything in `[social]`. You obviously account for this with the user-defined `[params]`, but just a thought. I'm currently just treating "authors" as a content section on a [site I'm building for my team at work](https://digitalredesigncomp.netlify.com/digital-team/) (no judgment; definitely still a WIP, but the team section shows that creating content files for each author, referencing `authors:[]` in front matter, and then creating a partial (`author-articles.hmlt`) is negligible.
2. Is the current problem one that might be better solved via improvements to list pages? This is brought up in multiple locations across the discussion forum. For example, have "authors" be a taxonomy like anything else, with the ability to add attributes (descriptions, bios, social, etc) to the list pages that Hugo currently creates for taxonomies? Or is the idea that this PR and benefit is specific to theme creation? 

Nevertheless, I think there are some nice conveniences added with what I could glean from the bang-up docs @digitalcraftsman put out there. I'm also a fan of being able to create a site author through the CLI. 

I'm all for it; thanks again to all the awesome Hugo maintainers for making the world's best SSG:)

derekperkins:
While I understand the reasoning, it's still really frustrating to have this put off again.

derekperkins:
@rdwatters 

1. I think this is going to be especially helpful for sites with 50-100 authors. It's a good idea to be able to define an author archetype for use with the CLI.
1. There will be author archives, that's just waiting on the Node vs Page upgrades.

Reasons why having first class author support is better:

1. **Portability** - like you mentioned, you can create your own hierarchy and lookups, but that will always be site specific
1. **Common** - Almost every site using Hugo is going to want to use authors at some point on a blog section, so making everyone re-invent the wheel makes adoption harder. Inside the author, many sites provide social links in an author byline, so those are also included as has having first class support
1. **Templates** - There are a lot of SEO and social benefits from having that data available that we can start building into Hugo. It enables extra settings for OpenGraph, Twitter cards, Author schema markup, etc
1. **Convenience** - It enables things like built in sorting by first name, last name, display name, etc. That requires significantly more work in a home-rolled solution

rdwatters:
@derekperkins 

In response to the bold ordered list:

1. Agreed.
2. Agreed. Making everyone reinvent the wheel is indeed a bad thing.
3. Agreed. But I think you're overstating it. I haven't had issues creating json+ld for schema.org, twitter cards, OGP, etc, in a few re-usable partials. And there is more to the aforementioned schemas than just author information, but I can see how standardizing it would help create better internal partials that new users won't have to think about.
4. Agreed++ These are things I didn't think of that would make for much more convenience:smile:

Out of curiosity, does the added author feature have any effect on performance/build times?

derekperkins:
Yeah, the "not having to think about it" is the real benefit. For someone like you who knows what they're doing, it's not difficult to roll those partials. For most other users who have never built those out or just installed an SEO plugin on a previous CMS, they're likely not including those at all, when they should be included by default in every theme IMO.

Authors won't impact performance any differently than having them in a section. You'd likely see improved performance by having all the lookups in the core rather than piping through partials and templating like you are doing today.

dencold:
Sorry! I didn't mean to imply that #1850 should be held off until 0.18, just that the social component should be. Does that not make sense? If we feel like the current implementation of socials as a map is not a good precedent to set, maybe we just drop it entirely and only set the direct `Author` metadata as the 0.17 release.

Do we have a timeline for when 0.17 will be released?

derekperkins:
Ok, all author PRs (minus @digitalcraftsman's docs) have been consolidated here.

I could use some help tracking down #2469. I don't know how multi-lingual stores multiple languages. Are they treated as different sites entirely? If that's the case, then I can see how it might not load.

I also put in treatment for #2467 that allows for language specific overrides, but if multi-lingual sites are treated as separate sites, it may not work. If that commit looks good, I'll add tests.

trans:
@rdwatters Makes me think of an interesting idea. Instead of putting authors in `data/_authors`, put them under `content/authors` as [hcards](http://microformats.org/wiki/hcard). The data can be parsed out from there for templates, and even included whole hog and let CSS handle formating.

pksunkara:
Any idea when this will be merged?

digitalcraftsman:
It is planned to merge it shortly after the release of the next version of Hugo. We (sadly) had to postpone this feature due to issues that came up on the last mile. Currently, we all wait that @spf13 makes the final release.

pksunkara:
The `.twitter` in this line needs to be edited into `.Social.twitter` with this PR.

pksunkara:
Sorry, here is the line, https://github.com/spf13/hugo/blob/master/tpl/template_embedded.go#L216

derekperkins:
@pksunkara Good catch, I just fixed that.

tpl/template_embedded.go
```diff
@@ -213,7 +210,7 @@ func (t *GoHTMLTemplate) EmbedTemplates() {
 {{ with .Site.Social.twitter }}<meta name="twitter:site" content="@{{ . }}"/>{{ end }}
 {{ with .Site.Social.twitter_domain }}<meta name="twitter:domain" content="{{ . }}"/>{{ end }}
 {{ range .Site.Authors }}
```
pksunkara: I think you need to change the `.Site.Authors` to plain `.Authors` so that it refers to the `Authors` function in `page.go`. Similarly, I know there are other occurrences. You might want to grep for `.Site.Author` and check everyone of it closely once.

hugolib/author.go
```diff
@@ -43,3 +78,205 @@ type Author struct {
...
+		case "email":
+			author.Email = cast.ToString(data)
+		case "weight":
+			author.Weight = cast.ToInt(data)
+		case "social":
+			author.Social = normalizeSocial(cast.ToStringMapString(data))
+		case "params":
+			author.Params = cast.ToStringMapString(data)
+		case "languages":
+			if author.languages == nil {
```
bep: I don't understand your `languages` implementation -- but looking at the code below, with the languageOverride func, it suspect it is too elaborate. We have widespread way of configuration overrides of maps in Hugo (and Viper), and this should be done at configuration load time.

In my head the `languages` in the author context isn't *languages the author speaks*, but different configurations for the different languages (aka different `shortbio` in Norwegian etc.)

So, pseudo speaking:

```
# defaults 
givenName = "bep"
bio = "Developer from Norway"

[languages]
[languages.no]
bio = "Utvikler fra Norge"
```

derekperkins: `languages` functions exactly how you describe. All it does is take the language set on the `Node`, then does a lookup on properties and replaces them in the defaults if the language specific property isn't empty. If you would like it implemented in a different fashion, show me a configuration override that you'd rather have me use.

derekperkins:
@bep Is there a good multi-author integration test already set up? I'd rather not start from scratch. I'm still not familiar with how they function, so I believe #2469 is still outstanding. If someone with a better idea how they are implemented could chime in, that'd be helpful.

hugolib/author.go
derekperkins: Either there is a nested map like I have here inside the author, or you map authors by language and copy all the default info there. The way I have it implemented now doesn't copy any info, but incurs an extra if statement per access + 1 map access per nested property.

bep:
>@bep Is there a good multi-author integration test already set up?

Not multi-author, but you can find good examples in:

* Complex: hugo_sites_test.go
* Basic: alias_test.go

The main stuff to understand here, which is new in Hugo 0.17: 

* The entire chain (excluding the CLI) is now easily testable
* Reloading is also testable by manually sending in file change events 
* Every (almost) file operation is backed by Afero fs, so it is easy to test the real stuff.

I have even written a blog post about it: http://bepsays.com/en/2016/10/12/pragmatic-and-effective-testing-in-go/

hugolib/author.go
derekperkins: Authors are only defined in `/data/_authors` and the author key/ID is referenced in front matter.

If you don't have any specific suggestions, I will just leave the implementation as is, which works, is clear and is tested.

tvb:
Status?

derekperkins:
It needs to be refactored due to the recent Node/Page changes. It just feels like a constantly moving target to get it merged.

digitalcraftsman:
> It needs to be refactored due to the recent Node/Page changes. 

I rebased the code related to this feature and updated the tests as well. You can find my changes [here](https://github.com/digitalcraftsman/hugo/commits/authors). 

Would it be possible to add commit access to the branch of this PR @derekperkins? Otherwise we have to merge changes from different repos.

derekperkins:
I just granted you collaborator access to my fork.

digitalcraftsman:
Thanks

derekperkins:
@bep Since I can't comment on #3088, are you proposing to scrap this PR?

derekperkins:
@bep I'd like to comment on your author proposal in #3088, but I don't have permissions. Where would you like me to put my feedback?

moorereason:
@derekperkins, I unlocked #3088.

sirinath:
Can this be updated reviewed? But instead of data make it flexible to configure where data of authors is stored. It can be in data or taxonomy as many sites use taxonomies for multiple authors. This data can be made available through `.Site.Authors` and `.Page.Authors`.

coolaj86:
# 2021 Update - Good News 🎉

There are no official docs, but someone has documented how `Author` and `Authors` work by looking into the source code - specifically, how `index.xml` is rendered.

See here:
[discourse.gohugo.io/t/site-author-usage/31459/8](https://discourse.gohugo.io/t/site-author-usage/31459/8?u=coolaj86)

PR #1852, opened 2016-02-13
Export MainSite var so I can fix hacdias/caddy-hugo#46

hacdias:
Hello!

There has been an [issue](https://github.com/hacdias/caddy-hugo/issues/43) on ```caddy-hugo``` plugin for Caddy related to multisite functionality.

I ended up discovering that to fix that bug, I would have to redefine ```MainSite``` to ```nil``` so it didn't use the configurations of a previous generated website, creating a new instance.

That's why I'm making this pull request. I hope it can be merged asap.

Thanks,

### DIFF
--- a/commands/benchmark.go
+++ b/commands/benchmark.go
@@ -14,9 +14,10 @@
 package commands
 
 import (
-	"github.com/spf13/cobra"
 	"os"
 	"runtime/pprof"
+
+	"github.com/spf13/cobra"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


bep:
Merged in ad17605

Please note the modified commit message for future reference.

AND: We need to figure out a better API ...

hacdias:
Thank you @bep. And yes, that would be better and I've some ideas. Later I'll share them.

Issue #1635, opened 2015-11-27
Feature request: create `index.json` or `index.js` for client-side search libraries like Tipue and Lunr.js.


rdwatters:
### 1. Roll Your Own
- `range` to loop through all `.Site.Pages`, escape the html/md (similiar to `| jsonify` in Liquid), and write it to a file with a `.js` or `.json` file extension. This is spitballing, but to keep everything HTML, maybe it's as simple as a parameter in `layout/searchIndex.html` that allows for a different file extension to be copied to `public/`. 
### 2. [Lunr.js](http://lunrjs.com/)
- Loops through every content page and pulls title, subtitle, description, tags, categories, dates, and body copy
- Excludes all .md files with `inSearch: false` / `inSearch = false`   in the toml/yaml of individual file; ie, if the following is set to `true`...
- Global param in `config.toml` /  `config.yaml` with `createJsonIndex = true`. Maybe the default behavior should be set to `false` in case something like this would slow down build time for larger sites. The default destination could be `static/js` since this dovetails with the folder structure for themes.
- Lunr requires some JS acumen to display your contents to the page. Larger sites would need a webworker. That might get trickier for newbs. Also, this is less customizable than option 1.  

This would then make it really easy to add weights to properties and have lunr.js index content on the client and display results by score. 
### 3. [Tipue](http://www.tipue.com/search/)
- The same as #2, but allow to write to a `.js` file instead for a `var tipue_search = {}...`

digitalcraftsman:
:+1: This feature would be really nice and could replace a lot of Gulp/Grunt scripts that are used the generate the search index.

If you range over all files, take care that you only include files that aren't marked as drafts.

rdwatters:
@digitalcraftsman Excellent point and something I should have mentioned above. At the page level, the index would look at two conditionals: 

1. `insearch != false`
2. `draft != true`

Thanks!

bep:
Agree about the "usefulness", but I'm pretty sure this is somehow a duplicate of some other issue.

digitalcraftsman:
This is related to #144

bep:
See https://discuss.gohugo.io/t/simple-json-site-index-in-hugo/2854

rdwatters:
I'm closing this issue since it's now a non-issue with the upcoming custom outputs...

Issue #144, opened 2013-12-01
Save data to "data/pages.json"


robotamer:
How about saving all the data to json files.
These could be used for ajax calls from client site.
Could also help with i18 (Multi Language) etc

oskarrough:
This would be a great feature! For the schema there's always http://jsonapi.org/ to look at.

rhewitt22:
Creating a template that outputs JSON would be great in order to build a site index.  The index would be used for client-side search with something like lunrjs.

digitalcraftsman:
There is some progress in this direction. See #1853.

bep:
There is fairly good JSON support for this use case in Hugo now. Improvements to this is covered by other issues.

Issue #1744, opened 2016-01-03
Native and complete i18n support


abourget:
As stated in https://github.com/spf13/hugo/pull/1734

---

@spf13 I was also looking at https://godoc.org/github.com/nicksnyder/go-i18n/i18n ...wouldn't it be awesome if, in addition to good support for i18n in terms of content, we had great translation capabilities?

We could load `i18n/{{$.Site.RenderLanguage}}.yaml` as an `go-i18n` translation on start, just like we load `.Site.Data`.. This would require `Multilingual` to be enabled.

I can see quite a nice mapping from the examples there to something Hugo-ish:
```
T("Hello {{.Person}}", map[string]interface{}{
    "Person": "Bob",
})
```
to something like `{{ i18n "Hello {{.Person}}" (dict "Person" "Bob") }}` or something like that.

Or `T("You have {{.Count}} unread emails.", 2)` -> `{{ i18n "You.. {{.Count}} unread mails" .Site.MailCountWhatever }}`

I'd add a native `i18n` function, or `T` (or both, one aliasing the other ?) to handle that.

What do you think ? What's your take on vendoring other libraries ?

spf13:
Fine with vendoring.

I think this should be a bit different. I think we should integrate it a
bit deeper and tighter. How about this...

Multilingual doesn't need to be enabled. If we have translations / language
files than it just works.


For the translation I think we have a dedicated folder (can be inside data)
for translation maps.

Then we have a template function & shortcode that looks up the appropriate
value for the current language.

so Hello {{ T "person" }}

would read from

/data/translations/en.yaml
/data/translations/es.yaml

etc.

Just ideas of what feels cleaner to me. I think having the mapping in the
content / templates is odd.

Does that make sense?



--
Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

abourget:
yeah it does :) I just took a bite at this one.. :) wanna take a look ? https://github.com/spf13/hugo/commit/1881c36569e2c05a8c90688427b43f509debcb6b

abourget:
about vendoring, I've seen this tool here: https://github.com/kardianos/govendor which manipulates the vendor dirs.. this would seem to be the best way to do vendoring, because it would all be native, yet still have some annotations..

On other projects, I used `git subtree` but sometimes it's more difficult to manage for others.. and it's less discoverable.. `govendor` is pretty discoverable.

abourget:
[...]

Issue #1128, opened 2015-05-11
Add ical, xcal support


bep:
See

http://discuss.gohugo.io/t/calendars-ical-xcal-co/1147/3

trentfisher:
I'm not sure what the intent here is.  As for me, I'd like to have an ical file of the dates and times of posts.  Extra credit for another ical with future posts included.  When I ran wordpress, I had an extension which did both of those and it was quite handy.

Perhaps this could be generated at the same time as the RSS?

gour:
I'm also interested to be able to render calendars synced via ownCloud in my web sites. Here is one [project](https://github.com/PuloV/ics-golang) which might help...

BrianHicks:
Another use case: I'm organizing a conference (www.elm-conf.us) where I'm using Hugo. I'd like to render an iCal feed of my talks for people to subscribe to.

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

juh2:
In the Pandoc Community the answer of such a question would be: Write a filter or write your own writer. Does Hugo support filters? I think that there are ways to take an input file (Markdown or YAML etc.) and write out an ical file, eg. with a Python script, but this would require a second process and slow down the build time. 

But perhaps Hugo can get a feature that while switched on collects all the metadata of all files and hand it over to an external process that eg. writes out ical files from the collected metadata. So the feature would be a switch which tells Hugo to provide a big metadata file and pipe it through a given filter before finishing its own task. If the filtering is only about writing ical files the filter can signal Hugo to go on immediately so that build time is not slowed down. If the filter is doing something that would affect the Hugo output, Hugo has to wait until the filter stops.

[...]

PR #1853, opened 2016-02-13
Add generator for a search index

digitalcraftsman:
The generator creates an index of all content
files and it's metadata.

See #1635 #144

### DIFF
--- a/docs/content/templates/variables.md
+++ b/docs/content/templates/variables.md
@@ -181,6 +181,8 @@ Also available is `.Site` which has the following:
 **.Site.Language** This indicates which language you are currently rendering the website for.  This is an object with the attributes set in your language definition in your site config. For the language code, use `.Site.Language.Lang`.<br>
 **.Site.LanguagePrefix** This can be used to prefix  theURLs with whats needed to point to the correct language. It will even work when only one language defined. See also the functions [absLangURL and relLangURL]({{< relref "templates/functions.md#abslangurl-rellangurl" >}}).<br>
 **.Site.Languages** An ordered list (ordered by defined weight) of languages.<br>
+**.Site.DisableSearchJSON**  A boolean (by default false) to indicate whether to build a content index. Defined in the site configuration. Read more about the [search]({{< relref "templates/search.md" >}}) feature.<br>
+
 
 ## File Variables
 
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -176,6 +176,8 @@ type SiteInfo struct {
 	Permalinks            PermalinkOverrides
 	Params                map[string]interface{}
 	BuildDrafts           bool
+	DisableSearchJSON     bool
+	SearchIndexLink       string
 	canonifyURLs          bool
 	preserveTaxonomyNames bool
 	paginationPageCount   uint64
@@ -821,6 +823,11 @@ func (s *Site) render() (err error) {
 	}
 	s.timerStep("render and write robots.txt")
 
+	if err = s.renderSearchJSON(); err != nil {
+		return
+	}
+	s.timerStep("render and write search.json")
+
 	return
 }
 
@@ -914,6 +921,8 @@ func (s *Site) initializeSiteInfo() {
 		GoogleAnalytics:                lang.GetString("GoogleAnalytics"),
 		RSSLink:                        permalinkStr(lang.GetString("RSSUri")),
 		BuildDrafts:                    viper.GetBool("BuildDrafts"),
+		DisableSearchJSON:              viper.GetBool("DisableSearchJSON"),
+		SearchIndexLink:                viper.GetString("baseURL") + viper.GetString("searchuri"),
 		canonifyURLs:                   viper.GetBool("CanonifyURLs"),
 		preserveTaxonomyNames:          lang.GetBool("PreserveTaxonomyNames"),
 		AllPages:                       &s.AllPages,
@@ -2017,6 +2026,28 @@ func (s *Site) renderSitemap() error {
 	}
 
 	return nil
+}
+
+func (s *Site) renderSearchJSON() error {
+	if viper.GetBool("DisableSearchJSON") {
+		return nil
+	}
+
+	n := s.newNode("search")
+	n.Data["Pages"] = s.Pages
+
+	rLayouts := []string{"search.json", "_default/search.json", "_internal/_default/search.json"}
+	outBuffer := bp.GetBuffer()
+	defer bp.PutBuffer(outBuffer)
+	err := s.renderForLayouts("search", n, outBuffer, s.appendThemeTemplates(rLayouts)...)
+
+	searchuri := viper.GetString("SearchUri")
+
+	if err == nil {
+		err = s.writeDestFile(searchuri, outBuffer)
+	}
+
+	return err
 }
 
 func (s *Site) renderRobotsTXT() error {
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -140,6 +140,7 @@ var (
 	disable404            bool
 	disableRSS            bool
 	disableSitemap        bool
+	disableSearchJSON     bool
 	draft                 bool
 	forceSync             bool
 	future                bool
@@ -228,6 +229,7 @@ func initHugoBuildCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&disable404, "disable404", false, "Do not render 404 page")
 	cmd.Flags().BoolVar(&disableRSS, "disableRSS", false, "Do not build RSS files")
 	cmd.Flags().BoolVar(&disableSitemap, "disableSitemap", false, "Do not build Sitemap file")
+	cmd.Flags().BoolVar(&disableSearchJSON, "disableSearchJSON", false, "Do not build Search JSON file")
 	cmd.Flags().StringVarP(&source, "source", "s", "", "filesystem path to read files relative from")
 	cmd.Flags().StringVarP(&contentDir, "contentDir", "c", "", "filesystem path to content directory")
 	cmd.Flags().StringVarP(&layoutDir, "layoutDir", "l", "", "filesystem path to layout directory")
@@ -319,6 +321,9 @@ func InitializeConfig(subCmdVs ...*cobra.Command) error {
 		}
 		if flagChanged(cmdV.Flags(), "enableRobotsTXT") {
 			viper.Set("EnableRobotsTXT", enableRobotsTXT)
+		}
+		if flagChanged(cmdV.Flags(), "disableSearchJSON") {
+			viper.Set("DisableSearchJSON", disableSearchJSON)
 		}
 		if flagChanged(cmdV.Flags(), "pluralizeListTitles") {
 			viper.Set("PluralizeListTitles", pluralizeListTitles)
--- a/docs/content/overview/configuration.md
+++ b/docs/content/overview/configuration.md
@@ -101,6 +101,10 @@ Following is a list of Hugo-defined variables that you can configure and their c
     disableLiveReload:          false
     # Do not build RSS files
     disableRSS:                 false
+    # Do not build search.json file
+    disableSearchJSON:          false
+    # Change the destination of the search index
+    searchURI:                  "search.json"
     # Do not build Sitemap file
     disableSitemap:             false
     # Build robots.txt file
--- a/docs/content/templates/rss.md
+++ b/docs/content/templates/rss.md
@@ -7,7 +7,7 @@ linktitle: RSS
 menu:
   main:
     parent: layout
-next: /templates/sitemap
+next: /templates/search
 notoc: one
 prev: /templates/partials
 title: RSS (feed) Templates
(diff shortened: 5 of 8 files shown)


digitalcraftsman:
The generator is working so far but the implemenation isn't finished yet (see TODO).

Is there anything ground-breaking to complain about the generator?

TODO:

- [ ] add CLI flag that triggers the build
- [ ] exclude pages with `excludefromindex` variable in frontmatter
- [ ] add option to exclude whole content types

Currently, I've only an option turn the generator on/off. Does it make sense to stop the regeneration in watch mode (`hugo server`) and only trigger it in production (`hugo`)?
Should more data be included in the index?
How would you set up a test function?

/cc @rdwatters @bep @spf13

rdwatters:
@digitalcraftsman This is pretty fantastic. Thanks for working on what I see as a really powerful feature (and one I have kind of been nagging about). 

**Question for you:** rather than just the ability to create an index that can be built/not built with a flag, how difficult it would it be to just extend Hugo's abilities to write to any .json file using the same templating logic? Would a feature like this (Jekyll has the ability to write JSON files, which comes in pretty handy) slow down builds to the point of not being worth it? 

Do you think writing to any-file.json rather than site-index.json would provide the most flexibility (ie, w/r/t using ajax, etc), or is the primary objective to allow for client-side search a la something like Tipue or lunr.js? Cheers.

digitalcraftsman:
> rather than just the ability to create an index that can be built/not built with a flag, how difficult it would it be to just extend Hugo's abilities to write to any .json file using the same templating logic? Would a feature like this (Jekyll has the ability to write JSON files, which comes in pretty handy) slow down builds to the point of not being worth it?

I think that would be a powerful addition to the current set of template functions. I'm not sure how much it would slow down the generation of pages. A possible implemenation could be the use of a global object shared object, like [Scratch](http://gohugo.io/extras/scratch/), but with a few more parameters (filename, destination). Inside a template a coder could speficy, if necessary with if-else statements, what  should be included in which file. 

> Do you think writing to any-file.json rather than site-index.json would provide the most flexibility (ie, w/r/t using ajax, etc), or is the primary objective to allow for client-side search a la something like Tipue or lunr.js?

This is a good question. I would like to integrate both. Just to have setting a config variable to true is the most user-friendly way. Your approach would allow much more flexibility, but the user/theme creator maybe needs to include in many different places. Imagine a user has different layouts for different content types. He would need to add the logic in each template file. Shortcodes would be a handy way to avoid redundant code.

Thank you @rdwatters for sharing your thoughts and ideas. 

> I think HUGO is easily the best SSG around. Cheers!

This project made good progess in it's rather short lifetime :smile:. I'm curious what we will see in the v1.0 release.

rdwatters:
@digitalcraftsman Good points all round. I guess there would be the two options, and ultimately I'm bringing up two separate feature requests:

1. Create site index with flag or config boolean - no need for everyone to build a site index on every build. That said, do you think it would be best to include an `excludefromindex` at the page level, the section level, or both? How would you go about disincluding, eg, list pages?
2. Ability to write to json files in general. You're right that this would have to be a separate process in that forcing devs to write templating to account for *all* content/section areas in a single site-index.json would be a little tedious.

bep:
>Ability to write to json files in general.

There is an open issue somewhere about rendering custom content-types, like JSON, ical ... whatever. We should do that.

digitalcraftsman:
> There is an open issue somewhere about rendering custom content-types, like JSON, ical ... whatever. We should do that.

If you, @bep, agree with @rdwatters and me we should consider this as two different issues. See #1128

spf13:
These are two different issues.

The PR is effectively a sitemap in JSON which will enable lots of nice integrations.

A second issue is for Hugo to support rendering into variable and multiple multiple different formats.
The second issue is a pretty considerable one which would require quite a bit of restructure. Including the integration of the text.Template library along side of the html.Template one.

hugolib/content_index.go
```diff
@@ -0,0 +1,86 @@
...
+
+type index struct {
+	Pages []*entry
+}
+
+type entry struct {
+	Title      string      `json:"title"`
+	Content    string      `json:"content"`
+	Permalink  string      `json:"permalink"`
+	Tags       interface{} `json:"tags"`
```
spf13: We shouldn't assume these taxonomies are being used. I think this is a very limiting approach.

hugolib/content_index.go
```diff
@@ -0,0 +1,86 @@
...
+				Categories: p.Params["categories"],
+				Date:       p.Date,
+				LastMod:    p.Lastmod,
+			}
+
+			rawIndex.Pages = append(rawIndex.Pages, entry)
+		}
+	}
+
+	jsonIndex, err := json.Marshal(rawIndex.Pages)
```
spf13: While I see the benefit of using json.Marshal to generate the JSON, this approach is also quite rigid defining the explicit fields that will be rendered/generated. 

What if instead we used a similar approach to how the sitemap works using a template instead to generate the output. It would be much more flexible.

rdwatters:
@digitalcraftsman Spitballing on this, but is there utility in implementing a stopwords list when creating the index? Here's a decent default list.

<http://www.ranks.nl/stopwords>

If the intention is client-side search, it looks like it's the same [stopwords used by Tipue](https://github.com/Tipue/Tipue-Search/blob/master/tipuesearch/tipuesearch_set.js) and similar to the [stopword filter for lunr.js](http://lunrjs.com/docs/#stopWordFilter). That said, if search results were designed to surface, say, the "description" key in front matter, SERPs would look weird if every definite and indefinite article were omitted from the page. Then again, maybe eliminating stopwords from the index before it's sent could make filesize smaller and potentially reduce demand on the client. 

It goes without saying that internationalization efforts being worked on outside this thread would have a different list.

moorereason:
If you want to remove stopwords in Go, check out https://github.com/bbalet/stopwords.  It's multi-lingual and has already been discussed on the forums for adding a related posts feature (https://github.com/bbalet/gorelated).

rdwatters:
:+1: @moorereason

bep:
The use of stopwords is within the role of the tokeniser/indekser. This PR is badly named, as it doesn't create a search index, it exports the content in a format suitable for indexing.

digitalcraftsman:
> This PR is badly named, as it doesn't create a search index, it exports the content in a format suitable for indexing.

After revisiting some disucssion here and in the forum I agree. As this PR is currently a WIP, it should just output a json file that is intended for searching the content (with lunr.js or similar tools). Using a stopword filter would consequently be the next step for optimizations.

As I discussed with @rdwatters before, we should create a seperate `jsonify` template function that can query content like the user wants it.

digitalcraftsman:
Since we have a `jsonify` template func and and a good proof-of-concept for a [content index](https://github.com/bep/bepsays.com/commit/1d7bf7fd6e7c637f28a04f93a97ebfed084c27ff) (thanks @bep) it would be perfect to do it this way in this PR. This would satisfy @spf13 wish for more flexibility.

However, I saw that @bep needed to create a new content file just set the url properly. Wouldn't it be better to add a `saveas` template func that saves a string, JSON object etc. as file under a given path:

```go
{{ $contentList | jsonify | saveas "/index.json" }}
```

The path would be relative to `static/`.

/cc @bep @moorereason

bep:
Yes, the extra content file is not good, we need better support for custom file types (json, ical etc.), but the answer isn't `saveas`(where would you call that from?)

digitalcraftsman:
>  but the answer isn't `saveas` (where would you call that from?)

I would call it from inside a template, like in the example above:

    {{ $contentList | jsonify | saveas "/index.json" }}

The function itself would have a signature like 

```go
func saveas(path string, data interface{}) error {}
```

digitalcraftsman:
I revisited this issue and implemented the feature with a template as @spf13 suggested. That gives users the miximal flexibility. Kudos to @bep for implementing the `jsonify` template func and for providing a [good starting point](https://github.com/bep/bepsays.com/commit/1d7bf7fd6e7c637f28a04f93a97ebfed084c27ff) for the internal template.

I would appreciate a review. According to the contribution guidelines the commit message should mentioned the modified package as prefix. Since I modified multiple packages which should I use?

@rdwatters you asked for an option to exclude certain pages. Have a look at the docs :wink: 

Furthermore, @rdwatters  and @moore suggested the usage of a stop word filter? Should this be realized with a template function (in a seperate pull request)?

Last but not least I would like to keep an eye on the localization support (#1744). Having search results in multiple languages doesn't make sense in my opinion. Should we offer an option to generate a content index per locale?

moorereason:
First, my handle is `moorereason`.  No need to spam whoever `moore` is.

Second, for the commit message prefixes, you want to use the primary affected package (I use that phrase in my updated but yet-to-be-merged [contributing guide](https://github.com/moorereason/hugo/blob/contrib-guide/CONTRIBUTING.md#git-commit-message-guidelines)).  In this case, commit  bb688f7 would use `hugolib`, in my view, since that's where the most important change is made.  Choose which package you feel is most relevant to call out in the commit message.

In your subsequent commits, I'd use `commands`, `hugolib`, and `docs`, respectively.  The idea is to give someone looking over the git logs a quick identifier of where the changes are occurring without them having to read the full commit message or look at the diffs.

I get the feeling I'm going to need update  the contributing guide to give a fuller explanation and rationale for the subject prefix.

digitalcraftsman:
I'm sorry for misspelling your handle.

The commit messages have been updates with their corresponding package as prefix. However, at first I just wasn't sure if the commits should be squashed or not.

digitalcraftsman:
I implemented the search feature in the material-docs theme and it works like a charme. But the usability of the default template could be improved. 

Currently, we are only linking the pages who match the search query. It would be much better if we also could link the headers of section that contains (parts) of the query. MkDocs uses the [headers as dividers](https://squidfunk.github.io/mkdocs-material/mkdocs/search_index.json) for the content and adds each of them as new search result.

hugolib/site.go
```diff
@@ -784,6 +791,8 @@ func (s *Site) initializeSiteInfo() {
 		GoogleAnalytics:       viper.GetString("GoogleAnalytics"),
 		RSSLink:               s.permalinkStr(viper.GetString("RSSUri")),
 		BuildDrafts:           viper.GetBool("BuildDrafts"),
+		DisableSearchJSON:     viper.GetBool("DisableSearchJSON"),
+		SearchIndexLink:       viper.GetString("baseURL") + viper.GetString("searchuri"),
```
digitalcraftsman: @bep Is there any helper functions that can prepend the baseurl?

docs/content/templates/variables.md
```diff
@@ -158,6 +158,7 @@ Also available is `.Site` which has the following:
 **.Site.Permalinks** A string to override the default permalink format. Defined in the site configuration.<br>
 **.Site.BuildDrafts** A boolean (Default: false) to indicate whether to build drafts. Defined in the site configuration.<br>
 **.Site.Data**  Custom data, see [Data Files](/extras/datafiles/).<br>
+**.Site.DisableSearchJSON**  A boolean (by default false) to indicate wether to build a content index. Defined in the site configuration. Read more about the [search]({{< relref "templates/search.md" >}}) feature.<br>
```
moorereason: s/wether/whether/

hugolib/site.go
moorereason: Looks like `s.permalinkStr()` is used above for the `RSSLink`.

digitalcraftsman: That's what I've done before. I printed the URL in a template and got `http://localhost:1313/search/index.json/` instead of `http://localhost:1313/search.json`

docs/content/templates/variables.md
digitalcraftsman: What do you mean with "s/wether/whether/"?

moorereason: Sorry, old sed/perl syntax shorthand.  In english: wherever you have "wether", substitute "whether".

hugolib/site.go
moorereason: That sounds like a bug.

docs/content/templates/variables.md
digitalcraftsman: I fixed the typo.

hugolib/site.go
digitalcraftsman: @bep do you know if this behavior is intended or how it can be avoided?

bep: As to helper, see what is used by absURL template func.

bep:
As to the discussion of stop-words:

1) I think it is in this case the responsibility of the search lib.
2) Hugo could have used such a feature (but it is hard: I have seen some of the stop-words lists for Norwegian, and they are crappy), but then as a cross-cutting concern that could be used by others, in this case as a filter.

This this PR should be about geting the data in a parseable format, aka JSON.

derekperkins:
Is this going to make it into 0.18?

digitalcraftsman:
I'm closing this pull request in favor of #2828. Custom output types would be much more flexible. Users could create content in a format they want by using templates and by specifying the output type (e.g. JSON).

My approach would be to specific and de facto deprecated once you can achieve the same with custom output types.

PR #1856, opened 2016-02-13
Clarify that Amazon WS is, actually, AWS

ajlanghorn:
Amazon WS is an uncommon way of referring to Amazon Web Services, which is usually referred to either by its full name, or as AWS.
Fixes this.

### DIFF
--- a/docs/content/overview/usage.md
+++ b/docs/content/overview/usage.md
@@ -182,7 +182,7 @@ Here is the command:
               --bind=87.245.198.50
 {{< /nohighlight >}}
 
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Hello @ajlanghorn,

I agree with you that either AWS or the full name are more common. Thank you for the fix. Your commit has been merged as ce7c267140610ea3435a810100dc79a5dee01d5f (master) and 67c7d8d39cca1728bab847e83afcf9beac3fc69c (v0.15.docs).

Best regards,
Digitalcraftsman

PR #1859, opened 2016-02-14
Unexport mainSite and hugoCmd; create a very simple API

hacdias:
@bep I don't know if this API (if it can be called that way) is a better way to use Hugo. If so, it can be improved in the future to add more functionalities.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1,0 +1,15 @@
+package api
+
+import (
+	"github.com/spf13/hugo/commands"
+	"github.com/spf13/viper"
+)
+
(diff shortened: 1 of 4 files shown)


hacdias:
I'm not understanding why this is failing [here](https://travis-ci.org/spf13/hugo/jobs/109166956#L181).

bep:
Try to rebase against master.

hacdias:
Still same error. But that works!!! Affs.

hacdias:
Moved to #1860

PR #1868, opened 2016-02-17
Add cdnoverview.com to the showcase

wouterspee:


digitalcraftsman:
Hi @wouterspee,

thanks for adding your website to the showcase. It's not necessary to trigger wercker again because a failed build could be caused by some other commits that changed the actual codebase of Hugo. However, I've merged you addition as 086d6540c32bad0745697e9e823298a7dddf7f75 (master) and 30d089b59892d057b9d1444f55819575f0563dce (v0.15.docs).

You should see you website within a few minutes in the showcase.

Cheers,
Digitalcraftsman

PR #1870, opened 2016-02-17
Close file handle when finished reading file

davidcroda:
the resGetCache and resGetLocal functions open file handles but do not close them when finished. This can cause "too many open files" errors when using `hugo serve` on files with getJSON calls.


bep:
Good catch. Thanks. Merged in 284e959.

PR #1878, opened 2016-02-18
Updated Dockerfile to use golang:1.6 base image

ifraixedes:


bep:
This is the pedantic me, but if you use docker with Hugo you can also use `git commit --amend` to correct the commit message to be inline with the fairly simple guidelines.

ifraixedes:
@bep Thanks to point me to it.

I hope that it fits now to the code guidelines.

bep:
Merged in 0945673

PR #1888, opened 2016-02-22
ported Shekhar Gulati Hugo tutorial to Hugo's Official quickstart

shekhargulati:


digitalcraftsman:
Hello @shekhargulati,

thank you for contributing your great introduction guide to Hugo. Before merging:

You used the HTML `` tag. Since your post itself is already written in Markdown you could use it for the inclusion of image too with with the syntax `optional description when hovering over image`.

Furthermore, your images don't show up in the introduction. It's seems to be an issue with the image path. Since all images of [tutorials](https://github.com/spf13/hugo/tree/master/docs/static/img/tutorials) and those of the [showcase](https://github.com/spf13/hugo/tree/master/docs/static/img) are stored in the `static` folder, I would store your screenshots there too. Maybe at `docs/static/img/quickstart`.

@spf13: are you fine with the current version?

---

TODO for myself:

- [ ] compress / optimize the screenshots

shekhargulati:
@digitalcraftsman I have made changes as per your suggestion. The reason I used `` instead of Markdown format was to limit the image size. Let me know if you have any other feedback.

digitalcraftsman:
Many thanks so far.

Know I see the problem with the image size. We can fix that with a small CSS snippet in [style.css](https://github.com/spf13/hugo/blob/master/docs/static/css/style.css):

```css
img {
    max-width: 100%;
    height: auto;
}
```

The next problem is that @spf13 needs to update the cache of the website manually. This is at least the experience from earlier design changes.

Would it be possible to create cross-references with the [relref](http://gohugo.io/extras/crossreferences/) shortcode? Those are used all over the docs and make the links relative to the base url and not to the root of the server.

shekhargulati:
@digitalcraftsman pushed the changes suggested by you.

bep:
@digitalcraftsman could you merge this?

digitalcraftsman:
I could. But I don't know if we face an issue with the stylesheets. Cloudflare needs to update their cache. Otherwise the layout of the content is broken because the screenshots aren't aligned correctly and create a horinzontal scrollbar.

bep:
If that is a problem, then inline the single CSS-change temporarily.

PR #1899, opened 2016-02-27
Add template function cat

digitalcraftsman:
Cat returns the concatenated version of multiple
strings that are delimited by a separator.

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -575,6 +575,16 @@ CJK-like languages.
 ```
 
 
+### cat
+
+Like the eponymous UNIX-command, `cat` con**cat**enates multiple strings to a single one. The first parameter is always a separator that delimits the strings. The separator is followed by all strings that should be concatenated.
+
+```html
+{{ cat ", " "foo" "bar" "buzz" }}
+<!-- returns the string "foo, bar, buzz" -->
+```
+
+
 ## URLs
 
 ### absURL, relURL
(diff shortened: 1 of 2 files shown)


bep:
`cat` on Unix doesn't do this ... This looks like something we have already.

digitalcraftsman:
> This looks like something we have already.

Hugo provides a `delimit` function. But it requires a list of strings. Out of the box it's not possible to create a slice in Go's template language (as far as I know). Cases like [this one](https://discuss.gohugo.io/t/concatenating-strings-with-print/711?u=digitalcraftsman) could be handled with `cat`, since cat doesn't requires as slice of strings. Instead it takes the strings as single arguments.

> cat on Unix doesn't do this

Usually, cat concatenates the content of files in Unix and returns them to Stdout. I treated the file's content as string (as analogy). But we can change name / description if this PR isn't rejected.

moorereason:
Delimit can also take a map, so can you use dict on the first arg to delimit?

digitalcraftsman:
That's a clever workaround. As example:

    {{ delimit (dict "key1" "foo" "key2" "bar") ", " }}

You can achieve this already with built-in tools but you've to add add a key, which isn't used at all. You need to make the tradeoff between potential redundancy and (in my eyes) a more elegant use.

moorereason:
True: the `dict` requires useless, unique keys.

I would rather see us add a `slice` template func instead of this `cat` proposal:

```go
func slice(args ...interface{}) []interface{}
```
Use case:
```go
{{ delimit (slice "foo" "bar") ", " }}
```

digitalcraftsman:
That sounds like a good compromise. I'll close this issue and open a new one for your suggested `slice` function.

digitalcraftsman:
See #1902

PR #1913, opened 2016-03-02
docs: use https links in readme where appropriate

nikolas:


digitalcraftsman:
@anthonyfok you stated in #1729 that we should stick with `http` links for now. On the other hand Steve seems to tend to `https` as longterm approach.

> HTTPS will use http2 now so we should use that when possible going forward.

Since this is the main idea of this pull request I would like to know if we should switch fully to `https` (where possible) or not.

digitalcraftsman:
Hello @nikolas,

thank you for your improvement of the README. I've merged your contribution as 970f15cf7a6b4418a190b78895ad7d33b10e4d2b (master).


Cheers,
Digitalcraftsman

PR #1931, opened 2016-03-07
Added exec template function.

sba1:
A new template function called exec is added. This function can be used to
include the standard output of an external command. Commands that are
allowed to be executed must be listed within the newly introduced
execWhiteList setting.

This is very similar to PR #1548, especially the idea of the white list was
inspired by that request.

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -28,6 +28,7 @@ import (
 	"html/template"
 	"math/rand"
 	"os"
+	_exec "os/exec"
 	"reflect"
 	"sort"
 	"strconv"
@@ -40,6 +41,7 @@ import (
 	"github.com/spf13/cast"
 	"github.com/spf13/hugo/helpers"
 	jww "github.com/spf13/jwalterweatherman"
+	"github.com/spf13/viper"
 )
 
 var funcMap template.FuncMap
@@ -1137,6 +1139,35 @@ func highlight(in interface{}, lang, opts string) (template.HTML, error) {
 	return template.HTML(helpers.Highlight(html.UnescapeString(str), lang, opts)), nil
 }
 
+// exec executes a command and returns its output as string.
+func exec(in interface{}, args ...string) (string, error) {
+	str, err := cast.ToStringE(in)
+
+	if err != nil {
+		return "", err
+	}
+
+	// Check if command is accepted (in white list)
+	commandAccepted := false
+	for _, validCommand := range viper.GetStringSlice("ExecWhiteList") {
+		if str == validCommand {
+			commandAccepted = true
+			break
+		}
+	}
+
+	if !commandAccepted {
+		return "", fmt.Errorf("Executing %s is not allowed. Check execWhiteList settings.", str)
+	}
+
+	out, err := _exec.Command(strings.TrimSpace(str), args...).Output()
+	if err != nil {
+		return "", err
+	}
+
+	return string(out[:]), nil
+}
+
 var markdownTrimPrefix = []byte("<p>")
 var markdownTrimSuffix = []byte("</p>\n")
 
@@ -1684,6 +1715,7 @@ func init() {
 		"div":          func(a, b interface{}) (interface{}, error) { return doArithmetic(a, b, '/') },
 		"echoParam":    returnWhenSet,
 		"eq":           eq,
+		"exec":         exec,
 		"first":        first,
 		"ge":           ge,
 		"getCSV":       getCSV,
--- a/tpl/template_funcs_test.go
+++ b/tpl/template_funcs_test.go
@@ -58,6 +58,56 @@ func tstIsLt(tp tstCompareType) bool {
 	return tp == tstLt || tp == tstLe
 }
 
+func TestExecFuncInTemplate(t *testing.T) {
+
+	viper.Reset()
+	defer viper.Reset()
+
+	viper.Set("ExecWhitelist", []string{"echo"})
+
+	in := "{{exec \"echo\" \"test\"}}"
+	expected := "test\n"
+
+	templ, err := New().New("test").Parse(in)
+	if err != nil {
+		t.Fatal("Got error on parse", err)
+	}
+
+	var b bytes.Buffer
+	err = templ.Execute(&b, nil)
+
+	if err != nil {
+		t.Fatal("Got error on execute", err)
+	}
+
+	if b.String() != expected {
+		t.Errorf("Got\n%q\nExpected\n>%q<", b.String(), expected)
+	}
+}
+
+func TestExecFuncInTemplateCmdNotInWhitelist(t *testing.T) {
+
+	viper.Reset()
+	defer viper.Reset()
+
+	viper.Set("ExecWhitelist", []string{})
+
+	in := "{{exec \"echo\" \"test\"}}"
+	expectedErrSuffix := "Executing echo is not allowed. Check execWhiteList settings."
+
+	templ, err := New().New("test").Parse(in)
+	if err != nil {
+		t.Fatal("Got error on parse", err)
+	}
+
+	var b bytes.Buffer
+	err = templ.Execute(&b, nil)
+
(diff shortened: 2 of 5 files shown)


tpl/template_funcs.go
```diff
@@ -1132,6 +1134,35 @@ func highlight(in interface{}, lang, opts string) (template.HTML, error) {
...
+	commandAccepted := false
+	for _, validCommand := range viper.GetStringSlice("ExecWhiteList") {
+		if str == validCommand {
+			commandAccepted = true
+		}
+	}
+
+	if !commandAccepted {
+		jww.ERROR.Printf("Executing %s is not allowed. Check execWhiteList settings.", str)
+		return "", nil
```
bep: Return an error instead of logging it.

tpl/template_funcs.go
```diff
@@ -1132,6 +1134,35 @@ func highlight(in interface{}, lang, opts string) (template.HTML, error) {
...
+
+	if err != nil {
+		return "", err
+	}
+
+	// Check if command is accepted (in white list)
+	commandAccepted := false
+	for _, validCommand := range viper.GetStringSlice("ExecWhiteList") {
+		if str == validCommand {
+			commandAccepted = true
```
bep: Cosmetic, but a break would be nice here, as it is no need to look further.

tpl/template_funcs.go
sba1: >> +
>> +    if !commandAccepted {
>> +        jww.ERROR.Printf("Executing %s is not allowed. Check
> execWhiteList settings.", str)
>> +        return "", nil
> 
> Return an error instead of logging it.

Done. The test for the failure case looks a bit awkward now, let me know 
if it can be improved (never did anything with Go before).

tpl/template_funcs_test.go
```diff
@@ -57,6 +57,56 @@ func tstIsLt(tp tstCompareType) bool {
...
+
+func TestExecFuncInTemplateCmdNotInWhitelist(t *testing.T) {
+
+	viper.Reset()
+	defer viper.Reset()
+
+	viper.Set("ExecWhitelist", []string{})
+
+	in := "{{exec \"echo\" \"test\"}}"
+	expectedErrSuffix := fmt.Sprintf("Executing echo is not allowed. Check execWhiteList settings.")
```
moorereason: No need to use `fmt.Sprintf` here.

tpl/template_funcs.go
```diff
@@ -1132,6 +1134,35 @@ func highlight(in interface{}, lang, opts string) (template.HTML, error) {
 	return template.HTML(helpers.Highlight(html.UnescapeString(str), lang, opts)), nil
 }
 
+// execute a command and returns its output as string
```
moorereason: Use a proper godoc comment.  Start with the func name and use complete sentences with punctuation.

`// exec executes a command and returns its output as a string.`

tpl/template_funcs_test.go
```diff
@@ -57,6 +57,56 @@ func tstIsLt(tp tstCompareType) bool {
...
+	templ, err := New().New("test").Parse(in)
+	if err != nil {
+		t.Fatal("Got error on parse", err)
+	}
+
+	var b bytes.Buffer
+	err = templ.Execute(&b, nil)
+
+	if !strings.HasSuffix(err.Error(), expectedErrSuffix) {
+		t.Errorf("Expected suffix %v in %v", expectedErrSuffix, err.Error())
```
moorereason: Why are you using `%v` here?  I think `%q` would work better.

tpl/template_funcs.go
sba1: >> @@ -1132,6 +1134,35 @@ func highlight(in interface{}, lang, opts
> string) (template.HTML, error) {
>>      return
> template.HTML(helpers.Highlight(html.UnescapeString(str), lang,
> opts)), nil
>>  }
>> 
>> +// execute a command and returns its output as string
> 
> Use a proper godoc comment. Start with the func name and use complete
> sentences with punctuation.
> 
> // exec executes a command and returns its output as a string.

In my opinion, sentences starting with a small letter don't look right. 
In this case, I would rather like to write

```
// Function exec executes a command and returns its output as a string.
````

But as I haven't found this in other places I modified it as you 
suggested.

I've also implemented all the other suggestions. Thanks for your help.

moorereason:
lgtm

CircleCI is dumb.

digitalcraftsman:
Does this pull request replace #847?

sba1:
Am 2016-03-12 12:41, schrieb digitalcraftsman:
> Does this pull request replace #847 [1]?

Seems so. At least in parts. It doesn't add the short code yet (though I 
have this in my local branch).

bep:
As neither of these solves the fundamental security issue, it doesn't really matter.

sba1:
> As neither of these solves the fundamental security issue, it doesn't
> really matter.

In fact, I was not aware of PR #847 and the discussion when creating 
this PR, however, as in the final version of this request (which was not 
responded though), a white list config string for the global 
configuration was introduced following your suggestion. This is also the 
case in this PR as described in the doc and commit message. If this is 
not correct (in fact my knowledge about Hugo internals is little) I'm 
open for suggestions.

For instance, another possibility would be to add a new command line 
parameter to ```huhu``` enable this feature.

bep:
@digitalcraftsman suggested in another issue that we add a `savetofile` template func. Combina that with `exec` and you will have the perfect platform for sophisticated Trojan horses. 

A whitelist and a --unsafe flag is the minimum here. But that is, in my eyes, still not enough. People make mistake ans suddenly the unsafe boolean is switched and the whitelist check could be bypassed by terminating the string with an emoji :-) ...

digitalcraftsman:
Especially in this combination I can understand your concerns.

sba1:
> @digitalcraftsman [1] suggested in another issue that we add a
> savetofile template func. Combina that with exec and you will have the
> perfect platform for sophisticated Trojan horses.

Actually, one of my wishes that I can directly embed some code (in some 
programming language) which is compiled, executed, its output is then 
embedded in the final output (so this exec is only the first step). Like 
in report generation such as in knitr or pweave, but initially much 
simpler.

> A whitelist and a --unsafe flag is the minimum here. But that is, in
> my eyes, still not enough. People make mistake ans suddenly the unsafe
> boolean is switched and the whitelist check could be bypassed by
> terminating the string with an emoji :-) ...

I would use hugo only as a static site generator. I agree that the 
feature can be troublesome when one would use "hugo server" in 
production. Is this the intended use case for hugo? Generally, I 
wouldn't mind if this exec feature would simply not work for "hugo 
server".

bep:
Unless someone has some really enlightening to add, I suggest we close this and related issues/PRs -- and focus on what we can do. A cross-platform `cat` (file reader) would be a good start.

sba1:
> Unless someone has some really enlightening to add, I suggest we close
> this and related issues/PRs -- and focus on what we can do. A
> cross-platform cat (file reader) would be a good start.

Of course, it is all your choice.

But what is with the suggestion to disable the feature for the "hugo 
server" option? With this question I'm just trying to understand what 
the exact security concerns are.

For me, hugo appears to be a static web site generator, so it converts 
some input files to some output files similar to a compiler ahead of 
time, i.e., there is usually a separate publishing step. If I understood 
it correctly, your concern is that one could do everything with an exec 
feature, which is of course right as it was the purpose of such a 
command (though the feature has to be explicitly enabled and the user 
has every control over it). This is essentially a feature other 
programming language offers (also e.g. in LaTeX).

Also, I've just learned about the "external helpers" concept of hugo, 
which means that hugo already calls external (predefined) programs. 
Apparently, some flexibility is useful (and because one doesn't have 
control over these tools one also cannot say that they never do 
potentially bad things).

bep:
http://discuss.gohugo.io/ is a better place for discussions.

bep:
I'm closing this now. We may revisit this in the future. Maybe.

PR #1932, opened 2016-03-07
Add md5 and sha1 template functions

### LINKED ISSUE #1910: Add hash template functions
It will be nice to have ability to hash content on template.

Two possibilities:

1. A generic hash function `{{ hash <CONTENT> <ALGORITHM> }}`
2. Common algorithm hash function pipeline `{{ .Site.Params.my_var | md5 }}`

Good candidates should be:

- `md5` 
- `sha1`
- `sha2`

--- 

*Use case*: generate *Gravatar* url from `email`

```
<meta property="og:image" content="http://www.gravatar.com/avatar/{{ .Site.Params.gravatar.email | md5 }}">
<meta property="og:image" content="http://www.gravatar.com/avatar/{{ hash .Site.Params.gravatar.email "md5" }}">
```

digitalcraftsman:
Fixes #1910.

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -595,6 +595,33 @@ CJK-like languages.
 ```
 
 
+### md5
+
+`md5` hashes the given input and returns its MD5 checksum.
+
+```html
+{{ md5 "Hello world, gophers!" }}
+<!-- returns the string "b3029f756f98f79e7f1b7f1d1f0dd53b" -->
+```
+
+This can be useful if you want to use Gravatar for generating a unique avatar:
+
+```html
+<img src="https://www.gravatar.com/avatar/{{ md5 "your@email.com" }}?s=100&d=identicon">
+```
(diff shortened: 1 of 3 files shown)


digitalcraftsman:
@bep since the template functions' name would conflict with the corresponding package names I added an `_` as [prefix](https://github.com/digitalcraftsman/hugo/blob/tplfuncs/hashing/tpl/template_funcs.go#L18-L19). Let me know if you're not happy with the way how the conflict is handled.

bep:
>an _ as prefix. 

That's fine. But you might add the error case to the tests (see the expect false interface{} construct used elsewhere).

Also add the two functions to the "smoke test at the top of the file".

digitalcraftsman:
I updated the test file accordingly.

bep:
Merged in 94c3825

digitalcraftsman:
Thanks.

PR #1938, opened 2016-03-08
Use default sitemap configuration for homepage (Fixes #1304)

### LINKED ISSUE #1304: Sitemap.xml forces homepage to priority zero, even if a different default is set
See http://gohugo.io/sitemap.xml for an example:

```
<url>
<loc>http://gohugo.io/</loc>
<lastmod>2015-07-21T10:00:00+00:00</lastmod>
<priority>0</priority>
</url>
```

The homepage priority is always 0, even if a different value is set in the config file, e.g.:

```
[sitemap]
  priority = "0.5"
```

(As an aside, the [sitemap](https://github.com/spf13/hugo/blob/master/hugolib/sitemap.go) config setting is not documented anywhere, the configuration page mentions it but doesn't define it).

s7anley:


bep:
Please put the "Fixes ..." in the commit message body and without the parens. See guideline in README.

bep:
Merged in be3519f

s7anley:
@bep sorry, wasn't on my laptop yesterday to fix it. I will do it properly
next time :)

2016-03-09 10:09 GMT+01:00 Bjørn Erik Pedersen <notifications@github.com>:

PR #1943, opened 2016-03-09
tpl: Add default function

moorereason:


digitalcraftsman:
:+1:

moorereason:
Hold up.  I found an issue that I need to fix.

digitalcraftsman:
I didn't planned to merge anything myself :smile:

tpl/template_funcs_test.go
```diff
@@ -1839,6 +1840,44 @@ func TestDateFormat(t *testing.T) {
 	}
 }
 
+func TestDefault(t *testing.T) {
+	for i, this := range []struct {
+		dflt     interface{}
+		given    interface{}
```
bep: Need nil-variants for these.

docs/content/templates/functions.md
```diff
@@ -28,6 +28,17 @@ and other basic tools; these are listed in the
 
 ## General
 
+### default
+Checks whether a given value is set and returns a default value if it is not.
+"Set" in this context means true for booleans; non-zero for numeric types;
+non-zero length for strings, arrays, slices, and maps; any struct value; or
+non-nil for any other types.
```
bep: I think you should add Zero for time.Time (time.IsZero) to the list of not set.

moorereason: Good idea.

moorereason: Fixed.

tpl/template_funcs_test.go
moorereason: Fixed.

moorereason:
I had to rework the func signature to use a variadic parameter for `given`; otherwise, the following construct would given an error about argument count mismatch when the key was undefined:

```
{{ index . "key" | default "foo" }}
```
Apparently, the Go template engine doesn't pass a nil(?) value through the pipe.

I can rebase and squash if this looks good to everyone.

bep:
Looks mostly good, but I have one question:

```
{map[string]interface{}{"images": []string{}}, `{{ default "default.jpg" (index .images 0) }}`, `default.jpg`, true},
```

The above testcase fails. And while I understand why it fails, it is confusing for the users that this works for maps, but not for slices. I use @derekperkins "list of images" syntax, and then have a rather clumsy construct to pick the first if found. I would rather use `default`.

bep:
On an added note: I don't expect you to "hook into" the Golang internals to get this to work. I don't see how. But I'm just thinking out loud: If we could override the `index` func somehow, I think it would be better if it returned `nil` and not `index out of range` error in this case.

moorereason:
@bep,
I have an overloaded `index` staged in the branch used by this PR.  It works with your test case above.  Can I push it with this PR or do you want a separate PR?

bep:
@moorereason push it with this PR, add my test case + reference the issue I just created.

bep:
But: When "overloading" the index func I would prefer if we copied the entire original func, if possible, maded it clear that the origin is Go; then just comment out the "index out of range" and return nil. Then it is clear to those who come after us that this is "almost the same" as index.

moorereason:
Already pushed.  :smile:  I did pretty much everything you mention except I didn't leave the original code there as a comment; however, I did leave a comment about the deviation from stdlib and reference where I copied it from (there were 3 funcs total).

bep:
It looks good. It just have to obvious what to replace/change if they fix this at some point int the stdlib. I will have a look at and merge this later.

bep:
Merged in 2d0650d Good work.

rdwatters:
👍🏻👍🏻👍🏻

moorereason:
@bep, this PR was never merged.  What was merged was a custom `index` implementation that makes the `default` implementation more useful.

Reopening and will modify the treatment of handling of booleans based on #1953.

moorereason:
Nevermind, this was merged.  I should sleep more.

/me walks away in shame.

PR #1945, opened 2016-03-10
tpl: Add `intersect` operator to `where` template function

christophermancini:
The main reason I added this operator was to filter `.Site.Pages` to a list of related pages based on taxonomy. Doing this in the template side is wordy and inefficient, and prevents you from being able to limit your results using the `first` template function.

My 1200 page site went from 30sec to build down to 9 by implementing this new feature.

Before this change:

```
        {{ if .Params.tags }}
          <h4>Related Articles</h4>
          <div id="related-articles" class="row-fluid">
          {{ $page_link := .Permalink }}
          {{ $tags := .Params.tags }}
          {{ range $i, $page := .Site.Pages }}
              {{ $has_common_tags := intersect $tags $page.Params.tags | len | lt 0 }}
              {{ if and $has_common_tags (ne $page_link $page.Permalink) }}
                {{ $page.Render "li" }}
              {{ end }}
            {{ end }}
          {{ end }}
          </div>
          <div class="clearfix"></div>
        {{ end }}
```

After this change:

```
{{ if .Params.tags }}
  <h4>Related Articles</h4>
  <div id="related-articles" class="row-fluid">
  {{ range first 20 (where .Site.Pages ".Params.tags" "intersects" .Params.tags) }}
    {{ if ne $.Permalink .Permalink }}
      {{ .Render "li" }}
    {{ end }}
  {{ end }}
  </div>
  <div class="clearfix"></div>
{{ end }}

```

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -67,7 +67,7 @@ Creates a dictionary `(map[string, interface{})`, expects parameters added in va
 Invalid combinations like keys that are not strings or uneven number of parameters, will result in an exception thrown.
 Useful for passing maps to partials when adding to a template.
 
-e.g. Pass into "foo.html" a map with the keys "important, content" 
+e.g. Pass into "foo.html" a map with the keys "important, content"
 
     {{$important := .Site.Params.SomethingImportant }}
     {{range .Site.Params.Bar}}
@@ -78,9 +78,8 @@ e.g. Pass into "foo.html" a map with the keys "important, content"
 
     Important {{.important}}
     {{.content}}
-    
 
-or create a map on the fly to pass into 
+or create a map on the fly to pass into
 
     {{partial "foo" (dict "important" "Smiles" "content" "You should do more")}}
 
@@ -313,6 +312,15 @@ Following operators are now available
 - `<`, `lt`: True if a given field value is lesser than a matching value
 - `in`: True if a given field value is included in a matching value. A matching value must be an array or a slice
 - `not in`: True if a given field value isn't included in a matching value. A matching value must be an array or a slice
+- `intersect`: True if a given field value that is a slice / array of strings or integers contains elements in common with the matching value. It follows the same rules as the intersect function.
+
+*`intersect` operator, e.g.:*
+
+    {{ range where .Site.Pages ".Params.tags" "intersect" .Params.tags }}
+      {{ if ne .Permalink $.Permalink }}
+        {{ .Render "summary" }}
+      {{ end }}
+    {{ end }}
 
 *`where` and `first` can be stacked, e.g.:*
 
@@ -340,7 +348,7 @@ e.g.
 
 ### readDir
 
-Gets a directory listing from a directory relative to the current project working dir. 
+Gets a directory listing from a directory relative to the current project working dir.
 
 So, If the project working dir has a single file named `README.txt`:
 
@@ -349,7 +357,7 @@ So, If the project working dir has a single file named `README.txt`:
 ### readFile
 Reads a file from disk and converts it into a string. Note that the filename must be relative to the current project working dir.
  So, if you have a file with the name `README.txt` in the root of your project with the content `Hugo Rocks!`:
- 
+
  `{{readFile "README.txt"}}` → `"Hugo Rocks!"`
 
 ## Math
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -646,6 +646,7 @@ func checkCondition(v, mv reflect.Value, op string) (bool, error) {
 
 	var ivp, imvp *int64
 	var svp, smvp *string
+	var slv, slmv interface{}
 	var ima []int64
 	var sma []string
 	if mv.Type() == v.Type() {
@@ -668,6 +669,9 @@ func checkCondition(v, mv reflect.Value, op string) (bool, error) {
 				imv := toTimeUnix(mv)
 				imvp = &imv
 			}
+		case reflect.Array, reflect.Slice:
+			slv = v.Interface()
+			slmv = mv.Interface()
 		}
 	} else {
 		if mv.Kind() != reflect.Array && mv.Kind() != reflect.Slice {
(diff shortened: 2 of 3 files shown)


bep:
Looks mostly good to me:

* Look at the commit message guidelines
* Needs a line and an example in the documentation
* English isn't my first language, but shouldn't it be "intersect"? "I want all the pages where the tags intersect these tags"

What do you think @tatsushid ?

christophermancini:
@bep Ok, so I added it to docs and I attempted to rebase -i, but when I force push, it won't get rid of the original commit. I tried deleting remote branch and repushing, but all it did was close my PR, it still kept the original commit. Any help here would be appreciated. The only thing I can think to do is create a new PR from my master that has the properly squashed commit.

W.R.T. intersect vs intersects, I looked up the definition and they used intersects in most of their examples and I think it flows nicely in code, e.g.:

`occasionally the water table intersects the earth's surface, forming streams and lakes`

`where .Site.Pages ".Params.tags" intersects .Params.tags`

However, if others think that it should be singular instead, I am fine with changing it.

bep:
It should be `intersect`, he intersects, they intersect.

As to rebase and push -i; i do this all the time, so if it doesn't work for you, you must do something wrong, but It is impossible to know what.

christophermancini:
I will change it to intersect and attempt to get it consolidated into a single commit again.

christophermancini:
@bep fixed.

docs/content/templates/functions.md
```diff
@@ -302,6 +302,15 @@ Following operators are now available
 - `<`, `lt`: True if a given field value is lesser than a matching value
 - `in`: True if a given field value is included in a matching value. A matching value must be an array or a slice
 - `not in`: True if a given field value isn't included in a matching value. A matching value must be an array or a slice
+- `intersect`: True if a given field value that is a slice / array of strings or integers contains elements in common with the matching value. It follows the same rules as the intersect function.
+
+*`intersect` operator, e.g.:*
+
+    {{ range where .Site.Pages ".Params.tags" "intersect" .Params.tags }}
+      {{ if ne .Permalink $.Permalink }}
```
bep: I'm pretty sure you can do `ne . $.` and compare objects. Even better would be to add this to the where above (I believe you would have to do a double where) -- then it can be used in a paginator ...

tpl/template_funcs.go
```diff
@@ -739,6 +743,25 @@ func checkCondition(v, mv reflect.Value, op string) (bool, error) {
...
+		if err != nil {
+			return false, err
+		}
+		switch r.(type) {
+		case []string:
+			if len(r.([]string)) > 0 {
+				return true, nil
+			}
+			return false, nil
+		case []int:
```
bep: This is too restrictive, I think. There are are int64 and all the rest of the family.

christophermancini: Since the source data would be in toml, is it possible for there to be a different type of int?

docs/content/templates/functions.md
christophermancini: I think placing it within the where above would be too expensive because you would be doing `for n` where it is only possible to have 1 match.

tpl/template_funcs.go
moorereason: We need to handle data coming from anywhere, not just TOML front-matter.  You need to be prepared to handle any type that `intersect` could return, which currently includes strings and all of the int and float types.

christophermancini: Fair enough, I will update it.

christophermancini: @moorereason  @bep I split out support for the individual int and float data types as well as added a test for the floats.

tpl/template_funcs_test.go
```diff
@@ -1211,6 +1212,33 @@ func TestWhere(t *testing.T) {
...
+			key: "b", op: "intersect", match: []int{4, 10, 12},
+			expect: []map[string][]int{
+				{"a": []int{1, 2, 3}, "b": []int{4, 5, 6}}, {"a": []int{7, 8, 9}, "b": []int{10, 11, 12}},
+			},
+		},
+		{
+			sequence: []map[string][]float32{
+				{"a": []float32{1.0, 2.0, 3.0}, "b": []float32{4.0, 5.0, 6.0}}, {"a": []float32{7.0, 8.0, 9.0}, "b": []float32{10.0, 11.0, 12.0}}, {"a": []float32{13.0, 14.0, 15.0}, "b": []float32{16.0, 17.0, 18.0}},
+			},
+			key: "b", op: "intersects", match: []float32{4, 10, 12},
```
moorereason: s/intersects/intersect/

Be sure to run tests.

moorereason:
Try using a commit message like this:

```
tpl: Add intersect operator to where function

Returns true if a given field value that is a slice / array of strings or integers contains elements in common with the matching value. It follows the same rules as the intersect function.
```

moorereason:
Also, please add test cases for each type.

Thanks for the contribution.  You're doing great. :+1:

tpl/template_funcs_test.go
christophermancini: My bad, I had started modifying from another machine that was out of date and didn't catch it. I will fix it.

christophermancini:
@moorereason updated with full data type test coverage, expanded commit message and fixed typo.

bep:
... that long switch looks kind of ugly when all you want is to get if the length of a slice is > 0. The reflect package should answer that in fewer lines, and would also support the unlikely case that we'll get a slice of `uint` ...

christophermancini:
@bep agreed that it looks ugly. I will give the reflect package a shot and update the code.

bep:
Any progress on the switch?

christophermancini:
@bep Sorry I have been super busy lately. I will try to get it done by Monday.

bep:
No problem.

christophermancini:
@bep I rebased the branch on master, swapped the huge switch with reflect code and re-squashed the commit.

christophermancini:
@bep bump. :)

Issue #1290, opened 2015-07-20
url in front matter should allow + signs to be passed on to permalinks


jacquesmattheij:
A url like '/this+is+a+test' in the front matter ends up as /thisisatest in .Permalink, but the page is generated in the right spot which makes any link(s) in the rest of the site to that page unclickable. 

The + sign is a legal character in a url and should be preserved wherever it is used.

update: it seems this also goes for the % sign (which I thought I might use as a work-around by encoding the + sign...).

PR #1966, opened 2016-03-12
Add '+' as one of the valid characters in urls specified in the front matter

egonSchiele:
The UnicodeSanitize function contains a list of valid characters, `+` should be one of those chars.

### DIFF
--- a/helpers/path.go
+++ b/helpers/path.go
@@ -96,7 +96,7 @@ func UnicodeSanitize(s string) string {
 	target := make([]rune, 0, len(source))
 
 	for _, r := range source {
-		if unicode.IsLetter(r) || unicode.IsDigit(r) || unicode.IsMark(r) || r == '%' || r == '.' || r == '/' || r == '\\' || r == '_' || r == '-' || r == '#' {
+		if unicode.IsLetter(r) || unicode.IsDigit(r) || unicode.IsMark(r) || r == '%' || r == '.' || r == '/' || r == '\\' || r == '_' || r == '-' || r == '#' || r == '+' {
 			target = append(target, r)
 		}
 	}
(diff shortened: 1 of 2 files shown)


bep:
Why should it be one of the valid characters?

egonSchiele:
It is one of the characters that is valid in urls http://stackoverflow.com/a/1547940

bep:
Hugo have its own restrictions. Thanks for your report.

spf13:
@egonSchiele Not all of the characters in that list are acceptable in commonly used filesystems. Hugo will only permit characters that are both acceptable in all filesystems and valid URL characters. Additionally we avoid characters that may be filesystem acceptable but will interfere with programs (finder, explorer, ls, dir, etc) which are more particular. 

We did end up adding the "%" character even though it's an issue on windows due to its extensive use in URLs.

Is there a similarly valid use case for allowing the "+"?

egonSchiele:
I don't know, @spf13 . To be honest, this is not an issue for me...I was merely trying to help by fixing issue #1290. Since it was marked `status/please contribute`, I assumed y'all did want this change and were merely looking for a PR.

spf13:
Aah.. Thanks for providing the background. We appreciate the fix. I did some diligence and I can't find a reason to prohibit this character. It's fine on linux, mac and windows.. even back to Win95. 

I'm fine with merging unless @bep has a reason against it.

bep:
No reason, please merge.

spf13:
merged as c42982f

Thanks @egonSchiele. Sorry this one took a while to merge in. Looking forward to further contributions.

egonSchiele:
Happy to help @spf13 !

PR #1967, opened 2016-03-13
Check for duplicate permalinks and throw an error (Fixes #1867)

### LINKED ISSUE #1867: Hugo permits conflicting content URLs
On Hugo v0.15 two or more pieces of content can resolve to the same URL, and one overwrites the other. There are many ways this can happen: for example if one specifies a `slug` that matches the filename of the other content (which does not specify a `slug`).

In this scenario it would be helpful if Hugo emitted a warning or perhaps an error. There might be a scenario where this is a legitimate feature, but I can't think of one. Let me know if there is :-)

Thank you

egonSchiele:
Here's one way issue #1867 could be fixed: by checking for duplicate permalinks after all the pages have been created.

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -713,6 +713,11 @@ func (s *Site) Process() (err error) {
 	if err = s.CreatePages(); err != nil {
 		return
 	}
+
+	if err = s.CheckPermalinks(); err != nil {
+		return
+	}
+
 	s.setupPrevNext()
 	if err = s.BuildSiteMeta(); err != nil {
 		return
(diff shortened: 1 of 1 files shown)


hugolib/site.go
```diff
@@ -955,6 +960,29 @@ func (s *Site) ConvertSource() chan error {
 	return errs
 }
 
+func (s *Site) CheckPermalinks() error {
+	permalinks := make(map[string]string)
+	for _, page := range []*Page(s.Pages) {
```
bep: for page := range s.Pages should be simpler.

hugolib/site.go
```diff
@@ -955,6 +960,29 @@ func (s *Site) ConvertSource() chan error {
...
+			return err
+		}
+
+		// the ending slash shouldn't matter
+		if permalink[len(permalink)-1] == '/' {
+			permalink = permalink[0 : len(permalink)-1]
+		}
+
+		if otherPage, exists := permalinks[permalink]; exists {
+			return fmt.Errorf("Pages %s and %s use the same permalink: %s", page.FullFilePath(), otherPage, permalink)
```
bep: A jww.ERROR would be better, me thinks, esp. considering the server mode.

bep:
@egonSchiele thanks for this suggestion; this works, but I think it would be simpler and more effective to track this on the rendering sitde of it.

Issue #1952, opened 2016-03-10
tpl: intersect needs some love


moorereason:
The following test cases fail when added to `tpl/template_funcs_test.go:TestIntersect`:

``` go
{[]string{"a", "b", "c"}, []string{"a", "b", "b"}, []string{"a", "b"}},
{[]interface{}{"a", "b", "c"}, []interface{}{"a", "b", "b"}, []interface{}{"a", "b"}},
```

Test results:

``` text
--- FAIL: TestIntersect (0.00s)
        template_funcs_test.go:791: [0] got [a b b] but expected [a b]
        template_funcs_test.go:791: [1] got [] but expected [a b]
```

Regarding test 0, `in` is used by `intersect` to avoid adding duplicates to the new set.  In this case, `in` is receiving a `reflect.Value` struct, but the `reflect.Struct` case is not handled by `in`.

Regarding test 1, `intersect` does not handle the `reflect.Interface` case.  See [Questions about "intersect" function](https://discuss.gohugo.io/t/questions-about-intersect-function/2877/6) forum discussion.

Tests were done in Go 1.6 and Hugo HEAD.

bep:
Regarding this and all the reflective code in that file: It should be more consolidated, if possible (there are many common code paths there, with where, intersect, in, ...)

robertbasic:
@moorereason @bep is this issue addressed? (I'm looking for things to contribute to, so just want to avoid working on already fixed stuff).

moorereason:
@robertbasic,
The first test above is resolved (#1972).

I have a fix for the second test.  It's in one of my personal branches (8e81d2e5dd93d995ae281b9353b3815bad2df2a5).  However, there's still a piece missing, which this test will hit:

```go
{[]string{"a", "b", "c"}, []interface{}{"a", "b", "b"}, []string{"a", "b"}},
```
If you want to take my branch as a starting point and finish it out, go for it!  I probably won't have time to finish it for a few days.

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
[...]

PR #1972, opened 2016-03-13
tpl: Send actual values to in from intersect

moorereason:
The `intersect` function uses `in` to avoid adding duplicates to the
resulting set.  We were passing `reflect.Value` items when we should
have been using `Value.Interface()` to send the actual data structure.
This fixes that.

See #1952

### DIFF
--- a/tpl/template_funcs_test.go
+++ b/tpl/template_funcs_test.go
@@ -766,7 +766,7 @@ func TestIntersect(t *testing.T) {
 		sequence2 interface{}
 		expect    interface{}
 	}{
-		{[]string{"a", "b", "c"}, []string{"a", "b"}, []string{"a", "b"}},
+		{[]string{"a", "b", "c", "c"}, []string{"a", "b", "b"}, []string{"a", "b"}},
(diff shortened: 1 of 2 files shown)


tpl/template_funcs_test.go
```diff
@@ -779,6 +779,7 @@ func TestIntersect(t *testing.T) {
 		{[]int{2, 4}, []int{1, 2, 4}, []int{2, 4}},
 		{[]int{1, 2, 4}, []int{3, 6}, []int{}},
 		{[]float64{2.2, 4.4}, []float64{1.1, 2.2, 4.4}, []float64{2.2, 4.4}},
+		// {[]interface{}{"a", "b", "c"}, []interface{}{"a", "b", "b"}, []interface{}{"a", "b"}}, // #1952
```
bep: ??

moorereason: Okay,  I'll remove it.  :smile:  That was a remind of the other issue reported in #1952.

moorereason: Done

bep:
Merged in df44b19

PR #1979, opened 2016-03-15
bufferpool: Check returned error in test

moorereason:
Check the error returned by `buff.WriteString` plus trivial little
cleanups.

### DIFF
--- a/bufferpool/bufpool_test.go
+++ b/bufferpool/bufpool_test.go
@@ -1,4 +1,4 @@
-// Copyright 2016-present The Hugo Authors. All rights reserved.
+// Copyright 2016 The Hugo Authors. All rights reserved.
 //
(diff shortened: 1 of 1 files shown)


bufferpool/bufpool_test.go
```diff
@@ -14,14 +14,21 @@
...
+	"github.com/stretchr/testify/assert"
 )
 
 func TestBufferPool(t *testing.T) {
+	v := "do be do be do"
+
 	buff := GetBuffer()
-	buff.WriteString("do be do be do")
-	assert.Equal(t, "do be do be do", buff.String())
+	_, err := buff.WriteString(v)
```
bep: WriteString will never return an error; and if it had ever returned an error, the content assertion would faill.

moorereason:
Fair enough.

PR #1983, opened 2016-03-16
tpl: Treat booleans as set in default function

moorereason:
Booleans and `default` don't really make sense together, so we'll always treat
booleans as "set" and return the given value.


bep:
Makes sense. Merged in b0326a1

Issue #1950, opened 2016-03-10
Always set Markup variable in Page struct


ypnos:
Theme behavior may depend on the markup processor used. An example is custom CSS for markdown tables that would ruin asciidoc styling.

Right now, the `Page` struct has a `Markup` variable. It can be set explicitely to override markup-type guessing. However, if the type is guessed, it remains unset, therefore being unavailable to the theme.

This would be a place where the  variable could be set:
https://github.com/spf13/hugo/blob/924028a9be4e468b38f3b7a530253c7c09a0d10e/hugolib/page.go#L813

However that might be an unexpected side-effect. I would suggest to rename `guessMarkupType` in `determineMarkupType`, then call it explicitely as an initialization step before rendering and always use `Page.Markup` when rendering.

bep:
>However that might be an unexpected side-effect. 

Yea, not really. It is fairly controlled: One could argue that Page.Markup is a cached version ...

PR #1999, opened 2016-03-20
Save auto-detected markup type in Page.Markup

ypnos:
Always make markup type available to the theme.
Solution to issue #1950.

Also renames guessMarkupType() to determineMarkupType() as discussed in #1950.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -263,7 +263,7 @@ func (p *Page) renderBytes(content []byte) []byte {
 		}
 	}
 	return helpers.RenderBytes(
-		&helpers.RenderingContext{Content: content, PageFmt: p.guessMarkupType(),
+		&helpers.RenderingContext{Content: content, PageFmt: p.determineMarkupType(),
 			DocumentID: p.UniqueID(), Config: p.getRenderingConfig(), LinkResolver: fn, FileResolver: fileFn})
 }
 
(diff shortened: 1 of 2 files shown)


hugolib/page.go
```diff
@@ -800,16 +800,17 @@ func (p *Page) Render(layout ...string) template.HTML {
...
 	// First try the explicitly set markup from the frontmatter
 	if p.Markup != "" {
-		format := helpers.GuessType(p.Markup)
-		if format != "unknown" {
-			return format
-		}
+		p.Markup = helpers.GuessType(p.Markup)
+	}
+
+	if p.Markup == "unknown" {
```
bep: This can be pushed up inside the previous conditional.

bep:
Can you squash the commits into one -- and add a commit message in line with the guidelines in the README.

bep:
There is a breaking shortcode test. I don't understand why it's broken ...

hugolib/page.go
```diff
@@ -800,16 +800,16 @@ func (p *Page) Render(layout ...string) template.HTML {
...
-		if format != "unknown" {
-			return format
+		p.Markup = helpers.GuessType(p.Markup)
+		if p.Markup == "unknown" {
+			p.Markup = helpers.GuessType(p.Source.Ext())
 		}
 	}
 
-	return helpers.GuessType(p.Source.Ext())
+	return p.Markup
```
moorereason: This line is wrong.  If `p.Markup == ""`, you return `""` instead of `"unknown"`, which [shortcodes are not prepared for](https://github.com/spf13/hugo/blob/a89035bdaaa8bb1525a74d82e068ef80bfa28aed/hugolib/shortcode.go#L254).  Follow the original logic:

```go
	// First try the explicitly set markup from the frontmatter
	if p.Markup != "" {
		p.Markup = helpers.GuessType(p.Markup)
		if p.Markup != "unknown" {
			return p.Markup
		}
	}

	p.Markup = helpers.GuessType(p.Source.Ext())
	return p.Markup
```

ypnos:
Sorry that code was plain wrong. Thank you @moorereason for your analysis!

New solution builds on the fact that `helpers.GuessType(p.Markup)` returns 'unknown' as a default (i.e. also on empty input string).

bep:
Merged in 9f6b5ad. Thanks.

PR #2002, opened 2016-03-21
hugolib: Use field keys in composite literals

moorereason:
Small fix in shortcode tests to use field keys in a composite literal.


bep:
Merged.

PR #2003, opened 2016-03-21
hugo: Add check tasks to Makefile

moorereason:
Add some helpful tasks to the Makefile to make it easier to test code quality
prior to committing.  Use `make check` to run the standard checks.  As of now,
the cyclo and lint tasks are not part of the standard checks since we fail
miserably on those.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ COMMIT_HASH=`git rev-parse --short HEAD 2>/dev/null`
 BUILD_DATE=`date +%FT%T%z`
 LDFLAGS=-ldflags "-X github.com/spf13/hugo/hugolib.CommitHash=${COMMIT_HASH} -X github.com/spf13/hugo/hugolib.BuildDate=${BUILD_DATE}"
 
+DIRS=$(shell go list -f {{.Dir}} ./...)
+
 all: gitinfo
 
 install: install-gitinfo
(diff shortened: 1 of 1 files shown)


bep:
>fail miserably on those.

That is a said perspective. These `go vet` fixes are useful in the bigger picture of things -- but we have many open and unresolved defects that I would say is a few miles higher up on the priority ladder.

moorereason:
I agree with you, but as a newcomer that's not a programmer by trade, I can vouch for the helpfulness of lower cyclomatic complexity and good godoc-style comments (golint).  I realize we have more pressing things to deal with, but adding golint and gocyclo to the default checks is a long term goal of mine.  At this point, I just wanted to get this Makefile pushed in so that we can tell contributors to run `make check` prior to submitting a PR and possibly add a step to the CI tools to fail if the checks fail.

bep:
Just so we agree that "failing" a few lint requirements (which in some cases are highly opinionated) isn't making us *fail miserably*. That wold also put Go's stdlib into this failing category, and that is the cleanest code I've seen in a while.

moorereason:
I amended the commit message to cut us some slack.  :smile:

bep:
Merged in e8cf8b40dfd93f84fbbd63b57295e2e39e828d19 ... and now `make check` actually passes ...

PR #2006, opened 2016-03-21
filesystem: Consolidate if conditions

srinivasreddy:
`go fmt` has been run.


moorereason:
@srinivasreddy, please fix the commit subject.  The prefix should be `source`, not `filesystem`.

We're working on clarifying this in the contributor docs, but the prefix is the **Go package** name.

srinivasreddy:
@moorereason  done. thanks

bep:
Merged in b319c18

PR #2017, opened 2016-03-24
commands: Add quiet mode

### LINKED ISSUE #1218: Quiet mode
Hi. Many apps have a quiet mode `-q` or `--quiet` that only prints in case of error. How about a quiet mode for Hugo? I'd find this useful for a cronjob.

digitalcraftsman:
Does this flag require some sort of testing?

Fixes #1218

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -137,6 +137,7 @@ var (
 	noTimes               bool
 	pluralizeListTitles   bool
 	preserveTaxonomyNames bool
+	quiet                 bool
 	renderToMemory        bool // for benchmark testing
 	uglyURLs              bool
 	verbose               bool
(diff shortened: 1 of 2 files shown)


bep:
>Does this flag require some sort of testing?

Only manual.

digitalcraftsman:
> How quiet is this?

Hugo will show no error messages, e.g. if you include a partial that doesn't exist or your call a template function with a wrong number of arguments.

> And fatal isn't the same as error.

Which of JWW's [logging level](https://godoc.org/github.com/spf13/jWalterWeatherman#pkg-variables) should I use instead? It's need have a higher priority than `ERROR`.

bep:
As I see it quiet mode should only show ERROR or FATAL -- nothing else, not even the stats.

bep:
What is the status on this one?

digitalcraftsman:
I'll work next on it. Quick question: arcoding to [jww's docs](https://godoc.org/github.com/spf13/jWalterWeatherman#Feedback) `jww.FEEDBACK` seems to write to the console regardless which logging level is set. 

For example should Hugo show that the [web server](https://github.com/spf13/hugo/blob/c2277fcbc1f7f38c9f990dd65c9dd2f24b7703a5/commands/server.go#L210) is available? Should such messages also suppressed?

bep:
I don't thinks this is a simple task, probably not worth it.

`quiet` means .... quiet, so nothing to stdout or stderr that isn't prefixed with ERROR or FATAL. Nothing.

moorereason:
We should either stop using FEEDBACK or FEEDBACK should be altered in some way.  It should either respect the logging thresholds or provide some other way to squelch the output.

digitalcraftsman:
Feedback uses `fmt.Println` and which isn't controlled by the logger. It would require that jww has an option for disabling this behavior. But you're right: Is it worth it?

bep:
>But you're right: Is it worth it?

No. Don't spend time on this. This can easily be handled by the user by redirecting output.

digitalcraftsman:
Then I'll close this pull request.

PR #2030, opened 2016-03-30
tpl: Fix panic in sort of nil sequence

moorereason:
Properly handle a nil sequence in the sortSeq function.  Test included.
Discovered while using `range sort .Site.Data.source.Undefined "foo"`.


bep:
Merged in fe92911

PR #2034, opened 2016-03-31
Disable syntax guessing for PygmentsCodeFences

### LINKED ISSUE #1861: Disable language guessing with PygmentsCodeFences
The `PygmentsCodeFences` option enables highlighting for all \`\`\` ... \`\`\` blocks, even if no language is specified. This leads to unwanted highlighting since the [-g flag is passed to pygments](https://github.com/spf13/hugo/blob/26d23f7f4becb2bbfa80d91ae9b27ffc76e13e3a/helpers/pygments.go#L99) in that case.

I think that the highlighting should only be be enabled for fenced blocks with an explicitly specified language. This would be in line with [Github's fenced code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/).

To disable automatic highlighting we could insert additional checks for `(lang != "")` in [line 38](https://github.com/spf13/hugo/blob/0f6b334b6715253b030c4e783b88e911b6e53e56/helpers/content_renderer.go#L38) and [line 81](https://github.com/spf13/hugo/blob/0f6b334b6715253b030c4e783b88e911b6e53e56/helpers/content_renderer.go#L81) in `content_renderer.go`.

phil-opp:
This disables highlighting for fenced code blocks without explicitly specified language. It also introduces a new `PygmentsCodeFencesGuessSyntax` config option (defaulting to false).

To enable syntax guessing again, add the following to your config file:

```
PygmentsCodeFencesGuessSyntax = true
```

This is a breaking change.

Fixes #1861.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -309,6 +309,7 @@ func loadDefaultSettings() {
 	viper.SetDefault("DisablePathToLower", false)
 	viper.SetDefault("HasCJKLanguage", false)
 	viper.SetDefault("EnableEmoji", false)
+	viper.SetDefault("PygmentsCodeFencesGuessSyntax", false)
 }
 
 // InitializeConfig initializes a config file with sensible default configuration flags.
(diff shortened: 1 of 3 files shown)


bep:
@phil-opp could you see my comment in #1861 and adjust this PR?

phil-opp:
@bep done

bep:
Good merged in 43b5dfa

PR #2035, opened 2016-04-01
Make ByCount sort consistently

### LINKED ISSUE #1930: Stable .ByCount index sort?
Currently the .ByCount sorter for indices does not order things reliably within a given count.  This means that every time I make a build my index of tags changes order.  It's not a huge deal but it's sortav annoying and adding a secondary sort seems like it would only help.  I could probably draft a patch but I wanted to verify that others are interested first.

robertbasic:
When two or more taxonomies have the same number of pages,
sort them by name to have consistent ByCount sorting of
taxonomies.

Fixes #1930


bep:
Thanks for this, merged in 7d5c9fb

Issue #245, opened 2014-03-30
Date.Format Internationalization


tubia:
Hi,
Maybe I don't understand well the logic of the .Date.Format variable, but it's possible to change the language of the months (for example) when they are displayed?
Something like {{ .Date.Format.IT "30 marzo 2006" }}

variadico:
Hi, there.

I wrote a general-purpose package, called [lctime], that might be able to fix this issue. It uses locales and `strftime` to format dates internationally. In addition, I use `go-bindata` to bundle locale data, which means devs don't have to distribute anything extra with their app.

I was thinking this issue could be easily solved if we could add `lctime.Strftime` as a template function. Then people could use this in their templates.

```
<p>{{strftime "%c" .Date}}</p>
```

The default locale is based on [environment variables]. However, if people want to change the locale, they can start Hugo like this.

```
LC_TIME=es_MX hugo server
```

Or maybe they could add a `locale` field in their `config.yaml` and Hugo could call `lctime.SetLocale` to explicitly set the locale.

Let me know if this is something that sounds interesting. I wouldn't mind opening a PR. Happy holidays! :christmas_tree: 

[lctime]: https://github.com/variadico/lctime
[environment variables]: https://godoc.org/github.com/variadico/lctime

PR #2041, opened 2016-04-02
tpl: Add localizedDate template function

soundrussian:
This adds a template function to implement date localization, as was discussed in issue #245. All credit should really go to @variadico , I've just created a pull request to learn myself some Go.

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -450,6 +450,207 @@ e.g.
 {{humanize "myCamelPost"}} → "My camel post"
 ```
 
+### localizeDate
+Converts the textual representation of the datetime into the localized form using the locale set by `languageCode`.
+
+These are the supported format directives. They're loosely based on [The Open Group Base Specifications Issue 7](http://pubs.opengroup.org/onlinepubs/9699919799/functions/strftime.html).
+
+<table class="table table-bordered">
+<thead>
+<tr>
+<th>Format directive</th>
+<th>Output</th>
+</thead>
+
+<tbody>
+<tr>
+<td>`%a`</td>
+<td>locale's abbreviated weekday name</td>
+</tr>
+
+<tr>
+<td>`%A`</td>
+<td>locale's full weekday name</td>
+</tr>
+
+<tr>
+<td>`%b`</td>
+<td>locale's abbreviated month name</td>
+</tr>
+
+<tr>
+<td>`%B`</td>
+<td>locale's full month name</td>
+</tr>
+
+<tr>
+<td>`%c`</td>
+<td>locale's appropriate date and time representation</td>
+</tr>
+
+<tr>
+<td>`%C`</td>
+<td>year divided by 100 and truncated to an integer</td>
+</tr>
+
+<tr>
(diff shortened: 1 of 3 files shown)


digitalcraftsman:
Hi @soundrussian,

it's a good idea to implement such a template function, especially while looking at #1744. 

> The default locale is based on [environment variables]. However, if people want to change the locale, they can start Hugo like this.

We should find a workaround once we have i18n support as a feature. Inside the template we will have access to the locale of the current page. Depending on the locale visitors should see the date formatted respectively. But this seems only to be possible when we set / change the environment accordingly.

[environment variables]: https://godoc.org/github.com/variadico/lctime

variadico:
Actually, it is possible to set the locale during runtime by calling `lctime.SetLocale`.

```go
lctime.SetLocale("es_MX")
fmt.Println(lctime.Strftime("%c", time.Now()))
// prints: lun 14 dic 2015 22:31:56 PST
```
From the [readme](https://github.com/variadico/lctime).

digitalcraftsman:
My bad. Thanks for the correction. Anyhow, it's a powerful combination.

bep:
This one needs some thinking, the bigger picture and all.

soundrussian:
I've updated the code to set locale from `languageCode` instead of environment variable, this seems more user-friendly. There might be some differences though between the formats of `languageCode` and expected locale, which I tried to address, but I'm not sure how `languageCode` is used in the wild.

digitalcraftsman:
Could you also add a few words for the documentation? Take a look at the examples in `docs/content/templates/functions.md`.

Maybe it seems trivial but why I would use `localizeDate` instead of <code>localize<b>d</b>Date</code> as function name.

soundrussian:
I agree that `localizeDate` is a better name, since it implies some action on the provided arguments. Changed the name and added an extensive documentation.

digitalcraftsman:
> bep: This one needs some thinking, the bigger picture and all.

How would you define the bigger picture? The multilingual feature should provide a frame. A more precise context about parts that need some thoughts could help.

I don't want to rush by adding this as a last-minute feature if we release the next version of Hugo soon. It's just a nice addition that fills a gap.

moorereason:
I don't like the `strftime` formatting.  That's the bigger picture I'm concerned about.  If this PR used Go time formats, I'd be on-board.

bep:
>I don't like the strftime formatting. That's the bigger picture I'm concerned about.

That is an extremely narrow minded picture ... I don't see the bigger picture, but I'm hoping someone else, specialists outside this project, would paint it for me. We're in the process of pulling lots of small localize* into a fairly fragmented and hard to maintain set of functions; there are language rulesets, dictionaries etc. duplicated, most likely not complete (I notice some shortcomings in my own language) ... 

We're not adding this in 0.17. What gets added is very hard to remove.

variadico:
@moorereason, Go makes international dates literally impossible. That's why `lctime` exists.
https://github.com/variadico/lctime#the-problem-with-the-go-standard-library

This is a trade off between slightly more idiomatic code—and giving non-English users the luxury of dates.

PR #2069, opened 2016-04-10
tpl: Do not write to cache when ignoring cache

### LINKED ISSUE #2067: Simplify ignoreCache
```
"Ignores the cache directory for reading but still writes to it"
```
Adjust that to ignore on both read and write -- and adjust the command flag description. No real win writing to cache when not reading from it. Keep it simple stupid.

See #2066

robertbasic:
Fixes #2067

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -222,7 +222,7 @@ func initHugoBuildCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&contentDir, "contentDir", "c", "", "filesystem path to content directory")
 	cmd.Flags().StringVarP(&layoutDir, "layoutDir", "l", "", "filesystem path to layout directory")
 	cmd.Flags().StringVarP(&cacheDir, "cacheDir", "", "", "filesystem path to cache directory. Defaults: $TMPDIR/hugo_cache/")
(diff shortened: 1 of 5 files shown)


commands/hugo.go
```diff
@@ -222,7 +222,7 @@ func initHugoBuildCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&contentDir, "contentDir", "c", "", "filesystem path to content directory")
 	cmd.Flags().StringVarP(&layoutDir, "layoutDir", "l", "", "filesystem path to layout directory")
 	cmd.Flags().StringVarP(&cacheDir, "cacheDir", "", "", "filesystem path to cache directory. Defaults: $TMPDIR/hugo_cache/")
-	cmd.Flags().BoolVarP(&ignoreCache, "ignoreCache", "", false, "Ignores the cache directory for reading but still writes to it")
```
bep: All below /docs/commands are autogenerated (we typically do it before a new releasee), so -- although this doesn't hurt it makes the changes harder to read, so revert it.

robertbasic: I did not know that! (I was wondering why so many duplicated docs, though)

PR #2075, opened 2016-04-12
Chomp Unicode BOM if present

wader:
Useful if using or sharing files with users that use editors that
append a unicode byte order marker header (like Windows notepad).

This will still assume files are UTF-8 encoded.


wader:
Actually for me it was Wordpad that added BOM but Notepad seems to be preserve it

PR #2076, opened 2016-04-13
tpl: Extend where to iterate over maps

moorereason:
Refactor and extend where to iterate over maps.

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -38,16 +38,13 @@ import (
 	"time"
 	"unicode/utf8"
 
-	"github.com/spf13/viper"
-
-	"github.com/spf13/afero"
-	"github.com/spf13/hugo/hugofs"
-
 	"github.com/bep/inflect"
-
+	"github.com/spf13/afero"
 	"github.com/spf13/cast"
 	"github.com/spf13/hugo/helpers"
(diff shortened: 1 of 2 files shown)


moorereason:
@bep, can you test this PR out?  It needs to be extended to cover more cases, but I'm stuck.

I've added a test case that I believe matches a data structure from `.Site.Data`.  My results are exactly the same (print `%#v` in the test case to see it), but `DeepEquals` returns false.  I can't figure out what I'm missing here.  Need another set of eyes on it.  Thanks

bep:
Change the check in the test to:

```go
if !reflect.DeepEqual(results, this.expect) {
				t.Errorf("[%d] Where clause matching %v with %#v, got\n%#v but expected\n%v", i, this.key, this.match, results, this.expect)
			}
```

And it will be more visible.

bep:
They look the same to me, too .. Which, I guess, must mean they are different in the types they contain (int vs int64?)

moorereason:
Found it.  I was returning `rv` (reflect.Value) instead of `rv.Interface()` (a map).

bep:
@moorereason could you do a rebase of this one against master?

moorereason:
@bep, done.

Issue #2074, opened 2016-04-12
Project website isn't rebuild automatically


digitalcraftsman:
Within the last days I merged pull request for the documentation. But the docs aren't rebuild. A look in the build log of Wercker showed that it was unable to `go get` an updated version of `afero`.

/cc @bep

bep:
This builds on a @spf13 server -- so no use in pinging me.

digitalcraftsman:
I'm sorry. I thought it would be possible to fix this in the `wercker.yml` file.

PR #2094, opened 2016-04-20
Update link to the contributing guide

mortonfox:


moorereason:
The README contains the correct URL.  The real problem is #2074.

Issue #2105, opened 2016-04-24
Export "detectFrontMatter" to fix a part of caddy-hugo plugin


hacdias:
Hello!

I would like to know if that function can be exported again because I'm using it on caddy-hugo plugin.

Thanks.

C.C.: @spf13 @bep

PR #2108, opened 2016-04-26
Export "detectFrontMatter" to fix caddy-hugo

hacdias:
See #2105

@bep @spf13


bep:
Closed, but since it's now exported, golint complains about missing godoc... Could you fix that?

hacdias:
Thanks. Yup, I'll do that now.

PR #2113, opened 2016-04-27
helpers: Use net/url for URL parsing in AbsURL

### LINKED ISSUE #2112: helpers: AbsURL schema detection is too simplistic
Currently, `AbsURL` checks for a `http` prefix in the input string to detect if a path is absolute.  This breaks if a user passes a valid path of `http/foo` and also fails to recognize other schemas.

We should use `net/url` to parse the path string.

moorereason:
Fixes #2112


moorereason:
Merged as e4ee1b89ad21f46dab96c9099f554bef4602650b

PR #2114, opened 2016-04-27
docs: Clarifies use of default archetypes when themes are active

yosemitebandit:
I previously thought that, when you ran something like `hugo new notes/test.md`, hugo would automatically look in `themes/THEMENAME/archetypes/deafult.md` if `archetypes/` did not exist, but that is not the case.  

You need to specify the theme for this to work, ala: `hugo new --theme=THEMENAME notes/test.md`.  This tweaks the documentation to remind users about that argument.


yosemitebandit:
`--theme` is no longer a valid flag with `hugo new`.  I'm now just specifying the theme in my `config.toml` and the archetype is correctly picked up when I run `hugo new`.

Issue #2057, opened 2016-04-07
Fix anchor mess


bep:
See https://github.com/spf13/hugo/issues/2056

This will be a breaking change (well, not as in glass breaking), but I think this will make more sense to most users. Duplicate ids are the special case here, not the other way around. 

This is a fairly common support question, too -- so If noone protest, I'm just doing it.

/cc @spf13 @halostatue

halostatue:
If this change is made, we probably need to make it so that plain ID anchors are *not* used by default on index pages (rather than actual pages). It’s when you might have the body of multiple posts in place that you *should* have unique anchors.

halostatue:
Also, this should be shouted really loudly in the release notes—someone who expects the current behaviour needs to know to change the default back.

bep:
>If this change is made, we probably need to make it so that plain ID anchors are not used by default on index pages (rather than actual page)

Yes, that would be nice, but that would mean double-rendering of the content page, and I'm reluctant to go that path. Lets think a little on this; I agree about the big bold notice in release notes, I think the sites that needs this are pretty rare.

octref:
I don't think anyone would use footnotes in non-post pages.
Not having crazy URL for footnote is my preference but I feel it's a more sensible default.

halostatue:
@octref You would be __wrong__. The reason we implemented the hashed anchors for footnotes & other anchors is *specifically* because people (18 months ago) were saying that their homepage wasn’t validating because they had multiple anchors and footnote anchors showing up on their homepage.

If you use numeric footnotes, and your main page renders multiple posts *in full* (which was, and probably still is, the default of the default Hugo template, although my site template does not do that), then you *will* have conflicting anchors and you *probably* want to have these uniquely generated anchors. The point is that aside from controlling how many posts render in full (or even render in part if there is a footnote before the break)…this may happen even for sites you don’t think it will happen for.

I agree with @bep that this is *probably* the right change, and that it will confuse people who don’t have this problem, but this code was introduced to fix a *very specific* problem that was pointed out. (And, BTW, it happens *even without* footnotes if you use headings so that the anchors to headings are repeated across posts.)

[...]

PR #2118, opened 2016-04-30
Enhance mmark extension support

### LINKED ISSUE #1970: helpers: Mmark support incomplete, e.g. missing EXTENSION_ABBREVIATION
As @mikhail73 reported in miekg/mmark#60 and https://discuss.gohugo.io/t/the-abbreviations-mmark-does-not-work-in-the-hugo/2867, the current experimental mmark support in Hugo is incomplete.  For example, `EXTENSION_ABBREVIATION` is not enabled, and whether mmark extensions can be customized is a question to me too.  Would be nice to get mmark support on par with Blackfriday support in Hugo.

anthonyfok:
Add missing EXTENSION_ABBREVIATION and several other extensions, and allow users to specify extensions and extensionsmask under a `[mmark]` table similar to the existing `[blackfriday]` support in `config.toml`.

Fixes #1970

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -113,6 +113,7 @@ var blackfridayExtensionMap = map[string]int{
 
 // Mmark holds configuration values for Mmark rendering.
 type Mmark struct {
+	Smartypants     bool
 	HrefTargetBlank bool
 	PlainIDAnchors  bool
 	Extensions      []string
@@ -122,6 +123,7 @@ type Mmark struct {
 // NewMmark creates a new Mmark filled with site config or some sane defaults.
 func NewMmark() *Mmark {
 	combinedParam := map[string]interface{}{
+		"smartypants":     false,
 		"hrefTargetBlank": false,
 		"plainIDAnchors":  true,
 	}
@@ -328,6 +330,10 @@ func getMmarkHTMLRenderer(defaultFlags int, ctx *RenderingContext) mmark.Rendere
 
 	htmlFlags := defaultFlags
 	htmlFlags |= mmark.HTML_FOOTNOTE_RETURN_LINKS
+
+	if ctx.getMmarkConfig().Smartypants {
+		htmlFlags |= mmark.HTML_USE_SMARTYPANTS
+	}
 
 	return &HugoMmarkHTMLRenderer{
 		mmark.HtmlRendererWithParameters(htmlFlags, "", "", renderParameters),
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -289,11 +289,15 @@ func getMarkdownExtensions(ctx *RenderingContext) int {
 	for _, extension := range ctx.getConfig().Extensions {
 		if flag, ok := blackfridayExtensionMap[extension]; ok {
 			flags |= flag
+		} else {
+			jww.ERROR.Printf("blackfriday.extensions: %q is not a valid extension", extension)
 		}
 	}
 	for _, extension := range ctx.getConfig().ExtensionsMask {
 		if flag, ok := blackfridayExtensionMap[extension]; ok {
 			flags &= ^flag
+		} else {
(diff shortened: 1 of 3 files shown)


helpers/content.go
```diff
@@ -359,6 +429,15 @@ func (c *RenderingContext) getConfig() *Blackfriday {
 	return c.Config
 }
 
+func (c *RenderingContext) getMmarkConfig() *Mmark {
+	c.configInit.Do(func() {
+		if c.MmarkConfig == nil {
+			c.MmarkConfig = NewMmark()
```
bep: This is will probably be correct in the end (I haven't looked closely), but just looking at this makes me ask "that doesn't look correct? Why don't you init all the configs?"

So what I would was:

Make a MmarkRenderingContext and BlackfridayRenderingContext and embed the common stuff from RenderingContext.

anthonyfok: Indeed, I was feeling bit uneasy about that getMmarkConfig() thingy but didn't know any better.

Thank you Bjørn for your advice!  I will try to understand your method better and work on it in the coming week.

bep: Just ask. This is good stuff, btw. Mmark looks more and more useful.

anthonyfok: Indeed.  I just found out that mmark has implemented EXTENSION_UNIQUE_HEADER_IDS (when detecting identical, anchors add a sequence number -1, -2 etc.) and this would probably make solving #2057 easier.  There are some really nice thought-out features that make mmark an attractive alternative to Blackfriday, though I must admit I haven't quite yet switched over to mmark myself.

bep: The sequence number doesn't really solve Hugo's problem -- as there will still be potential id conflicts when rendering pages on the home page (two pages with the same title in the same position), but it would probably be "good enough" for most people.

anthonyfok:
Note to self:

See https://discuss.gohugo.io/t/smart-quotes-in-titles-and-mmark-bodies/3485

And remember to do the equivalent of the following (but in cleaner way using @ bep's recommendation, and for other mmark options too):

```diff
diff --git a/helpers/content.go b/helpers/content.go
index a631336..362b3ba 100644
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -108,6 +108,7 @@ var blackfridayExtensionMap = map[string]int{
 
 // Mmark holds configuration values for Mmark rendering.
 type Mmark struct {
+	Smartypants     bool
 	HrefTargetBlank bool
 	PlainIDAnchors  bool
 	Extensions      []string
@@ -117,6 +118,7 @@ type Mmark struct {
 // NewMmark creates a new Mmark filled with site config or some sane defaults.
 func NewMmark() *Mmark {
 	combinedParam := map[string]interface{}{
+		"smartypants":     false,
 		"hrefTargetBlank": false,
 		"plainIDAnchors":  true,
 	}
@@ -317,6 +319,10 @@ func getMmarkHTMLRenderer(defaultFlags int, ctx *RenderingContext) mmark.Rendere
 	htmlFlags := defaultFlags
 	htmlFlags |= mmark.HTML_FOOTNOTE_RETURN_LINKS
 
+	if ctx.getMmarkConfig().Smartypants {
+		htmlFlags |= mmark.HTML_USE_SMARTYPANTS
+	}
+
 	return &HugoMmarkHTMLRenderer{
 		mmark.HtmlRendererWithParameters(htmlFlags, "", "", renderParameters),
 	}
```

To test, add the following in config.toml in a test site:

```toml
[mmark]
smartypants = true
```

bep:
What is the status of this one?

bep:
@anthonyfok -- I will close this one as it has not been updated in ages and the PR itself has drifted pretty far away from master.

anthonyfok:
Thanks @bep!  Good call!  And sorry for my inactivity.

PR #2130, opened 2016-05-07
helpers: Remove ToReader funcs

moorereason:
Remove StringToReader and BytesToReader in favor of using the stdlib directly.


moorereason:
Merged as e2aea65170985d7428c468ce0591557ad95af884

PR #2146, opened 2016-05-13
Add --ignoreFiles command-line flag

TheOnlyRew:
This PR adds an `--ignoreFiles [string]` general CLI flag that will slice a provided CSV, and append the []string to the ignoreFiles array set in the cfg file.

It should be noted that I have built and "tested" this code by hand, but I'm not familiar enough with this codebase to know if there's somewhere I could/should be adding automated tests for command-line interactions. It should also be noted that this is the first Golang project I've ever touched, so I might be doing something _real_ stupid without knowing it. Any and all guidance / criticism appreciated.

It's also potentially worth discussing that the use case that motivated this addition would be better served by an additive mechanism (`--renderMatching`? `--ignoreFilesExcept`?), rather than the subtractive tool that's currently in place. This was an easy addition, though, and I don't feel confident enough to make that kind of larger UX change without 1) more experience in Go and 2) input from the community/contributors.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -155,6 +155,7 @@ var (
 	logFile     string
 	theme       string
 	source      string
+	ignoreFiles string
 )
 
 // Execute adds all child commands to the root command HugoCmd and sets flags appropriately.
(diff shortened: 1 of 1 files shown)


TheOnlyRew:
Tangential point of discussion; including an empty string in the flag (`--ignoreFiles ""`) _or_ the cfg file (`ignoreFiles: [""]`) will result in no files rendering. As this is the current behavior -- and because an empty RegEx really will match on literally everything -- I didn't modify the behavior. If this should change, the below line could be added to [this file](https://github.com/spf13/hugo/blob/master/source/filesystem.go#L153-L154)

```
if len(ignorePattern) == 0 { return false }
```

bep:
Hugo 0.16 adds environment variables support. I think that covers this.

TheOnlyRew:
w/ env var support (and being 8 months out of date) I'm gonna go ahead and close out the PR.

PR #2151, opened 2016-05-16
corrects git clone for hugo_theme_robust

nurp:


dim0627:
lgtm:+1:

digitalcraftsman:
Hello @nurp,

the path itself is not wrong. This format is used for SSH connections. But not all users have to be familiar with SSH. You switch to HTTPS works always and is perhaps the better default protocol. 

I've merged your contribution as b542ba70f12b7b333f278c78dcac55ade75eefd5 (master) and b542ba70f12b7b333f278c78dcac55ade75eefd5 (v0.15.docs).

Cheers,
Digitalcraftsman

PR #2153, opened 2016-05-18
Docs: Add an example of .PrevInSection

revelt:
It's not a common knowledge what the "pointer" is so let's add an example of how to use .PrevInSection with .Permalink


digitalcraftsman:
Show, don't tell. This addition would make it clear. However, I would put the `For example,` also inside the if statement in case a previous page doesn't exist.

digitalcraftsman:
Hello @revelt,

thanks again for this addition. I've merged your contribution as 5611631539cdff75ce849b85f0f324e831e2e8a5 (master) and 8e46c56431cff91f59ec45b19a5e2f36cb68c9ab (v0.15.docs).

Cheers,
Digitalcraftsman

PR #2155, opened 2016-05-18
docs: Document .File variables for pages

### LINKED ISSUE #2154: docs: Document .Page.File
Each page has a `.File` object accessible (of type `source.File`).  We need to document that interface in the Template Variables documentation.

digitalcraftsman:
Fixes #2154.

### DIFF
--- a/docs/content/templates/variables.md
+++ b/docs/content/templates/variables.md
@@ -159,6 +159,20 @@ Also available is `.Site` which has the following:
 **.Site.BuildDrafts** A boolean (Default: false) to indicate whether to build drafts. Defined in the site configuration.<br>
 **.Site.Data**  Custom data, see [Data Files](/extras/datafiles/).<br>
 
+## File Variables
+
+The `.File` variable gives you additional information of a page.
+
+> **Note:** `.File` is only accessible on *Pages* but **not** on *Nodes* (like the homepage and list pages).
+
+Available are the following attributes:
+
+**.File.Path** The original relative path of the page, e.g. `content/posts/foo.md`<br>
+**.File.LogicalName** The name of the content file that represents a page, e.g. `foo.md`<br>
+**.File.BaseFileName** The filename without extension, e.g. `foo`<br>
+**.File.Ext** or **.File.Extension** The file extension of the content file, e.g. `md`<br>
+**.File.Dir** Given the path `content/posts/dir1/dir2/`, the relative directory path of the content file will be returned, e.g. `posts/dir1/dir2/`<br>
+
 ## Hugo Variables
 
 Also available is `.Hugo` which has the following:


digitalcraftsman:
@moorereason I tested the variable names with a theme and the following questions popped up:


- `.File.String`, `.File.Bytes` and `.File.Contents` are accessible but were either nil or empty
- What exactly is the difference between `.File.BaseFileName` and `.File.LogicalName`?
- Again what is the difference between `.File.Section` and `.File.Dir`? The last one just appends a slash à la posts/

bep:
Two items to add here:

1) As @digitalcraftsman  spotted, there are duplicates or almost duplicates here that needs to be fixed. I think @spf13 added some of these over time without much tought.
2) These structs are embeded on Page and, in my head, should be documented with their shortest path. So it is `.UniqueID` (already documented, btw) and not `.File.UniqueID`.

moorereason:
* `Content` is an io.Reader and is probably already at EOF by the time the templates are rendered.  `String` and `Bytes` use `Content`.  I wouldn't document these as template functions.
* Given `foo.txt`, `BaseFilename` is `foo` while `LogicalName` is `foo.txt`.
* I'm assuming `Dir` could return `content/post/dir2/dir3` when `Section` would just return `post`.

I usually look at the test cases to see how these should behave, but there are no tests for much of the `File` package.

bep:
Yes, some of these are really internals, and only exported because they live in a different package.

docs/content/templates/variables.md
```diff
@@ -159,6 +159,17 @@ Also available is `.Site` which has the following:
 **.Site.BuildDrafts** A boolean (Default: false) to indicate whether to build drafts. Defined in the site configuration.<br>
 **.Site.Data**  Custom data, see [Data Files](/extras/datafiles/).<br>
 
+## File Variables
+
+The `.File` variable gives you additional information of pages. Available are the following attributes:
+
+**.File.Path** The original relative path of the page, e.g. content/posts/foo.md<br>
```
moorereason: Use back ticks around the example output for each variable.

docs/content/templates/variables.md
```diff
@@ -159,6 +159,17 @@ Also available is `.Site` which has the following:
 **.Site.BuildDrafts** A boolean (Default: false) to indicate whether to build drafts. Defined in the site configuration.<br>
 **.Site.Data**  Custom data, see [Data Files](/extras/datafiles/).<br>
 
+## File Variables
+
+The `.File` variable gives you additional information of pages. Available are the following attributes:
```
moorereason: We need to make it clear that the `.File` object is only available on Pages and not Nodes (index, lists).

digitalcraftsman:
According to your feedback I updated my pull request.

> @bep: These structs are embeded on Page and, in my head, should be documented with their shortest path. So it is .UniqueID (already documented?) and not .File.UniqueID.

`UniqueID` isn't documented so far. Should I remove the reference here and document it instead under the `.Pages` variables?

digitalcraftsman:
I documented `UniqueID` in 20358f46c8b813ab627cc27e08b1689687e41410 and removed `.File.UniqueID`

digitalcraftsman:
Merged as 04889a5ee6247060cef4bde133a013ce7d4e3b61.

PR #2165, opened 2016-05-24
Remove empty toc list items to fix rendering of toc on pages that don't use full heading hierarchy

tebruno99:
This is my first contribution/modification of a Go project so I'm just looking for some thoughts and feedback on best solutions. This is the solution to a render issue that I encountered that is currently working for me.

When a Theme already puts the Title in the template as `<h1>` it is preferred in my case that the markdown of the page not include an `<h1>` (this looks very bad).  However, if `<h1>` is left out of the markdown then empty TOC list items are created which messes up the `TOC` of the page causing some strange situations.

I remove the empty list elements from the `TOC` after the `TOC` is generated so the `TOC` stays good looking regardless of the heading hierarchy.

---

Example of a document that starts at `<h3>` 
Before the change:
pre change

---
After the change:
post change

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -335,6 +335,15 @@ func ExtractTOC(content []byte) (newcontent []byte, toc []byte) {
 
 	newcontent = append(content[:startOfTOC], content[endOfTOC:]...)
 	toc = append(replacement, origContent[startOfTOC+len(first):endOfTOC]...)
+
+	emptyTop := []byte(`<ul>
+<li>
+`)
+	emptyBottom := []byte(`</ul></li>
+`)
+
+	toc = []byte(strings.Replace(string(toc), string(emptyTop), "", 10))
+	toc = []byte(strings.Replace(string(toc), string(emptyBottom), "", 10))
 	return
 }
 
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -341,9 +341,9 @@ func ExtractTOC(content []byte) (newcontent []byte, toc []byte) {
 `)
 	emptyBottom := []byte(`</ul></li>
 `)
-
-	toc = []byte(strings.Replace(string(toc), string(emptyTop), "", 10))
-	toc = []byte(strings.Replace(string(toc), string(emptyBottom), "", 10))
+	// Replace empty toc list items with empty strings
+	toc = []byte(bytes.Replace(toc, emptyTop, []byte(""), -1))
+	toc = []byte(bytes.Replace(toc, emptyBottom, []byte(""), -1))
 	return
 }
 
--- a/hugolib/page_test.go
+++ b/hugolib/page_test.go
@@ -261,9 +261,39 @@ in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
 pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui
 officia deserunt mollit anim id est laborum.`
 
-	pageWithToC = `---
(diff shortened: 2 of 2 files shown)


helpers/content.go
```diff
@@ -335,6 +335,15 @@ func ExtractTOC(content []byte) (newcontent []byte, toc []byte) {
...
 	toc = append(replacement, origContent[startOfTOC+len(first):endOfTOC]...)
+
+	emptyTop := []byte(`<ul>
+<li>
+`)
+	emptyBottom := []byte(`</ul></li>
+`)
+
+	toc = []byte(strings.Replace(string(toc), string(emptyTop), "", 10))
+	toc = []byte(strings.Replace(string(toc), string(emptyBottom), "", 10))
```
bep: * There is a `bytes.Replace` ... making this simpler.
* If you use a variable once, you might as well get rid of it (empty*)
* The last argument should be 1 (one replacement, if there are more there is something weird happening)

tebruno99:
Cool thanks I will rework it to use bytes.  The reason why there may be more than 1 is that it is empty list item for each heading not used.  So if the document's author starts with h6 there will be 5 empties (h1,h2,h3,h4,h5).

bep:
There is a test failing.

One could argue that this should be fixed where the ToC is produced, but I'm all for being practical if this is all for the better. Someone other than me should chime in ... @anthonyfok @digitalcraftsman ?

rdwatters:
@tebruno99 I agree that having the TOC default to `<h1>` as the first title in TOC doesn't make sense if you are following the old standard of a single `<h1>` on an article page. That said - and I don't know how difficult this would be to implement - I think the most powerful `{{.TableOfContents}}` variable would be one that accepts an argument for the starting level for the headings, since not everyone is going to start body-copy headings with `<h2>` either, depending on the nature of the content. In other words, `{{.TableOfContents "3" }}` would start the `<aside>` and ordered list with the first `<h3>` with `<h4>` as children, etc...if that makes sense.

tebruno99:
I think that does make sense. However, this doesn't remove the ability to start with `<h1>` it only removes the empties if `<h1>` wasn't started with.  For documents that do start with `<h1>` there should be no effect and they continue on unaware.

helpers/content.go
bep: Ah, OK, I misunderstood about the 10 ... then you can just add -1 (replace all).

bep:
I agree with @tebruno99 -- this PR kind of fixes a bug; we all want a better TOC, which would be a data structure that the user could present as he wish. But that is another train.

tebruno99:
I've added some of the suggestions.  Never done a github pull request before so sorry for the noise if I did it incorrectly.

bep:
That looks good, but you need to fix the failing test -- probably just need to adjust the expected result.

rdwatters:
Got it. Thanks for the clarification. And I agree that the TOC can continue to be improved, and I appreciate this small iterative improvement you guys are implementing now as well. Cheers!

tebruno99:
I've fixed the broken test by splitting it into 2 tests. One for `<h1>` and another for removing empties. I think that is best because the conditions for each are different and we want to be sure one doesn't break the other.

tebruno99:
I'm going to close this because I've found issues with the ToC that is generated after and agree that we should fix the empty elements in the generator not in the renderer. (Turns out that if you go from h3 to h2 the ToC still places the h2 as a child of the h3 after this patch).

Thanks for the input and exp!

PR #2166, opened 2016-05-24
Added content-type based partials overriding

XCMer:
Adds overriding of partials for specific content-type.

If you have a content type called "article", then:

`layouts/article/partials/sidebar.html` will override `layouts/partials/sidebar.html`, or the one in the theme.

This allows theme designers to bifurcate their layouts into partials that can be overridden by the end-user on a "content-type" basis. A useful use case (for which I implemented this), was to have different content in the sidebar for each content-type, which was basically a link to a list of recommended resources.

**Tests**
I've not written any test for this, and would like some support for doing it. The `template_test.go` file does not have a `partial` function test yet (or I have not found it).

### DIFF
--- a/tpl/template.go
+++ b/tpl/template.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


bep:
I believe what this PR does is what views were built for. /cc @spf13

bep:
I will close this. This is the role of view, and we're in the process of revamping the layout selection, and this does not fit into that world view.

PR #2179, opened 2016-06-02
Properly handle temp files while rebuilding

spf13:
@bep The fix turned out to be pretty simple, but it took a while to identify the cause. 

I've kept this as 3 commits because I feel they are 3 independent changes which address different issues, but together they resolve the general issue of temp file handling. I can squash if you prefer.


bep:
Yes, I can understand that ... the easiest to fix is the hardest to track. The commits looks fine; I can merge this, but we should let this linger for a day until we release.

PR #2186, opened 2016-06-06
Remove unneeded casts in page.getParam

sambooo:
Some calls to cast.To in getParam are no-ops that can be removed, since they just add type information that is discarded on return


moorereason:
Curious how you found these.  Are you using a linter?

sambooo:
Just caught my eye when looking at Page's implementation. Why do you ask?

moorereason:
Then you have a good eye.  :+1: I was asking about the linter because I don't know of an existing linter that would detect that issue.

lgtm

sambooo:
Force pushed to my fork like an idiot and got another commit stuck in the mix, no changes in the new version

PR #2191, opened 2016-06-08
Fix urlesc import url in Dockerfile

codyopel:
Point to the correct repo location.


anthonyfok:
Thank you!  Merged as acbedf6 (master) and 0066070 (v0.16.docs).

PR #2197, opened 2016-06-11
tpl: Modify tpl.humanize to add ordinals to integer input.

michaelorr:
- [x] Initial functionality
- [x] Add tests
- [x] Sign CLA
- [x] Update Docs
- [x] Squash commits + revised commit message

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1696,8 +1696,12 @@ func countRunes(content interface{}) (int, error) {
 	return counter, nil
 }
 
-// humanize returns the humanized form of a single word.
+// humanize returns the humanized form of a single parameter.
+// If the parameter is either an integer or a string containing an integer
+// value, the behavior is to add the appropriate ordinal.
 // Example:  "my-first-post" -> "My first post"
+// Example:  "103" -> "103rd"
+// Example:  52 -> "52nd"
 func humanize(in interface{}) (string, error) {
 	word, err := cast.ToStringE(in)
 	if err != nil {
@@ -1708,6 +1712,11 @@ func humanize(in interface{}) (string, error) {
 		return "", nil
 	}
 
+	_, ok := in.(int)           // original param was literal int value
+	_, err = strconv.Atoi(word) // original param was string containing an int value
+	if ok == true || err == nil {
+		return inflect.Ordinalize(word), nil
+	}
 	return inflect.Humanize(word), nil
 }
 
--- a/tpl/template_funcs_test.go
+++ b/tpl/template_funcs_test.go
@@ -1801,11 +1801,16 @@ func TestHighlight(t *testing.T) {
 func TestInflect(t *testing.T) {
 	for i, this := range []struct {
 		inflectFunc func(i interface{}) (string, error)
-		in          string
+		in          interface{}
 		expected    string
 	}{
 		{humanize, "MyCamel", "My camel"},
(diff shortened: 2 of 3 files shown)


tpl/template_funcs.go
```diff
@@ -1699,13 +1699,21 @@ func countRunes(content interface{}) (int, error) {
...
 // Example:  "my-first-post" -> "My first post"
+// Example:  "103" -> "103rd"
 func humanize(in interface{}) (string, error) {
 	word, err := cast.ToStringE(in)
 	if err != nil {
 		return "", err
 	}
-	return inflect.Humanize(word), nil
+
+	_, err = strconv.Atoi(word)
```
michaelorr: Note: I used `strconv.Atoi` here instead of `cast.ToIntE` because `ToIntE` degrades a little more gracefully on some conditions than we want in this case. Floats and bools for example.

michaelorr:
The AppVeyor failure looks like a networking blip. Is there a way to re-trigger it?

tpl/template_funcs.go
bep: Agree. This should be integers; that said I think it would be cleaner if you pull the integer check to the top and did a type assertion for the integer types. Aka: It looks funky to first convert to a string, then back to int when that was what you had in the first place. The downside is that there are lots of integer types, but we should have a helper for that (like cast.ToInt, but not as permissive ...

michaelorr: Hrm... I had structured this way for two reasons. I was trying to support strings that were integer-y i.e. `"103"` but hadn't considered the case of a literal `int` being passed. But also because `inflect.Ordinalize` expects a string param, so we'll need to do that conversion regardless of if we end up calling `inflect.Ordinalize` or `inflect.Humanize`.
It's simple enough to add the logic for handling literal `int` arguments to the template func. I'll add that.

tpl/template_funcs_test.go
```diff
@@ -1799,10 +1799,13 @@ func TestHighlight(t *testing.T) {
 func TestInflect(t *testing.T) {
 	for i, this := range []struct {
 		inflectFunc func(i interface{}) (string, error)
-		in          string
+		in          interface{}
 		expected    string
 	}{
 		{humanize, "MyCamel", "My camel"},
+		{humanize, "41", "41st"},
+		{humanize, 103, "103rd"},
```
bep: Add test variants for int64 and int32.

michaelorr: Will do.

michaelorr: I'm adding the int64 test case but we cannot support int32 as easily here because the underlying `cast.ToStringE` doesn't support it either:
https://github.com/spf13/cast/blob/master/caste.go#L225-L262
https://github.com/spf13/cast/blob/master/cast_test.go#L45-L60

bep: OK, int64 and int are the important ones since those are the one we'll get.

michaelorr:
test added for int64
appveyer is complaining that the branch isn't mergable but I've just rebased off the most recent changes in upstream/master so I'm not sure what's going on there.
I know I need to update docs and squash my commits as well, but then I think this should be good to go

michaelorr:
Docs updated and commits squashed.
As far as I'm concerned, this is ready to be merged, thoughts @bep ?

PR #2204, opened 2016-06-13
Add tests for embedded shortcodes

### LINKED ISSUE #1956: tpl: Write tests for embedded shortcodes
These will never show up in code coverage reports, but we need to write tests for the embedded shortcodes.

Shortcodes are in `tpl/template_embedded.go`.

ks888:
Fixes #1956

### DIFF
--- a/hugolib/embedded_shortcodes_test.go
+++ b/hugolib/embedded_shortcodes_test.go
@@ -1,0 +1,308 @@
+// Copyright 2016 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


ks888:
The builds fail due to cyclic import. The new test in tpl imports hugolib, and hugolib imports tpl.

I guess I need to find ways to test embedded shortcodes without importing hugolib.

bep:
I think the pragmatic thing to do here is to move that test file to /hugolib.

ks888:
Certainly moving the test file is simpler way. I will try.

ks888:
I fixed the problem and the build is successful now. Please take a look.

bep:
Very good!

ks888:
Sorry this merged PR breaks AppVeyor build. I created the new PR #2225 to fix this issue.

Issue #2208, opened 2016-06-13
"hugo server --port=80" gives "Error: Port 80 already in use" even when it is just "permission denied"


anthonyfok:
As of Hugo v0.16, running `hugo server --port=80` as an unprivileged user always returns an error message saying `Error: Port 80 already in use` regardless if the port is actually in use or not.

For comparison, also as an unpriviledged user, running netcat as `nc -l 80` to listen on port 80 correctly returns the error `nc: Permission denied`.

The code in question resides in the server() function in https://github.com/spf13/hugo/blob/v0.16/commands/server.go#L123-L137:

```go
	l, err := net.Listen("tcp", net.JoinHostPort(serverInterface, strconv.Itoa(serverPort)))
	if err == nil {
		l.Close()
	} else {
		if flagChanged(serverCmd.Flags(), "port") {
			// port set explicitly by user -- he/she probably meant it!
			return newSystemErrorF("Port %d already in use", serverPort)
		}
		jww.ERROR.Println("port", serverPort, "already in use, attempting to use an available port")
		sp, err := helpers.FindAvailablePort()
		if err != nil {
			return newSystemError("Unable to find alternative port to use:", err)
		}
		serverPort = sp.Port
	}
```

The fix should be simple: just need to add some checks what the error actually is instead of assuming "Can't open a port" === "The port is already in use".

moorereason:
Fix will likely involve [`os.IsPermission`](https://golang.org/pkg/os/#IsPermission).

g3wanghc:
```
listen tcp 127.0.0.1:80: bind: permission denied
Error: Port 80 already in use
```

anthonyfok:
Thank you for your effort, @g3wanghc!

However, as far as I know, "Port < 1024 can only be opened by root" is not a universal rule.  For example, I think non-UNIX platforms like MS Windows has a different way of determining whether an unprivileged user can open a certain port or not.

Furthermore, even on Linux, there are ways to allow non-root users to listen on port < 1024, see, for example:

* http://stackoverflow.com/questions/413807/is-there-a-way-for-non-root-processes-to-bind-to-privileged-ports-1024-on-l
* https://debian-administration.org/article/386/Running_network_services_as_a_non-root_user.

I just tested the `setcap` method, and I was happily running `hugo server --port=80` as a non-root user!

So, instead of imposing artificial rules in the Hugo code, it is better to let the undelying OS tell us whether a port can be opened or not, and why.  :wink:

[...]

PR #2214, opened 2016-06-14
#2208: Specify permission error for ports under 1024

g3wanghc:
`bind: permission denied` is not recognized by [os.isPermission](https://golang.org/pkg/os/#IsPermission) so its checking if the user is `root` instead.

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -126,6 +126,9 @@ func server(cmd *cobra.Command, args []string) error {
 	} else {
 		if flagChanged(serverCmd.Flags(), "port") {
 			// port set explicitly by user -- he/she probably meant it!
+			if serverPort < 1024 && os.Geteuid() != 0 {
+				return newSystemErrorF("Port under 1024 can be opened only by root")
+			}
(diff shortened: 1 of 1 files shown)


anthonyfok:
Thank you for your effort, @g3wanghc!

However, as far as I know, "Port < 1024 can only be opened by root" is not a universal rule.  For example, I think non-UNIX platforms like MS Windows has a different way of determining whether an unprivileged user can open a certain port or not.

Furthermore, even on Linux, there are ways to allow non-root users to listen on port < 1024, see, for example:

* http://stackoverflow.com/questions/413807/is-there-a-way-for-non-root-processes-to-bind-to-privileged-ports-1024-on-l
* https://debian-administration.org/article/386/Running_network_services_as_a_non-root_user.

I just tested the `setcap` method, and I was happily running `hugo server --port=80` as a non-root user!

So, instead of imposing artificial rules in the Hugo code, it is better to let the undelying OS tell us whether a port can be opened or not, and why.  :wink:

PR #2217, opened 2016-06-15
commands: Port binding net.OpError handling

g3wanghc:
As @bep suggested, I appended the os err message to create a generic Hugo error.

New messages are:

```
Error: Port 80 permission denied
Error: Port 1025 address already in use
```

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -126,7 +126,7 @@ func server(cmd *cobra.Command, args []string) error {
 	} else {
 		if flagChanged(serverCmd.Flags(), "port") {
 			// port set explicitly by user -- he/she probably meant it!
-			return newSystemErrorF("Port %d already in use", serverPort)
+			return newSystemErrorF("Server startup failed: %s", err)
 		}
 		jww.ERROR.Println("port", serverPort, "already in use, attempting to use an available port")
 		sp, err := helpers.FindAvailablePort()


commands/server.go
```diff
@@ -126,7 +126,13 @@ func server(cmd *cobra.Command, args []string) error {
 	} else {
 		if flagChanged(serverCmd.Flags(), "port") {
 			// port set explicitly by user -- he/she probably meant it!
-			return newSystemErrorF("Port %d already in use", serverPort)
+			switch t := err.(type) {
+			case *net.OpError:
+				err_msg := t.Err.(*os.SyscallError).Err.Error()
+				return newSystemErrorF("Port %d "+err_msg, serverPort)
+			default:
```
bep: All of this looks too complicated. What do you get if you just print the original error?

bep: As in return return newSystemErrorF("Failed to bind to port: %s", err)

g3wanghc: @bep `Error: Failed to bind to port 80: listen tcp 127.0.0.1:80: bind: permission denied`

bep: That is close; I guess this would be fine then:

newSystemErrorF("Server startup failed: %s", port, err)

g3wanghc: @bep Sounds good, that makes the code a lot simpler.

g3wanghc:
@bep New error messages are:

`Error: Server startup failed: listen tcp 127.0.0.1:80: bind: permission denied`
`Error: Server startup failed: listen tcp 127.0.0.1:1025: bind: address already in use`

Issue #347, opened 2014-06-25
Can't put irc:// links in configuration menu


krig:
Linking to irc channels in a menu option doesn't work (the link gets stripped).

Example (TOML):

[[menu.main]]
name = "IRC: #golang at freenode"
url = "irc://irc.freenode.net/#golang"

anthonyfok:
Just a quick note that a solution has been found by the use of template.HTMLAttr and a new "safeHtmlAttr" action.  For example, when the relevant code has been added to Hugo, we may replace

    <a class="" href="{{ .Url }}">...</a>

with

    <a class="" {{ printf "href=%q" .Url | safeHtmlAttr }}>

and the irc:// link would work.

Still working on the documentation.  Will commit tonight or tomorrow.  :-)

krig:
:+1:

anthonyfok:
After reading the html/template docs more carefully and searching for similar functions, it turns out there is a template.URL that I kept missing, so we can do a `safeUrl` too like this:

    <a href="{{ .Url | safeUrl }}">...</a>

which is more compact than `<a {{ printf "href=%q" .Url | safeHtmlAttr }}>`, though I suppose someone will find use other uses for `safeHtmlAttr` sooner or later.  Will add `safeUrl` in another commit.

There are actually [template.JS](http://golang.org/pkg/html/template/#JS) and [template.JSStr](http://golang.org/pkg/html/template/#JSStr) too, which we may use for new `safeJs` and `safeJsStr` too in case anyone needs them.

anthonyfok:
Hello @krig, @tatsushid:

After much consideration, I have decided to disable `safeHtmlAttr` template function that @tatsushid has kindly contributed, and have instead promoted the simpler `safeUrl` for use cases like the one described by @krig in this issue report.

What happened is that when I was trying to write the documentation for these new template functions, I struggled to think of any use cases of `safeHtmlAttr` aside from this one which is better served with `safeUrl`.  I have then decided to disable `safeHtmlAttr` for the time being, until it becomes indispensable for certain use cases.  Ditto for the yet unimplemented `safeJs` and `safeJsStr` whose functionality the underlying Go html/template has already provided.

(Yes, I am wary of potential security issues with `safeHtmlAttr`, `safeJs` and `safeJsStr`, especially when there is currently no demand for them, at least not yet.  May be in the future.)

Thank you for your understanding.

anthonyfok:
[...]

PR #2234, opened 2016-06-22
Enable safeHTMLAttr

webguerilla:
Hi together,

I'm using Hugo shortcodes to generate HTML forms from JSON. I like to set HTML attributes dynamically, but was not able to do so because the attributes were replaced with "ZgotmplZ". The template functions safeHTML, etc. did not work.

After some research I found the template function "safeHTMLAttr", which was already implemented, but disabled, because the developer has seen no use case for it. See https://github.com/spf13/hugo/issues/347#issuecomment-70612848

I have now enabled the function again, because it would be really useful for me. If you like to merge it, I could also write a test and improve the documentation.

Best regards,
Marco

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -546,7 +546,6 @@ rendering the whole string as plain-text like this:
 <p>© 2015 Jane Doe.  &lt;a href=&#34;http://creativecommons.org/licenses/by/4.0/&#34;&gt;Some rights reserved&lt;/a&gt;.</p>
 </blockquote>
 
-<!--
 ### safeHTMLAttr
 Declares the provided string as a "safe" HTML attribute
 from a trusted source, for example, ` dir="ltr"`,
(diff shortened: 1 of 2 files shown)


moorereason:
I like this PR.  We live in the Go Template world with it's features and quirks.  We should give users the tools they need to build whatever they want.

bep:
I like it, too.

anthonyfok:
Ditto.  :-)  Looking back at #347, it seems silly to me that I would have been so overly cautious back then.  So, let's merge it!  It's long overdue.

Merged as c21e2b3 after amending the commit message per the contribution guideline.
Thank you for your contribution, @webguerilla!

webguerilla:
@anthonyfok Thank you for merging, however I have not contributed much, just uncommented your work :smiley:

anthonyfok:
@webguerilla Thank you for your compliments!  But the credit actually goes to @tatsushid who contributed this feature in the first place.

PR #2239, opened 2016-06-27
Add automatic page date fallback

vbatoufflet:
Hi,

This PR adds an automatic value for `Page.Date` using the file's `ModTime` as fallback when none is provided?

Please let me know if it suits to you.

Regards.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -663,6 +663,13 @@ func (p *Page) update(f interface{}) error {
 		p.Draft = !*published
 	}
 
+	if p.Date.IsZero() {
+		fi, err := hugofs.Source().Stat(filepath.Join(helpers.AbsPathify(viper.GetString("ContentDir")), p.File.Path()))
+		if err == nil {
+			p.Date = fi.ModTime()
+		}
+	}
+
 	if p.Lastmod.IsZero() {
 		p.Lastmod = p.Date
 	}


bep:
I guess this is a fair compromise.

hugolib/page.go
```diff
@@ -663,6 +664,13 @@ func (p *Page) update(f interface{}) error {
 		p.Draft = !*published
 	}
 
+	if p.Date.IsZero() {
+		fi, err := os.Stat(filepath.Join(helpers.AbsPathify(viper.GetString("ContentDir")), p.File.Path()))
```
bep: You need to use the https://github.com/spf13/hugo/blob/master/hugofs/fs.go#L30 filesystem.

vbatoufflet:
Hi,

Thanks for the feedback. I just updated my PR to use hugofs.

luixxiul:
This causes a trouble if you are using Page in terms of WordPress: about/sitemap/privacy policy/etc which do not have ``.Date`` suddenly start to appear in feeds or the list of recent posts.

vbatoufflet:
Hi all,

Maybe we can define a configuration option to toggle automatic date feature?

WDYT?

bep:
>WDYT?

Yes, we need that, and set it to default = false.

luixxiul:
That would be great!

2016/07/01 16:25 "Bjørn Erik Pedersen" <notifications@github.com>:

vbatoufflet:
Yes, I will create a PR.

vbatoufflet:
Hi All,

Any suggestion on the setting name, e.g. `AutomaticFallbackDates` or `AutoDates`?

bep:
@vbatoufflet see my comment on https://github.com/spf13/hugo/issues/2244

PR #2257, opened 2016-07-05
tpl: Add a querify function to generate query strings inside templates

jimmysawczuk:
The `querify` function will take a set of parameters specified like a dict and return a url.Values object which can be .Encode'd into a query string.

Example:

```html
<a href="http://www.google.com?{{ (querify "q" "test" "page" 3).Encode | safeHTML }}">Search</a>
```

Returns:

```html
<a href="http://www.google.com?page=3&q=test">Search</a>
```

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -27,6 +27,7 @@ import (
 	"html"
 	"html/template"
 	"math/rand"
+	"net/url"
 	"os"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


bep:
Looks good. Could you "correct" your commit messages (I know, pedantic, but we like the imperative mood and titles without period at the end).

jimmysawczuk:
@bep no problem, fixed (I think) (and squashed into one commit).

Issue #2022, opened 2016-03-27
Wish a feature for TODO lists (Incorporate shurcooL/github_flavored_markdown?)


weaming:
```
* [ ] foo
* [x] bar
* [ ] baz
```

Refer: [FND/markdown-checklist: Python Markdown extension for lists of tasks with checkboxes](https://github.com/FND/markdown-checklist)

bep:
Your odds getting this improves drastically if you open an issue at https://github.com/russross/blackfriday

weaming:
@bep 
...

Done.

weaming:
[Wish a feature for TODO lists · Issue #249 · russross/blackfriday](https://github.com/russross/blackfriday/issues/249)

anthonyfok:
Interesting!  @shurcooL, who helps maintain blackfriday, replied at russross/blackfriday#249:

> FWIW, this is already implemented in github_flavored_markdown, which is a renderer that implements the blackfriday.Renderer interface. It's listed here:
>
> https://github.com/russross/blackfriday#other-renderers
> 
> (It wasn't hard, just... shurcooL/github_flavored_markdown@2b74a5b .)

So, it appears that it can be incorporated into Hugo, perhaps in a special "GitHub Flavored Markdown" mode.

phtan:
I want to tackle this. As far as I can see, nobody is claiming this issue.

shurcooL:
Let me know if you have any questions or if I can provide helpful guidance with this.

phtan:
@shurcooL , I appreciate your offer. I found what I needed when I looked through your tests at the github_flavored_markdown repository.

bep:
Replaced by more concrete issues.

phtan:
@shurcooL , was `github_flavored_markdown.Markdown()` meant to be a replacement of  blackfriday? Can a developer use github_flavored_markdown while still using blackfriday?

shurcooL:
@phtan, yes, `github_flavored_markdown.Markdown` method is very high level and does everything needed to convert source Markdown to output HTML, using GitHub-Flavored Markdown rendering rules.

It also performs HTML sanitization (since the GFM rendering method does too).

It uses `blackfriday` for parsing and rendering, and `bluemonday` for sanitization.

You can look at the [code](https://github.com/shurcooL/github_flavored_markdown/blob/b3921bea4a899100a4fd522dd940b819af56e437/main.go#L31-L59):

```Go
[...]

PR #2268, opened 2016-07-09
Wish a feature for TODO lists (Incorporate shurcooL/github_flavored_markdown?)

phtan:
This pull request addresses #2022


bep:
This shouldn't  its own markup type. I'll create some more clear issues.

phtan:
So... there will be new PRs to address the new issues #2269 and #2270 

Should I close this pull request, or will you do it yourself?

bep:
>Should I close this pull request, or will you do it yourself? 

Do as you please. You may change this PR to fix one of them. #2269 is the priority as that is the top used markdown.

shurcooL:
I agree that simply adding support for task lists into the existing Markdown mode (or both of them) is better than adding a 3rd new mode that users have to opt into. 👍

Basically apply a change like https://github.com/shurcooL/github_flavored_markdown/commit/2b74a5b054c18cbdf8589529327a7f9feba8b205 to the existing renderers you have.

PR #2281, opened 2016-07-14
Per #2279: `querify` func was double-escaping reserved characters

### LINKED ISSUE #2279: URL-escaping function for templates / shortcodes?
Hi folks,

I'm hoping to implement a "Tweet this" shortcode for my newly migrated blog; this would take a message and some optional args like 'via', 'hashtag', 'url' or 'accounts' and would construct a Twitter intent URL, and wrap the tweeted phrase in an intent link to tweet it.

The problem is, there doesn't appear to be a way to url-escape text from templates/shortcodes easily.

If I'm correct and this is a missing feature, I could look into adding an url-escape function? Or perhaps I'm missing something, but I did try to find this in the docs to no avail..

Thanks!

cathalgarvey:
Test case modified and expanded for `querify` to reflect original bug and changes.

Fixes #2279, but changes escaping behavior for "+" characters.

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1769,7 +1769,7 @@ func querify(params ...interface{}) (string, error) {
 	}
 
 	for name, value := range vals {
-		qs.Add(name, url.QueryEscape(fmt.Sprintf("%v", value)))
+		qs.Add(name, fmt.Sprintf("%v", value))
 	}
 
 	return qs.Encode(), nil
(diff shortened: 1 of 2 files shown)


cathalgarvey:
Thanks!

On 14 July 2016 14:53:48 IST, "Bjørn Erik Pedersen" <notifications@github.com> wrote:
>Merged #2281.
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/spf13/hugo/pull/2281#event-723019697

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Issue #2286, opened 2016-07-18
Wanted: HTMLEscape / HTMLUnescape template functions


cathalgarvey:
Hey folks,

I've been trying to make a shortcode for the Twitter "intent" link format, but keep needing functionality to get things working. The missing piece now is HTML entity unescaping in shortcode templates, because otherwise the URLs get constructed with escaped data when I use `{{%` shortcodes, breaking things.

I've made changes to add `htmlescape` and `htmlunescape` functions, which simply defer to `html.EscapeString` and `html.UnescapeString`. I've also added test cases and documentation for same. PR incoming. :)

Thoughts welcome!

PR #2287, opened 2016-07-18
Added `htmlescape` and `htmlunescape` template functions

cathalgarvey:
See #2286


bep:
Looks almost very good:

1) The naming standard is camel casing, so it should be `htmlEscape` and `htmlUnescape`.
2) Fix the commit message (add vs added)

cathalgarvey:
Sorry for that! Fixed, after some embarrassing rebase shenanigans.

PR #2300, opened 2016-07-24
hugolib: Add the alphabetical and diacritics stripped option to taxonomies ordering

### LINKED ISSUE #2180: Alphabetical sorting doesn’t support accents
While sorting taxonomies alphabetically, names beginning with diacritics such as 'Č' are placed to the end of list. Would it be possible at least strip diacritics and put them to the same position as items beginning with 'C'? You can test it for example on following three items: rock, rôle, rose.

ks888:
Fixes #2180

I added the new option `AlphabeticalAndDiacriticsStripped` to sort taxonomies alphabetically. Unlike the existing `Alphabetical` option, diacritics such as 'Č' are stripped when sorted. Displayed key names are still same.

Although it is possible to modify the `Alphabetical` option, I think it is not always desirable to strip diacritics when sorted.

### DIFF
--- a/docs/content/taxonomies/ordering.md
+++ b/docs/content/taxonomies/ordering.md
@@ -33,6 +33,15 @@ Taxonomies can be ordered by either alphabetical key or by the number of content
     {{ end }}
     </ul>
 
+It is possible to strip diacritics such as Č when ordered. Displayed key names are still same.
+
+    <ul>
+    {{ $data := .Data }}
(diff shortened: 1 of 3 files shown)


jry2:
@spf13 any chance to review or comment this PR? It works for our website, taxonomies are sorted as expected, but we must use custom Hugo build.

This bug must affect all languages with diacritics, I'm confused there are no more reports.

jry2:
After further investigation I'm not so sure about this PR, please see a new comment https://github.com/spf13/hugo/issues/2180#issuecomment-290962894

PR #2303, opened 2016-07-26
Finish the multilingual feature

bep:
This is currently slightly slower than master in a bench mark with 4 fairly big sites:

```
benchmark           old ns/op      new ns/op      delta
BenchmarkHugo-4     3146033049     3225819889     +2.54%

benchmark           old allocs     new allocs     delta
BenchmarkHugo-4     10557704       10850988       +2.78%

benchmark           old bytes      new bytes      delta
BenchmarkHugo-4     1628022920     1632976488     +0.30%
```

I guess this is mostly about doing "more work". I have done some general tuning in another PR (see #2378) after some profile work.

* [x] Add `Translations` and `AllTranslations` to Node.
* [x] Rework the site build process. Looping every site/language and doing the same for all of them kind of works, but isn't very effective nor elegant. Needs to extract the common tasks and do it once only.
* [x] Rework the documentation.
* [x] Fix all the `TODO(bep) multilingo` and  `TODO(bep) ml` comments.
* [x] Check the relevant URL funcs, `absURL` etc.
* [x] Add more tests
* [x]  Write redirect page to default lang on / when multilingual
* [x]  Write a global sitemap.xml
* [x]  Check that taxonomies can be configured on a per language basis
* [x] Add YAML config test  
* [x]  Check sitemap per language
* [x]  Check list command etc.
* [x] Move shortcode handling to "pre render". The site isn't really ready at the current point where the shortcodes gets rendered, which comes as a surprise to many. This is even more true now with the multilanguage feature (Site.Pages etc. will be empty). So we need to move this handling to a later stage.
* [x]  Watch: Handling adding and removing languages
* [x]  Check Blackfriday config per site
* [x]  Livereload inject
* [x] Check ref, relref  
* [x] implement https://github.com/spf13/hugo/issues/2312
* [x] Improve the default i18n vs missing bundles and translations 
* [x] See https://github.com/spf13/hugo/pull/2303/files#diff-d7b04948857565053b55f353ccb2bab2R296 while this is a temporary workaround until we get a proper registry for "all nodes and pages", this may be confusing to users expecting the home object in the translations object to be a full blown *home*. Now even the title is kind of wrong when it is set per language. We should fix this if it isn't too hard. A temporary "node map" maybe.
* [x] check data in Page.Site.Data from shortcodes + add some data to the ML tests
* [x] fix variables docs section -- there are some old and not valid info there now 
* [x] Double check all "reuse of nodes for RSS etc"
* [x] Double check the collection transfers
* [x] Check i18n template func in shortcode 
* [x]  Check shortcode changes vs livereload
* [x]  Fix Go 1.7 vet issues


the1900:
I really delightful about this feature. If you don't mind I think taxonomy's should be multilingual too.

When see the taxonomy the url is fixed to `domain/taxonomy` but if url were `domain/lang/taxonomy` then it is more flexible I guess.

bep:
@the1900 I meant it when I said "no comments, please". I have a shitload of work to get the foundation up and running and I don't need "it would be nice if ...".

bep:
OK, this should work. A little pragmatic in some areas, but it is well tested.

As to how it works, see https://github.com/spf13/hugo/blob/66971e7e299facd61fb0e2e037ed290143dbc46e/docs/content/content/multilingual.md

Happy code review!

/cc @spf13

hacdias:
I'll check this out asap. :) Thanks for mentioning.

bep:
Note to self, some weird recursive things happen if you get the TOML mapping wrong like this (see the blackfriday config). Probably related to the config watcher:

```
[Languages.en]
weight = 1

[Languages]
[Languages.nn]
weight = 2
title = "bep.is (Nynorsk)"
[Languages.en.blackfriday]
angledQuotes = true

[Languages]
[Languages.nn]
weight = 3
title = "bep.is (Nynorsk)"
[Languages.en.blackfriday]
angledQuotes = true
```

gntech:
There is some problem reading the Language config when using a yaml-config file.
```
---
baseurl: "http://replace-this-with-your-hugo-site.com/"
languageCode: "en-us"
title: "My New Hugo Site"

Languages:
  en:
    weight: 1
  fr:
    weight: 2
```
```
Error: Error building site: Failed to parse multilingual config: Language config is not a map: map[]
```
When I try the same config in a toml-format it works as it should.

docs/content/content/multilingual.md
```diff
@@ -0,0 +1,151 @@
...
+title: Multilingual Mode
+weight: 68
+toc: true
+---
+Hugo supports multiple languages side-by-side (added in `Hugo 0.17`). Define the available languages in a `Languages` section in your top-level `config.yaml` (or equivalent).
+
+Example:
+
+```
+DefaultContentLanguage = "en"
```
gntech: This is not yaml-syntax

gntech: This is not yaml-syntax

bep: You are right, it is TOML ... Will fix the doc.

docs/content/templates/variables.md
```diff
@@ -58,6 +58,8 @@ matter, content or derived from file location.
 **.IsPage** Always true for page.<br>
 **.Site** See [Site Variables]({{< relref "#site-variables" >}}) below.<br>
 **.Hugo** See [Hugo Variables]({{< relref "#hugo-variables" >}}) below.<br>
+**.Translations** A map to other pages with the same filename, but with a different language-extension (like `post.fr.md`).  Populated only if `Multilingual` is enabled in your site config.
+**.Lang** Taken from the language extension notation.  Populated only if `Multilingual` is enabled for your site config.
```
gntech: Populated only if 'Languages' is enabled in your site config?

docs/content/content/multilingual.md
gntech: Actually I was referring to only line 17. That line is TOML while line 18-32 is YAML

gntech:
I am unable to use my data-files from shortcodes. This was possible before through ```.Page.Site.Data.``` Is this a bug or should I use some other variable now with the multilingual feature?

bep:
@gntech I will check data and add a test for it.

bep:
@gntech see https://github.com/spf13/hugo/commit/85345d92df927fc1240343dca0b2502f65594374 I cannot reproduce it. I would rather say that this was an issue pre this branch. Now we evaluate the shortcodes as late as possible.

bep:
To see what is needed to update a site to multilingual, see this commit from one of my sites:

https://github.com/bep/bep.is/commit/4d027b86b8aa8dfad0f7b4ed7b1baaf6b71c5616

It is a one-pager, so maybe not very typical, but you get the drift.

See it in action here:

http://bep.is/

nils-holger:
Config.yaml doesn't work for me: `Error: Error building site: Failed to parse multilingual config: Language config is not a map: map[]`. I've been using the exact code as in the docs (except the toml issue of @gntech above). Config.toml works great.

bep:
@nils-holger ok, I'll add a test for a yaml config, I'll admint I haven't tested that, but thought that should just work.

helpers/url.go
```diff
@@ -147,47 +147,98 @@ func MakePermalink(host, plink string) *url.URL {
...
+}
+
+func getLanguagePrefix() string {
+	defaultLang := viper.GetString("DefaultContentLanguage")
+	defaultInSubDir := viper.GetBool("DefaultContentLanguageInSubdir")
+
+	if !viper.GetBool("Multilingual") {
+		return ""
+	}
+	currentLang := viper.Get("CurrentContentLanguage").(*Language).Lang
```
moorereason: Check for `nil` before accessing `Lang`.

helpers/url.go
```diff
@@ -147,47 +147,98 @@ func MakePermalink(host, plink string) *url.URL {
...
+		}
+	}
+	return MakePermalink(baseURL, in).String()
+}
+
+func getLanguagePrefix() string {
+	defaultLang := viper.GetString("DefaultContentLanguage")
+	defaultInSubDir := viper.GetBool("DefaultContentLanguageInSubdir")
+
+	if !viper.GetBool("Multilingual") {
```
moorereason: Move this `if` to the top of the function to exit earlier.

helpers/url.go
bep: I need to panic if it isn't there (which suggests a programming error anyhow), so it might as well be a panic with a nil pointer.

hugolib/hugo_sites.go
```diff
@@ -0,0 +1,566 @@
...
+	}
+
+	if len(h.Sites) > 1 {
+		pages := h.Sites[0].AllPages
+		allTranslations := pagesToTranslationsMap(h.multilingual, pages)
+		assignTranslationsToPages(allTranslations, pages)
+	}
+}
+
+// preRender performs build tasks that needs to be done as late as possible.
```
moorereason: s/needs/need/

hugolib/node.go
```diff
@@ -144,10 +180,147 @@ type URLPath struct {
 	Section   string
 }
 
+func (n *Node) URL() string {
+	return n.addLangPathPrefix(n.URLPath.URL)
+}
+
+func (n *Node) Permalink() string {
+	return permalink(n.URL())
+}
```
moorereason: I know we do this already on the `Page` struct (see `.Page.Section()`), but I don't like these two funcs (`URL()` & `Permalink()`).  We're embedding `URLPath` into the `Node`, which embeds the `URL` and `Permalink` fields.  Overshadowing those fields with funcs smells bad to me.

We don't need to fix all of Hugo's problems in this PR, but I'd like to see us change the `Node` struct to have `urlPath URLPath` and then access `n.urlPath.URL`.  We'd need to add the `Slug()` and `Section()` funcs to `Node` and update code all over the place, but at least we'd avoid this overloading of embedded fields business and some confusion about how those embedded `URLPath` fields are supposed to be used, esp. with i18n thrown into the mix.

bep: Yes, I kind of agree ... but ... you will find a lot of stuff that should be done, but I think we should draw some line. I have put a lot of sweat into making better tests, so changes like you suggest should be safer to do. My big motivation for the change above was to get a simple 5 lines template example listing all the translations of a page or a node without all the IsPage and IsHome special cases. I feel that trumps a little code impurity for a while.

bep:
@nils-holger see https://github.com/spf13/hugo/pull/2303/commits/663f09d949b21d3200453dfb1c84ee5c599ff91c re YAML not working. Thanks for the report. I was of course too lazy to test JSON ...

docs/content/templates/functions.md
```diff
@@ -740,6 +746,23 @@ This can be useful if you want to use Gravatar for generating a unique avatar:
...
+translations.  The translations can exist in both the theme and at the
+root of your repository.
+
+e.g.: `{{ i18n "translation_id" }}`
+
+
+### T
+
+`T` is an alias to `i18n`. E.g. `{{ T "translation_id" }}`.
+>>>>>>> Add multilingual support in Hugo
```
nils-holger: Line 765 can just be deleted. This todo is done (Yay!). Maybe we could add a link to /content/multilingual/#translation-of-strings to clarify.

docs/content/templates/variables.md
nils-holger: There should be `<br>` at the end of line 61. I would also add a link to `/content/multilingual` after line 61.
`.Lang` works nice. But the map .`Translations` is empty for me.

nils-holger: I've been testing the variables in a template: `.Lang` works nice. But the map .`Translations` is empty for me.

docs/content/templates/functions.md
bep: That is a sloppy Git merge from me ... thanks.

bep: And a link is a good idea.

mitchchn:
Apologies if this isn't the right PR to report this, but I've found a bug: partials do not have access to `$.Site.Data` when rendering translated pages (it's set to a `nil` map).

bep:
@mitchchn I have tested that ... Is it something special about your case? Is the data in your theme? What OS?

bep:
@ mitchchn I will check, thanks.

bep:
@mitchchn OK, I see whats happening. Stupid mistake. Will fix.

bep:
@mitchchn I fixed the Data issue. Please take it for a spin :-)

gntech:
@bep commit 3b4607f fixed the problem I had with reaching my data from shortcodes!

mitchchn:
@bep That fixed the Data, thanks. There still seem to be things missing from .Site in translated content. $.Site.Taxonomies only includes translated pages for instance.

bep:
That is by design. Taxonomies are site local, but they can have translations. I will double check others.

gntech:
I have found a peculiar bug with the Permalink to translated content when I am on a node page. I have Swedish and English on my site with Swedish being the default language. When I am on the Swedish node page the link to the corresponding english node page works as expected however when I am on the English nodepage the Permalink back to Swedish nodepage points to the RSS "index.xml" of that node. A bit hard to explain but Very peculiar.

To summarize: The permalink to the translated counterpart of a node goes to the RSS page of the translated node.

bep:
@gntech I understand that issue, but I thought I had fixed every places. Earlier the node instances from, say the home page and the home page's rss feed, where reused. Now when we use the node pointers to create a "translation map", this didn't work. It's an easy fix and thanks for testing.

bep:
@gntech i have commited a fix to your issue.

docs/content/templates/variables.md
```diff
@@ -118,10 +121,13 @@ includes taxonomies, lists and the homepage.
 **.IsPage** Always false for nodes.<br>
 **.Site** See [Site Variables]({{< relref "#site-variables" >}}) below.<br>
 **.Hugo** See [Hugo Variables]({{< relref "#hugo-variables" >}}) below.<br>
+**.Translations** A list of translated versions of the current node. All nodes (except the pager nodes) can have translated counter parts. See [Multilingual]({{< relref "content/multilingual.md" >}}) for more info. <br>
+**.HasTranslations** Whether there are any translations to display.<br>
```
nils-holger: I've tried these variables in a single template:
Line 124: .Translations gives an empty list [] instead of the expected result
Line 125: .HasTranslations gives an error: `HasTranslations is not a field of struct type *hugolib.Page`

hugolib/site.go
```diff
@@ -909,10 +909,10 @@ func (s *Site) initializeSiteInfo() {
 		Languages:                      languages,
 		defaultContentLanguageInSubdir: defaultContentInSubDir,
 		GoogleAnalytics:                lang.GetString("GoogleAnalytics"),
-		RSSLink:                        permalinkStr(viper.GetString("RSSUri")),
+		RSSLink:                        permalinkStr(lang.GetString("RSSUri")),
 		BuildDrafts:                    viper.GetBool("BuildDrafts"),
 		canonifyURLs:                   viper.GetBool("CanonifyURLs"),
-		preserveTaxonomyNames:          viper.GetBool("PreserveTaxonomyNames"),
+		preserveTaxonomyNames:          lang.GetBool("PreserveTaxonomyNames"),
```
moorereason: Does this mean that `RSSUri` and `PreserveTaxonomyNames` can modified on a per-language basis?  If so, we should note that in the documentation.

bep: It is noted in the documentation in a vague way ... and I think it should stay this vague until it is carved in stone.

mitchchn:
One more: the `i18n` function does not appear to work in shortcodes.

```
ERROR: 2016/08/12 15:54:19 shortcode.go:590: error processing shortcode shortcodes/callout.html 
 ERR: template: shortcodes/callout.html:9:5: executing "shortcodes/callout.html" at <i18n "help">: error calling i18n: i18n not initialized, have you configured everything properly?
```

The same function call works fine in layouts.

bep:
@mitchchn @gntech and others: My last commit fixes some issues when running in server watch mode, with reloading of changed shortcodes and similar. It would be helpful if you could take it for a spin.

I consider this branch solid enough now. @spf13 may have some to say, so I'll assign this issue to him.

mitchchn:
@bep It works very well now. Nothing else has come up in my testing. :+1:

gntech:
@bep I have tested the latest code now and I have found no issues. Kjempebra! :)

nils-holger:
@bep I've tested all documented features and all is working great now. Thanks for all time and efforts.

bep:
For some performance numbers, see https://github.com/bep/hugo-benchmark

gour:
I've built 0.17multilingual and after reading the docs tried to run:

`hugo --i18n-warnings | grep i18n
i18n|MISSING_TRANSLATION|en|wordCount`

but it reports about:

`Error: unknown flag: --i18n-warnings`

What do I miss?

abourget:
What's `lugo`?!

abourget:
Would it be useful that I review some more ? Are we good to go with this or does someone else need to take a look ?

I'll be happy to migrate my sites with any changes that were introduced here.. (since m17n) !

gntech:
I think we are awaiting @spf to take a look at this and give the official go ahead

yitzhakbg:
### [Multipass build](https://heartbt.co.il/how/Multi-Language/)
simplifies multilingual blogs under one roof without having to invent wheels.

yitzhakbg:
### [Multipass build](https://heartbt.co.il/how/Multi-Language/)
enables building each language site on its own, yet have them all under a single roof. Simplifies multilingual. No need to invent new wheels. The idea in short is to work with each content branch as if it were its own blog, unconcerned about other languages and merge the at build time. Layout, themes, asset branches etc. can be common to all.

bep:
@yitzhakbg please use the forum for questions and marketing.

yitzhakbg:
Pardon me. Thanks for the correction

-- 
Yitzhak Bar Geva <https://www.linkedin.com/in/yitzhakbg>

On 1 September 2016 at 00:56:27, Bjørn Erik Pedersen (
notifications@github.com) wrote:

@yitzhakbg <https://github.com/yitzhakbg> please use the forum for
questions and marketing.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/spf13/hugo/pull/2303#issuecomment-243915090>, or mute
the thread

bep:
@spf13 has promised to take some time off his schedule to look at this PR. If he does not object to its content in the next few days, I will merge it.

bep:
Merged!

shaform:
If we have a global sitemap.xml, why don't we have a global robots.txt?

Issue #1587, opened 2015-11-17
One change in layout file triggers two builds in watch mode


bep:
BTW not related to the latest "in memory" changes. Not sure when this happened -- but here is a potential for a easy 100% speedup! Tested with both ACE and Go templates.

OSX latest.

```
hange detected, rebuilding site
2015-11-17 21:44 +0100
0 of 3 drafts rendered
0 future content
132 pages created
64 paginator pages created
16 tags created
10 categories created
0 series created
in 250 ms

Change detected, rebuilding site
2015-11-17 21:44 +0100
0 of 3 drafts rendered
0 future content
132 pages created
64 paginator pages created
16 tags created
10 categories created
0 series created
in 232 ms
``

PR #2305, opened 2016-07-26
Be less aggressive with CHMOD filesystem events

melo:
On 4679fbee41d3, rebuild was disabled on CHMOD filesystem events, but the code is overly aggressive.

In some situations, specially with older Mac's (using a Late 2008 Macbook), the events we receive might be aggregated. On my particular laptop, I get this events:

    INFO: 2016/07/26 18:08:51 hugo.go:737: Received System Events: ["<path>": WRITE|CHMOD]

These events are ignored because the code only checks for Chmod. This commit fixes this by checking that the event is also not a Write or Create.

Related to #1587.

----

This is my first contribution to any Go project, so please forgive any Go-newbiness going on. Any comments are appreciated.

I also don't know how to write tests about this. I don't see a `hugo_test.go` and frankly I don't know how to simulate filesystem events to reproduce this. I do know that Hugo now rebuilds correctly on these events on my slow Mac :). Help is appreciated here.

Thank you,

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -775,7 +775,10 @@ func NewWatcher(port int) error {
 					// A workaround is to put your site(s) on the Spotlight exception list,
 					// but that may be a little mysterious for most end users.
 					// So, for now, we skip reload on CHMOD.
-					if ev.Op&fsnotify.Chmod == fsnotify.Chmod {
+					// We do have to check for WRITE though. On slower laptops a Chmod
+					// could be aggregated with other important events, and we still want
+					// to rebuild on those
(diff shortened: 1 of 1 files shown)


bep:
Don't worry about the tests -- I know how painful it is to write file system events tests (see https://github.com/fsnotify/fsnotify/blob/master/integration_darwin_test.go#L22 for a test I wrote to simulate atomic writes on OS X); Hugo's problem is that it must support lots of browsers, and it is impossible (or very hard) to test them all ...

melo:
Thanks!

PR #2327, opened 2016-08-03
Change .yaml extension to .yml (.yaml still works)

MarkDBlackwell:
As you may know, the conventional extension for YAML files is `.yml` (rather than `.yaml`).

`.yaml` remains supported, presumably because spf13/viper supports it (e.g. [here](https://github.com/spf13/viper/blob/a0cdbddebd2b4e614f360b2220f32055d1afe7dd/viper.go#L220) and [here](https://github.com/spf13/viper/blob/a0cdbddebd2b4e614f360b2220f32055d1afe7dd/viper.go#L181)).

Most changes are documentation, some are error messages.

(BTW, I found `git diff --word-diff=color` useful in showing these changes.)

See #2 and (for example) use of ".yml" in issue 162.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -209,7 +209,7 @@ func initHugoBuilderFlags(cmd *cobra.Command) {
 }
 
 func initRootPersistentFlags() {
-	HugoCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is path/config.yaml|json|toml)")
+	HugoCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is path/config.yml|json|toml)")
 
 	// Set bash-completion
 	validConfigFilenames := []string{"json", "js", "yaml", "yml", "toml", "tml"}
--- a/commands/import_jekyll.go
+++ b/commands/import_jekyll.go
@@ -193,7 +193,7 @@ func loadJekyllConfig(jekyllRoot string) map[string]interface{} {
 	exists, err := helpers.Exists(path, fs)
 
 	if err != nil || !exists {
-		jww.WARN.Println("_config.yaml not found: Is the specified Jekyll root correct?")
+		jww.WARN.Println("_config.yml not found: Is the specified Jekyll root correct?")
 		return nil
 	}
 
@@ -245,18 +245,8 @@ func createConfigFromJekyll(inpath string, kind string, jekyllConfig map[string]
 		"languageCode":       "en-us",
 		"disablePathToLower": true,
 	}
-	kind = parser.FormatSanitize(kind)
 
-	by, err := parser.InterfaceToConfig(in, parser.FormatToLeadRune(kind))
-	if err != nil {
-		return err
-	}
-
-	err = helpers.WriteToDisk(filepath.Join(inpath, "config."+kind), bytes.NewReader(by), hugofs.Source())
-	if err != nil {
-		return
-	}
-
+	createConfig(inpath, kind, in)
 	return nil
 }
 
--- a/commands/new.go
+++ b/commands/new.go
@@ -135,7 +135,7 @@ func doNewSite(basepath string, force bool) error {
 			return errors.New(basepath + " already exists and is not empty")
 
 		case !isEmpty && force:
-			all := append(dirs, filepath.Join(basepath, "config."+configFormat))
+			all := append(dirs, filepath.Join(basepath, parser.ConfigBasename(configFormat)))
 			for _, path := range all {
 				if exists, _ := helpers.Exists(path, hugofs.Source()); exists {
 					return errors.New(path + " already exists")
@@ -148,7 +148,12 @@ func doNewSite(basepath string, force bool) error {
 		hugofs.Source().MkdirAll(dir, 0777)
 	}
 
-	createConfig(basepath, configFormat)
+	in := map[string]interface{}{
+		"baseurl":      "http://replace-this-with-your-hugo-site.com/",
+		"title":        "My New Hugo Site",
+		"languageCode": "en-us",
+	}
+	createConfig(basepath, configFormat, in)
 
 	jww.FEEDBACK.Printf("Congratulations! Your new Hugo site is created in %q.\n\n", basepath)
 	jww.FEEDBACK.Println(`Just a few more steps and you're ready to go:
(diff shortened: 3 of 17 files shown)


bep:
>As you know, the conventional extension for YAML files is .yml (rather than .yaml).

I didn't know that.

MarkDBlackwell:
In 2014, the Stack Overflow question [Is it .yaml or .yml?](http://stackoverflow.com/questions/21059124/is-it-yaml-or-yml) measured Google search counts for YAML and YML to be 15,400 and 49,100 respectively.

Now, those figures are [4,110,000](https://www.google.com/search?q=yaml) and [6,010,000](https://www.google.com/search?q=yml). Furthermore, the "YAML" count is unfairly high because it includes mention of the markup language outside its use as an extension.

Searching this repository's issues for [yml](https://github.com/spf13/hugo/search?q=yml&type=Issues) results in 21 hits. Therefore a number of people are using ".yml" with this repository, despite its suggestion to use ".yaml".

The Wikipedia article on [YML](https://en.wikipedia.org/w/index.php?title=YML&oldid=634957727) says ".yml" is _"the_ file extension for the YAML file format". Its [YAML](https://en.wikipedia.org/w/index.php?title=YAML&oldid=732701368) article lists both extensions.

Of course, both of these extensions are abbreviations of a long, possible extension, ".yamlaintmarkuplanguage". But programmers (and users) don't want to type all that! In fact, we want to type as little as possible, and yet still be unambiguous and clear. Typing how many characters accomplishes both these goals?

Any suggestion that this (real life usage) preference was caused by a 8.3 character DOS-ish limitation IMO is a [red herring](https://en.wikipedia.org/w/index.php?title=Red_herring&oldid=732783672) (erroneous and misleading).

Instead, generally, few extensions start with "Y". On the other hand, many extensions start with "J" (for "Java") or "X" (for "cross", "extensible", "extreme", etc.). So starting with "Y" already conveys much information.

Perhaps this is why we see in practice that the "linguistic" pressure to lengthen the abbreviation in question to four (4) characters is weak, and the "linguistic" pressure to shorten this abbreviation to three (3) characters is strong.

Apparently, on balance, as a result, the abbreviated, three (3) character extension is the one predominantly in use for YAML, despite the inventor's preference. Well—he can't be right about everything. ;/

bep:
I agree abut the 3 letter extension being good, so this looks good to me.

MarkDBlackwell:
Some command (I think `hugo new {something}`) generated an extension `.yaml` file (I noticed).

So I added 'WIP' to the title. I plan to investigate further.

MarkDBlackwell:
Now, commands:

1. `hugo new site --format yaml`; and
1. `hugo import jekyll`

both generate `config.yml`—I added a test.

This PR adds two (2) tests.

MarkDBlackwell:
> remove the edits below docs/content/commands/

Removed.

whereisaaron:
Nice troll @MarkDBlackwell, I had to check my calendar :-)

You're claiming that users voted with their feet to move to a three-letter contraction of the acronym, not in any part from the decades of influence of the three-letter extension limitation of CP/M and DOS. Which is still the convention in Windows, the overwhelmingly dominant desktop OS. You claim that has no influence, a 'red herring' even (the link is to the definition BTW, not to any data or justification). And instead the dominant, overriding, heartfelt concern of users was to save that keystroke?

By your own Google data, the balance is shifting. It looks like by about 2018 you'll be able to come back here and use the _exact same argument_ to recommend a change back to '.yaml'.

> "YAML" count is unfairly high because it includes mention of the language by name

Woah, how confusing, a file extension that mentions the language by name  :-P

@bep think how many keystrokes we could save if we change all our '.go' files to '.g'. I know that is not the inventors' preference or the standard. "Well—they can't be right about everything. ;/"

@MarkDBlackwell better get to work on your crusades for '.json' to '.jsn' and '.toml' to '.tml'. It's not the Internet if we don't drp the vwls :-)

I'd prefer a less populist argument, but neither is it important either way, so, like, whatever dude 🐹

MarkDBlackwell:
@whereisaaron: Actually (as the PR's title states), extension ".yaml" would still work.

Hugo [accepts both](https://github.com/spf13/hugo/blob/661c9787ea9bd005375ca295cf50ed602ff342af/commands/hugo.go#L207) ".json" and ".js" for JSON; ".yaml" and ".yml" for YAML; and ".toml" and ".tml" for TOML.

Please be reassured that the changes in the PR wouldn't prevent anyone from using the extension ".yaml" for their YAML-formatted Hugo files, if they so choose.

Certainly, either extension any particular user would like to use for their YAML-formatted files is fine with Hugo.

The PR doesn't advise people *against* using the ".yaml" extension. Does Hugo's documentation need more language stating that the extension ".yaml" is fine to use as well? I believe I've already noticed something, along these lines.

The purpose of the PR is to aid (in a small way) Hugo's evangelism by removing a needless, disconcerting discrepancy between (on the one hand):

1. Hugo's documentation and generated files; and (on the other hand)
1. The predominant practice IMO of developers who aren't yet familiar with Hugo.

It aids this evangelism by:

1. Changing Hugo's documentation, from recommending the extension ".yaml", to recommending ".yml"; and
1. Changing Hugo, from generating ".yaml", to generating ".yml" files.

Again, this is in line with the practice IMO predominant among developers, especially those outside the world of Hugo.

It also removes a possible perception that Hugo is simply unaware of which extension (for YAML files) is predominant in actual use. Unjustly, that perception might lessen Hugo's reputation slightly among potential adopters.

BTW, Go is IMO the C-related language most attractive to developers who use dynamic languages, due to Go's modern language features.

For evangelistic purposes, we should match those developers' expectations, wherever doing so doesn't hurt us.

The PR is all about a comfort thing; it will remove a small hurdle for new Hugo developer-users IMO. As the saying goes, "Every little bit helps."

whereisaaron:
I agree the PR is harmless and doesn't prevent people using either extension as they wish. If @bep is for it, then I respect his decision.

I personally think your arguments for the change are specious, and part of a [long-standing personal campaign](http://markdblackwell.blogspot.com/2013/07/use-file-extension-yml-for-yaml.html) rather than based on any actual evidence that Hugo users are confused by the .yaml extension. I respect your opinion and personal preference, but disagree with your reasoning. Also, sorry for calling you a troll, I can see you genuinely believe what you say.

bep:
This change is not important enough compared to the amount of changes needed.

Issue #2328, opened 2016-08-03
Templates should be able to convert timestamps to Unix time


waddles:
The `dateFormat` function takes a timestamp string and formats it according to fields provided in the layout example, however there is no way to provide a format for Unix time. The return value from `dateFormat` is always a reformatted string.

Templates should be able to directly use fields of a `time.Time` structure, to be able to do things like:
```
---
start: 2016-08-03
end: 2016-08-05T07:04:03Z
```
```
{{ $start := (time .Params.start) }}
{{ $end := (time .Params.end) }}
{{ $start.Unix }}
{{ $end.Local }}
```
It might also be cool to do something like:
```
{{ (sub $end $start).Minutes }}
```

bep:
>{{ (sub $end $start).Minutes }}

Could you create another issue for that. I agree that it would be cool. Both add and sub should support time.

PR #2329, opened 2016-08-03
tpl: Add time template func

waddles:
First attempt at fixing #2328


digitalcraftsman:
@waddles could you also add a few words to the documentation about the [template function](https://github.com/spf13/hugo/blob/master/docs/content/templates/functions.md)?

PR #2335, opened 2016-08-06
docs: Improve Quickstart Guide

### LINKED ISSUE #2331: Quickstart Guide broken at Step 8
The Quickstart Guide seems broken at [Step 8. Customize robust theme](https://github.com/spf13/hugo/blob/914e4d85350521f417c8b844824c73398fa9561e/docs/content/overview/quickstart.md#step-8-customize-robust-theme). The Guide's recommended actions don't change the appearance of the browser, fully as the Guide claims.

The Guide is correct, through the point of code evolution where the browser tab appears [as here](https://gohugo.io/img/quickstart/bookshelf-new-default-image.png) (including the post title and text).

In particular (in the next step), the Guide says, "We have removed details of the book so that only image is shown" and says the results should be [as here](https://gohugo.io/img/quickstart/bookshelf-only-picture.png). However, in actuality, the post's title and text remain, just as before. (Other test alterations I tried to the new, overriding file `bookshelf/layouts/_default/li.html` did appear in the browser.)

To a Hugo novice, this seems mysterious.

Of course, this brokenness might lessen Hugo's uptake of new users. (It frustrates them at a critical, early moment: when they're following the Quickstart Guide.)

I speculate this breakage may be from:

1. Changes in the Hugo executable's functionality; or
1. Changes in the theme, [hugo_theme_robust](https://github.com/dim0627/hugo_theme_robust).

This kind of problem IMO would seem to diminish if somehow the `spf13/hugo` repository itself were to version control some Hugo theme for the Quickstart Guide to use (just as it version controls the guide itself). This isn't true already (supposedly).

Seemingly, another solution would lie in changing the Quickstart guide so that it specifies checking out the above theme from a particular commit known to work (rather than from its latest commit—assuming the theme is the problem.) For example:

```bash
    $ git clone https://github.com/dim0627/hugo_theme_robust.git
    $ cd hugo_theme_robust
    $ git checkout b8ce466
```

That approach might be explainable to new users.

As to my Hugo version, I built Hugo from pull request commit 689e7e19e56ae5bd0e50de1fdddaffaad4299013 forked at d7f364c27e3adc4e012ee07e71f1c5ad5cfeb56f (but IMO that seems unlikely to have caused the problem).

    $ hugo version
    Hugo Static Site Generator v0.17-DEV BuildDate: 2016-08-03T14:45:33-04:00

OS: Windows XP, Git for Windows

MarkDBlackwell:
Without these changes, the Quickstart Guide fails to work.

I followed the Quickstart Guide as amended here, and now it all works. Fixes #2331.

### DIFF
--- a/docs/content/overview/quickstart.md
+++ b/docs/content/overview/quickstart.md
@@ -1,5 +1,5 @@
 ---
-lastmod: 2016-02-22
+lastmod: 2016-08-19
 date: 2013-07-01
 linktitle: Quickstart
 menu:
@@ -27,7 +27,13 @@ appropriate version for your OS and architecture.
 
 Save the main executable as `hugo` (or `hugo.exe` on Windows) somewhere in your `PATH` as we will be using it in the next step.
 
-More complete instructions are available at [Installing Hugo]({{< relref "overview/installing.md" >}})
+More complete instructions are available
+at [Installing Hugo]({{< relref "overview/installing.md" >}}).
+
+If you're on Windows, this quickstart will assume
+you're using [Git Bash](https://git-for-windows.github.io/)
+(also known as Git for Windows).
+Thus all commands will begin with the Bash prompt character (which is `$`).
 
 Once `hugo` is installed, make sure to run the `help` command to verify `hugo` installation. Below you can see part of the `help` command output for brevity.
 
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Excuse the delay. I've merged your pull request as 0614a59b0f35344987062133cfab67520e88e3b8.

Cheers,
Digitalcraftsman

Issue #889, opened 2015-02-10
safeHtml throwing ERROR: wrong type for value; expected string; got float64


thomasmodeneis:
Hi,

I have the following problem:

tst_float64.md 
```
xyz blz bla 12.866.431; 27,3% xyz bla bla.
```

tst.html
```
{{ .tst64 | safeHtml }}
```

ERROR:
```
ERROR: 2015/02/10 Error while rendering page tst_float64.md: template: tst.html:32:59: executing "tst.html" at <safeHtml>: wrong type for value; expected string; got float64
```
Does anyone knows witch part should I look for possible solutions?
Thank you.

bep:
I don't know where to look for the problem ...

What's `.tst64`?

Apparently a float64.

Maybe printf could work around this, but safeHtml should maybe accept more than just strings.

thomasmodeneis:
.tst64 is just a field inside the .md file that contains a float64 in the middle of the string.
I will try printf as workaround when I get home, thanks for the tip.

bep:
safeHTML now accepts "anything", including floats.

PR #2352, opened 2016-08-12
Feature-Request: Make absURL and relURL accept anything

kulla:
The template functions `relURL` and `absURL` currently only accept strings. However, after #889 many template functions like `safeHtml` were opened to accept more than strings. So I suggest that also `relURL` and `absURL` should share this behavior. For example the following code currently throws an error

    {{ .RSSLink | relURL }}

because `.RSSLink` is of type `template.HTML` (it throws an error with my version `Hugo Static Site Generator v0.16-DEV BuildDate: 2016-02-06T18:14:17+01:00` shipped with Ubuntu 16.04).

I followed 3a2a4c3b07f . Note that I never programmed in go before, so someone should check my changings :-) I am also not sure whether [code injections](https://en.wikipedia.org/wiki/Code_injection) are possible after the changes in my commit...

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1803,7 +1803,7 @@ func htmlUnescape(in interface{}) (string, error) {
 
 func init() {
 	funcMap = template.FuncMap{
-		"absURL":       func(a string) template.HTML { return template.HTML(helpers.AbsURL(a)) },
+		"absURL":       func(a interface{}) template.HTML { return template.HTML(helpers.AbsURL(cast.ToString(a))) },
 		"add":          func(a, b interface{}) (interface{}, error) { return helpers.DoArithmetic(a, b, '+') },
 		"after":        after,
 		"apply":        apply,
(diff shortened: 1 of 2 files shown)


bep:
Agree.

tpl/template_funcs.go
```diff
@@ -1803,7 +1803,7 @@ func htmlUnescape(in interface{}) (string, error) {
 
 func init() {
 	funcMap = template.FuncMap{
-		"absURL":       func(a string) template.HTML { return template.HTML(helpers.AbsURL(a)) },
+		"absURL":       func(a interface{}) template.HTML { return template.HTML(helpers.AbsURL(cast.ToString(a))) },
```
bep: Please use the ToStringE variant and return any error as second return value from the template func. Same with relURL.

moorereason: He learned from the best. :wink:  Do the changes in https://github.com/spf13/hugo/commit/3a2a4c3b07f82527765cbeaa33f53ab7e2314491 need errors returned, too?

bep: @moorereason please stick to this PR. Taking the entire commit history of Hugo into consideration for this simple change is ... just stupid.

bep:
We'll handle the errors as part of https://github.com/spf13/hugo/issues/2354

kulla:
Thanks :smile:

PR #2357, opened 2016-08-13
Correct "it's" and "its"

MarkDBlackwell:
Now in the entire repository all uses of "it's" and "its" are correct.


digitalcraftsman:
Once again thanks for the submission. Merged as be429d22681f9486b9ee47b53b93c3beb97e3b17.

PR #2365, opened 2016-08-15
#2364 Exposed funcMap as FuncMap

### LINKED ISSUE #2364: tpl: Expose funcMap as FuncMap
Hugo has the best template funcmap that I've seen and to be able to use it in any other Go project I'm faced with either:

1. Fork Hugo, change visibility of funcMap to FuncMap, vendor forked Hugo
2. Copy the bits of the funcMap I need, attempt to ensure all licenses and credits are copied, maintenance nightmare
3. Ask if Hugo would consider just changing the visibility of funcMap
4. Extract tpl into it's own project to supply improved templating and functions to all Go projects

I'm thinking option 3 is the most trivial for now and achieves the goal. The license already permits the code re-use, but the default unexposed visibility does not. This change would be just to make the code follow the spirit of the license.

PS: My usage scenario is that I'm porting a Django project to Go and need a few simple functions within my templates, and whether it's `dfault` or `in` Hugo has the best so it makes sense to use these.

buro9:
This allows other Go projects to use the template functions from Hugo whilst
still allowing those projects to load and manage templates however they wish.

Would resolve #2364

### DIFF
--- a/tpl/template.go
+++ b/tpl/template.go
@@ -94,10 +94,10 @@ func New() Template {
 
 	localTemplates = &templates.Template
 
-	for k, v := range funcMap {
+	for k, v := range FuncMap {
 		amber.FuncMap[k] = v
 	}
-	templates.Funcs(funcMap)
+	templates.Funcs(FuncMap)
 	templates.LoadEmbedded()
 	return templates
 }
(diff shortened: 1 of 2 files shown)


bep:
See comment in #2364

PR #2370, opened 2016-08-17
Fix typos in tests

moorereason:


alenbasic:
Looks good to me, but I would probably change the "something error occurred" to "some error has occurred" or similar. Might as well fix it all up at the same time.

hugolib/pageGroup_test.go
```diff
@@ -149,7 +149,7 @@ func (page *Page) DummyPageMethodReturnThreeValueForTest() (string, string, stri
 }
 
 func (page *Page) DummyPageMethodReturnErrorOnlyForTest() error {
-	return errors.New("something error occured")
+	return errors.New("something error occurred")
```
bep: "some error occured" or similar.

moorereason:
Fixed.  Now using "some error occurred".

PR #2371, opened 2016-08-17
hugolib: Use named keys in composite literals

moorereason:
Make `go vet` great again


bep:
Please wait with this, I have some massive commits in this area, and I'm not too keen on the merge job.

bep:
OK, the go vet errors was more annoying than a merge ... :-)

PR #2372, opened 2016-08-17
docs: Fix URLs in examples for sites with path prefix.

kulla:
When baseurl ist something like "http://example.com/hugo/" a link to
"/tags/..." does not work. Therefore I fixed the examples in http://gohugo.io/taxonomies/displaying/ so that they also work in cases where the website is not saved at the server's root.
See also https://discuss.gohugo.io/t/how-shall-i-link-a-taxonomy-page/3920 for more details.

### DIFF
--- a/docs/content/taxonomies/displaying.md
+++ b/docs/content/taxonomies/displaying.md
@@ -38,7 +38,7 @@ each content piece are located in the usual place
 
     <ul id="tags">
       {{ range .Params.tags }}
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Hello 

thank you for submitting this addition. I've merged your contribution as de1ca477b8402ad195b121d60eddb2c3e4d2f506.

Cheers,
Digitalcraftsman

PR #2375, opened 2016-08-19
Fix typos in tests

bogem:


hugolib/pageGroup_test.go
```diff
@@ -149,7 +149,7 @@ func (page *Page) DummyPageMethodReturnThreeValueForTest() (string, string, stri
 }
 
 func (page *Page) DummyPageMethodReturnErrorOnlyForTest() error {
-	return errors.New("something error occured")
+	return errors.New("something error occurred")
```
bep: If you want to fix this and the similar below, you should fix both language errors in the same sentence.

bogem: Excuse me, but I don't understand you :(

bep: "something error occurred" is still not valid English.

moorereason:
This is a dupe of #2370 

@bep, I made the same edits and left the strange error message in place.  I thought about changing it, but it's not that relevant because it's a bogus error that is expected to be returned.  The error returned is never really seen.  That test code originally came from user tatsushid.  If you still think we should change it, say so in #2370, and I'll fix it.

PR #2387, opened 2016-08-27
tpl: Add template function 'asciidocify'

lafisrap:
I added 'asciidocify' function according to the 'markdownify' function. Also added a test.

`go build` and `go test ./...` show no errors.

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1218,6 +1218,18 @@ func markdownify(in interface{}) (template.HTML, error) {
 	return template.HTML(m), nil
 }
 
+// asciidocify renders a given string from Markdown to HTML.
+func asciidocify(in interface{}) (template.HTML, error) {
+	text, err := cast.ToStringE(in)
+	if err != nil {
+		return "", err
+	}
+	m := helpers.RenderBytes(&helpers.RenderingContext{Content: []byte(text), PageFmt: "asciidoc"})
+	//m = bytes.TrimPrefix(m, markdownTrimPrefix)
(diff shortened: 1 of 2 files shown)


lafisrap:
I checked the AppVeyor Errors, but don't understand. I'd need help there, I guess.

Wercker doesn't show an Error log.

moorereason:
https://ci.appveyor.com/project/spf13/hugo/build/1.0.3652#L786

```
=== RUN   TestFuncsInTemplate
ERROR: 2016/08/27 09:17:10 content.go:468: asciidoctor / asciidoc not found in $PATH: Please install.
                  Leaving AsciiDoc content unrendered.
--- FAIL: TestFuncsInTemplate (0.02s)
	template_funcs_test.go:245: Diff:
		["asciidocify: **BatMan**" "<p><strong>BatMan</strong></p>" "</div>" "asciidocify: <div class=\"paragraph\">"]
```

lafisrap:
Ok. Then its not a problem of the new code, but that the asciidoc functionality needs asciidoctor installed on the test system (gem install asciidoctor). Where are the tests running?

bep:
@lafisrap the solution would be to conditionally run the test if assciidoctor is installed (see `helpers.HasAsciidoc` or something). But that is not the main problem with this PR. Having asciidoc running once per content page is slow enough, but having it available as a template func may be pushing it, with one hardware thread per func invocation.

bep:
I will close this now. We might revisit this in the future when we get native Asciidoc support.

PR #2395, opened 2016-08-29
Title strings with enclosing # will truncate permalinks

kylehq:
I found that using a title in my permalinks would truncate the link created if the title contained the # symbol - the url parser deems it to be valid as expected being a fragment. Therefore I've proposed this fix to remove the value on pageToPermalinkTitle() call.

I'm not sure how far up the chain this should go - eg are there .Site.Permalink values I should be concerned with also? I'm not sure but altering this method fixes the issue so far as I can tell. 

Also as a seperate issue, you can create `new` files with # values in the filename. I think they should also be stripped. I can create this PR also if required.

My setup 

```
# config.toml
[permalinks]
    test    = "/:title/"
```

Then alter any frontmatter content file to have a `#` in the title 

```
+++
categories = []
date = "2016-08-29T07:50:07+03:00"
draft = true
tags = []
title = "a new #1 file post that stinks like #2s"
+++
```

Then ensure that you are spitting out the permalink 

```
<a href="{{ .Permalink }}">{{ .Title | markdownify }}</a>
```
### Before with url truncation.

screen shot 2016-08-29 at 9 08 41 am
### After - no truncation on href and a happy url =]

after

`go test -run TestPermalink -v` shows permalink tests are aok.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ install:
   - make govendor
 script:
   - make check
+  # Test 64-bit alignment on 32-bit builds
+  - env GOARCH=386 make test
   - go build -race
   - ./hugo -s docs/
   - ./hugo --renderToMemory -s docs/


kylehq:
NOTE: I have already submitted my CLA - where the `Kyle Clarke seems not to be a GitHub user` is coming from is anybody's guess. How else did I create this PR without being a Github user?

digitalcraftsman:
GitHub uses email addresses to identify users. It seems you committed with an email address that is unknown to Github. Just add this to the emails in the settings and confirm it. This way, GitHub and the CLA bot can associate the commit with you.
 
This solved the problem for previous pull request submitters.

kylehq:
Thanks @digitalcraftsman - I have moved to Gitlab now so that address was my default and assigned in the PR. I have verified now =]

kylehq:
Ok this PR needs more work, I managed to break permalinks again by having a ' in the title of my blog. So all non url chars AND the # should be stripped. 

I also think I found another bug where using a number with Taxonomy failed to compile with
```
[taxonomies]
    tag = "tags"
    category = "categories"
    series = "series"

<ul>
  {{ range .Site.Taxonomies.series.29days }} <-- Goes kaboom due to 29
    <li><a href="{{ .URL }}">{{ .Name }}</a></li>
  {{ end }}
</ul>
```
I will confirm and lodge/PR.

thiagomgd:
@kylehq I think that you can create a new function to strip special characters, and then remove `#` and any other character that may break the links.

I don't know if it's related, but if I create a post titled "Hey Mr. John Doe", the {{ .Permalink }} contains the dot, but, of course, I get a 404 when I click it. But removing it manually from the url at the address bar gets me to the correct post.

moorereason:
@kylehq,
I believe your issue with `.29days` is a characteristic of Go templates.  You need to do something like this:

```
{{ range (index .Site.Taxonomies.series "29days") }}
```

kylehq:
@thiagomgd that's exactly related, I managed to break my posts again by describing `America` as `` `Murica`` =] I will create (or confirm a method doesn't already exist) to do just that.

kylehq:
@moorereason I didn't know that thanks - but I suggest we change the docs to reflect this https://github.com/spf13/hugo/blob/master/docs/content/taxonomies/displaying.md#2-listing-content-with-the-same-taxonomy-term

I will create another PR.

kylehq:
I'm going to have to park this for a bit - first first thoughts were to create a regexp to remove any chars & html entities that are not url safe. Excluding spaces in order to get a text representation of a title value that could be urlize'd.

But then good ol unicode comes along and `title: ラーメン` hmmmm

A bit blocked on how best to tackle this 😒

kylehq:
Ok so CR is required. But I learned that you can do unicode regexp's today - awesome!

moorereason:
@kylehq,
Can you squash and resolve merge conflicts?

PR #2397, opened 2016-08-30
docs: Update documentation for URLs

sethmacleod:
Using absolute URLs overrides relative URL behavior. This should be explicit in the docs.


digitalcraftsman:
Thank you for the addtion.

PR #2400, opened 2016-08-30
Some fixes

bogem:
Made some fixes according to `gometalinter`

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -617,11 +617,7 @@ func copyStatic() error {
 
 	// because we are using a baseFs (to get the union right).
 	// set sync src to root
-	err := syncer.Sync(publishDir, helpers.FilePathSeparator)
-	if err != nil {
-		return err
-	}
-	return nil
+	return syncer.Sync(publishDir, helpers.FilePathSeparator)
 }
 
 // getDirList provides NewWatcher() with a list of directories to watch for changes.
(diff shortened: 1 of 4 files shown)


moorereason:
@bogem,
Thanks for your contribution.  If you're able, please update your commit messages to denote the package.  For example, `commands: Delete unused newUserErrorF` and `parser: Use strings.Contains instead of strings.Index`.

More info in our [commit message guidelines](https://github.com/spf13/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines).

bogem:
Ok, I'll do it tomorrow

bogem:
Done

moorereason:
That's better.  Thanks.

LGTM except, perhaps, removing `newUserErrorF`.  I'm conflicted.  @spf13, what do you think?

bogem:
It's private function and it's never used in projects. I run tests and everything is working

commands/hugo.go
```diff
@@ -77,10 +77,6 @@ func newUserError(a ...interface{}) commandError {
 	return commandError{s: fmt.Sprintln(a...), userError: true}
 }
 
-func newUserErrorF(format string, a ...interface{}) commandError {
```
bep: It is not used, but it completes a natural four method API, and it is just a question of time until it is getting used, so don't remove it.

bogem:
Okay, later I will restore that

bep:
@bogem never mind,  I see it has been enough chatter in this PR already. I'll merge.

bogem:
Thanks!

Issue #1252, opened 2015-07-07
Builtin pagination template doesn´t work with sub root in BaseURL and CanonifyUrls = "false"


donlzx:
If .Site.BaseURL contain paths (e.g., example.com/en/), pagination doesn't work correctly.  The variables URL, Next.URL, Prev.URL, etc. return incorrect values in the pagination templates, resulting in the following behavior:

http://example.com/en/products/page/2/ (correct link) => http://example.com/products/page/2/ (links in generated pages)

I make it working by hacking the source code (hugolib/pagination.go) around line 276:

> urlFactory := newPaginationURLFactory(section)

modified as: 

> urlFactory := newPaginationURLFactory(helpers.RelURL(section))

Since I am new to Hugo and haven't looked at the source codes before, can someone check this issue and fix it in a correct way?

bep:
I'm pretty sure the Next.URL etc. are relative to `context root`, so the examples you give cannot be the string value returned. Can you check?

donlzx:
@bep Actually, I mean the final URL in the generated pages.

bep:
And how do you construct the "full URL"?

donlzx:
For testing, I'm just using the default internal template and nothing special:

```
  {{ range (.Paginator 20).Pages }}
       ....
  {{ end }}
  {{ template "_internal/pagination.html" . }}
```
The full URLs are constructed by the browser.  To be exact, the links to the pages in generated HTML looks like:

```
 <a href="/products/page/2/">2</a>
```
It should be:
```
<a href="/en/products/page/2/">2</a>
```

bep:
OK, now I understand it.

The current workaround is to either set `CanonifyUrls = "true"` or use a custom paginator template.

`helpers.RelURL` would be the right medicine, but I would do it with the `rel` variable here:

https://github.com/spf13/hugo/blob/master/hugolib/pagination.go#L367

donlzx:
@bep Using ```helpers.RelURL()``` with ```rel``` variable doesn't work for me, because ```rel``` variable is already of the form ```/section/page/... ``` and helper.RelURL()``` doesn't change it.

bep:
Hmm, that sound a little bit strange. I guess this should be a safe bet, too:

https://github.com/spf13/hugo/blob/master/hugolib/pagination.go#L273

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

[...]

PR #2403, opened 2016-08-31
hugolib: Make URLs in pagination relative.

kulla:
A fix for #1252 (works for me). It may be good to add a unit test for this change as well (I haven't enough experience with hugo to write it by myself...)


Philpax:
Have there been any updates on this? Using `CanonifyUrls` isn't the most ideal workaround for this.

bep:
Can you give some concrete samples that fails for you?

PR #2438, opened 2016-09-13
docs: Improve Traversing Local Files

MarkDBlackwell:
1. Clarify its English;
1. Remove an unnecessary [HTML style attribute](https://www.w3.org/TR/2002/WD-css-style-attr-20020515);
1. Give the shortcode a more accurate name;
1. Better distinguish filesystem paths from URL paths;
1. Mention, for function `readDir`, that relative and absolute paths have the same effect;
1. Correct a bug in the shortcode, whereby the links it provided were broken by conversion to lower case, for filenames containing capitalization such as `HPstyles.css`;
1. Remove source code duplication between the content file and the shortcode file; and
1. Provide a new shortcode which reads and displays any file (for now: just the main shortcode file).


digitalcraftsman:
thanks again for the improvements. Merged as 0d1e96701fd3907d0a3e15e068925fe42ba3f662.

PR #2457, opened 2016-09-16
tpl: Add EnableMissingTranslationPlaceholders option

### LINKED ISSUE #2451: i18n: return empty string by default when no match
If `i18n` finds no matches, it should return a nil value (empty string?) instead of the string "[i18n] %s". 

It could also simply return the identifier, but a nil value is idiomatic go and would let us check if  a match exists in template code, e.g. `{{ with i18n "some_identifier" }}`.

digitalcraftsman:
Fixes #2451

### DIFF
--- a/tpl/template_i18n.go
+++ b/tpl/template_i18n.go
@@ -82,6 +82,10 @@ func SetI18nTfuncs(bndl *bundle.Bundle) {
 					return translated
 				}
 			}
+
+			if !viper.GetBool("EnableMissingTranslationPlaceholders") {
+				return ""
+			}
 			return fmt.Sprintf("[i18n] %s", translationID)
 		}
 	}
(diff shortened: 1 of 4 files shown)


moorereason:
You need to `gofmt` `tpl/template_i18n.go`.

tpl/template_i18n.go
```diff
@@ -82,6 +82,10 @@ func SetI18nTfuncs(bndl *bundle.Bundle) {
 					return translated
 				}
 			}
+
+			if(viper.GetBool("EnableMissingTranslationPlaceholders")) {
```
bep: You will have to add this conditional above, too.

bep: if not ...

digitalcraftsman: Done

digitalcraftsman:
`gofmt` - done.

mitchchn:
Thank you @digitalcraftsman!

Issue #2119, opened 2016-04-30
Provide an extra zip/tar.gz release archive with Go vendoring included for reproducibility


anthonyfok:
Due to the dynamic nature of the Go library packages that Hugo depends on, our released source code in zip or tar.gz form could soon become unusable by the end user.  For example, it is probably very difficult for an end user to build hugo v0.15 from source from https://github.com/spf13/hugo/archive/v0.15.tar.gz without going back in time for some dependencies.  Even more difficult to re-create the exact same released v0.15 binaries from source.

I recommend providing an extra zip/tar.gz source archive with the vendoring information added, e.g., generated by the following commands:

```bash
export GOPATH=/tmp/hugo-gopath
govendor get -v github.com/spf13/hugo
cd $GOPATH/src/github.com/spf13
mv hugo hugo-0.16
GZIP=-9v tar cvzf hugo-0.16-full.tar.gz hugo-0.16
```

using govendor from https://github.com/kardianos/govendor.

bep:
But I think we should just start do proper vendoring and this will be solved.

anthonyfok:
Cc: @spf13 

@ bep wrote:
> But I think we should just start do proper vendoring and this will be solved.

Should we start by checking in vendor/vendor.json before the 0.16 release?
From https://github.com/kardianos/govendor/blob/master/doc/dev-guide.md:

> ## Always check in the "vendor/vendor.json" file
>
> You can add the ignore rule `vendor/*/` to ignore source files. This way a consumer of your package has a chance at reproducing your package tests if something appears to break later.

I think this would remedy the sad situation where https://github.com/spf13/hugo/archive/v0.15.tar.gz quickly became almost impossible to compile (within a month?) due to API changes in upstream packages.

bep:
Is that a standard vendoring format used by all the vendoring tools?

anthonyfok:
> Is that a standard vendoring format used by all the vendoring tools?

Good question, one that I hadn't thought of.

* Godep: Godeps/Godeps.json
* govendor: vendor/vendor.json
* gvt: vendor/manifest

So it looks like it differs among various vendoring tools.

bep:
And then there is Dave Cheney's GDB ... We should pick one, but not by random.

moorereason:
For reference see [Go best practices, six years in: Dependency Management](https://peter.bourgon.org/go-best-practices-2016/#dependency-management).

Random comments:

[...]

PR #2461, opened 2016-09-16
Add vendoring with govendor

moorereason:
Doesn't check in vendored packages.  Rewrites Makefile to use govendor helpers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,6 @@ COMMIT_HASH=`git rev-parse --short HEAD 2>/dev/null`
 BUILD_DATE=`date +%FT%T%z`
 LDFLAGS=-ldflags "-X github.com/spf13/hugo/hugolib.CommitHash=${COMMIT_HASH} -X github.com/spf13/hugo/hugolib.BuildDate=${BUILD_DATE}"
 
-DIRS=$(shell go list -f {{.Dir}} ./...)
-
 all: gitinfo
 
 install: install-gitinfo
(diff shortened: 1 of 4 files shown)


bep:
The vendor.json file; is that a standard format?

moorereason:
There is no standard format.  That's govendor's format.  The only standard is that the Go tools look for vendored packages in the `vendor/` directory.

moorereason:
Tagged for 0.17.  FIxes #2119.

bep:
I merged this as it does solve the problem ... but we should do an evaluation of the vendoring options for 0.18.

nathany:
When I read the [release notes](https://github.com/spf13/hugo/releases/tag/v0.17) for 0.17, I just expected the code under vendor/ to be checked in.

Should `make govendor` be mentioned in the README?

Makefile
```diff
@@ -34,39 +32,45 @@ docker:
 	docker cp hugo-build:/go/bin/hugo .
 	docker rm hugo-build
 
+govendor:
+	go get -u github.com/kardianos/govendor
+	go install github.com/kardianos/govendor
```
nathany: FYI

> "Get downloads the packages named by the import paths, along with their
dependencies. It then installs the named packages, like 'go install'." - `go get -h`

moorereason:
The Makefile has already been fixed in master.

And yes, we may need to update the README.

Makefile
nathany: Which is to say the `go install` line is superfluous. 

It still appears here atm: https://github.com/spf13/hugo/blob/master/Makefile#L37

moorereason: You're right.  I was looking at this on my phone and misunderstood where your comment was in the Makefile.  The `go install` does indeed need to be removed.

PR #2468, opened 2016-09-17
tpl: Fix i18n for missing translations and add tests

mitchchn:
`i18n` now follows the correct logic for missing translations when the default translation file is empty or missing. Previously it just reverted to the `go-i18n` behaviour and spat out the identifier.

I also added test cases for the `i18n` function for various kinds of missing translations.

I noticed one odd consequence of the `EnableMissingTranslationPlaceholders` flag introduced in https://github.com/spf13/hugo/pull/2457: even when it is set, placeholders do not appear if the translation is missing in the current language but present in the default language. It seems that with this new flag we would want to highlight those cases as well.

I think the ideal logic would be:

* _EnableMissingTranslationPlaceholders_: Look for a translation in the current language. If it is missing, always flag it, even if there is an available default translation.
* Default: Look for a translation in the current language. If it is missing, look for a default translation. If that is missing, return "".

All this would require is bumping the flag check up a few lines. Let me know what you all think and I can update this PR as necessary. cc: @digitalcraftsman @bep

### DIFF
--- a/tpl/template_i18n.go
+++ b/tpl/template_i18n.go
@@ -63,30 +63,26 @@ func SetI18nTfuncs(bndl *bundle.Bundle) {
 
 	for _, lang := range bndl.LanguageTags() {
 		currentLang := lang
-		tFunc, err := bndl.Tfunc(currentLang)
 
-		if err != nil {
-			jww.WARN.Printf("could not load translations for language %q (%s), will use default content language.\n", lang, err)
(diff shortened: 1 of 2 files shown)


bep:
>I think the ideal logic would be:

Yes, that makes sense to me.

mitchchn:
Updated the logic and tests. I think we now have a good, consistent set of behaviours for development and production. :)

Issue #2467, opened 2016-09-17
Feature request: multilingual author profiles


digitalcraftsman:
Some content creators might blog in multiple languages. Consequently I would be nice to translate certain author informations as well, like in an "About me" section at the end of an article. 

Otherwise, the content creators
1. are either bound to a single language
2. have to maintain multiple author profiles with redundant information
3. or have to use params to create some kind of workaround

Alternatively, authors could add language-specific containers to their author's profile in which they can overwrite default values - much like the current multilingual setup in config files. 

Authors, whose native language doesn't use the latin alphabet could now write their name with a different character set, e.g. in Chinese, Arabic or Japanese.

Example:

``` toml
# file: data/_authors/john.toml

# The following information are the default values
givenName    = "John"
familyName   = "Doe"
displayName  = "John Doe"
shortBio     = "I blog about interesting stuff"
bio          = "An extended biography..."

# de has to match the language prefix from the config file
# shortBio and bio of this section replace the values above in the German version
[languages.de]
shortBio     = "Ich blogge über interessantes Zeug"
bio          = "Ausführliche Biographie..."
```

derekperkins:
@digitalcraftsman Look at #1850, it uses the Node/Page lang to override the base language properties if they are set.

jmooring:
`.Page.Author` and `.Page.Authors` were [deprecated a year ago](https://github.com/gohugoio/hugo/commit/097fd588ca2928056122991caaa879e835c5ba62) in favor of using the taxonomy system. For this use case, using the taxonomy system is much simpler than (a) language-specific site parameters, (b) translation tables, or (c) language-specific data files.

Working example:

```text
git clone --single-branch -b hugo-github-issue-2467 https://github.com/jmooring/hugo-testing hugo-github-issue-2467
cd hugo-github-issue-2467
hugo server
```

PR #2470, opened 2016-09-17
Multi-Lingual Authors

derekperkins:
@digitalcraftsman, is this ballpark what you're looking at for #2467?

Obviously the accessors would need be be altered to allow for overrides, so all the struct fields would have to be unexported and then turned into functions. Is there a `Node` language to use?

### DIFF
--- a/hugolib/author.go
+++ b/hugolib/author.go
@@ -42,20 +42,57 @@ func (a Authors) Get(id string) Author {
 // Author contains details about the author of a page.
 type Author struct {
 	ID          string
-	GivenName   string            // givenName OR firstName
-	FirstName   string            // alias for GivenName
-	FamilyName  string            // familyName OR lastName
-	LastName    string            // alias for FamilyName
-	DisplayName string            // displayName
-	Thumbnail   string            // thumbnail
-	Image       string            // image
-	ShortBio    string            // shortBio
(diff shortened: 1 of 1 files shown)


derekperkins:
@bep Even if you don't want to tackle the multi-lingual piece of this now, I think that we should use functions instead of fields for the 0.17 release, and I also realized that I wasn't mapping weight in, so that has to get merged before release.

bep:
This all tells me we also needs some tests for all of this.

bep:
Also, functions just for the sake of a function just adds lines of code. If we add some side effects to these fields later, we can replace the field(s) with functions without the clients knowing it.

My main point being: Solving a concrete problem by also solving a set of non-problems is suboptimal.

derekperkins:
Ok, now they aren't functions just for the sake of functions. I'm just trying to understand if this is the right direction to take this PR for multi-lingual author support.

digitalcraftsman:
Maybe @ bep has a better approach to tackle this problem. The current approach requires that we have to pass the current language as argument to the accessors. Couldn't we instead read the current language from viper?

derekperkins:
@digitalcraftsman I closed this in favor of #1850, where `Author` and `Authors` always return localized data, without having to provide a language parameter.

Issue #2469, opened 2016-09-17
Author profiles aren't accessible in multlingual setup


digitalcraftsman:
If I add the configuration for another language then the author profiles can't be accessed inside the templates. `.Site.Authors` for example remains an empty list. Have a look on the "Authors" header on the homepage.

The author profiles return if I [comment out](https://github.com/digitalcraftsman/hugo-multilingual/blob/master/config.toml#L45-L57) the config for the German translation.

bep:
@derekperkins

derekperkins:
I started my PR long before languages became a thing, and to be honest, I don't know anything about the implementation. It's just loading from data files inside `data/_authors`. Is that path being redirected somehow?

On that note, right now there isn't any support inside of author profiles for anything multi-lingual. How would you anticipate wanting to interact with it? Looking at the setup, it seems like most fields would be redundant if you nested them inside languages, but that might be the easiest way to go about it. Obviously I'd prefer to mimic however multi-lingual is handled elsewhere in Hugo.

```
[languages.en]

given_name      = "Derek"
family_name     = "Perkins"
display_name    = "Derek Perkins"
thumbnail       = "img link"
image           = "img link"
short_bio       = "super short bio"
bio             = "Here's Derek's awesome bio"
email           = "derek@email.com"

[social]
    facebook    = "derekperkins"
    github      = "derekperkins"
    twitter     = "derek_perkins"
    googleplus  = "DerekPerkins1"

[params]
   random       = "whatever you want"

[languages.de]

given_name      = "Derek"
family_name     = "Perkins"
display_name    = "Derek Perkins"
thumbnail       = "img link"
image           = "img link"
short_bio       = "super short bio in German"
bio             = "Here's Derek's awesome bio in German"
email           = "derek@email.com"

[social]
    facebook    = "derekperkins"
    github      = "derekperkins"
    twitter     = "derek_perkins"
    googleplus  = "DerekPerkins1"

[params]
   random       = "whatever you want"
```

derekperkins:
Sorry, I didn't see #2467 before responding, I'll check that out.

bep:
#2467 has nothing to do with this feature. That is "future work".

derekperkins:
I disagree, it's clearly related since something multi-lingual is what broke it in the first place. It doesn't mean that #2467 has to happen right now, but if I'm going to go in and investigate, I'd prefer to at least think about how to handle multi-lingual authors.

moorereason:
[...]

PR #2471, opened 2016-09-17
Fix Author data for multi-lingual sites

derekperkins:
Should fix authors breaking in multi-lingual sites

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -103,6 +103,7 @@ type Site struct {
 	timer          *nitro.B
 	targets        targetList
 	targetListInit sync.Once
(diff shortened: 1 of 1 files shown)


derekperkins:
@digitalcraftsman Can you test this commit out and see if it fixes #2469?

derekperkins:
@bep This would only get caught in a full site integration test. `func (s *Site) readDataFromSourceFS()` used to only get called once, but I believe it's calling it once per language now.

digitalcraftsman:
`.Site.Authors` is still an empty array if I add a second language.

PR #2472, opened 2016-09-17
Map weight into Author

derekperkins:
While the default sort uses Author.Weight to sort, it wasn't ever read in from the data file.


derekperkins:
Closing since I included this fix in #2473

PR #2478, opened 2016-09-18
Add .Site.Author map

### LINKED ISSUE #2464: Breaking API-change re Author
From my own site:

```
ERROR: 2016/09/17 14:38:27 general.go:212: Error while rendering taxonomy category rss: template: _default/rss.xml:7:54: executing "_default/rss.xml" at <.Site.Author.email>: can't evaluate field Author in type *hugolib.SiteInfo
ERROR: 2016/09/17 14:38:27 general.go:212: Error while rendering taxonomy tag rss: template: _default/rss.xml:7:54: executing "_default/rss.xml" at <.Site.Author.email>: can't evaluate field Author in type *hugolib.SiteInfo
ERROR: 2016/09/17 14:38:27 general.go:212: Error while rendering section post rss: template: _default/rss.xml:7:54: executing "_default/rss.xml" at <.Site.Author.email>: can't evaluate field Author in type *hugolib.SiteInfo
ERROR: 2016/09/17 14:38:27 general.go:212: Error while rendering section page rss: template: _default/rss.xml:7:54: executing "_default/rss.xml" at <.Site.Author.email>: can't evaluate field Author in type *hugolib.SiteInfo
ERROR: 2016/09/17 14:38:27 general.go:212: Error while rendering homepage rss: template: _default/rss.xml:7:54: executing "_default/rss.xml" at <.Site.Author.email>: can't evaluate field Author in type *hugolib.SiteInfo
```

This needs to be fixed.

derekperkins:
Not sure where to log out deprecation notices, since it isn't loaded by itself and access is through a map, not a function.

Fixes #2464

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -165,6 +165,7 @@ type SiteInfo struct {
 
 	BaseURL               template.URL
 	Taxonomies            TaxonomyList
+	Author                map[string]interface{} // deprecated and will be removed in 0.18
 	Authors               Authors
 	Social                SiteSocial
 	Sections              Taxonomy


bep:
Adding a random map on Site wasn't really what I asked for in the issue description -- and would not solve the API breakage problem described. 

This issue is the reason for the revert of the "author" pull request. I can understand the frustration, but that frustration would be pretty pale compared to my frustration when the "my site is broken" reports started to pile up.

derekperkins:
@bep I'm trying to give you the benefit of the doubt that you aren't trying to come across as condescending, but I'd appreciate it if you stopped treating me like I don't know what I'm doing. 

All this commit does is revert the change to exactly how it was before, so yes, it would solve the API breakage problem.
https://github.com/spf13/hugo/blob/master/hugolib/site.go#L179

bep:
I see 1 added line. Is that "exactly how it was before"?

derekperkins:
Yes, nothing else referenced that map, so we could leave it in forever and anyone using it would never have to know.

bep:
It was referenced, but anyhow: This and the related issues didn't shout **stability** to me.

derekperkins:
>Is that "exactly how it was before"?

That comes across as very condescending. 

Like any new feature, there are going to be some issues. There have been dozens of bug-fixes on the multi-lingual code in the last 2 weeks, which is expected. I submitted 2 PRs to fix the 2 reported issues with the Author code. Maybe it doesn't line up with your vision, which is fine and I respect, but you're not making people want to stick around by insulting a genuine attempt to solve a problem.

As one of the lead maintainers of https://github.com/chartjs/Chart.js, I'm very aware of what it takes to run a big open source repo, so I can especially appreciate the hard work you put in to keep Hugo going. That's why I understand the reasoning to revert the Author changes before this release, even if it is personally frustrating.

bep:
I don't even know what condescending means (I'm from Norway). I treat open source with the same amount of requirements to quality as for paying customers. 

So when I describe a solution with a func that returns the old map (and content) accompanied with a descriptive log message telling the user that he/she needs migrate to a new API, and what I get is an empty map ... Then I have no problem telling you that "this is not good enough." It wasn't meant as condescending, but I can get slightly sarcastic when I feel that people are wasting my time. I'm not sure what level of quality of you aspire in the `Chart.js` project, but that isn't really this project. You are right about the bug fixing on the multilingual feature, but I would say that that feature is more complex.

You normally do solid work, but in this latest author trilogy there has been too much jiggle (whatever that means) in my book. I could take the silent approach and just ignore the PRs I don't like, but that wouldn't be very fruitful.

derekperkins:
My only point is that there were 2 separate issues with authors: 

1. Caused by the multi-lingual changes - I submitted #2471 
1. I removed `.SiteInfo.Author map[string]interface{}` in the original PR, which caused people issues who assigned things to said map, so in this PR, I put the same map back in the same location. That's the only solution that won't break the existing API. Turning it into a function means that it would break for anyone who assigned anything into the map.

I'm not trying to waste your time, I just submitted what appeared to me to be the only way to maintain the existing API.

PR #2491, opened 2016-09-24
commands/server: fix for protocol relative baseurl

innerand:
fixURL() didn't add http scheme to protocol relativ baseurls resulting in
malformed server urls (something like //localhost//example.com/)


innerand:
Ignore this. Issue was already fixed.

Issue #2417, opened 2016-09-08
Update multilingual example


digitalcraftsman:
The current [multilingual](https://github.com/spf13/hugo/tree/master/examples/multilingual) example used workarounds to create an internationalized website. But since Hugo has [native i18n](https://github.com/spf13/hugo/blob/master/docs/content/content/multilingual.md) support (as part of Hugo v0.17) it would be consquently to update the example as well.

aubort:
I'd be happy to contribute, if it can wait a week or two before being contributed.

aubort:
I have started to work on this today and will update the code [here](https://github.com/aubort/hugo/tree/multilingual-example/examples/multilingual) as I make progress.

aubort:
Just a quick update on this, I've updated the code and have a very basic version of multilingual site based on the previous example. 
I've updated the branch (https://github.com/aubort/hugo/tree/multilingual-example/examples/multilingual). 

Happy to hear your comments/feedback. I am not sure how complex this example should be, but I still want to add the multilingual menu because I think it is important.

PR #2501, opened 2016-09-27
added dutch language to examples

fabulousduck:
added the dutch language to the examples to show how multiple languages can be used.


digitalcraftsman:
Hello @fabulousduck,

this version of the example will be deprecated soon. The next version of Hugo, which will be released  soon, integrates i18n as a core-feature. This example was designed as a proof-of-concept. However, the setup with the new multilingual integration will change.

Furthermore, this example will be updated with a new version that makes use of the upcomming multilingual feature (see #2417).

Nevertheless, thanks for your submission.

PR #2502, opened 2016-09-28
Fix Ace templates link

jasonm23:


digitalcraftsman:
Hello @jasonm23,

I checked the link with the HEAD of master and the links worked fine. Why did you removed the `relref` shortcode?

jasonm23:
I was viewing the docs through GitHub, and the other links were just valid markdown, and worked as expected.

My assumption is that the links should just work, everywhere, via any markdown processor.

digitalcraftsman:
The `relref` shortcode is part of the [cross-reference](https://gohugo.io/extras/crossreferences/) feature. Hugo allows you to completely rewrite the url of a content file. 

For example: links would break if you link to `post/foo.md` manually and somebody rewrites the url in the frontmatter. In this case the content isn't accessible under the manually defined url. `relref` would take things like this into account.

> My assumption is that the links should just work, everywhere, via any markdown processor.

The usage in the documentation is in some places inconsistent. In most places `relref` is used. If we follow the cited assumption we would have to remove `relref` everywhere. It would be nice if links would work with other processors -- but at the end it's a tradeoff.

jasonm23:
TBH the inconsistency is something that should be fixed.

`relref` _appears_ to be applicable to internal and external links.  Is that correct?

jasonm23:
(And `ref` for absolute refs!)

digitalcraftsman:
> TBH the inconsistency is something that should be fixed.

I agree

> `relref` _appears_ to be applicable to internal and external links.  Is that correct?

I've tested if external ressources are ignored. This is not the case. Hugo tries to link a content file based on the scheme that you to passed to `relref`. "https://google.com/foo/bar.html" in multiple variants always produced a error message that's displayed in the log.

> (And ref for absolute refs!)

Correct. `ref` prepends the base url.

jasonm23:
I will look at fixing some of these inconsistencies and submit pull requests for separate sections / documents as appropriate.

digitalcraftsman:
Any updates here?

digitalcraftsman:
Closing this as stall. Feel free to reopen if their are any updates for a pull request.

rdwatters:
@digitalcraftsman As an FYI, all good on POC, including aliases: https://hugodocs.info/templates/alternatives/#ace-templates

PR #2508, opened 2016-09-29
V0.16.docs

firmao:


anthonyfok:
Hi @digitalcraftsman!  Sorry, I have missed your message until today (following from @rdwatters’s #3175):

> @anthonyfok wouldn't it be more elegant to handle all articles about Hugo in a data file, similar to the tweets on the homepage. This way we don't have to deal with the (in my opinion) horrible Markdown syntax for tables.

That is a marvellous idea!  Yes, it is a great candidate for Hugo's data file feature!

PR #2511, opened 2016-10-01
Allow to set cache dir in config file

dmerejkowsky:
Use case:
I'm using `pygments`, so after every reboot, `hugo` re-runs `pygments` for _every_ code snippet in _any_ page of the blog, which takes quite some time.

Granted, I could set `TPMDIR` before running hugo, or use the command line flag, but I find it convenient to just have `cacheDir = cache` in the config file.
(And of course a matching entry in my blog's `.gitignore`)

Going further, maybe this could be the default?

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -370,6 +370,11 @@ func InitializeConfig(subCmdVs ...*cobra.Command) error {
 	}
 
 	if cacheDir != "" {
+		viper.Set("CacheDir", cacheDir)
+	}
+
+	cacheDir = viper.GetString("cacheDir")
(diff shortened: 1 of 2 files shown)


bep:
This sounds like a rare problem.

Have you tried to set a value for `CacheDir` in your config (or as HUGO_CACHEDIR env var)?

bep:
Your implementation is buggy. Please correct it and I will merge.

dmerejkowsky:
Oups you're right. Using `--cacheDir` from the command line was no longer working.

Fixed :)

PR #2513, opened 2016-10-01
doc: Improving language in archetypes doc

timrourke:
Improve language in the documentation for archetypes. As a new user of Hugo, this documentation is important to me, and I thought I would give it some love. I think making archetypes easy to understand is especially important for users coming from another blogging system that may have its own notion of how to define a "custom post type".


digitalcraftsman:
Hello @timrourke,

thank you for extending and improving the documentation of Hugo.

PR #2552, opened 2016-10-11
tpl: Add partialCached template function

### LINKED ISSUE #1368: Add PartialCached variant
See

http://discuss.gohugo.io/t/how-to-make-hugo-faster/1597/6

```
{{ range $.Site.Pages.ByTitle }}
  {{ if and (in .Params.categories "Category") (in .Params.categories "sub-category") }}
    <li><a href="{{ .Permalink }}">{{ .Title }}</a>
        {{ if .Draft }} * {{ end }}</li>
  {{ end }}
{{ end }}
```

The above is an extract of a partial used on every page. It is an ineffective construct (slightly better in 0.15, I guess) and it gets really bad when page count goes up.

But it's static. And, although it is impossible to make a general assumption about how "cacheable" a given partial is (I think?), the user will know -- and should be able to say so.

So, for a site with 2000 pages, the above should only be needed to be processed once.

@tatsushid not sure how Ace include works in this area?

moorereason:
Fixes #1368

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1392,6 +1392,52 @@ func replace(a, b, c interface{}) (string, error) {
 	return strings.Replace(aStr, bStr, cStr, -1), nil
 }
 
+// partialCache represents a cache of partials protected by a mutex.
+type partialCache struct {
+	sync.RWMutex
+	p map[string]template.HTML
+}
+
+// Get retrieves partial output from the cache based upon the partial name.
+// If the partial is not found in the cache, the partial is rendered and added
+// to the cache.
+func (c *partialCache) Get(key, name string, context interface{}) (p template.HTML) {
+	var ok bool
+
+	c.RLock()
+	p, ok = c.p[key]
+	c.RUnlock()
+
+	if ok {
+		return p
+	}
+
+	c.Lock()
+	if p, ok = c.p[key]; !ok {
+		p = partial(name, context)
+		c.p[key] = p
+	}
+	c.Unlock()
+
+	return p
+}
+
+var cachedPartials = partialCache{p: make(map[string]template.HTML)}
+
+// partialCached executes and caches partial templates.  An optional variant
+// string parameter (a string slice actually, but be only use a variadic
+// argument to make it optional) can be passed so that a given partial can have
+// multiple uses.  The cache is created with name+variant as the key.
+func partialCached(name string, context interface{}, variant ...string) template.HTML {
+	key := name
+	if len(variant) > 0 {
+		for i := 0; i < len(variant); i++ {
+			key += variant[i]
+		}
+	}
+	return cachedPartials.Get(key, name, context)
+}
+
 // regexpCache represents a cache of regexp objects protected by a mutex.
 type regexpCache struct {
 	mu sync.RWMutex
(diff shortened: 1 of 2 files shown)


moorereason:
@bep, can  you double-check my test code?  I used `InitializeT()`, but I'm not sure if that's the proper way to setup the template system.  Without it, I was getting a panic in `Lookup()` since `tmpl` is nil.

I would have copied the partial tests, but we don't have any!  :)

moorereason:
@bep, just noticed your proposed API.  I'll take a look at updating my PR.

bep:
Yes, my thought was to make it possible to use the same partial with different contexts without making a copy of the partial template.

tpl/template_funcs.go
```diff
@@ -1392,6 +1392,41 @@ func replace(a, b, c interface{}) (string, error) {
...
+
+// Get retrieves partial output from the cache based upon the partial name.
+// If the partial is not found in the cache, the partial is rendered and added
+// to the cache.
+func (c *partialCache) Get(name string, contextList ...interface{}) (p template.HTML) {
+	var ok bool
+
+	if p, ok = c.get(name); !ok {
+		p = partial(name, contextList...)
+		c.set(name, p)
```
bep: It may not have much practical difference in this case, but this lock construction with a `set` and `get` method makes it impossible to get the locking right -- and since people are going to copy this code into their university coding assignments, we might as well do it right.

* The creation of the partial isn't protected by a lock, so there will be multiple invocations of the same `partial`
* Also it would be good to double-check the map once you get the write lock; the partial may have been added in the open window between unlock and lock.

moorereason:
@bep, I updated the locking code.  I agree with your original comment.  I've moved the partial rendering into the `set` func.  Also reworked to support an optional variant parameter.

If you like what you see, I'll add some docs to the PR prior to merging.

tpl/template_funcs.go
```diff
@@ -1392,6 +1392,47 @@ func replace(a, b, c interface{}) (string, error) {
...
+	p := partial(name, []interface{}{context}...)
+	c.p[key] = p
+	c.mu.Unlock()
+	return p
+}
+
+var cachedPartials = partialCache{p: make(map[string]template.HTML)}
+
+func partialCached(name string, context interface{}, variant ...string) template.HTML {
+	key := name + strings.Join(variant, "")
```
bep: This is nitpicky, but I have seen some fairly heavy use of partials lately, so this would be better:

```go
key := name
if len(variant) > 0 {
   key += ...
}

File: tpl/template_funcs.go [hunk 3]
```diff
@@ -1392,6 +1392,47 @@ func replace(a, b, c interface{}) (string, error) {
...
+}
+
+func (c *partialCache) get(key string) (p template.HTML, ok bool) {
+	c.mu.RLock()
+	p, ok = c.p[key]
+	c.mu.RUnlock()
+	return
+}
+
+func (c *partialCache) set(key, name string, context interface{}) template.HTML {
```
bep: There is still a window between get/set where the partial could have been created by others. That is why I commented on the get/set construction. You can fix it by adding an additional map lookup inside the write lock, but then the set/get method names will be slightly misleading. As such I prefer reading the whole thing as one big code block. You would still need a double-check inside the write lock, but mistakes will be easier to spot.

Note that in normal cache situations (like a regexp cache), I may leave out the double-check because it doesn't matter if it gets invoked 1 or 2 times -- but in this case it has potential to get unexpected results (different partials).

tpl/template_funcs.go
```diff
@@ -1392,6 +1392,44 @@ func replace(a, b, c interface{}) (string, error) {
...
+	p  map[string]template.HTML
+}
+
+// Get retrieves partial output from the cache based upon the partial name.
+// If the partial is not found in the cache, the partial is rendered and added
+// to the cache.
+func (c *partialCache) Get(key, name string, context interface{}) (p template.HTML) {
+	var ok bool
+
+	c.mu.Lock()
```
bep: By double-checking i meant ... double. In this case you could use a regular mutex -- but I'm pretty sure proper use of a rwmutex will get much faster in this case.

So:

```
RLock
p, ok = m[key]
RUnlock

if ok return p

Lock

// double check
if p, ok = m[key]

else p = partial ...
Unlock

```

tpl/template_funcs.go
```diff
@@ -1392,6 +1392,44 @@ func replace(a, b, c interface{}) (string, error) {
 	return strings.Replace(aStr, bStr, cStr, -1), nil
 }
 
+// partialCache represents a cache of partials protected by a mutex.
+type partialCache struct {
+	mu sync.RWMutex
```
bep: I prefer to embed the mutex when the type is unexported and the mutex is so connected to the type, then you can do `c.Lock()` etc.

moorereason:
Travis was hung, so I squashed and rebased.

Issue #2399, opened 2016-08-30
Better info message when rendering Asciidoc


essen:
Hugo Static Site Generator v0.16 BuildDate: 2016-08-30T11:56:44+02:00

Sometimes Asciidoc errors out without any output. The problem is that Hugo does not tell us which file it is rendering, so finding where the error happens can be difficult:

```
INFO: 2016/08/30 12:40:44 content.go:476: Rendering with /usr/bin/asciidoc ...
ERROR: 2016/08/30 12:40:44 content.go:482: exit status 1
```

The verbose flags don't seem to help.

choeppler:
In #2567 there is a partial solution in that the errors in asciidoc / asciidoctor's stderr would now be reported by hugo:
```
Started building sites ...
ERROR: 2016/10/13 07:30:53 content.go:558: asciidoctor: ERROR: <stdin>: line 139: mismatched macro: endif::backen-docbook[], expected endif::backend-docbook[]
Built site for language en:
<snip/>
```
As the source filename is not available in the function `getAsciidocContent` which hands over the file contents to asciidoc / asciidoctor via stdin, reporting of which file is being rendered is still missing.
AFAICT -- being on a hacktoberfest stroll, I'm both very new to the used technology and hugo's codebase... -- the last place in the call stack where that information is still readily available is `commonConvert` in hugolib/handler_page.go. However, adding a log message there about which page is about to be converted obviously does not help much if more than one file is being converted concurrently and the log messages start to get interleaved. 
Thoughts?

PR #2567, opened 2016-10-13
Add logging of asciidoc/-tor errors

choeppler:
Add logging of the errors which asciidoc and asciidoctor output to their
stderr stream when converting asciidoc documents. Note that
asciidoctor's exit code may be SUCCESS even if there are ERROR messages
in its stderr output (tested with Asciidoctor 0.1.4 and 1.5.5).
Therefore log the stderr output whenever it is non-empty.

See #2399

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -546,9 +546,19 @@ func getAsciidocContent(content []byte) []byte {
 	jww.INFO.Println("Rendering with", path, "...")
 	cmd := exec.Command(path, "--no-header-footer", "--safe", "-")
 	cmd.Stdin = bytes.NewReader(cleanContent)
-	var out bytes.Buffer
+	var out, cmderr bytes.Buffer
 	cmd.Stdout = &out
-	if err := cmd.Run(); err != nil {
+	cmd.Stderr = &cmderr
+	err := cmd.Run()
+	// asciidoctor has exit code 0 even if there are errors in stderr
+	// -> log stderr output regardless of state of err
+	for _, item := range strings.Split(string(cmderr.Bytes()), "\n") {
+		item := strings.TrimSpace(item)
+		if item != "" {
+			jww.ERROR.Println(item)
+		}
+	}
+	if err != nil {
 		jww.ERROR.Println(err)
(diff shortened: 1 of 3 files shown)


helpers/content.go
```diff
@@ -546,9 +546,19 @@ func getAsciidocContent(content []byte) []byte {
...
 	cmd.Stdout = &out
-	if err := cmd.Run(); err != nil {
+	cmd.Stderr = &cmderr
+	err := cmd.Run()
+	// asciidoctor has exit code 0 even if there are errors in stderr
+	// -> log stderr output regardless of state of err
+	for _, item := range strings.Split(string(cmderr.Bytes()), "\n") {
+		item := strings.TrimSpace(item)
+		if item != "" {
+			jww.ERROR.Println(item)
```
bep: This is good -- but it would be useful if the page path was in the error print.

choeppler: Thanks -- and I agree. Haven't found a good way to pass the page path (see comment in #2399). Can the page path somehow be retrieved from the RenderingContext?

bep: Ah, OK -- I didn't see the entire function, you have no page here. Yes, we should add some more context to the `RenderingContext` that can used to give better feedback ... We have `DocumentID` (which is useless for logging) and we could add `DocumentName` where we can put the page's fileName.

Can you do that as part of this PR? (we can fix the rest of the logging later)

choeppler: Yes, I can do that. Putting `page.Path` in `DocumentName` we can get error messages like
```
ERROR: 2016/10/13 09:47:12 content.go:560: asciidoctor: ERROR: post/fail.ad: line 139: mismatched macro: endif::backen-docbook[], expected endif::backend-docbook[]
```
I'll add a second commit to this pull request with the changes for you to review. Ok?

bep: Yes, that looks perfect, thanks.

choeppler: Good. Done. See 441a2c90888b4f8e6463780fbd3e0caaa66c1f8b

PR #2571, opened 2016-10-13
all: Unify case of config variable names

bogem:
All config variables start with low-case and use camelCase.

If there is abbreviation at the beginning of the name, the whole abbreviation will
be written in low-case
If there is abbreviation at the end of the name, the whole abbreviation will
be written in upper-case.
For example, rssURI.

Also, the performance will be better, because `viper` lowers case of keys with function `strings.ToLower`, which calls `strings.Map(unicode.ToLower, str)` and this function works so, what the later there will be upper-case rune, the faster it will work.

Conversation: https://discuss.gohugo.io/t/viper-why-hugo-sets-keys-in-upper-case


bogem:
Any questions or offers?

bep:
>Any questions or offers?

No. I'm just waiting to merge a potential conflicting PR, that's all.

bogem:
Ok

bep:
And I was right. a10b2cd372798c4e4b862f0ec03010d2aea2ff1e created some conflicts in this PR. Sorry for dumping this in your lap, so to speak, but I just had to get that commit just right without too much distractions.

bogem:
No problem. I will solve it now

bogem:
Solved

bep:
Thanks for this, it looks so much better now.

@digitalcraftsman the docs got a cleanup of this, too. I'm not aware of any functional changes since Hugo 0.17 -- maybe we should just rebase the docs branch against master?

digitalcraftsman:
> @digitalcraftsman the docs got a cleanup of this, too. I'm not aware of any functional changes since Hugo 0.17 -- maybe we should just rebase the docs branch against master?

This would be a good idea. I would rather preferring to cherrypick this single commit in order to minimize merge conflicts.

PR #2576, opened 2016-10-14
Implement support for alias templates

natefinch:
> 
> This change adds a canonical alias.html template that is used for page
> redirects, and passes the page as data to the template under .Page

### DIFF
--- a/docs/content/extras/aliases.md
+++ b/docs/content/extras/aliases.md
@@ -94,3 +94,11 @@ Assuming a baseurl of `mysite.tld`, the contents of the html file will look some
 ```
 
 The `http-equiv="refresh"` line is what performs the redirect, in 0 seconds in this case.
+
+## Customizing
+
+You may customize this alias page by creating an alias.html template in the
+layouts folder of your site.  In this case, the data passed to the template is
+
+* Permalink - the link to the page being aliased
+* Page - the Page data for the page being aliased
--- a/docs/content/templates/overview.md
+++ b/docs/content/templates/overview.md
@@ -71,4 +71,6 @@ Used to render the XML sitemap
 ### [404](/templates/404/)
 This template will create a 404.html page used when hosting on GitHub Pages
 
+### [Alias](/extras/aliases/#customizing)
+This template will override the default page used to create aliases of pages.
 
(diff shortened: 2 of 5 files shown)


natefinch:
I tested this on my site, so for example, here's my alias.html in the root of my layouts directory: https://github.com/natefinch/npf/blob/master/layouts/alias.html

bep:
The implementation looks fine, but we need:

1) A line in the docs in /templates somewhere
2) One integration test

natefinch:
Great, yeah, I was going to put those up if the implementation looked ok.  Didn't want to do all the work on the docs if the implementation was going to change :)  Can you give me a pointer to where to put the integration tests... having tests in Hugo is new since I last worked on it ;)

bep:
It is a small test so `page_test.go` would be suitable -- or you can just create `alias_test.go` in hugolib (we may add more tests in the future). I did *some* work on the integration tests for Hugo 0.17 -- but not everything is as clean as I would like. 

I just rewrote a similar test to use Afero:

https://github.com/spf13/hugo/commit/3ce904669a0d24debc5849c9f1a2f3f5f8fc9f11

When we have a virtual file system for (almost) everything, we might as well use it in tests.

natefinch:
Tests are in, docs are in.  Please check out the docs, not sure if they are sufficient or not.

natefinch:
Not sure what's up with the test failures in CI.  I get one test failure locally:
```
--- FAIL: TestShortcodesInSite (0.13s)
	shortcode_test.go:616: sect/doc6/index.html content expected:
		"<div class=\"highlight\"><pre><code class=\"language-bash\" data-lang=\"bash\"><span></span>b: b c: c\n</code></pre></div>\n"
		got:
		"<div class=\"highlight\"><pre><code class=\"language-bash\" data-lang=\"bash\">b: b c: c\n</code></pre></div>\n"
FAIL
FAIL	github.com/spf13/hugo/hugolib	1.626s
```

But I don;t really understand why that is failing.

bep:
>But I don;t really understand why that is failing.

It is probably a `pygments` test that is too specific (in that it obviously fails in your pygments version) -- I will fix that, so just ignore that one.

Travis is in a global meltdown so ignore that, too for now. The code/tests/doc looks good.

PR #2580, opened 2016-10-14
Use govendor sync instead of get

moorereason:


moorereason:
@bep, I'm pretty sure this will fix our Travis builds, but Travis appears to be having a rough day.

PR #2594, opened 2016-10-17
Replace toc_token with table of contents if present

jlfwong:
In the process of migrating my blog from Jekyll to Hugo, I couldn't find an easy way to place a table of contents inline in the middle of content.

Jekyll allows you to specify a magic token (I use `{:toc}`) in the middle of your content that will be replaced with the table of contents. The associated code in Jekyll is here: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/converters/markdown/rdiscount_parser.rb#L18.

I wanted the same thing in Hugo, so I wrote it.

### DIFF
--- a/hugolib/hugo_sites.go
+++ b/hugolib/hugo_sites.go
@@ -14,6 +14,7 @@
 package hugolib
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


bep:
Sorry for this late response. agree that this would be useful; but this, if not already possible, should be done via a shortcode.

Issue #1526, opened 2015-10-27
`livereload.js` not being injected properly


pandorahlz:
Per some forum posts and another issue, it seems that it gets injected before the closing body tag, but I think this isn't right. HTML5 supports omitting (implied) body and head tags, so I think a line with the script tag should just be appended to the file on the fly upon client service. That said, there's still some sort of bug, because even with opening `html`, `head`, and `body` tags, the script isn't being injected.

bep:
You need a closing body tag. We are not changing that. We need something to search for.

pandorahlz:
What? Why? HTML5 is abundantly clear that there's no need for head and body tags. It's code clutter in your markup to include something unnecissarily. Why wouldn't yo just append it?

pandorahlz:
Also, there's a bug at least on OSX Yosemite, which is another issue in this bug that you haven't addressed. I think the suggestion to lose the requirement is an enhancement idea, but there's additionally an actual bug here, where even when meeting the ostensible requirement, it doesn't work as it should.

bep:
The HTML5 spec doesn't say that you can not have a end body tag. The Hugo spec says that you need an end body tag to use live reload. You can easily follow both HTML5 and Hugo spec by adjusting your template.

pandorahlz:
Don't you think the burden of justification should be on the software package imposing extra complexity?

pandorahlz:
And, according to HTML5, you really don't need to have beginning, or end, html, head, or body tags.

bep:
Please take discussions and questions at the forum.

pandorahlz:
Why so salty?

bep:
Because my time doing free work for the open source community is a limited resource. So, bring your questions to the forum where there are more people to answer.

PR #2599, opened 2016-10-17
Make livereload depend less on presence of body tag

vyuh:
Add livereload script before the first matching `endTag` among the following:
- `</body>`
- `</BODY>`
- `</html>`
- `</HTML>`
  If all above attempts fail to inject livereload, then append livereload to end of document if it starts with `<!DOCTYPE html>`.

This makes livereload work for HTML pages that omit optional tags including `body`
- Earlier, livereload only worked on webpages generated from layouts that have a body closing tag.
- There was no way to achieve this new behaviour using present command line flags.
- This behaviour is desired by all people who wish to omit optional HTML tags, including `body`, in layouts. See, https://google.github.io/styleguide/htmlcssguide.xml#Optional_Tags
- It is harmless to inject the livereload script at the end of content as a last resort when all attempts to replace patterns (like `</body>`) fail.
- Anyone who wants to avoid this new behaviour (why?) can use `--disableLiveReload` with `hugo server`.

See #1526

### DIFF
--- a/transform/livereloadinject.go
+++ b/transform/livereloadinject.go
@@ -21,18 +21,27 @@ import (
 )
 
 func LiveReloadInject(ct contentTransformer) {
-	endBodyTag := "</body>"
-	match := []byte(endBodyTag)
+	endTags := []string{"</body>", "</BODY>", "</html>", "</HTML>"}
 	port := viper.Get("port")
 	replaceTemplate := `<script data-no-instant>document.write('<script src="/livereload.js?port=%d&mindelay=10"></' + 'script>')</script>%s`
-	replace := []byte(fmt.Sprintf(replaceTemplate, port, endBodyTag))
+	var newcontent []byte
 
-	newcontent := bytes.Replace(ct.Content(), match, replace, 1)
+	for _, endTag := range endTags {
+		replacement := []byte(fmt.Sprintf(replaceTemplate, port, endTag))
+		match := []byte(endTag)
+		newcontent = bytes.Replace(ct.Content(), match, replacement, 1)
+		if len(newcontent) != len(ct.Content()) {
+			break
+		}
+	}
 	if len(newcontent) == len(ct.Content()) {
-		endBodyTag = "</BODY>"
-		replace := []byte(fmt.Sprintf(replaceTemplate, port, endBodyTag))
-		match := []byte(endBodyTag)
-		newcontent = bytes.Replace(ct.Content(), match, replace, 1)
+		doctype := "<!DOCTYPE html>"
+		if 0 == bytes.Index(newcontent, []byte(doctype)) {
+			// this is an HTML document with body and html tag omited
+			replace := fmt.Sprintf(replaceTemplate, port, "")
+			newcontent = append(ct.Content(), replace...)
+			// append livereload to end of document
+		}
 	}
 
 	ct.Write(newcontent)


bep:
>It is harmless to inject the livereload script at the end of content as a last resort when all attempts to replace patterns 

This assumes, of course, that all content is HTML.

vyuh:
@bep if some other kind of content is generated, and its proper function depends on non injection of `livereload.js`  based on  absence of body tag, then I agree that I have not considered that case.

bep:
The non HTML content is rare currently, but I have a site that produces a JSON file that would be messed up by this PR.

Footnote remark: Your replacement logic is also slightly flawed: It seems to add an end body tag and it is more effective to use append instead of bytes.Join.

vyuh:
Thank you for pointing out flaws. I have improved the code.
Now, it:
    - Does not end body tag when it was not there in first place
    - uses `append` instead of `bytes.Join`

This is the first time I am writing something in Go. I did
look up for something like `bytes.Append` but there wasn't
one, so I had settled for first thing I found, that did the job.

vyuh:
I searched the master branch for usages (direct and indirect) of 
`LiveReloadInject`.

This is the rough tree:
- `LiveReloadInject` in `transform/livereloadinject.go`
    - `renderAndWritePage` in `hugolib/site.go`
        - `pageRenderer` in `hugolib/site.go`
        - `taxonomyRenderer` in `hugolib/site.go`
        - `renderListOfTaxonomyTerms` in `hugolib/site.go`
        - `renderSectionLists` in `hugolib/site.go`
        - `renderHomePage` in `hugolib/site.go`

For this pull request to work, it is necessary that `hugo` calls these
functions only to generate content that is HTML.

To support other content types, it is sensible to provide a 
page specific `nolivereload` variable.

bep:
We have had this discussion before, and my argument is this:

If you want livereload and do not have a body in your HTML template: Add a body tag to your HTML template.

vyuh:
For users who omit optional tags, adding a body close tag is a workaround,
not a solution. They have no proper way to get `livereload` feature. Having
to change layouts to something inconsistent with their style of authoring
HTML is disheartening.  More so, because this is easily avoidable.

The requirement that injector function must allow non HTML content to pass
through cleanly, is an unclear design restriction considering that:

- The injector function is called only in situations where an HTML page is
  being generated.
- Injection of `livereload` script makes sense only for HTML content.
  No other kind of content requires to pass through this.
- Support for correct generation of other formats, can be added by creating
  clear and distinct interfaces. For example, by adding a `renderAndWriteJSON`
  function for handling JSON. See `renderAndWriteXML` which does not pass
  content through `livereload` injector.

vyuh:
For now, I have added a check to ensure that `livereload` is injected only to
documents beginning with "<!DOCTYPE html>". I would have started with a more
inclusive check, but using `regexp` might slow down the service of html pages
too much. This is the most unintrusive way that I could think of, to resolve
this.

Still, if this fallback behaviour conflicts with other goals of `hugo`
project, please consider adding a command line flag like `--forceLiveReload`
for this use case. Usability of `livereload` feature should not depend on
how one authors HTML in layout files.

bep:
I'm closing this. This works fine as is, and making it more complex will introduce other problems, for non HTML files etc. Thanks for your contribution.

Issue #2557, opened 2016-10-11
Add validation for invalid frontmatter markers


tarsisazevedo:
I have a post with `draft = true` and it's on my rss feed and on my index page without title.
My conf:

++++++
draft = true
title = "how to have tsuru + services"
date = "2016-10-12T23:54:04-02:00"
++++++

$ hugo version
Hugo Static Site Generator v0.17 BuildDate: 2016-10-07T11:42:05-03:00

To build my blog I use:
`hugo --destination=/path chdir=/tmp/site`

The last line which no title is the post in draft.
![screen shot 2016-10-11 at 10 13 19 am](https://cloud.githubusercontent.com/assets/118348/19271547/85ef973a-8f9b-11e6-8a38-1181f19094fd.png)

I can send a PR if you guide me to the bug :)
Thanks for this amazing software.

gmonnerat:
+1

I Have the same issue.

$ hugo version
Hugo Static Site Generator v0.17-DEV BuildDate: 2016-09-12T12:44:52-03:00

bep:
The draft feature works fine in Hugo 0.17 -- which and the missing title suggests an issue with the TOML parser (we changed parsers in Hugo 0.17), which is strange. I will investigate a little.

bep:
Do you see some ERROR in the console?

gmonnerat:
<pre>
(jessie)gabriel@localhost:~/dev/gmonnerat.github.com$ cat content/post/example.md | head -n 5
++++++
title = "how to have tsuru + services"
draft = "True"
++++++

(jessie)gabriel@localhost:~/dev/gmonnerat.github.com$ hugo
Started building sites ...
Built site for language en:
0 draft content
0 future content
0 expired content
19 pages created
0 non-page files copied
2 paginator pages created
7 tags created
13 categories created
total in 259 ms
</pre>

No error in the console and "0 draft content" instead of "1 draft content"

tarsisazevedo:
My hugo's posts is [here](https://github.com/tarsisazevedo/site). I don't get any error on console too.

bep:
OK, I see now what is happening. I added a test for this, see https://github.com/spf13/hugo/commit/d2bc64bee3190d5f1850bec45bf7f51375f13c79

* Your frontmatter is invalid ("++++++" vs "+++")
* In an ideal world be should handle this better, either matching it more losely or throw a noisy error
* RIght now we treat this as not having front matter at all, the simplest solution

[...]

PR #2600, opened 2016-10-18
[WIP] Log when a page is rendered without header

tarsisazevedo:
This PR is related to #2557 

What do you think? What I'm missing?


bep:
This should be handled when the parsing fails. Having no front matter is not an error in itself, and certainly not FATAL.

Issue #2433, opened 2016-09-11
Table of Contents (ToC) not always rendered


targodan:
I wanted to add a ToC to some of my posts and found that it is not being rendered under certain circumstances.

## My setup

Using the blackburn theme, added a shortcode called `/layouts/shortcodes/toc.html`:

``` html
<nav id="toc">
    {{ .Page.TableOfContents }}
</nav>
```

Added a post called `/content/post/test.md`:
``` markdown
+++
title = "Test"
+++

# ToC

{{% toc %}}

# a

a

# b

b

# c

c
```

Hugo version: `Hugo Static Site Generator v0.16 BuildDate: 2016-09-01T17:43:30+02:00`

## The rendered result

When I just execute hugo like this

``` bash
$ hugo -d public/
```

The ToC is not rendered / is empty.
The same thing happens when I execute hugo as a server.

``` bash
$ hugo server
```

When executing as server and I modify and save the `/layouts/shortcodes/toc.html` it renders the ToC.
When I now change and save the `/content/post/test.md` again, the ToC disappears again.

Under any circumstances I did not get any errors.

By "does not render / is empty" I mean the result looks like this in the markup.

``` html
<nav id="toc">
</nav>
```

stefano-pogliani:
I have a similar problem when trying to render the summary of other pages.

I use a taxonomy called `pagesets` to group pages and then I have a shortcode to render the list of pages in the taxonomy:
```html
{{ $pageset := .Get 0 -}}
{{ $pages := .Site.Taxonomies.pagesets.Get $pageset }}

{{ range $pages }}
<div class="pageset-block">
  <h2>
    <a href="{{ .Page.RelPermalink }}">{{ .Page.LinkTitle }}</a>
  </h2>
  <p>{{ .Page.Summary }}</p>
</div>
{{ end }}
```

I am not familiar with the Go programming language but I was able to follow the build process to the following lines:

  * Where (I believe) shortcodes are processed: https://github.com/spf13/hugo/blob/v0.17/hugolib/hugo_sites.go#L469
  * Where (I believe) TOC, Content, and Summary are computed: https://github.com/spf13/hugo/blob/v0.17/hugolib/hugo_sites.go#L490

My use case is a sort of "index page" for the pages in the `pageset`.
As such I suspect that, for me, the shortcode in the "index page" is processed before the pages referenced in it are.

@targodan where you able to find a solution/work around for your problem?

targodan:
[...]

PR #2623, opened 2016-10-23
Handle ToC before handling shortcodes.

targodan:
The generation of the ToC is now done before the shortcodes are
handled. This makes the use of ToC in shortcodes possible.

The test checks whether or not the ToC is successfully rendered when
used from within a shortcode.

This relates to issue #2433.


hugolib/hugo_sites_test.go
```diff
@@ -659,6 +670,106 @@ func TestChangeDefaultLanguage(t *testing.T) {
...
+	cfg := BuildCfg{}
+
+	err := sites.Build(cfg)
+
+	if err != nil {
+		t.Fatalf("Failed to build sites: %s", err)
+	}
+
+	// Not checking the heading-ids because they are not generated by the ToC.
+	assertFileContentRegexp(t, "public/en/post/simple/index.html", true, tocPageSimpleExpected)
```
bep: You can turn off those funky heading IDs by setting plainIDAnchors = true or something -- it is set in another test, so search for it.

targodan: Will get to it in a minute. Should I leave the function `assertFileContentRegexp` in for future use or should I remove it so as to not have dead code in it?

bep: Just leave it be, it will be useful at some point.

targodan: Ok, I just tried adding `plainIDAnchors = true` to the `[blackfriday]` section of the config but it didn't change anything. According to the [documentation](https://gohugo.io/overview/configuration/) it is true by default and it isn't what we are looking for.

My idea was that my test should not break when/if the id-generation is altered. I have found a different sollution for that. My test is now using the `#heading {#id}` method of secifying ids.

bep:
Thanks for this!

ketan:
Interestingly, I'm looking to generate TOC after shortcodes. My requirement is that I have some shortcodes that include partials. Some of these partials contain headers, which are not known being rendered by the toc.

Any thoughts on working around this issue?

labkode:
Same problem as @ketan, I use a shortcode that generates configuration directives and they are not rendered into the TOC because they live in a shortcode.

plinss:
Also similar problem as @ketan, I'm trying to use shortcodes in the headings to create icons, and the TOC now contains `HAHAHUGOSHORTCODE`.

PR #2628, opened 2016-10-24
all: Delete some redundant code

bogem:
Code looks a little bit cleaner now

### DIFF
--- a/commands/genman.go
+++ b/commands/genman.go
@@ -14,7 +14,6 @@
 package commands
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -36,7 +35,7 @@ in the "man" directory under the current directory.`,
 		header := &doc.GenManHeader{
 			Section: "1",
 			Manual:  "Hugo Manual",
-			Source:  fmt.Sprintf("Hugo %s", helpers.HugoVersion()),
+			Source:  "Hugo " + helpers.HugoVersion(),
 		}
 		if !strings.HasSuffix(genmandir, helpers.FilePathSeparator) {
 			genmandir += helpers.FilePathSeparator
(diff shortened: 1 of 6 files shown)


bogem:
Any questions?

bep:
How is this an improvement?

bogem:
Haha. This is really not an improvement (except of simplifying generating of hugo version ([code](https://github.com/spf13/hugo/pull/2628/files#diff-3793dfaf7c918125b94a1f1d0baf59d4L39))). But the code is a liiiiiiittle bit cleaner now 😅

bep:
>But the code is a liiiiiiittle bit cleaner now

How is that? I assume the return values from most of these are of type `error` -- and while I agree that not handling these error is bad and should fixed, just removing the "_" is more like sweeping the bugs under the carpet.

bogem:
Ok. Then I will edit it and will check error instead of just ignoring it

commands/genman.go
```diff
@@ -36,7 +35,7 @@ in the "man" directory under the current directory.`,
 		header := &doc.GenManHeader{
 			Section: "1",
 			Manual:  "Hugo Manual",
-			Source:  fmt.Sprintf("Hugo %s", helpers.HugoVersion()),
```
bep: I think the fmt version is easier to understand.

bogem: But I think the concatenation of two simple strings is also not so hard to understand.

bep: So both are easy, why change it?

bogem: Because it's a bit faster😅

bogem:
Ok, this PR was not good idea. I promise, this PR was the last mediocre contribution

bep:
@bogem the potential error fixes is a good idea, but I don't like cosmetic changes just for changes sake, typos are easy to make, and this is how bugs happen.

PR #2646, opened 2016-10-29
Schema org

brunoamaral:
resolves spf13/hugo#2635 , Only show post's own keywords

### DIFF
--- a/tpl/template_embedded.go
+++ b/tpl/template_embedded.go
@@ -233,7 +233,7 @@ func (t *GoHTMLTemplate) EmbedTemplates() {
 {{ end }}{{ end }}
 
 <!-- Output all taxonomies as schema.org keywords -->
-<meta itemprop="keywords" content="{{if .IsPage}}{{ range $index, $tag := .Params.tags }} {{ $tag }},{{ end }}{{ else }}{{ range $plural, $terms := .Site.Taxonomies }}{{ range $term, $val := $terms }}{{ printf "%s," $term }}{{ end }}{{ end }}" />
+<meta itemprop="keywords" content="{{ if .IsPage}}{{ range $index, $tag := .Params.tags }}{{ $tag }},{{ end }}{{ else }}{{ range $plural, $terms := .Site.Taxonomies }}{{ range $term, $val := $terms }}{{ printf "%s," $term }}{{ end }}{{ end }}{{ end }}" />
 {{ end }}`)
 
 	t.AddInternalTemplate("", "google_analytics.html", `{{ with .Site.GoogleAnalytics }}
--- a/tpl/template_embedded.go
+++ b/tpl/template_embedded.go
@@ -233,7 +233,7 @@ func (t *GoHTMLTemplate) EmbedTemplates() {
 {{ end }}{{ end }}
 
 <!-- Output all taxonomies as schema.org keywords -->
-<meta itemprop="keywords" content="{{ range $plural, $terms := .Site.Taxonomies }}{{ range $term, $val := $terms }}{{ printf "%s," $term }}{{ end }}{{ end }}" />
+<meta itemprop="keywords" content="{{if .IsPage}}{{ range $index, $tag := .Params.tags }} {{ $tag }},{{ end }}{{ else }}{{ range $plural, $terms := .Site.Taxonomies }}{{ range $term, $val := $terms }}{{ printf "%s," $term }}{{ end }}{{ end }}" />
 {{ end }}`)
 
 	t.AddInternalTemplate("", "google_analytics.html", `{{ with .Site.GoogleAnalytics }}


bep:
@digitalcraftsman  Approving non-mergable PRs doesn't really bring us very much closer to resolving this PR.

digitalcraftsman:
> Approving non-mergable PRs doesn't really bring us very much closer to resolving this PR.

Except that this PR needs to be rebased I approve the changes made to the internal template.

bep:
>What aspect makes this PR non-mergable?

It needs a rebase. I'm not saying you should do it, but someone could maybe ask the PR starter nicely? And that could maybe have been done as part of the "reviewing workflow". Not sure.

moorereason:
@brunoamaral, can you rebase this PR, please?

digitalcraftsman:
> It needs a rebase. I'm not saying you should do it, but someone could maybe ask the PR starter nicely? And that could maybe have been done as part of the "reviewing workflow". Not sure.

Because you requested a review from me I thought you just wished a second opinion for it. On the hand it's would be good to have this PR already rebased to merge with a few clicks would be a good idea. On the other hand smaller changes that aren't affected by a large merge conflict could be merged with GitHubs automatic rebase feature. New commits that don't touch the part of the contribution at all shouldn't be a blocker for a PR.

On

tpl/template_embedded.go
```diff
@@ -233,7 +233,7 @@ func (t *GoHTMLTemplate) EmbedTemplates() {
 {{ end }}{{ end }}
 
 <!-- Output all taxonomies as schema.org keywords -->
-<meta itemprop="keywords" content="{{ range $plural, $terms := .Site.Taxonomies }}{{ range $term, $val := $terms }}{{ printf "%s," $term }}{{ end }}{{ end }}" />
+<meta itemprop="keywords" content="{{ if .IsPage}}{{ range $index, $tag := .Params.tags }}{{ $tag }},{{ end }}{{ else }}{{ range $plural, $terms := .Site.Taxonomies }}{{ range $term, $val := $terms }}{{ printf "%s," $term }}{{ end }}{{ end }}{{ end }}" />
```
jhabdas: I have two concerns to share:

First, we could be using `delimit` to clean it up:

```html
<meta name="keywords" content="{{ if .IsPage }}{{ delimit .Params.tags ", " }}{{ end }}">
```

Second, this assumes the schema template is only being used once in a document. However, on list pages it's quite likely schema will need to be output multiple times. As a result, we should not duplicate the `.Site.Taxonomies` information using an accumulator as the output in a list would be duplicative.

brunoamaral: Guys, terribly sorry about this. I missed the notification where moorereason asked me to rebase the PR.

@jhabdas Agree with your first point and have a question about the second one. I only use the schema.org partial on the head of the document, and  therefore it should be used only once. What use case am I missing?

Also this should only be applied in regards to pages or single posts, not in lists of pages.

jhabdas: > What use case am I missing?

The case where an schema object is repeated in a list and the partial is applied multiple times.

screen shot 2017-04-09 at 8 10 46 pm

You can dig into it [via the validator](https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fhackcabin.com) if you like.

The following schema _Things_ support the attributes (`dateModified`, `datePublished`, `keywords`) provided in this internal template:

- https://schema.org/CreativeWork
- http://schema.org/BlogPosting (inherits from CreativeWork)

If we output a list of site keywords we lose the ability reuse this partial more than once per page (which is a major shortcoming of Jekyll SEO Tag IMO.

Does that make sense?

jhabdas: Spotted another issue. This code is outputting hyphenated keyphrases, and the output should be humanized (preferably in lower case), e.g.

```
{{ lower (humanize $term) }}
```

jhabdas: Not to poop and swoop. My suggested resolution is to do this:

```
<meta itemprop="keywords" content="{{ .Keywords }}">
```

If if you're trying to do this at the page level _only_, you would do something like this:

```html
<meta name="keywords" content="{{ if .Keywords }}{{ delimit .Keywords ", " }}{{ else if .IsPage }}{{ delimit $.Params.tags ", " }}{{ else }}{{ range $term, $val := .Site.Taxonomies.categories }}{{ lower (humanize $term) }}, {{ end }}{{ end }}">
```

The only deficiency in the above page-level code AFAICT is there is an extra comma at the end of the list of commas. Annoying for humans, parsers won't care.

brunoamaral: Thank you for this @jhabdas. I'll try to apply your suggestions once I get a chance and maybe submit a fresh PR for hugo 0.20. :)

jhabdas: Thanks so much @brunoamaral! I'd love to be able to use the internal snippet on my pages. As it stands the only things preventing me from doing so are the keywords issue described as well as the inclusion of `wordCount` (invalid for `CreativeWork` but valid for `BlogPosting`).

jhabdas: @brunoamaral here's what I got so far regarding a page-level meta, in case it helps some WRT slicing and dicing for Schema: https://github.com/comfusion/after-dark/commit/4b1f0f1ae2785fc636d84380a8f803d0e689f5e7

Issue #1251, opened 2015-07-05
Support for minification of generated HTML files


chaseadamsio:
When generating `HTML`, it'd be really nice to have the code minified at least as a single line of `HTML` (instead of just output with lots of spacing and line breaks). Would it be hard to implement this?

Dr-Terrible:
For these kind of tasks there is an handy go library that could be easily integrated in the Hugo's pipeline: https://github.com/tdewolff/minify

bep:
This has been discussed, and this may be relevant to the plugin discussion. @Dr-Terrible I haven't looked at minify -- but if it is small and well written, one could argue that HTML minify should be a core feature off Hugo (and not a plugin).

But with plugins we could pull in all kinds of stuff, even stuff not written in Go.

chaseadamsio:
For the time being, I'm using a gulp task (since I'm already using it for deploying to s3) to minify and update asset hashes. 

Personally, I'm down with the plugin architecture...I'd be interested to hear more about the discussion. I'll look around and lurk. Thanks!

bep:
I use gulp for similar stuff myself. I would love to ditch that. It should be possible, the hardest part getting a decent LESS implementation in Go, but maybe that's not necessary.

Search for `plugin` here and at http://discuss.gohugo.io/

Dr-Terrible:
> @Dr-Terrible I haven't looked at minify -- but if it is small and well written, one could argue that HTML minify should be a core feature off Hugo (and not a plugin).

@bep That is why I have suggested _minify_ in the first place. It's relatively small, well written, and extremely fast too (<10ms per page). I'm completely against the idea of plugins (I saw the discussion on http://discuss.gohugo.io/ some time ago); for that kind of features there is npm, gulp, grunt, and tons of tools alike. Unfortunately, there aren't decent _plugins_ for HTML tidying/minification, and I agree with you that one could argue that it should be a core function of Hugo (but only for HTML page, no CSS/JS/SVG/XML etc). I like that idea, since HTML pages generated by Hugo have a lot of dead weight on due to the Go template system injecting white spaces and new lines like there is not tomorrow. I'd like Hugo to be able to trim that _fat_ too (with Jekyll there is a specific plugin to remove the extra noise from Shopify) .

[...]

PR #2664, opened 2016-11-03
transform: Adds --optimize and MinifyHTML

d4l3k:
When --optimize is set, MinifyHTML uses github.com/tdewolff/minify to minify all the
 pages being rendered. It minifies the HTML as well as the embedded CSS, JS and JSON.

In the case of CSS, there is a check if "-amp-start" is present and if so, it won't
minimize. This is to avoid breaking amp-boilerplate.

https://github.com/ampproject/amphtml/blob/master/spec/amp-boilerplate.md

See #1251

I'm not entirely sure if the AMP check is a good fit for this. Would probably be better to structure it in some other way, but it was the quickest way to get this working for my use case.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -225,6 +225,7 @@ func initHugoBuildCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool("canonifyURLs", false, "if true, all relative URLs will be canonicalized using baseURL")
 	cmd.Flags().StringVarP(&baseURL, "baseURL", "b", "", "hostname (and path) to the root, e.g. http://spf13.com/")
 	cmd.Flags().Bool("enableGitInfo", false, "Add Git revision, date and author info to the pages")
+	cmd.Flags().Bool("optimize", false, "Minify HTML")
 
 	cmd.Flags().BoolVar(&nitro.AnalysisOn, "stepAnalysis", false, "display memory and timing of different steps of the program")
 	cmd.Flags().Bool("pluralizeListTitles", true, "Pluralize titles in lists using inflect")
@@ -395,6 +396,7 @@ func initializeFlags(cmd *cobra.Command) {
 		"forceSyncStatic",
 		"noTimes",
 		"noChmod",
+		"optimize",
 	}
 
 	for _, key := range persFlagKeys {
--- a/transform/minifyhtml.go
+++ b/transform/minifyhtml.go
@@ -1,0 +1,48 @@
+// Copyright 2016 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package transform
+
+import (
+	"bytes"
+	"regexp"
+
+	"github.com/pkg/errors"
+	"github.com/tdewolff/minify"
+	"github.com/tdewolff/minify/css"
+	"github.com/tdewolff/minify/html"
+	"github.com/tdewolff/minify/js"
+	"github.com/tdewolff/minify/json"
+)
+
+var minifier *minify.M
+
+func init() {
+	minifier = minify.New()
+	minifier.AddFunc("text/css", css.Minify)
+	minifier.Add("text/html", &html.Minifier{
+		KeepDefaultAttrVals: true,
+		KeepWhitespace:      true,
+		KeepDocumentTags:    true,
+	})
+	minifier.AddFunc("text/javascript", js.Minify)
+	minifier.AddFuncRegexp(regexp.MustCompile("[/+]json$"), json.Minify)
+}
+
+// MinifyHTML minifies the HTML and the embedded CSS, JS and JSON.
+func MinifyHTML(ct contentTransformer) {
+	buf := bytes.NewReader(ct.Content())
+	if err := minifier.Minify("text/html", ct, buf); err != nil {
+		panic(errors.Wrap(err, "error minifying html"))
+	}
+}
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -317,6 +317,78 @@
 			"revisionTime": "2016-12-17T20:04:45Z"
 		},
 		{
+			"checksumSHA1": "ItM9x14tp2vRPEwTufY42XkJFe8=",
+			"path": "github.com/tdewolff/buffer",
+			"revision": "0edfcb7b750146ff879e95831de2ef53605a5cb5",
+			"revisionTime": "2016-04-27T18:11:22Z"
+		},
+		{
+			"checksumSHA1": "aYOP+Gtn+AK81aMpgqGKkrt+rMI=",
+			"path": "github.com/tdewolff/minify",
+			"revision": "5788bd40c07d1e1d4fc4eb8a955760d509c3159c",
+			"revisionTime": "2016-11-05T17:17:08Z"
+		},
+		{
+			"checksumSHA1": "uRo+3Qy3vwfbwBsjQNEprWBHKHo=",
+			"path": "github.com/tdewolff/minify/css",
+			"revision": "5788bd40c07d1e1d4fc4eb8a955760d509c3159c",
+			"revisionTime": "2016-11-05T17:17:08Z"
+		},
+		{
+			"checksumSHA1": "Ytcf7dKpnLbz62zxwB/VdSCBS7w=",
+			"path": "github.com/tdewolff/minify/html",
+			"revision": "5788bd40c07d1e1d4fc4eb8a955760d509c3159c",
+			"revisionTime": "2016-11-05T17:17:08Z"
+		},
+		{
+			"checksumSHA1": "8Vj9VTxgIn/BaZuGuwi1u/ckWZ8=",
+			"path": "github.com/tdewolff/minify/js",
+			"revision": "5788bd40c07d1e1d4fc4eb8a955760d509c3159c",
+			"revisionTime": "2016-11-05T17:17:08Z"
+		},
+		{
+			"checksumSHA1": "w9vp7arHrzUEi6KyAM9K8hAP3hM=",
+			"path": "github.com/tdewolff/minify/json",
+			"revision": "5788bd40c07d1e1d4fc4eb8a955760d509c3159c",
+			"revisionTime": "2016-11-05T17:17:08Z"
+		},
+		{
+			"checksumSHA1": "dW3/qMmvhmqylTirRfPK0by6bVE=",
+			"path": "github.com/tdewolff/parse",
+			"revision": "5c35632ca368ebf46247ede2b96e9f6c0258fae0",
+			"revisionTime": "2016-10-09T09:18:41Z"
+		},
+		{
+			"checksumSHA1": "QnZ3/9oDs+BrgtD/gS7x1oeNI2U=",
+			"path": "github.com/tdewolff/parse/css",
+			"revision": "5c35632ca368ebf46247ede2b96e9f6c0258fae0",
+			"revisionTime": "2016-10-09T09:18:41Z"
+		},
+		{
+			"checksumSHA1": "+gDXaHsyhgy8EB9beJKKH932qAk=",
+			"path": "github.com/tdewolff/parse/html",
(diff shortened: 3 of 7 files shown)


vendor/vendor.json
```diff
@@ -21,7 +21,7 @@
 			"revisionTime": "2016-07-26T15:08:25Z"
 		},
 		{
-			"checksumSHA1": "dIidE/pUkqulIvTkuMBTpH3/lnU=",
+			"checksumSHA1": "A3v5qAlaOKeq3c+yDCL4jzAB0UA=",
```
bep: What are these changes? Not related to your commit -- but I'm curious as to why the checksums differs ...

d4l3k: I ran `make govendor` in the assumption that it would install govendor and download the dependencies.

I guess it also updated them?

d4l3k: Removed that file.

transform/minifyhtml.go
```diff
@@ -0,0 +1,62 @@
...
+	})
+	m.AddFunc("text/javascript", js.Minify)
+	m.AddFuncRegexp(regexp.MustCompile("[/+]json$"), json.Minify)
+	minifier = m
+}
+
+// MinifyHTML minifies the HTML and the embedded CSS, JS and JSON.
+func MinifyHTML(ct contentTransformer) {
+	var buf bytes.Buffer
+	buf.Write(ct.Content())
```
bep: I will look closer at this PR later, but 3 comments for now:

* If you need a `io.Reader` from a `byte` slice, use `bytes.NewReader`, it is 200 times faster and allocates no memory.
* Not all content produced by Hugo is HTML, a user can set the `url` in to `index.json` as an example.
* You need to pull in the dependencies into the vendor file for the CI build to go green. See the `govendor` doc.

moorereason:
Haven't looked at the patch in detail, but I would suggest naming the option "minify" instead of "optimize".

d4l3k:
@moorereason I used "optimize" since the discussion from the linked issue (#1251) suggested it. Happy to switch to minify though.

bep:
No, keep the `optimize`.

bep:
But what is the performance of this.

I notice this comment:

>@bep That is why I have suggested minify in the first place. It's relatively small, well written, and extremely fast too (<10ms per page). 

Seeing Hugo is currently doing <1ms per page, <10ms doesn't sound "extremely fast".

d4l3k:
Doesn't seem like the performance is much of a consideration. In the tests of my personal site, there's a very small difference in time.

```
rice@config  ~/Developer/fn.lc   master  rm -r public/; time hugo
Started building sites ...
Built site for language en:
0 draft content
0 future content
0 expired content
22 pages created
0 non-page files copied
0 paginator pages created
8 tags created
0 categories created
total in 77 ms
hugo  0.07s user 0.01s system 81% cpu 0.106 total

rice@config  ~/Developer/fn.lc   master  rm -r public/; time hugo --optimize
Started building sites ...
Built site for language en:
0 draft content
0 future content
0 expired content
22 pages created
0 non-page files copied
0 paginator pages created
8 tags created
0 categories created
total in 92 ms
hugo --optimize  0.07s user 0.03s system 85% cpu 0.117 total
```

While there's some fluctuation and my site is rather small, there's only about a 20% increase in time taken.

d4l3k:
@bep Any thoughts on that test failure? Not sure what would be causing something like that.

d4l3k:
@bep Any more feedback?

transform/minifyhtml.go
```diff
@@ -0,0 +1,48 @@
...
+		KeepWhitespace:      true,
+		KeepDocumentTags:    true,
+	})
+	m.AddFunc("text/javascript", js.Minify)
+	m.AddFuncRegexp(regexp.MustCompile("[/+]json$"), json.Minify)
+	minifier = m
+}
+
+// MinifyHTML minifies the HTML and the embedded CSS, JS and JSON.
+func MinifyHTML(ct contentTransformer) {
```
aomarks: Should this be called Minify() and the file minify.go, since it doesn't just minify HTML?

transform/minifyhtml.go
```diff
@@ -0,0 +1,48 @@
...
+	buf := bytes.NewReader(ct.Content())
+	if err := minifier.Minify("text/html", ct, buf); err != nil {
+		panic(err)
```
aomarks: Does this panic get recovered somewhere?

Maybe add "Error minifying" or something to the message in case minify's error is not disambiguating.

transform/minifyhtml.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/tdewolff/minify/css"
+	"github.com/tdewolff/minify/html"
+	"github.com/tdewolff/minify/js"
+	"github.com/tdewolff/minify/json"
+)
+
+var minifier *minify.M
+
+func init() {
+	m := minify.New()
```
aomarks: Nit: should be safe to assign directly to the minifier variable from this init without the m temporary. Or is it just to keep the lines short?

hugolib/hugo_sites_test.go
```diff
@@ -891,6 +893,7 @@ var multiSiteJSONConfig = `
   "defaultExtension": "html",
   "baseURL": "http://example.com/blog",
   "disableSitemap": false,
+	"optimize": false,
```
d4l3k: Whoops, fixed

transform/minifyhtml.go
d4l3k: It only accepts HTML. You can embed other things within the HTML to be minified, but it calls `minifier.Minify("text/html", ...)`.

transform/minifyhtml.go
d4l3k: It was to keep the lines short, but expanding it is probably clearer.

transform/minifyhtml.go
d4l3k: Added a bit of clarification text.

Ideally we would modify `transform.trans` type definition to have an error returned from it. Would make error handling a lot cleaner. Thoughts, @bep?

bep:
>@bep Any more feedback?

I haven't had any time to look at this.

metachris:
Just wanted to give this effort a thumbs and a bump. Would be nice to have build-in minification support for Hugo. In particular for all of HTML, JS and CSS.

bep:
@avanderhoorn please use the thumbs up or something. I suspect that these "Bump" comments work against their purpose.

avanderhoorn:
@bep Understood. Its just that the thumbs don't have any outward reflection in terms of notifications or interest from the issue list. But now that I know you are looking at those, I'm happy with that.

bep:
>But now that I know you are looking at those, I'm happy with that.

I don't, but it is a little "the boy who cried wolf". I get a notification on everything that happens in Hugo (and many other repos). So I look into it, expecting it to be important. A "me want this, too" does not fall into the "important" category. So, if enough people do that (i.e. "the boy who cried wolf"), I may have to stop looking into the notifications on existing issues. That's all.

tdewolff:
> Seeing Hugo is currently doing <1ms per page, <10ms doesn't sound "extremely fast".

Maybe a little late, but I just wanted to correct this statement. It does 3 ms for the BBC website, which is about 110 kB. Given that most Hugo pages are 1 kB or 10 kB I suspect, it would be well under 1 ms per page. I have put a major amount of time in optimizing it, and I don't think there is any way to optimize it much without removing the abstractions and generalities. Also, as soon as thunk-function inline works by the Go compiler, expect a ~25% speed increase as well.

Jos512:
> Given that most Hugo pages are 1 kB or 10 kB I suspect, it would be well under 1 ms per page.

Does it matter whether it adds < 1ms or < 10ms per page? I can imagine that people use the minification only when creating a deploy. For those situations the 10ms per page is more than acceptable. Preferable even, because my Gulp task takes me more than a second per page.

How about we disable minfication by default for `hugo server` and enable it for the `hugo` command? (Just like the new "Fast Render Mode" is now enabled by default for `hugo server`, but doesn't kick in with `hugo` simply because it isn't needed there.)

With the command flag people can easily turn off this behaviour when building with `hugo`.

> Also, as soon as thunk-function inline works with the Go compiler, expect a ~25% speed increase as well.

I would suspect, but this is for much further in the process, that Hugo can also improve the minification when using caching. Just like when we cache partials, some of the minified output might also be cacheable.

transform/minifyhtml.go
```diff
@@ -0,0 +1,48 @@
...
+
+var minifier *minify.M
+
+func init() {
+	minifier = minify.New()
+	minifier.AddFunc("text/css", css.Minify)
+	minifier.Add("text/html", &html.Minifier{
+		KeepDefaultAttrVals: true,
+		KeepWhitespace:      true,
+		KeepDocumentTags:    true,
```
tdewolff: I would suggest getting rid of `KeepDocumentTags`. It is only there to support malfunctioning third-party tools. `KeepWhitespace` and `KeepDefaultAttrVals` are only there to allow improper use of block tags in combination with whitespace or CSS/JS selectors on HTML default attribute values.

transform/minifyhtml.go
```diff
@@ -0,0 +1,48 @@
...
+	})
+	minifier.AddFunc("text/javascript", js.Minify)
+	minifier.AddFuncRegexp(regexp.MustCompile("[/+]json$"), json.Minify)
+}
+
+// MinifyHTML minifies the HTML and the embedded CSS, JS and JSON.
+func MinifyHTML(ct contentTransformer) {
+	buf := bytes.NewReader(ct.Content())
+	if err := minifier.Minify("text/html", ct, buf); err != nil {
+		panic(errors.Wrap(err, "error minifying html"))
```
tdewolff: Is it normal to panic out of a content transformer? Is there a friendlier way?

vendor/vendor.json
```diff
@@ -316,6 +316,78 @@
 			"revision": "2402e8e7a02fc811447d11f881aa9746cdc57983",
 			"revisionTime": "2016-12-17T20:04:45Z"
 		},
+		{
```
tdewolff: You can update the `minify` and `parse` packages, there is a big performance improvement with the latest release (see https://github.com/tdewolff/minify/commit/9904fe70e1c1cb5217ed45b24bf9728de3ff10c2). Also, the new `parse` package has incorporated `buffer` and `strconv`, so these dependencies are not required anymore.

commands/hugo.go
```diff
@@ -225,6 +225,7 @@ func initHugoBuildCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool("canonifyURLs", false, "if true, all relative URLs will be canonicalized using baseURL")
 	cmd.Flags().StringVarP(&baseURL, "baseURL", "b", "", "hostname (and path) to the root, e.g. http://spf13.com/")
 	cmd.Flags().Bool("enableGitInfo", false, "Add Git revision, date and author info to the pages")
+	cmd.Flags().Bool("optimize", false, "Minify HTML")
```
tdewolff: Maybe lose the "HTML" (here and in other places)?

transform/minifyhtml.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/tdewolff/minify/html"
+	"github.com/tdewolff/minify/js"
+	"github.com/tdewolff/minify/json"
+)
+
+var minifier *minify.M
+
+func init() {
+	minifier = minify.New()
+	minifier.AddFunc("text/css", css.Minify)
```
tdewolff: You don't see to add the XML and SVG minifiers, however it is increasingly common to embed SVG images either verbatim withing HTML or as a data URI (base64 usually) within CSS. Both will be minified if you add the SVG minifier. XML (besides SVG) is less common, but it is possible people add `MathML` in their sources.

jonathanstiansen:
This is odd, the `licence/cla`(sic) is seperate from the `license/cla` - I don't really know the protocol but one (probably the misspelled one) needs to be deleted.

In the meantime @d4l3k could you sign the `license/cla` I am ITCHING for a built in, consistent way of minification in my Hugo projects.

d4l3k:
@jonathanstiansen Just signed it. Not sure why I had to sign it twice.

@tdewolff I'd love to polish this up and get it merged, but it seems like @bep has zero interest so I don't really want to waste my time

cristim:
@d4l3k I'd love to have this merged, what's still blocking it?

d4l3k:
@cristim Complete and utter lack of interest from the Hugo maintainers.

spookyUnknownUser:
Why don't they just close this then?

bep:
Sorry about this, but we will not merge this at this point. We need to get some kind of assets processing design (and maybe some plugins support) before doing this and similar.

killua99:
What options do we have to minimizeHTML then? run a gulp process ? or an external thirdparty tool ?

jonathanstiansen:
Also would have been nice to hear this years ago, *sometime* around when this was put forward.

d4l3k:
@jonathanstiansen I was really excited to contribute to Hugo when I first started using it, but have really been put off by the maintainers apathy towards contributors. It's really a shame

PR #2669, opened 2016-11-04
docs: Link latest change as commit in the footer

digitalcraftsman:


digitalcraftsman:
I'll interpret your "thumbs up" as approval @bep. So I would merge this one.

bep:
You don't need my approval ... I just "thumbed up" because I thought it was a good idea.

Issue #2651, opened 2016-10-31
Improve reStructuredText performance 


tychoish:
While we wait on a mature (or even moderately capable) native go reStructuredText support (e.g. #1436), I would like to open a new issue as a proposal for an intermediate solution. 

I'm preparing a site with ~1600 rst posts. Using the current implementation, it takes 2 and a half to three minutes to render the site. On a theory that we were loosing some time to python process startup, I prototyped a solution where: 
- a very basic flask app that takes a POST request with some and returns the rendered restructured task: 
  
  ``` python
  import docutils.core
  
  from flask import Flask
  from flask import request
  
  app = Flask(__name__)
  
  @app.route("/", methods=["POST", "GET"])
  def main():
      if request.method == "GET":
          return "rst2html service; use the POST method to convert"
      else:
          return docutils.core.publish_parts(request.data, writer_name="html")['body'].strip()
  
  if __name__ == '__main__':
      app.run()
  ```
- a modification to the existing `getRstContent()` function that does the appropriate HTTP request for converting the content.

This gets the rendering time down for my site down to ~30 seconds. I tried converting all the rst to markdown (which didn't go well,) and rendering was ~3 seconds, so this isn't ideal, and doesn't obviate the need for #1436, but puts rst support into the realm of usable. 

The challenge is how to make this solution production worthy. My initial inclination is to write the script above to a tempfile, and then start it in a goroutine in the background, but I'm not sure if hugo has a concept, of a "background job" that we could hook into, and by the time we need to render rst the first time, its too late. At the same time, rst clearly has a niche audience within hugo, and it seems disruptive to force RST to restructure the way that hugo processes text (though, I think that's not historically inconsistent, based on my memory of adding the existing rst support.) 

Would it make sense to write a "go rst library" with a functional interface, that uses this approach internally that hugo could consume? My largest concern is potentially leaking the flask app, but I assume that's possible to overcome. 

Is there another option that I'm missing?

bep:
>The challenge is how to make this solution production worthy.

I would say that unless 1000s of people are shouting for speedier RST, your solution will never be accepted into Hugo. It adds complexity that I don't want to maintain/support.

You are right about the `os/exec` overhead -- if that could be reduced by batching the RST conversion somehow (it needs some refactoring, but Hugo could send all (or batches of) the pages in one Go to the external helper), that could be interesting. But that should also somehow include Aciidoc, which I guess is more popular than RST.

bep:
[...]

PR #2672, opened 2016-11-06
use shimmed approach to translating rst and asciidoc

tychoish:
This is, as a draft and to further the discussion from #2651. 

Mostly just a wrapper around https://github.com/tychoish/shimgo (which we can and should discuss further..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,4 @@ script:
   - ./hugo -s docs/
   - ./hugo --renderToMemory -s docs/
 before_install:
-  # gem install must be run with sudo on OSX 
-  - sudo gem install asciidoctor | gem install asciidoctor 
-  - sudo pip install docutils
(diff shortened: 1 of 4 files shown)


bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

bep:
@tychoish sorry, I lost track of this issue. I will make a not on returning back to this when I get some other stuff off my plate.

Could you rebase this against the current master?

bep:
OK, I will close this now. The idea is good, but I suggest starting a new PR.

Issue #2535, opened 2016-10-08
Widget mechanism


lebarde:
Hello,

I have used an existing theme for [my new website](http://www.topologix.fr). But I have to modify it to add (for instance) VCards or other beautiful stuff. If ever I want to update my theme or switch to another one, I will lose the code parts.

It would be great if there was a wordpress-like, reusable widget mechanism. I think we would have to:
- create a widget (as a piece of template? with more extensionability?);
- insert a widget area inside the theme;
- instantiate widgets inside `config.toml` or `config.yaml`.

For now I understand that there is no normalization for that process, which IMO prevents well-meaning people to publish finely tuned (and customizable) site blocks, just like in any content management system.

Do you think like me that it would be a good enhancement?

bep:
This has been discussed before, but noone has really taken it any further.

We need someone to write down some kind proposal/spec of how this should work.

A start would be to list some common widgets and how they should work and then sketch a solution based on that.

lebarde:
Okay!

So below is the proposition.

# Widget examples

With a short brainstorm, I imagined the following widgets:
1. VCard: a widget that [displays a hcard](https://en.wikipedia.org/wiki/HCard) (it is parsed and used by google)
2. List categories
3. List the *n* last posts (with options for excerpts, thumbnails, author, date, …)
4. Display an arbitrary text (with options allowing markdown, html, exec -- see #796)
5. Showcase/slider to display slides for a homepage, for instance
6. Links. Display arbitrary links.
7. Displaying menus!

# Specifications / capacities

- The widgets would live in `layouts/widgets/`
- They could be git repositories or maybe (I don't know how it works) with `go get`. One another (big but nice IMO) solution would be a hosting centralization, as for every CMS. The `widgets.gohugo.io` or `extensions.gohugo.io` website would host a bunch of registered plugins. That would enable creating a `hugo install` routine. But that would be for (much) later.
- The themes would declare widget areas in a way like `{{ widget_area "footer" }}`, `{{ widget_area "content_bottom" }}` or `{{ widget_area "sidebar" }}`.
- The main config file would declare a `widgets` collection, describing what is in each widget area. Each widget would be called with its name (identifier given by the widget's creator), and the user would set some other options. These options are variables specified by the widget's creator. The widgets' declaration would be possible in two ways:
 1. declaring the widgets as a collection inside the `config.toml` or `config.yaml` file. Example (for YAML mode):
```yaml
widgets:
  main-sidebar:
[...]

PR #2684, opened 2016-11-12
Call partials with specific prefixes

### LINKED ISSUE #2683: Dynamic call to template: partial call with prefix
Hello everyone,

What do you think about that?

lebarde:
You can now call a partial template with a prefix by doing this:

```
{{ partial $SomePartialName $SomePrefix . }}
```

Fixes #2683

### DIFF
--- a/tpl/template.go
+++ b/tpl/template.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/eknkc/amber"
 	"github.com/spf13/afero"
+	"github.com/spf13/cast"
 	bp "github.com/spf13/hugo/bufferpool"
 	"github.com/spf13/hugo/helpers"
 	"github.com/spf13/hugo/hugofs"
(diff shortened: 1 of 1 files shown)


lebarde:
Sorry, I don't know much about travis, which is not passing.

bep:
You need to run `gofmt` on your code.

What do you need this prefix for? Can you give a concrete example that isn't solved without it?

lebarde:
Thanks for your answer, I'll run that.
Actually it is to implement the widgets mechanism (#2535): I've done some work since my last message (sorry…).

Widget templates are called [from an `_internal` template](https://github.com/lebarde/hugo/blob/widgets/tpl/template_embedded.go#L154). Those templates are gathered from the `widgets/` directory, so I have to prefix them.

What do you think?

lebarde:
Ooops that was a mistake. I am trying to get travis work.

bep:
>What do you think?

I think at widget feature would have to be accepted or rejected as a whole, and not a lot of small pieces. I will close this now.

lebarde:
You're totally right, as I can see no use case for that excepted for widgets.

PR #2686, opened 2016-11-12
Multilingual crosspost

### LINKED ISSUE #2529: Crosspost a post on multiple languages on multilingual sites
I propose that there should be a convenient way to crosspost a post in multiple languages.
## An example use case

On my personal site I currently write articles in English first and then I create a translated version of that in Swedish. 

But I am thinking about focusing more on writing only in English. But I still want my visitors that go to the swedish site to be able to see the new content even though it is only available in English. I want my visitors to be aware that there are new content on the site.

Those posts/articles that do have a Swedish translation should of course show that translation on the Swedish site but for posts/articles that is available only in English I want to be able to crosspost them so they also show up on the swedish site.
## How to do this

Currently we have .Pages (pages in current language) and .AllPages (all languages). One possiblity is adding an optional parameter in the frontmatter of the post.

`crosspost: true`

This post would then be added to the .Pages list for all languages disregarding which language it was originally written in.

Another solution is to make a clever query between .Pages and .AllPages and looking for .Params.crosspost to list the "cross posts". This could be achieved without doing any changes to Hugo itself.

gntech:
This is the best way I have come up with for making a crosspost between languages. Discussion/critique/feedback are very welcome :)

### Crosspost between languages

Sometimes you want a certain post to show up in the index for another language, this is called a "crosspost".

For example you may have written an English post `post.en.md` that you want to show up in the index for the Swedish posts.

One way to do this is to create an empty corresponding Swedish post called `post.sv.md` but with the parameter `crosspost: en` in the frontmatter. This means that the post with language "en" will be rendered in-place of the Swedish post.

The date param is also necessary so the post will show in the right chronological order.

```yaml
---
date: 2016-10-27
crosspost: "en"
---

```

Then you have the following logic in your pages loop on the index template. What this does is that it replaces which context get rendered if the post is marked as crosspost. In this case it renders the `post.en.md` instead of `post.sv.md`.

```html
{{ if isset .Params "crosspost" }}
  {{ $c := .Params.crosspost }}
  {{ range .Translations }}
    {{ if eq $c .Lang }}
      {{ .Render "listitem" }}
    {{ end }}
  {{ end }}
{{ else }}
  {{ .Render "listitem" }}
{{ end }}
```

Fixes #2529


gntech:
@bep, @digitalcraftsman and others: What is your opinion/feedback on the solution I propose for the docs here?

bep:
I don't understand it, which isn't promising for the rest of the Hugo users. And it is very hacky and we should be able to do better. But I haven't thought too hard about the problem.

gntech:
Thanks for the feedback, I see your point with it being hacky, but I currently struggle to find a better way.

I will however definitely think this through one more time. Feel free to chip in if you have any suggestions/hints or ideas on how to better tackle this :)

digitalcraftsman:
Is there any reason why you would limit a crosspost to a single language? I would suggest that `crosspost` should take a list with language codes, e.g.

    crosspost = ["sv", "no", "de"]

If English is our default language it would also crosspost this post to the Swedish, Norwegian and German version.

From the perspective of a theme developer it would be much more complicated to loop over all pages. Instead Hugo itself should recognize this crossposts and replace the empty content of the Swedish post with the English article.

bep:
I'm pretty new to this crosspost consept, and more people should chime in that this is a great ide before we do lot of work in this area.

My in my head this should be a "first class citizen" in Hugo, i.e. I shouldn't have to creates some empty content files.

* I should define a list of languages I want to crosspost to (maybe also some additional metadata)
* The crossposts should be automatically inserted into the relevant languages' page collection, with a attribute marking them as a crosspost.
* Have some way to link back to the original ...
* ...?

Not sure.

gntech:
@bep I agree with the proposed functionality you describe above. This is exactly the ideal way I would like crossposts to work. That is a very clean way to do it. 

This pull request is focused on describing/documenting how to do best do it with the functionality that Hugo offers today.

bep:
>This pull request is focused on describing/documenting how to do best do it with the functionality that Hugo offers today.

Yes I understand that, but I think a thread on the forum would be better. Do we have a real issue about crossposts?

bep:
I will close this as per discussion above. We can do better.

PR #2687, opened 2016-11-13
Add a widget mechanism to Hugo

### LINKED ISSUE #2683: Dynamic call to template: partial call with prefix

lebarde:
Here is a first attempt to get widgets into Hugo. General tests are ok on my dev environment, but I have not already written tests for widgets, nor documentation.

Discussion is most welcome!

1. Create a site
2. Create a widget directory inside `/widgets`. It should look like this:
```
widgets/
└── text
    ├── layouts
    │   └── widget.html
    └── README.md
```
Note that the name `widget.html` is mandatory. *Currently the context is the content of the config parameter `widgets.[mywidgetarea].[mywidget].options`*. Variables are accessible with `.content` for a text widget like the following and as described in the config below.
```
{{- if isset . "content" -}}
  {{- .content | safeHTML -}}
{{- else -}}<pre>Here is a text widget, but there is nothing to print. Please define options.content inside every text widget in your config.</pre>
{{- end -}}
```

3. Configure your site with a `widgets` variable describing widgets inside widget areas:
```
widgets:
  sidebar:
    - type: text
      options:
        content: "<h1>IT WORKS from config</h1>"
        parser: html
  showcase:
    - type: text
      options:
        content: "Here lies a showcase."
  footer:
    - type: text
      options:
        content: "Powered by Hugo with widgets."
        foo: bar
```

4. Create a template using the `widgets` call. This can be done like this: `{{ widgets "sidebar" . }}`.
5. Create content. You can also use the widget's shortcode: `{{% widgets "showcase" %}}`
6. Build and enjoy.

- Currently the widgets' context is only the content of the config variable. We should add a wider context (easy).
- I have not studied the impact on performances.
- Else?

Fixes #2683
See #2535

### DIFF
--- a/hugolib/config.go
+++ b/hugolib/config.go
@@ -66,6 +66,7 @@ func loadDefaultSettings() {
 	viper.SetDefault("dataDir", "data")
 	viper.SetDefault("i18nDir", "i18n")
 	viper.SetDefault("themesDir", "themes")
+	viper.SetDefault("widgetsDir", "widgets")
 	viper.SetDefault("defaultLayout", "post")
 	viper.SetDefault("buildDrafts", false)
 	viper.SetDefault("buildFuture", false)
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -100,6 +100,7 @@ type Site struct {
 	Sections       Taxonomy
 	Info           SiteInfo
 	Menus          Menus
+	Widgets        Widgets
 	timer          *nitro.B
 	targets        targetList
 	targetListInit sync.Once
@@ -177,6 +178,7 @@ type SiteInfo struct {
 	rawAllPages           *Pages // Includes absolute all pages, including drafts etc.
 	Files                 *[]*source.File
 	Menus                 *Menus
+	Widgets               *Widgets
 	Hugo                  *HugoInfo
 	Title                 string
 	RSSLink               string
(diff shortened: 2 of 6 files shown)


lebarde:
This last commit defines the variables that can be accessed in the widget's template. The context is:
- `.$` is the general context
- `.wa` is the widget area name
- `.w` is the widget name
- `.Params` is the widget's configuration variable

bep:
When #2649 is done, it may be more clear what a `widget` is or isn't. But it is not a new special thing with special code. If *everything is a page*, then a `widget` is also a `kind` of page which can contain other widgets -- and should fit into Hugo's model, not create another on the side.

And please don't ask me *how do we do that?* I have plenty on my plate at the moment, and this isn't priority.

lebarde:
This is nice. Thank you for explaining this about widgets as nodes.

BTW, when I ask something in the forum, please consider that I am asking it to the whole community, not especially to you in particular. I will be happy if we discuss together (indeed !), but will not be offended if it happens that you do not answer to me (more if the threads are closed as if I were *persona non grata*). We are with you.

Anyway I know you are doing a hard work on Hugo nodes and this is extraordinary! I am looking forward how all this will become.

bep:
>BTW, when I ask something in the forum, please consider that I am asking it to the whole community, not especially to you in particular

Good, let us keep the discussion there.

PR #2696, opened 2016-11-15
source, tpl: fix staticcheck complaints

bogem:
tpl/template_funcs.go:1019:3: the surrounding loop is unconditionally terminated
source/lazy_file_reader.go:66:5: err != nil is always true for all possible
values ([nil:error] != [nil:error])


bep:
Merged, but note the fixed commit message.

bogem:
Oh, sorry. I forgot😓

PR #2698, opened 2016-11-16
commands: Make benchmark command more useful

### LINKED ISSUE #2432: benchmark: add --benchmem flag to benchmark command
I want to see, how many memory `hugo` consumes in the same way as `go test -bench=. --benchmem` do.

bogem:
* Add information about average time, memory consumption and
allocations.
* Fix situation, when user provides memprofile and cpuprofile, but
cpuprofile will not created.

Fixes #2432

I want to make benchmark command more useful and see, how many memory hugo consumes and average time per build.  It's just a concept, so if you have any offers, don't hesitate to express them.

### DIFF
--- a/commands/benchmark.go
+++ b/commands/benchmark.go
@@ -15,15 +15,18 @@ package commands
 
 import (
 	"os"
+	"runtime"
 	"runtime/pprof"
+	"time"
 
 	"github.com/spf13/cobra"
+	jww "github.com/spf13/jwalterweatherman"
 )
 
 var (
 	benchmarkTimes int
-	cpuProfilefile string
-	memProfilefile string
+	cpuProfileFile string
+	memProfileFile string
 )
 
 var benchmarkCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


commands/benchmark.go
```diff
@@ -37,48 +40,68 @@ func init() {
...
+		pprof.StopCPUProfile()
+		cpuProf.Close()
+	}
+
+	runtime.ReadMemStats(&memStats)
+	totalMemAllocated := memStats.TotalAlloc - memAllocated
+	totalMallocs := memStats.Mallocs - mallocs
 
+	jww.FEEDBACK.Println()
+	jww.FEEDBACK.Printf("Average time pro operation: %vms\n", int(1000*totalTime.Seconds()/float64(benchmarkTimes)))
```
bep: My English isn't native, but should it not be "per operation" here and below?

bogem: Yes, I mixed up with german

PR #2704, opened 2016-11-18
all: Use jww instead of fmt for printing

bogem:
This is an attempt to unify method for printing. I tried to discuss about these theme [here](https://discuss.gohugo.io/t/jww-vs-fmt-what-should-one-use-for-printing/4592). Yes, maybe it's not very important theme. But it's very ambiguous for new developers to understand which method should you use for printing. I want to contribute to `hugo` and make really significant contributions, but for a start I should understand and clarify some ambiguous but important moments like this or [solved problem with keys for viper](https://github.com/spf13/hugo/pull/2571).
I hope, you merge it. But if there are some problems or bugs in this PR, please tell me, I will try to fix it.

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -14,10 +14,10 @@
 package commands
 
 import (
-	"fmt"
 	"runtime"
 
 	"github.com/spf13/cobra"
+	jww "github.com/spf13/jwalterweatherman"
 )
 
 var envCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


commands/genman.go
```diff
@@ -36,7 +35,7 @@ in the "man" directory under the current directory.`,
 		header := &doc.GenManHeader{
 			Section: "1",
 			Manual:  "Hugo Manual",
-			Source:  fmt.Sprintf("Hugo %s", helpers.HugoVersion()),
+			Source:  "Hugo " + helpers.HugoVersion(),
```
bep: We have discussed this before ...

bogem: Yes😅
But I think, it's nicer and less to read (and a little bit faster too). But if you don't like it, I can revert it.

bogem: Ok, you're right, I will delete all irrelevant changes and will keep only print changes.

bep:
Could you do a rebase against master?

bogem:
But there is a merge conflict. I should first of all solve it

bogem:
Or is there an easier way?

bep:
>But there is a merge conflict. I should first of all solve it

A `rebase master` will handle merge conflicts as part of the process.

bogem:
Thanks
This PR is ready to merge

PR #2705, opened 2016-11-18
all: Work around fmt statements

bogem:
Just some obvious and insignificant PR. But maybe it can improve the reading of code and performance, especially in hugolib/site.go.


bep:
Could you do a rebase against master?

PR #2706, opened 2016-11-19
Quickstart: add a note about disabling caching when changing images

katzien:
When applying changes to images, sometimes they're not immediately applied due to Hugo's caching being turned on by default. This commit adds a note about how to disable caching to be able to view changes immediately.

### DIFF
--- a/docs/content/overview/quickstart.md
+++ b/docs/content/overview/quickstart.md
@@ -371,6 +371,8 @@ We will use the default image shown below.
 
 Hugo will sync the changes and reload the website to use new image as shown below.
 
+Note: if you don't see your changes applied immediately, restart your server with the `--ignoreCache` flag to disable caching.
+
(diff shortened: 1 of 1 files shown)


bep:
To my knowledge Hugo isn't caching any images; we cache hightlighting from Pygments and responses from Twitter.

katzien:
@bep oh, interesting! I just went through the quickstart and when it came to changing the default theme image my changes weren't reloaded immediately. When I restarted the server with caching disabled all changes were reflected instantly. The docs implied that when I change the image the server will automatically reload the changes, but nothing got triggered for me (based on the console output). I'm not 100% sure if it's due to caching or something else as I'm fairly new to Hugo, but this is what seemed to be the cause for me. Happy to do more investigation if you have any thoughts of what this could've been caused by?

bep:
Please use http://discuss.gohugo.io/ for questions/troubleshooting.

PR #2727, opened 2016-11-23
vendor: remove afero sftp deps

moorereason:
Note about updated checksums:  my hunch is that bep had extraneous files in his local gitmap and cast packages that govender used to compile a different checksum.  You can check for extra files with

```bash
cd $GOPATH/src/github.com/bep/gitmap && find . | xargs git check-ignore
```


bep:
>bep had extraneous files in his local gitmap and

Correct, but that seems to be a shady behavior by Govendor ...

PR #2730, opened 2016-11-24
commands, helpers: Add and fix docs

bogem:


helpers/language.go
```diff
@@ -93,6 +93,8 @@ func (l *Language) Params() map[string]interface{} {
 	return l.params
 }
 
+// SetParam sets param with the given key and value to l.
```
moorereason: SetParam sets a param with the given key and value.

Issue #2701, opened 2016-11-17
Refactor the globals out of site build


bep:
There currently is a controlled, but careful, dance in the tests to set a set of global state before each test. While this is controlled and has no performance impact for the end result, it it not best practice, it is hard to reason about and it prevents us from using `t.Parallel` to potentially speed up the test suite.

We have done some considerably work around this area, so when we get some of the big PRs in the pipeline merged, it may be time.

From my head this means:

* HugoSites should contain:
- [x] viper
- [x] template
- [x] hugofs
- [x] jww logger
- [x] translator/i18n bundles handling
- [x] image config cache

There are plenty of details left out in the above, but it should be complete in general.

bep:
Just a bench compare from 0.18.1:

```
▶ benchcmp -best /tmp/1.bench /tmp/2.bench
benchmark           old ns/op       new ns/op       delta
BenchmarkHugo-4     11344988814     10447443972     -7.91%

benchmark           old allocs     new allocs     delta
BenchmarkHugo-4     43590588       43615350       +0.06%

benchmark           old bytes      new bytes      delta
BenchmarkHugo-4     9136722432     9104173576     -0.36%
```

bogem:
I think, it would be better to rename and `Deps(Cfg).Cfg` to `Deps(Cfg).Provider`. The provider means, what it provides the config. 
It should be rename, because `cfg.Cfg` says nothing.

bep:
You should create a new issues for new issues, a closed issue is ... closed.

However, *provider* does not tell what it provides -- `Cfg` at least tell that it contains configuration.

PR #2731, opened 2016-11-25
helpers, hugolib, tpl: Get rid of ConfigProvider

bogem:
Less abstraction, more simplicity, more performance.

`hugo benchmark -n 30 -s examples/multilingual` before:
```
Average time per operation: 22ms
Average memory allocated per operation: 1933kB
Average allocations per operation: 30286
```

after:
```
Average time per operation: 19ms
Average memory allocated per operation: 1920kB
Average allocations per operation: 30150
```

It's just concept, but fully-working (at least it passes the tests). 

See https://github.com/spf13/hugo/issues/2728#issuecomment-262789492

More tests should be added before the merging.


bep:
What do you suggest we do when we want to add more *Spec types?

bogem:
Use `New*SpecFromCurrentLanguage` or `New*SpecFromViper`

bep:
There are several problems with your approach. At first glance it looks like a DRY problem, but that could be OK, although duplicating the property reading in the New funcs is error phrone. 

But that also leads to using a totally different code path in many of the tests vs real code.

I think the ConfigProvider is a nice and needed abstraction. I'm not sure what your benchmarks really show / do, but I'm pretty sure that this isn't worth it.

bogem:
Yes, the duplication is a problem and the obvious solution is usage of some abstraction (like `ConfigProvider`) but I thought, what this problem could be solved in more elegant way than one more abstraction. But my solution is also not an elegant way though.

bogem:
And what did you mean [here](https://github.com/spf13/hugo/pull/2731/files#diff-2fd22c31939cd0e842f01d0bd044c980L29)? What should be get rid of?

bogem:
The problem of duplication in my PR could be solved by using the `viper` as empty `Language`, so all `Get*` calls will be redirected to `viper`, although it's a workaround😅
`helpers/pathspec.go`
```
...
func NewPathSpecFromViper() *PathSpec {
    l := helpers.Language{}
    return NewPathSpecFromLanguage(l)
}
...
```

bep:
>calls will be redirected to viper, although it's a workaround

We don't do workarounds when we don't have to. The existing solution is fine, this doesn't move the issue you starting anywhere closer to a fix. The `ConfigProvider` interface may show its value more clearly when we finish the #2701. 

In the future, you may save yourself some work by discussing your plan with others before spending too much time implementing it.

bogem:
Ok, I'll try to contribute to fix #2701 issue and will discuss some ideas and solutions at first before the implementation.

bep:
Just to make it clear: I appreciate your contributions, they show real potential, and there will lots of fun tasks to solve in Hugo in the near future, some of the fun being that it is actually being used by a lot of users.

bogem:
Thank you, @bep 😌
Yes, I understand it :)

Issue #2743, opened 2016-11-28
Allow default title to have initial caps


mdhender:
Would like to specify the name of new content in all lower case and have the default title created in the front matter converted to initial caps.
```
$ hugo new post/init-caps.md
```
Current output:
```
+++
date = "2016-11-27T18:58:19-06:00"
title = "init caps"

+++
```
Would like to see
```
+++
date = "2016-11-27T18:58:42-06:00"
title = "Init Caps"

+++
```

bep:
Note that title casing is a style not used in all languages, also title casing isn't probably what you really want in this case, see #989.

So, to get this right would be a challenging task.

mdhender:
If no one objects, I'll name the config option "newContentTitleFormat" with this fix looking for "initCaps". That way I can avoid title case completely.

bep:
This will be handled in round-about way in #2746

Issue #2746, opened 2016-11-28
Allow specifying title from the command line, and auto-derive the file name


kaushalmodi:
Hello,

This is related to https://github.com/spf13/hugo/issues/2743, but this feature request is to do the opposite... The user specifies the title from the command line with whatever case they like, and then the file name is auto-generated from that. 

So if the title specified is "My New Post!", the file name will be "posts/my-new-post.md". Here are the transformation rules:

- The file name is made all-lowercase. 
- Spaces are replaced with hyphens.
- Punctuation marks like `!` and `?` are removed.
- The file is created in `posts/`. May be the content sub dir `posts` be specified as a part of the command line argument? Example: `hugo <new option for this feature> posts 'My New Post!'`. So the first argument is the name of the content sub dir. And the rest of the arguments are part of the title. *Single quoting is used because of the `!` character.*

I have implemented something like above as a bash wrapper script [here][1].

[1]: https://discuss.gohugo.io/t/auto-generate-file-name-based-on-title/4648/2?u=kaushalmodi

bep:
I suggest adding a couple of flags to new, so:

```
hugo new --title="My Cool Post!" --section=post --markdown=md
```

Or without, as it is today:

```
hugo new /post/mycoolpost.md
```

kaushalmodi:
Thanks. That would be awesome, as that can be easily aliased too. Looking forward to it.

kaushalmodi:
> hugo new --title="My Cool Post!" --section=post --markdown=md

A quick question.. That will result in `post/my-cool-post.md`, correct? (Note the replacement of spaces with hyphens).

bep:
Proably

```
post/my-cool-post!.md
```

kaushalmodi:
I was thinking something along these lines. .to completely sanitize the file names:

```bash
    # Using ${title}
    #  - Replace spaces with underscores
    #  - Lower-case the whole title
    #  - Replace '&' with "and"
    #  - Remove all characters except a-z, 0-9 and '-'
    # For example, converts "My New \& Awesome Post!" to "my-new-and-awesome-post.md"
    # (Note that we need to escape & with \ above, in the shell.)
    fname=$(echo "${title}" \
                | \sed -r -e 's/ /-/g' \
                       -e 's/./\L\0/g' \
                       -e 's/&/and/g' \
                       -e 's/[^a-z0-9-]//g').md;
```

bep:
We have functions for that already in Hugo.

[...]

PR #2756, opened 2016-12-01
Add new helper MakeTitleInitCaps

mdhender:
Updates the hugo new command to allow creating the title using an
initial capital letter format. Adds a new helper function,
MakeTitleInitCaps, and a new config file option, coerceTitleFormat.
If set to "initCaps", then createMetaData calls the new helper
instead of MakeTitle.

Implements #2743

### DIFF
--- a/create/content.go
+++ b/create/content.go
@@ -132,7 +132,13 @@ func createMetadata(archetype parser.Page, name string) (map[string]interface{},
 	}
 
 	if _, ok := metadata["title"]; !ok {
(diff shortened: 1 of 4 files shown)


bep:
No, we're not adding more magic options that subtly changes the behavior. It is hard to explain to people (needs doc), is hard to understand, is of limited use for Norwegian users and many more -- and it has to be maintained.

People (should) care about titles not about file names, so it makes more sense to create articles based on that and not the other way around.

This will be handled properly in #2746.

PR #2772, opened 2016-12-10
Docs: clarify using archetypes from a theme

ebr:
Adding a clarification to archetype docs.

I spent a few hours bashing my head against this, because I was running `hugo server --theme=THEME`, but the `THEME` was not in my `config.toml`. So none of the theme archetypes would work. Eventually I found https://discuss.gohugo.io/t/help-with-archetypes/1118/5, and :bulb:.

Hopefully this helps someone.


digitalcraftsman:
Hello @ebr,

Thanks for submitting this clarification. I merged it as 38f556d.

moorereason:
@ebr, would you rather be able to do `hugo new --theme=THEME foo/page.md`?  Currently, the `--theme` option is not available to the `new` sub-command.

ebr:
@moorereason that's a great idea and it does make sense to keep options consistent between subcommands. It would be useful in my build scripts, but by no means critical, now that the behaviour is clearly understood.

PR #2777, opened 2016-12-12
new command: hugo gen index

joescharf:
This command will create an index of site content and output a json file (that can be used with lunr.js) or update an index on Algolia.

I regenerated the docs, and just committed the `hugo_gen.md` and `hugo_gen_index.md` files that changed/were new, something is wrong when I `hugo serve` the site though, 404 on the hugo_gen_index file.


bep:
To get the build to go green you need to add the new dependency via `govendor`

bep:
Thanks for your contribution. The code looks solid enough, but we sill solve the "custom output" problem in a more general way.

PR #2778, opened 2016-12-12
docs: Add Extras > GitInfo

### LINKED ISSUE #2670: Document the GitInfo feature
See e8380e612fae18493556d620726a9069a17128db.

moorereason:
This PR contains an additional commit to cleanup the Extras menu ordering.  It was an ugly mess.  I removed the weights from the pages to allow the sorter to use the Title so that we don't need manage weights by hand.  Make sure I didn't break something by doing that.

PS - Notice the definition list in the GitInfo page.  I'd much rather see us use that structure in pages such as Template Variables.

Fixes #2670

### DIFF
--- a/docs/content/extras/gitinfo.md
+++ b/docs/content/extras/gitinfo.md
@@ -18,6 +18,8 @@ Hugo provides a way to integrate Git data into your site.
 
 1. The Hugo site must be in a Git-enabled directory.
 1. The Git executable must be installed and in your system `PATH`.
+1. Enable the GitInfo feature in Hugo by using `--enableGitInfo` on the command
+   line or by setting `enableGitInfo` to `true` in your site configuration.
 
 ## The GitInfo Object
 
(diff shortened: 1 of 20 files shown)


docs/content/extras/gitinfo.md
```diff
@@ -0,0 +1,48 @@
...
+Hash
+: commit hash, e.g. `866cbccdab588b9908887ffd3b4f2667e94090c3`
+
+Subject
+: commit message subject, e.g. `tpl: Add custom index function`
+
+
+## Performance Considerations
+
+The Git integrations should be fairly performant, but it does add some time to the build, which depends somewhat on the Git history size.
```
bep: We could add this later, but it would be tempting to add some info about the `enableGitInfo` in this context. It is also a "Prerequisite" of sorts, but it is also both a config option and a flag ... and the flag is useful for bigger sites when in dev mode.

Issue #2788, opened 2016-12-14
MakePath function produces unexpected URLs with certain input


bryanjeal:
When I `|urlize` certain strings such as:
- "♪ Short musical interlude ♬"
- "Foo & bar"

I would **expect** to get:
- "Short-musical-interlude"
- "Foo-bar"; 

however, I **actually** get: 
- "-Short-musical-interlude-"
- "Foo--bar"

The double `--` and the `-` at the beginning and end of the url isn't what other slugifiers would produce (https://docs.djangoproject.com/en/1.10/_modules/django/utils/text/#slugify)

bryanjeal:
I have submitted a pull request with a possible fix for this issue. #2789

bep:
We will have to think a little about this. The URL produced are perfectly valid, and changing this would be somewhat a breaking change.

bryanjeal:
My previous experience has been with PHP and Django, so that is what has shaped my expectations and prompted this bug report.

The slugifiers I have used all produce outlook similar to the fix (although do default to lowercase as well as eliminating multiple/trailing dashes).

Here are a couple other slugifiers so you can see what I mean:
- https://github.com/cocur/slugify
- https://www.npmjs.com/package/slugify
- https://github.com/slugify/slugify

Thank you for your time on this issue.

Artiavis:
Not sure if this is the best place to put it, but I have a related concern that I didn't see in the official docs. I have an article titled "F# vs Clojure Shootout", which gets urlized as "f#-vs.-clojure-toy-problem-shootout", which is obviously a problem since "#" gets interpreted as an anchor when it's actually a semantically valid piece of the title. Can you either warn about this case or escape it automatically? (I'm porting my blog from another site generator where titles weren't automatically mapped to URL's so I didn't have this problem.)

bep:
See https://godoc.org/github.com/russross/blackfriday#hdr-Sanitized_Anchor_Names

jmooring:
The behavior changed a long time ago. With v0.111.3 and default site configuration:

```text
{{ urlize "♪ Short musical interlude ♬" }} --> short-musical-interlude
{{ urlize "Foo-bar" }} --> foo-bar
```

PR #2789, opened 2016-12-14
helpers: Remove extra '-' from MakePath (Fix for issue 2788) 

bryanjeal:
This Pull Request is a fix for issue #2788. 

MakePath function would leave extra '-' characters in its output when Unicode Characters were removed after strings.Replace is run.


bep:
You need to run `go fmt` on your code to get the CI to go green.

bryanjeal:
Sorry about that @bep. 

I fixed a conflict in github's editor and that messed with the formatting. Should be fixed now.

moorereason:
@bryanjeal, please sign the CLA.  Otherwise, you can close this PR.  Thanks.

helpers/path.go
```diff
@@ -124,7 +127,7 @@ func (p *PathSpec) UnicodeSanitize(s string) string {
 	for i, r := range source {
 		if r == '%' && i+2 < len(source) && ishex(source[i+1]) && ishex(source[i+2]) {
 			target = append(target, r)
-		} else if unicode.IsLetter(r) || unicode.IsDigit(r) || unicode.IsMark(r) || r == '.' || r == '/' || r == '\\' || r == '_' || r == '-' || r == '#' || r == '+' || r == '~' {
+		} else if unicode.IsLetter(r) || unicode.IsDigit(r) || unicode.IsMark(r) || r == '.' || r == '/' || r == '\\' || r == '_' || r == '-' || r == '#' || r == '+' || r == '~' || r == ' ' {
```
bep: This should be superflous if you adjust the order of the above.

bep:
What is the performance implications of this change?

helpers/path.go
bryanjeal: Please excuse my ignorance, but I can't seem to reorder the operations in the `MakePath` function to make adding the space to the `UnicodeSanitize` unnecessary.

Changing to: Regex ReplaceAll -> UnicodeSanitize -> Trim 
causes the test case: `{"Foo & bar", "Foo-bar", true}` to fail.

bryanjeal:
@bep please find the benchmark results below:

### Code

```
func BenchmarkMakePath(b *testing.B) {
	v := viper.New()
	l := NewDefaultLanguage(v)
	v.Set("removePathAccents", true)
	p, _ := NewPathSpec(hugofs.NewMem(v), l)

	for i := 0; i < b.N; i++ {
		_ = p.MakePath("Foo & bar")
	}
}
```

### Results

Results based on the code currently in `gohugoio/hugo` **master**:
```bash
$ go test -bench=MakePath
BenchmarkMakePath-8   	  500000	      2976 ns/op
PASS
```

Results based on the code in this change request:
```bash
$ go test -bench=MakePath
BenchmarkMakePath-8   	  300000	      3621 ns/op
PASS
```

PR #2804, opened 2016-12-20
Update multilingual.md

asifm:
There was a mix of TOML and YAML in the first example config. Changed the YAML to TOML. Please check if the syntax is correct (I'm new to TOML).


digitalcraftsman:
Thanks for the submission @asifm.

PR #2817, opened 2016-12-22
command: Add `themesDir` command argument to specify themes directory

avinassh:
Currently, I don't see a way to specify themes directory to look for, while running the `hugo` command. So I added this command line argument which takes the path where themes are located. 

Example usage:

    $ hugo server --theme=docuapi --themesDir=/Users/avi/snakes/hugo-themes
    $ hugo --theme=docuapi --themesDir=/Users/avi/snakes/hugo-themes

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -146,6 +146,7 @@ var (
 	destination string
 	logFile     string
 	theme       string
+	themesDir   string
 	source      string
 )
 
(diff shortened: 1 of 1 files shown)


bep:
I guess you can do 

```
env HUGO_THEMESDIR="/my/themes/ hugo --theme=docuapi
```

But you are right, this should be a CLI option.

avinassh:
^Oh! I did not know about that option. Even Google also didn't help.

Is there any changes I need to make for this PR?

PR #2825, opened 2016-12-23
helpers: Remove "no theme set" warning

### LINKED ISSUE #2821: Remove warning "WARN: 2016/12/22 23:04:30 hugo.go:469: No theme set"

moorereason:
Fixes #2821

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -472,7 +472,9 @@ func getStaticSourceFs() afero.Fs {
 	useStatic := true
 
 	if err != nil {
-		jww.WARN.Println(err)
+		if err != helpers.ErrThemeUndefined {
+			jww.WARN.Println(err)
+		}
 		useTheme = false
 	} else {
 		if _, err := source.Stat(themeDir); os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


bep:
The fact that `getThemeDirPath` returns an error looks suspicious. I would expect it to return just a "" to signal "no theme" -- and let the caller decide if that is OK or not -- which it obviously is.

moorereason:
I considered that, but looking at how `getThemeDirPath` is being used, this looked like the cleanest solution.  Take the following example from [hugolib.site.go#L754-L757](https://github.com/spf13/hugo/blob/34c87421b82b967a8ae85261ca9a3e9fe7b156da/hugolib/site.go#L754-L757):

```go
	themeDataDir, err := helpers.GetThemeDataDirPath()
	if err == nil {
		dataSources = append(dataSources, &source.Filesystem{Base: themeDataDir})
	}
```

If we would modifed `getThemeDirPath` to `return "", nil` when no theme is set, we would need to update these calls to something like this:

```go
	themeDataDir, err := helpers.GetThemeDataDirPath()
	if err == nil && themeDataDir != "" {
		dataSources = append(dataSources, &source.Filesystem{Base: themeDataDir})
	}
```

What I've submitted avoids these extra checks and lets us assume that an error returned means we can ignore the returned path.

PR #2831, opened 2016-12-25
commands: Support human-readable YAML boolean values in undraft

kropp:
In addition to true/false, undraft now supports yes/no, y/n and on/off in different cases.


kropp:
YAML boolean type description in documentation: http://yaml.org/type/bool.html

bep:
Thanks for the effort, but I feel that this complicates the code more than the value it gives.

kropp:
What if `undraft` would remove draft attribute altogether? This would eliminate the need to change its value and handle multiple options and will produce correct results in all cases.

Issue #2752, opened 2016-11-30
TOML array not parsed


stp-ip:
Within my front-matter I have the following toml:
```
+++
title = "$title"
joined = "2011-01-01"
img = "3.jpg"
position = "$position"
social = [
  ["fa-twitter", "#"],
  ["fa-facebook", "#"],
  ["fa-linkedin", "#"]
]
+++
```
Rendering {{ .Params }} returns an empty social [ ].

I retried using yaml:
```
---
title: "$title"
joined: "2011-01-01"
img: "3.jpg"
position: "$position"
---
social:
  -
    - "fa-twitter"
    -  "#"
  -
    - "fa-facebook"
    -  "#"
  -
    - "fa-linkedin"
    -  "#"
---
```

Version used: Hugo master and Hugo 0.17 were tested. Both render an empty social [  ].

bep:
This was fixed like in the last week -- are you **really** sure you have tested this with the latest Hugo?

stp-ip:
Will definitely recheck, but I did use "go get -u + go build, which should have updated and rebuild".

stp-ip:
Hugo Static Site Generator v0.18-DEV BuildDate: 2016-11-30T08:57:46Z

stp-ip:
git log for the golang source: "971d1ba commands: Add noChmod option to disable perm sync"

bep:
That is sadly not a proof since the fix was in the go-toml library. We use govendor to do vendoring, and I must admit that I'm no expert here ... But the makefile should help out.

stp-ip:
Triple checked. Updated go vendor, updated deps, rebuild. Using make and without. Toml still comes up empty.

stp-ip:
Might this be an issue with using arrays within arrays? It seems with the right syntax on both yaml and toml, both return an empty array.

bep:
You are right, Hugo does something weird in this case.

bep:
@moorereason I'm pretty sure this is not a 0.18 regression.

PR #2837, opened 2016-12-26
parser: Refactor frontmatter parser and add tests

moorereason:
Lots of cleanups here:

- Refactor InterfaceToConfig and InterfaceToFrontMatter to use io.Writer.
- Simplify InterfaceToFrontMatter by wrapping InterfaceToConfig.
- Export FrontmatterType since we return it in DetectFrontMatter.
- Refactor removeTOMLIdentifier to avoid blindly replacing "+++".
- Update HandleJSONMetaData to return an empty map on nil input.
- Add godoc comments.

Coverage for frontmatter.go increased from 10.7% to 90.7%.

Updates #2752

### DIFF
--- a/parser/frontmatter.go
+++ b/parser/frontmatter.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"io"
 	"strings"
 
 	"github.com/chaseadamsio/goorgeous"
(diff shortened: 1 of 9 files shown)


moorereason:
Pushed an additional commit to add tests and godocs to page.go.

Coverage of page.go increased from 66.4% to 81.9%.

parser/frontmatter.go
```diff
@@ -174,17 +178,51 @@ func HandleTOMLMetaData(datum []byte) (interface{}, error) {
...
+// removeTOMLIdentifier removes, if necessary, beginning and ending TOML
+// frontmatter delimiters from a byte slice.
 func removeTOMLIdentifier(datum []byte) []byte {
-	return bytes.Replace(datum, []byte(TOMLDelim), []byte(""), -1)
+	if len(datum) < 8 {
+		return datum
+	}
+
+	b := bytes.TrimPrefix(datum, []byte(TOMLDelim))
+	if len(datum)-len(b) == 3 {
```
bep: Something tells me that you can use bytes.TrimFunc or similar to handle this and below.

parser/frontmatter_test.go
```diff
@@ -14,9 +14,210 @@
...
+		{
+			map[string]interface{}{"title": "test 1"},
+			'+',
+			[]byte("title = \"test 1\"\n"),
+			false,
+		},
+
+		// YAML
+
+		// FIXME(moorereason): ?? returns "{}\n"
```
bep: ??

moorereason: The following commented-out test for an empty map returns `{}\n` for YAML conversion when I expect it to return `nil`.  Didn't take time to look into it before, but YAML apparently can define an empty map as `{}`.  I'd prefer that the results of each format all return a `nil` slice in that situation.  What do you think?  Do we check for `[]byte("{}\n")` and return `nil`?

```go
// {map[string]interface{}{}, '-', nil, false},
```

parser/frontmatter.go
moorereason: Added a commit to simplify this func.  Please review.

parser/frontmatter_test.go
moorereason: Fixed.

moorereason:
Rebased, squashed, and added Org frontmatter tests.

moorereason:
Pushed another commit to track changes from chaseadamsio/goorgeous#23.

parser/page_test.go
```diff
@@ -63,8 +63,8 @@ func TestPage(t *testing.T) {
 			orgPageFrontMatter,
 			true,
 			map[string]interface{}{
-				"Title":      "Org Test 1",
-				"Categories": []string{"a", "b"},
+				"TITLE":      "Org Test 1",
+				"CATEGORIES": []string{"a", "b"},
```
bep: Was this an improvement, you think?

moorereason: Yes.

bep: If case insensitivity was the motivation, you have gone from testing mix-case to upper case only.

But it's not important.

I forgot about this PR. I didn't want to pull it in late in the 0.19 game, but now is the time, but now there are some conflicts.

moorereason: Fair point.  I'll add mixed case when I rebase.

moorereason:
Updated org-mode tests and the vendored goorgeous package.  Rebased and squashed.

PR #2841, opened 2016-12-28
hugolib: Allow arrays of arrays in frontmatter Params

### LINKED ISSUE #2752: TOML array not parsed

moorereason:
Fixes #2752

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -1018,6 +1018,8 @@ func (p *Page) update(f interface{}) error {
 							p.Params[loki] = vvv
 						case map[string]interface{}: // Proper parsing structured array from JSON based FrontMatter
 							p.Params[loki] = vvv
+						case []interface{}:
+							p.Params[loki] = vvv
 						default:
 							a := make([]string, len(vvv))
 							for i, u := range vvv {
--- a/hugolib/page_test.go
+++ b/hugolib/page_test.go
@@ -1261,6 +1261,61 @@ func TestDraft(t *testing.T) {
 	}
 }
 
+var pagesParamsTemplate = []string{`+++
+title = "okay"
+draft = false
+tags = [ "hugo", "web" ]
+social= [
+  [ "a", "#" ],
+  [ "b", "#" ],
+]
++++
+some content
+`,
+	`---
+title: "okay"
+draft: false
+tags:
+  - hugo
+  - web
+social:
+  - - a
+    - "#"
+  - - b
+    - "#"
+---
+some content
+`,
+	`{
+	"title": "okay",
+	"draft": false,
+	"tags": [ "hugo", "web" ],
+	"social": [
+		[ "a", "#" ],
+		[ "b", "#" ]
+	]
+}
+some content
+`,
+}
+
+func TestPageParams(t *testing.T) {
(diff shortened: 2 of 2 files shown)


hugolib/node_as_page_test.go
```diff
@@ -385,6 +386,25 @@ title = "Deutsche Hugo"
 	assertFileContent(t, expectedFilePath(ugly, "public", "nn", "sect1", "regular2"), true,
 		"Single Title: Page 02")
 
+	// Page Params
+	pages := sites.findAllPagesByKind(KindPage)
+	for _, p := range pages {
+		if p.Path() == "sect1/regular1.en.md" {
```
bep: Are you sure the above is truthful on Windows?

bep: Also, when you need to add an import to get to the assert package, you by definition is breaking the standard set in the file. I think the require package is suitable in most situations. If something fails, it is bad and I want it to stop.

bep: Also, and I know I have been a sinner in area, adding more test cases to an existing very broad test to test for a very specific thing smells a little. I think we have narrower unit-typish tests for these params conversions -- and if not, we should.

moorereason: 1. Good catch.  I need to use `filepath.Join`.

2. I didn't notice your standard.  Apparently, I have a different standard.  :grin: I use `assert` where I would use `t.Errorf` and `require` where I would use `t.Fatalf`.  I only use `Fatalf`/`require` to avoid accessing nil pointers and such in subsequent tests.  So, when I use `require`, it's a signal to me that we've got a serious problem.  If a Test func contains numerous assertions, I like to see them all fail instead of just the first one.  With `require` I'd only see the first failured assertion.  If you don't like my reasoning, I'll change to `require` in this file.

3. I agree, but I was trying to follow your lead.  :grinning:  Sometimes I'm not quite sure how to setup all the site scaffolding to run a simple unit test.  I'll take another look at the tests in this PR to see if I can pull them out into a more narrow test.

bep: The node_as_page_test was something I built to be able to move from node to page, and is a little bit of everything, and I was thinking of maybe remove it, or move the tests.

Testing TOML to interface{} conversions sounds like a job for a unit test.

As to require: If you have one failing test you have a serious problem -- and fixing that one test is usually enough to fix any tests that come after it. But up to you, my point was: If there are 100 require in a test file, then adding one single assert is at least breaking a pattern.

PR #2851, opened 2016-12-29
tpl/lang: Add NumFmt function

### LINKED ISSUE #1444: Outputting numbers with thousands separators

moorereason:
numFormat formats a number with a given precision using the requested decimal,
grouping, and negative characters.

Fixes #1444

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -460,6 +460,24 @@ e.g.
 
 * `{{ int "123" }}` → 123
 
+### lang.NumFmt
+
+`NumFmt` formats a number with the given precision using the *decimal*,
+*grouping*, and *negative* options.  The `options` parameter is a
+string consisting of `<negative> <decimal> <grouping>`.  The default
+`options` value is `- . ,`.
+
+Note that numbers are rounded up at 5 or greater.
+So, with precision set to 0, 1.5 becomes `2`, and 1.4 becomes `1`.
+
+```
+{{ lang.NumFmt 2 12345.6789 }} → 12,345.68
+{{ lang.NumFmt 2 12345.6789 "- , ." }} → 12.345,68
+{{ lang.NumFmt 0 -12345.6789 "- . ," }} → -12,346
+{{ lang.NumFmt 6 -12345.6789 "- ." }} → -12345.678900
+{{ -98765.4321 | lang.NumFmt 2 }} → -98,765.43
+```
+
 ## Strings
 
 ### printf
--- a/tpl/lang/init.go
+++ b/tpl/lang/init.go
@@ -34,6 +34,16 @@ func init() {
 			[][2]string{},
 		)
 
+		ns.AddMethodMapping(ctx.NumFmt,
+			nil,
+			[][2]string{
+				{`{{ lang.NumFmt 2 12345.6789 }}`, `12,345.68`},
+				{`{{ lang.NumFmt 2 12345.6789 "- , ." }}`, `12.345,68`},
+				{`{{ lang.NumFmt 6 -12345.6789 "- ." }}`, `-12345.678900`},
+				{`{{ lang.NumFmt 0 -12345.6789 "- . ," }}`, `-12,346`},
+				{`{{ -98765.4321 | lang.NumFmt 2 }}`, `-98,765.43`},
+			},
+		)
 		return ns
 
 	}
--- a/tpl/lang/lang.go
+++ b/tpl/lang/lang.go
@@ -14,6 +14,11 @@
 package lang
 
 import (
+	"errors"
+	"math"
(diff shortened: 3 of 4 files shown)


moorereason:
Replaces #1450.

This should be simpler than the previous attempt.  Formats the number using `strconv` and then inserts/replaces  characters where needed.  Should be easier to maintain.

bep:
@moorereason if you rebase this, I will merge it. I see many people ask for it, and once we get some aliasing and namespaces in order, I'm not too worried about having "many" funcs. But double check the arg order (and example) so it fits nicely into a pipe.

moorereason:
Added a pipe example and rebased.

docs/content/templates/functions.md
```diff
@@ -460,6 +460,19 @@ e.g.
...
+
+`numFormat` formats a number with a given precision using the requested *decimal*,
+*grouping*, and *negative* characters.  The function signature is
+`numFormat <decimal> <grouping> <negative> <precision> <number>`.
+
+```
+{{ numFormat "," "." "-" 2 12345.6789 }} → 12.345,68
+{{ numFormat "." "" "-" 6 -12345.6789 }} → -12345.678900
+{{ numFormat "." "," "-" 0 -12345.6789 }} → -12,346
+{{ -12345.6789 | numFormat "." "," "-" 2 }} → -12,346.68
```
digitalcraftsman: Shouldn't the result be `-12,345.68`

bep: While you're in that file, you may also specify what rounding rule used.

moorereason: Doh! typo

moorereason:
Pushed a new commit.  Please review.

moorereason:
@bep, we can merge this after #3415, but I made some changes from my original PR:

- Moved this function to the `lang` namespace since this is really a localization concern.
- Removed shorthand `numFormat` from funcMap.  Can only access via namespace.
- Renamed from `numFormat` to `NumFmt`.

moorereason:
@bep, this is ready for review.

docs/content/templates/functions.md
```diff
@@ -460,6 +460,22 @@ e.g.
...
+
+`NumFmt` formats a number with a given precision using the requested *decimal*,
+*grouping*, and *negative* characters.  The function signature is
+`NumFmt <decimal> <grouping> <negative> <precision> <number>`.
+
+Note that numbers are rounded up at 5 or greater.
+So, with precision set to 0, 1.5 becomes `2`, and 1.4 becomes `1`.
+
+```
+{{ lang.NumFmt "," "." "-" 2 12345.6789 }} → 12.345,68
```
bep: Thinking about this, the signature looks kind of klunky for a template func.

* I kind of want to put the format into my site config, so people can set their own.
* I doubt this will work: `myNumFormat = [",", ".", "-"]`
* Suspect this will be better and then use `strings.Fields`:  `myNumFormat = ", . _"`

Other than that it looks fine.

moorereason: Good thoughts.  How about we build it with the intention that it will work with l10n later.  Like i18n, the options are defined in the site config.

```go
func NumFmt(precision int, n float64, opts ...interface{}) string
```

```
// Assume that they've defined locale number formatting in the site config.
{{ lang.NumFmt 2 123.456 }}
{{ 123.456 | lang.NumFmt 2 }}

// Given multiple locales defined in config, allow to specify which one to use.
{{ lang.NumFmt 2 123.456 "loc:en_US" }}

// Allow use without a locale definition?
{{ lang.NumFmt 2 123.456 "dec:." "grp:," "neg:-" }}
```

We don't have to support all of those options, but the func signature would allow us flexibility in the implementation as we move toward l10n.

If we need to think through l10n before merging this, I'm okay with that.

bep: We have no concept of locale or "locale formatting". What you are saying is that we should wait until we see how that looks like before we revisit this one.

bep: But I think we should just drop the locale debate, and just pack the three arguments into one string. It should be possible to do some magic locale stuff later if needed.

moorereason:
@bep, how's this look?  I still need to add site config support.

bep:
This looks great. Don't bother with the site config support; we will have to revisit this at some point. With this, people can easily put that config in site params.

Issue #225, opened 2014-03-09
customize the template for .TableOfContents


ktmud:
I want to make it work with Bootstrap's [scrollspy](http://getbootstrap.com/javascript/#scrollspy), but the html structure seems to be hardcoded.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

sagikazarmark:
I would like to bring this up again, because it would be a nice feature. Currently the TOC generator outputs quite an ugly structure which is hard to use together with existing TOC styling and scrollspy plugins (like the ones in Twitter Bootstrap and Materialize CSS).

Heydon:
+1 for greater template control.

In addition, *all* of my table of contents renders inside a single `<li>`. This is not correct, semantically — essentially it is to say that the table of contents is a list of one item?

neeklamy:
@Heydon, the table of contents placed inside a top level `li` is by design, you can see the test code for this in [page_test.go#L807](gohugoio/hugo/blob/873a6f18851bcda79d562ff6c02e1109e8e31a88/hugolib/page_test.go#L807). I’m at a loss as to why, I’ve been trying to figure out the same thing.

I’ve asked about this over at the [\[Resolved\] .TableOfContents variables displays an extra bullet point –Hugo Discussions](https://discourse.gohugo.io/t/resolved-tableofcontents-variables-displays-an-extra-bullet-point/2303).

Heydon:
@neeklamy Thanks for the follow-up. I imagine it's simply a quirk of the logic to build the nested list. But it should be amended.

kizu:
For those who'd want to have a quick and dirty solution for this:

```
{{ .TableOfContents | replaceRE "<nav([^>]*)>\\s*<ul>\\s*<li>([\\s\\S]+)</li>\\s*</ul>\\s*</nav>" "<nav$1>$2</nav>" | safeHTML }}
```

This would remove the initial wrapping `ul>li`. We can also use `replaceRE` to change things inside further if needed.

[...]

PR #2852, opened 2016-12-29
Add a TocEntries variable to Page

bjchambers:
This is generated by recording the headers processed during markdown
rendering.

This allows rendering the TOC using a template rather than relying on
the built-in markdown formatted TOC.

### DIFF
--- a/docs/content/extras/toc.md
+++ b/docs/content/extras/toc.md
@@ -19,10 +19,10 @@ Simply create content like you normally would with the appropriate
 headers.
 
 Hugo will take this Markdown and create a table of contents stored in the
-[content variable](/layout/variables/) `.TableOfContents`
+[content variable](/layout/variables/) `.TableOfContents`.
 
 
-## Template Example
+### Template Example
 
 This is example code of a [single.html template](/layout/content/).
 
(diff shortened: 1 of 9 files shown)


bjchambers:
I'm sure I've made a huge number of style and/or convention mistakes, since I haven't written much Go and wasn't really sure how things in hugo fit together.

I've wanted the ability to render a custom table of contents based on the headers within a document for a while, so figured I'd throw something together and see if there was any interest.

bjchambers:
I think this change could also be useful for addressing #225 since it allows creating the actual TOC using a partial.

bjchambers:
Some things I'm not completely sure about here:

1. How to handle dangling TOC entries (eg., skipping to a level 3 nested in a level 1). To preserve the tree-like structure, there needs to be a level entry within the level 1 that contains the level 3 entry. In the current code, this is handled by introducing a placeholder node, which allows the template to either skip the placeholders, or include just their contents, etc.

2. Representation of the TocEntry structure. Since it gets built up during rendering it needs to be mutable, but the use of the array of pointers to entries seems ugly. An alternative I considered was just doing a list-like approach, where each entry has a child pointer and a next pointer. Then rather than using `Clone` to create the template-visible value, these pointers could be traversed and turned into some easier to consume form.

3. Whether the TocEntry built during markdown rendering needs to be a tree. Another option I considered (which may be cleaner) is to just expose the list of entries reported, and then have some functions available to the templates to access those in different ways (eg., turn the list into a tree with placeholders, turn the list into a tree discarding dangling entries, etc.)

Happy to discuss these or any other considerations.

moorereason:
Please open a thread in the `dev` section of the forums to discuss this.

bep:
I'm sorry for this late response, but this is too big a change for what it is. I agree about the end result, but this can be done much simpler in Blackfriday.

PR #2856, opened 2016-12-29
Add link to HugoPhotoSwipe tool

GjjvdBurg:
HugoPhotoSwipe is a command line tool to create and manage PhotoSwipe galleries with Hugo. It resizes images for responsive layouts in PhotoSwipe and generates the Markdown necessary for Hugo to create the gallery. I think this can be a useful tool for Hugo users.


digitalcraftsman:
Thanks for sharing this tool.

PR #2859, opened 2016-12-31
tpl: add now function

nishanths:
Add a timeNow template function that returns the current time as
time.Time. Also, update related docs.

Don't know what the right way to test this change is (if necessary), so I'm open to suggestions.

Personally, I'm looking to use use the unix timestamp of the current time in the [query string for cache busting](https://css-tricks.com/strategies-for-cache-busting-css/#article-header-id-1) in a theme. 

```
<link rel="stylesheet" href="/css/reset.css?t={{ timeNow.Unix }}">
```

I'm sure hugo users will come up with other creative uses for current time.

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -2146,6 +2146,7 @@ func initFuncMap() {
 		"modBool":       modBool,
 		"mul":           func(a, b interface{}) (interface{}, error) { return helpers.DoArithmetic(a, b, '*') },
 		"ne":            ne,
+		"now":           func() time.Time { return time.Now() },
 		"partial":       partial,
 		"partialCached": partialCached,
 		"plainify":      plainify,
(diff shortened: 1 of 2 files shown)


moorereason:
`time.Now` is [already accessible on the Page object](https://github.com/spf13/hugo/blob/bcb3ef74131f8119e290ddb68908ce27760a80a7/hugolib/page.go#L1578-L1580).

However, I've started a discussion referencing this PR: https://discuss.gohugo.io/t/4968

nishanths:
Updated the function name to `now`

tpl/template_funcs.go
```diff
@@ -2186,6 +2186,7 @@ func initFuncMap() {
 		"substr":       substr,
 		"title":        title,
 		"time":         asTime,
+		"now":          func() time.Time { return time.Now() },
```
bep: I know this fits fine next to time, but the rest of them are alphabetic ordered.

nishanths: done, thanks

PR #2880, opened 2017-01-03
tpl: Fix Unix time test case in TestDateFormat()

### LINKED ISSUE #2855: TestDateFormat Unix epoch test is timezone sensitive and fails in the Americas
The following test case in TestDateFormat() in tpl/template_func_tests.go:

```go
		{"Monday, Jan 2, 2006", 1421733600, "Tuesday, Jan 20, 2015"},
```

is timezone sensitive and thus prone to FAIL in the Western Hemisphere (maybe not the most accurate term, but yeah, at least the Americas continents.)

For example, in Edmonton, Canada, where the current timezone is MST (-0700),

```
$ TZ=America/Edmonton go test ./...
?   	github.com/spf13/hugo	[no test files]
ok  	github.com/spf13/hugo/bufferpool	0.006s
ok  	github.com/spf13/hugo/commands	0.026s
ok  	github.com/spf13/hugo/create	0.015s
ok  	github.com/spf13/hugo/helpers	3.090s
ok  	github.com/spf13/hugo/hugofs	0.020s
ok  	github.com/spf13/hugo/hugolib	11.818s
?   	github.com/spf13/hugo/livereload	[no test files]
ok  	github.com/spf13/hugo/parser	0.024s
ok  	github.com/spf13/hugo/source	0.027s
ok  	github.com/spf13/hugo/target	0.022s
--- FAIL: TestDateFormat (0.00s)
	template_funcs_test.go:2293: [3] DateFormat got Monday, Jan 19, 2015 but expected Tuesday, Jan 20, 2015
FAIL
FAIL	github.com/spf13/hugo/tpl	1.180s
ok  	github.com/spf13/hugo/transform	0.009s
?   	github.com/spf13/hugo/utils	[no test files]
?   	github.com/spf13/hugo/watcher	[no test files]
```

whereas in Europe, e.g. Oslo, Norway, where the timezone is CET (+0100):

```
$ TZ=Europe/Oslo go test ./...
?   	github.com/spf13/hugo	[no test files]
ok  	github.com/spf13/hugo/bufferpool	0.005s
ok  	github.com/spf13/hugo/commands	0.030s
ok  	github.com/spf13/hugo/create	0.014s
ok  	github.com/spf13/hugo/helpers	3.142s
ok  	github.com/spf13/hugo/hugofs	0.009s
ok  	github.com/spf13/hugo/hugolib	9.984s
?   	github.com/spf13/hugo/livereload	[no test files]
ok  	github.com/spf13/hugo/parser	0.004s
ok  	github.com/spf13/hugo/source	0.011s
ok  	github.com/spf13/hugo/target	0.022s
ok  	github.com/spf13/hugo/tpl	1.002s
ok  	github.com/spf13/hugo/transform	0.009s
?   	github.com/spf13/hugo/utils	[no test files]
?   	github.com/spf13/hugo/watcher	[no test files]
```

ToTimeE() in github.com/spf13/cast uses time.Unix(), which converts a Unix epoch timestamp into a _local_ time.  I haven't figured out how to properly fix this test, and the best I could come out with is to remove the "day of week" and the "day" (so only test for year and month), or maybe to somehow force the timezone to UTC for the duration of these tests?

(If v0.18.1 is going to be based off `master` and would include commit bc060ee1c9779f47078871e662a0247c0082f818 with the latest spf13/cast, it would be nice to fix this test before the v0.18.1 regression fix release.)

Many thanks!

Cc: @bep, @moorereason

anthonyfok:
Fixes #2855

### DIFF
--- a/tpl/template_funcs_test.go
+++ b/tpl/template_funcs_test.go
@@ -2272,7 +2272,8 @@ func TestDateFormat(t *testing.T) {
 		{"Monday, Jan 2, 2006", "2015-01-21", "Wednesday, Jan 21, 2015"},
 		{"Monday, Jan 2, 2006", time.Date(2015, time.January, 21, 0, 0, 0, 0, time.UTC), "Wednesday, Jan 21, 2015"},
 		{"This isn't a date layout string", "2015-01-21", "This isn't a date layout string"},
-		{"Monday, Jan 2, 2006", 1421733600, "Tuesday, Jan 20, 2015"},
+		// The following test case gives either "Tuesday, Jan 20, 2015" or "Monday, Jan 19, 2015" depending on the local time zone
+		{"Monday, Jan 2, 2006", 1421733600, time.Unix(1421733600, 0).Format("Monday, Jan 2, 2006")},
 		{"Monday, Jan 2, 2006", 1421733600.123, false},
 		{time.RFC3339, time.Date(2016, time.March, 3, 4, 5, 0, 0, time.UTC), "2016-03-03T04:05:00Z"},
 		{time.RFC1123, time.Date(2016, time.March, 3, 4, 5, 0, 0, time.UTC), "Thu, 03 Mar 2016 04:05:00 UTC"},


bep:
@anthonyfok wouldn't it be easier if you moved to a compatible time zone?

moorereason:
@bep, the problem is that `dateFormat` will always return a string based on `time.Local` when converting Unix epochs (thanks to `time.Unix()`).  I scratched my head for a while trying different ideas, but there's not really anywhere to alter what `dateFormat` is doing.  Anthony's PR is a simple fix that should pass in any location.

bep:
@moorereason It was a joke! (reminder to self: Put smilies after jokes)

PR #2882, opened 2017-01-03
Add truncate template function

biilmann:
This commit adds a truncate template function for safely truncating text without
breaking words. The truncate function is HTML aware, so if the input text is a
template.HTML it will be truncated without leaving broken or unclosed HTML tags.

    {{ "this is a very long text" | truncate 10 " ..." }}
    {{ "With [Markdown](/markdown) inside." | markdownify | truncate 10 }}

The HTML truncation is based on Django's truncatehtml template helper, so while I would normally be really cautious about doing anything regexp based with HTML, this is a very battle tested implementation. I've also tested it on a corpus of 2600+ blog posts from a large publication.

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -664,12 +664,13 @@ e.g.
 
 ### truncate
 
-Truncate a text to a max length without cutting words or HTML tags in half. Since go templates are HTML aware, truncate will handle normal strings vs HTML strings intelligently.
+Truncate a text to a max length without cutting words or leaving unclosed HTML tags. Since Go templates are HTML-aware, truncate will handle normal strings vs HTML strings intelligently. It's important to note that if you have a raw string that contains HTML tags that you want treated as HTML, you will need to convert the string to HTML using the safeHTML template function before sending the value to truncate; otherwise, the HTML tags will be escaped by truncate.
 
-e.q.
+e.g.
 
-* `{{ "this is a text" | truncate 10 " ..." }}` → this is a ...
-* `{{ "With [Markdown](#markdown) inside." | markdownify | truncate 10 }}` → With &lt;a href='#markdown'&gt;Markdown …&lt;/a&gt;
+* `{{ "this is a text" | truncate 10 " ..." }}` → `this is a ...`
+* `{{ "<em>Keep my HTML</em>" | safeHTML | truncate 10 }}` → `<em>Keep my …</em>`
+* `{{ "With [Markdown](#markdown) inside." | markdownify | truncate 10 }}` → `With <a href='#markdown'>Markdown …</a>`  
 
 ### split
 
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -662,6 +662,15 @@ e.g.
 * `{{slicestr "BatMan" 3}}` → "Man"
 * `{{slicestr "BatMan" 0 3}}` → "Bat"
 
+### truncate
+
+Truncate a text to a max length without cutting words or HTML tags in half. Since go templates are HTML aware, truncate will handle normal strings vs HTML strings intelligently.
+
+e.q.
+
+* `{{ "this is a text" | truncate 10 " ..." }}` → this is a ...
+* `{{ "With [Markdown](#markdown) inside." | markdownify | truncate 10 }}` → With &lt;a href='#markdown'&gt;Markdown …&lt;/a&gt;
+
 ### split
 
 Split a string into substrings separated by a delimiter.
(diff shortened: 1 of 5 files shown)


tpl/template_funcs.go
```diff
@@ -239,6 +246,127 @@ func slicestr(a interface{}, startEnd ...interface{}) (string, error) {
...
+			}
+			ellipsis = template.HTML(html.EscapeString(s))
+		}
+	default:
+		return "", errors.New("too many arguments passed to truncate")
+	}
+	if err != nil {
+		return "", errors.New("text to truncate must be a string")
+	}
+	text, err = cast.ToStringE(textParam)
```
moorereason: Use `:=` and remove `var text string` above to delay the allocation.

tpl/template_funcs.go
```diff
@@ -239,6 +246,127 @@ func slicestr(a interface{}, startEnd ...interface{}) (string, error) {
...
+	if html, ok := textParam.(template.HTML); ok {
+		return truncateHTML(length, ellipsis, html)
+	}
+
+	if len(text) <= length {
+		return template.HTML(html.EscapeString(text)), nil
+	}
+
+	var count int
+	words := strings.Fields(text)
```
moorereason: I'd bet that something like [this](https://github.com/spf13/hugo/blob/19b6fdf2b1a03ca49d754810dd4d9d414e59ef1a/helpers/content.go#L469) would be more efficient.  Doesn't use `strings.Fields` that would generate more allocs.

biilmann:
Thanks @moorereason I've added both performance optimizations.

moorereason:
@bep,
Can you point out where it assumes each character is a byte?  Ranging on a string iterates the runes.

Can we stop creating `tpl/template_something.go` already?  We're in the `tpl` directory.  Feels like the file names are stuttering.  I'd rather see `tpl/func_truncate.go`.  I want to rename most of the files in the directory, but I'm afraid I'd break everyone's PRs. :grinning:

bep:
@moorereason it should be pretty obvious if you add those missing test cases. As to naming, I just follow a naming convention already there. Let us take that discussion somewhere else.

biilmann:
Spot on with the unicode comments. I've made some changes that fixes the issues with unicode slicing of the texts and that should handle languages with no spaces.

Take a look and let me know if you spot any other issues.

The two code paths between text and HTML truncation are much more similar now, and I'll want to refactor a bit to have just one path before this is ready to merge...

bep:
Yea, this looks more like a `truncate` func people would want to steal for their CMS project ... I will have a closer look later.

bep:
Here is a failing test case for you:

```go
{2, template.HTML("<p>P1</p><p>P2</p>"), nil, template.HTML("<p>P1 …</p>"), false},
```

biilmann:
Good catch - fixed that edge case and got rid of the duplication between the text and HTML truncation.

tpl/template_func_truncate.go
```diff
@@ -0,0 +1,156 @@
...
+	var lastWordIndex, lastNonSpace, currentLen, endTextPos, nextTag int
+
+	for i, r := range text {
+		if i < nextTag {
+			continue
+		}
+
+		if isHTML {
+			// Make sure we keep tag of HTML tags
+			slice := string(text[i:])
```
moorereason: `text` is a string now.

biilmann: Fixed

docs/content/templates/functions.md
```diff
@@ -662,6 +662,15 @@ e.g.
 * `{{slicestr "BatMan" 3}}` → "Man"
 * `{{slicestr "BatMan" 0 3}}` → "Bat"
 
+### truncate
+
+Truncate a text to a max length without cutting words or HTML tags in half. Since go templates are HTML aware, truncate will handle normal strings vs HTML strings intelligently.
```
moorereason: I would say:

Truncate a text to a max length without cutting words or leaving unclosed HTML tags.  Since Go templates are HTML-aware, `truncate` will handle normal strings vs HTML strings intelligently.  It's important to note that if you have a raw string that contains HTML tags that you want treated as HTML, you will need to convert the string to HTML using the `safeHTML` template function before sending the value to `truncate`; otherwise, the HTML tags will be escaped by `truncate`.

```
`{{ "<em>Keep my HTML</em>" | safeHTML | truncate 10 }}` → `<em>Keep my …</em>`
```

biilmann:
Rewrote the tag closing logic and updated the docs. Added a few more edge case test cases as well around tag closing.

PR #2910, opened 2017-01-07
WIP: hugolib: Fix .Site.LastChange setting

### LINKED ISSUE #2909: .Site.LastChange should be time of newest content
Currently, `.Site.LastChange` is set to the first page in the default sort order ([hugolib/site.go#L135](https://github.com/spf13/hugo/blob/2ac9817eae7ddbb84c9d17c998d3ed004acc4b95/hugolib/site.go#L1351)), which may not be the most recent content if weights were added to the front matter.

See https://discuss.gohugo.io/t/4992

moorereason:
If weights are added to the content front matter, the default sort order
can't be relied upon to set `.Site.LastChange`.

Fixes #2909

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -1341,8 +1341,7 @@ func readCollator(s *Site, results <-chan HandledResult, errs chan<- error) {
 func (s *Site) buildSiteMeta() (err error) {
 	defer s.timerStep("build Site meta")
 
-	ln := len(s.Pages)
-	if ln == 0 {
+	if len(s.Pages) == 0 {
 		return
 	}
 
@@ -1357,7 +1356,12 @@ func (s *Site) buildSiteMeta() (err error) {
 
 	s.assembleMenus()
 
-	s.Info.LastChange = s.Pages.ByLastmod()[ln-1].Lastmod
+	for _, p := range s.Pages.ByLastmod().Reverse() {
+		if p.shouldBuild() {
+			s.Info.LastChange = p.Lastmod
+			break
+		}
+	}
 
 	return
 }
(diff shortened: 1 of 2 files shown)


hugolib/site.go
```diff
@@ -1355,7 +1356,7 @@ func (s *Site) buildSiteMeta() (err error) {
 
 	s.assembleMenus()
 
-	s.Info.LastChange = s.Pages[0].Lastmod
+	s.Info.LastChange = s.Pages.ByLastmod()[ln-1].Lastmod
```
bep: 1) I suspect there may be other and conflicting date setting somewhere else
2) Sorting the pages to get the last date sounds ineffective
3) Test

moorereason: Where's "that other place"?

I've considered adding a simple check in `Page.update()` for `p.Lastmod.After(p.Site.LastChange)`, but that wouldn't work during watch mode due to partial rebuilds.

Once we can find an efficient way to do this, I can write tests.

bep: >Where's "that other place"?

I don't know. I can find it, but then I have to do some work.

bep: And please add a remark about "work in progress" or a question or something if it is a pull request that needs further work etc.

moorereason:
@bep,
I haven't had time to find a better solution, particularly after a partial rebuild.  I've wired in tests inside an existing test.  This PR is still a WIP in my opinion, but I wanted to push up what I have so you can take a look at it.

bep:
I'm mocking around with the disableKinds feature, and I will have that in mind.

PR #2916, opened 2017-01-09
Add pygmentize to the hugo snap

### LINKED ISSUE #2914: Snap binary doesn't work with os/exec?
See https://discuss.gohugo.io/t/solved-pygmentize-not-working-on-ubuntu-16-04/5063/2

/cc @anthonyfok @elopio

elopio:
closes #2914

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -20,3 +20,5 @@ parts:
     go-importpath: "github.com/spf13/hugo"
     build-packages:
       - git
+    stage-packages:
+      - python-pygments


elopio:
This is one way to solve it.
In general, snapcraft lets you handle your build process however you like, so we could also install from pip, or from pygmets source. Or as I told you, relax the isolation security to access other binaries from the system.

elopio:
Thanks @bep. If you have further problems with the snap, I'd be happy to help.

PR #2939, opened 2017-01-16
Sitemap: sitemap generation may not respect standards for <changefreq>, <priority>

magikstm:
Still needs some work to handle #.## and strings in sitemap.priority correctly.

I'm a novice in Go, so it may not respect standards fully and it's unfortunately incomplete at the moment.

I created tests and did my best to complete it, but I've had difficulties to handle #.## and strings correctly for sitemap.priority.

I think it should be a good start for someone to pick from and finish. I did as best as I could.

Tests should be complete.

(*) I didn't handle "loc" yet and I would probably suggest splitting handling that one in another issue and another fix as it doesn't seem at the same place in the code.

### DIFF
--- a/hugolib/sitemap.go
+++ b/hugolib/sitemap.go
@@ -16,6 +16,7 @@ package hugolib
 import (
 	"github.com/spf13/cast"
 	jww "github.com/spf13/jwalterweatherman"
+	"strings"
 )
 
 // Sitemap configures the sitemap to be generated.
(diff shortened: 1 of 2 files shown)


magikstm:
This line:
> priority := cast.ToFloat64(value)

Uses this: https://github.com/spf13/cast/blob/master/cast.go

I've found this to possibly handle strings:
http://stackoverflow.com/questions/36391664/go-convert-string-to-float32

I've also found this to possibly handle #.##:
http://stackoverflow.com/questions/24852317/how-to-count-decimal-places-of-float

I'm unsure what the best way according to projects standards to handle these two validations would be.

magikstm:
I think these 3 lines and their related lines may also need to be reviewed:

> viper.SetDefault("sitemap", Sitemap{Priority: -1, Filename: "sitemap.xml"})
(see: https://github.com/spf13/hugo/blob/master/hugolib/config.go)

> Keywords:    []string{}, Sitemap: Sitemap{Priority: -1},
(see: https://github.com/spf13/hugo/blob/master/hugolib/page.go)

> if page.Sitemap.Priority == -1 {
(see: https://github.com/spf13/hugo/blob/master/hugolib/site_render.go)

magikstm:
My last commits should make this complete. I think I may only need to squash my commits.

Please review it.

PR #2945, opened 2017-01-17
tpl: Fix deadlock in cached partials

### LINKED ISSUE #2935: partialCached slows down Hugo (much)
```
PS I:\site\website> hugo env
Hugo Static Site Generator v0.18.1 BuildDate: 2016-12-30T10:03:28+01:00
GOOS="windows"
GOARCH="amd64"
GOVERSION="go1.7.4"
```

In the footer of each page, I reference the following partial:

```
{{ partialCached "cached/footer-morearticles.html" . .RelPermalink }}
```

That partial contains:

```
<!-- List of recent articles -->
<ul>
  {{ if not (eq .Section "") }}

    {{ range (where .Site.Pages "Section" .Section) | shuffle | first 6 }}
      <li><a href="{{ .Permalink }}" title="{{ .Title }}">{{ .Title }}</a></li>
    {{ end }}

  {{ else }}

    {{ range first 6 .Site.Pages }}
      {{ if not (eq .Section "") }}
        <li><a href="{{ .Permalink }}" title="{{ if isset .Params "categories" }}{{ index .Params.categories 0}} - {{ end }}{{ .Title }}">{{ .Title }}</a></li>
      {{ end }}
    {{ end }}

  {{ end }}
</ul>
```

Now when I try to build Hugo (`hugo server --buildDrafts`), it takes a tremendous amount of time before Hugo builds. For instance, here I waited more than 5 minutes and Hugo still hadn't build:

```
PS I:\site\website> hugo server --buildDrafts | timestamp
 8:38:08 Started building sites ...
PS I:\site\website> timestamp
 8:44:21
```

Now if I don't use `partialCached` but instead a regular partial:

```
{{ partial "cached/footer-morearticles.html" . }}
```

And I don't change anything else in my theme, then the Hugo build time is less than 2 seconds:

```
PS I:\site\website> hugo server --buildDrafts | timestamp
 8:35:58 Started building sites ...
 8:36:00 Built site for language en:
 8:36:00 36 of 36 drafts rendered
 8:36:00 0 future content
 8:36:00 0 expired content
 8:36:00 402 regular pages created
 8:36:00 5 other pages created
 8:36:00 0 non-page files copied
 8:36:00 0 paginator pages created
 8:36:00 0 n1 created
 8:36:00 0 n2 created
 8:36:00 total in 1690 ms
 8:36:00 Watching for changes in I:\site\website\{data,content,layouts,static,themes}
 8:36:00 Serving pages from memory
 8:36:00 Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
 8:36:00 Press Ctrl+C to stop
```

I of course referenced the [`partialCached` documentation](https://gohugo.io/templates/partials/#cached-partials), but I'm not seeing anything in the way that I use `partialCached` to explain this behaviour.

----

To help troubleshoot, the issue also happens when I use:

```
{{ partialCached "cached/footer-morearticles.html" . }}
```

So it doesn't seem to be caused by using an unique cache key. 

That seems to suggest that one of the functions used in the partial don't work nicely with `partialCached`. I tried to rule out that myself, but:

- Removing `shuffle` from the partial had no effect (problem persisted).
- Removing `first 6` from partial had no effect.
- Removing the 'piping operation' (` | shuffle | first 6`) had no effect.
- Changing `.Site.Pages` to the new `.Site.RegularPages` had no effect.
- Replacing `(where .Site.RegularPages "Section" .Section)` with `.Site.RegularPages` to rule out `where` had no effect.

---------------

To rule out even more, on my pages I used this:

```
{{ partialCached "foot.html" . }}
```

Now when I change that to `partial` (instead of `partialCached`) and in `foot.html` use the following code:

```
{{ partialCached "cached/footer-morearticles.html" .Permalink }}
```

Then the same problem happens (Hugo remains stuck on 'Started building sites...'). So it doesn't seem to be caused by a cached partial that references a cached partial.

----------

If you have other ideas I can test, let me know.

moorereason:
Cached partials that contained cached partials would create a deadlock.

Fixes #2935

### DIFF
--- a/tpl/template_funcs.go
+++ b/tpl/template_funcs.go
@@ -1541,7 +1541,10 @@ func (tf *templateFuncster) Get(key, name string, context interface{}) (p templa
 
 	tf.cachedPartials.Lock()
 	if p, ok = tf.cachedPartials.p[key]; !ok {
+		tf.cachedPartials.Unlock()
 		p = tf.t.partial(name, context)
+
+		tf.cachedPartials.Lock()
 		tf.cachedPartials.p[key] = p
 	}
 	tf.cachedPartials.Unlock()


Jos512:
Thanks for fixing that issue!

PR #2978, opened 2017-01-22
hugolib: improve menu generation for section pages with content

### LINKED ISSUE #2974: Menu weight and linktitle should be inherited from the front-matter in section pages with content
The lazy blogger setting to automatically generate menu entries for section pages is quite useful. However, if one wants to influence order, or translate link names, it can't be used, and the entire menu has to be described in `config.toml`.

Now that we have section pages with content, in that case we could inherit the weight and linktitle from the front-matter of those section pages. This will allow to use the lazy blogger setting, and still influence order and translate links directly from the content page front-matter, which is a little bit more comfortable.

aroig:
When using the lazy blogger setting to automatically generate menu entries from section pages, we now recognize section pages that have content, and use the weight and linktitle configured in the frontmatter.

This way, we can use the lazy blogger automatic generation, and influence menu order and translations, directly from the frontmatter.

Fixes #2974

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -1460,7 +1460,7 @@ func (s *Site) assembleMenus() {
 	if sectionPagesMenu != "" {
 		// Create menu entries for section pages with content
 		for _, p := range pages {
-			if p.Section() != "" && p.Kind == KindSection {
+			if p.Kind == KindSection {
 				// menu with same id defined in config, let that one win
 				if _, ok := flat[twoD{sectionPagesMenu, p.Section()}]; ok {
 					continue
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -1461,15 +1461,16 @@ func (s *Site) assembleMenus() {
 		// Create menu entries for section pages with content
 		for _, p := range pages {
 			if p.Section() != "" && p.Kind == KindSection {
+				// menu with same id defined in config, let that one win
+				if _, ok := flat[twoD{sectionPagesMenu, p.Section()}]; ok {
+					continue
+				}
+
 				me := MenuEntry{Identifier: p.Section(),
 					Name:   p.LinkTitle(),
 					Weight: p.Weight,
 					URL:    p.RelPermalink()}
 
-				// menu with same id defined in config, let that one win
-				if _, ok := flat[twoD{sectionPagesMenu, me.KeyName()}]; ok {
-					continue
-				}
 				flat[twoD{sectionPagesMenu, me.KeyName()}] = &me
 			}
 		}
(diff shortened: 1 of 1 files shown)


hugolib/site.go
```diff
@@ -1431,27 +1432,47 @@ func (s *Site) assembleMenus() {
...
+		// Create menu entries for section pages with content
+		for _, p := range pages {
+			if p.Section() != "" && p.Kind == KindSection {
+				me := MenuEntry{Identifier: p.Section(),
+					Name:   p.LinkTitle(),
+					Weight: p.Weight,
+					URL:    p.RelPermalink()}
+
+				// menu with same id defined in config, let that one win
+				if _, ok := flat[twoD{sectionPagesMenu, me.KeyName()}]; ok {
```
bep: This conditional can be pulled above the me above.

hugolib/site.go
```diff
@@ -1431,27 +1432,47 @@ func (s *Site) assembleMenus() {
...
-	sectionPagesMenus := make(map[string]interface{})
-	//creating flat hash
 	pages := s.Pages
-	for _, p := range pages {
 
-		if sectionPagesMenu != "" {
+	if sectionPagesMenu != "" {
+		// Create menu entries for section pages with content
+		for _, p := range pages {
+			if p.Section() != "" && p.Kind == KindSection {
```
bep: Why check for p.Section()? Every page have a Kind -- and all section pages (with or without content) == KindSection.

hugolib/site.go
```diff
@@ -1431,27 +1432,47 @@ func (s *Site) assembleMenus() {
...
+
+				// menu with same id defined in config, let that one win
+				if _, ok := flat[twoD{sectionPagesMenu, me.KeyName()}]; ok {
+					continue
+				}
+				flat[twoD{sectionPagesMenu, me.KeyName()}] = &me
+			}
+		}
+
+		// Create entries for remaining content-less section pages
```
bep: Not sure what you mean by "content-less" section pages? The code below should be superfluous with the code above.

hugolib/site.go
aroig: The check for p.Section() was already there before I touched it. You do not want to synthesize a section menu entry for the empty string, and that would happen if there are pages that do not belong to a section.

hugolib/site.go
aroig: Done. I can squash all the commits if you wish, once we finish reviewing.

hugolib/site.go
aroig: To be honest... I forgot most of it.

This piece of code was kind of ugly. It ran through all pages, collecting a list of all sections they belong to, and from this collection of section, synthesized menu entries.

What I remember, is that I thought I could do better, by running through all section pages, and assembling the menu entry via some method in Page, using the actual page metadata. It didn't work, it turns out there was a reason for doing the old ugly thing. Probably content-less section pages are not yet there when the menu is assembled.

Then I ended up refining the old method, by using the section page if it is there, and doing the old thing of synthesizing a menu entry for mentioned sections (not actual section pages) otherwise.

Maybe you, that are more familiar with the code can fill in the blanks. Otherwise I can try squeeze my brain a little more and see if anything else comes out.

aroig:
I'll try to add some tests tomorrow. Thanks for looking into this!

hugolib/site.go
aroig: Oh, sorry, in this case you are right, on the second pass is when this test is supposed to be relevant.

bep:
I will finish this ins https://github.com/spf13/hugo/pull/3067

I'll need to do some general test cleanup in this department, and I might as well clean up after myself.

Thanks for starting this and bringing this to my attention.

aroig:
Thanks! I'll try to do some testing of master before the upcoming release.

PR #2984, opened 2017-01-23
tpl: Remove twitter:domain tag from internal shortcode

alanorth:
It seems this metadata tag is no longer used by Twitter, as it has been removed from their official [Cards Markup Tag Reference](https://dev.twitter.com/cards/markup). There is a reference to this on the [W3 mailing list in 2013 as well](https://lists.w3.org/Archives/Public/www-validator/2013Oct/0025.html).

Users should remove `.Site.Social.twitter_domain` from their configuration as well.


alanorth:
Rebased on top of current master. This is a trivial cleanup, hoping this makes it into 0.19.

PR #2989, opened 2017-01-26
Fix regression of .Truncated evaluation in manual summaries

socialhacker:
This fixes the behavior of .Truncated that was introduced with commit
bef496b97e1dc2df8c52b080152d539a58752801 which was later broken.  The
desired behavior is that .Truncated would evaluate to false when there
was nothing after the user defined summary marker.

This also adds a simple unit test to ensure that this feature isn't
broken again.  The check for content after the user defined summary
marker is done on the raw content instead of the working copy because
some of the markup renderers add elements after the marker, making it
difficult to determine if there is actually any content.

The behavior (evaluating to false when there is no content, just
summary) is also now documented.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -379,6 +379,23 @@ var (
 	internalSummaryDivider = []byte("HUGOMORE42")
 )
 
+// We have to replace the <!--more--> with something that survives all the
+// rendering engines.
+// TODO(bep) inline replace
+func (p *Page) replaceDivider(content []byte) []byte {
+	sections := bytes.Split(content, helpers.SummaryDivider)
+
+	// If the raw content has nothing but whitespace after the summary
+	// marker then the page shouldn't be marked as truncated.  This check
+	// is simplest against the raw content because different markup engines
+	// (rst and asciidoc in particular) add div and p elements after the
+	// summary marker.
+	p.Truncated = (len(sections) == 2 &&
+		len(bytes.Trim(sections[1], " \n\r")) > 0)
+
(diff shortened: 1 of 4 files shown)


hugolib/page.go
```diff
@@ -380,6 +380,17 @@ var (
 	internalSummaryDivider = []byte("HUGOMORE42")
 )
 
+// If the raw content has nothing but whitespace after the summary marker then
+// the page shouldn't be marked as truncated.  This check is simplest against
+// the raw content because different markup engines (rst and asciidoc in
+// particular) add div and p elements after the summary marker.
+func (p *Page) anyContentAfterUserDefinedSummary() bool {
+	sections := bytes.Split(p.rawContent, helpers.SummaryDivider)
```
bep: I care about performance, and I'm pretty sure we have done the above split before. So no need to do it again.

socialhacker: Performance is a good goal, and a nice feature of Hugo, and I'm happy to improve this.  This split is not actually done anywhere on the raw content, it is done on the work content once it has been modified by renderContent, but that is too late.  The same approximate amount of work is also done on the work content right before renderContent in commonConvert (after shortcode and emoji processing).  That is where the  is replaced by the internal marker.  It should be fine to do this check there instead of on the true raw content, so I'll look into changing how that replacement is done so that the truncate flag information can be computed then.

socialhacker:
Hi Bjørn, I've updated to only do the raw data split once, I've also included in this version the removal of some dead code, that might have been a mistake though.  Let me know if you would like me to make a separate pull request for the removal.

PR #3015, opened 2017-02-08
Adding instructions about using customized domain

yizhexu:
Just some simple documentation related change that I find to be useful - let me know if you prefer me to do this in a different way.


digitalcraftsman:
Hello @yizhexu,

thank you for submitting this addition.

Cheers,
Digitalcraftsman

PR #3022, opened 2017-02-08
Add ability to sort by frontmatter parameters

fj:
This allows users to sort by the value of a specified frontmatter key, including anything that would otherwise be available in `.Params`. Comparisons (and thus the resulting sorting) are done on a string basis, so, for example, `xyz123` comes before and not after `xyz23`.

This is useful for general-purpose sorting for which `weight` is either not sufficient or where it duplicates information that's already present in the frontmatter. See discussions [here](https://discuss.gohugo.io/t/can-a-sections-pages-be-sorted-by-arbitrary-frontmatter-variables/5358).

If the committers are happy with this implementation, I'll also add the correspondingly necessary documentation for this feature.

### DIFF
--- a/hugolib/pageSort.go
+++ b/hugolib/pageSort.go
@@ -278,7 +278,7 @@ func (p Pages) Reverse() Pages {
 	return pages
 }
 
-func (p Pages) ByParam(paramsKeyStr string) Pages {
+func (p Pages) ByParam(paramsKey interface{}) Pages {
 	paramsKeyStr := cast.ToString(paramsKey)
 	key := "pageSort.ByParam." + paramsKeyStr
 
--- a/hugolib/pageSort.go
+++ b/hugolib/pageSort.go
@@ -288,6 +288,13 @@ func (p Pages) ByParam(paramsKey interface{}) Pages {
 		s1 := cast.ToString(v1)
 		s2 := cast.ToString(v2)
 
+		// Sort nils last.
+		if s1 == "" {
+			return false
+		} else if s2 == "" {
+			return true
+		}
+
 		return s1 < s2
 	}
 
--- a/hugolib/pageSort.go
+++ b/hugolib/pageSort.go
@@ -14,6 +14,7 @@
 package hugolib
 
 import (
+	"fmt"
 	"sort"
 )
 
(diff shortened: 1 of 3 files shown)


hugolib/pageSort.go
```diff
@@ -275,3 +276,20 @@ func (p Pages) Reverse() Pages {
 
 	return pages
 }
+
+func (p Pages) ByParam(paramsKeyStr string) Pages {
+	key := "pageSort.ByParam." + paramsKeyStr
+
+	paramsKeyComparator := func(p1, p2 *Page) bool {
+		v1, _ := p1.Param(paramsKeyStr)
```
bep: You understand that `Param` will look in the site config if param is not present in the page? I thought a little about this, and I think it makes sense.

hugolib/pageSort.go
```diff
@@ -275,3 +276,20 @@ func (p Pages) Reverse() Pages {
...
+		v2, _ := p2.Param(paramsKeyStr)
+		s1 := fmt.Sprintf("%v", v1)
```
bep: Use `cast.ToString` to convert to string. Don't think it will make much practical difference in this case, but then we at least have one way of doing it.

hugolib/pageSort_test.go
```diff
@@ -113,6 +113,26 @@ func TestPageSortReverse(t *testing.T) {
 	assert.True(t, probablyEqualPages(p2, p1.Reverse()))
 }
 
+func TestPageSortByParam(t *testing.T) {
+	unsorted := createSortTestPages(10)
+	firstValue, _ := unsorted[0].Param("arbitrary")
```
bep: You're missing the test case where the param isn't set. I would say that those pages should come at the end.

hugolib/pageSort.go
```diff
@@ -275,3 +276,20 @@ func (p Pages) Reverse() Pages {
 
 	return pages
 }
+
+func (p Pages) ByParam(paramsKeyStr string) Pages {
```
bep: Make `paramsKeyStr` an `interface{}`-- it comes from the templates, and it *may* be something else.

hugolib/pageSort.go
fj: Yes, that was intentional. That way one can provide site-wide defaults. I'll make sure to document that behavior in the follow-up.

hugolib/pageSort_test.go
fj: That seems reasonable to me.

hugolib/pageSort.go
fj: Agreed.

hugolib/pageSort.go
fj: :+1:, working on this.

fj:
This should address the remarks you had earlier, @bep. Let me know how else I can help.

PR #3033, opened 2017-02-12
Use Page.Params more consistently when adding metadata

fj:
Right now, `.Params` is documented as being "everything you put into the frontmatter, except for a long list of specific variables". In particular, values which are otherwise subsumed by `Page` fields generally don't wind up in `.Params`. But this has exceptions -- for instance, `Description` is both a `Page` field and will be present in `Params`, even though the documentation says it _won't_ be present in `Params`.

The net result is that it's a little confusing and inconsistent about what does or does not wind up in `Params`.

This problem makes it difficult to use certain other Hugo functions in your templates. For example, one cannot do `{{ .Param "date" }}` because there is no `date` stored in `Params`, even though `date` is in your frontmatter. That means that if you invoke such values dynamically, e.g., `{{ .Param $k }}`, you can't be sure if it will work, even if `$k` is the name of a valid key in your frontmatter.

I think a more consistent and less surprising result is that `Params` should match your frontmatter as closely as possible. This PR attempts to fix this by making sure such assignments always happen.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -919,29 +919,37 @@ func (p *Page) update(f interface{}) error {
 		switch loki {
 		case "title":
 			p.Title = cast.ToString(v)
+			p.Params[loki] = p.Title
 		case "linktitle":
 			p.linkTitle = cast.ToString(v)
+			p.Params[loki] = p.linkTitle
 		case "description":
 			p.Description = cast.ToString(v)
-			p.Params["description"] = p.Description
+			p.Params[loki] = p.Description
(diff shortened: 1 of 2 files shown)


bep:
I agree that this is a good change, but it would be good to get the heads up from some others ... @digitalcraftsman @moorereason ?

fj:
Note: I've updated this branch to remove the conflicts that arose from the `Test.Parallel` enhancements. If there's anything else I need to do to make this qualify for 0.19 please let me know.

Issue #2598, opened 2016-10-17
Allow .Param to search nested maps


bep:
So for key `foo.bar`:

1) Check top-level for `foo.bar`
2) Traverse the path `foo`, `bar` etc.

PR #3034, opened 2017-02-13
Enhance `.Param` to permit arbitrarily nested parameter references

fj:
Currently, the only way to access params with nested keys is to be explicit and static:

    {{ .Params.author.favorites.color }}

This, for example, doesn't work:

    {{ $k := "author.favorites.color" }}
    {{ .Param $k }} // Param doesn't understand nesting.

This doesn't work either:

    {{ $k := slice "author" "favorites" "color" }}
    {{ index .Param $k... }} // Go templates don't understand variadic args.

What would be ideal is to be able to say something like:

    {{ $k := slice "author" "favorites" "color" }}
    {{ .Param $k }}

and have it traverse the `Page.Params` and `Site.Params` maps.

This PR introduces a bound function, `Page.Traverse`, that produces these results. It has behavior identical to `Page.Param`, but it will traverse a map to an arbitrary depth instead of to just the first level, and it accepts a `[]string` instead of a single `interface{}`. If a key is specified that does not exist, `nil` is returned instead; this is not considered a panic-worthy error.

If we'd rather that this behavior be subsumed into `.Param` directly in lieu of a separate function, I'm happy to do that and to amend the commits accordingly. Either way, if we believe this is useful, I will also be happy to document accordingly.

### DIFF
--- a/docs/content/templates/list.md
+++ b/docs/content/templates/list.md
@@ -250,7 +250,7 @@ The below example sorts a list of posts by their rating.
       <!-- ... -->
     {{ end }}
 
-If the frontmatter field of interest is nested beneath another field you can
+If the frontmatter field of interest is nested beneath another field, you can
 also get it:
 
     {{ range (.Date.Pages.ByParam "author.last_name") }}
--- a/docs/content/templates/list.md
+++ b/docs/content/templates/list.md
@@ -238,6 +238,7 @@ your list templates:
     {{ end }}
 
 ### Order by Parameter
+
 Order based on the specified frontmatter parameter. Pages without that
 parameter will use the site's `.Site.Params` default. If the parameter is not
 found at all in some entries, those entries will appear together at the end
@@ -246,6 +247,13 @@ of the ordering.
 The below example sorts a list of posts by their rating.
 
     {{ range (.Data.Pages.ByParam "rating") }}
+      <!-- ... -->
+    {{ end }}
+
+If the frontmatter field of interest is nested beneath another field you can
+also get it:
+
+    {{ range (.Date.Pages.ByParam "author.last_name") }}
       <!-- ... -->
     {{ end }}
 
--- a/docs/content/templates/variables.md
+++ b/docs/content/templates/variables.md
@@ -103,10 +103,55 @@ which would render
 **See also:** [Archetypes]({{% ref "content/archetypes.md" %}}) for consistency of `Params` across pieces of content.
 
 ### Param method
-In Hugo you can declare params both for the site and the individual page.  A common use case is to have a general value for the site and a more specific value for some of the pages (i.e. an image).
+
+In Hugo you can declare params both for the site and the individual page. A
+common use case is to have a general value for the site and a more specific
+value for some of the pages (i.e. a header image):
+
 ```
-$.Param "image"
+{{ $.Param "header_image" }}
 ```
+
+Thee `.Param` method provides a way to resolve a single value whether it's
+in a page parameter or a site parameter.
+
+When frontmatter contains nested fields, like:
+
+```
+---
+author:
+  given_name: John
+  family_name: Feminella
+  display_name: John Feminella
+---
+```
+
+then `.Param` can access them by concatenating the field names together with a
+dot:
+
+```
+{{ $.Param "author.display_name" }}
+```
+
+If your frontmatter contains a top-level key that is ambiguous with a nested
+key, as in the following case,
+
+```
+---
+favorites.flavor: vanilla
+favorites:
+  flavor: chocolate
+---
+```
+
+then the top-level key will be preferred. In the previous example, this
+
+```
+{{ $.Param "favorites.flavor" }}
+```
+
+will print `vanilla`, not `chocolate`.
+
 ### Taxonomy Terms Page Variables
 
 [Taxonomy Terms](/templates/terms/) pages are of the type `Page` and have the following additional variables. These are available in `layouts/_defaults/terms.html` for example.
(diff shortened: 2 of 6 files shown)


bep:
See https://github.com/spf13/hugo/issues/2598

I'm not sure adding another method makes sense here. The "dotted" syntax seems to be familiar for most people. 

Also, if you refactor this into that, please add "Fixes #2598" at the bottom of your commit message.

fj:
The problem with the "dotted" syntax is that you can't be dynamic -- you must know the key in advance. I was trying to illustrate that above with the example I gave in the description.

Maybe a more concrete example would help. Consider something like this:

```
# index.html
Upcoming Movies:
  {{ partial item-list.html (dict "resource" "movies" "key" "movie.release_date" "content" .) }}
Upcoming Events:
  {{ partial item-list.html (dict "resource "events" "key" "event.date_range" "content" .) }}

# item-list.html
{{ range where .content.Site.Pages "Section" .resource }}
  date: {{ .Params.[???] }}
{{ end }}
```

There's nothing one can put in the `[???]` that will accommodate both `events` and `movies`. But with this approach, you could instead do:

```
  {{ $k := split .key }}
  date: {{ .Param $k }}
```

I'll refactor this into `.Param`, as you've suggested, and will fix #2598 accordingly. Thanks!

fj:
Actually, @bep, apologies, I may have misunderstood you. Are you proposing an interface like this:

```
  {{ .Param "some.arbitrary.key" }}
```

instead of the `[]string` approach? If so, I considered that, but this would technically break the API. Right now, `.Param` would look for a key that was literally named `"some.arbitrary.key"`, as in:

```
---
some.arbitrary.key: example-value
---
```

I could do the "dotted" approach, if you prefer that. Otherwise, the `[]string` approach doesn't introduce any backwards-compatibility issues.

bep:
> This situation in your frontmatter, for example, would be ambiguous:

Sure, but so would this:

```
{{ .Params.author.favorites.color }}
```

Open up an issue about this on Go's issue tracker, and you will be scolded by Rob Pike (he is a legend, so I guess he deserves to be a little harsh on GitHub).

It that is really an issue (I think it is a constructed issue), it should be checked when the params get parsed.

But we're not throwing away the best and most obvious API because of it.

Here is what I suggest. 

In `Param`, for the string input:

1. Check it as is, if match, return that.
2. If not match, and if the input string contains a ".", then split the string and do a deep search.

fj:
This now implements the procedure described by @bep:

> In Param, for the string input:
> 
> Check it as is, if match, return that.
> If not match, and if the input string contains a ".", then split the string and do a deep search

and fixes #2598; it also adds documentation accordingly.

hugolib/page.go
```diff
@@ -330,13 +330,65 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
...
+}
+
+func traverse(keys []string, m map[string]interface{}) (interface{}, error) {
+	// Shift first element off.
+	firstKey, rest := keys[0], keys[1:]
+	result := m[firstKey]
+
+	// No point in continuing here.
+	if result == nil {
+		return result, fmt.Errorf("no such key: " + fmt.Sprintf("%v", firstKey))
```
bep: This isn't an error.

hugolib/page.go
```diff
@@ -330,13 +330,65 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
...
 		return nil, err
 	}
+
 	keyStr = strings.ToLower(keyStr)
+	result, _ := p.traverseDirect(keyStr)
+	if result != nil {
+		return result, nil
+	}
+
+	keySegments := strings.Split(keyStr, ".")
```
bep: if len(keySegments) == 1 here there is no need to look further.

fj: Resolved. Thanks for spotting this.

hugolib/page.go
```diff
@@ -330,13 +330,69 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
...
+	if len(keySegments) == 1 {
+		return nil, nil
+	} else {
```
bep: OK, this is nit picky, but I guess the Go linter will complain about this: You should drop the else here, as you return early above.

hugolib/page.go
```diff
@@ -330,13 +330,69 @@ func (p *Page) Param(key interface{}) (interface{}, error) {
...
+		result, _ = p.traverseNested(keySegments)
+		if result != nil {
+			return result, nil
+		}
+	}
+
+	return nil, nil
+}
```
moorereason: Don't ignore the error from `traverseNested()`.  The end of this func could be simplified to:

```go
if len(keySegments) == 1 {
    return nil, nil
}

return p.traverseNested(keySegments)
```

docs/content/templates/list.md
```diff
@@ -249,6 +250,13 @@ The below example sorts a list of posts by their rating.
       <!-- ... -->
     {{ end }}
 
+If the frontmatter field of interest is nested beneath another field you can
```
moorereason: Add a comma after "another field."

docs/content/templates/variables.md
```diff
@@ -103,10 +103,55 @@ which would render
...
+
+In Hugo you can declare params both for the site and the individual page. A
+common use case is to have a general value for the site and a more specific
+value for some of the pages (i.e. a header image):
+
+```
+{{ $.Param "header_image" }}
+```
+
+Thee `.Param` method provides a way to resolve a single value whether it's
```
moorereason: s/Thee/The/

hugolib/page.go
fj: Good catch, thanks.

hugolib/page.go
fj: Agreed -- changed this to include @moorereason's comment below as well.

fj:
I believe this should integrate the feedback from @bep and @moorereason.

bep:
This is solid, thanks for this, @fj

PR #3073, opened 2017-02-21
tpl: Use og:updated_time OpenGraph tag on nodes

alanorth:
Quick fix for a minor error I found in my earlier [commit improving the OpenGraph tags](https://github.com/spf13/hugo/pull/2979) generated by Hugo's internal opengraph template. The commit message there was correct, but the code left out one of the stated changes: nodes are of type "website" and according to the [Facebook docs](https://developers.facebook.com/docs/reference/opengraph/object-type/website/) they should use `og:updated_time`.


alanorth:
Sorry about that, @bep. Quick fix for a minor issue in #2979.

PR #3078, opened 2017-02-21
new go-version, switch to alpine (reduce size by 500mb)

hloeffler:
* use go 1.8
* switch to alpine for faster build and smaller image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,13 @@
-FROM golang:1.6.2
+FROM golang:1.8-alpine
 MAINTAINER Sven Dowideit <SvenDowideit@home.org.au>
 
 ENV GOPATH /go
 ENV USER root
 
+RUN apk update && apk add git
+
 # pre-install known dependencies before the source, so we don't redownload them whenever the source changes
+# TODO can this list be generated?
 RUN go get github.com/stretchr/testify/assert \
 	&& go get github.com/kyokomi/emoji \
 	&& go get github.com/bep/inflect \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,13 @@
...
+FROM golang:1.8-alpine
 MAINTAINER Sven Dowideit <SvenDowideit@home.org.au>
 
 ENV GOPATH /go
 ENV USER root
 
+RUN apk update && apk add git
+
 # pre-install known dependencies before the source, so we don't redownload them whenever the source changes
+# TODO can this list be generated?
```
digitalcraftsman: it should be possile with `go get ./...`

bep: Should use `govendor list`

moorereason: To expand on that, let `govendor` and `make` do the heavy lifting:

```Dockerfile
RUN go get github.com/kardianos/govendor \
 && govendor get github.com/spf13/hugo \
 && cd $GOPATH/github.com/spf13/hugo \
 && make install test
```

hloeffler: i want to use this Dockerfile for testing current code
(and to keep "make docker")
i'll add an commit
but maybe we want more Dockerfiles:
* one for clean build current github version
* one for fast build and test current $PWD version
* minimal image with latest stable huge version for becoming official dockerhub umage

hloeffler:
i think its not that easy :-)
what do we want this Dockerfile to be?
see discussion: https://github.com/spf13/hugo/pull/3078/commits/8165792467aa9e27a2b4f70e3e210ee51862a734

PR #3080, opened 2017-02-21
print error if --kind is used and cant find content type

hloeffler:
add error output  if `hugo new --kind` is used but can't find archetype for `--kind`
closes https://github.com/spf13/hugo/issues/3076

maybe even return error

### DIFF
--- a/create/content.go
+++ b/create/content.go
@@ -38,6 +38,10 @@ func NewContent(s *hugolib.Site, kind, name string) (err error) {
 
 	location := FindArchetype(s, kind)
 
+	if location == "" && kind != "" {
+		jww.ERROR.Println("Content Type", kind, "not found")
+	}
+
 	var by []byte
 
 	if location != "" {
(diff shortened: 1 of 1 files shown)


bep:
If we log an error, we should not continue. So, instead of logging an ERROR, return an error from the function.

hloeffler:
thanks,
i also think return error is better
i can do it later, but last time it broke tests

bep:
Not sure what the test failure is, but if you do a `rebase master` against the current Hugo, it should be fine.

hloeffler:
failed:
https://github.com/spf13/hugo/blob/4b3ca38b2e8aa07f6fecbba0cc150b1fab9227e3/create/content_test.go#L48
easy fix, remove this line :)
more work, extend check for return values

bep:
@hloeffler I suspect there has been too much changes in this area for this to merge cleanly, and it would probably be easier to create a new PR.

PR #3084, opened 2017-02-22
docs: Adds 'Using _index.md' page

### LINKED ISSUE #2827: docs: Improve docs for `_index.md` etc.
From what I understood in the new documentation and [the discussion forum](https://discuss.gohugo.io/t/hugo-0-18-ignores-index-md-in-the-section/4896/12) Hugo 0.18 supports rendering of `_index.md` files in sections, while `index.md` should not work. However, I am seeing varying behaviours, including the opposite one.  An example is provided via the [attached hugo018index.zip file](https://github.com/spf13/hugo/files/671347/hugo018index.zip).

Here is what is in the example:
```
hugo018index $ find . -type file
./bare/config.toml
./bare/content/_index.md
./bare/content/test/index.md
./bare/layouts/_default/single.html
./bare/layouts/index.html
./under/config.toml
./under/content/_index.md
./under/content/test/_index.md
./under/layouts/_default/single.html
./under/layouts/index.html

hugo018index $ diff -r bare under
Only in under/content/test: _index.md
Only in bare/content/test: index.md

hugo018index $ diff bare/content/test/index.md under/content/test/_index.md; echo $?
0

hugo018index $ 
```
In other words, the example has two directories with identical data, except in the `bare` directory the `test` section has `index.md`, while in the `under` directory the `test` section has identical `_index.md`.

Opposite my expectation things **do work** when running `hugo server` in the `bare` directory, and **do not work** when running `hugo server` in the `under` directory.

Are my expectations wrong, or there is a Hugo issue?

PuffinBlue:
There has been confusion regarding the use of _index.md pages following
the shift to 'Everything is a Page' in Hugo v0.18:

https://discuss.gohugo.io/t/hugo-0-18-ignores-index-md-in-the-section/4896

This has lead to attempts at clarification:

https://discuss.gohugo.io/t/-index-md-pages-in-section-taxonomies-etc-are-not-rendered-using-single-html-template/5466

There is also an open issue regarding improving the docs on this topic:

https://github.com/spf13/hugo/issues/2827

This commit adds a dedicated page explaining how _index.md files fit
into Hugo's template hierarchy, what they can contain and how to render
content from them onto the respective list/terms/home page.

What this commit doesn't do is address confusion over former behaviour
of the 'index.md' files (note the lack of _). Instead it focuses on
clarifying the new behaviour in the hope of rendering the old moot.

Fixes #2827

### DIFF
--- a/docs/content/content/using-index-md.md
+++ b/docs/content/content/using-index-md.md
@@ -15,7 +15,7 @@ weight: 70
 ---
 # \_index.md and 'Everything is a Page'
 
-As of version v0.18 Hugo now treats '[everything as a page](http://bepsays.com/en/2016/12/19/hugo-018/)'. This allows you to add content and frontmatter to any page - including List pages like [Sections](/content/sections/), [Taxonomies](/taxonomies/overview/), [Taxonomy Terms pages](/templates/terms/) and even to 'special case' pages like the [Home page](/templates/homepage/).
+As of version v0.18 Hugo now treats '[everything as a page](http://bepsays.com/en/2016/12/19/hugo-018/)'. This allows you to add content and frontmatter to any page - including List pages like [Sections](/content/sections/), [Taxonomies](/taxonomies/overview/), [Taxonomy Terms pages](/templates/terms/) and even to potential 'special case' pages like the [Home page](/templates/homepage/).
 
 In order to take advantage of this behaviour you need to do a few things. 
 
@@ -40,25 +40,31 @@ Now that you've got a handle on templates lets recap some Hugo basics to underst
 
 1. Sections and Taxonomies are 'List' pages, NOT single pages.
 2. List pages are rendered using the template heirarchy found in the [Content - List Template](http://localhost:1313/templates/list/) docs.
-3. The Home page is not technically a list page but has [it's own template](/templates/homepage/) at layouts/index.html.
+3. The Home page, though technically a List page, can have [it's own template](/templates/homepage/) at layouts/index.html rather than \_default/list.html. Many themes exploit this behaviour so you are likely to encounter this specific use case. 
 4. Taxonomy terms pages are 'lists of metadata' not lists of content, so [have their own templates](/templates/terms/). 
 
-Let's put this all together clearly because it's a source of much confusion:
+Let's put all this information together:
 
->**\_index.md files used in List pages, Terms pages or the Home page are NOT rendered as single pages or with Single Content templates, they are used to 'inject' frontmatter/content into the respective List pages, Terms pages or the Home page and those pages are rendered using templates listed above.**
+> **\_index.md files used in List pages, Terms pages or the Home page are NOT rendered as single pages or with Single Content templates.**
+
+> **All pages, including List pages, can have frontmatter and frontmatter can have markdown content - meaning \_index.md files are the way to _provide_ frontmatter and content to the respective List/Terms/Home page.**
 
 Here are a couple of examples to make it clearer...
 
-| \_index.md location               | Page affected             | Rendered by                   |
-| -------------------               | ------------              | -----------                   |
-| /content/post/\_index.md          | site.com/post/            | /layouts/section/post.html    |
-| /content/categories/hugo/index.md | site.com/categories/hugo/ | /layouts/taxonomy/hugo.html   |
+| \_index.md location                 | Page affected             | Rendered by                   |
+| -------------------                 | ------------              | -----------                   |
+| /content/post/\_index.md            | site.com/post/            | /layouts/section/post.html    |
+| /content/categories/hugo/\_index.md | site.com/categories/hugo/ | /layouts/taxonomy/hugo.html   |
+
+## Why \_index.md files are needed
+
+With a Single page such as a post it's possible to add the frontmatter and content directly into the .md page itself. With List/Terms/Home pages this is not possible so \_index.md files can be used to provide that frontmatter/content to them. 
 
 ## How to display content from \_index.md files
 
-From the information above it should follow that content within an \_index.md file won't be rendered in its own Single Page, it'll be 'injected' into the respective List/Terms/Home page. 
+From the information above it should follow that content within an \_index.md file won't be rendered in its own Single Page, instead it'll be made available to the respective List/Terms/Home page. 
 
-To **_actually display that content_** you need to ensure that the relevant template responsible for rendering the List/Terms/Home page contains (at least) `{{ .Content }}`. 
+To **_actually render that content_** you need to ensure that the relevant template responsible for rendering the List/Terms/Home page contains (at least) `{{ .Content }}`. 
 
 This is the way to actually display the content within the \_index.md file on the List/Terms/Home page. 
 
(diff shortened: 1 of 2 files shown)


docs/content/content/using-index-md.md
```diff
@@ -0,0 +1,113 @@
...
+- [Home page template](/templates/homepage/)
+- [Content List templates](/templates/list/)
+- [Single Content templates](/templates/content/)
+- [Taxonomy Terms templates](/templates/terms/)
+
+Now that you've got a handle on templates lets recap some Hugo basics to understand how to use an \_index.md file with a List page.
+
+1. Sections and Taxonomies are 'List' pages, NOT single pages.
+2. List pages are rendered using the template heirarchy found in the [Content - List Template](http://localhost:1313/templates/list/) docs.
+3. The Home page is not technically a list page but has [it's own template](/templates/homepage/) at layouts/index.html.
```
bep: I would argue that the home page is a *list page*.

My current definition of a list page is:

A page that:

* Has one or more child pages in `.Pages` or the alias `.Data.Pages`
* Will get its default template definition from `_default/list.html`

For the home page, it is true that you can override the list template by putting an `index.html` template in the root /layouts folder, but such a statement is also true for the other list pages.

docs/content/content/using-index-md.md
```diff
@@ -0,0 +1,113 @@
...
+
+>**\_index.md files used in List pages, Terms pages or the Home page are NOT rendered as single pages or with Single Content templates, they are used to 'inject' frontmatter/content into the respective List pages, Terms pages or the Home page and those pages are rendered using templates listed above.**
+
+Here are a couple of examples to make it clearer...
+
+| \_index.md location               | Page affected             | Rendered by                   |
+| -------------------               | ------------              | -----------                   |
+| /content/post/\_index.md          | site.com/post/            | /layouts/section/post.html    |
+| /content/categories/hugo/index.md | site.com/categories/hugo/ | /layouts/taxonomy/hugo.html   |
+
```
bep: Missing underscore in index in categories example?

docs/content/content/using-index-md.md
```diff
@@ -0,0 +1,113 @@
...
+1. Sections and Taxonomies are 'List' pages, NOT single pages.
+2. List pages are rendered using the template heirarchy found in the [Content - List Template](http://localhost:1313/templates/list/) docs.
+3. The Home page is not technically a list page but has [it's own template](/templates/homepage/) at layouts/index.html.
+4. Taxonomy terms pages are 'lists of metadata' not lists of content, so [have their own templates](/templates/terms/). 
+
+Let's put this all together clearly because it's a source of much confusion:
+
+>**\_index.md files used in List pages, Terms pages or the Home page are NOT rendered as single pages or with Single Content templates, they are used to 'inject' frontmatter/content into the respective List pages, Terms pages or the Home page and those pages are rendered using templates listed above.**
+
```
bep: I don't like the word "inject" -- as it implies that something "magic" is going on. 

**All pages, including list pages, can have frontmatter and frontmatter can have markdown content.**

docs/content/content/using-index-md.md
```diff
@@ -0,0 +1,113 @@
...
+| \_index.md location               | Page affected             | Rendered by                   |
+| -------------------               | ------------              | -----------                   |
+| /content/post/\_index.md          | site.com/post/            | /layouts/section/post.html    |
+| /content/categories/hugo/index.md | site.com/categories/hugo/ | /layouts/taxonomy/hugo.html   |
+
+## How to display content from \_index.md files
+
+From the information above it should follow that content within an \_index.md file won't be rendered in its own Single Page, it'll be 'injected' into the respective List/Terms/Home page. 
+
```
bep: Ditto about inject.

docs/content/content/using-index-md.md
```diff
@@ -0,0 +1,113 @@
...
+This is the way to actually display the content within the \_index.md file on the List/Terms/Home page. 
+
+A very simple/naive example of this would be:
+
+```html
+{{ partial "header.html" . }}
+	<main>
+        {{ .Content }}
+		{{ $paginator := .Paginate (where .Site.Pages "Type" "post") }}
+		{{ range $paginator.Pages }}
```
bep: To make this less confusing, I would change the above to:

```
{{ range .Paginator.Pages }}
```

Which is syntax that is portable to use in the default list template. The above looks like some home page syntax.

docs/content/content/using-index-md.md
PuffinBlue: Yes, I agree with you but I wanted to capture something of how it can be a 'special case' and inadvertently overlooked how all such pages could have a similar behaviour. I'll reword the doc to clarify.

docs/content/content/using-index-md.md
PuffinBlue: Ha! I don't like it either! 

I used it because (and I may misunderstand here) it seems if you add an _index.md file with frontmatter it automatically applies that frontmatter to the rendered output. Is this correct?

Of course it doesn't 'inject' content, you have to render that with {{ .Content }}. 

So what would be a good way of capturing both concepts do you think?

docs/content/content/using-index-md.md
PuffinBlue: Do you mean like this:

```
<main>
        {{ .Content }}
        {{ range .Paginator.Pages }}
		{{ partial "summary.html" . }}
	{{ end }}
	{{ partial "pagination.html" . }}
	</main>
```

I'm not very good with writing templates yet :-(

bep: Yes.

docs/content/content/using-index-md.md
PuffinBlue: I'm rewording and going with 'provide'.

bep: I don't know, but I think you just add to the confusion by trying to something "clearly", when it comes out as muddy. 

I think it is better to list:

1. Where the different pages get their content and frontmatter from.
2.  What templates the different pages will be renedered with.

docs/content/content/using-index-md.md
```diff
@@ -40,25 +40,31 @@ Now that you've got a handle on templates lets recap some Hugo basics to underst
...
-| \_index.md location               | Page affected             | Rendered by                   |
-| -------------------               | ------------              | -----------                   |
-| /content/post/\_index.md          | site.com/post/            | /layouts/section/post.html    |
-| /content/categories/hugo/index.md | site.com/categories/hugo/ | /layouts/taxonomy/hugo.html   |
+| \_index.md location                 | Page affected             | Rendered by                   |
+| -------------------                 | ------------              | -----------                   |
+| /content/post/\_index.md            | site.com/post/            | /layouts/section/post.html    |
+| /content/categories/hugo/\_index.md | site.com/categories/hugo/ | /layouts/taxonomy/hugo.html   |
+
+## Why \_index.md files are needed
```
bep: I think you should also remove this section, as it also implies "something magic" about these md files -- they are basically content files that must be named a special way so we can recognise them and so they do not interfere with other content files. The "need" should be obvious.

PuffinBlue: Could just change that to "Why index.md files are used"  and not 'needed' to remove any notion of 'specialness'?

The section above in this in the quote says what the _index.md files do, this section says *why* they are used to do it - i.e. you can't add frontmatter/content to the List pages without the use of them. 

I think it's important to keep the 'why' part in there as it's not immediately obvious to all people (me included) as some of the forums topics have shown so I'd argue it's important to spell things out as fully as possible. 

That said, it's your choice, I can remove it if that's the preference as it would be good to get the rest of the info up.

bep: OK, thats good -- we can change it later if we change our mind.

digitalcraftsman:
Let me know if this additions can be merged.

digitalcraftsman:
Merged as eaabecf5.

PR #3087, opened 2017-02-23
Press: Add blog post about Hugo CI

jacob-long:
Added article from gettingthingstech.com.

I also cleaned up some poorly-aligned entries near the top of markdown table (though it doesn't affect the HTML output either way).


digitalcraftsman:
Hello @jacob-long,

thank you for the clean up and the addition of the article. Merged as 950d1aa1.

Cheers,
Digitalcraftsman

PR #3096, opened 2017-02-24
Fix minor, but potentially confusing typo

kiley0:
"...your website has three actions..." should be "...your website has three sections..."


digitalcraftsman:
Hello @kiley0,

thank you for submitting this addition. Section makes more sense in this context.

Cheers,
Digitalcraftsman

PR #3106, opened 2017-02-27
Correct  commit link for 'now' template function.

vsopvsop:
When mentioning the commit for 'now' template function the previous line's github link is repeated unfortunately. Now you can fix it via this PR


digitalcraftsman:
Thank you for fixing this :+1:

terinjokes:
Would be great if this could also be fixed on the GitHub Releases page as well (which is where I view release notes)

moorereason:
@terinjokes, done.  :+1:

PR #3115, opened 2017-03-01
Taxonomies: Populate the KindTaxonomyTerm Pages list

socialhacker:
Previously this was left empty, but it is very handy to have a list
of term pages for a given taxonomy.  This list can now be paginated
like other page lists.  It makes it possible to render summary
content from each terms index page for instance.  It also makes it
possible to sort the term pages in the same way that other page
lists can be sorted.

### DIFF
--- a/docs/content/templates/rss.md
+++ b/docs/content/templates/rss.md
@@ -48,6 +48,14 @@ Hugo will use the following prioritized list. If a file isn’t present, then th
 * /themes/`THEME`/layouts/\_default/rss.xml
 * [Embedded rss.xml](#the-embedded-rss-xml:eceb479b7b3b2077408a2878a29e1320)
 
+### Taxonomy Terms RSS
+
+* /layouts/taxonomy/`SINGULAR`.terms.rss.xml
+* /layouts/\_default/rss.xml
+* /themes/`THEME`/layouts/taxonomy/`SINGULAR`.terms.rss.xml
+* /themes/`THEME`/layouts/\_default/rss.xml
+* [Embedded rss.xml](#the-embedded-rss-xml:eceb479b7b3b2077408a2878a29e1320)
+
 
 ## Configuring RSS
 
--- a/docs/content/templates/terms.md
+++ b/docs/content/templates/terms.md
@@ -57,6 +57,7 @@ Taxonomy Terms pages will additionally have:
 
 * **.Data.Singular** The singular name of the taxonomy
 * **.Data.Plural** The plural name of the taxonomy
+* **.Data.Pages** (or as **.Pages**) The taxonomy Terms index pages
 * **.Data.Terms** The taxonomy itself
 * **.Data.Terms.Alphabetical** The Terms alphabetized
 * **.Data.Terms.ByCount** The Terms ordered by popularity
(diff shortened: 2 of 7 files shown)


bep:
Thanks,

If you look at the recent commit log (just before the release), I added this functionality, but then reverted it. I will have to think about this before I commit to it. I'm not sure it is such a good idea as it appears to be.

socialhacker:
Ahh yes, I see that now, I should have looked harder or asked first.  :)  What was it that was problematic?  I would really like to be able to access the individual term pages from the list page.  I'd be happy losing the pagination capability, but I'm guessing that wasn't the major hangup.

bep:
I'm not totally sure. It was mainly a "this will bite me in the ass in the future" -- and when I started looking at the changes I forgot re. RSS feed etc., I started to question the value of having a page collection where none of the pages have anything in common (they "have a tag").

socialhacker:
Actually, looking over your old commit it appears that you were populating the Pages variable with all of the Pages that used the taxonomy.  This is slightly different from what I did.  My version populates Pages with the term index pages.  This behaves a bit more like a poor mans section hierarchy.  From a TaxonomyTerm page you can list the individual Taxonomy pages, which can each list the pages that are associated with that term.  So in this case the pages do have something in common, they are all the term pages for a taxonomy, instead of all of the pages that contain any term from the taxonomy.

bep:
OK, I will think about this a little. What you say makes sense.

socialhacker:
You bet, and thank you.

Do you know why the builders think hugolib/page.go is not correctly formatted?  There is no useful output in the logs that I can see and my systerm (go version 1.8) doesn't see anything wrong with the file.

bep:
>Do you know why the builders think hugolib/page.go is not correctly formatted

Run `go fmt` on that file. I use `LiteIDE` as my editor for Go, and it formats on save.

socialhacker:
Ahh, I see that I have been using "go fmt" incorrectly and it is silent when it has nothing to do, which I guess makes sense.  :)  I see the problem now.  Thanks.

socialhacker:
New version up, I believe I've addressed your comments.  I also fixed what looks like a bug in HugoSites.createMissingPages.

bep:
If there were a bug in createMissingPages, please create a separate issue and do not include any changes to that method in this PR.

socialhacker:
Will do.

socialhacker:
I've moved the fix for createMissingPages to a separate PR.  However since GitHub doesn't seem to have a way of specifying PR dependencies this PR contains both commits, I hope that makes sense.  I assume that once the fix has landed that this PR will just contain the top commit.  Thanks.

bep:
Could you rebase with master and force push so we get one commit here. I will have to look into the test race failure on Go 1.8. This does not fail locally for me, and it is a mystery how your change could have anything to do with this.

bep:
See #3129 for the failing test.

socialhacker:
Yeah, I noticed that test failure on the builder as well, but also couldn't reproduce it locally.  I've rebased this PR against master.

bep:
@socialhacker I have added a temporary fix in master, so it would be good if you could do a new rabase and see if this PR goes to green.

socialhacker:
Looks good.

Issue #3139, opened 2017-03-07
using apply with built-in functions


wildhaber:
## Introduction
This is a follow up report based on the topic from the discussion board (https://discuss.gohugo.io/t/apply-printf-on-a-sequence/5722) related using built-in functions with `apply`.

## Reproduction information

Assuming I have a list of numbers 1 to 30 (for example) and I want want to have them all as 5 digit length.

So for a single number I would use:

```go
{{ printf "%05d" 13 }} # 00013
```

Now I thought about using `apply` to range through my sequence list and apply this `printf` like:

```go
{{ apply (seq 30) "printf" "%05d" "." }}
# nope: at <apply (seq 30) "prin...>: error calling apply: can't find function printf
# cannot find function printf?  <- confusing
```

Next try - combining "printf" and its parameter `%05d`:

```go
{{ apply (seq 30) (printf "%05d") "." }}
# nope: at <apply (seq 30) (prin...>: error calling apply: can't find function %!d(MISSING)
# ok, so forget about this.
```

Another try providing the context (`.`) within the printf method bracket:

```go
{{ apply (seq 30) (printf "%05d" .) }}
# guess what!
# it breaks the page but outputs the right things a million times in the console including a bunch of errors :smiley:
#
# Console: 
# [] [] [] %!d(bool=0true) %!d(bool=false) %!d(string=00000) {63617444825 454000000 00000} {00000 00000 00000} {00702 00800 00004 00000} %!d(string=markdown) %!d(string=00000) %!d(string=article) %!d(bool=0true) %!d(string=00000) [%!d(string=article/single.h
tml) %!d(string=_default/single.html) %!d(string=theme/article/single.html) %!d(string=theme/_default/single.html)] %!d(string=00000) [00043 00043 00043 00013 00010 00100 00097 00116 00101 00032 00061 00032 00034 00050 00048 00049 00054 00045 00049 00050 00045 000
49 00054 00084 00048 00048 00058 00050 00055 00058 00048 00053 00046 00052 00053 00052 00090 00034 00013 00010 00112 00117 00098 00108 00105 00115 00104 00100 00097 00116 00101 00032 00061 00032 00034 00050 00048 00049 00054 00045 00049 00050 00045 00049 00054 000
[...]

PR #3141, opened 2017-03-07
tplimpl: Add built-in print funcs to FuncMap

moorereason:
Add print, printf, and println to the Hugo FuncMap so that they are
accessible to the apply template func.

Updates #3139


moorereason:
I didn't include any other built-in funcs because I don't know how relevant they are to the `apply` use case.  These print funcs may be enough to close the parent issue.

digitalcraftsman:
`printf` is already part of the template funcs.

moorereason:
I think you're missing point of this PR.

digitalcraftsman:
I should have dug deeper. Bjørn already stated the problem: 

> Yes. apply looks in the Hugo internal map for existing funcs (printf is a built-in)... which is limiting. But nothing I can do without fixing Hugo. Can you create an issue on GitHub?

Just forget what I wrote.

PR #3162, opened 2017-03-10
hugolib: Make tag path and permalink respect disablePathToLower

### LINKED ISSUE #1641: Misc case issues for URLs
I think there may be a regression for the URLs of uppercase tags.

In https://github.com/spf13/hugo/issues/491 urlize was changed to lowercase URLs. In https://github.com/spf13/hugo/issues/557 I think this may have been reverted?

Now when I added tag links to my post list page:

```
<a href="/tags/{{ . | urlize }}">#{{ . }}</a>
```

The links were broken as URLs come out as mixed case (contain uppercase letters). As a workaround, I can do this and it works.

```
<a href="/tags/{{ . | urlize | lower }}">#{{ . | lower }}</a>
```

I don't think most themes use `| lower`.

cc/ @bep @spf13 who participated in both issues.

xofyarg:
Use MakePathSanitized instead of MakePath to leverage the config
setting. Fix #1641.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -819,7 +819,7 @@ func (p *Page) createPermalink() (*url.URL, error) {
 		return url, nil
 	}
 
-	dir := strings.TrimSpace(p.s.PathSpec.MakePath(filepath.ToSlash(strings.ToLower(p.Source.Dir()))))
+	dir := strings.TrimSpace(p.s.PathSpec.MakePathSanitized(filepath.ToSlash(p.Source.Dir())))
 	pSlug := strings.TrimSpace(p.s.PathSpec.URLize(p.Slug))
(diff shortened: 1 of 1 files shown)


bep:
Hey, and thanks for the contribution. It looks very good, but I will wait with this until I'm finished with my current "multiple outputs" PR, because I'm doing some major rewrites in path handling.

bep:
@xofyarg this PR isn't mergeable as it is now. Can you confirm that this is still an issue with the latest Hugo?

xofyarg:
I think it's obsolete as it's already covered by a PR you write for path handling. But the problem seems still exist in HEAD: #3374.

Anyway, I'm closing this PR.

PR #3163, opened 2017-03-10
Editing Step 2 to include themes directory

theamazingcase:
I noticed when going through the Quickstart guide that the "themes" directory  is currently not included in the description that follows `tree -a` in **Step 2** even though it is autogenerated for you by `new site`.

I added some description of what the directory is for and included it inside the bash readout as well.

It might seem like a small thing, but I did get lost for a minute when trying `hugo server` for the first time and getting a warning message about a missing theme without having read anything about themes earlier in the guide.

### DIFF
--- a/docs/content/overview/quickstart.md
+++ b/docs/content/overview/quickstart.md
@@ -78,12 +78,13 @@ $ tree -a
 |-- content
 |-- data
 |-- layouts
-`-- static
+|-- static
+`-- themes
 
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Hello @theamazingcase,

thank you for submitting this addition. I've merged your contribution as 0e44f75e .

Cheers,
Digitalcraftsman

PR #3168, opened 2017-03-11
tpl: Add union template func

digitalcraftsman:


docs/content/templates/functions.md
```diff
@@ -214,6 +214,14 @@ e.g.
     </ul>
 
 
+### intersect
```
bep: Union.

docs/content/templates/functions.md
```diff
@@ -214,6 +214,14 @@ e.g.
...
+Given two arrays (or slices), this function will return a new set containing the elements out of both arrays. The elements supported are strings, integers and floats (only float64).
```
bep: That is not the best "union" description I've seen. What elements?

tpl/tplimpl/template_funcs_test.go
```diff
@@ -1021,6 +1023,52 @@ func TestIntersect(t *testing.T) {
...
+	for i, this := range []struct {
+		sequence1 interface{}
+		sequence2 interface{}
+		expect    interface{}
+	}{
+		{[]string{"a", "b", "c", "c"}, []string{"a", "b", "b"}, []string{"a", "b", "c"}},
+		{[]string{"a", "b"}, []string{"a", "b", "c"}, []string{"a", "b", "c"}},
+		{[]string{"a", "b", "c"}, []string{"d", "e"}, []string{"a", "b", "c", "d", "e"}},
+		{[]string{}, []string{}, []string{}},
+		{[]string{"a", "b"}, nil, make([]interface{}, 0)},
```
bep: This is a surprise; I would have expected a,b here.

tpl/tplimpl/template_funcs_test.go
```diff
@@ -1021,6 +1023,52 @@ func TestIntersect(t *testing.T) {
...
+		{nil, []string{"a", "b"}, make([]interface{}, 0)},
+		{nil, nil, make([]interface{}, 0)},
+		{[]string{"1", "2"}, []int{1, 2}, make([]string, 0)},
```
bep: I also think this is a surprising result. One could argue that the result should be either

* "1", "2", 1, 2 (which is impossible)
* "1", "2"
* 1, 2
* error

But not an empty string slice. Probably easiest to make it an error, not sure.

docs/content/templates/functions.md
digitalcraftsman: Fixed

docs/content/templates/functions.md
digitalcraftsman: I updated the description. What other general term would you use for an element of a set? Maybe  objects? The intersection function above refers to them as elements.

tpl/tplimpl/template_funcs_test.go
digitalcraftsman: The [intersection func](https://github.com/spf13/hugo/blob/master/tpl/tplimpl/template_funcs.go#L352-L353) also enforces that both array have to be of the same type. `[]string{}` ≠ `[]int{}`.

If there is some ambiguity in play we could just return the first or second slice but I prefer to be more explicit about this error so that users aren't confused at the end.

tpl/tplimpl/template_funcs_test.go
digitalcraftsman: As described below I enforced that both arrays have to be of the same type. But `nil` indicates the absence of a value and without context also a type.

Hence I would agree with you and would add a check for this case.

bep: I agree that nil == empty set in this context. But What is `{1,2,3} ∪ {}`? What you `expect` is the `intersect`, not the `union`.

tpl/tplimpl/template_funcs_test.go
bep: I agree about being "explicit", but this isn't explicit error handling. You should return and assert the error.

tpl/tplimpl/template_funcs_test.go
digitalcraftsman: You're right! I'll add check for the `nil` case.

digitalcraftsman: I've added the check and returning now the non-empty set of both.

docs/content/templates/functions.md
```diff
@@ -214,6 +214,23 @@ e.g.
     </ul>
 
 
+### union
+Given two arrays (or slices) A and B, this function will return a new array that contains the elements or objects that belong to either A or to B or to both. The elements supported are strings, integers and floats (only float64).
```
bep: This func was motivated by the OR/AND discussion re. the `where` func, but that would only work if if it also support `Pages` (or more general: Slices of structs and pointers).

I guess this func may be useful as is, but it was not what I envisioned.

digitalcraftsman: Good point. Then the intersection function should support "slices of structs and pointers" type as well.

bep: It is out of scope of this PR, but another PR with that would be super-useful.

digitalcraftsman: Can this PR merged as it is? I can open an issue for adding support for Pages.

PR #3175, opened 2017-03-13
docs: Press page remove dead links and YYYY-MM-DD date format

rdwatters:
This page was a real chore chore on the POC. In the event that a new doc site won't be approved by v20, this will at least clean up the broken links I found during my audit. I would like to eventually create this as a [data file](https://github.com/spf13/hugo/pull/2508#commitcomment-18705617) since md tables get unwieldy quickly.


anthonyfok:
@rdwatters, Thank you for your great work in cleaning up the list of Press articles!
Just one comment:

> make formatting consistent

Actually, the existing formatting **is** consistent, depending on one’s point of view.  The seemingly random spaces are there so that the column widths are consistent and everything would line up.  It has been like that since the beginning when @spf13 started it, supposedly because it is an automatic feature of [vim-spf13](http://vim.spf13.com/): whenever you edit a row, upon typing `|`, a vim plugin included in vim-spf13 automatically makes everything line up!

But of course, the problem here is that each line is like 300 characters long, and one would have to have a really large monitor, or make the font in the terminal really small, to see the full-width of the almost-perfectly-lined-up table.  :-p  (I am partially guilty of making the table that wide because I decided to add `<span lang="xx">` markups for foreign languages.)  And indeed, most people who have made pull requests to add links didn't know that, and I recall making several commits in the past to re-align those added rows.

So, which is more consistent?  Strictly one space before and after each `|` character?  Or padding with as many spaces as necessary to equal column widths?  I don't know either.

That said, I think re-implementing that page using data files is a marvellous idea!  :+1:

spf13:
This page predates data files. The switch to them is long overdue.

rdwatters:
> So, which is more consistent? 

Ah, I should have been clearer @anthonyfok.  I agree the parts you already mention are consistent. I meant the date formatting, etc, which is now easier to see at a glance with YYYY-MM-DD. The spacing between the rows in the table wasn't what I was referring to and didn't necessarily both me. I made it tighter in spacing mostly for OCD-related reasons... 

> This page predates data files. The switch to them is long overdue.

Agreed: https://github.com/rdwatters/hugo-docs-concept/issues/19

rdwatters:
FTR @anthonyfok [here it is in TOML](https://github.com/rdwatters/hugo-docs-concept/blob/master/data/articles.toml) in the event we can't get the new docs site up in time for v20 (or we can just wait 😄 ).

Issue #3042, opened 2017-02-16
Add template funcs namespaces


bep:
See https://github.com/spf13/hugo/issues/3040

I don't mind adding more template funcs, but we should have some kind of name spacing to make them easier to manage/document, i.e. `strings.trimPrefix` etc. Not sure how the "." works in this case, but then find something else ("_"?)

We should keep the "old" names as aliases, of course, but this will lead to more natural names like:

* `image.config`
* `image.resize`
* etc.

@moorereason @digitalcraftsman

robsonsobral:
"tempalte"? Typo on title.

thiagomgd:
I think it's way easier to have fewer functions, but with the namespaces to add more options. Thumbs up for this issue!

moorereason:
I made a proof of concept locally that allows `strings.TrimPrefix`.  The trick is to create a func that returns a struct which has the namespace funcs defined on it:

```go
func stringsNS() *stringsNamespace {
	// should probably use a global instead of a new instance
	return &stringsNamespace{}
}

type stringsNamespace struct{}

// TrimPrefix returns s without the provided leading prefix string. If s doesn't
// start with prefix, s is returned unchanged.
func (_ *stringsNamespace) TrimPrefix(prefix, s interface{}) (string, error) {
	p, err := cast.ToStringE(prefix)
	if err != nil {
		return "", err
	}

	ss, err := cast.ToStringE(s)
	if err != nil {
		return "", err
	}

	return strings.TrimPrefix(ss, p), nil
}

// funcMap := template.FuncMap{
//     "strings":  stringsNS,
```

Usage:

```
{{ strings.TrimPrefix "$" "$BatMan" }}
{{ "$BatMan" | strings.TrimPrefix "$" }}
```

So, we can accomplish what we want.  We just need to come up with a more formal proposal to redo the funcmap with namespaces.

digitalcraftsman:
Adding namespaces is a good idea if we add more functionality to the templates. The docs already loosely groups template functions by "namespaces" / usecases.

However, do you want to put every function in a namespace, even standard ones like `div` in `math.div`?

How do we handle this redundancy of template functions from a long-term perspective? 

Will we depecrate them (not immediately, but one day in the future) or do we continue with this double-barreled approach? Deprecating them would be affect nearly every Hugo site.

[...]

PR #3177, opened 2017-03-14
tpl: Add namespaced template functions

moorereason:
This commit adds a `tpl/strings` package that provides a set of standalone
and namespaced template functions.  It's not complete, but it
demonstrates how we might implement namespacing.

The standalone funcs are useful for replacing existing functions without
needing to instantiate a Namespace object.  Not sure I'm sold on this
approach.  Open to suggestions.

TODO:
- [x] `apply` should support a namespaced funcMap
- [x] Add `lang` namespace.
- [x] Add `data` namespace.


moorereason:
@bep, I didn't understand the usefulness of creating a TplFuncsNamespacer interface.  The approach I've taken is rather simple and easy to reason about.

Updates #3042.

bep:
@moorereason the namespace construct was the first thing that popped into my head, your approach looks bettter.

moorereason:
I've pushed up a proof of concept for your 1st point.  Moved `imageConfig` into its own namespace.  It needs access the underlying `deps.Fs`, so `images.NewNamespace` accepts that dependency as a param.

It looks like `markdownify` needs `deps.ContentSpec` and `deps.Cfg`, so I'll probably simplify this and just pass `deps` itself to the `NewNamespace` funcs to make it more consistent and less confusing.

To add the stateful namespace to the funcMap as `images`, the Namespace has a `Namespace` function that returns itself (at `template_funcs.go:2037`).

Also, I agree with your 2nd point: the more code we can get out of `tplimpl`, the better.  We'll just need to come up with a plan for each func and what namespace it should live in.

bep:
>The standalone funcs are useful for replacing existing functions without
needing to instantiate a Namespace object. Not sure I'm sold on this
approach. Open to suggestions.

They may be useful as a refactoring step, but in the end they are duplicates and in violation of DRY.

moorereason:
@bep,
Take a look now.  I've added two commits since your last comments (after rebasing with master).

moorereason:
I'm ready for this PR to be kicked around.  It moves just about everything out of `template_funcs.go`.

Much of this was just tedious copying and rewriting (and adding) of tests.  There's probably some more cleanup to do regarding tests in `template_funcs_test.go`, but we can do that later (I'm tired of looking at it right now).  I mainly want to get others to test the namespacing and see where I may have missed something.

One area that could use more eyes is `collections.Namespace.funcMap`.  It's a copy of the `tplimpl.templateFuncster.Tmpl.funcMap`.  The funcMap is needed by `collections.Apply`.  If there's a better way to do this, let me know.

bep:
Yea, that looks like much boring work, much appreciated ... Will look at it later today.

bep:
The packages look good, but I have one fundamental comment (which I think I also have mentioned before). 

One of the motivations was to clean up the "blob" of code in `tplimpl`, which we mostly achieve with this, but there is still this long definition list: https://github.com/moorereason/hugo/blob/59dbb5e4f1dcff3666bc391a536ddf0fb9b791f8/tpl/tplimpl/template_funcs.go#L76

With a very strong coupling between `tplimpl` and the internals of all the different packages.

* If I want to do something in the `math` package (adding a func, changing an alias), I want to make the change in that package, not also in some other package.
* If I want to add a new package with some new `funcs`, I expect to add one more package import somewhere in tplimpl, but that is it -- I don't want to add it to n other places.

(also, nice to have: If I want to write a `docshelper.go` that creates a JSON with all the funcs and aliases, I want to be able to do that with ease).

moorereason:
@bep,

* If you want to add a method to an existing namespace, you can do that strictly within the namespace package.  This PR achieves that desire.  Note the namespaces are at the top of the funcMap.

* Adding a new namespace currently involves a little more work:  1) you need the namespace in the funcMap (no way to avoid this); 2) you need to [add an instance of the namespace](https://github.com/moorereason/hugo/blob/59dbb5e4f1dcff3666bc391a536ddf0fb9b791f8/tpl/tplimpl/templateFuncster.go#L43-L55) to `templateFuncster`; and 3) [initialize the namespace](https://github.com/moorereason/hugo/blob/59dbb5e4f1dcff3666bc391a536ddf0fb9b791f8/tpl/tplimpl/templateFuncster.go#L65-L77) in `newTemplateFuncster`.

I was trying to be consistent as I built out this PR, but I can see now that not all of the namespaces need the `deps` context and, therefore, don't need an instance tied to `templateFuncster`.  So, for a simple namespace, we should be able to remove steps 2 & 3.  I'll work on that.  :+1:

moorereason:
@bep, I've pushed a commit that demos how we could decouple simple namespaces from templateFuncster.  The major downside is that we now alloc a new namespace struct with each func call.  I don't think I like this approach, but I wanted you to see it.

I wrestled with several approaches over the last two months before landing on the existing framework, but I'm open to more ideas.

bep:
>I wrestled with several approaches over the last two months before landing on the existing framework, but I'm open to more ideas.

If you want to go where I want, you need one global `TemplatetFunsMapFactoryRegistry` (I come from a Java background) or something -- which is a list of `func (d *deps.Deps) FuncMap` (or something)  ... Which the different namespace packages adds to in a init func. Something like that.

moorereason:
> If you want to go where I want, ...

I understand what you want, but I don't know how to get there from here.  At least, not yet.  Every attempt I've made has been overly complex or unworkable (similar to my experience with Java ;-p).  What I have here should work from the user's perspective.  I'd prefer that we merge the existing framework and refactor the underlying mechanics of the funcMap building in a separate PR.

bep:
Yes, that is fine. That would be a small change compared to this.

bep:
I have browsed through the changes, and all in all, it looks very good.

A couple of comments:

* The `NewNameSpace()` sounds stuttery; there is only one thing the package creates, so `New()` would be better.
* I see no use (or potential use) of the `NameSpace()` method.
* I'm pretty sure that `Namespace` is a more common term.
* Of the packages below, I agree to most of them, but:
* `security` sounds a little ambitious: I suggest `safe`, so it is `safe.HTML` etc.
* `transform` also sounds off, but I cannot think of a better suggestion now ...

```
+	collections *collections.Namespace
 +	crypto      *crypto.Namespace
 +	encoding    *encoding.Namespace
 +	images      *images.Namespace
 +	inflect     *inflect.Namespace
 +	math        *math.Namespace
 +	os          *os.Namespace
 +	security    *security.Namespace
 +	strings     *hstrings.Namespace
 +	time        *time.Namespace
 +	transform   *transform.Namespace
 +	urls        *urls.Namespace
```

moorereason:
> * The `NewNameSpace()` sounds stuttery; there is only one thing the package creates, so `New()` would be better.

Done.

> * I see no use (or potential use) of the `NameSpace()` method.

The text/template funcMap values must be functions or I get a panic during fuzzing tests.  Also, the method is `Namespace()` (lower "s").

> * I'm pretty sure that `Namespace` is a more common term.

I'm not sure what you're referring to here.

> * Of the packages below, I agree to most of them, but:
>   * `security` sounds a little ambitious: I suggest `safe`, so it is `safe.HTML` etc.
>   * `transform` also sounds off, but I cannot think of a better suggestion now ...

Renamed `security` to `safe`.  More discussion about naming [in the forums](https://discuss.gohugo.io/t/plan-for-namespacing-template-funcs/5800).  I've tried to keep my original post up-to-date.

I've also submitted a new commit to add `safe.SanitizeURL` (which just calls `helpers.SanitizeURL`) and cleaned up redundant tests in `template_funcs_test.go`.

bep:
I see a TODO item in the first post. Is this ready for a merge?

moorereason:
Push a new commit to add a `data` namespace.  Should be ready for a merge.

bep:
It looks good, and I would just merge it -- but some of the commit messages look a little bit temporary, and I would hate to just squash it ...?

moorereason:
I noticed that.  I'm find with squashing it to clean up the commit history, which is pretty messy.  I had intended to squash sooner in the process but forgot to.

bep:
But please squash it yourself, so I would not have to "guess" a proper commit message.

moorereason:
Squashed.

PR #3180, opened 2017-03-14
[WIP] org: use Emacs as org mode interpreter

chaseadamsio:
Goorgeous was a good first run at the basic use cases for Org mode but there are two major problems with trying to keep a go parser (similar to asciidoc and rst):

1. Org mode itself is a moving target, varying from major versions and adding new functionality which would mean constantly trying to stay up to date with it's features for power users.
2. Different users are different versions of Org mode (9 was recently released and it has many difference from 8), which means trying to make everyone happy for their org version.

This PR leverages emacs as a subprocess the same way rst and asciidoc are used with one exception:

- `org-mode` uses its "frontmatter" to make decisions about how to render content, so rather than reading the frontmatter out and leaving the "content", the entire file's contents are left intact and a peek function gets the frontmatter to be passed along to the org parser.

https://github.com/chaseadamsio/goorgeous/issues/32#issuecomment-286092742

# Advantages

- Now users can use the Emacs Org version they feel comfortable with and can know that what they experience in Emacs for org-html-as-html is going to be what they get in Hugo.

# Disadvantages

- A new dependency on Emacs is introduced, which means the CI environment (or whatever environment is used to render and publish)

## Solution

- Docker with Emacs v 25 and diff. major versions of org-mode.

Rollup of changes:

- use Emacs as the org interpreter
- move org headers parsing from chaseadamsio/goorgeous to spf13/hugo
- remove goorgeous as dependency altogether

cc'ing the following users who've opened issues in goorgeous to give any feedback or concerns they have. @strow @yssource @priyadarshan @jackbaty @kaushalmodi 

@strow's workflow is to have `*.org` files in his `content` directory which means Hugo will overwrite his handrolled files. In the case that users want to ignore using Emacs org mode in Hugo, do we want to add a config option to [ignore invocation as suggested here](https://github.com/chaseadamsio/goorgeous/issues/32#issuecomment-285964518)? My gut reaction is to [suggest the workflow I recommended here](https://github.com/chaseadamsio/goorgeous/issues/32#issuecomment-286031146), but if we want to have an option to ignore rendering org files, I'm happy to add that.  

# Todo

- Update docs to reflect that Emacs is a requirement where rst docs exists.

### DIFF
--- a/docs/content/content/supported-formats.md
+++ b/docs/content/content/supported-formats.md
@@ -13,7 +13,7 @@ weight: 15
 toc: true
 ---
 
-  Since 0.14, Hugo has defined a new concept called _external helpers_. It means that you can write your content using Asciidoc[tor], reStructuredText or Org-Mode. If you have files with associated extensions ([details](https://github.com/spf13/hugo/blob/77c60a3440806067109347d04eb5368b65ea0fe8/helpers/general.go#L65)), then Hugo will call external commands to generate the content (the exception being Org-Mode content, which is parsed natively).
+  Since 0.14, Hugo has defined a new concept called _external helpers_. It means that you can write your content using Asciidoc[tor], reStructuredText or Emacs org-mode. If you have files with associated extensions ([details](https://github.com/spf13/hugo/blob/77c60a3440806067109347d04eb5368b65ea0fe8/helpers/general.go#L65)), then Hugo will call external commands to generate the content.
 
   This means that you will have to install the associated tool on your machine to be able to use those formats.
 
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -24,7 +24,6 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	"github.com/chaseadamsio/goorgeous"
 	"github.com/miekg/mmark"
 	"github.com/mitchellh/mapstructure"
 	"github.com/russross/blackfriday"
@@ -416,7 +415,7 @@ func (c ContentSpec) RenderBytes(ctx *RenderingContext) []byte {
 	case "rst":
 		return getRstContent(ctx)
 	case "org":
-		return orgRender(ctx, c)
+		return getOrgContent(ctx)
 	}
 }
 
@@ -666,9 +665,44 @@ func getRstContent(ctx *RenderingContext) []byte {
 	return result[bodyStart+7 : bodyEnd]
 }
 
-func orgRender(ctx *RenderingContext, c ContentSpec) []byte {
+func getEmacsExecPath() string {
+	path, err := exec.LookPath("emacs")
+	if err != nil {
+		return ""
+	}
+	return path
+}
+
+// HasEmacs returns whether Emacs is installed on this computer.
+func HasEmacs() bool {
+	return getEmacsExecPath() != ""
+}
+
+// getOrgContent calls Emacs as an external helper
+// to convert Org content to HTML.
+func getOrgContent(ctx *RenderingContext) []byte {
 	content := ctx.Content
-	cleanContent := bytes.Replace(content, []byte("# more"), []byte(""), 1)
-	return goorgeous.Org(cleanContent,
-		c.getHTMLRenderer(blackfriday.HTML_TOC, ctx))
+	cleanContent := bytes.Replace(content, SummaryDivider, []byte(""), 1)
+
+	path := getEmacsExecPath()
+	if path == "" {
+		jww.ERROR.Println("Emacs not found in $PATH: Please install.\n",
+			"                 Leaving Org content unrendered.")
+		return content
+	}
+
+	jww.INFO.Println("Rendering", ctx.DocumentName, "with", path, "...")
+	cmd := exec.Command(path, "--no-site-file", "--batch", "--eval",
+		"(let ((org-document-content \"\") this-read) (while (setq this-read (ignore-errors (read-from-minibuffer \"\"))) (setq org-document-content (concat org-document-content \"\n\" this-read))) (with-temp-buffer (org-mode) (insert org-document-content) (org-html-export-as-html nil nil nil t nil) (princ (buffer-string))))")
+
+	cmd.Stdin = bytes.NewReader(cleanContent)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
+
+	if err != nil {
+		jww.ERROR.Printf("%s rendering %s: %v", path, ctx.DocumentName, err)
+	}
+
+	return normalizeExternalHelperLineFeeds(out.Bytes())
 }
--- a/parser/frontmatter.go
+++ b/parser/frontmatter.go
@@ -14,13 +14,15 @@
 package parser
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
+	"regexp"
 	"strings"
 
-	"github.com/chaseadamsio/goorgeous"
 	toml "github.com/pelletier/go-toml"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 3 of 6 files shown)


bep:
My take on this is:

1) If this Pull Request was submitted before we started any talk about Org-mode support, I would not have merged it.
2) So, given 1), in my head, the question is whether to remove Org-mode support entirely or not. Since we only had it for one version, the breakage in the wild should be minimal.

chaseadamsio:
> If this Pull Request was submitted before we started any talk about Org-mode support, I would not have merged it.

Can you elaborate on why?

We can remove org support if that's the only option and I'll maintain my own fork with this solution.

priyadarshan:
A non-profit I have been helping is considering switching to Hugo, because of the recently introduced support for org-syntax.

They curate a pretty large site offering several thousands books, all using org-syntax.

Org-syntax is, sometimes, the go-to syntax for projects where markdown is not enough. Org-syntax is, in a way, like YAML, only with exceptional support on the editor's side (Emacs).

I hope Hugo will keep org-syntax support, no matter in which way.

Using native Emacs would be of course ideal, but I am afraid it would take a serious toll on building a site.

Right now, the site takes a few hours to render, while using gnu `parallel` (to help run more Emacs instances at once)
If there is a way to keep Emacs in memory, without invoking an instance for each file to be built, perhaps even an instance per core, that would make native Emacs feasible.

Please, keep supporting org-syntax.

Hugo is such a good tool, it would be a pity to miss on such major text syntax.

chaseadamsio:
Source for emacs lisp: http://joelmccracken.github.io/entries/reading-writing-data-in-emacs-batch-via-stdin-stdout/

Implementation is from @joelmccracken's jekyll implementation.

bep:
>@bep Can you elaborate on why? (I'm mainly curious why it's a viable solution for rst & asciidoc but not for org-mode)

I didn't merge any of those, either. But if we, for the argument's sake, say I did:

Both of these are clearly marked as experimental and a temporary workaround until a native Go implemenation happens. Not sure about RST, but for Asciidoc it is just a matter of time.

What you are saying is the opposite. We're dropping the permanent and native solution in favor of the "temporary permanent" variant.

I know people testing out their 1000 pages Asciidoc site gets disappointed. This may bee sen in relation to #2651  -- but as it is no, it is a no-go in my head.

/cc @spf13

joelmccracken:
Glad to hear this is still helping.

I actually replaced the code in that post because it was too slow; basically it fired up an emacs instance to recompile each org file all the time. You might not have the problem if hugo is smart enough to only regenerate a few files at a time though.

If you're interested in what I replaced it with:

https://github.com/joelmccracken/joelmccracken.github.com/blob/writing/_lib/org-convert.el
https://github.com/joelmccracken/joelmccracken.github.com/blob/writing/_plugins/org-converter.rb

It starts up an emacs process that acts as a TCP server with a dumb protocol. It receives the org mode file content, and returns the compiled output.

bep:
@chaseadamsio just to add the obvious: I totally sympathise with your reasons for not wanting to continue working on/maintaining your Org-mode Go project.

chaseadamsio:
@bep in that care what if we wrap current org support as an experimental config and I'll get it as close as possible to parity? 

I want to support org mode natively but it's a totally different beast from markdown in terms of what it does. 

I'm good with whatever you decide and would be glad to support that decision, but I definitely think not supporting org in the long run misses a lot of opportunities.

bep:
I will have to think about this ...

chaseadamsio:
No problem. I think the usability requirements should be:

- if you want to use `org-mode` with native go through hugo, add a value to your `config.toml` to opt in
- otherwise the old workflow for how people used org before just works the way it works.

I can create a branch for the config value check with goorgeous as well so it's there at least and I'll continue to work towards parity with goorgeous until you make a decision.

strow:
I am the long-time user of org-mode with hugo, and the new native org-mode converter completely broke my workflow as googeous is not fully compliant with orgmode.  But, I am just a user, not a contributor here.  I find it so so easy to use emacs to produce org-->html that hugo converts (ever so quickly) to my webpages.  When I am editing an org file, I just hit (Ctrl-e Ctrl-b h h) and the .html is created.  The absolute beauty of orgmode is that the .org source can also be used to produce beautiful latex-->pdf output.  (The issues are almost always to do with figures.)  

My (strong) suggestion, so I can continue to use hugo upgrades, is for you to provide a config switch to turn off the native orgmode converter.  If you continue to have the orgmode converter, it would also be nice to use it at times.  If hugo only converts .org files when a .html file of the same prefix did NOT exist, my workflow would still be intact, even if I turned off the proposed switch to disable the org converter.

I suspect using emacs as an external converter will cause a flurry of problems from people who haven't written proper org files, don't know emacs well enough, etc. etc.

Anyway, I love hugo and love orgmode.  I married them two years ago, and I have really nice web pages full of figures.  I also just started using fancybox too, very very easy.  So, thanks for all your work on hugo!

bep:
@strow @chaseadamsio ok, so the core of the problem is that Emacsers are putting *.org files into /content and use Emacs to produce HTML? Why not put those files into **some other folder** as they're 

1) Not intended to be processed by Hugo
2) Not intended to part of the end result, i.e. the static content

yssource:
No matter which way, PLEASE, PLEASE, PLEASE, keep supporting org-syntax.

strow:
Putting the *.org files elsewhere is a real big pain since I then have to write code in the org setup on where to put every single file I edit (using their "publish" functionality, as far as I can tell from org).  I'm not using that fancy publishing functionality of org, I'm just doing a very simple .org to .html export.  If I move the .org elsewhere, then every single time I edit it I have to move the .html.   What I am doing is so so simple.   I don't think a hugo upgrade should break something that has worked fine for two years.

I may not understand the full details of how org can handle files, but it sure will take a lot of configuration to change, and I have *lots* of files that would need to be "fixed".

bep:
@strow put your (other) ORG files in /static where they belong -- no changes to your work flow. /content is reserved for HUGO content. Adding a ORG handler that handles files in that directory is hardly a breaking change.

bep:
To add to the above:

* We may add some way to turn off / configure suffix etc. for the handlers (including this one)
* But we will not replace the native org-mode handler with a "emacs exec" if the above is the motivation. I think most people would be happy to have something that supported "the bulk of the Org-mode" spec.

chaseadamsio:
@bep 👍 new PR to come for the config knob and continue to add features to goorgeous.

bep:
That is great; and understand that (most) people understand that it takes time to build software on the spare time, and don't expect it to be feature complete (or bug free) right away.

strow:
Bep, you misunderstood me.  I do not think the "emacs exec" is a good idea, so I agree with you.

But, I also think it is a bad idea to break hugo when it worked fine for me for two years.  So, please please add the turn off m/ configure suffix to hugo in the next release.

It's a cardinal rule to not break something that worked fine for years!!

bep:
>It's a cardinal rule to not break something that worked fine for years!!

OK, then we agree on one point and disagree on another. That is fine. That is open source.

chaseadamsio:
@strow I'm going to add the lever for overriding and then you'll be back to your normal workflow, then everyone wins!

strow:
Thank you, and I do indeed appreciate open source, a lot!  I am also very willing to test .org to .html converter whenever requested.

priyadarshan:
As far as I know, The need to support "legacy" features would not apply to open source projects in `0.x` version, like Hugo. They are supposed to be free to change API's, if not even encouraged to do so where proper, so progress in solidity and features can be achieved at a quicker pace, and in a more integral way.

joelmccracken:
It's worth looking at the second FAQ item on http://semver.org/

PR #3182, opened 2017-03-14
commands: Support human-readable YAML boolean values in undraft

kropp:
Simplified version of #2831

### DIFF
--- a/commands/undraft.go
+++ b/commands/undraft.go
@@ -25,8 +25,8 @@ import (
 
 var undraftCmd = &cobra.Command{
 	Use:   "undraft path/to/content",
-	Short: "Undraft changes the content's draft status from 'True' to 'False'",
-	Long: `Undraft changes the content's draft status from 'True' to 'False'
+	Short: "Undraft resets the content's draft status",
+	Long: `Undraft resets the content's draft status
 and updates the date to the current date and time.
 If the content's draft status is 'False', nothing is done.`,
 	RunE: Undraft,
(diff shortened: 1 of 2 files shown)


bogem:
Sorry for the stupid question, but how does it work, if it replaces nothing.

kropp:
@bogem there is a loop over all frontmatter entries. Before this change, when `draft` key was found, it's value was changed from `true` to `false`. After this change this entry is skipped and not written to resulting frontmatter. Absent `draft` key means that the entry is not draft.

bogem:
Ok, I understood.
When the docs should be also edited, because initially `undraft` command sets undraft field from true to false. Now you just delete. This should be noticed.

bep:
How is `yes` interpreted by `go-yaml`?

kropp:
@bogem thanks, nice catch! I've updated my change.

kropp:
@bep all values according to [spec](http://yaml.org/type/bool.html) are supported: https://github.com/go-yaml/yaml/blob/v2/resolve.go#L37-L42

bep:
@kropp that was not my question.

kropp:
@bep `yes` is interpreted as `true`

bogem:
Could you delete please the last line of help (`If the content's draft status is 'False', nothing is done.`). Your change removes the `draft` property despite the value.

kropp:
@bogem No, `undraft` command does the check first and breaks if the given page is not draft (see https://github.com/kropp/hugo/blob/bf8aa82a5b638e88ba9529a26c7b81dfbda2e350/commands/undraft.go#L119-L122).

The documentation is therefore correct.

bogem:
Ah ok, sorry for my inattention. 
Then it looks good to me.
What do you think, @bep?

PR #3191, opened 2017-03-17
docs: Make install command compatible with Fish, Zsh and Fish

### LINKED ISSUE #3190: Installing script on gohugo.io doesn't work with Fishshell
Hello guys,

First of all, thanks for the amazing work on Hugo !

The double ampersand is not supported by Fishshell

thylong:
"&&" is not supported by Bash but ";" is universally accepted.
The behavior of semicolon is a little bit different than double
ampersand but in a case of a package manager update then install,
it makes sense.

Fixes #3190

### DIFF
--- a/docs/content/overview/installing.md
+++ b/docs/content/overview/installing.md
@@ -37,7 +37,7 @@ Ideally, you should install it somewhere in your `PATH` for easy use.
 `/usr/local/bin` is the most probable location.
 
 On macOS, if you have [Homebrew](http://brew.sh/), installation is even
-easier: just run `brew update && brew install hugo`.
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Hello @thylong,

first of all thank you for making this submission. Even if I understand the intention behind the PR the reactions seem to be mixed based on the votes on the comment above.

@moorereason could you elaborate why you downvoted this PR?

bep:
I voted up without thinking. 

The headline does not match the description and "&&" is perfectly fine in Bash -- and has a totally different meaning than ";". I'm closing this.

PR #3192, opened 2017-03-17
Fix documentation of non-existent 'editor' config

callahad:
Looks like this was erroneously added two years ago in commit 1827680beaf10c0710b3daed2cf1f6679c9a070a. I suspect the author misinterpreted the `--editor` [command line option](https://github.com/spf13/hugo/blob/1827680beaf10c0710b3daed2cf1f6679c9a070a/commands/hugo.go#L92) which [overrides the `newContentEditor` configuration option](https://github.com/spf13/hugo/blob/1827680beaf10c0710b3daed2cf1f6679c9a070a/commands/hugo.go#L173-L175).

### DIFF
--- a/docs/content/overview/configuration.md
+++ b/docs/content/overview/configuration.md
@@ -134,8 +134,6 @@ along with their current, default values:
     disable404:                 false
     # Do not inject generator meta tag on homepage
     disableHugoGeneratorInject: false
-    # edit new content with this editor, if provided
-    editor:                     ""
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Hello @callahad, 

thank you for submitting this addition. `editor` is indeed not a correct [config option](https://github.com/spf13/hugo/blob/4b3ca38b2e8aa07f6fecbba0cc150b1fab9227e3/create/content.go#L80).

PR #3199, opened 2017-03-19
WIP hugolib: Add template analysis feature

moorereason:
This commit adds a templateAnalysis option to turn on tracking of
execution stats for referenced templates.

### DIFF
--- a/tpl/tplimpl/template.go
+++ b/tpl/tplimpl/template.go
@@ -186,17 +186,12 @@ func (t *GoHTMLTemplate) executeTemplate(context interface{}, w io.Writer, layou
 		}
 
 		if templ != nil {
-			var start time.Time
 			if t.Cfg.GetBool("templateAnalysis") {
-				start = time.Now()
+				defer t.AddTemplateTiming(layout, time.Now())
 			}
 
 			if err := templ.Execute(w, context); err != nil {
 				helpers.DistinctErrorLog.Println(layout, err)
-			}
-
-			if t.Cfg.GetBool("templateAnalysis") {
-				t.AddTemplateTiming(layout, time.Now().Sub(start))
 			}
 
 			worked = true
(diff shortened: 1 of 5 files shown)


moorereason:
This is a proof of concept.  Assuming it can be improved upon.  I'm not sure how to move the timings map up higher into the build state so that we can track the parent templates that are executed in `hugolib`.

tpl/tplimpl/template.go
```diff
@@ -62,6 +63,9 @@ type GoHTMLTemplate struct {
 	amberFuncMap template.FuncMap
 
 	*deps.Deps
+
+	sync.RWMutex
```
bep: As you never use the read lock, a regular Mutex is a better choice.

moorereason: True.  I was using RLock in an earlier iteration.

moorereason:
Refactored to push the timings up into `deps.Deps`.  This may be an abuse of the Deps concept, but I'm now able to track all template executions.

bep:
It is a very good idea, and, with one important addition, would have saved me lots of debugging time the other day (I ended up adding something similar temporarily).

But you are right, it does not belong in `Deps`.

I suggest that this is important enough for its own package ( `metrics`?) and a simpler to use abstraction (`Timer.Time(func(), keys ...string)`?).

Also add a `Count` to the metrics (invocation count per key(s)).

The above probably means an adjustment of how the template methods are used (must consolidate the lookup and execute), but that seems worth doing.

I also suggest that the `GoHTMLTemplate` owns its own metrics, and also provides the metric report on demand (maybe implementing an optional interface with a `PrintMetrics(io.Writer)` method.

bep:
Note: I'm about to simplify the Template interface to be able to do both text and html template, so it might be worth it to wait a little.

moorereason:
Replaced by #3812

PR #3208, opened 2017-03-21
commands: Ignore "." dirs in hugo --cleanDestinationDir

bogem:
Fixes #3202

PR #3210, opened 2017-03-21
commands: Added new option `cleanDestinationDirKeepDotFolder`

eamonnmcevoy:
Add a new option to ignore 'dot' folders when deleting files in destination dir.

Fixes #3208

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -213,6 +213,7 @@ func initRootPersistentFlags() {
 // Called by initHugoBuilderFlags.
 func initHugoBuildCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool("cleanDestinationDir", false, "Remove files from destination not found in static directories")
+	cmd.Flags().Bool("cleanDestinationDirKeepDot", false, "Remove files from destination not found in static directories, but keep dot folders")
 	cmd.Flags().BoolP("buildDrafts", "D", false, "include content marked as draft")
 	cmd.Flags().BoolP("buildFuture", "F", false, "include content with publishdate in the future")
 	cmd.Flags().BoolP("buildExpired", "E", false, "include expired content")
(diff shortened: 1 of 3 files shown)


eamonnmcevoy:
relates to: https://github.com/spf13/hugo/issues/3202

commands/hugo.go
```diff
@@ -578,9 +580,16 @@ func (c *commandeer) copyStatic() error {
...
 	// We can effectively clean the publishDir on initial sync
-	syncer.Delete = c.Cfg.GetBool("cleanDestinationDir")
+	syncer.Delete = c.Cfg.GetBool("cleanDestinationDir") || c.Cfg.GetBool("cleanDestinationDirKeepDotFolders")
+
 	if syncer.Delete {
 		c.Logger.INFO.Println("removing all files from destination that don't exist in static dirs")
+
+		if c.Cfg.GetBool("cleanDestinationDirKeepDotFolders") {
+			syncer.DeleteFilter = func(f os.FileInfo) bool {
+				return f.IsDir() && len(f.Name()) > 0 && string(f.Name()[0]) == "."
```
bep: Use strings.HasPrefix(f.Name(), ".")

docs/content/commands/hugo_benchmark.md
```diff
@@ -21,37 +21,38 @@ hugo benchmark
 ### Options
 
 ```
```
bep: All doc below /commands is autogenerated so no need to edit it by hand.

commands/hugo.go
```diff
@@ -578,9 +580,16 @@ func (c *commandeer) copyStatic() error {
 	syncer.DestFs = c.Fs.Destination
 	// Now that we are using a unionFs for the static directories
 	// We can effectively clean the publishDir on initial sync
-	syncer.Delete = c.Cfg.GetBool("cleanDestinationDir")
+	syncer.Delete = c.Cfg.GetBool("cleanDestinationDir") || c.Cfg.GetBool("cleanDestinationDirKeepDotFolders")
```
bep: We have enough flag. 

I would say that having a cleanDestinationDir command that also deleted hidden folders ("." files) was a surprise for many. So I suggest we just make this the new default and make sure to document it. People who want to remove the entire folder can use `rm -rf`.

eamonnmcevoy: No problem, I'm not sure where to update the documentation, do I just update `commands/hugo.go`?

ln:215:
`cmd.Flags().Bool("cleanDestinationDir", false, "Remove files from destination not found in static directories. Folders beginning with "." are ignored")`

PR #3213, opened 2017-03-22
docs: Expand the description of "after"

mdhender:
Make the last part of the description clearer by adding
to the example.

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -155,14 +155,17 @@ e.g.
 
 ### after
 Slices an array to only the items after the <em>N</em>th item. Use this in combination
-with `first` to use both halves of an array split at item _N_.
+with `first` to use both (not necessarily equal) halves of an array split at item _N_.
 
 Works on [lists](/templates/list/), [taxonomies](/taxonomies/displaying/), [terms](/templates/terms/), [groups](/templates/list/)
 
 e.g.
 
-    {{ range after 10 .Data.Pages }}
-        {{ .Render "title" }}
+    {{ range after 2 .Data.Pages }}
+        {{ .Render "caesars" }}
+    {{ end }}
+    {{ range first 2 .Data.Pages }}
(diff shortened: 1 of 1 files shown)


docs/content/templates/functions.md
```diff
@@ -155,14 +155,17 @@ e.g.
...
-with `first` to use both halves of an array split at item _N_.
+with `first` to use both (not necessarily equal) halves of an array split at item _N_.
 
 Works on [lists](/templates/list/), [taxonomies](/taxonomies/displaying/), [terms](/templates/terms/), [groups](/templates/list/)
 
 e.g.
 
-    {{ range after 10 .Data.Pages }}
-        {{ .Render "title" }}
+    {{ range after 2 .Data.Pages }}
```
moorereason: I don't know if adding an example of `first` is necessary here.  Also, I think it may help the illustration if we use a simple slice in the example instead of `.Data.Pages`.

```
{{ $data := slice "one" "two" "three" "four" }}

{{ range after 2 $data }} → ["three", "four"]
{{ . }}
{{ end }}
```

mdhender: Just trying to make that last part clear. Agree that using the slice helps with the example.

rdwatters:
Thanks @moorereason && @mdhender. You both inspired me :smile:

https://hugodocs.info/functions/after/

Maybe I should add more to connect it w/ @bep 's `seq` function example from the forums?

https://discuss.gohugo.io/t/how-can-i-start-a-sequence-with-1/5879/6?u=rdwatters

digitalcraftsman:
Is this PR still part of a discussion or ready to be merged?

rdwatters:
Let @moorereason decide.  

Thanks @digitalcraftsman

digitalcraftsman:
> Let @moorereason decide.

@moorereason ?

bep:
This is superfluous with the new docs site.

PR #3219, opened 2017-03-23
RSS description tag should contain hugo .summary

DarwinJS:


tpl/tplimpl/template_embedded.go
```diff
@@ -84,7 +84,7 @@ func (t *GoHTMLTemplate) EmbedTemplates() {
       <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
       {{ with .Site.Author.email }}<author>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</author>{{end}}
       <guid>{{ .Permalink }}</guid>
-      <description>{{ .Content | html }}</description>
+      <description>{{ .Summary | html }}</description>
```
rdwatters: @DarwinJS  Seems the description page variable with a default summary value makes more sense since it provides the user more control.

```
{{ .Description | default .Summary }}
```

bep: 1) The description is something completely different and would be a pretty breaking change.
2) This PR is itself a breaking change and, while agree it is a good thing, we should not just merge it without thought.

DarwinJS: @rdwatters - that is a nuance of Hugo that I am not aware of.  I would go with whatever is considered best by PR reviewers.

All I know is that RSS feeds don't generally include entire articles in a feed, but rather a summary the norm.

rdwatters: @DarwinJS Sure thing. I think the point of confusion is the terms we are using, and I apologize for any confusion.

> **tl;dr
>  1. Hugo `.Summary` == "lead"
>  2. Hugo `.Description` == user-defined summary of content (e.g., synopsis or meta description)
>  3. `.Content` is part of Atom and not RSS

**Explanation**

The `.Summary` in Hugo is really a ["lead" in publishing](https://en.wikipedia.org/wiki/Lead_paragraph). So, a first paragraph, or a specific word count, *could* be a summary of the article intended to draw the reader in, but I think this assumes too much from authors, at least with respect to creating a Rich Site Summary. Of course, the feature is pretty darn handy, especially with the ability for users to define their own via `<!--more-->`. The summaries Hugo creates automatically are really *excerpts* or *extracts*. This point has been called out more specifically in the [new Hugo docs](https://hugodocs.info/content-management/summaries/#user-defined-manual-summary-splitting) in hopes of making this clearer. (Please provide feedback if it doesn't! :smile:) 

A *description* is a more focused and deliberate paragraph of text that does in fact summarize the full content of the page, although&mdash;I'm guessing&mdash;the way it was originally intended in Hugo was in the case of meta descriptions (usually <160 characters)...which also makes sense.  

As far as the RSS `<description>` tag containing the full-length `.Content`, this is a case of confusing RSS with [Atom](https://en.wikipedia.org/wiki/RSS#RSS_compared_to_Atom), which allows for a *optional* summary *or* full content, whereas RSS requires it. Here is an excellent [resource from Harvard on the RSS spec](http://cyber.harvard.edu/rss/rss.html#hrelementsOfLtitemgt). NB, `<description>` in RSS can either mean "channel description" or "item synopsis." 

So that's my reasoning behind the `.Description => .Summary` since it follows a model of `deliberate usage => sane default`. 

In case you are asking about the `default` syntax specifically, and excuse me if you know this, but I love this function and feel like it's underused: you can see the [docs on it here](https://hugodocs.info/functions/default-function/) as well.

HTH.

DarwinJS: @rdwatters - that's a rocking explanation - I'm going to update my custom rss template!

DarwinJS: @rdwatters - question regarding best implementation on my setup - should I run a custom template for mailchimp only and leave the system default as is?

bep: @digitalcraftsman you decide this.

rdwatters: @DarwinJS I'm at a loss on that one man, haha. I don't actually use MailChimp, but FWIW, I don't think it should be up to mailchimp to redefine the spec. Let's see what @digitalcraftsman says about it and go from there. BTW, thanks for submitting this PR; I hadn't ever looked at the built-in template.

DarwinJS: @rdwatters - I guess I'm function under the assumption MailChimp has picked the broadest compatibility option since they are integrated to by many services?  Not sure if you had a chance to actually look at their page?

Thanks for all the help.

digitalcraftsman: Even if the Atom spec allows the usage of the full content I think the original intention of RSS is to deliver a sneak-peak for content I subscribed to. Hence the usage of `.Summary` as replacement makes sense. Many themes just show the summary on the homepage as well.

bep: Agree. But I think @rdwatters is right about the Description | Summary. I'm not sure how many uses that actively. But I think we'll start with this one.

irl: This change has broken blog aggregators that expect the full content to be available in the feed. See for example any of the feeds syndicated by:

* http://planet.debian.org/
* http://planet.fsfe.org/
* http://planet.57north.org.uk/
* http://fedoraplanet.org/

I could give more examples.

irl: For anyone stumbling across this from Google and looking for a workaround, I have published a fix on my blog at https://iain.learmonth.me/blog/2017/2017w396/

Issue #3205, opened 2017-03-20
Steve's vision for Hugo


digitalcraftsman:
This issue acts as a reminder for the presentation that @spf13 put together to share his thoughts about ["The Future of Hugo"](https://drive.google.com/file/d/0B-s2yFMZO3LeejNNeXVuRVZIUHc/view). 

At the time of writing certain things mentioned like i18n support and "make everything a node"-transition are already built-in to Hugo.

vassudanagunta:
@digitalcraftsman when did @spf13 put together that presentation? Has his vision been updated since? Has anyone else presented their thoughts on a coherent vision (as opposed to grab-bag of pet features) since?

digitalcraftsman:
Steve created this presentation to communicate his vision on a private channel to the other maintainers. I don't know the exact date when he shared this document but it was months before I created this issue.

> Has his vision been updated since?

No changes have been made since then. Some parts are can be ignored with the latest developments (i.e. the whole nodes vs. pages distinction). Others still have to finds it's way into Hugo.

> Has anyone else presented their thoughts on a coherent vision (as opposed to grab-bag of pet features) since?

Not in this kind of presentation. @bep as lead developer will finally decide on which (new) features will be in the focus. He has a "mental roadmap" of improvements and major features that he would like to see implemented before making a v1.0 release.

AFAIK he hasn't written down what the missing pieces are to complete the puzzle. But the milestones and currently active issues should give you a good picture of where the development is heading towards.

Nonetheless, it would be still interesting to have some kind of outline of his vision.

jhabdas:
I'm particularly interested in the concept of multi-part/multilingual metadata inheritance (Ref: [Slide 16–17](https://drive.google.com/file/d/0B-s2yFMZO3LeejNNeXVuRVZIUHc/view)) as from what I can tell right now there's quite a bit of duplication in my bundles. Perhaps this is something which could be added with the introduction of multi-part pages, if that feature is still on the table, as I saw Steve mention meta inheritance as a feedback point in the multilingual implementation but met some resistance where I feel the original i18n implementor may not have considered the general use case.

jmooring:
Outdated, not actionable.

PR #3231, opened 2017-03-24
extras/shortcodes: Add Reuse Page Shortcode concept

savsheniuk:
Add a Reuse Page shortcode that allows tech writers to reuse same pages in different places without need in copy-pasting them every time.
It also adds a concept of Page snippets and dynamic links resolution.

### DIFF
--- a/docs/content/extras/shortcodes.md
+++ b/docs/content/extras/shortcodes.md
@@ -66,6 +66,44 @@ The `%` characters indicates that the shortcode's inner content needs further pr
 The `<` character indicates that the shortcode's inner content doesn't need any further rendering, this will typically be pure HTML:
 
     {{</* myshortcode */>}}<p>Hello <strong>World!</strong></p>{{</* /myshortcode */>}}
+    
+### Page Reuse shortcodes
+
+The `~` character indiciates that the shortcode's inner content must be taken from other referred page. For example, 
+this will take the rendered content (including shortcodes/images/links etc.) of the page "mypage.md" located by path "/path/to/page"
+starting from the content folder:
+ 
+    {{~/* myshortcode ref="/path/to/page/mypage" */~}}
+    
+Page Reuse shortcode must contain at least one parameter "ref" if _named parameters_ are used or the first one if _positional parameters_ are used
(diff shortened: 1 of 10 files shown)


savsheniuk:
selection_084

moorereason:
@savsheniuk,
The problem with the CLA is likely that your git commits use a different email address than your GH account.

savsheniuk:
@moorereason Oh, yes. You are right :(

bep:
Hey and welcome to Hugo devs, @savsheniuk !

I have read your code, and it looks solid. So, I understand that you are a good developer.

But I'm not sure how this fits into the "vision" #3205, and a number of changes tell me that I don't want to commit to this right now.

I would probably have preferred to make it simpler:

* Just mark them as snippets in a front matter and put them in a temporary collection and just look in all page collections when referenced (in some order)
* Instead of some new shortcode syntax, I would prefer just to create a Hugo internal shortcode (`hugo-snippet` or whatever, with a reserved namespace).

But even with those simplifications, I'm hesitant at the moment. 

But what would have been a good idea, and I feel it would solve the same problem, would be some kind of _caching internal decorator shortcode_, so I would do something like this:

```
{{< hugo-cached >}}
{{< readAndMarkdownify filename="/some-file.md" >}}
{{< /hugo-cached >}}
```

Where `hugo-cached` would be a special internal shortcode that would somehow use all the parameters for the inner shortcodes as a cache key.

savshen:
Thanks, @bep . I understand your concern. Just one note that markdownify function doesn't handle shortcodes in the specified page and in order to do that I needed to get to the context of the page rendering... so the decision was to do that on the stage of the shortcodes processing what pointed me onto writing a new kind of shortcode that works as any other shortcode but allows to use the content of the page as an inner content... it looked to me as a pretty generic solution... although I spontaneously added a few other features to that shortcode like Dynamic links processing and Page snippets.... so now it could seem a little overflown...

savshen:
Fixed the first point: new parameter "issnippet=true" can be specified inside the front-matter to mark the page as a snippet.

bep:
I will close this as I think this is too big a change for what it is.

Most of this will be gained from  #3612

PR #3250, opened 2017-03-30
docs: Fix path that govendor installs hugo in

nikolas:


moorereason:
The existing instructions work with Go 1.8's default of GOPATH=$HOME/go.  If you know how to set your own GOPATH to point somewhere else, you likely know not to use the example in the guide.  With this PR, we'll have to explain to newbies how to define a GOPATH, which we want to avoid.

PR #3252, opened 2017-03-31
hugolib: add .Frontmatter variable to Page

miltador:
There is an access to raw content but no access to raw frontmatter. This PR fills this gap.
I need the raw frontmatter in my rendered content because I want to postprocess it in some way and I need whole metadata header inside the htmls that I will replace later.

### DIFF
--- a/docs/content/templates/variables.md
+++ b/docs/content/templates/variables.md
@@ -34,6 +34,7 @@ matter, content or derived from file location.
 **.Description** The description for the page.<br>
 **.Draft** A boolean, `true` if the content is marked as a draft in the front matter.<br>
 **.ExpiryDate** The date where the content is scheduled to expire on.<br>
+**.Frontmatter** Raw metadata header.<br>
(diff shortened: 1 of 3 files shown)


bep:
You will have to present a really useful use case for this to get merged. The `RawContent` accessor was a mistake, and I have been bitten by it once, and still think it should be deprecated/removed. There is currently no real reason to store the raw front matter -- but if we expose it to the end users, we just created one that we will have to live with.

rdwatters:
@miltador Seems like this is easily covered with a `range` through `.Page.Params` and something like `plainify`. 

Perhaps the real question is whether Hugo should create features for other processors/renderers rather than focus on a core set of features just for itself. My immediate thoughts is that `.Frontmatter` is adding a feature for a single user, or at least, a *very* specific use case...so my current vote is "no."

miltador:
@rdwatters what if I want to iterate over nested fields?

Anyway it is no longer needed by me and I will feel ok if it's never going to be upstream.

Issue #3264, opened 2017-04-02
invalid link in content/using-index-md page


kamilsk:
invalid link located here https://github.com/spf13/hugo/blob/master/docs/content/content/using-index-md.md
```
2. List pages are rendered using the template heirarchy found in the [Content - List Template](http://localhost:1313/templates/list/) docs.
```

Just remove `http://localhost:1313` part.

Issue #3228, opened 2017-03-24
Move current gohugo.io and themes.gohugo.io to Netlify


bep:
This is a technical prerequisite for getting the refreshed gohugo.io site out the door, but as the current build is broken on several levels (for one, the deployment machine fails on the Git command -- and the docs was the main motivation for adding that feature in the first place ... -- and it has a too old version of Hugo). 

**So I suggest we just do this quick and hard for the current static sites.**

We already have a Netlify build for the main site, so the suggested game plan is:

1) @spf13 tries to get a theme site up and running (with the scripts and all) on Netlify. (I have no access to configure the Git link)
2) When 1) is OK, @spf13 make the DNS switch.

Sounds OK?

* And please, please don't make this into a discussion about any future site setup. This is about the **current production environment**, and that only.

/cc @budparr @rdwatters

rdwatters:
> We already have a Netlify build for the main site

Which site?  Actually, it probably doesn't matter since both the docs and the news site are on Netlify. 

I'm cool with whatever process you guys want. As I mentioned to @budparr, I already updated some of the styles per his quick and dirty suggestions, so the contrast between the new home and docs is more seamless. 

@bep, ideally, how much time are you thinking? I ask just so that I know how much time I have to finish the 1 or 2 WIP pages. . Thanks!

rdwatters:
Also, @digitalcraftsman and @budparr; I haven't been keeping up with the themes site as much. I can do a PR to line up the styles. I'm okay with getting this done now and refining the theme/styles in future iterations.

bep:
@rdwatters I will say this one final time: **This issue is about the CURRENT gohugo.io and themes.gohugo.io.**

budparr:
I imagine @biilmann could be helpful with this process, and it may be helpful to get the sites on the "Pro" plan, which is free for open source projects. Let me know if there's anything specific I can do to help.

rdwatters:
I'm cool with whatever you want to do. Still. 

I agree that the the guys at Netlify would probably be of great help. I shouldn't go on Github when I wake up at 3am randomly.

spf13:
I support this and will do whatever I need to do / can do to help.

bep:
>I support this and will do whatever I need to do / can do to help.

@spf13 I will turn this around: I assigned this issue to *you*, so you will have to drive this issue. You have the needed access to your DNS provider etc., so it isn't very practical for anyone else to do it.

But I will do anything I can do to help ...

[...]

PR #3265, opened 2017-04-02
fix issue #3264

kamilsk:
remove localhost part from link


rdwatters:
@bep This makes sense to remove, but FYI, it's non-issue with the new docs site since `using-index-md` is no longer a page.

bep:
@rdwatters don't include my name in every small little tiny detail. I get plenty of notifications as is. And judging by the momentum of #3228, we will have to maintain the current docs site for a while longer.

PR #3267, opened 2017-04-03
tplimpl: Return false in "in" func when looking for nil

### LINKED ISSUE #3225: tpl: in function should check for nil params
Panic with this test:

```
{[]string{"a", "b", "c"}, nil, false},
```

Other types will panic, too.

See https://discuss.gohugo.io/t/5903

ajiyakin:
When looking for nil in set of string, the in function should return false.
E.g.: in({[]string{"a", "b", "c"}, nil) will return false

Fixes #3225

### DIFF
--- a/tpl/tplimpl/template_funcs.go
+++ b/tpl/tplimpl/template_funcs.go
@@ -473,6 +473,11 @@ func (ic *imageHandler) config(path interface{}) (image.Config, error) {
 
 // in returns whether v is in the set l.  l may be an array or slice.
 func in(l interface{}, v interface{}) bool {
+
+	if v == nil {
+		return false
+	}
+
 	lv := reflect.ValueOf(l)
 	vv := reflect.ValueOf(v)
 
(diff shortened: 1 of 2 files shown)


ajiyakin:
Thanks @moorereason, so the `l` itself doesn't allowed to store nil value?
I'm going to fix it.

And should I rebase my commit or it's ok to add new commit for this pull request?

tpl/tplimpl/template_funcs.go
```diff
@@ -474,6 +474,11 @@ func (ic *imageHandler) config(path interface{}) (image.Config, error) {
 
 // in returns whether v is in the set l.  l may be an array or slice.
 func in(l interface{}, v interface{}) bool {
+
+	if nil == v {
```
moorereason: Not needed are [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions).

ajiyakin: Fixed :+1:

bep:
@ajiyakin this is totally my fault, I did not follow up on this PR. Your code looks very good, but in the meantime, we have moved the code around. The func in question is now below `tpl/collections`. Could you move your fix to the correct file?

ajiyakin:
Sorry @bep, I'm late.
I'm gonna check it out.

ajiyakin:
I think there's already line that handle this within [master](https://github.com/spf13/hugo/blob/master/tpl/collections/collections.go#L236) branch. So we can close #3225 then.

PR #3277, opened 2017-04-05
Add NewTab property to Hugo menus

felicianotech:
This PR adds support for opening Hugo menu items in a new tab/window.

I've seen old Issues and Discourse posts on new tab support in Hugo but they all references opening links in content/Markdown. Hugo now has a config option to tell BlackFriday to do that and it works great. This PR is specifically for Hugo menus and allowing links to open in a new tab.

Many power users who surf the web know how to middle click a link to open in a new tab or to use the content menu or even control click. Not all users are power users. I've come across many website owners who don't want to lose visitors to certain external links, thus this feature comes in handy.

Also included is a small docs tweak to also mention this new menu item property.

I tried to follow [CONTRIBUTING.md](CONTRIBUTING.md) to the tee here though I'm not sure what would be the best way to modify the tests file to include this property.


felicianotech:
Polite bump. :)

lebarde:
Hello @felicianotech,

And why not directly put an `Args` field instead of `NewTab`? In this case you can specify arbitrary code, for instance classes. That is also nice if you want to trigger CSS on that menu entry.

bep:
>And why not directly put an Args field

The idea is good, but I would call it `Params` to be in line with ... the other params.

lebarde:
This is nice :-)

bep:
... but we should think a little before we do this. The bad thing about being generic in this situation is that it makes it hard for themes to use it.

/cc @digitalcraftsman

vjeantet:
Hello, the I like `Params` idea.
As a theme dev, with this kind of feature i could describe menuEntry (in config.toml or page's frontmatter) with params like "icon", "description"...

And use it in layouts

digitalcraftsman:
This option isn't a generic approach but a reasonable addition to current options. It makes it easier to link to external sites in a menu without a dirty workaround.

I played with the thought to add a `.Params` field to the menu items as well, similar to the same-named block in the config. `.Params` gives the users all the flexibility they need, especially for edge cases.

Instead of preventing that this feature becomes the wild west regarding theme standards we should rather try to add a few more options that are very common / often requested. This helps theme maintainers to adopt this option and letting it become a standard. For everything else they can fall back to `.Params`.

felicianotech:
@digitalcraftsman So are you suggestion adding my initial PR as well as `.Params` to menu items?

digitalcraftsman:
@felicianotech I would like to see your PR merged. Let's see if the rest greenlights this decision.

Before adding anything like `.Params` I would like to take this discussion a bit further to get more feedback from others.

bep:
I think we in the future may delegate more of the menu rendering to Hugo internals, but this property fits nicely into that future; so we'll go with the `NewTab` property.

But we need a simple test.

vjeantet:
Thinking about the term `Tab`... is this word apply to every environnement, is fully understandable ?

When i saw the first notification about this PR I thought it was about rendering menus with tabs. (I'm not English native, i could come only from me 😊 )

lebarde:
That was also my first reaction. But if there is no obvious naming, the documentations could do a good job on this IMO.

felicianotech:
I'm not sure how I would add a test for this.

bep:
See #3600

felicianotech:
# Updated PR

1. Rebased on `master` since this PR was *old*.
2. Moved the docs portion of this PR to gohugoio/hugoDocs#273 since the docs was split into its own repo.
3. Added a simple test as @bep requested. The test was added to the old menu_test file which explicitly says to not add anymore tests to. :disappointed: Problem is, the new test file doesn't have a setup for config menus, just page menus so I didn't know how to add it there.
4. In issue #3600 it was brought up that this PR only adds the option for `target="_blank"`. Let me explain why. The target attribute for the HTML anchor tag can have 5 different values. "_self" which is the default and opens in the same window/tab, "_blank" which opens in a new window/tab, and 3 others related to frames. As frames are deprecated in HTML 5, this 3 other possible values are also deprecated leaving just "_self" or "_blank" which lends to this feature being a boolean.

felicianotech:
If I rebase, we lose the test I made. Should i just close this PR at this point?

PR #3280, opened 2017-04-05
installing.md: Added missing `go install` step

metachris:
The previous "Fetch from GitHub" installation instructions were missing the `go install github.com/spf13/hugo` step.


bogem:
Does 'govendor get' not install the package?

metachris:
It didn't for me yesterday.

bogem:
It works me now. Try to `go get -u github.com/kardianos/govendor` and then `govendor get github.com/spf13/hugo`

moorereason:
`govendor get` should install the package.  If it doesn't, submit an issue upstream.  Also, the GOPATH is deliberately not included in the instructions.  See #3250.

Issue #3275, opened 2017-04-04
Broken syntax-highlighting on current master branch


metachris:
Hello,

I've been trying out the code of the master branch today, and it seems that it does not do syntax-highlighting at all (no `<pre>` or `<code>` tags when using a highlight shortcode such as `{{< highlight go >}} A bunch of code here {{< /highlight >}}`).

It's working in the v0.19 release of course. 

Btw: It would be great if you could release v0.20 soon because it fixes the bug with period in tags (which is why I've been trying the master code in first place ;)!

<3

bep:
>Btw: It would be great if you could release v0.20 soon because it fixes the bug with period in tags (which is why I've been trying the master code in first place ;)!

So we should release v0.20 even if the hightlight shortcode isn't working for you?

metachris:
No need to be snarky.

I was just reporting a bug I've experienced, and on an only slightly related note expressed my desire to see the next release which fixes another pretty important issue.

bep:
>No need to be snarky.

Yes it was.

bep:
And I cannot reproduce this either, and we do have tests for this running fine on "all" systems, so I suspect others would have shouted already, so I suspect "something else" is going on.

Do you have a public sourced site with this problem?

metachris:
```
$ hugo env
Hugo Static Site Generator v0.20-DEV darwin/amd64 BuildDate: 2017-04-04T10:51:42+02:00
GOOS="darwin"
GOARCH="amd64"
GOVERSION="go1.8"
```

I was able to reproduce the bug (syntax highlighting shortcode getting ignored) by just following the steps in the [Hugo Quickstart Guide](https://gohugo.io/overview/quickstart/): 

* Create a new site: `$ hugo new site bookshelf`
* Add content: `$ hugo new post/good-to-great.md`
  * Add `{{< highlight go >}} A bunch of code here {{< /highlight >}})` to the post
* Add a theme: 
```
$ cd themes
$ git clone https://github.com/dim0627/hugo_theme_robust.git
$ (cd hugo_theme_robust; git checkout b8ce466)
```

* Serve the site: `$ hugo server --theme=hugo_theme_robust --buildDrafts`

---

Viewing the page there is simply no syntax highlighting, and this is the resulting HTML:

```
<div class="article-body" itemprop="articleBody">
      <p> A bunch of code here .</p>
</div>
```

bep:
[...]

PR #3284, opened 2017-04-05
docs/installing.md: Info about Pygments installation

metachris:
Additional information about the Pygments dependency in the "installing from source" chapter.

See #3275


bogem:
It's already written: https://gohugo.io/overview/installing/#installing-pygments-optional

bogem:
I don't think we should duplicate it.

metachris:
My problem with the docs was that I came to this site when looking for building from source: http://gohugo.io/overview/installing/#installing-from-source. Pygments is mentioned too far above and was not in the picture here.

PR #3294, opened 2017-04-06
helpers, hugolib: Refactoring

bogem:
helpers: Delete RenderingContext.getConfig
helpers, hugolib: Use only non-pointer RenderingContext 
hugolib: Prevent decoding pageParam in common cases

`hugo benchmark -s docs`
Before:
```
Average memory allocated per operation: 119514kB
Average allocations per operation: 2057862
```

After:
```
Average memory allocated per operation: 118679kB
Average allocations per operation: 2042662
```

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -19,6 +19,7 @@ package helpers
 
 import (
 	"bytes"
+	"fmt"
 	"html/template"
 	"os/exec"
 	"unicode"
@@ -33,7 +34,6 @@ import (
 	jww "github.com/spf13/jwalterweatherman"
 
 	"strings"
-	"sync"
 )
 
 // SummaryLength is the length of the summary that Hugo extracts from a content.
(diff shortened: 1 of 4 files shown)


bep:
It looks fine, but I will wait to pull this in until we ship 0.20.

bogem:
So, do I merge it?

bogem:
I think, I should return pointers to RenderingContext, because it really impacts the performance (+~5%)

bep:
Yes, I think some other reason I made it a pointer ... But I cannot remember. Sorry that I haven't had (or ...) time to look into this.

bogem:
Delete `helpers, hugolib: Use only non-pointer RenderingContext` commit.
New benchmarks:
Before:
```
Average time per operation: 425ms
Average memory allocated per operation: 125821kB
Average allocations per operation: 2115851
```

After:
```
Average time per operation: 421ms
Average memory allocated per operation: 125184kB
Average allocations per operation: 2100659
```

Issue #285, opened 2014-05-11
Extract date and permalink from filename


xaprb:
One of the Jekyll features I enjoy is extracting the date and slug/permalink from the filename so I don't need to specify them in the front matter. It would be nice if Hugo would support this as well.

jon49:
Here's a workaround until the enhancement made for people on linux/macs, or if you're smart and figure out how to do it on windows.

```make
POST_SRC := $(subst ./,,$(shell find ./src/post/ -regextype posix-extended -regex '.*[0-9]{4}-[0-9]{2}-[0-9]{2}-.*\.md'))
POST_DEST := content/post/

default: all

define POSTS_PARSE
_target := $(POST_DEST)$(shell echo $1 | grep -Po '\w+\.md')
POSTS := $$(POSTS) $$(_target)
$$(_target): $1
	date=$$$$(basename $$< | grep -Po '\d{4}-\d{2}-\d{2}'); \
		sed "1 adate: $$$$date" $$< > $$@
endef

$(foreach x,$(POST_SRC),$(eval $(call POSTS_PARSE,$x)))

posts: $(POSTS)

init:
	mkdir -p content/post/

all: posts

.PHONY: all posts init
```

put your original files in `./src/post` and it will export to `./content/post`. Just type `make` for one time use and `watch make` when actively editing files.

robsonsobral:
It's easier to keep the files organized too.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

borisschapira:
@bep 

* Easier to keep the files organized
* No need to charge the Front-Matter with the info (same goes for the language for example)

borisschapira:
And I'll add: easier to migrate from Jekyll :)

xaprb:
I agree, it's a feature request and valuable for the reasons Boris mentioned.

robsonsobral:
Please, don't give up on this.

vassudanagunta:
@bep, I recommend you add this to the next milestone, and perhaps also #3805, rather than one of the specific solution proposals. There are at least three of those (#3310, #3762, #3859) and I will be adding a fourth! 😄

PR #3310, opened 2017-04-09
Allow dates to be parsed from filenames

devjack:
This has been partially discussed in #285.

This allows the date variable to be parsed from a filename with the aim to ease content migration.  Date can be 'datestamped' in any part of a filename, with the example being `2017-01-31-mypost.md`.

Filename fallback is enabled via the `useFilenameDateAsFallback` config value and can find custom date formats using `filenameDateFallbackPattern` (config) and parsed into the `date` param using `filenameDateFallbackFormat` (config).


devjack:
First time golang and hugo contrib & I've had some trouble getting tests to pass locally too (both on master and my patch). Would appreciate any pointers on configuring/running/fixing these for the PR. Cheers!

bep:
Re. testing I would recommend using `make`:

I.e. 

* make sure you have govendor installed and the vendored lib synced: `make vendor`
* Run all tests and checks: `make check`

I will comment on the PR itself in another commebnt.

hugolib/page.go
```diff
@@ -1021,6 +1021,14 @@ func (p *Page) update(f interface{}) error {
 			p.Date = fi.ModTime()
 			p.Params["date"] = p.Date
 		}
+	} else if p.Date.IsZero() && p.s.Cfg.GetBool("useFilenameDateAsFallback") {
```
bep: We have a little too many config variables as is, and I think we can drop this one and just check that `filenameDateFallbackPattern` is set. But other than this minor detail, this looks very good.

devjack: @bep thats a good point.  With that in mind, perhaps one config value to provide the regex that defaults to nil? That way the parsing (regardless of format separator) can be configured but there's BC compat if Nil or not configured?

bep: I'm not sure I follow the logic here. Are you saying we could reduce the number of configs to 1? How?

devjack: I'm thinking so - its balance between config and 'magic'.

If the regex and time format is always the same, you can infer it from named groups.

For example:
```
(?P<datestamp>(?P<stdYear>\d{4})\-(?P<stdZeroMonth>\d{2})-(?P<stdZeroDay>\d{2}))
```

The catch is that time format const's aren't public so there would need to be a copy of them in Hugo for mapping (probably just day/month/year with the 0 padded variants too).

So its either:
 a) add 2x config params and let people choose format and pattern
 b) add 1x config param and force people to use the same format as the pattern.

bep: Sorry, I forgot this, as this will likely be copy-pasted from the documentation: 

* If the code to create a time format string from that regexp is not too complex, then 1 config is fine.
* else just add two config vars.

devjack:
Cheers @bep,
Picked this one back up and patched (38a313f) as per your suggestion. My speculative automagic turned out to be too complex so settled on two config values with an updated sample in the docs.
Cheers!

maxandersen:
I've made an updated patch that resolves the conflict on docs files, you can see it here: https://github.com/maxandersen/hugo/tree/devjack-filename-date-fallback

It seem to work great for me so I hope this can make it in as I really need this one for my migration of ~40 old blog posts using date in filenames.

maxandersen:
btw. imo this should really just be enabled by default or at least have a default value for `filenameDateFallbackPattern` for iso dates - just so you don't have to put "complex" regular expressions into your config.

rdwatters:
This is a great feature and will surely help with those migrating content from other SSGs :smiley:

That said, I wouldn't enable this by default with so many sites in the wild, IMHO.

maxandersen:
> That said, I wouldn't enable this by default with so many sites in the wild, IMHO.

true - but having a "useFilenameDateAsFallback = true" would then imo be better since I would argue in 99% of all cases the dateformat will be standard iso (yyyy-mm-dd).

One thing I noticed while using this patch is that it should also set the "slug" to remove the date from the path - at least as an option. so instead of `/post/2017-08-01-my-new-blog` as url it would say `/post/my-new-blog`.

maxandersen:
i'll try get my patch updated to handle proper `slug` extraction too. will need some sleep first though ;)

maxandersen:
also, just looked in jekyll and awestruct and they both basically assume `<yyyy>-<mm>-<dd>-<slug>` with some variance that can be better handled with logic rather than letting users set the regular expressions for matching. I'll get an updated patch that handles this as good as jekyll/awestruct seem to do it.

hugolib/page_test.go
```diff
@@ -847,6 +848,22 @@ func TestPageWithDate(t *testing.T) {
...
+	cfg, fs := newTestCfg()
+
+	writeSource(t, fs, filepath.Join("content", "2017-01-31-simple.md"), simplePageNoDate)
+
+	s := buildSingleSite(t, deps.DepsCfg{Fs: fs, Cfg: cfg}, BuildCfg{SkipRender: true})
+
+	require.Len(t, s.RegularPages, 1)
+
+	p := s.RegularPages[0]
+	d, _ := time.Parse(time.RFC3339, "2017-01-31")
```
maxandersen: this always fails as RFC3339 requires timezone info.

lambrospetrou:
What is the status of this? Is it going to be integrated, released?

bep:
Sorry for the delay,

I haven't wanted to commit to this change (i.e. encode more information into the filename) until I saw the end of my current task (page bundles, a task that I have been revisiting for a pretty long time now).

This change may look innocent enough (and it probably is), but I have learnt some hard lessons in Hugo: Structural changes (i.e. stuff that makes people change their themes/content) is easy to add, but very hard to change/remove); even the smallest variable changes have had significant cost (`.Now` to `now` etc.).

These PRs are not forgotten and I understand them pretty well. But if you for some reason need to remove your forks, that is understandable.

maxandersen:
> But if you for some reason need to remove your forks, that is understandable.

don't need to remove them I just wish I could stop having to maintain separate fork :)

btw. this change shouldn't change any behavior - all past behavior is as before and if you dont want dates in title to affect anything you can disable the behavior.

But thanks for the update - hope it can get added soon :)

andriisoldatenko:
@devjack are you going to continue or do you need any help?

devjack:
@andriisoldatenko unfortunately I have no time at present to dedicate to this again in the immediate future. It’s open for anyone else that would like the feature etc.

PR #3328, opened 2017-04-12
tpl/collections: Add support for interfaces to intersect

### LINKED ISSUE #1952: tpl: intersect needs some love

moorereason:
Fixes #1952

### DIFF
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -298,21 +298,49 @@ func (ns *Namespace) Intersect(l1, l2 interface{}) (interface{}, error) {
 					l2vv := l2v.Index(j)
 					switch l1vv.Kind() {
 					case reflect.String:
-						if l1vv.Type() == l2vv.Type() && l1vv.String() == l2vv.String() && !ns.In(r.Interface(), l2vv.Interface()) {
-							r = reflect.Append(r, l2vv)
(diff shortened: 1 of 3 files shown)


moorereason:
Not sure why Netlify failed, but I've rebased this PR with the latest template refactoring. Should be good to go.

PR #3363, opened 2017-04-18
add option to render summary or full content in RSS feed

kropp:
Added option to generate full feed or only summaries.

RSS feed generation changed in 0.20. Some prefer new behavior, some prefer old (at least me and there is also #3341 I just noticed), thus an option. It is set to `true` by default to honor current (new) behavior.

I believe a custom template is too complex for this task, because this is probably the only thing I'd like to change and maintaining yet another template will require much more attention. And there is already `rssLimit` option, so why not? 😃

### DIFF
--- a/docs/content/overview/configuration.md
+++ b/docs/content/overview/configuration.md
@@ -174,6 +174,8 @@ along with their current, default values:
     pygmentsUseClasses:         false
     # maximum number of items in the RSS feed
     rssLimit:                   15
+    # if true RSS contains only article summaries
+    rssSummary:                 true
     # default sitemap configuration map
     sitemap:
     # filesystem path to read files relative from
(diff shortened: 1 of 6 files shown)


Cj-Malone:
Either this or my [pr](https://github.com/spf13/hugo/pull/3341) would please me, although I think it's better for the client to decide to show a description or the full post, and to know if it's the full content. Maybe using the content namespace in this PR would be best?

```
<description>{{ .Summary | html }}</description>
{{ if ne .Site.RSSSummary }}<content:encoded>{{ .Content | html }}</content:encoded>{{ end }}
```

This way it's still an option for the site owner, and informs the client if it should display a "Read more" link.

rdwatters:
> although I think it's better for the client to decide to show a description or the full post

By this logic, wouldn't it make more sense to include `.Description`, the page variable> Summaries are typically auto-generated and confused with what is really the idea of a "lead" paragraph. This conversation has taken place before on a different PR/issue...

Cj-Malone:
See [this](https://github.com/spf13/hugo/pull/3341/commits/190b097abc3866fd5859c9831323456ab37d8da7) to prefer `.Description` over `.Summary`. I still it should be up to the client (and the publisher) to decide on full content vs a "description" not the CMS.

PR #3365, opened 2017-04-19
Add check-vendor target to Makefile

moorereason:
The check-vendor target will exit with an error if any files in vendor/
differ from what's checked into git.

govendor checksums can get out of sync if extraneous files are in the
local, third-party repo (see kardianos/govendor#129).  govendor also
updates the vendor.json upon "sync."  We can use that quirk to detect
out of sync checksums in CI, so I've added check-vendor to the check
target dependencies.


moorereason:
The vendor file is currently out of sync, so I expect CI to fail.

bep:
@moorereason I merged this, but then removed it from make check. I could not get the build to succeed, so I suspect there are some flaw(s) in the script.

PR #3369, opened 2017-04-20
Allow "after" to skip all elements

dobesv:
Previously if after would skip all elements, it would return an error.  However, there's nothing wrong with skipping all elements and returning an empty list.


dobesv:
See also https://github.com/spf13/hugo/issues/3368

bep:
I agree, but it would be even greater with a test case.

bep:
@dobesv would it be possible to add a test case?

dobesv:
Sorry I don't have time. If you care about the feature go for it!

PR #3384, opened 2017-04-24
Set extldflags to static when building

### LINKED ISSUE #3382: Hugo binary no longer statically linked on Linux (starting 0.20.3)
Hi there

Since version 0.20.3, it appears the Hugo binary is no longer being statically linked.

This means when running Hugo within an alpine container, 'no such file or directory' is output (because the linked libraries are not present). I'm assuming this is to do with the transition to using goreleaser, as part of this issue: https://github.com/spf13/hugo/issues/3358

From a quick glance, it looks like goreleaser allows us to specify ld flags. etc for use when building, so this should be possible!

Thanks

munnerz:
This sets the `-static` ldflag when building Hugo, so that it can be run in a minimal Docker container.

EDIT: Does in fact work fine when building for Windows

Closes #3382

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -1,7 +1,7 @@
 build:
   main: main.go
   binary: hugo
-  ldflags_template: -s -w -X hugolib.BuildDate={{.Date}} -extldflags "-static"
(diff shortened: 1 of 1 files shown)


munnerz:
After testing, it seems all is fine on Windows!

PR #3391, opened 2017-04-25
Generalize BasePathFs usage

moorereason:
Related to spf13/afero#76

### DIFF
--- a/hugofs/fs.go
+++ b/hugofs/fs.go
@@ -34,7 +34,7 @@ type Fs struct {
 
 	// WorkingDir is a read-only file system
 	// restricted to the project working dir.
-	WorkingDir *afero.BasePathFs
+	WorkingDir afero.Fs
 }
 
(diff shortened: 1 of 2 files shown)


bep:
See my comments on the other thread.

moorereason:
I'm not sure what the specific security concern is here.  How is `hugofs.Fs.WorkingDir` vulnerable to abuse?  If it's open to abuse/misuse somehow, we should change the `hugofs.Fs` data structure to protect the underlying afero.Fs implementations with Setters and type assertions.

bep:
>'m not sure what the specific security concern is here. How is hugofs.Fs.WorkingDir vulnerable to abuse? 

I want to know that the file access is confined to a restricted directory (i.e. `WorkingDir`) to prevent obvious stuff like `readFile "/etc/passwd"`.

But I'm now sure why I should make any arguments here, you are the one who proposes a change.

PR #3406, opened 2017-04-28
Fix about 7 very minor typos in docs/

tom-harvey:


bogem:
But please read the [CONTRIBUTING](https://github.com/spf13/hugo/blob/master/CONTRIBUTING.md) to have correct commit messages next time:

PR #3416, opened 2017-04-30
Replace cdn.mathjax.org with cdnjs.cloudflare.com

takuti:
MathJax officially informed that [their self-hosted *cdn.mathjax.org* is retired](http://docs.mathjax.org/en/latest/configuration.html#loading-mathjax-from-a-cdn), and they currently recommend to use *cdnjs.cloudflare.com*.

Thus, this PR fixes corresponding MathJax document.

Note: Since *cdnjs.com* does not provide a link to the **latest** version of MathJax, sample snippet specifies to use version 2.7.1 (i.e., the latest version at this moment).

### DIFF
--- a/docs/content/tutorials/mathjax.md
+++ b/docs/content/tutorials/mathjax.md
@@ -20,10 +20,10 @@ This is not an introduction into actually using MathJax to render typeset mathem
 
 ## Enabling MathJax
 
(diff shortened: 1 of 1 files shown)


bogem:
Please sign the CLA

takuti:
@bogem signed.

bogem:
But please read the CONTRIBUTING to have correct commit messages next time:

PR #3419, opened 2017-05-01
tpl: Add basic tests for all namespace init funcs

moorereason:


bep:
This is the kind of tests that I only like when I can get them for free. 700 LOC in 18 files isn't free -- and making sure every package has at least 1 example test will accomplish the same test coverage.

PR #3424, opened 2017-05-02
add ability to disable content based on extension

chaseadamsio:
allows Hugo users to define content extensions they want to ignore during build. As an example, if a user wants to ignore files in their content directory that have `.org` extensions, they can set the following configuration in their `config.toml`:

```
[disableContentType]
  org = true
```

Original request from @strow: https://github.com/chaseadamsio/goorgeous/issues/32
Also mentioned in this PR: https://github.com/spf13/hugo/pull/3180#issuecomment-286499092

### DIFF
--- a/hugolib/config.go
+++ b/hugolib/config.go
@@ -112,4 +112,5 @@ func loadDefaultSettingsFor(v *viper.Viper) {
 	v.SetDefault("defaultContentLanguageInSubdir", false)
 	v.SetDefault("enableMissingTranslationPlaceholders", false)
 	v.SetDefault("enableGitInfo", false)
+	v.SetDefault("ignoreFiles", make([]string, 0))
 }


bep:
* We have an `ignoreFiles` setting, does this not work for this?
* If no, why not fix that?

strow:
Thanks, but, although the ignoreFiles seems to work fine from the
command line (hugo), it does not work when in server mode (hugo server)
*if* the offending file (a .org file) changes while in server mode.

This is very repeatable.

Larrabee

Bjørn Erik Pedersen writes:

> * We have an ignoreFiles setting, does this not work for this? 
> * If no, why not fix that? 
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.


-- 
L. Larrabee Strow
UMBC Physics Department
Email: strow@umbc.edu
Cell: 724-288-6933

bep:
@strow sure, but that is a known issue (it even has a GitHub issue somewhere); that is fixed by ... fixing that issue, not by adding another setting which may or may not have the same problem ...

PR #3442, opened 2017-05-07
docs: Use single link element for RSS in example

tobbez:
The rel attribute supports specifying a set of values, not only a single
one. Using two link elements can also cause browsers to show the feed
twice in menus.


XhmikosR:
This is definitely right and should be merged.

digitalcraftsman:
Hello @tobbez,

thank you for submitting this addition. I've merged your contribution as  00d63215.

Cheers,
Digitalcraftsman

/cc @rdwatters

PR #3451, opened 2017-05-10
tpl/time: Support overlapping namespace and template func

### LINKED ISSUE #3421: "time" template func clashes with "time" namespace
I have remamed the alias "asTime" for now, but that will break some templates in the wild.

moorereason:
Fixes #3421

### DIFF
--- a/tpl/time/init.go
+++ b/tpl/time/init.go
@@ -32,8 +32,7 @@ func init() {
 				// If no args are passed to `time`, assume namespace usage and
 				// return namespace context.
 				//
-				// If args are passed, show a deprecation warning and attempt to
-				// simulate the old "as time" behavior.
+				// If args are passed, call AsTime().
 
 				if len(v) == 0 {
 					return ctx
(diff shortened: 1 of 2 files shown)


tpl/time/init.go
```diff
@@ -25,8 +25,26 @@ func init() {
...
-			Context: func() interface{} { return ctx },
+			Name: name,
+			Context: func(v ...interface{}) interface{} {
+				// Handle overlapping "time" namespace and func.
+				//
+				// If no args are passed to `time`, assume namespace usage and
+				// return namespace context.
+				//
+				// If args are passed, show a deprecation warning and attempt to
+				// simulate the old "as time" behavior.
```
bep: Can you fix the description?

tpl/time/init.go
```diff
@@ -42,7 +60,7 @@ func init() {
 		)
 
 		ns.AddMethodMapping(ctx.AsTime,
-			[]string{"asTime"}, // TODO(bep) handle duplicate
+			[]string{"asTime"},
```
bep: Remove this; we don't need this when we have `time`.

PR #3452, opened 2017-05-10
doc: RSS uses .Summary instead of .Content since 0.20.

carandraug:
Release notes for 0.20 say:

> `RSS` description in the built-in template is changed from full `.Content` to `.Summary`. This is a somewhat breaking change, but is what most people expect from their RSS feeds. If you want full content, please provide your own RSS template.

However, docs had not been updated. This fixes that.

I noticed one other difference on the template:

    <atom:link href="{{.Permalink}}" rel="self" type="application/rss+xml" />

instead of

     {{ with .OutputFormats.Get "RSS" }}
        {{ printf "<atom:link href=%q rel=\"self\" type=%q />" .Permalink .MediaType | safeHTML }}
    {{ end }}


digitalcraftsman:
Hello @carandraug,

thank you for submitting this addition. You might noticed that we moved Hugo into it's own GitHub organization. Along the way we moved the docs in it's own repo. Hence you can find your pr merged in https://github.com/gohugoio/hugoDocs/commit/9b38a89d7f0006ea0145177d1cb5fc6775694f77.

Cheers,
Digitalcraftsman

PR #3456, opened 2017-05-11
Fixed incorrect specification of directory structure

TejasQ:
Hugo looks for the list/single templates of a taxonomy in `/layouts/taxonomy`, not `/layouts/taxonomies` as correctly mentioned in [this doc page](https://gohugo.io/templates/list/). This change is simply a minor reconciliation.


bogem:
But please read the [CONTRIBUTING](https://github.com/spf13/hugo/blob/master/CONTRIBUTING.md) to have correct commit messages next time:

Issue #3092, opened 2017-02-24
tplimpl: isset should return an error on unsupported type


moorereason:
`switch` should have a default case that returns an error when an unsupported `reflect.Kind` is encountered.

See https://discuss.gohugo.io/t/help-understanding-isset/5577/2

kaushalmodi:
Related: https://github.com/spf13/hugo/issues/3089

Thanks for fixing this!

kaushalmodi:
Confirmed the fix.. now if I do `isset .Site "DisqusShortname"`, I get:

> ERROR 2017/03/01 18:19:13 Error while rendering "page posts/installing-go-toolchain.md": template: /home/kmodi/hugo/kaushalmodi.gitlab.io/themes/refined/layouts/_default/baseof.html:45:10: executing "after_main" at <isset .Site "DisqusS...>: error calling isset: unsupported type "ptr"

bep:
Reopened -- as this breaks lots of themes and the theme site. Will have to be re-evaluated.

vielmetti:
Looks like this affects the "Casper" theme.

saso94:
In "hugo-coder" theme, when I start to building site, in the command line there is this warning:

WARNING: calling IsSet with unsupported type "invalid" (<nil>) will always return false.

PR #3480, opened 2017-05-18
tpl/collections: Log an error on unsupported types in IsSet

moorereason:
Unsupported types are currently silently ignored by IsSet.  An earlier
attempt was made to solve the issue by returning an error.  That attempt
was reverted since it broke some existing themes.

So instead, we'll log an error.  Hopefully, people will stop using IsSet
in this way, and we can eventually return an error outright.

Updates #3092


bep:
An ERROR is almost as breaking. CI-builds will fail (including the themes). I suggest you do something ala 

```
DistinctFeedbackLog.Printf("WARNING: bla bla 
```

moorereason:
Good catch.  I've added a new commit to use the Feedback logger.

PR #3498, opened 2017-05-22
docs: Remove Unmaintained Frontends from Tools

onedrawingperday:
The unmaintained Rango frontend and the Hugopit "experiment" should be removed from the Hugo Tools/Frontends section of the documentation as they are unusable.

Rango can mess up with other dependencies installed in the Go Path as reported here: 
https://github.com/stayradiated/rango/issues/4

Hugopit's developer himself stated that this is an unmaintained experiment and he seemed a bit surprised that it is included in Hugo's documentation over here: https://github.com/sjardim/Hugopit/issues/3

### DIFF
--- a/docs/content/tools/_index.md
+++ b/docs/content/tools/_index.md
@@ -77,10 +77,8 @@ If you don't want to use [Wercker for automated deployments]({{< relref "tutoria
 
 Do you prefer an graphical user interface over a text editor? Then give these frontends a try:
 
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Thank you @onedrawingperday for cleaning up the tools section

onedrawingperday:
@digitalcraftsman There may be a problem with Lipi also. I've also mentioned you at the forum.  You may want to test Lipi yourself. It failed on me.

PR #3506, opened 2017-05-22
tpl/collections: Support interfaces in union

### LINKED ISSUE #3411: tpl: Union needs better support for interface{} values
See https://discuss.gohugo.io/t/6382.  Similar issue raised about `intersect` in #1952.

moorereason:
Fixes #3411 

Also includes an additional commit to add missing uint support to `In`

### DIFF
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -261,6 +261,13 @@ func (ns *Namespace) In(l interface{}, v interface{}) bool {
 						return true
 					}
 				}
+			case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+				switch vv.Kind() {
+				case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+					if vv.Uint() == lvv.Uint() {
+						return true
+					}
+				}
 			case reflect.Float32, reflect.Float64:
 				switch vv.Kind() {
 				case reflect.Float32, reflect.Float64:
@@ -564,7 +571,6 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
 			if l1v.Type() != l2v.Type() &&
 				l1v.Type().Elem().Kind() != reflect.Interface &&
 				l2v.Type().Elem().Kind() != reflect.Interface {
-
 				return r.Interface(), nil
 			}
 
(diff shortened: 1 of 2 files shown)


tpl/collections/collections.go
```diff
@@ -561,21 +568,95 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
...
 			}
 
 			for j := 0; j < l2v.Len(); j++ {
-				elem := l2v.Index(j)
-				if !ns.In(r.Interface(), elem.Interface()) {
-					r = reflect.Append(r, elem)
+				l2vv := l2v.Index(j)
+
+				switch l1vv.Kind() {
+				case reflect.String:
```
bep: Why do you need all these case switches? I.e. could you not delegate that to ns.In?

tpl/collections/collections.go
```diff
@@ -561,21 +568,95 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
...
+					l2t, err := toFloat(l2vv)
+					if err == nil && !ns.In(r.Interface(), float32(l2t)) {
+						r = reflect.Append(r, reflect.ValueOf(float32(l2t)))
+					}
+				case reflect.Float64:
+					l2t, err := toFloat(l2vv)
+					if err == nil && !ns.In(r.Interface(), l2t) {
+						r = reflect.Append(r, reflect.ValueOf(l2t))
+					}
+				case reflect.Interface:
```
bep: When I read "support interfaces" the last time, I expected a more complete support set than this; i.e. I wanted `*Page`. You may argue that is outside of scope, or the next step or whatever, but to me that is a sign that we have to look at the problem again and see if it can be made simpler/more generic, which would cover more bases.

tpl/collections/collections.go
moorereason: Good idea.  I'll investigate.

moorereason: What I'm doing is this PR is casting to like types.  So, `Union([]int{1, 2}, []interface{}{"3"})` would return `[]int{1, 2, 3}`.  `In` doesn't do that -- it only compares matching types.

Much of this matching logic could be pushed to `In`, but `Union` would still have to get a matching type for the returned slice (whose type matches that of `l1`).

I'm open to suggestions.

tpl/collections/collections.go
moorereason: That's outside the scope of this PR.  I'm mainly trying to support data from JSON and friends (which is generally a bunch of `interface{}` values).

PR #3508, opened 2017-05-23
docs: Add tutswiki.com to the showcase

chankeypathak:
Adding a site which is inspired from Hugo.


digitalcraftsman:
Hello @chankeypathak,

thank you for adding your website to the showcase of Hugo. I've merged your contribution as  077dabe3.
Hugo's showcase section should be updated within the next minutes.

Cheers,
Digitalcraftsman

bep:
@digitalcraftsman adding showcase sites that is "inspired from Hugo" is a tad outside of the scope. I think the minimum should be that its actually built with Hugo, i.e. has a Hugo generator tag on home page.

chankeypathak:
@bep Sorry the word "inspired" gave it another meaning. The site is built using "Hugo" only. I'll add the generator tag.

PR #3521, opened 2017-05-24
Add noindex tag to HTML generated by Hugo Aliases

onedrawingperday:
So that Googlebot (and other search engine bots) can stop indexing these old URLs.

### DIFF
--- a/hugolib/alias.go
+++ b/hugolib/alias.go
@@ -30,8 +30,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


XhmikosR:
@bep; I don't suppose the aliases are added to the sitemap? If so they should be excluded.

As for this PR, it makes total sense.

onedrawingperday:
@XhmikosR 
No aliases URLs are not added to the sitemap.

PR #3531, opened 2017-05-26
tpl: Add uniq function

adiabatic:
As I [mentioned in the forums](https://discuss.gohugo.io/t/adding-uniq-to-template-functions/6710).

### DIFF
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -373,6 +373,15 @@ e.g.
        {{ .Content }}
     {{ end }}
 
+
+### uniq
+
+Takes in a slice or array and returns a slice with subsequent duplicate elements removed.
+
+    {{ uniq (slice 1 2 3 2) }}
+    {{ slice 1 2 3 2 | uniq }}
+    <!-- both return [1 2 3] -->
+
 ## Files
 
 ### readDir
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -587,3 +587,41 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
 		return nil, errors.New("can't iterate over " + reflect.ValueOf(l1).Type().String())
 	}
 }
+
+// Uniq takes in a slice or array and returns a slice with subsequent
+// duplicate elements removed.
+func (ns *Namespace) Uniq(l interface{}) (interface{}, error) {
+	if l == nil {
+		return make([]interface{}, 0), nil
+	}
+
+	lv := reflect.ValueOf(l)
+	lv, isNil := indirect(lv)
+	if isNil {
+		return nil, errors.New("invalid nil argument to Uniq")
+	}
+
+	var ret reflect.Value
+
+	switch lv.Kind() {
+	case reflect.Slice:
+		ret = reflect.MakeSlice(lv.Type(), 0, 0)
+	case reflect.Array:
+		ret = reflect.MakeSlice(reflect.SliceOf(lv.Type().Elem()), 0, 0)
+	default:
+		return nil, errors.New("Can't use Uniq on " + reflect.ValueOf(lv).Type().String())
+	}
+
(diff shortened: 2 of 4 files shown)


adiabatic:
I made it fail nicely for `{1, 1, true},` and added support for arrays, although I'm not sure how they'd get passed in.

`{"foo", "fo", false},`, on the other hand, I changed to `{"foo", "fo", true},` because, on further reflection, I don't want this to operate on strings because it'll encourage normalization-dependent behavior. Depending on whether diacritical marks are part of the letter's code point or not, "foaö" could Uniq down to either "foaö" (if ö is one code point) or "foä" (if o and ¨ are separate code points).

tpl/collections/collections.go
```diff
@@ -587,3 +587,48 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
...
+	}
+
+	switch lv.Kind() {
+	case reflect.Array, reflect.Slice:
+	// okay
+	default:
+		return nil, errors.New("Can't uniquify a " + reflect.ValueOf(lv).Type().String())
+	}
+
+	var ret reflect.Value
```
moorereason: Can you refactor this a bit?  Define `ret` before the `switch` and merge the following `if` statements as cases in the `switch`.

tpl/collections/collections.go
```diff
@@ -587,3 +587,48 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
...
+// duplicate elements removed.
+func (ns *Namespace) Uniq(l interface{}) (interface{}, error) {
+	if l == nil {
+		return make([]interface{}, 0), nil
+	}
+
+	lv := reflect.ValueOf(l)
+	lv, isNil := indirect(lv)
+	if isNil {
+		return nil, errors.New("Can't uniquify a nil value")
```
moorereason: "uniquify" doesn't exist.  Change this to simply `invalid nil argument to Uniq`.

tpl/collections/collections.go
```diff
@@ -587,3 +587,48 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
...
+	}
+
+	switch lv.Kind() {
+	case reflect.Array, reflect.Slice:
+	// okay
+	default:
+		return nil, errors.New("Can't uniquify a " + reflect.ValueOf(lv).Type().String())
```
moorereason: Change to `can't use Uniq on`.

tpl/collections/collections.go
```diff
@@ -587,3 +587,48 @@ func (ns *Namespace) Union(l1, l2 interface{}) (interface{}, error) {
...
+
+	for i := 0; i != lv.Len(); i++ {
+		lvv := lv.Index(i)
+		lvv, isNil := indirect(lvv)
+		if isNil {
+			continue
+		}
+
+		inRet := ns.In(ret.Interface(), lvv.Interface())
+		if !inRet {
```
moorereason: Simplify to

```go
if !ns.In(ret.Interface(), lvv.Interface()) {
```

docs/content/templates/functions.md
```diff
@@ -373,6 +373,15 @@ e.g.
        {{ .Content }}
     {{ end }}
 
+
+### uniq
+
+Loops through an array and creates a new array with only the first occurrence of any given value.
```
moorereason: Generally, try to make the docs description match the func's godoc comment.  We eventually plan to auto-generate these from godoc.

adiabatic:
`make test` and `make test-race` work fine on my machine, but one of the Travis builds broke in `hugolib.TestShortcodeFigure` on linux. Is this something I should try to correct, or was this caused by someone else? I'm still new at using git with other people.

moorereason:
This looks great, @adiabatic.  Thanks for persevering!

The failed Travis builds are unrelated to this PR.

PR #3549, opened 2017-05-31
comments out indent code causing build failure

umamiMike:
I go get and built on master, the build failed on this line.  Build succeeded after.  I am probably wrong, havent looked into it in any detail.


bep:
You need Go >= 1.7.

PR #3556, opened 2017-06-03
Add option to strip trailing slashes from URLs

tgwizard:
This is a follow-up to https://github.com/spf13/hugo/issues/492 and https://discuss.gohugo.io/t/links-without-a-trailing-slash/831, especially "PRs are welcome."

I'm not that familiar with Hugo, nor with Go in general. Any feedback is welcome, but especially:

1. Are these all the places that need to respect the new flag? (It does work for me when running on one of my projects.)
2. `removeTrailingSlashesFromURLs` is kinda long. Is there some better name?
3. What tests should I add for this?


bep:
1. I think so. Maybe also the URLs in menus.
2. `trimURLSlash` ...?
3. I suggest you look for a simple site test with pages + paginators and test for both off/on in a loop.

Note: 

* Docs below /docs/commands is autogenerated
* I don't want a CLI flag for this option. This should live in the config.toml only; if you want to set it as a flag, set an OS env var.

hugolib/pagination.go
```diff
@@ -527,6 +527,11 @@ func newPaginationURLFactory(d targetPathDescriptor) paginationURLFactory {
 
 		targetPath := createTargetPath(pathDescriptor)
 		targetPath = strings.TrimSuffix(targetPath, d.Type.BaseFilename())
+
+		if d.PathSpec.RemoveTrailingSlashesFromURLs {
+			targetPath = strings.TrimSuffix(targetPath, "/")
+		}
```
bep: What wuld be better would be to put this config option into the `pathDescriptor`. This way you can handle it in one place and add test(s) here:

https://github.com/spf13/hugo/blob/master/hugolib/page_paths_test.go#L40

tgwizard: By `pathDescriptor` do you mean `targetPathDescriptor`?

If so: how do I send in the configuration for when the `targetPathDescriptor` is initialized? Should I update `initTargetPathDescriptor` to take a config object? That seems to require changes in multiple places.

tgwizard:
Thanks for the feedback! I've amended to commit to call the option `trimTrailingURLSlashes`, and only allow it to be set through the config file.

Depending on the answer to the question ^ I'm not sure exactly where to add tests. (Regarding 1. -
 the change already affects URLs in menus).

tgwizard:
Bump @bep

bep:
@tgwizard re. testing etc. You should add this to `createTargetPath` and test it in `page_paths_test.go`.

lights7:
with a slash the table of content won't work, for example the link of the 2nd section is https://example.com/post/post1/#2ndsection
while clicking it will not go to that section. However without backslash https://example.com/post/post1#2ndsection, the browser will move to 2nd section. 

Anyone have the same problem like what I have?

PR #3557, opened 2017-06-04
docs: Add thislittleduck.com to the showcase

petercossey:
We just launched our company site using Hugo - had a great time using it and I think it show's off Hugo's flexibility.


bep:
Nice site, but I don't see any Hugo generator tag in the source on the home page. That is a requirement to get on the showcase page. Hugo adds this on the index page by default, but you seem to have disabled that feature.

petercossey:
@bep the missing Hugo generator tag is a mystery for us... we didn't turn it off in our config.toml and I think out theme isn't doing anything out of the ordinary. We'll follow it up and get it added.

petercossey:
@bep we've fixed the issue, the homepage now includes the following tag:

```html
<meta name="generator" content="Hugo 0.19" />
```

Issue #100, opened 2013-10-08
Add Hugo import from wordpress


spf13:
'hugo import ....'

anthonyfok:
There is also a relatively new [WordPress to Hugo Exporter][], which works both as a WordPress plugin as as a PHP CLI script, which Cyrill Schumacher (@SchumacherFM) wrote based on the [WordPress to Jekyll Exporter](https://github.com/benbalter/wordpress-to-jekyll-exporter) that @xaprb has used about a year ago:

 * https://github.com/SchumacherFM/wordpress-to-hugo-exporter

[WordPress to Hugo Exporter]: https://github.com/SchumacherFM/wordpress-to-hugo-exporter

It must be pretty good: A recent [blog post](http://baty.net/2014/12/2014-12-31-my-hugo-experiment/) by Jack Baty (@jackbaty) has nothing but praises for @SchumacherFM's great tool:

> Converting from Wordpress was surprisingly easy. I used a plugin by Cyrill Schumacher and had all of my 1800+ posts, images, etc. converted for use with Hugo in less than an hour. All links were preserved so I don’t need to create a bunch of rewrite rules in Apache like every other time I’ve done this. Every step was easier than expected so I just kept going!

Moving forward, should we:

1. considered this issue solved, maybe except documenting the existence of this wonderful WordPress plugin/CLI tool?  Or...
2. create a `hugo import ...` command that would fetch the [WordPress to Hugo Exporter][] and guide the user to having it installed and run?  Or, even more ambitiously...
3. create a `hugo import ...` command which essentially does similar things as the PHP code, but all written in Go?

battlemidget:
@anthonyfok i like 1 for the path of least resistance. most people are probably used to checking the website for importers similar to what jekyll displays on their site.

anthonyfok:
Oh, one more thing: Jekyll *does* have its own WordPress importer:

 * http://import.jekyllrb.com/docs/wordpress/

as featured in http://www.smashingmagazine.com/2014/08/01/build-blog-jekyll-github-pages/

Though even in Jekyll's case, it seems that using the excellent [WordPress to Jekyll Exporter](https://github.com/benbalter/wordpress-to-jekyll-exporter) by GitHub staff Ben Balter (@benbalter) **is** the recommended route.

[...]

PR #3559, opened 2017-06-04
New file import_wp.go for importing WordPress export files

eklausme:
As indicated in https://github.com/spf13/hugo/issues/100 this is the integration to Hugo from a stand-alone Go-program.

### DIFF
--- a/commands/import_wp.go
+++ b/commands/import_wp.go
@@ -1,0 +1,516 @@
+// Copyright 2016 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"bufio"
+	"fmt"
+	"github.com/spf13/cobra"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"unicode"
+)
+
+func init() {
+	importCmd.AddCommand(importWordpressCmd)
+}
+
+var importWordpressCmd = &cobra.Command{
+	Use:   "wordpress",
+	Short: "hugo import from WordPress",
+	Long: `hugo import from WordPress.
+
+Import from WordPress requires list of files, e.g. ` + "`hugo import wordpress export1.xml export2.xml`.",
+	RunE: importFromWordpress,
+}
+
+func importFromWordpress(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return newUserError(`Import from WordPress requires list of files, e.g. ` + "`hugo import wordpress export1.xml export2.xml`.")
+	}
+
+	// WordPress export may contain multiple files
+	for _, i := range args[0:] {
+		fmt.Println("\tWorking on file: ", i)
+		finp, err := os.Open(i) // for read access
+		if err != nil {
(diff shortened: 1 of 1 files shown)


bep:
@eklausme this looks extensive and the code looks solid. To add this as a command to Hugo, though, we need to know that it is 

1) Useful
2) Robust i.e. handles "most Word Press" versions without us having to tweak it all the time.

If I understand it, this importer handles XML export files. I remember doing a WP conversion myself at one time, and found that I lost lots of stuff doing it this way (custom shortcodes etc.), and went with a conversion that used the WP rendered output as the base for conversion.

@SchumacherFM have done some work in this are, and may share some thoughts?

SchumacherFM:
I get rarely feedback regarding the PHP WordPress to Hugo converter script. But AFAIK most people install it into their WP installations, hack on it, and then export all the pages and throw away the whole WP installation.

Maintaining the WP XML in Go within the Hugo binary can be a huge task.

Maybe and this is just an idea: Instead of implementing the import_wp.go into Hugo, you can just run it standalone and mention the importer besides my Wordpress PHP plugin. After some time you can see from the feedback from the community on how intensive the maintenance would be. If it runs well without lots of modifications ... add it to the Hugo binary.

eklausme:
User Feedback
---------------------

#### Previous Feedback

I collected various comments made during the last 4 years regarding issue #100, i.e., adding Hugo import feature from WordPress.

[spf13](https://github.com/spf13) commented on Oct 9, 2013: 'hugo import ....'

[anthonyfok](https://github.com/anthonyfok) commented on Jan 21, 2015: ...and had all of my 1800+ posts, images, etc. converted for use with Hugo in less than an hour. ... create a hugo import ... command which essentially does similar things as the PHP code, but all written in Go?

[bep](https://github.com/bep) commented on Jan 21, 2015: I couldn't get the "Wordpress to Hugo Exporter" to work. Issues with newer WP versions.

[cgsmith](https://github.com/cgsmith) commented on Apr 30, 2015: It would be nice to have it just parse the Wordpress XML ...

[eklausme](https://github.com/eklausme) (me) commented on Apr 24: I am very much interested in seeing this as hugo import wordpress.

To summarize the above: The creator of Hugo, spf13, wants an import feature; an approved  committer, anthonyfok, wants it written in Go and run quickly; bep seems to have some issues with the import facilities provided thus far.

The proposed pull request is a direct response to the above comments: Written in Go, being fast, and working directly on the export file without any need to install plugins in WordPress, which by the way is not possible with [wordpress.com](https://wordpress.com/).

#### Recent Feedback

I had written on this [import functionality](https://eklausmeier.wordpress.com/2017/04/24/converting-wordpress-export-file-to-hugo/) in April 2017, this blog post has now been viewed more than 100 times. I haven't received any comments so far. This could mean either:
1. People download my code and everything works just fine.
2. People recognize they have to install Go and compile a program on their own and are scared off.
3. None of the above, e.g., just serve reader's curiosity.


WordPress Version Problems
----------------------------------------
Regarding the version issue mentioned above by bep, I tested a WordPress export file from 2013 and 2015 which I still have in my backup directory. Both versions were imported to Hugo flawlessly. I checked the output of both with theme [Hugo Bootstrap v4 Blog](https://themes.gohugo.io/hugo-theme-bootstrap4-blog/). Blog from 2013 was converted in 0.1 seconds, blog from 2015 was converted in 1.1 seconds.

If bep could test his troublesome WordPress export file with this converter, or send me his problematic file, then I will take a look.


Proposal
------------
Following the initial proposal from spf13, it would lower the burden for the usual food blogger, travel blogger, parenting blogger, etc., to just have a single binary and import ones own stuff into Hugo and see results almost immediately and then play with content and themes, but not with technical specialities, like compiling, PATH variables, etc.

If there are still problems I am still around and eager to assist in case of any glitches.

bep:
>To summarize the above: The creator of Hugo, spf13, wants an import feature; an approved committer, anthonyfok, wants it written in Go and run quickly; bep seems to have some issues with the import facilities provided thus far.

If @spf13 chimes in with a "I will support/maintain this future", then we can consider.

PR #3574, opened 2017-06-09
Add Blackfriday 'joinLines' extension support.

choueric:


choueric:
Sorry for that, I will remember to attach some information next time. Thank you for your suggestion.

PR #3575, opened 2017-06-09
docs: in templates/terms, change .Site to $.Site to fix examples

mbresson:
In the documentation's templates -> terms page, changed **.Site** to **$.Site** because the global context was masked by the term's context created by **range**.
Otherwise, the example in the documentation cannot work.


digitalcraftsman:
Hello @mbresson,

thank you for submitting this addition.

Cheers,
Digitalcraftsman

PR #3662, opened 2017-06-29
Add log function for templates

artem-sidorenko:
It might be very useful for building nice tag clouds

Example:

Tag cloud built without log:

untitled

Tag cloud built with log:

untitled1

### DIFF
--- a/tpl/math/math.go
+++ b/tpl/math/math.go
@@ -15,7 +15,10 @@ package math
 
 import (
 	"errors"
+	"math"
 	"reflect"
+
+	"github.com/spf13/cast"
 )
 
 // New returns a new instance of the math-namespaced template functions.
@@ -32,6 +35,16 @@ func (ns *Namespace) Add(a, b interface{}) (interface{}, error) {
 
 func (ns *Namespace) Div(a, b interface{}) (interface{}, error) {
 	return DoArithmetic(a, b, '/')
+}
+
+func (ns *Namespace) Log(a interface{}) (float64, error) {
+	af, err := cast.ToFloat64E(a)
+
+	if err != nil {
+		return 0, errors.New("Log operator can't be used with non integer or float value")
+	}
+
+	return math.Log(af), nil
 }
 
 // Mod returns a % b.
--- a/docs/content/templates/functions.md
+++ b/docs/content/templates/functions.md
@@ -433,6 +433,12 @@ favicon.ico: {{.Width}} x {{.Height}}
 </tr>
 
 <tr>
+<td><code>math.Log</code></td>
+<td>Natural logarithm of one float.</td>
+<td><code>{{math.Log 1.0}}</code> → 0</td>
+</tr>
+
+<tr>
 <td><code>mod</code></td>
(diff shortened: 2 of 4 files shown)


digitalcraftsman:
Go's `math` package names the function for the natural logarithm (log_e) just `Log`. You use this name for the template func as well. However, users might find this name ambiguous because I would rather expect that `Log` returns log_10 instead of log_e.

Hence I would prefer to name the template function `ln` (for logarithm naturalis).

Let's take this a step further. log_e might be limiting in some cases and log_10 might be needed instead. While the `math` package has a function for this too it only adds just another function to the namespace.

Hence I suggest the addition (or even as a replacement) of a more generic logarithm template func where users can set the base as a parameter. The `math` package doesn't provide a function that takes the base as a parameter. 

But it's trivial to obtain the result in the wished by with log_10. Let's say I want to calculate log_2(8). This can be written as log_10(8) / log_10(2).

So implementing a generic function should be trivial.

tpl/math/math.go
```diff
@@ -34,6 +35,22 @@ func (ns *Namespace) Div(a, b interface{}) (interface{}, error) {
 	return DoArithmetic(a, b, '/')
 }
 
+func (ns *Namespace) Log(a interface{}) (float64, error) {
+	av := reflect.ValueOf(a)
+	var af float64
+
+	switch av.Kind() {
```
bep: Use `cast. ToFloat64`.

moorereason: Or `cast.ToFloat64E`

moorereason:
@digitalcraftsman,
The [Go math package](https://golang.org/pkg/math/) offers `Log`, `Log10`, `Log1p`, `Log2`, and `Logb`.  We should follow their lead as to what `Log` means.  I'm not in favor of adding all of these funcs to Hugo until someone needs them, though.

Also, I don't like `ln`.  That's too short (saves one character and loses most of its intuitiveness).  Makes me think of creating symlinks.  If I were searching for a logarithm func, I'd be looking for "log" not "ln."

**If** we put these in the global funcmap, I say we simply use the same names:  `log`, `log10`, `log1p`, `log2`, and `logb`.

Creating a single, multipurpose `log` func feels like over-engineering.

bep:
>I'd be looking for "log" not "ln."

Without any context, I would say that `log` does some debug logging or something. I would say that 

1. Follow the names from `math.Log` etc.
2. No aliases

digitalcraftsman:
> I'm not in favor of adding all of these funcs to Hugo until someone needs them, though.

Hence the generic func which is neither difficult to use nor to implement.

> Also, I don't like ln. That's too short (saves one character and loses most of its intuitiveness). Makes me think of creating symlinks. If I were searching for a logarithm func, I'd be looking for "log" not "ln."

Well, `ln` is the button you'll find on a calculator.

tpl/math/math.go
artem-sidorenko: @bep @moorereason  done, thanks

artem-sidorenko:
@moorereason done

PR #3666, opened 2017-06-30
perf(docker): radical reduce the image size

ellerbrock:
# Docker Image Size

Hello its me again :)
Its weekend which means time to clean up all the mess i done over the week :)
I radical cleaned up the Docker Image to end up with a smaller image size.

Cheers Maik

woop woop

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,33 @@
+# Developer: Maik Ellerbrock <opensource@frapsoft.com>
+#
+# GitHub:  https://github.com/ellerbrock
+# Twitter: https://twitter.com/frapsoft
+# Docker:  https://hub.docker.com/u/ellerbrock
+# Quay:    https://quay.io/user/ellerbrock
+
 FROM golang:alpine3.6
+
+LABEL maintainer "Maik Ellerbrock <opensource@frapsoft.com>"
 
 ENV GOPATH /go
 
 RUN \
   adduser -h /site -s /sbin/nologin -u 1000 -D hugo && \
-  apk add --no-cache dumb-init && \
-  apk add --no-cache --virtual .build-deps \
-    git \
-    make && \
-  go get github.com/kardianos/govendor && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,24 +4,21 @@ ENV GOPATH /go
...
-  apk add --no-cache dumb-init && \
-  apk add --no-cache --virtual .build-deps \
-    git \
-    make && \
-  go get github.com/kardianos/govendor && \
-  govendor get github.com/gohugoio/hugo && \
+  apk add --no-cache \
+    dumb-init \
+    git && \
+  go get github.com/gohugoio/hugo && \
```
bep: You cannot just "go get" Hugo due to the vendoring. It will never be correct. So while the Docker file may be smaller, it is now not correct.

bep:
@ellerbrock thanks for this, but we need to keep the "vendored version" of this Docker file.

ellerbrock:
hi @bep,

first at all thanks for your awesome work on hugo, nice to talk to you as person.
but i guess in this case you are wrong because i build and run this container already in production. you can see and test it here: 

- Docker Hub: https://hub.docker.com/r/ellerbrock/alpine-hugo/
- Dockerfile: https://hub.docker.com/r/ellerbrock/alpine-hugo/~/dockerfile/
- Build Details: https://hub.docker.com/r/ellerbrock/alpine-hugo/builds/b3jjfeaxv45l3vbee8xvj7x/

Run Example (to get hugo --help):

`docker run ellerbrock/alpine-hugo`

Cheers Maik

bep:
>Or do i miss something?

Yes. Without vendoring it _may work_, but you have no idea what library versions Hugo uses, so it may break or do the unexpected at any moment.

ellerbrock:
hi @bep,

i see, yeah thats should not be a problem ...
let me change that and test and then i push a updated version ...

PR #3668, opened 2017-07-01
docs: Update christianmendoza.me in the showcase

christianmendoza:
This PR updates the thumbnail and adds the source link for my site's entry in the showcase.


bep:

PR #3701, opened 2017-07-14
WIP: hugolib: Add section-based deep menu

miltador:
This adds new kind of 'autogenerated' menu based on content structure where the menu key is top-level section and all pages are hierarchically added reflecting directory structure.

This actually reduces amount of boilerplate metadata for articles. We have 1,5k+ pages and a majority of it goes under sectioned menu with different levels with `parent` articles. It's hard to maintain those menu`s `identifier` and `parent` fields. We want the menu to be folded reflecting directory structure of our content.
What content structure looks like: https://goo.gl/gZaUrr . The requirement is to have a `index` file for a folder which will be set as `parent` for items inside the folder.

I am not sure if I called it properly or what name would reflect the intent of it. This also probably needs some polishing and introducing tests to cover this so I flagged it as WIP.


bep:
The idea is good, but I think it can be made simpler. I don't think what we need is another "special thing".

Hugo has (at least) two tree structures:

1) Menus
2) Sections 

I'm not entirely sure why you'd want a `Menu` is you can just create the menu from the `Page` (plenty of examples doing that).

But, if there were good reasons for always having a menu (i.e. "one thing"), then I would much prefer some kind of adapter methods, i.e. `.Page.ToMenu()`

miltador:
I definitely missed out that `nested sections` feature with Hugo releases. Probably that together with [hugotest](https://github.com/bep/hugotest) example covers what this PR was intended to implement but in terms of `Page`s. Will take a closer look on that.

PR #3705, opened 2017-07-15
Add GOEXE to support building with different versions of `go`

mdhender:
Add a variable to the makefile and benchmark scripts to let users change the command used to build. Doesn't impact tools like govendor.

### DIFF
--- a/benchSite.sh
+++ b/benchSite.sh
@@ -1,9 +1,12 @@
 #!/bin/bash
 
+# allow user to override go executable by running as GOEXE=xxx make ...
+GOEXE="${GOEXE-go}"
+
 # Send in a regexp mathing the benchmarks you want to run, i.e. './benchSite.sh "YAML"'. 
 # Note the quotes, which will be needed for more complex expressions.
 # The above will run all variations, but only for front matter YAML.
 
 echo "Running with BenchmarkSiteBuilding/${1}"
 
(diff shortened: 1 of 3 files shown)


benchSite.sh
```diff
@@ -1,9 +1,12 @@
...
+
 # Send in a regexp mathing the benchmarks you want to run, i.e. './benchSite.sh "YAML"'. 
 # Note the quotes, which will be needed for more complex expressions.
 # The above will run all variations, but only for front matter YAML.
 
 echo "Running with BenchmarkSiteBuilding/${1}"
 
-go test -run="NONE" -bench="BenchmarkSiteBuilding/${1}$" -test.benchmem=true ./hugolib -memprofile mem.prof -cpuprofile cpu.prof
\ No newline at end of file
+"${GOEXE}" test -run="NONE" -bench="BenchmarkSiteBuilding/${1}$" -test.benchmem=true ./hugolib -memprofile mem.prof -cpuprofile cpu.prof
```
bep: An adde note not related to this commit (which looks really good): Running this script with a very vague expression (like YAML in the example) will run a whole lot of variations that may or may not complete. So it is recommended to pin it down a little. Run it with a "wide setting" until you understand the matching pattern, then kill it and restart it with something more specific.

PR #3714, opened 2017-07-18
helpers: Add --trace to asciidoctor args

miltador:
This will help to understand and fix errors by seeing stacktrace of an error.

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -573,7 +573,12 @@ func getAsciidocContent(ctx *RenderingContext) []byte {
 	}
 
 	jww.INFO.Println("Rendering", ctx.DocumentName, "with", path, "...")
-	cmd := exec.Command(path, "--no-header-footer", "--safe", "-")
+	cmd := exec.Command(path, "--no-header-footer", "--safe")
+	if strings.HasSuffix(path, "asciidoctor") {
+		// asciidoctor-specific arg to show stack traces on errors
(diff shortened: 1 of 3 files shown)


bep:
I'm by no means an expert here, but we do fallback to `asciidoc` if `asciidoctor` isn't available -- and I don't see that flag there.

miltador:
I think `asciidoc` will just ignore it. Or I can add a check to only add `--trace` for `asciidoctor`.

anthonyfok:
Hi @miltador,

> I think `asciidoc` will just ignore it.

```
$ asciidoc --version
asciidoc 8.6.9
$ asciidoc --trace
asciidoc: illegal command options
$ echo $?
1
```

So `asciidoc` does complain about an unknown command-line option and does exit with an error.

> Or I can add a check to only add `--trace` for `asciidoctor`.

Please do so.  :+1:  Thank you for your contribution!  :wink:

miltador:
@anthonyfok, thanks for the test! Added a check for `asciidoctor`.

miltador:
~Thanks for a test! Added a check.~

Will add a smarter check (WIP).

miltador:
Added smarter check for `asciidoctor`.
@anthonyfok it is ready for review now.

miltador:
@anthonyfok, switched the order to first check for Asciidoctor and then for Asciidoc.

PR #3722, opened 2017-07-20
fix typo error

yajiez:
Just a small typo error.


yajiez:
Thanks for your guidance!

PR #3735, opened 2017-07-25
Make SummayLength configurable

Shuenhoy:


Shuenhoy:
It seems that my code is out of date

helpers/content.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 // SummaryLength is the length of the summary that Hugo extracts from a content.
-var SummaryLength = 70
+var SummaryLength = 200
```
bep: Don't change the default value. That will break many, many sites.

hugolib/config.go
```diff
@@ -132,4 +133,5 @@ func loadDefaultSettingsFor(v *viper.Viper) {
 	v.SetDefault("enableGitInfo", false)
 	v.SetDefault("ignoreFiles", make([]string, 0))
 	v.SetDefault("disableAliases", false)
+	v.SetDefault("SummaryLength", 200)
```
bep: Follow the pattern ans start with lower char: SummaryLength > summaryLength

helpers/content.go
bep: As this is moved to a config variable, this variable should be removed.

hugolib/config.go
bep: As noted above: The default above should remain at 70 (changing that will break many many sites).

hugolib/page.go
```diff
@@ -611,9 +611,9 @@ func (p *Page) setAutoSummary() error {
 	var summary string
 	var truncated bool
 	if p.isCJKLanguage {
-		summary, truncated = helpers.TruncateWordsByRune(p.PlainWords(), helpers.SummaryLength)
+		summary, truncated = helpers.TruncateWordsByRune(p.PlainWords() , p.s.Cfg.GetInt("SummaryLength"))
```
bep: summaryLength

bep:
And this needs a line somewhere in the documentation (just add it to this PR).

moorereason:
Replaced by #3924

Issue #3730, opened 2017-07-23
Reduce Docker image size further using "multi-stage builds" in Docker CE 17.06


anthonyfok:
Thanks to @ellerbrock's optimization effort in #3666 and #3674, the Hugo Docker image generated using the Dockerfile was greatly reduced from 428MB to 277MB, but it still looks a bit big.

It turns out that about 250MB is taken up by /usr/local/go, which is needed for building Hugo, but not needed for running Hugo.  So, that got me wondering: Could we build the Hugo Docker image with golang:alpine3.6 (258 MB) like we do now, but use the much smaller alpine:latest (3.96 MB) for the running image?

```
$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
golang              alpine3.6           9f687d7d2a96        3 weeks ago         258 MB
alpine              latest              7328f6f8b418        3 weeks ago         3.96 MB
```

That way, perhaps future Hugo Docker images could be smaller than 25 MB!

It turns out that it was not a trivial thing to do before, but with the hot new Docker CE 17.06 (released 2017-06-28) and its "Multi-stage Builds" feature, it looks like this goal is within reach:

* [Even Smaller Docker Image Sizes](https://blog.replicated.com/optimizing-docker-images-with-multi-stage-builds/)
* [Announcing Docker 17.06 Community Edition (CE) - Docker Blog](https://blog.docker.com/2017/06/announcing-docker-17-06-community-edition-ce/)

@ellerbrock, could you please look into this when you are free?  (No hurries though, as I am sure it will take a while for Docker CE 17.06 to propagate, e.g. Debian is still at Docker 1.13.1 released in February 2017.)  Many thanks!

ellerbrock:
Hi @anthonyfok, good idea.
There are a few ways how we could archive this:

1.) In most cases i build and cleanup inside the container. In that case its isolated in the docker container without any outside dependencies. The important part here is to install and cleanup in one RUN command to avoid hitting the docker cache layer to keep the container size small.

2.) Another solution is to setup a proper CI pipeline (https://docs.travis-ci.com/user/languages/go/) and just copy the executable inside the container. I would only use such a solution if all steps are transparent to ensure that no malicious stuff finds a way inside the container and my network (with [Travis](https://travis-ci.org/) which you use anyway this is possible).

[...]

PR #3738, opened 2017-07-25
perf(dockerfile): xtrem awesome container size optimization

ellerbrock:
Hi @anthonyfok,

i just ask myself:

cool or what?

Few days before we talked about `size matters` and how to optimize it in #3730.
We started with 428M and optimized it to 277M.
What would you say to make it even a little bit smaller and give you a 27M Container without waiting for any Docker releases and just do it the good old way?

hugo

cheers maik

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,9 +8,9 @@ RUN \
   apk add --no-cache \
     dumb-init && \
   apk add --no-cache --virtual .build-deps \
-	gcc \
-	musl-dev \
-	go \
+    gcc \
+    musl-dev \
+    go \
     git && \
   mkdir -p \
     ${GOPATH}/bin \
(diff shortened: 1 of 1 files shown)


ellerbrock:
hi @anthonyfok,

#9ed48c1 thats the spirit!
you really squeezed the last bits out for the optimization - great!

hey what do you guys think about making a official repo for hugo?
would be nice to have some official hugo container on dockerhub.
like mentioned before i could help with the setup and maintain the docker things.
to be able to start doing that i guess i need at least be part of the team with read permission for the repo to create a automatic build. i have on my local machine quite some other useful things hacked together with `docker-compose.yml` which could be helpful for others as well (for example bring hugo server up, automatic compile sass with a super simple setup ...), so maybe create a docker repo for hugo with useful starter could be an option as well.
i maintain the [Bash-it](https://github.com/Bash-it) docker repo and there we moved the docker stuff to its own repo for example ...

but anyway, its up to you guys.
you can discuss about if you like the idea or not and ping me when you came up with a solution ...

cheers maiky

PR #3739, opened 2017-07-25
Replace http://example.org/ with https://example.org/

### LINKED ISSUE #3728: Replace http://example.org/ with https://example.org/ in examples etc.
We may as well use the HTTPS version in all instances that are facing the public; i.e. in documentation and in the "new site" template.

Note that document changes should live in https://github.com/gohugoio/hugoDocs -- but I'll keep this as one issue.

mpcabd:
As per issue #3728 @bep suggests changing all instances of
http://(www.)?example.com to https://(www.)?example.com

Fixes #3728


bep:
@mpcabd thanks for this; I have updated the issue to be more specific. I never meant a full "search and replace" in this repo, but that is what happens when I create quick issues so I don't forget.

mpcabd:
I should have asked for more details when I suspected it yesterday, I thought that it's probably not just a replace-all task 😅 I was looking for something quick to finish before I go to sleep.

I'll fix it again as you described today @bep.

digitalcraftsman:
@mpcabd has become Schrödinger's cat for CLA agreements. The CLA bot shows on the two comments above that the CLA was signed and **not** signed.

digitalcraftsman:
@bep are the replacements now as intended?

Issue #3514, opened 2017-05-23
RFE: add --debug option to hugo server


vassudanagunta:
It would be nice to have a hugo server mode that helps you figure why a page isn't rendering the way you expected. 

For example, the [Single Page Template Lookup Order](https://hugodocs.info/templates/single-page-templates/#single-page-template-lookup-order) is not trivial, and results can differ from expectations because a theme may or may not define a template, and so forth. It would be tremendously helpful if the path to the template used were printed at the bottom of each page. 

`--debug` output can be enhanced over time, but for starters it would useful to have the following debug info appended to the bottom of every page served:

- the path to the source content file (`.File.Path`)
- the path to the layout template used
- any [Page Variables](https://hugodocs.info/variables/page/#page-variables) useful for debugging 
- all values stored in Scratch (`.Scratch`)

vassudanagunta:
Once I get the time to learn enough Go to implement this, I'll give it a stab if no one else has.

PR #3741, opened 2017-07-27
add --debug option to be improved on over time

maxandersen:
Why:

 * first time using hugo I got very little info from --verbose output
   but I noticed there is quite a lot of useful DEBUG logging
 * asked for in other issues like https://github.com/gohugoio/hugo/issues/3514

This change addreses the need by:

 * adding a simple --debug flag which simply turns on debug level in stdout
   and logoutput if enabled.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -143,6 +143,7 @@ var (
 	renderToMemory bool // for benchmark testing
 	verbose        bool
 	verboseLog     bool
+	debug          bool
 	quiet          bool
 )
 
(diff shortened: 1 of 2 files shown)


maxandersen:
so i've signed the CLA but it still wont let me pass ;/

digitalcraftsman:
Hello @maxandersen,

first of all thanks for your contribution. I've seen another issue with the same problem. Please click on the recheck button in the CLA comment above if you haven't already done this.

I browsed the issues in the repo for the CLA assistant. Is it possible that you use two GitHub accounts? A problem described in one issue was that the submitter used account A to sign the CLA sometime ago but used account B to submit a pull request.

bep:
@digitalcraftsman I think I know what the problem with the CLA is, I will check.

maxandersen:
okey, seems CLA is now only showing up once and is all good - now it says "appveyor was unable to build non-m..." and I can't see the details on that to know what is failing ? any ideas ?

maxandersen:
btw. looking some more I'm starting to think it might make more sense to add a ``--verboselevel string  set level of verbose logging level. Possible values are TRACE, DEBUG, INFO, WARN, ERROR, CRITICAL, FATAL``

Suggesting this as i've found both trace and debug levels in the source code.

WDYT ?

Issue #3744, opened 2017-07-27
tpl: TestMethodToName fails with gccgo


anthonyfok:
With gccgo-7:

```
$ go-7 version
go version go1.8.1 gccgo (Debian 7.1.0-10) 7.1.0 linux/amd64
```

`go-7 test` produces the following error:

```
=== RUN   TestMethodToName
--- FAIL: TestMethodToName (0.00s)
        Error Trace:    templatefuncRegistry_test.go:32
	Error:		Not equal: "MyTestMethod" (expected)
			        != "$thunk0" (actual)
		
FAIL
FAIL	github.com/gohugoio/hugo/tpl/internal	0.916s
```

Is it safe to just skip the test on gccgo with `// +build !gccgo` in tpl/internal/templatefuncRegistry_test.go (which contains only one test, TestMethodToName())?

Or is this a concern that may cause a gccgo-compiled Hugo to fail with certain templates?

Many thanks!

References:

* https://github.com/golang/go/issues/3901
* https://bugs.launchpad.net/gocheck/+bug/1250253

bep:
>Or is this a concern that may cause a gccgo-compiled Hugo to fail with certain templates?

No, then there would be more template func test failures. I don't remember the code path here, but I expect there to be a symmetry here, so `$thunk0` is an identifier good as any.

anthonyfok:
>Or is this a concern that may cause a gccgo-compiled Hugo to fail with certain templates?
>> No, then there would be more template func test failures. I don't remember the code path here, but I expect there to be a symmetry here, so `$thunk0` is an identifier good as any.

Thank you for your reassurance!  I'll file a PR to accommodate gccgo in TestMethodToName then.

PR #3745, opened 2017-07-27
releaser: Add mips{,64}{,le}, ppc64{,le} and s390x archs

anthonyfok:
To my pleasant surprise, cgo is amazing at cross-compiling even for architectures that it does not yet provide an official compiler for, especially for MIPS, and the resulting `hugo` binaries really do work!

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -15,6 +15,13 @@ build:
     - 386
     - arm
     - arm64
+    - mips
+    - mipsle
+    - mips64
+    - mips64le
+    - ppc64
+    - ppc64le
+    - s390x
   ignore:
     - goos: openbsd
       goarch: arm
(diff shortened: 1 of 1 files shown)


bep:
This would add too much to build time, build complexity and maintenance -- we're not adding this just because we can. Also, we don't have CI test setups for these.

anthonyfok:
> This would add too much to build time, build complexity and maintenance -- we're not adding this just because we can. Also, we don't have CI test setups for these.

* Build time and build complexity: no so much, since these 7 architectures are valid only on Linux, so there are just 7 more to build, rather than 14 or 21 or 28.
* Maintenance and lack of CI test setup: I guess we could tell the users that these are provided for the end user's convenience but not officially supported?  :-)

But then, it came to the real deal breaker: Is there any user demand for these extra architectures?  Right now, I guess not, as I haven't noticed anyone asking for them on GitHub or on Discourse, so, until someone actually wants them, this issue is moot.

Though I guess, if there is any demand, it could be partially fulfilled by the cross-platform support provided by Debian and the likes.  Not quite there yet, as seen on 
https://buildd.debian.org/status/package.php?p=hugo&suite=sid, but I think I finally figured out how to solve the build failures on Debian mips, mipsel, mips64el and s390x, see #3744 and golang/go#18031.  Hope to finally to get them to build correctly soon!

PR #3749, opened 2017-07-29
Task list at beginning of markdown now renders properly

### LINKED ISSUE #3710: Does not render task-list when it's the first token
The following page:

```md
---
title: "Test"
date: 2017-07-17T12:07:51+01:00
draft: false
---

* [ ] TaskList
```

Outputs the following HTML

```html
<ul> class="task-list"
<li><input type="checkbox" disabled class="task-list-item"> TaskList</li>
</ul>
```

Which renders like this

Task List

---

Steps to fix - add any leading content before the task-list

```md
---
title: "Test"
date: 2017-07-17T12:07:51+01:00
draft: false
---

leading content

* [ ] TaskList
```

danieka:
Fixes #3710

The problem was that when the task list is at the beginning of the output from blackFriday does not begin with a newline as this code expected. So I added code so that if there is a newline we forward past that before doing our operation.

This is the first time working on Hugo so please point out if this bug should be fixed somewhere else in the code.

### DIFF
--- a/helpers/content_renderer.go
+++ b/helpers/content_renderer.go
@@ -100,11 +100,16 @@ func (r *HugoHTMLRenderer) List(out *bytes.Buffer, text func() bool, flags int) 
 	r.Renderer.List(out, text, flags)
 	if out.Len() > marker {
 		list := out.Bytes()[marker:]
+		if list[0] == 10 {
+			out.Write(list[:1])
+			list = list[1:]
+			marker++
+		}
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this contribution. We have now to solutions for the same problem, and I have looked at both and both do the job, but #3750 looks slightly better. Mostly a matter of taste.

danieka:
I agree, I prefer that solution as well.

PR #3752, opened 2017-07-30
calling isSet on array with key type string no longer panics but retu…

### LINKED ISSUE #3681: isset should return error on integer slices when passed a string value
Just updated hugo from 0.19 to 0.25 via "brew upgrade hugo".  Ran hugo and got
```scsmac081:lpr-geeks bwbecker$ hugo
Started building sites ...
WARNING: calling IsSet with unsupported type "invalid" (<nil>) will always return false.

panic: reflect: call of reflect.Value.Int on string Value [recovered]
	panic: reflect: call of reflect.Value.Int on string Value [recovered]
	panic: reflect: call of reflect.Value.Int on string Value [recovered]
	panic: reflect: call of reflect.Value.Int on string Value

goroutine 133 [running]:
text/template.errRecover(0xc4205cb830)
	/usr/local/Cellar/go/1.8.3/libexec/src/text/template/exec.go:146 +0x62
panic(0x1710e80, 0xc4203649a0)
	/usr/local/Cellar/go/1.8.3/libexec/src/runtime/panic.go:489 +0x2cf
text/template.errRecover(0xc4205cab38)
	/usr/local/Cellar/go/1.8.3/libexec/src/text/template/exec.go:146 +0x62
panic(0x1710e80, 0xc4203649a0)
	/usr/local/Cellar/go/1.8.3/libexec/src/runtime/panic.go:489 +0x2cf
```
etc.

Uninstalled hugo  (brew uninstall --force hugo), reinstalled, with the same result.

MacBook Pro running MacOS 10.12.5.

danieka:
…rns false

Fixes #3681 

Is this the kind of solution your looking for? I assumed that all kinds of Int should be accepted.

### DIFF
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -313,7 +313,7 @@ func (ns *Namespace) IsSet(a interface{}, key interface{}) (bool, error) {
 
 	switch av.Kind() {
 	case reflect.Array, reflect.Chan, reflect.Slice:
-		if kv.Kind() != reflect.Int && kv.Kind() != reflect.Int8 && kv.Kind() != reflect.Int16 && kv.Kind() != reflect.Int32 && kv.Kind() != reflect.Int64 {
+		if !isInt(kv.Kind()) && !isUint(kv.Kind()) {
 			helpers.DistinctFeedbackLog.Printf("WARNING: calling IsSet with with key of type %q will always return false for container of type %q (%T).\n", kv.Kind(), av.Kind(), a)
 			return false, nil
 		}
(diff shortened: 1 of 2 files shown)


bep:
@moorereason knows this better than I do, but there is a `isNumeric` etc. func in there somewhere, which is cleaner than this massive and repeated switch statements.

danieka:
Since floats would also panic I checked against isInt and isUint instead.

kaushalmodi:
@moorereason Looks like this PR was approved but never merged?

PR #3756, opened 2017-07-31
Remove unused dependencies from vendor.json

jorinvo:
I realized that some dependencies are not in use anymore. To see which dependencies are still in use I ran [`dep init`](https://github.com/golang/dep) and compared [the results](https://github.com/gohugoio/hugo/files/1187802/Gopkg.toml.txt) to the `vendor.json` file.
All unused dependencies are removed now. All `make` commands are still working.
Hope I'm not missing anything.

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -166,12 +166,6 @@
 			"revisionTime": "2017-05-10T13:15:34Z"
 		},
 		{
-			"checksumSHA1": "KQhA4EQp4Ldwj9nJZnEURlE6aQw=",
-			"path": "github.com/kr/fs",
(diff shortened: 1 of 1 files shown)


bep:
Thanks for this, this is very useful.But I assume the `dep init` only uses the `build` tag for the system you are running on, and will miss out on `windows` only deps etc. (see failing build).

jorinvo:
Was a bit naive to think more than half of the dependencies are unused 😄 
Didn't include the third-party dependencies. Also, other than `govendor`, `dep` only references the root and not single packages (e.g. `golang.org/x/text`, not `golang.org/x/text/runes` and `golang.org/x/text/languages` separately).
There are still 14 dependencies that can be removed with tests passing.

PR #3759, opened 2017-08-01
Switch from fork bep/inflect to markbates/inflect

jorinvo:
Original package has received updates the fork hasn't.
Without fork updates are easier to maintain.

Not sure if I'm missing a previous reason why the project was forked.

### DIFF
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -32,7 +32,7 @@ import (
 
 	"github.com/gohugoio/hugo/media"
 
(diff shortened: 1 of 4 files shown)


bep:
Both are forks from the same Bitbucket repo; his fork did not exist when I made my fork. But I'm happy that someone is actually picking up the thread and does some actual maintenance.

bep:
Or: I guess his fork did exist, but it was pretty far behind the original -- not it isn't.

PR #3762, opened 2017-08-01
Better more complete fix for allowing filename to set date and slug

maxandersen:
This is a better and more complete PR for https://github.com/gohugoio/hugo/pull/3310 which:

- has actual working tests :)
- uses a single boolean to decide if should enable filename as fallback or not
   (this allows us to add if needed more and more complex parsing of the header rather than lock it down to user specified regex)
- also sets the slug so the date is not part of the url


hugolib/page.go
```diff
@@ -1070,13 +1070,44 @@ func (p *Page) update(f interface{}) error {
...
+
 		fi, err := p.s.Fs.Source.Stat(filepath.Join(p.s.PathSpec.AbsPathify(p.s.Cfg.GetString("contentDir")), p.File.Path()))
 		if err == nil {
+			p.s.Log.DEBUG.Printf("using file modification time as fallback for page %s", p.File.Path())
 			p.Date = fi.ModTime()
 			p.Params["date"] = p.Date
 		}
 	}
 
+	if p.Date.IsZero() && p.s.Cfg.GetString("useFilenameDateAsFallback") != "" {
```
bep: * useFilenameDateAsFallback is a bool (not a string)
* also, I think it should default to `true`

hugolib/page.go
```diff
@@ -1070,13 +1070,44 @@ func (p *Page) update(f interface{}) error {
...
+		dateExp := regexp.MustCompile("(?P<year>\\d{4})\\-(?P<month>\\d{2})\\-(?P<day>\\d{2})-(?P<slug>.*)\\..*")
```
bep: This regexp looks reasonable slow to compile so it should be pulled out into a variable and reused.

hugolib/page.go
maxandersen: fixed use of bool and I changed the default to `true` including updating the docs/commit messages to match.

hugolib/page.go
maxandersen: fixed.

maxandersen:
I've updated with a fix that ensures even if post has no frontmatter dates gets calculated correctly.

Anything else I need to do here  to get it merged in ?

maxandersen:
now rebased against master.

hugolib/page.go
```diff
@@ -1348,10 +1388,11 @@ func (p *Page) parse(reader io.Reader) error {
 	}
 
 	if meta != nil {
-		if err = p.update(meta); err != nil {
+		if err = p.readMetadata(meta); err != nil {
 			return err
 		}
 	}
+	p.updateMetadata()
```
maxandersen: this is added so when you render pages that are missing frontmatter they don't get more broken than they need to be. Took me ages to figure out that title/slug and lastmod changes were not handled unless I had an explicit title in frontmatter. With this change it at least still sorts and gets rendered somewhat properly.

maxandersen:
rebased again - hope this can go in soon :)

maxandersen:
this is still relevant - just awaiting for update on it from dev.

vassudanagunta:
This takes away the flexibility of @devjack's PR #3310, hardcoding the filename pattern. I think that's the wrong direction to go. I'm going to offer Yet Another Take when I get the chance in the next few of days. 

BTW, the four component commits in this PR are strange. The first only has the second half of @devjack's PR #3310, the third is attributed to @bep and has a diff of +38 −14,450, and the final attributed to @maxandersen has a diff of +14,574 −14. :open_mouth:

maxandersen:
the flexibility about customizing pattern was taken away to make the initial feature simpler and avoid overengineering it.

bep:
>This takes away the flexibility of @devjack's PR #3310, hardcoding the filename pattern. 

This only works if it is simple. Both to document/explain and to maintain/test and it makes your content as portable as possible. So one filename pattern it is.

I'm sorry for neglecting this issue for so long. I don't need it myself, but I understand that many others are. I have added the tracking issue to the 0.35 milestone. Let's see if we can agree on a solution by then.

PR #3764, opened 2017-08-02
Added template function 'asciidocify'

sluedecke:
In order to use AsciiDoc in shortcodes and other template places, I have added 'asciidocify' as a new template function similar to 'markdownify'

### DIFF
--- a/tpl/transform/transform_test.go
+++ b/tpl/transform/transform_test.go
@@ -168,6 +168,33 @@ func TestMarkdownify(t *testing.T) {
 	}
 }
 
+func TestAsciidocify(t *testing.T) {
+	t.Parallel()
+
+	ns := New(newDeps(viper.New()))
+
+	for i, test := range []struct {
+		s      interface{}
+		expect interface{}
+	}{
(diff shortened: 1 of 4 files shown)


tpl/transform/transform_test.go
```diff
@@ -168,6 +168,33 @@ func TestMarkdownify(t *testing.T) {
...
+		s      interface{}
+		expect interface{}
+	}{
+		{"Hello *World!*", template.HTML("Hello <strong>World!</strong>")},
+		{[]byte("Hello **Bytes World!**"), template.HTML("Hello <strong>Bytes World!</strong>")},
+		{tstNoStringer{}, false},
+	} {
+		errMsg := fmt.Sprintf("[%d] %s", i, test.s)
+
+		result, err := ns.Markdownify(test.s)
```
kevingimbel: Your test is using Markdownify instead of Asciidocify. I guess this should be adjusted to validate actual AsciiDoc code instead.

bep:
This has been discussed before, see my comments in this thread: https://github.com/gohugoio/hugo/pull/2387

I appreciate the effort, but we're not adding this now.

PR #3794, opened 2017-08-11
tpl: Add strings.TrimLeft and TrimRight

moorereason:


moorereason:
See https://discourse.gohugo.io/t/7915 for rationale.

tpl/strings/init.go
```diff
@@ -112,6 +112,34 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.TrimLeft,
+			nil,
+			[][2]string{
+				{`{{ strings.TrimLeft "++Batman++" "+" }}`, `Batman++`},
```
bep: It would be better (in my eyes) if all examples (where possible) were on "pipe form", i.e. 

```
{`{{ "++Batman++" | strings.TrimLeft  "+" }}`, `Batman++`},
```

moorereason:
@bep, the same thought crossed my mind after I pushed up this PR.  I was trying to match TrimPrefix and TrimSuffix, which was trying to match Trim.

I don't know if we want to go about changing Trim at this point, but I've rebased this PR to only include TrimLeft and TrimRight.  We can fix TrimPrefix and TrimSuffix in a separate PR.

tpl/strings/init.go
```diff
@@ -112,6 +112,20 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.TrimLeft,
+			nil,
+			[][2]string{
+				{`{{ "++Batman++" | strings.TrimLeft "+" }}`, `Batman&#43;&#43;`},
```
bep: With the HTML entity encoding thing happening, I would also consider picking another example. It may look fine when rendered somewhere, but it is confusing to look at in the code.

PR #3799, opened 2017-08-13
use shimmed approach to translating rst and asciidoc

tychoish:
This revives #2672, which I dropped the ball on, but am still interested in getting merged at some point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,4 @@ script:
   - ./hugo -s docs/
   - ./hugo --renderToMemory -s docs/
 before_install:
-  # gem install must be run with sudo on OSX
-  - sudo gem install asciidoctor | gem install asciidoctor
+  - sudo pip install docutils asciidoc flask 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,9 +9,7 @@ clone_folder: C:\GOPATH\src\github.com\gohugoio\hugo
 
 install:
   - git submodule update --init --recursive
-  - gem install asciidoctor
-  - pip install docutils
-
+  - pip install docutils flask asciidoc
 build_script:
   - make hugo-race check
   - hugo -s docs/
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -50,6 +50,7 @@ import (
 	jww "github.com/spf13/jwalterweatherman"
 	"github.com/spf13/nitro"
 	"github.com/spf13/viper"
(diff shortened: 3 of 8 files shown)


miltador:
Did you do some benchmarks on how it speeds up build time?
I am really interested in seeing this PR "in action". Cause in our project we have ~1,5k asciidoc files and our build time is about 2 minutes.
Anyway I will test it out in spare time, but it looks promising.

tychoish:
It takes a 2-3 minute build time for me (restructured text, of comparable size), and gets it down to under a minute.

tychoish:
I've resubmitted the pull request with a fix for the tests. 

I ran tests locally using the configuration from the .travis file (on linux) and wasn't able to reproduce the test failure. Is there something I should be doing to be able to reproduce this locally?

miltador:
I am not sure but maybe tests fail cause you removed installation of `asciidoctor` gem? It looks like `asciidoctor` doesn't process the input correctly, `div`s which it usually adds are gone.

bep:
```
<div class=\"paragraph\">
```

There are some "summary split logic" that considers the output of Asciidoctor (and RST). So, if the output has changed, for some reason, that has to be fixed. As in: We need to understand what's happening and make sure that the Asciidoc test is actually run and asserted OK.

tychoish:
I've been able to root out most of these cases with the extra div tags that are no longer being added, but not being able to run the tests locally means I have to keep repushing to this branch to let travis run the tests, as they all pass on my machine. 

As of this build, we won't be using asciidoctor, which I have removed from the travis dependency list and do not have installed locally. 

I'm making progress because it does not fail on appveyor anymore.

miltador:
Oh... now I've got this and what `shimgo` does too.
Unfortunately this doesn't suit for me cause `asciidoc` python implementation seems to be stale and `asciidoctor` has pushed it far further with a lot of advanced features we use in our project.
Removing `asciidoctor` support is what I am against of.

tychoish:
I think it would be reasonable to extend the shimgo package to handle asciidoctor in a similar fashion.

The cost of the old approach is in process creation, not in parsing so it'll work. If you want to add support in shimgo to handle a simple sinatra (or similar) ruby app with asciidoctor, it should work fine.

bep:
Can someone clean up and close PRs that is not active anymore? This or #3842 ??

miltador:
This one can be closed, #3842 is active.

PR #3803, opened 2017-08-14
tpl: Cleanup strings.TrimPrefix and TrimSuffix

moorereason:
These funcs were added during the move to namespaces but were
undocumented.  This commit fixes the order of the arguments and adds the
funcs to the method mapping.

### DIFF
--- a/tpl/strings/init.go
+++ b/tpl/strings/init.go
@@ -119,10 +119,26 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.TrimPrefix,
+			nil,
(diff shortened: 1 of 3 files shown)


tpl/strings/init.go
```diff
@@ -112,6 +112,22 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.TrimPrefix,
+			[]string{"trimPrefix"},
```
bep: I don't think these funcs reach the "need an alias" bar.

PR #3812, opened 2017-08-18
metrics: Add simple template metrics feature

moorereason:
Tracks regular templates, shortcodes, partials, aliases, etc.  I don't think I missed anything.

Sample output:

```none
$ hugo --templateMetrics
Started building sites ...

Template Metrics:

     cumulative       average       maximum         
       duration      duration      duration  count  template
     ----------      --------      --------  -----  --------
      254.221µs      36.317µs     111.885µs      7  alias
       512.42µs      512.42µs      512.42µs      1  404.html
      791.222µs     791.222µs     791.222µs      1  taxonomy/category.terms.html
     5.683547ms    5.683547ms    5.683547ms      1  taxonomy/tag.terms.html
    13.754115ms     118.569µs     275.939µs    116  _internal/shortcodes/highlight.html
    27.061924ms   13.530962ms    26.18361ms      2  theme/fixed/single.html
    59.547255ms      41.152µs    4.930039ms   1447  _internal/shortcodes/relref.html
    61.138274ms   61.138274ms   61.138274ms      1  index.html
   101.781005ms      47.942µs   27.666183ms   2123  theme/partials/default_foot.html
   103.187682ms  103.187682ms  103.187682ms      1  _internal/_default/sitemap.xml
   269.729605ms     19.2664ms   70.802937ms     14  taxonomy/category.html
   329.442505ms  329.442505ms  329.442505ms      1  fixed/archives.html
   356.977108ms  356.977108ms  356.977108ms      1  indexes/post.html
   523.095035ms    2.254719ms   44.263001ms    232  theme/rss.xml
   562.297122ms     264.859µs   54.887203ms   2123  partials/sidebar.html
    760.71056ms     358.318µs    32.50068ms   2123  partials/head.html
   1.636059596s     770.635µs   55.143587ms   2123  partials/default_head.html
   1.911769186s    8.809996ms  142.008155ms    217  taxonomy/tag.html
  59.028253864s   31.331344ms  174.884436ms   1884  post/single.html

Built site for language en:
0 draft content
0 future content
0 expired content
1887 regular pages created
468 other pages created
0 non-page files copied
0 paginator pages created
217 tags created
14 categories created
total in 19216 ms
```

Things I don't like:

- metrics.Store.LogFeedback(): I wish `jww.Feedback` implemented `io.Writer` (but it's really a logger).  Or better yet, stdlib `log.Logger` should be an interface.
- Don't like the repeated calls to `.GetBool("templateMetrics")`, but the templates don't have access to the Site Info.

I intentionally kept this simple so that it's easy to reason about and should be simple to merge.  Also, I built this to specifically track template execution timings.  If there are dreams of doing more general purpose metrics for other things, we'll have to evolve the metrics package to handle that.

**Update (2017.08.18)**: Updated the sample output to reflect sorting by sum.


tpl/partials/partials.go
```diff
@@ -72,6 +73,10 @@ func (ns *Namespace) Include(name string, contextList ...interface{}) (interface
 			b := bp.GetBuffer()
 			defer bp.PutBuffer(b)
 
+			if ns.deps.Cfg.GetBool("templateMetrics") {
+				defer ns.deps.Metrics.MeasureSince(templ.Name(), time.Now())
+			}
+
 			if err := templ.Execute(b, context); err != nil {
```
bep: It looks to me that every `MeasureSince` is wrapping a call to `templ.Execute` (or similar). `templ` is a Hugo interface, and we should put the `MeasureSince` directly in the `Execute` method (there are some embedding going on there currently, but that is easy to work around with some delegation. This will not only be much cleaner, but it will also solve the `GetBool` issue, as you can set that as a config once when you create the template harness.

metrics/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+	w.Printf("  %13s  %12s  %12s  %5s  %s\n", "----------", "--------", "--------", "-----", "--------")
+
+	s.mtx.Lock()
+
+	// sort keys
+	var tkeys []string
+	for k := range s.metrics {
+		tkeys = append(tkeys, k)
+	}
+	sort.Strings(tkeys)
```
bep: I suspect it would be better for most use cases (find bottle necks) to sort by the cumulative duration.

tpl/partials/partials.go
moorereason: I agree, but we're using `Execute` from external packages (like `html/template`).  Can you help build out the template handlers so that we can intercept the `Execute` calls?

bep: I will make a commit tomorrow that will make what I think about more clear.

metrics/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+// MeasureSince adds a measurement for key to the metric store.
+func (s *Store) MeasureSince(key string, start time.Time) {
+	s.mtx.Lock()
+	s.metrics[key] = append(s.metrics[key], time.Since(start))
+	s.mtx.Unlock()
+}
+
+// LogFeedback prints metrics in a pretty format to a jwalterweatherman.Feedback.
+func (s *Store) LogFeedback(w *jww.Feedback) {
+	w.Printf("  %13s  %12s  %12s  %5s  %s\n", "cumulative", "average", "maximum", "", "")
```
moorereason: Sorting the results by sum helped the readability quite a bit.  Plus, I'm using the `time.Duration` Stringer interface to print these values, so I'd rather not reinvent the wheel.  I think what we have is good enough.

metrics/metrics.go
moorereason: Done.

hugolib/site.go
```diff
@@ -1705,6 +1705,12 @@ func (s *Site) appendThemeTemplates(in []string) []string {
 // Stats prints Hugo builds stats to the console.
 // This is what you see after a successful hugo build.
 func (s *Site) Stats() {
+	if s.Cfg.GetBool("templateMetrics") {
+		s.Log.FEEDBACK.Println()
+		s.Log.FEEDBACK.Printf("Template Metrics:\n\n")
+		s.Metrics.LogFeedback(s.Log.FEEDBACK)
```
moorereason: Done.

moorereason:
Not that important, but running this against the hugoDocs site gives me full paths to some templates:

```none
    34.704285ms     294.104µs   20.562918ms    118  shortcodes/note.html
   110.567168ms     250.151µs   17.011391ms    442  theme/partials/boxes-section-summaries.html
   147.235229ms     568.475µs   17.763642ms    259  theme/partials/nav-links-docs-mobile.html
   157.586418ms    7.163019ms   21.833037ms     22  /home/foo/dev/src/github.com/gohugoio/hugoDocs/themes/gohugoioTheme/layouts/news/single.html
   205.734511ms     794.341µs   21.616462ms    259  theme/partials/site-nav.html
   213.237464ms   15.231247ms   58.033026ms     14  /home/foo/dev/src/github.com/gohugoio/hugoDocs/themes/gohugoioTheme/layouts/_default/list.html
   217.907486ms   13.619217ms   39.875413ms     16  /home/foo/dev/src/github.com/gohugoio/hugoDocs/themes/gohugoioTheme/layouts/_default/taxonomy.html
   234.479311ms     905.325µs   18.575748ms    259  theme/partials/nav-mobile.html
    307.36153ms    9.914888ms   43.728581ms     31  theme/partials/pagelayout.html
   497.119896ms    1.919381ms    43.52786ms    259  theme/partials/site-footer.html
   1.734563418s    7.412664ms   34.629905ms    234  theme/partials/nav-links-docs.html
   2.696173466s   13.347393ms   58.012074ms    202  /home/foo/dev/src/github.com/gohugoio/hugoDocs/themes/gohugoioTheme/layouts/_default/single.html
```

bep:
@moorereason see https://github.com/gohugoio/hugo/commit/d000cf605091c6999b72d6c632752289bc680223

metrics/metrics.go
```diff
@@ -0,0 +1,111 @@
...
+type MetricsProvider interface {
+	MeasureSince(key string, start time.Time)
+	WriteMetrics(w io.Writer)
+	Reset()
+}
+
+// Store provides storage for a set of metrics.
+type Store struct {
+	metrics map[string][]time.Duration
+	mtx     *sync.Mutex
```
bep: The naming standard seems to be `mu` for mutexes.

moorereason: Done.

moorereason:
@bep, I see d000cf605091c6999b72d6c632752289bc680223, but how do we go about accessing the Site/Deps state from the site build?  It looks like we'd have circular dependencies.

I was thinking we would need to put this in `tmplimpl`.

bep:
>@bep, I see d000cf6, but how do we go about accessing the Site/Deps state from the site build? It looks like we'd have circular dependencies.

Why do you need deps in there? Just set the `MetricsProvider` if flag set when creating the `TemplateAdapter`; I don't see how that can create a circular dependency.

bep:
@moorereason it would be great to get this into 0.27. Let me know if you need any more help.

moorereason:
@bep, I've been too busy to work on it lately and likely won't have time until Saturday or Monday.  Feel free to take what I have and run with it.  You won't hurt my feelings.

moorereason:
@bep,
Took another stab at this PR.  Rebased but didn't squash so you could see my progression.  I'd like to squash these commits before we merge, but let me know what you think.

tpl/tplimpl/template.go
```diff
@@ -112,15 +112,25 @@ func (t *templateHandler) Lookup(name string) *tpl.TemplateAdapter {
 		// in the text template collection.
 		// The templates are stored without the prefix identificator.
 		name = strings.TrimPrefix(name, textTmplNamePrefix)
-		return t.text.Lookup(name)
+
+		te := t.text.Lookup(name)
+		if te != nil {
+			te.Metrics = t.Deps.Metrics
```
bep: You should set the `Metrics` at template creation time.

tpl/template.go
```diff
@@ -66,13 +68,16 @@ type TemplateDebugger interface {
...
 	Template
+	Metrics metrics.Provider
 }
 
 // Execute executes the current template. The actual execution is performed
 // by the embedded text or html template, but we add an implementation here so
 // we can add a timer for some metrics.
 func (t *TemplateAdapter) Execute(w io.Writer, data interface{}) error {
-	// TODO(moorereason) metrics fmt.Println("Execute:", t.Name())
+	if t.Metrics != nil {
```
bep: I may read this wrong, but to me, it looks like `t.Metrics` is always set (i.e. a NoOpProvider if not enabled). So with that construct, you should get rid of this conditional ... but:

I would rather prefer that this was actual nil or zero or whatever (and drop the NoOp construct). It does not look expensive, but I would prefer not to create lots of time objects etc. in the normal run mode.

moorereason: Done.

tpl/tplimpl/template.go
moorereason: I think you're going to have to help me out here.

bep: OK.

PR #3842, opened 2017-08-28
use shimmed approach to translating asciidoctor, asciidoc and rst

miltador:
This is a continuation of #3799 with updated [shimgo](https://github.com/tychoish/shimgo) to support Asciidoctor conversion and fixed tests.
Supersedes #3799.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,8 @@ script:
   - ./hugo -s docs/
   - ./hugo --renderToMemory -s docs/
 before_install:
+  # more recent version of ruby is required for sinatra
+  - rvm install 2.2.4
   # gem install must be run with sudo on OSX
-  - sudo gem install asciidoctor | gem install asciidoctor
+  - sudo gem install asciidoctor sinatra --no-ri --no-rdoc | gem install asciidoctor sinatra --no-ri --no-rdoc
+  - sudo pip install docutils flask
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,16 +1,20 @@
 init:
   - copy c:\MinGW\bin\mingw32-make.exe c:\MinGW\bin\make.exe
   - set PATH=%PATH%;C:\MinGW\bin;%GOPATH%\bin
+  # Take default Ruby out of path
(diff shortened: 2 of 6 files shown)


miltador:
Depends on tychoish/shimgo#8. After merge of that PR I will update this one and see if tests pass.

miltador:
BTW I don't know what causes AppVeyor build to hang, my changes to it's config are minimal.
It's strange it is stuck after `test386` task. Tested locally on my Windows machine, everything is fine.

bep:
>BTW I don't know what causes AppVeyor build to hang, my changes to it's config are minimal.

Your changes are not minimal. I suggest you add some debug statements to figure out where/why it hangs.

miltador:
Issue with hanging builds was due to not released resources after tests run. It is now solved.
`shimgo` is updated with possible data race fixes.
This PR is ready for review now.

tychoish:
@bep what are the barriers to getting this merged?

tychoish:
I'm very interested in getting this merged, and would like to know how to proceed. It seems like this patch works against all tests, against a proposal #2672 that seemed well received. If this patch isn't mergable in this form, I would very much enjoy learning what works remains.

bep:
I have been busy with work++ lately, and I cannot give a date as to when I will. This isn't a straightforward merge, several issues to consider. But I will have to come back to that later.

bep:
I have reviewed this and this will take some considerable amount of my time to get into Hugo. Also, adding 7K lines of inline Ruby and Python code etc. to Hugo as a temporary workaround is a good way on the side of "too much" to be really worth it. This adds some unknown amount of maintenance and support work, which I'm not willing to contribute. This is a good idea, but it is just not worth it. When we get some plugin support for stuff like this, this can be revisited.

tychoish:
Thanks so much for your time, but is there any way you can reconsider?

The existing code just shells out to code which Hugo can't include or
enforce exists, which is much more fragile, and likely leads to a more
confusing experience.

Additionally, the inline code that you mention is in a venodred library,
which is not Hugo's maintenance burden.

PR #3853, opened 2017-09-01
tpl: Add os.fileExists template function

### LINKED ISSUE #3839: Add template func to check if a file exists
I'd like to be able to do something like

```
{{ if fileExists (printf "static/%s.jpg" .BaseFileName) }}
  <img src="{{ printf "static/%s.jpg" .BaseFileName }}" />
{{ end }}
```

For when a file may or may not exist.

FWIW, I'd be okay if `imageConfig` and/or `readFile` just returned nil instead of generating errors if the file doesn't exist, but just querying the existence of a file would save resources in my case. In a perfect world, though, `fileExists` (or whatever you want to call it) would exist *and* `imageConfig`/`readFile` would return nil if the file doesn't exist or can't be read.

digitalcraftsman:
This pull request is intended to fix #3839

Currently, I've an issue with the test cases. `fileExists` returns a boolean and an error / `nil`.

Due to the type assertion `false` can become ambiguous: it can either mean that a file does not exist or that an error is expected.  

If the file does not exist, `fileExists` returns `false` (correctly), implicating that no error occurred, but the test fails because `false` is interpreted like an expected error.

Any ideas how this ambiguity can be circumvented?

### DIFF
--- a/docs/content/functions/fileExists.md
+++ b/docs/content/functions/fileExists.md
@@ -1,0 +1,29 @@
+---
+title: "fileExists"
+linktitle: "fileExists"
+date: 2017-08-31T22:38:22+02:00
+description: Checks whether a file exists under the given path.
+godocref:
+publishdate: 2017-08-31T22:38:22+02:00
+lastmod: 2017-08-31T22:38:22+02:00
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+signature: ["fileExists PATH"]
+workson: []
+hugoversion:
+relatedfuncs: []
+deprecated: false
+aliases: []
+---
+
+`fileExists` allows you to check if a file exists under a given path, e.g. before inserting code into a template:
+
+```
+{{ if (fileExists "static/img/banner.jpg") -}}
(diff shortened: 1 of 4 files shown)


bep:
>Any ideas how this ambiguity can be circumvented?

Check the error returned from `Stat`.

``go
if !os.IsNotExist(err) {
   return err
}

return false
```

**Comment by bep** [comment]
In general, it would be great if you could check if your PR goes green before you leave it to dry.

**Comment by NicoHood** [comment]
This function seems to only work with absolute paths. Could some please help me (add this to the docs as well!) to use a relative path (checking if an image inside the posts folder exists), or update the function to support both?

I used this dirty workaround:
```
        {{ $imgPath := print "/content/" $.Dir "header.png" }}
        {{ if (fileExists $imgPath) -}}
              <img alt="{{ .Title }}" class="img-fluid mb-2" src="header.png" />
        {{ end }}
```

moorereason:
@NicoHood, please don't comment on issues that were closed 3 years ago.  Open a new ticket or ask on the discussion forums.

PR #3858, opened 2017-09-03
Add/3rd party libs paths to config

funkyboy:
Allows to specify the path of third party libs in the config file. Handy if you need to install such tools in a folder that is not included in PATH.
As per https://discourse.gohugo.io/t/specify-the-path-of-pygmentize/8242/6 

This is just for pygmentize. 
I'd like to know if you like this approach/style before going on with rst2html and asciidoc.

Cheers.

### DIFF
--- a/helpers/content_renderer_test.go
+++ b/helpers/content_renderer_test.go
@@ -42,8 +42,8 @@ func (c ContentSpec) renderWithMmark(input string) string {
 }
 
 func TestCodeFence(t *testing.T) {
-
-	if !HasPygments() {
+	v := viper.New()
+	if !HasPygments(v) {
 		t.Skip("Skipping Pygments test as Pygments is not installed or available.")
 		return
 	}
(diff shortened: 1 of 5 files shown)


bep:
Nothing wrong with your approach, but I don't like sending that `cfg` around (it may also be a performance issue; probably not in this case, but doing Viper lookups in hot paths are not performant). Since this is about consolidating I suggest you follow the example of `PathSpec` etc. and create a new `ExtHelperSpec` which can take that `cfg` and do all the path logic on creation time -- and will then own the `HasPygments()` etc. Don't hesitate to ask.

funkyboy:
@bep I admit I am a bit lost :)
Do you have an example of the pattern you illustrated? 
Even a high level overview is fine.

bep:
See `helpers.PathSpec`.

funkyboy:
Updated discussion here: https://discourse.gohugo.io/t/specify-the-path-of-pygmentize/8242
I think this modification doesn't make a lot of sense now.
If you agree feel free to close this PR.

PR #3864, opened 2017-09-07
Improve docker image build

tjamet:
Hi, Thanks for this project, very nice using it.

I noticed the Dockerfile at the root of the project, which is nice, and found some little optimisations thanks to  [latest docker updates (17.05:)](https://docs.docker.com/engine/userguide/eng-image/multistage-build/)

This pull request adds support for multi-stage build allowing to have shorter build time when updating hugo code itself.

As a side effect it builds the image from the working directory instead of getting master from github and reduces, in my measurements, the size of the image from 20MB to 16MB

On another hand, I didn't find an _official_ image in the docker registry. Would it be an option to create an automated build from this repo?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,18 @@
+FROM golang:1.9.0-alpine3.6 AS build
+
+RUN apk add --no-cache --virtual git musl-dev
+RUN go get github.com/kardianos/govendor
+RUN govendor get github.com/gohugoio/hugo
+WORKDIR /go/src/github.com/gohugoio/hugo
+RUN rm -f $GOPATH/bin/hugo
+RUN go install -ldflags '-s -w'
+
 FROM alpine:3.6
-
-ENV GOPATH /go
-ENV PATH $GOPATH/bin:$PATH
-
 RUN \
   adduser -h /site -s /sbin/nologin -u 1000 -D hugo && \
   apk add --no-cache \
-    dumb-init && \
-  apk add --no-cache --virtual .build-deps \
-    gcc \
-    musl-dev \
-    go \
-    git && \
-  mkdir -p \
-    ${GOPATH}/bin \
-    ${GOPATH}/pkg \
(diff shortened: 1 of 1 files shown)


bep:
>On another hand, I didn't find an official image in the docker registry. Would it be an option to create an automated build from this repo?

I recently created the `gohugo` organization on the Docker hub, so there will eventually show up some "official images", but I guess the most interesting case is stable versions, and not

vielmetti:
I'll make a pitch for "official images" to support the new Docker multiarch approach to builds - a single `docker run gohugo/hugo` would be able to run unmodified on an any architecture. The `mquery` program is an example of a build infrastructure that uses this, see https://github.com/estesp/mquery/issues/1 for an example.

bep:
@vielmetti an added note:

* I created the `gohugo` org at Docker hub some time ago (just to make sure no one else took it)
* I have at least one dockerfile in https://github.com/bep/dockerfiles/tree/master/ci-hugo-deploy that I should move to the Gohugoio GitHub org
* So, I guess it also makes sense to have a docker file in the `hugo` repo (i.e. the latest source?), we should consider creating a `gohugoio/dockerfiles` project (name?) with subfolders with some dockerfiles.

My time (and Docker knowledge ...) limits my capacity to actually do the stuff above, but if others want to sketch out a project structure etc. (with some simple build setup), then I'm happy to create the project and add some people as collaborators to that project.  We could mark it as "in progress" until it's ready.

vielmetti:
@bep sounds good to mark a separate issue for multiarch support in Docker and mark it a work in progress.

I know that @tianon has been helpful in proving assistance with the necessary tooling for this.

tianon:
If you're interested in contributing to the official images, see https://github.com/docker-library/official-images#contributing-to-the-standard-library and https://github.com/docker-library/official-images/blob/master/NEW-IMAGE-CHECKLIST.md. :+1:

PR #3865, opened 2017-09-07
travis: clean go versions

dvrkps:


.travis.yml
```diff
@@ -2,8 +2,8 @@ language: go
 sudo: required
 dist: precise
 go:
-  - 1.8.3
-  - 1.9
```
bep: Are you sure those two are comparable? Will the latter get us 1.9.1 when that comes out?

dvrkps: Yes, and Yes, 1.x will match 1.9.1 but also 1.10, ...
https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use

bep: No, that is not what it says in the documentation; 1.x will not match 1.9.1; it will match the latest major version, which is not what we want.

dvrkps: I check my old builds before go 1.9 and for 1.x build travis use go version 1.8.3
https://travis-ci.org/dvrkps/bima/jobs/258657992#L221

bep: OK, that may be true, but I like to be explicit about what Go versions we support. New versions come infrequently, so it is not "big work".

PR #3873, opened 2017-09-10
Release v0.27

bep:


bep:
@digitalcraftsman @moorereason I have just trimmed the release notes; add/edit if you want.

Note: We now release from `CircleCI` and this branch so to create a little less noise add `[ci skip]` to your commit messages. You will not accidently trigger a release.

That process is now in broad steps:

1. Release process is triggered from branch names release-<version>
2. There is a manual approve step after the basic build
3. First it creates a release notes draft.
4. A rename to "-ready.md" then triggers a second build-approve-release flow which does the actual binary building and releasing.

PR #3877, opened 2017-09-11
Added Hugo hosting via KeyCDN.

codyarsenault:


bep:
Hey; I should have noticed earlier, but you need to open this PR here:

https://github.com/gohugoio/hugoDocs

We have a "mirror" of the docs in this repo to get it in the source distribution, but we only add content that is tightly coupled with code changes.

Issue #2699, opened 2016-11-16
.IsTranslated behaviour with identical filenames


bcgwebdesign:
Using Hugo v0.17 on ubuntu 14.04.4 

In the case of 
```
/about/
/about/index.md
/about/index.de.md
/services/index.md
```

the ,isTranslated variable is true for /services/index.md and /about/index.md but both refer to the /about/index.de.md

Can the behaviour be improved to deal with this? 

The workaround I'm going with at the moment is to not have filenames the same throughout the site.

(using [this partial](https://gohugo.io/content/multilingual/#link-to-translated-content) for showing when a page has a translation)

bep:
@abourget

bep:
This issue has been automatically marked as stale because it has not been commented on for at least six months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

This issue will automatically be closed in the near future if no further activity occurs. Thank you for all your contributions.

borisschapira:
@bep A possible workaround would be to add support for a translation key inside the front matter metadatas. That way, a contributor would specify the same key for content he would like to see matched, without interfering with filenames.

bep:
@borisschapira yea, that is probably the way to go.

bademux:
Nested section multilanguage articles are broken.

Meanwhile I place here very limited and ugly hack

Show language dropdown
assertions:
"en" is your default language
path for article is unchanged
content placed in "/content/" dir 
allarticle files are named as "index\.(.{2})\.md"
PureCss framework is used
```html
    <div class="lang-menu pure-menu pure-menu-horizontal">
        <ul class="pure-menu-list">
        <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
            <a href="#" class="pure-menu-link"><img src="/imgs/flag_{{ .Lang }}.png" alt="{{ .Lang }}" /></a>
            <ul class="pure-menu-children">
{{- $langs := printf "%s" $.Site.Languages -}}
{{- $articleRegex := "index\\.(.{2})\\.md" -}}
{{- $path := printf "/content/%s" .File.Dir -}}
{{- range $files := readDir $path -}}
    {{- $lang := .Name | replaceRE $articleRegex "$1" -}}
    {{- if in $langs $lang -}}
[...]

PR #3889, opened 2017-09-18
hugolib: add new front matter field to use as the translation grouping key

mbarbon:
This commit implements the suggested solution in issue https://github.com/gohugoio/hugo/issues/2699

I'm not sure on the simplest way of writing an automated test for this: I thought about adding a new test case in site_output_test.go, but maybe a full integration test is too much for this?
 
As an aside: I'm not sure why the logic that uses the section as part of the translation key is only used for nodes and not for content pages (it seems a sensible default to me), so I left it that way in my patch. I can submit a follow-up pull request to change it if you think it makes sense.

Thanks!
Mattia

### DIFF
--- a/docs/content/content-management/front-matter.md
+++ b/docs/content/content-management/front-matter.md
@@ -134,6 +134,9 @@ There are a few predefined variables that Hugo is aware of. See [Page Variables]
 `title`
 : the title for the content.
 
+`translationKey`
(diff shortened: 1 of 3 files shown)


bep:
This seems to have gone under my radar. This should be implemented already. Thanks for the PR!

PR #3894, opened 2017-09-21
tpl/tplimpl: Add "Twitter" shortcode for consistency

pacollins:
This will increase consistency in UX by matching with `{{< instagram >}}`, `{{< gist >}}`, `{{< youtube >}}`, `{{< vimeo >}}`.

I'm not sure if this is the most efficient way to implement this, and while I realize the reasoning for {{< tweet >}}, it creates inconsistency when compared to other platforms.

### DIFF
--- a/tpl/tplimpl/template_embedded.go
+++ b/tpl/tplimpl/template_embedded.go
@@ -55,6 +55,7 @@ func (t *templateHandler) embedShortcodes() {
 {{ end }}`)
 	t.addInternalShortcode("gist.html", `<script src="//gist.github.com/{{ index .Params 0 }}/{{ index .Params 1 }}.js{{if len .Params | eq 3 }}?file={{ index .Params 2 }}{{end}}"></script>`)
 	t.addInternalShortcode("tweet.html", `{{ (getJSON "https://api.twitter.com/1/statuses/oembed.json?id=" (index .Params 0)).html | safeHTML }}`)
+	t.addInternalShortcode("twitter.html", `{{ (getJSON "https://api.twitter.com/1/statuses/oembed.json?id=" (index .Params 0)).html | safeHTML }}`)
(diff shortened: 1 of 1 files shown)


tpl/tplimpl/template_embedded.go
```diff
@@ -55,6 +55,7 @@ func (t *templateHandler) embedShortcodes() {
 {{ end }}`)
 	t.addInternalShortcode("gist.html", `<script src="//gist.github.com/{{ index .Params 0 }}/{{ index .Params 1 }}.js{{if len .Params | eq 3 }}?file={{ index .Params 2 }}{{end}}"></script>`)
 	t.addInternalShortcode("tweet.html", `{{ (getJSON "https://api.twitter.com/1/statuses/oembed.json?id=" (index .Params 0)).html | safeHTML }}`)
+	t.addInternalShortcode("twitter.html", `{{ (getJSON "https://api.twitter.com/1/statuses/oembed.json?id=" (index .Params 0)).html | safeHTML }}`)
```
bep: This copy and paste is a little bit lazy -- pull it out into a variable and reuse.

pacollins: That's part of the reason I said:

>I'm not sure if this is the most efficient way to implement this...

I know their is another way to do it, but I'm not aware of how to implement that since shortcodes thus far are singular things so I have nothing to reference. If I was making the actually change for UX consistency, it would be to remove `{{< tweet >}}` entirely, but that removes backwards compatibility. I am more familiar with front end and less familiar with GO, if anyone would like to direct me a little more clearly, I'd be happy to make adjustments, but at that point it's probably easier to just make that edit.

bep: To me it would be better to do:

```go
twitterTmpl :=  `{{ (getJSON "https://api.twitter.com/1/statuses/oembed.json?id=" (index .Params 0)).html | safeHTML }}`

t.addInternalShortcode("tweet.html", twitterTmpl)
t.addInternalShortcode("twitter.html", twitterTmpl)
```

pacollins: Oh! You meant make the function a variable, not tweet/twitter. I thought you were trying to not have tweet.html and twitter.html. So I was not sure how you wanted to accomplish that, I will make this change.

bep:
You need to run gofmt (or go fmt) on your changed Go file.

PR #3920, opened 2017-09-27
hugolib: Move metrics output to the end of the site build

moorereason:


hugolib/hugo_sites_build.go
```diff
@@ -66,6 +67,15 @@ func (h *HugoSites) Build(config BuildCfg, events ...fsnotify.Event) error {
 		h.Log.FEEDBACK.Printf("total in %v ms\n", int(1000*time.Since(t0).Seconds()))
 	}
 
+	if h.Deps.Cfg.GetBool("templateMetrics") {
```
bep: Nitpicking, but would it not be cleaner to say `if h.Metrics != nil`?

moorereason: Done.

PR #3921, opened 2017-09-27
tpl: Add float template function

### LINKED ISSUE #3307: Add `float` function to parse string to floating point value
Please add a `float` function so that we can convert a string into a floating point number, analogous to the `int` function for integers.

This will allow us to pass named floating point arguments from a shortcode.  

For my use-case, this is to pass a non-integer "aspect ratio" value.  My current workaround is to supply this as a rational two-integer tuple and to perform the division in the template, but is only an approximation when the value is irrational.

x3ro:
Add a template function that allows conversion to float. This is
useful, for example, when passing aspect ratios into templates,
which tend to not be integers.

Fixes #3307 

Feedback very welcome, this is the first time I've tampered with the hugo source 🎉

### DIFF
--- a/tpl/cast/cast.go
+++ b/tpl/cast/cast.go
@@ -30,6 +30,22 @@ type Namespace struct {
 
 // ToInt converts the given value to an int.
 func (ns *Namespace) ToInt(v interface{}) (int, error) {
+	v = convertTemplateToString(v)
+	return _cast.ToIntE(v)
+}
+
+// ToString converts the given value to a string.
+func (ns *Namespace) ToString(v interface{}) (string, error) {
+	return _cast.ToStringE(v)
+}
+
+// ToFloat converts the given value to a float.
+func (ns *Namespace) ToFloat(v interface{}) (float64, error) {
+	v = convertTemplateToString(v)
+	return _cast.ToFloat64E(v)
+}
+
+func convertTemplateToString(v interface{}) interface{} {
 	switch vv := v.(type) {
 	case template.HTML:
 		v = string(vv)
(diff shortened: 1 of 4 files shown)


tpl/cast/cast.go
```diff
@@ -30,6 +30,22 @@ type Namespace struct {
 
 // ToInt converts the given value to an int.
 func (ns *Namespace) ToInt(v interface{}) (int, error) {
+	v = convertTemplateToString(v)
```
bep: Please don't change exisiting methods. This doesn't look correct and I'm surprised there were not red tests.

tpl/cast/cast_test.go
```diff
@@ -81,3 +81,37 @@ func TestToString(t *testing.T) {
...
+func TestToFloat(t *testing.T) {
+	t.Parallel()
+
+	ns := New()
+
+	for i, test := range []struct {
+		v      interface{}
+		expect interface{}
+	}{
+		{"1", 1.0},
```
bep: These tests assumes that the input is strings only. I suspect your implementation will fail on `2, int64(2), float64(2)` etc.

tpl/cast/cast.go
x3ro: > Please don't change exisiting methods

Why not? Should I just copy-and-paste the code that was used in toInt? I just took what was there before and moved it into a separate helper function.

tpl/cast/cast_test.go
x3ro: I've pushed additional additional test cases, they all pass.

bep: Yea, I see now -- I read your original code slightly wrong, but it helped with the test cases.

PR #3923, opened 2017-09-29
metrics: Detect partialCached candidates

moorereason:
When template metrics are enabled, calculate a checksum for the output
of every partial and identify partials that always generate the same
output.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -24,6 +24,9 @@ import (
 
 // The Provider interface defines an interface for measuring metrics.
 type Provider interface {
+	// RecordPartialChecksum records the checksum of a partial template's output.
+	RecordPartialChecksum(name, checksum string)
+
 	// MeasureSince adds a measurement for key to the metric store.
 	// Used with defer and time.Now().
 	MeasureSince(key string, start time.Time)
@@ -37,15 +40,17 @@ type Provider interface {
 
 // Store provides storage for a set of metrics.
 type Store struct {
-	metrics map[string][]time.Duration
-	mu      *sync.Mutex
+	partialChecksums map[string]map[string]int
+	metrics          map[string][]time.Duration
+	mu               *sync.Mutex
 }
 
 // NewProvider returns a new instance of a metric store.
 func NewProvider() Provider {
 	return &Store{
-		metrics: make(map[string][]time.Duration),
-		mu:      &sync.Mutex{},
+		partialChecksums: make(map[string]map[string]int),
+		metrics:          make(map[string][]time.Duration),
+		mu:               &sync.Mutex{},
 	}
 }
 
@@ -53,6 +58,20 @@ func NewProvider() Provider {
 func (s *Store) Reset() {
 	s.mu.Lock()
 	s.metrics = make(map[string][]time.Duration)
+	s.mu.Unlock()
+}
+
+// RecordPartialChecksum records the checksum of a partial template's output.
+func (s *Store) RecordPartialChecksum(name, checksum string) {
+	s.mu.Lock()
+
+	mm, ok := s.partialChecksums[name]
+	if !ok {
+		mm = make(map[string]int)
+		s.partialChecksums[name] = mm
+	}
+	mm[checksum]++
+
 	s.mu.Unlock()
 }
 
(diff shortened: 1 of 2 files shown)


moorereason:
Sample output:

```text
Template Metrics:

     cumulative       average       maximum         
       duration      duration      duration  count  template
     ----------      --------      --------  -----  --------
   4.016195381s      2.1149ms   40.464406ms   1899  post/single.html
   1.501761942s     6.82619ms  158.542712ms    220  taxonomy/tag.html
   1.287319185s      88.481µs   50.391187ms  14549  theme/_default/li.html
   1.146139245s     535.328µs   21.760069ms   2141  partials/default_head.html
   543.050767ms     253.643µs   20.579276ms   2141  partials/head.html
   497.718056ms    2.117949ms   34.459169ms    235  theme/rss.xml
   363.451719ms     169.757µs   21.300448ms   2141  partials/sidebar.html
   311.709801ms   22.264985ms   95.811685ms     14  taxonomy/category.html
   303.597405ms  303.597405ms  303.597405ms      1  indexes/post.html
   268.807471ms    2.337456ms   57.851899ms    115  _internal/shortcodes/highlight.html
   263.727971ms  263.727971ms  263.727971ms      1  fixed/archives.html
*  116.935001ms      54.617µs   38.906473ms   2141  theme/partials/default_foot.html
   103.365146ms  103.365146ms  103.365146ms      1  _internal/_default/sitemap.xml
    52.778722ms      36.199µs    3.732557ms   1458  _internal/shortcodes/relref.html
    35.470809ms   35.470809ms   35.470809ms      1  index.html
     5.334418ms    5.334418ms    5.334418ms      1  taxonomy/tag.terms.html
     2.265137ms      75.504µs     262.042µs     30  _default/li2.html
     2.236642ms    1.118321ms    1.801326ms      2  theme/fixed/single.html
      930.834µs     930.834µs     930.834µs      1  taxonomy/category.terms.html
      787.871µs     787.871µs     787.871µs      1  404.html

* = this partial always generates the same output; consider using partialCached
```

bep:
This is fantastic. I would never (or it would have taken me long time) have thought of it. Simple and powerful.

metrics/metrics.go
```diff
@@ -24,6 +24,9 @@ import (
 
 // The Provider interface defines an interface for measuring metrics.
 type Provider interface {
+	// RecordPartialChecksum records the checksum of a partial template's output.
+	RecordPartialChecksum(name, checksum string)
```
bep: This all looks good, but I'm not a big fan of the method name here. Not very important, but I would have prefered `AddTemplateChecksum` -- we may use it for other templates in the future.

bep:
Thinking a little about it, I think this can be made even more powerful.

I suspect there are many partial use cases that is "almost cacheable", i.e. it can be cached with a small partial refactoring.

Say this pseudo partial that produces this:

```
# My Single Page

And then some very static and long text ...
```

The above could be refactored by pulling the `{{ .Title }}` out.

So:

Instead of storing the hashes, I suggest:

1. Save the content of the first rendering of a given partial.
2. For the subsequent partial rendering of the same partial: Diff it with the first and keep a "diff counter". If it is "totally different" you could maybe mark it to avoid having to diff all of them.
3. Then create a "cache score" in percentage at the end, i.e. 100% would represent what is currently in this PR.

metrics/metrics.go
moorereason: Done.

moorereason:
@bep,
Thanks for the feedback.  I intentionally kept this first round simple to make sure others like the general idea.

You mention detecting slight differences in the output.  I hadn't considered that approach.  I'll have to think about how to use that and convey results to the user.

My next idea was to give clues about potential uses of cached variants.  For instance, with the following change, you'll get a view of the checksum distributions:

```diff
diff --git a/metrics/metrics.go b/metrics/metrics.go
index d97b7c0..9a8c28f 100644
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -89,11 +89,10 @@ func (s *Store) WriteMetrics(w io.Writer) {
        // Find partialCached candidates
        candidates := make(map[string]bool)
        for name, sumMap := range s.partialChecksums {
-               if len(sumMap) == 1 {
-                       for _, v := range sumMap {
-                               if v > 1 {
-                                       candidates[name] = true
-                               }
+               for _, v := range sumMap {
+                       if v > 1 {
+                               candidates[name] = true
+                               fmt.Fprintf(w, "~ %5d  %s\n", v, name)
                        }
                }
        }
```

Running this against the Hugo docs, you'll see things like this:

```text
~   554  theme/partials/svg/link-ext.svg
~     7  theme/partials/svg/link-ext.svg

~     4  theme/partials/svg/twitter.svg
~   277  theme/partials/svg/twitter.svg
```

I'd like to come up with a simple math formula to highlight instances like these and recommend using partialCached variants.

moorereason:
Replaced by #3938.

PR #3924, opened 2017-09-29
Change SummaryLength to be configurable

### LINKED ISSUE #3734: Make SummayLength configurable
I saw the code "var SummaryLength = 70" in the hugo/helpers/content.go, however, I use Chinese and that length is too short. I Hope it can be configurable. Thanks!

bmon:
Move SummaryLength into the ContentSpec struct and refactor the
relevant summary functions to be methods of ContentSpec. The new
summaryLength struct member is configurable by the summaryLength config
value, and the default remains 70. Also updates hugolib/page to use the
refactored methods.

Resolves #3734


bep:
This looks solid, PRs as I like them, thanks.

PR #3927, opened 2017-09-30
 tpl: Add errorf template function

### LINKED ISSUE #3817: Add errorf template func
I'm working on a heavily data-driven hugo site; virtually all the content is in front matter, and it's all relatively structured.  As a result I tend to have a lot of templates, and they're more complicated than a blog or something where you mostly just need `{{ .Content }}`.

One thing I've struggled a bit with is making sure the input data is formatted correctly. I'd like to add a bit of logic to my templates to verify that required properties exist and, if not, emit an error message. I was thinking something like `errorf`, which would be just like `printf` but output to the console, and cause the build to fail.  That way I could do something like

```
{{ if not .Params.foo }}{{ errorf "foo parameter missing for %s" .URL }}{{ end }}
```

Without this, sometimes Hugo will crash (e.g., if I try to dereference a null pointer) or emit an error message (e.g., passing an unset value to jsonify results in an error about an incorrect number of arguments), but often Hugo will just quietly keep going and I end up with bad data.

I know doing this in the template isn't really ideal, but the only other thing I can think of would be validating against a schema, but I'm not aware of anything for markdown (which is the front matter format I'm using)… Maybe a way to automatically convert front matter to JSON then validate it against a JSON schema?

bmon:
Add template function that will build a string from the given format
string and arguments, then log it to ERROR. This has an intended
side-effect of causing the build to fail, when executed.

Resolves #3817 

I'm not 100% confident everything is done properly. Things I'm unsure about:

1. Could logging to error during `go test` have bad side effects?
2. Is `jww.ERROR` the correct logger to use for this? I tried to copy the style i saw in tpl/data.
3. Is there a way I can also log the template/content that executed the `errorf` call? I think it should be more verbose.

### DIFF
--- a/tpl/fmt/fmt.go
+++ b/tpl/fmt/fmt.go
@@ -15,15 +15,17 @@ package fmt
 
 import (
 	_fmt "fmt"
+	"github.com/gohugoio/hugo/helpers"
 )
 
 // New returns a new instance of the fmt-namespaced template functions.
 func New() *Namespace {
-	return &Namespace{}
+	return &Namespace{helpers.NewDistinctErrorLogger()}
 }
 
 // Namespace provides template functions for the "fmt" namespace.
 type Namespace struct {
+	errorLogger *helpers.DistinctLogger
 }
 
 // Print returns string representation of the passed arguments.
@@ -40,4 +42,9 @@ func (ns *Namespace) Printf(format string, a ...interface{}) string {
 // Println returns string representation of the passed arguments ending with a newline.
 func (ns *Namespace) Println(a ...interface{}) string {
 	return _fmt.Sprintln(a...)
+}
+
+func (ns *Namespace) Errorf(format string, a ...interface{}) string {
+	ns.errorLogger.Printf(format, a...)
+	return _fmt.Sprintf(format, a...)
 }
--- a/docs/content/functions/errorf.md
+++ b/docs/content/functions/errorf.md
@@ -1,0 +1,26 @@
+---
+title: errorf
+linktitle: errorf
+description: Evaluates a format string and logs it to ERROR.
+date: 2017-09-30
+publishdate: 2017-09-30
+lastmod: 2017-09-30
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [strings, log, error]
+signature: ["errorf FORMAT INPUT"]
+workson: []
+hugoversion:
+relatedfuncs: [printf]
+deprecated: false
+aliases: []
+---
+
+`errorf` will evaluate a format string, then output the result to the ERROR log.
+This will also cause the build to fail.
+
+```
+{{ errorf "Something went horribly wrong! %s" err }}
+```
--- a/tpl/fmt/init.go
+++ b/tpl/fmt/init.go
@@ -50,8 +50,14 @@ func init() {
 			},
 		)
 
-		return ns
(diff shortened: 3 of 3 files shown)


tpl/fmt/fmt.go
```diff
@@ -41,3 +42,8 @@ func (ns *Namespace) Printf(format string, a ...interface{}) string {
 func (ns *Namespace) Println(a ...interface{}) string {
 	return _fmt.Sprintln(a...)
 }
+
+func (ns *Namespace) Errorf(format string, a ...interface{}) string {
+	jww.ERROR.Printf(format, a...)
```
bep: What I suggest you do is:

* Create a Distinct error  https://github.com/gohugoio/hugo/blob/master/helpers/general.go#L304 and put it in Namespace

This will remove the spamming effect and make your test more robust.

tpl/fmt/fmt_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	jww "github.com/spf13/jwalterweatherman"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestErrorf(t *testing.T) {
+	errorCount := jww.LogCountForLevel(jww.LevelError)
+	ns := &Namespace{}
```
bep: Use the `New` method to create a namespace and make sure the logger (see comment above) is created in that one.

bmon:
I've made use of the distinct logger to avoid the error log spam.

With regard to the logging output format, I found that if the user wanted additional information to help them isolate where an error is being thrown, they can add it from within the template itself, like:

`{{ errorf "layouts/_default/single.html for %s said: %s" .File.Path "Something bad is going on!" }}`

tpl/fmt/fmt_test.go
```diff
@@ -0,0 +1,14 @@
+package fmt
+
+import (
+	jww "github.com/spf13/jwalterweatherman"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestErrorf(t *testing.T) {
+	errorCount := jww.LogCountForLevel(jww.LevelError)
```
bep: You are not checking the correct logger here. This is the global logger.

tpl/fmt/fmt_test.go
```diff
@@ -0,0 +1,14 @@
...
+	ns := New()
+	assert.EqualValues(t, ns.Errorf("%s\n", "This is an errorf test"), "This is an errorf test\n", "")
+	assert.EqualValues(t, errorCount+1, jww.LogCountForLevel(jww.LevelError), "Error log count not incremented")
```
bep: Error count should be 1 for this test. And you need to assert the correct logger.

bmon: I'm a bit lost here. Is there another way I should be checking the log count? 
I'm unsure how I would apply https://github.com/spf13/jWalterWeatherman/blob/master/log_counter.go#L34 on the distinct logger only.

bep: Let's put it this way: The log count check you currently have does not make sense (it does not make sense). I'm not sure if the distinct logger exports its internal logger, so it may be better to remove this check than to check for something that does not test what you do.

bmon: Ok. I've removed the test.

PR #3931, opened 2017-10-03
Add magefile

### LINKED ISSUE #3909: Replace make with mage
[mage](https://github.com/magefile/mage) is a makefile-replacement that uses go instead of bash as the programming language

One of the benefits of using go is that it is cross platform.  Currently, hugo's makefile won't work on windows without jumping through some hoops.  

In addition, having the build script written in go means that it's more friendly to contributors who already know go, but might be intimidated by lines like this -

```
@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
``` 

I think it would be pretty easy to port Hugo's makefile to mage.  I'd be happy to provide PR doing this, if the maintainers think it's a good idea.

natefinch:
This closes #3909.  The magefile should be identical in functionality
to the current makefile.  It uses the current master of mage.  I wasn't
sure if I should add mage to the vendor directory or what (probably
should).  Wanted to get this up for others to review.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -1,0 +1,230 @@
+// +build mage
+
+package main
+
+import (
+	"bytes"
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/sh"
+)
+
+const (
+	packageName  = "github.com/gohugoio/hugo"
+	noGitLdflags = "-X $PACKAGE/hugolib.BuildDate=$BUILD_DATE"
+)
+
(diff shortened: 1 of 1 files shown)


bep:
> I wasn't sure if I should add mage to the vendor directory or what

Excellent question, but since you, in most cases, uses `mage` to get `govendor`, there is a chicken and egg problem. I would think that `mage` would be like `make` in this department. You need to "get that first" somehow.

/cc @moorereason

natefinch:
So, some notes - Mage has no dependencies outside the stdlib, and builds with 1.7+ (probably even earlier versions, but I haven't tried it).  Go get should work fine basically always.

PR #3933, opened 2017-10-03
tplimpl: Output xmlns:xhtml only if there are translations available

jamieconnolly:
This removes `xmlns:xhtml="http://www.w3.org/1999/xhtml"` from the default sitemap when there are no translation links available. The translation links were added in #3348.


moorereason:
What issue does this resolve?

jamieconnolly:
Apologies! I've added in some detail to the PR now.

PR #3941, opened 2017-10-05
Changed last 'make' statements to 'mage'

hholi:


bep:
Takkar, Hallgeir. For future reference: I changed your commit title to be in line with the same (pedantic) guidelines you just edited ... I have no idea why the build failed, `mage` is new to me, too ...

PR #3951, opened 2017-10-09
Replaced unused 'viper' variable in server log

mdhender:
The verbose output for server was not displaying the current
configuration file name. The log used a global 'viper'
variable. Replaced that with the 'config' variable.

With this change, the verbose log now shows:

  INFO 2017/10/08 22:21:00 Using config file: config.toml

Note that there's still an issue when using multiple config
files. The routine is hard coded to show only the first file
read.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -391,7 +391,7 @@ func InitializeConfig(subCmdVs ...*cobra.Command) (*deps.DepsCfg, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


bep:
I could have fixed this myself, but I will point it out now to get it correct in future PRs: Check the contributor guideline vs your commit message.

mdhender:
Let me know if that commit message is what you're expecting.

bep:
Yea that looks good. Note that simple one-liner commit messages are fine for this, the important part here was "Removed ..." should be "Remove ..." or similar.

PR #3968, opened 2017-10-13
Make chomp return the type it receives

### LINKED ISSUE #2187: Make chomp return the type it receives
See https://discuss.gohugo.io/t/errors-with-type-template-html-should-be-string-with-hugo-0-16-0/3470/3

kropp:
fixes #2187

### DIFF
--- a/tpl/strings/strings_test.go
+++ b/tpl/strings/strings_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/gohugoio/hugo/deps"
+	"github.com/spf13/cast"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


tpl/strings/strings_test.go
```diff
@@ -35,12 +35,20 @@ func TestChomp(t *testing.T) {
...
+		{"\n a\n\r\n", "\n a"},
+		{"\n a\r\r", "\n a"},
+		{"\n a\r", "\n a"},
+		// html to html
+		{template.HTML("\n a\n"), template.HTML("\n a")},
+		{template.HTML("\n a\n\n"), template.HTML("\n a")},
+		{template.HTML("\n a\r\n"), template.HTML("\n a")},
+		{template.HTML("\n a\n\r\n"), template.HTML("\n a")},
+		{template.HTML("\n a\r\r"), template.HTML("\n a")},
+		{template.HTML("\n a\r"), template.HTML("\n a")},
```
moorereason: I would rather see the test cases be strings and the test check both types per case.

kropp: Sounds reasonable, updated

Issue #3564, opened 2017-06-06
Custom language codes not support in multilingual setup


bep:
https://github.com/nicksnyder/go-i18n/issues/72

/cc @bogem

bogem:
The solution is in https://github.com/nicksnyder/go-i18n/pull/74. Waiting for @nicksnyder's answer.

bep:
@bogem what is the status on this?

bogem:
@nicksnyder fixed it in https://github.com/nicksnyder/go-i18n/pull/79. We should update this library and integrate this feature in hugo

bep:
@bogem I have pulled in the lates go-i18n, could you have a look it at it? We could maybe label it as `Hacktoberfest` and hope someone will chime in during that month? Then we could just help him/her implementing it? Sounds good?

bogem:
@bep yes, sounds good. Anyway, it's easy to solve. If nobody will fix this issue on Hacktoberfest, I will do it.

kevingimbel:
Hey @bep, @bogem - I'd like to contribute but I'm not sure where to start and what needs to be done. Could you give me a heads-up?

As far as I understand the problem was loading languages which do not follow the Unicode CLDR standard naming. From the linked issue in go-i18n I understand that this can be fixed by registering the languages with `RegisterPluralSpec` so that we (hugo) can load whatever languages the user wants to have?

---

I've not yet worked with i18n in hugo, but I've been using hugo for a year or so and follow the development closely. :)

bogem:
> From the linked issue in go-i18n I understand that this can be fixed by registering the languages with RegisterPluralSpec so that we (hugo) can load whatever languages the user wants to have?

You are totally right. All the logic of loading the i18n is in [initFuncs](https://github.com/gohugoio/hugo/blob/873a6f18851bcda79d562ff6c02e1109e8e31a88/i18n/i18n.go#L58). Start there. 
And don't forget to write tests.

kevingimbel:
@bogem thanks! I'll start as soon as I get some free time!

bogem:
Good luck👍
If you need help, don‘t hesitate to write here.

kevingimbel:
@bogem I created a Pull Request #3971 

I had to update the `Update` function from `TranslationProvider `. This package is not tested - where should I add tests now? I tested the modification by creating a demo side and running my modified code against it with a language called "rando".

bogem:
[...]

PR #3971, opened 2017-10-15
i18n: Allow custom language codes

kevingimbel:
This PR uses the new `RegisterPluralSpec` function from [go-i18n](https://github.com/nicksnyder/go-i18n) to register all languages defined. Previously only language codes available within the Unicode CLDR  standard could be defined.

### DIFF
--- a/i18n/i18n_test.go
+++ b/i18n/i18n_test.go
@@ -157,8 +157,8 @@ func doTestI18nTranslate(t *testing.T, test i18nTest, cfg config.Provider) strin
 	ids := []string{}
 
 	for file := range test.data {
-		id := strings.Split(file, ".toml")
-		ids = append(ids, id[0])
+		id := strings.TrimSuffix(file, ".toml")
+		ids = append(ids, id)
 	}
 
 	language.RegisterPluralSpec(ids, &language.PluralSpec{})
--- a/i18n/i18n_test.go
+++ b/i18n/i18n_test.go
@@ -23,9 +23,11 @@ import (
 
 	"github.com/gohugoio/hugo/config"
 	"github.com/nicksnyder/go-i18n/i18n/bundle"
+	"github.com/nicksnyder/go-i18n/i18n/language"
 	jww "github.com/spf13/jwalterweatherman"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/require"
+	"strings"
 )
 
 var logger = jww.NewNotepad(jww.LevelError, jww.LevelError, os.Stdout, ioutil.Discard, "", log.Ldate|log.Ltime)
@@ -137,10 +139,29 @@ var i18nTests = []i18nTest{
 		expected:     "hello",
 		expectedFlag: "[i18n] hello",
 	},
+	// Non Unicode CLDR language code
+	{
+		data: map[string][]byte{
+			"dk.toml": []byte("[hello]\nother = \"hej\""),
+		},
+		args:         nil,
+		lang:         "dk",
+		id:           "hello",
+		expected:     "hej",
+		expectedFlag: "hej",
+	},
 }
 
 func doTestI18nTranslate(t *testing.T, test i18nTest, cfg config.Provider) string {
 	i18nBundle := bundle.New()
+	ids := []string{}
+
+	for file := range test.data {
+		id := strings.Split(file, ".toml")
+		ids = append(ids, id[0])
+	}
+
+	language.RegisterPluralSpec(ids, &language.PluralSpec{})
 
 	for file, content := range test.data {
 		err := i18nBundle.ParseTranslationFileBytes(file, content)
--- a/i18n/translationProvider.go
+++ b/i18n/translationProvider.go
@@ -55,8 +55,7 @@ func (tp *TranslationProvider) Update(d *deps.Deps) error {
 			langs = append(langs, r.BaseFileName())
 		}
 	}
-	// we need to register all language codes as "plural spec" to prevent errors
-	// with unknown language codes
+	// We need to register all language codes as "plural spec" to prevent errors with unknown language codes.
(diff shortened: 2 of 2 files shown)


kevingimbel:
Interesting. Travis fails on `mage -v check` but this did not fail on my machine. I'll take a look at it.

kevingimbel:
@bogem Could you help me with this? I added a test for "dk.toml" to the i18n_test.go file, see https://github.com/gohugoio/hugo/pull/3971/files#diff-e0ea71d98579d0c18004dbb09fdfaff7R142

If I run `mage test` the test passes, which means the previously unsupported language "dk" can now be loaded. 

However, with the new `RegisterPluralSpec` function in `translationProvider.go` I get a data race error. I don't know how to debug this - could you give me a hint?

i18n/i18n_test.go
```diff
@@ -137,10 +139,29 @@ var i18nTests = []i18nTest{
...
+	},
 }
 
 func doTestI18nTranslate(t *testing.T, test i18nTest, cfg config.Provider) string {
 	i18nBundle := bundle.New()
+	ids := []string{}
+
+	for file := range test.data {
+		id := strings.Split(file, ".toml")
+		ids = append(ids, id[0])
```
bogem: Use this:
```go
id : = strings.TrimSuffix(file, ".toml")
ids = append(ids, id)
```

i18n/translationProvider.go
```diff
@@ -37,6 +38,8 @@ func (tp *TranslationProvider) Update(d *deps.Deps) error {
 	dir := d.PathSpec.AbsPathify(d.Cfg.GetString("i18nDir"))
 	sp := source.NewSourceSpec(d.Cfg, d.Fs)
 	sources := []source.Input{sp.NewFilesystem(dir)}
+	ps := &language.PluralSpec{}
```
bogem: Move this just before the 60th line.

i18n/translationProvider.go
```diff
@@ -37,6 +38,8 @@ func (tp *TranslationProvider) Update(d *deps.Deps) error {
...
+	langs := []string{}
```
bogem: Move this just before the 54th line.

i18n/translationProvider.go
```diff
@@ -50,6 +53,14 @@ func (tp *TranslationProvider) Update(d *deps.Deps) error {
 
 	for _, currentSource := range sources {
```
bogem: Add a comment here that we should plural specs of languages to prevent the error of using custom language codes.
And add a link to #3564 too.

bogem:
See https://github.com/nicksnyder/go-i18n/pull/81

kevingimbel:
@bogem I made the following changes:
- Move variable initialization in `translationProvider.go`
- Add comment about language specs with a reference to the original issue

i18n/translationProvider.go
```diff
@@ -48,6 +49,18 @@ func (tp *TranslationProvider) Update(d *deps.Deps) error {
 
 	i18nBundle := bundle.New()
 
+	langs := []string{}
+	for _, currentSource := range sources {
+		for _, r := range currentSource.Files() {
+			langs = append(langs, r.BaseFileName())
+		}
+	}
+	// we need to register all language codes as "plural spec" to prevent errors
```
bogem: Add dots at the end of sentences and please start sentences with capital lettes.

kevingimbel:
@bogem I capitalized the "w" in "We" and added a dot at the end of the sentence.

bogem:
To prevent data race try to do the following:
1. Add sync.Mutex to translationProvider:
```go
// i18n/translationProvider.go

// TranslationProvider provides translation handling, i.e. loading
// of bundles etc.
type TranslationProvider struct {
    t Translator
    sync.Mutex // <--
}
```

2. Use mutex when registering plural spec:
```go
// i18n/translationProvider.go

ps := &language.PluralSpec{}
tp.Lock()
language.RegisterPluralSpec(langs, ps)
tp.Unlock()
```

bep:
>Actually I'm not sure, if it will help. But anyway let me know about the result of tests.

It will not help. You cannot just lock the write -- this is a read/write map race. Wasn't this fixed upstream? If so, you need to update your dependencies.

kevingimbel:
@bep the fix from @bogem was rejected upstream https://github.com/nicksnyder/go-i18n/pull/81

bogem:
> @bep the fix from @bogem was rejected upstream nicksnyder/go-i18n#81

[Exactly](https://github.com/nicksnyder/go-i18n/pull/81#issuecomment-340786766). And I find @nicksnyder's arguments convincing.

bep:
OK, but my point still holds. You cannot just lock the writer side of it.

bogem:
@bep how can we add locking for read operations? Can you help please?

bep:
I will look. Ideally we would do the RegisterPluralSpec before we start any processing, but this is a global used by every site, so I think concurrency is hard to avoid. But I will have a look.

bep:
See https://github.com/nicksnyder/go-i18n/issues/82

bep:
As to "getting this one working", all the races seems to be in the Update method, so this should work:

https://github.com/bep/hugo/commit/96ed4f535bd00e9a725c4550ada934b4cc97f47a

kevingimbel:
@bep thanks a lot for getting into this, I was completely lost unfortunately! 

Should I add the code to my PR or will you merge your branch?

bep:
Just copy and paste my code.

kevingimbel:
@bep I added your changes to my PR. `mage hugoRace` no longer fail on my machine, but `mage check` still fails.

bep:
@kevingimbel I will fix/merge this tomorrow. And for the long run, I will see if I can create a PR for the go-18n lib ... some time.

bep:
#4049 replaces and includes this

PR #3973, opened 2017-10-15
helpers: Remove style and script tags in plainify output

### LINKED ISSUE #3235: plainify should remove <style> and <script> sections
`plainify` doesn't strip contents of `<style>` tags

See source of https://www.liwen.id.au/barbershop-charts/, line 34

Where 
```
<meta property="og:description" content="{{ .Content | plainify | htmlEscape | truncate 200 }}">
```

is resulting in

```
<meta property="og:description" content="Here are some easy-to-follow flowcharts to help you remember the words to popular barbershop songs.
  figure { border: 1px solid lightgrey; } 
(Click to zoom)
Hello Mary Lou .gallery .img4 …">
```

`plainify` should probably strip out the contents of any non-displaying tags e.g. `<style>`, `<script>`, (any others?)

uncompiled:
Add new stripHTMLContainers function that strips out specific HTML tags including the contents contained within the element.

Fixes #3235

I originally wanted to see if I could implement this logic while walking through the string, but it seemed excessively complex to detect specific tags while walking through each codepoint in the string.  I ended up with this implementation because it felt easier to reason about.

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"html/template"
 	"os/exec"
+	"regexp"
 	"unicode"
 	"unicode/utf8"
 
@@ -187,6 +188,7 @@ func StripHTML(s string) string {
 		return s
 	}
 	s = stripHTMLReplacer.Replace(s)
+	s = stripHTMLContainers(s)
 
 	// Walk through the string removing all tags
 	b := bp.GetBuffer()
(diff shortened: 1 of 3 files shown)


bep:
There is a benchmark written for this. If you run this before/after:

```bash
▶ benchcmp pre.bench post.bench
benchmark                old ns/op     new ns/op     delta
BenchmarkStripHTML-4     3734          38027         +918.40%

benchmark                old allocs     new allocs     delta
BenchmarkStripHTML-4     4              74             +1750.00%

benchmark                old bytes     new bytes     delta
BenchmarkStripHTML-4     736           50808         +6803.26%
```

uncompiled:
That is helpful!  Let me see if I can re-implement this in some way that isn't painfully slower.  😢

uncompiled:
There was a big improvement in the benchmark when [walking through the string](https://github.com/gohugoio/hugo/pull/3973/commits/eb6e5b07afc6e4c3c8654d20fae68ead1f27b4a3):
```
▶ go test -bench=BenchmarkStripHTML ./helpers
BenchmarkStripHTML-4   	  100000	     17537 ns/op
ok  	github.com/gohugoio/hugo/helpers	2.223s
```

However, I found the biggest improvement by only doing the extra work if `<style>` or `<script>` tags are present in the input string.

The benchmark now seems to be in the same ballpark using either method:

```
▶ benchcmp pre.bench post.bench 
benchmark                old ns/op     new ns/op     delta
BenchmarkStripHTML-4     3854          4274          +10.90%
```

PR #3974, opened 2017-10-15
Added dot notation map access to index func

### LINKED ISSUE #3289: Add "dotted" map access as second try in the index func
My proposal to fix the index func in the Go std lib was rejected, so we are stuck with maintaining a patched version.

I'm working on some Hugo docs, and have a shortcode that creates HTML tables from a JSON data file:

```
{{< datatable "media" "types" "Type" "Suffix" >}}
```

I could probably have done the splitting in the template, but right now I have accepted that I have two levels, and my data list is available at `.Site.Data.media.types`.  This logic would break if it suddenly was `.Site.Data.hugo.media.types`.

It would be nice if I could do:

```
{{ $mylist := index .Site.Data "media.types" }}
```

I.e. if the index func 

1. First did a map lookup with "media.types"
2. If none found and the string contains a ".", split it and do a nested lookup

/cc @moorereason

blackmichael:
Resolves #3289

### DIFF
--- a/tpl/collections/index.go
+++ b/tpl/collections/index.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 )
 
 // Index returns the result of indexing its first argument by the following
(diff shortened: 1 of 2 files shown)


bep:
@blackmichael this looks great, I will merge once the tests go green. And again, sorry for the delay.

PR #3993, opened 2017-10-22
Fix for slashes in taxonomy terms: replace slashes with hyphens, add urlizeSegment function

tsuereth:
I recently noticed an issue on my Hugo site around taxonomy terms containing forward slashes.

My blog is about video games.  So for the taxonomy "games" I have terms such as:
* Kingdom Hearts 358/2 Days
* Retro/Grade
* SHENZHEN I/O

When these taxonomy terms are translated into url-safe slugs, the slashes are left untouched.  As a result, the path to taxonomy term metadata is e.g. `content/games/shenzen-i/o/_index.md`

... however, this doesn't really work correctly.  The term processing gets confused, and will sometimes only see the term as `shenzhen-i` (which is then unable to match up with posts I've made for `shenzhen-i/o`).  The "sometimes" seems to be a simple race condition in how Hugo's output HTML is processed, such that either it catches the metadata and misses the posts, or catches the posts and misses the metadata.

This wackiness can be avoided by forbidding slashes in a slugged/urlized taxonomy term, which is the approach I took in this pull request.  Taxonomy terms will replace slashes with hyphens before going through `MakePathSanitized` -- using `MakePathSegmentSanitized` instead.

Additionally, since my blog made use of shortcodes to link to these taxonomy terms, I made an additional helper function `urlizeSegment` which does the same as `urlize` but replacing slashes with hyphens.  (For example, I can take a shortcode for "SHENZHEN I/O" and use `urlizeSegment` to create a hyperlink to the page for `shenzhen-i-o`.)

There is a potential alternative approach, which is to allow slashes in slugged/urlized taxonomy terms, and somehow fix the front-matter processing to properly resolve these recursive paths.  I am not ambitious enough to try this! but I am curious if maintainers think that might be the better approach.


tsuereth:
I revised these changes and re-posted a PR at #4091

PR #4018, opened 2017-10-27
Add support for height argument to figure shortcode

### LINKED ISSUE #4014: figure shortcode does not parse the height argument
Looking at the [source][1], only the `width` argument is parsed in the `figure` shortcode, not the `height`. 

---

[Discourse thread][2]

[1]: https://github.com/gohugoio/hugo/blob/b6f2e3b4382de83479bfcf91321b7ee3a758b141/tpl/tplimpl/template_embedded.go#L22-L37

[2]: https://discourse.gohugo.io/t/can-i-find-out-the-image-size/2196/12?u=kaushalmodi

kaushalmodi:
Fixes #4014

### DIFF
--- a/docs/content/content-management/shortcodes.md
+++ b/docs/content/content-management/shortcodes.md
@@ -88,6 +88,7 @@ The `figure` shortcode can use the following named parameters:
 * `attr` (i.e., attribution)
 * `attrlink`
 * `alt`
+* `height`
 
 #### Example `figure` Input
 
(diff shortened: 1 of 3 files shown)


kaushalmodi:
The PR works, but it looks ugly with hundreds of characters on a single line. The nested with/if/else/end's on the same line makes it very much unreadable. I tried using `{{-` and `-}}`, but they don't work.

Is there any way to add newlines in those embedded templates?

kaushalmodi:
As a side, I see that the [documentation for the `figure` shortcode`][1] is missing mention of the `width` argument that was already supported. I would like to add that, and also mention the `height` argument support that this PR now adds.

What is the best place to do that? Because I believe that https://github.com/gohugoio/hugoDocs is only for released version of  `hugo`, right?

[1]: https://gohugo.io/content-management/shortcodes/#figure

kaushalmodi:
Pinging @digitalcraftsman to get help with the above documentation question.

bep:
@kaushalmodi every documentation change that goes hand-in-hand with the code (like this one): Update the docs in /docs (i.e. a Git subtree of the original). This will be merged into hugoDocs when we release.

kaushalmodi:
@bep OK, there wasn't much to add to docs other than add ``* `height` ``. I have updated my PR with that.

PR #4022, opened 2017-10-29
Hacktoberfest (plis is no spam)

ShadyPlayz:
You should make it more accessible.


kevingimbel:
@ShadyPlayz I reported your account for spamming. You can't just open random pull requests like this. All you do is waste people's time. 

If you want to participate in Hacktoberfest try to find an issue that matches your skill and try to fix it, then submit a pull request.

PR #4030, opened 2017-10-31
read me corrections

dasamantaraoaditya:


README.md
```diff
@@ -32,7 +32,7 @@ Hugo is designed to work well for any kind of website including blogs, tumbles,
 
 #### Supported Architectures
 
-Currently, we provide pre-built Hugo binaries for Windows, Linux, FreeBSD, NetBSD, macOS (Darwin), and [Android](https://gist.github.com/bep/a0d8a26cf6b4f8bc992729b8e50b480b) for x64, i386 and ARM architectures.
+Currently we provide pre-built Hugo binaries for Windows, Linux, FreeBSD, NetBSD, macOS (Darwin), and [Android](https://gist.github.com/bep/a0d8a26cf6b4f8bc992729b8e50b480b) for x64, i386 and ARM architectures.
```
bep: @anthonyfok my English is that of a Norwegian, but isn't the one one the left the correct version in English?

anthonyfok: Hi @ bep,

My English is that of a Chinese, but I agree with you that the original version with the comma after "Currently" is more grammatically correct.

From https://english.stackexchange.com/questions/131128/currently-the-environment-is-so-contaminated-vs-the-environment-is-currently:

> This comma is actually required, since "When an introductory adverbial element seems to modify the entire sentence and not just the verb or some single element in the rest of the sentence, put a comma after it." (http://grammar.ccc.commnet.edu/grammar/commas_intro.htm)

Seeing this pull request only removes the comma, thereby changing something which is originally correct into something wrong according to most English grammar books, I would say this PR should be closed without merging.

PR #4042, opened 2017-11-02
Add support for --destination sftp://.. style urls the hugo cmd

ziprandom:
This commit adds support for `sftp://[user[:pwd]@]host[:port]/destination/path` style urls as
the `--destination` flag when building a site with hugo.

The sftp implementation is based upon [spf13/afero/sftpfs](https://github.com/spf13/afero/blob/master/sftpfs/sftp.go) virtual file system. The `hugo` command
checks if the `--destination` (`-d`) flag provided is a valid sftp url. If so it  initializes
a connection to the server and a `afero/sftpfs.FS` and sets a `hugofs` with the sftpfs as the
Destination.

If username and/or password are not provided within the url the systemuser name will be
used and authentication is tried with the `~/.ssh/id_rsa` private ssh key.

Host certificates are validated against `~/.ssh/known_hosts`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -127,6 +127,12 @@
   revision = "ae77be60afb1dcacde03767a8c37337fad28ac14"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/kr/fs"
+  packages = ["."]
+  revision = "2788f0dbd16903de03cb8186e5c7d97b69ad387b"
+
+[[projects]]
   name = "github.com/kyokomi/emoji"
   packages = ["."]
   revision = "7e06b236c489543f53868841f188a294e3383eab"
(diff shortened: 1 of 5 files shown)


ziprandom:
I'm very new to go and the hugo project and have no idea whether this feature is wished for at all by the community of maintainers and users here . Tests are missing atm. but I am ready to write them if theres interest in this feature and I get some advise on the code.

ziprandom:
I don't understand the failed travis test thoug. seems it couldn't fetch mage?

bep:
This has been discussed before, and any "publish to remote server" task would have to come after the build and be part of some bigger processing chain -- if we want to go that route (i.e. HTML minification etc.). Rendering directly to the web server is most likely not what you want, and not what we want as some default behaviour in Hugo. You will have a fairly bad error handling situation, potentially ending up with lots of blank pages on the server.

For the future, it may save you some work if you create some kind of proposal on the forum before you start to implement it.

ziprandom:
fair enough, that's what I expected. I implemented the functionality because I needed it in a project and opened the pull request to give back to the comunity. So no wasted time here. I'm using hugo as a publishing tool for tech savvy clients in web projects and after adding this functionality the hugo binary and the hugo source is really everything I have to ship to them ...

PR #4047, opened 2017-11-03
Add internal shortcode 'table'

kaushalmodi:
This new shortcode allows specifying class and style for a Markdown table.

### DIFF
--- a/docs/content/content-management/shortcodes.md
+++ b/docs/content/content-management/shortcodes.md
@@ -107,6 +107,49 @@ The `figure` shortcode can use the following named parameters:
 </figure>
 {{< /output >}}
 
+### `table`
+
+`table` allows specifying `class` and `style` attributes to a Markdown table. This is useful for the cases where you do not want to apply styling for all tables globally.
+
+The `table` shortcode can use the following named parameters:
+
+- `class`
+- `style`
+
+If named parameters are not used, the *first* parameter is assigned to the
+`table` tag's `class` attribute, and the *second* parameter is assigned to the
+`style`. Both parameters are optional.
+
+#### Example `table` Input
+
+{{< code file="table-input-example.md" >}}
+{{%/* table class="toto" style="width:50%" */%}}
+| a | b |
+|---|---|
+| c | d |
+{{%/* /table */%}}
+{{< /code >}}
+
+#### Example `table` Output
+
+{{< output file="table-output-example.html" >}}
+<table class="toto" style="width:50%">
+<thead>
+<tr>
+<th>a</th>
+<th>b</th>
+</tr>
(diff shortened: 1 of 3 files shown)


kaushalmodi:
I have added tests + documentation for this new shortcode.

Below passes:
```
go test -run="TestTable.*Arg" ./...  
```

bep:
I'm on the fence on this one. The PR itself looks solid enough (I have a couple of comments, but can take that later), but my main concern is this:

I'm still hoping that we can solve this in the Markdown renderer with some processing extension syntax for all block level elements, e.g. write something like:

```bash
{: class="table-striped" }
| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| col 3 is      | right-aligned | $1600 |
| col 2 is      | centered      |   $12 |
| zebra stripes | are neat      |    $1 |
```

The above is just something I made up now.

So, while I agree that a `table` shortcode would be a useful *temporary* solution, I would rather have it live outside of the Hugo core -- where it would be hard to remove.

We should really do something about the "shortcode and snippets" distribution problem.

kaushalmodi:
I had to come up with *some* solution as issues for support for use cases like this one have been open on Blackfriday for **many years**, but seems like the union of the set of folks who want this feature and the set of folks who know how to implement this in Go is 0. Can you perhaps provide a PR to Blackfriday to support this?

Here's plan B:

Instead of a table-specific shortcode, would you be up for a generic internal shortcode like **div_wrap** that would simply wrap the HTML of the wrapped Markdown in `<div class="(.Get 0)">` .. `</div>`?

Another plan B idea (plan B2) is to have a generic internal shortcode like **classify** that would **insert** the `.Get 0` specified class in the **first HTML tag** generated as a result of the Markdown->HTML conversion.

### Example 1: table

```
{{% classify "toto" %}}
| a | b |
|---|---|
| c | d |
{{% /classify %}}
```

Output:

```
<table class="toto">
...
</table>
```

### Example 2: heading

```
{{% classify "red" %}}
## This heading needs to be red
{{% /classify %}}
```

Output:

```
<h2 class="red">This heading needs to be red</h2>
```

The `classify` plan would **partially** tend to use cases like these: https://discourse.gohugo.io/t/use-custom-css-classes-when-processing-markdown-content-files/9138/7

kaushalmodi:
I am still interested in a shortcode that wraps the Inner Markdown with a `div` tag with user provided attributes like `class`. This would be a generic solution, and would apply to tables and more!

kaushalmodi:
That's a bummer.

kaushalmodi:
https://discourse.gohugo.io/t/table-in-markdown/11434/

PR #4057, opened 2017-11-05
Support custom external formats/renderers

### LINKED ISSUE #234: ENH: custom markdown processor
I love how easy Hugo makes it to organize and display my content.

I am used to using pandoc to convert my markdown to various formats and would love to use it within Hugo. Is there any way we could shell out to a custom markdown processor within Hugo?

I realize that this might defeat a lot of the speed gains we get by using Hugo, but it would also add some flexibility for those who are willing to sacrifice a little performance.

betaveros:
Add the configuration variable externalFormats, which maps formats to a
custom external executable and arguments, as either a string parsed as
whitespace-separated tokens or a list. Shell out to these executables to
render content of those formats.

Fixes #234.

(I have very little experience with Go, so suggestions on code style or convention would be really welcome.)

### DIFF
--- a/docs/content/content-management/formats.md
+++ b/docs/content/content-management/formats.md
@@ -201,6 +201,20 @@ For example, for Asciidoc files, Hugo will try to call the `asciidoctor` or `asc
 
 To use these formats, just use the standard extension and the front matter exactly as you would do with natively supported `.md` files.
 
+You can also specify entirely custom external helpers for a particular format, possibly overriding the built-in renderers, by adding entries under the `externalFormats` variable in your [site configuration][config]. For example, if you wanted to use `pandoc` to render your markup with the flag `--strip-comments`, you might put the following in `config.toml`:
+
+```
+[externalFormats]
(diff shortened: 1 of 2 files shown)


bep:
Hey @betaveros  -- this PR looks solid, but this is conceptually the same as #796 in another area. 

I'm not merging this at the moment. This may change in the future.

betaveros:
Alright. What I really wanted from this was pandoc support, which seems similar to the existing code for shelling out to rst2html and asciidoc — If I wrote a PR to shell out to that specifically, would that be more likely to be merged?

bep:
Yes, that would be fine. It would also be cool if you somehow could consolidate the external "helper funcs" so this doesn't end up as another func doing almost the same as the others. Your general approach is halfway there, I guess, but with a "hardcoded set of executables".

betaveros:
I will close this for now to focus on #4060 since there's substantial overlapping code and this would need to be rewritten anyway.

PR #4061, opened 2017-11-07
Handle Taxonomy permalinks

### LINKED ISSUE #1208: Option to change taxonomies URL
As discussed [here on the forum](http://discuss.gohugo.io/t/change-the-url-for-categories-and-tags/1110), it would be nice to have the ability to change the default URLs generated by Hugo for taxonomies. 

Right now, we only have plural form, like `http://website.tld/tags/cat`. An option could give instead `http://website.tld/tag/cat`for the same URL.

Maybe a good solution would be to let user decide what he wants in the URL. That way, we could choose the singular instead of plural, but also an other word completely (`http://website.tld/pet/cat`), or nothing at all (`http://website.tld/cat`).

EDIT : as described [here](http://discuss.gohugo.io/t/change-the-url-for-categories-and-tags/1110/7?u=nicolinuxfr), there's an easy way to have singular instead of plural. But we could have an option to change entirely the URL.

betaveros:
Return the correct virtual Section for Taxonomy and TaxonomyTerm.
Restrict permalink expansion to only Pages and Taxonomies, but then
actually use expanded permalinks even for non-Pages.

Fixes #1208.

### DIFF
--- a/docs/content/content-management/taxonomies.md
+++ b/docs/content/content-management/taxonomies.md
@@ -132,7 +132,7 @@ Note that if you use `preserveTaxonomyNames` and intend to manually construct UR
 {{% note %}}
 You can add content and front matter to your taxonomy list and taxonomy terms pages. See [Content Organization](/content-management/organization/) for more information on how to add an `_index.md` for this purpose.
 
-Note also that taxonomy [permalinks](/content-management/urls/) are *not* configurable.
+Much like regular pages, taxonomy list [permalinks](/content-management/urls/) are configurable, but taxonomy term page permalinks are not.
 {{% /note %}}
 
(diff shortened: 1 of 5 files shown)


moorereason:
Can you add tests?

bep:
I suspect the lack of tests is because he is testing the waters as to "will this be merged". Yes, this looks good.

And it would look even better with a testcase or two (there should be a permalinks type of integration test somewhere that you can expand on) and some lines in the docs (in this repo).

betaveros:
Should have tests and documentation now. I added the tests to `taxonomy_test.go` because it was convenient; let me know if you think it doesn't belong there.

PR #4077, opened 2017-11-11
Docker build fix

### LINKED ISSUE #4076: Docker build fails
If you try a normal Docker build it fails with following message

```
$sudo docker build -t hugo:org .

Sending build context to Docker daemon  72.38MB
Step 1/16 : FROM golang:1.9.0-alpine3.6 AS build
1.9.0-alpine3.6: Pulling from library/golang
88286f41530e: Pull complete
3f27971ac235: Pull complete
91367c599182: Pull complete
8577260f1836: Pull complete
5b99bfbf214e: Pull complete
e21738584314: Pull complete
Digest: sha256:e91ddbf20f44daeba9a9eca328bc8dbaccf790d26d017dfc572bc003f556e42d
Status: Downloaded newer image for golang:1.9.0-alpine3.6
 ---> ed119d8f7db5
Step 2/16 : RUN apk add --no-cache --virtual git musl-dev
 ---> Running in edc90ba4ad12
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/x86_64/APKINDEX.tar.gz
(1/5) Installing libssh2 (1.8.0-r1)
(2/5) Installing libcurl (7.56.1-r0)
(3/5) Installing expat (2.2.0-r1)
(4/5) Installing pcre (8.41-r0)
(5/5) Installing git (2.13.5-r0)
Executing busybox-1.26.2-r5.trigger
OK: 25 MiB in 17 packages
 ---> b41a21807b15
Step 3/16 : RUN go get github.com/golang/dep/cmd/dep
 ---> Running in ec52bc4d2c4a
 ---> 92752b578452
Step 4/16 : WORKDIR /go/src/github.com/gohugoio/hugo
 ---> fee664d6fc57
Step 5/16 : RUN dep ensure
 ---> Running in b96ddc7caaeb
could not find project Gopkg.toml, use dep init to initiate a manifest
The command '/bin/sh -c dep ensure' returned a non-zero code: 1
```

I can easily fix it and will send a pull request.

Have you ever thought about compressing your Go binaries to further reduce the image size?
I could offer some code snippet to reduce image size to ~8.2MB

Also the Dockerfile could be updated in general and cleaned up.

gruebel:
This should fix #4076

In general I could offer to clean up your Dockerfile :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,26 @@
-FROM golang:1.9.0-alpine3.6 AS build
+FROM golang:1.9.2-alpine3.6 AS build
 
-RUN apk add --no-cache --virtual git musl-dev
-RUN go get github.com/golang/dep/cmd/dep
+RUN apk add --no-cache \
+    git \
+    musl-dev \
+ && go get github.com/golang/dep/cmd/dep
 
+COPY Gopkg.lock Gopkg.toml /go/src/github.com/gohugoio/hugo/
 WORKDIR /go/src/github.com/gohugoio/hugo
-ADD . /go/src/github.com/gohugoio/hugo/
-RUN dep ensure
+RUN dep ensure -vendor-only
+COPY . /go/src/github.com/gohugoio/hugo/
 RUN go install -ldflags '-s -w'
 
 FROM alpine:3.6
-RUN \
-  adduser -h /site -s /sbin/nologin -u 1000 -D hugo && \
-  apk add --no-cache \
+
+RUN adduser -h /site -s /sbin/nologin -u 1000 -D hugo \
+ && apk add --no-cache \
     dumb-init
 COPY --from=build /go/bin/hugo /bin/hugo
-USER    hugo
+USER hugo
 WORKDIR /site
-VOLUME  /site
-EXPOSE  1313
+VOLUME /site
+EXPOSE 1313
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--", "hugo"]
 CMD [ "--help" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,8 @@ RUN apk add --no-cache --virtual git musl-dev
 RUN go get github.com/golang/dep/cmd/dep
 
 WORKDIR /go/src/github.com/gohugoio/hugo
-RUN dep ensure
 ADD . /go/src/github.com/gohugoio/hugo/
+RUN dep ensure
 RUN go install -ldflags '-s -w'
 
 FROM alpine:3.6


Dockerfile
```diff
@@ -4,8 +4,8 @@ RUN apk add --no-cache --virtual git musl-dev
 RUN go get github.com/golang/dep/cmd/dep
 
 WORKDIR /go/src/github.com/gohugoio/hugo
-RUN dep ensure
```
tjamet: I would personally opt for adding `Gopkg.*` and running `dep ensure -vendor-only` instead of `dep ensure`.
Reason being that `dep ensure` takes a while (see https://github.com/golang/dep/blob/master/docs/FAQ.md#why-is-dep-slow) and that by nature, dep locks dependencies in a well-known file. We expect the dependencies lock to evolve less often than the code itself hence we can benefit docker build cache when re-building the image after hugo code update.

the following patch works for me:
```
diff --git i/Dockerfile w/Dockerfile
index 889584b6..119a972f 100644
--- i/Dockerfile
+++ w/Dockerfile
@@ -3,8 +3,9 @@ FROM golang:1.9.0-alpine3.6 AS build
 RUN apk add --no-cache --virtual git musl-dev
 RUN go get github.com/golang/dep/cmd/dep

+ADD Gopkg.* /go/src/github.com/gohugoio/hugo/
-RUN dep ensure
+RUN dep ensure -vendor-only
 ADD . /go/src/github.com/gohugoio/hugo/
 RUN go install -ldflags '-s -w'
```

gruebel:
No problem, I added the vendor-only flag. I also bumped up the build stage docker base image to 1.9.2.

In general I reformated the whole Dockerfile, while honoring the offical Dockerfile best practices.

- COPY instead of ADD (see https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy)
- removed the virtual flag from the apk add command, because there is no point in using it, when you are not planing to remove the dependencies afterwards

Dockerfile
```diff
@@ -1,23 +1,26 @@
-FROM golang:1.9.0-alpine3.6 AS build
+FROM golang:1.9.2-alpine3.6 AS build
 
-RUN apk add --no-cache --virtual git musl-dev
-RUN go get github.com/golang/dep/cmd/dep
+RUN apk add --no-cache \
+    git \
+    musl-dev \
+ && go get github.com/golang/dep/cmd/dep
```
tjamet: Why using multiple lines and a single layer of cache here?

I find short single lines more readable personally

gruebel: I always use multi-line for package listing, it makes it much easier to read and also PR to review. In the offical Dockerfile best practice is a short passage about sorting multi-line, which points already in the right direction of using multi-line in general for package listing https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#sort-multi-line-arguments

I definitley agree on using a oneliner, if you just install one package, but with two I prefer to use multi-line and starting with three there is no other choice, then going multi-line.

Regarding the merge of the two layers, it is just about saving a bit of storage, few kbs, https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#minimize-the-number-of-layers and speed, because Docker doesn't have to save the intermediate layer.

```
68667e787a6e        24 hours ago        /bin/sh -c go get github.com/golang/dep/cm...   39MB
5eb1bd0ea57e        24 hours ago        /bin/sh -c apk add --no-cache     git     ...   31.6MB
```
```
7218153155d7        24 hours ago         /bin/sh -c apk add --no-cache     git     ...   70.5MB
```

And lastly it's about constistancy, if you look at the final stage of the Dockerfile, there is both used multi-line package listing and merging two layers ;)

https://github.com/gohugoio/hugo/blob/42fbf15fb7727a95d135488f6e3d84d04d717361/Dockerfile#L12-L15

tjamet: I don't think the consistency nor the size are adding value here.

We are in the middle of the 'compilation' stage of a multi-stage build, both stages solves different problems.
At this stage, layers stays on the builder host and will never be pushed, compared to the latest stage where layers are actually pushed and where it is interesting to lower their number to gain on the image size.
At the same time, having more layers during this stage adds more steps for the user to use cache, interesting when it comes to re-building an image

However, I don't think that, in the current case, it has a significative impact, just a general thinking about guidelines and spirit of guidelines :)

gruebel: I think in the end it depends totally one personal taste.

The build stage is just garbage stage where you put all in, which you don't want to use in your final image, but need to make it.

And you are right, which ever approach we choose here it has almost no impact :D

So anyone else has an opinion on it? Then he/she could decide, which way to go.

rmetzler:
The current master has a defunct Dockerfile. I have the same problem as #4076 

I wonder why this PR hasn't merged yet and I guess it's because it does more than just fixing the bug.
Can we maybe have the bugfix and discuss the other topics later?

ellerbrock:
@gruebel fully agree with your answers and docker best practices!
i just saw your pr after i run in the problem with the build myself yesterday night and made a fix for it: https://github.com/gohugoio/hugo/pull/4154 

done here similar like you docker best practices, added multilines and changed [ADD to COPY](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) and updated alpine to the latest version.

after that i made a second pr with the production hugo image from scratch: https://github.com/gohugoio/hugo/pull/4155
and created a dockerhub and quay account to hopefully automate this process in the future.

hope that the maintainer (ping @anthonyfok)  would see the need for having a automated official hugo image for the community.

anyway, just wanted to say that i'm totally on your side with the answers what you said about the docker build stuff and best practices.

docker-best-practices

cheers maik

skoblenick:
Where is this PR at? I have seen the other optimization PR's but the immediate concern should be to fix the existing Dockerfile.

PR #4091, opened 2017-11-16
Implementing path-segment sanitization rules

### LINKED ISSUE #4090: Slashes in a taxonomy term break path resolution

* Kingdom Hearts 358/2 Days
* Retro/Grade
* SHENZHEN I/O

tsuereth:
Implementing path-segment sanitization rules, for things like taxonomy terms and titles in permalinks, so that slashes and pound signs get replaced (with hyphens) rather than considered as part of a filesystem or URL path

Includes a new template function urlizeSegment (and corresponding documentation) which can be used in content to emulate said path-segment sanitization

Fixes #4090


tsuereth:
I bungled my fork's git history, sorry for the noise.  Re-submitted at #4092

Issue #2835, opened 2016-12-26
Deprecate preserveTaxonomyNames


bep:
It is complicated and should not be needed anymore. It adds maintenance that I would love to get rid of.

moorereason:
Why is it not needed anymore?

bep:
>Why is it not needed anymore?

You can set title in front matter.

It is still convenient -- but I'm not willing to maintain it.

bep:
To put it in the context, I just spent 40 minutes on #2809 -- and I still have not fixed it (some other corner cases bumped to surface once I added some proper tests for it). It made sense when I added it, but it probably was a mistake.

moorereason:
0.18 is a game changer for many things, and I'm trying to make sure I understand.  You mean we can now set the Taxonomy Term title via frontmatter, such as in `context/categories/acteur/gerard-depardieu/_index.md` (as needed in the original issue #1180):

```
+++
title = "Gérard Depardieu"
+++
```

/cc @nicolinuxfr

bep:
>You mean we can now set the Taxonomy Term title via frontmatter

Yes. It will be some "work" if you have many such taxonomies, but I guess it would also be natural with a picture or some other metadata about Mr Depardieu.

But my main point about this issue is to make stuff simpler, so we can make changes without to much worrying about stuff blowing up in unexpected places.

bep:
But your example is slightly wrong, it should be:

```
content/acteur/gerard-depardieu/_index.md
```

bep:
It seems like this issue will be the "fix" for #2809 -- I don't see a simple solution to that. I have spent enough time already, so I think I'm just gonna bite the apple.

bep:
>It seems like this issue will be the "fix" for #2809 -- I don't see a simple solution to that.

OK, I think I found a way.

moorereason:
Good.  I think we're stuck with this feature until we go through a deprecation period.

mikhail73:
I have 200kb size (4000+ strigs) /data/names.yaml for generate taxanomy terms like this http://pokrovsky.newgod.su/authors/%D0%BA%D0%BE%D1%87%D0%B5%D1%88%D0%BA%D0%BE%D0%B2/. 
You want to "deprecate" it?

bep:
@mikhail73 not sure how your taxonomy mapping can have anything to do with `preserveTaxonomyNames`?

mikhail73:
[...]

PR #4092, opened 2017-11-16
Implementing path-segment sanitization rules

### LINKED ISSUE #4090: Slashes in a taxonomy term break path resolution

tsuereth:
Implementing path-segment sanitization rules, for things like taxonomy terms and titles in permalinks, so that slashes and pound signs get replaced (with hyphens) rather than considered as part of a filesystem or URL path

Includes a new template function urlizeSegment (and corresponding documentation) which can be used in content to emulate said path-segment sanitization

Fixes #4090

### DIFF
--- a/docs/content/functions/findRe.md
+++ b/docs/content/functions/findRe.md
@@ -45,3 +45,4 @@ If you are just learning RegEx, or at least Golang's flavor, you can practice pa
 [`plainify`]: /functions/plainify/
 [toc]: /content-management/toc/
 [`urlize`]: /functions/urlize
+[`urlizeSegment`]: /functions/urlizeSegment
--- a/docs/content/functions/urlizeSegment.md
+++ b/docs/content/functions/urlizeSegment.md
@@ -1,21 +1,21 @@
 ---
-title: urlize
-# linktitle: urlize
-description: Takes a string, sanitizes it for usage in URLs, and converts spaces to hyphens.
-date: 2017-02-01
-publishdate: 2017-02-01
-lastmod: 2017-02-01
+title: urlizeSegment
+# linktitle: urlizeSegment
+description: Takes a string, sanitizes it for usage in URLs, and converts spaces, slashes, and pound signs to hyphens.
+date: 2017-10-21
+publishdate: 2017-10-21
+lastmod: 2017-10-21
 categories: [functions]
 menu:
   docs:
     parent: "functions"
 keywords: [urls,strings]
 godocref:
-signature: ["urlize INPUT"]
+signature: ["urlizeSegment INPUT"]
 hugoversion:
 deprecated: false
 workson: []
-relatedfuncs: [urlizeSegment]
+relatedfuncs: [urlize]
 ---
 
 The following examples pull from a content file with the following front matter:
@@ -23,8 +23,8 @@ The following examples pull from a content file with the following front matter:
 {{< code file="content/blog/greatest-city.md" copy="false">}}
 +++
 title = "The World's Greatest City"
-location = "Chicago IL"
-tags = ["pizza","beer","hot dogs"]
+location = "Chicago IL/USA"
+tags = ["food/pizza","drink/beer","food/hot dogs","we're #1"]
 +++
 {{< /code >}}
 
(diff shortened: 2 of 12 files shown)


bep:
I understand the problem, but we cannot add yet another special thing with another set of special methods to handle this. 

Either we add this to the existing Urlize func, or we don't. I'm not sure. Will have to think about it. 

/cc @moorereason

moorereason:
I agree with @bep.

tsuereth:
Thanks @bep and @moorereason for your attention on this!

I agree with the sentiment of reducing special-case code paths -- that being said, the complication I ran into when beginning this work is that URL/path sanitization is applied somewhat inconsistently throughout Hugo's codebase.

In some places, path segments (which *should not* contain a slash) are sanitized before joining them together.  But in other places, the entire path (which *must* contain slashes) is sanitized at once.

(This is not to mention the hypothetical backwards-incompatibility for templates that might currently use `urlize` and expect slashes to come back out of it.)

Phrasing the functions `MakePathSegmentSanitized` and `URLizeSegment` as I did was intended to start separating these concerns.  If you believe the goal should be to make the formation of path segments into full paths (or URLs) follow more uniform rules, then there is substantially more code auditing to do.  And I suppose that some kind of new interface for templates would be required, to avoid breaking existing content.

Would you like help with this?  Per my examples, I'm somewhat invested in the cause.

moorereason:
You’re correct, which is why I said I agreed with @bep’s “will have to think about it” comment.

My recommendation at this point is to remove the template func stuff and just focus on fixing the internal path sanitizer issues for taxonomies and titles. Those are clear bugs.

How important is the template func for your use case?

tsuereth:
I use the template function (currently, `urlize`) to implement shortcodes for linking to a specific term's page.  I.e. in a post:

`Text text text, {{% game "SHENZHEN I/O" %}}link to the game SHENZHEN I/O{{% /game %}} because more text ...`

Then the shortcode, implemented in `layouts/shortcodes/game.html`:

```
<a href="{{ .Page.Site.BaseURL }}game/{{ index .Params 0 | urlizeSegment }}">{{ .Inner }}</a>
```

Is there a better way to generate a permalink to a specific term's page?  If there is, then I wouldn't have any need for urlizeSegment.

moorereason:
I prefer to use “ref” and “relref” for page links.

tsuereth:
The `ref`/`relref` template functions don't seem to fit my use-case, given that they need a pre-encoded document path.  If I were to hard-code my content with document paths for taxonomy terms, then all of my content would have to be revised by hand if the encoding rules for terms change.  (For example, if the rule for slashes is changed, as in this pull request!)

This would get doubly frustrating for term hyperlinks that are generated dynamically in a layout template, such as when my layout for a post links to the term page for each game tagged in that post.  The template needs both the "pretty" (for display) and sanitized (for hyperlink) forms of each term, so if there isn't a reliable way to translate one to the other, then content would need to provide *both* - perhaps in two separate lists? - to the template.

To be clearer, what I'm looking for in a template function is a way to reference or replicate the process by which Hugo determines where a taxonomy term "lives."  Then my content doesn't have to know any of the rules of this string-to-filepath translation process.

What `urlizeSegment` proposes in this PR - and how I'm currently using `urlize` - is to "replicate" that translation process.  A tactical alternative might be to "reference" it instead, i.e. if my content templates could search a `map` of friendly term names to their sanitized paths.  (Digging into this a bit, however, it would seem to lead to the problems noted in #2835 .)

bep:
>Is there a better way to generate a permalink to a specific term's page?

```bash
{{ .Site.GetPage "taxonomy" "game" "myterm" }}
```

bep:
That said. Let's move the discussion to #4090

tsuereth:
Aha!  This had escaped my notice -- using `GetPage` will solve my content template problem.

For demonstration, here's my revised shortcode template:

```
{{ $term := $.Site.GetPage "taxonomyTerm" "game" (index .Params 0) }}
{{ if $term }}
<a href="{{ $term.Permalink }}">{{ .Inner }}</a>
{{ else }}
{{ .Inner }}
{{ end }}
```

This does change one behavior, compared to the `urlize` implementation -- if I use the shortcode on a term that doesn't exist, then `GetPage` will return `nil` for it, and I can't formulate a hyperlink (whereas `urlize` would make a hyperlink to a page that doesn't exist => 404 error).

At least in my case, this behavior isn't a problem.

Thanks @bep !

bep:
> and I can't formulate a hyperlink 

I think you can do something like:

```
{{ $term := $.Site.GetPage "taxonomyTerm" "game" (index .Params 0) }}
{{ $url := cond (ne $term nil) $term.RelPermalink "http://notfound/" }}
```

But the bottom line here I think is this: The old `urlize` constructs comes from the time where Hugo did not have a complete object graph -- so taxonomies was just files written to disk. We should fix the documentation, and if there still are examples that isn't possible to do without manual URL concatenation in the templates, we should fix that.

PR #4119, opened 2017-11-24
Index Pages in the cache by their Title, if available

### LINKED ISSUE #4118: GetPage behavior differs between auto-generated and existing pages; existing pages cannot be looked up by Title
Following on the discussion in #4092 , I'm using `Site.GetPage` to retrieve information about a page for some arbitrary TaxonomyTerm.

Meanwhile, I also have manually created content for some (but not all) of the terms used on my site.

When GetPage is invoked with a term for which I *have not* created a content page, the lookup works as I expect: for example, `$.Site.GetPage "taxonomyTerm" "game" "Game Title"` will retrieve a Page which Hugo has automatically-generated for the "Game Title" term.

However, when GetPage is invoked with a term for which a content page exists, this no longer works: searching for "Game Title" will return `nil`.

From examining the surrounding code, this is because Page behavior differs for pages which have been generated by Hugo, versus those which exist and have been indexed by Hugo.  The former are indexed by title, while the latter are indexed by document path.

(Thus, if I were to `$.Site.GetPage "taxonomyTerm" "game" "game-title"` then it would retrieve the Page I wanted.  However, since my goal was to retrieve the page without replicating path-sanitization rules, this is contrary to my purpose.)

tsuereth:
Fixes #4118

By adding pages to the cache index keyed by Title, I can look up an existing Taxonomy Term page by the un-sanitized term (in the same way that auto-generated Taxonomy Term pages can be looked-up).

### DIFF
--- a/hugolib/page_collections.go
+++ b/hugolib/page_collections.go
@@ -61,12 +61,20 @@ func (c *PageCollections) refreshPageCaches() {
 				// shortcodes. If the user says "sect/doc1.en.md", he/she knows
 				// what he/she is looking for.
 				for _, p := range c.AllRegularPages {
+					// If the Page has a Title, cache by that as well as by the filepath.
+					if p.Title != "" {
+						cache[p.Title] = p
+					}
 					cache[filepath.ToSlash(p.Source.Path())] = p
 					// Ref/Relref supports this potentially ambiguous lookup.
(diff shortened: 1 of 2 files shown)


tsuereth:
Per notes in #4118 , this fix is inadequate for what I intended, so I'll revisit/reopen it later if appropriate.

PR #4120, opened 2017-11-25
Adding tests to show where GetPage fails to find pages by title

tsuereth:
Following discussion in #4118 -- currently this is expected to fail the following tests:

```
--- FAIL: TestGetPageByTitle (0.10s)
	site_test.go:1008: [0] Expected GetPage("page", "SimplePage") to retrieve a valid page, got nil
	site_test.go:1008: [1] Expected GetPage("page", "Complex Page: More Complicated Title / With Special Stuff") to retrieve a valid page, got nil
--- FAIL: TestGetPageByTaxonomyTerm (0.18s)
	site_test.go:1071: [0] Expected GetPage("taxonomyTerm", "tags", "SimpleTag") to retrieve a valid page, got nil
	site_test.go:1071: [1] Expected GetPage("taxonomyTerm", "tags", "Complex Tag: Tag Harder") to retrieve a valid page, got nil
```

It should pass the final iteration in TestGetPageByTaxonomyTerm, i.e. GetPage on a term page for which there is no associated content should work.

### DIFF
--- a/hugolib/site_test.go
+++ b/hugolib/site_test.go
@@ -963,3 +963,66 @@ func TestRefLinking(t *testing.T) {
 
 	// TODO: and then the failure cases.
 }
+
+var simpleTagPage = []byte(`+++
+title = "SimpleTag"
++++`)
+
+var complexTagPage = []byte(`+++
+title = "Complex Tag: Tag Harder"
++++`)
+
+var taggedPost = []byte(`+++
+tags = ["SimpleTag", "Complex Tag: Tag Harder", "Tag Without Content"]
+title = "Test Post Don't Ignore"
+date = 2017-11-25T00:00:00Z
++++
+This test post has multiple tags, some of which have associated content pages, and some of which don't.`)
+
+func setupTaxonomyTermsMockSite(t *testing.T) *Site {
+	sources := []source.ByteSource{
+		{Name: filepath.FromSlash("tags/simpletag/_index.md"), Content: simpleTagPage},
+		{Name: filepath.FromSlash("tags/complex-tag-tag-harder/_index.md"), Content: complexTagPage},
+		{Name: filepath.FromSlash("post/testpost.md"), Content: taggedPost},
+	}
+
+	taxonomies := make(map[string]string)
+	taxonomies["tag"] = "tags"
+
+	cfg, fs := newTestCfg()
+
+	cfg.Set("baseURL", "http://auth/")
+	cfg.Set("taxonomies", taxonomies)
+	cfg.Set("preserveTaxonomyNames", true)
+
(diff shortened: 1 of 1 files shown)


hugolib/site_test.go
```diff
@@ -963,3 +963,114 @@ func TestRefLinking(t *testing.T) {
...
+		pageKind          string
+		pageTitle         string
+		expectedPermalink string
+	}{
+		{KindPage, "SimplePage", "http://auth/simple/"},
+		{KindPage, "Complex Page: More Complicated Title / With Special Stuff", "http://auth/complex/"},
+	} {
+		// This call to GetPage simulates a template invocation such as:
+		// $.Site.GetPage "pageKind" "pageTitle"
+		p, err := site.Info.GetPage(test.pageKind, test.pageTitle)
```
bep: Where does it say that we support page title in `GetPage`? Didn't we agree previously that this was ambigous and not possible?

tsuereth: Hi @bep ,

In #4092 , when discussing how to formulate a link to a taxonomy term (given the un-sanitized form of the term), you suggested using GetPage like this:

> > Is there a better way to generate a permalink to a specific term's page?
>
> ```
> {{ .Site.GetPage "taxonomy" "game" "myterm" }}
> ```

What these tests demonstrate is that this will only work if the term does not have an associated content file.  If there is an associated content file, then the term's Page is indexed by the sanitized path to that file, instead.

If it isn't the intent of GetPage to retrieve pages by "title" - and the fact that this work for terms without a content file, is just a coincidence - then my request for "a better way to generate a permalink to a specific term's page" is still outstanding.

bep: Sure, but then you should create a test with a term with and without a content file and do 

 .Site.GetPage "taxonomy" "game" "termWithContentFile"
 .Site.GetPage "taxonomy" "game" "termWithOutContentFile"

Not some totally unrelated test.

tsuereth: If you look a few lines down, that's what `TestGetPageByTaxonomyTerm` tests.

The `TestGetPageByTitle` test is included because, based on my code investigation in #4118 , these two behaviors are very closely related.  They both present the same "problem" of retrieving a Page by its Title; the latter case just happens to be Titled with a taxonomy term.

As a user, I would find it odd if GetPage was fixed to allow retrieving a taxonomy term's Page by Title, but not other types of Page.

That said, since you're right that it isn't the specific focus of the bugs I've been discussing, I'll update my test PR to remove this distraction.

bep:
See #4122

So there is a bug here combined with `cfg.Set("preserveTaxonomyNames", true)` -- but not as described. `.GetPage` requires a `path` to the page -- no titles.

PR #4124, opened 2017-11-27
Update Go versions for Travis CI

moorereason:
Travis now supports wildcards.  Yea!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: go
 sudo: required
 dist: precise
 go:
-  - 1.8.5
-  - 1.9.2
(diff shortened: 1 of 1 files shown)


bep:
We have had this discussion before. We need to test with the version we release with, i.e. the exact version. So, while this may look like a time saver, it isn't what we (I) want.

bep:
@moorereason curious about the "confused smiley" ... Do you not think it is a good idea to test with the same version as we release with?

moorereason:
I'm assuming we're releasing with the latest stable release, but I don't care enough to argue about it.

Issue #4131, opened 2017-11-28
Add "contains" operator in where  function to support list pages in nested subdir


rfancn:
If we want to get all pages under specific subdir in some section, we can add _index.md for each section,  and the template corresponding logic  works like below:
```
{{ $currentDir := .Dir }}

{{ range .Site.RegularPages }}
            {{ if hasPrefix .Path $currentDir }}
                {{ .Title }}
            {{ end }}
 {{ end }}
```
Above logic works to list out all pages under each nested section(subdir), but it seems there is no way to generate a collection and build a paginator by `where` function, I also didn't see a way to generate a slice manually.

The quick way to implement above logic, it need add one operator like `contains` in `where`  function to compare the field value `contains` match value, or the more general way is to add a template function like `apply` to returns a new slice filter with a conditional check function.

rfancn:
PR: [4132](https://github.com/gohugoio/hugo/pull/4132) tested works for  layouts/_default/section.html to list all pages under nested subdir. (**Important**: all nested subdir need have corresponding _index.[langugage].md to get it works as expected)
```
{{ $currentDir := .Dir }}
{{ $paginator := .Paginate ( where (where .Site.AllPages "Kind" "page") "Path" "contains" $currentDir ) }}
{{ range $paginator.Pages }}
     {{ .Title }}
{{ end }}
```

bep:
You can easily list `.Sections` and `.Pages` and combine them with `union` and send them into a paginator.

The PR #4132 looks a little bit too big to me -- and would need a really important use case not possible any other way for me to consider merging.

rfancn:
@bep Thanks for your comments, I may not expressed clearly the use case I want to show here, but you may check if the .Section and .Union you mentioned can fix the problem I list here: http://hdget.com/hugo-list-pages-in-nested-dirs/

rfancn:
@bep I just want to filter all those pages under specific nested sub dirs not only root sector , as far as I know,  regardless how much deeper the sub dirs the page sits in, the section always be  the top root dir.

PR #4132, opened 2017-11-28
tpl: Add "contains" operator in where tpl function

rfancn:
Add "contains" operator in "where" tpl function to compare whether the given field value contains match value or not, in fact which does the similar logic as "in" operator but exchanged the comparison
objects. This is useful when filter pages collection under nested subdir.

See #4131

### DIFF
--- a/docs/content/functions/where.md
+++ b/docs/content/functions/where.md
@@ -76,6 +76,9 @@ The following logical operators are vailable with `where`:
 `not in`
 : `true` if a given field value isn't included in a matching value; a matching value must be an array or a slice
 
+`contains`
+: `true` if a given field value contains a matching value; a matching value must be an array or a slice
+
 `intersect`
 : `true` if a given field value that is a slice/array of strings or integers contains elements in common with the matching value; it follows the same rules as the [`intersect` function][intersect].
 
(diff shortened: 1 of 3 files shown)


rfancn:
Test Case:

1. Organize content files with nested subdir
```
- content/
   - blog/
       - program/
          - golang/
          - python/
       - cloud/
         ...
```
2. make sure _index.[language code].md files created in each subdir including root section
```
- content/
   - blog/
        - _index.en.md
        - _index.zh.md
        - program/
            - _index.en.md
            - _index.zh.md
            - golang/
                   - _index.en.md
                   - _index.zh.md
                   golang-pitfalls.md
                   golang-test.md
            - python/
       - cloud/
         ...
```
3.  create layouts/_default/section.html as below:
```
{{ $currentDir := .Dir }}
{{ $paginator := .Paginate ( where (where .Site.AllPages "Kind" "page") "Path" "contains" $currentDir ) }}
{{ range $paginator.Pages }}
     {{ .Title }}
{{ end }}
```
4. Tested with below urls to see if it list out all nested pages under given path/dir:

- http://localhost/blog/
- http://localhost/blog/program
- http://localhost/blog/program/golang/
...

moorereason:
@rfancn, please add tests to `where_test.go`.

rfancn:
@moorereason Fixed the previous commit error, and put "contains" operator logic in one block, also added tests again following cases:
- string contains substring
- slice contains another slice with the same type:
  - int8,int16,int32,int64
  - string
  - timeType
- interface{} object contains corresponding value

Issue #4083, opened 2017-11-15
Collection key type in YAML breaks map range order


dmgawel:
When defining a collection in Hugo data file using YAML format it is then parsed as `map[interface {}]interface {}`. This fundamentally differs from TOML and JSON which parses collections to `map[string]interface {}`.

Example (YAML):
```yaml
test:
  a: 1
  b: 2
```
is parsed to:
```go
map[interface {}]interface {}{"a":1, "b":2}
```

Example (TOML):
```toml
[test]
a = 1
b = 2
```
is parsed to:
```go
map[string]interface {}{"a":1, "b":2}
```

Example (JSON):
```json
{
  "test": {
    "a": 1,
    "b": 2
  }
}
```
is parsed to:
```go
map[string]interface {}{"a":1, "b":2}
```

As YAML is a superset of JSON one can suppose the same key type in both formats.

The following have a great effect on ranging over maps. From the [range docs](https://golang.org/pkg/text/template/#hdr-Actions):
> If the value is a map and the keys are of basic type with a defined order ("comparable"), the elements will be visited in sorted key order.

Apparently `string` is comparable, unlike `interface{}`.

dmgawel:
I came across another issue with this behaviour. `map[interface {}]interface {}` is not 'jsonifiable'. Given data in `test.yaml` executing: `{{ .Site.Data.test | jsonify }}` outputs:

```
ERROR 2017/11/15 14:43:43 Error while rendering "home": template: index.html:8:23: executing "index.html" at <jsonify>: error calling jsonify: json: unsupported type: map[interface {}]interface {}
```

bep:
I assume this is a `go-yaml` upstream issue -- and I think I have seen it before?

moorereason:
Duplicate of #2441

PR #4138, opened 2017-12-01
parser: Fix YAML maps key type

### LINKED ISSUE #2441: Convert YAML to JSON with menu fails
```
Error: json: unsupported type: map[interface {}]interface {}
```

dmgawel:
To support boolean keys, the `yaml` package unmarshals maps to `map[interface{}]interface{}` (see: https://github.com/go-yaml/yaml/issues/139). This behaviour is unexpected in Hugo case and breaks maps ranging and converting to json. This PR follows many `yaml` package dependent projects and changes all maps to `map[string]interface{}` like we would've gotten from `json`. Config exported by https://github.com/go-yaml/yaml would be perfect solution but it looks like it will not come quickly.

Fixes #2441, #4083

### DIFF
--- a/parser/frontmatter.go
+++ b/parser/frontmatter.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"strings"
 
@@ -201,7 +202,43 @@ func removeTOMLIdentifier(datum []byte) []byte {
 func HandleYAMLMetaData(datum []byte) (interface{}, error) {
 	m := map[string]interface{}{}
 	err := yaml.Unmarshal(datum, &m)
+
+	// To support boolean keys, the `yaml` package unmarshals maps to
+	// map[interface{}]interface{}. Here we recurse through the result
+	// and change all maps to map[string]interface{} like we would've
+	// gotten from `json`.
+	if err == nil {
+		for k, v := range m {
+			m[k] = stringifyYAMLMapKeys(v)
+		}
+	}
+
 	return m, err
+}
+
+// stringifyKeysMapValue recurses into in and changes all instances of
+// map[interface{}]interface{} to map[string]interface{}. This is useful to
+// work around the impedence mismatch between JSON and YAML unmarshaling that's
+// described here: https://github.com/go-yaml/yaml/issues/139
+//
+// Inspired by https://github.com/stripe/stripe-mock, MIT licensed
+func stringifyYAMLMapKeys(in interface{}) interface{} {
+	switch in := in.(type) {
+	case []interface{}:
+		res := make([]interface{}, len(in))
+		for i, v := range in {
+			res[i] = stringifyYAMLMapKeys(v)
+		}
+		return res
+	case map[interface{}]interface{}:
+		res := make(map[string]interface{})
+		for k, v := range in {
+			res[fmt.Sprintf("%v", k)] = stringifyYAMLMapKeys(v)
+		}
+		return res
+	default:
+		return in
+	}
 }
 
 // HandleJSONMetaData unmarshals JSON-encoded datum and returns a Go interface
--- a/parser/frontmatter_test.go
+++ b/parser/frontmatter_test.go
@@ -169,7 +169,8 @@ func TestHandleYAMLMetaData(t *testing.T) {
 	}{
 		{nil, map[string]interface{}{}, false},
 		{[]byte("title: test 1"), map[string]interface{}{"title": "test 1"}, false},
-		{[]byte("a: Easy!\nb:\n  c: 2\n  d: [3, 4]"), map[string]interface{}{"a": "Easy!", "b": map[interface{}]interface{}{"c": 2, "d": []interface{}{3, 4}}}, false},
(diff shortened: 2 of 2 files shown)


parser/frontmatter.go
```diff
@@ -201,9 +201,46 @@ func removeTOMLIdentifier(datum []byte) []byte {
...
+	case []interface{}:
+		res := make([]interface{}, len(in))
+		for i, v := range in {
+			res[i] = stringifyYAMLMapKeys(v)
+		}
+		return res
+	case map[interface{}]interface{}:
+		res := make(map[string]interface{})
+		for k, v := range in {
+			kStr, _ := k.(string)
```
bep: This assumes the key is actually a string, an assumption we cannot make without breakage. Also, this PR could need some more tests, esp. re the recursiveness of this + types.

dmgawel: You're right. I've changed key handling and non-string values are now converted to strings. Additionally I added tests for nested maps and different key types.

parser/frontmatter.go
```diff
@@ -201,9 +202,45 @@ func removeTOMLIdentifier(datum []byte) []byte {
...
+	case []interface{}:
+		res := make([]interface{}, len(in))
+		for i, v := range in {
+			res[i] = stringifyYAMLMapKeys(v)
+		}
+		return res
+	case map[interface{}]interface{}:
+		res := make(map[string]interface{})
+		for k, v := range in {
+			res[fmt.Sprintf("%v", k)] = stringifyYAMLMapKeys(v)
```
bep: This still assumes that this is a map with string keys. If I, for some reason, use integers, then I don't want Hugo to convert them into strings. 

* We can probably assume that all the keys are of the same type, so check one element before doing a conversion.
* Also, use `cast.ToStringE` to convert into a string (much faster)

dmgawel: I see your point but I'd like to point out that converting all keys to strings is a default behaviour for both TOML and JSON parsers in Hugo. What should we do about that? :-)

dmgawel: Hi @bep. Have you thought this through? IMO keeping consistent parsing behaviour between different formats is a better approach, but I leave the decision to you.

bep: "default behaviour for both TOML and JSON parsers in Hugo." I think you are right for the top level maps. But here you are doing a deep conversion, which does not make sense. Arrest me if I'm wrong because I have not looked at this code in a while.

dmgawel: Please review a simple test case I wrote and run on Hugo code without my changes:

```go
package main

import (
	"github.com/gohugoio/hugo/parser"
	"fmt"
)

func main() {
	res, _ := parser.HandleJSONMetaData([]byte("{\"inted\": {\"1\": \"a\"}, \"booled\": { \"true\": \"a\"}}"))
	fmt.Printf("%#v\n", res)
	// map[string]interface {}{"inted":map[string]interface {}{"1":"a"}, "booled":map[string]interface {}{"true":"a"}}

	res, _ = parser.HandleTOMLMetaData([]byte("[inted]\n1 = \"a\"\n[booled]\ntrue = \"a\""))
	fmt.Printf("%#v\n", res)
	// map[string]interface {}{"inted":map[string]interface {}{"1":"a"}, "booled":map[string]interface {}{"true":"a"}}

	res, _ = parser.HandleYAMLMetaData([]byte("inted:\n  1: a\nbooled:\n  true: a"))
	fmt.Printf("%#v\n", res)
	// map[string]interface {}{"inted":map[interface {}]interface {}{1:"a"}, "booled":map[interface {}]interface {}{true:"a"}}
}
```

It looks that indeed Hugo unmarshals TOML and JSON to `map[string]` not only for top-level maps. But maybe I did something wrong, I'm not a Go expert :-)

Deep conversion would made no sense for article front matter but it's intended for complex data stored in `data` directory. YAML seems easier and more readable to structure nested maps and arrays than TOML but because of this inconsistency one cannot sort or range over those maps.

vassudanagunta:
> This behaviour is unexpected in Hugo case and breaks maps ranging and converting to json.

Can you provide a test case that shows this top-level problem?

vassudanagunta:
@bep This merge broke data directory tests that were added since this PR's commit (5c89674, was made 2 months ago). Those tests were added (PR #4373) to increased coverage of the data directory as well as to illuminate **de facto** behavior, whether it desirable, so that it could be discussed and acted upon. That discussion (#4379) is still waiting. I point out the different treatment of YAML, the possible justification for it, and asked *Do we need to support non-string map keys in YAML data?* Above you both discussed:

@bep:
> "default behaviour for both TOML and JSON parsers in Hugo." I think you are right for the top level maps. But here you are doing a deep conversion, which does not make sense. Arrest me if I'm wrong because I have not looked at this code in a while.

@dmgawel 
> Deep conversion would made no sense for article front matter but it's intended for complex data stored in data directory. YAML seems easier and more readable to structure nested maps and arrays than TOML but because of this inconsistency one cannot sort or range over those maps.

Is the merge of this PR an implicit answer of "No, we don't need to support non-string map keys in YAML data", or was this change only meant for front matter as @dmgawel suggests below? The tests I added were intended to guard against unintentional changes.

If the former I will update the tests to reflect the new specification. If the latter I can add code so that YAML in the data directory is treated the old way, different from YAML front matter. It will be easy because I've already done that for JSON. 

It would be nice to have the #4379 discussion.

/cc @moorereason

PR #4144, opened 2017-12-03
chroma: Fix highlight for diff files

### LINKED ISSUE #4143: chroma: Incorrect syntax highlighting for .diff code
hugo version: v0.32-DEV-A2D81CE9

# Steps to reproduce

* in config.yml
```yaml

pygmentsStyle: native
pygmentsUseClasses: true

```

* in content/post/hello.md

```markdown
# Hello
{{< highlight diff >}}
- foo
+ bar
{{< /highlight >}}
```
# Actual

Generated html looks like:

```html

<div class="highlight">
...
<span class="err">+ bar</span>
...
</div>
```
Note the 'err' class.

# Expected

No `err` class.

# Notes

I think the problem is that the chroma lexer does not see the last `\n` and thus does not manage to properly parse the contents of the code block.

dmerejkowsky:
Chroma lexer does not like code that does not have a final `\n`,
so add it to the source code before passing it to the Tokenise()
function.

Fixes #4143

### DIFF
--- a/helpers/pygments.go
+++ b/helpers/pygments.go
@@ -196,6 +196,8 @@ func chromaHighlight(w io.Writer, source, lexer, style string, f chroma.Formatte
 		s = styles.Fallback
 	}
 
+	source += "\n"
+
 	it, err := l.Tokenise(nil, source)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


dmerejkowsky:
Note: maybe we should file a bug in Chroma instead.

bep:
>Note: maybe we should file a bug in Chroma instead.

Yes.

dmerejkowsky:
Bug opened here: https://github.com/alecthomas/chroma/issues/96

dmerejkowsky:
Good news! The fix in chroma was just as easy :)

Maybe we can wait for the next release of chroma and close this ?

PR #4145, opened 2017-12-03
use twitter_cards internal

marcanuy:
Currently _gohugoioTheme_ is using the `twitter_cards.html` partial, while it already exists as an internal template.


bep:
You need to create this PR in the theme: https://github.com/gohugoio/gohugoioTheme

PR #4149, opened 2017-12-06
tpl: Add os.PathSeparator

### LINKED ISSUE #2394: Add ToSlash template func
While debugging https://github.com/devopsdays/devopsdays-web/issues/989 I noticed that the `Path` to a source file uses back slashes on Windows, but forward slashes on Linux etc..

Sample shortcode:

```
{{ $path := split .Page.Source.File.Path "/" }}
```

While this is a normal problem with cross platform, Hugo does not seem to expose what the underlying Path Separator should be (`PathSeparator` in https://golang.org/pkg/os/#FileInfo) nor does it define a way to canonicalise a path (relative or absolute)

Would it be possible to either;
1. Offer an additional method on the File object for CanonicalPath
2. Modify Path to be canonical and offer a NativePath if you need access (This would be a breaking change)
3. Expose the OS Path Separator so we can use it in split operations, perhaps as a method off of `FileSystem` or global variable of some form

moorereason:
os.PathSeparator returns the OS-specific path separator as a string.

Fixes #2394

### DIFF
--- a/tpl/os/os.go
+++ b/tpl/os/os.go
@@ -46,6 +46,11 @@ func (ns *Namespace) Getenv(key interface{}) (string, error) {
 	return _os.Getenv(skey), nil
 }
 
+// PathSeparator returns the OS-specific path separator.
+func (ns *Namespace) PathSeparator() string {
+	return string(_os.PathSeparator)
+}
+
 // readFile reads the file named by filename relative to the given basepath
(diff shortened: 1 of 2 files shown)


bep:
As per discussion in #2394 -- this is not the solution.

moorereason:
Sorry, I misread the discussion.

PR #4154, opened 2017-12-10
fix(dockerfile): bugfix for broken docker build + optimization

ellerbrock:
Hello Hugo's,

i just found out that the Docker Build is broken, so i fixed that for you.
Further i cleaned up the Dockerfile and updated [Alpine Linux](https://hub.docker.com/_/alpine/) to Version 3.7 and the [Go Lang](https://hub.docker.com/_/golang/) Build Image to Version 1.10-rc.

_Another PR is coming soon with some further optimizations._

fixed-it

Cheers Maik

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,35 @@
-FROM golang:1.9.0-alpine3.6 AS build
+# GitHub:  https://github.com/gohugoio/
+# Docker:  https://hub.docker.com/r/gohugoio/
+# Twitter: https://twitter.com/gohugoio
 
-RUN apk add --no-cache --virtual git musl-dev
-RUN go get github.com/golang/dep/cmd/dep
+FROM golang:1.10-rc-alpine3.7 AS build
 
-WORKDIR /go/src/github.com/gohugoio/hugo
-RUN dep ensure
-ADD . /go/src/github.com/gohugoio/hugo/
-RUN go install -ldflags '-s -w'
+RUN \
+  apk add --no-cache \
+    git \
+    musl-dev && \
+  go get github.com/golang/dep/cmd/dep && \
+  go get github.com/kardianos/govendor && \
+  govendor get github.com/gohugoio/hugo && \
+  cd /go/src/github.com/gohugoio/hugo && \
+  dep ensure && \
+  go install -ldflags '-s -w'
 
-FROM alpine:3.6
+# ---
+
+FROM alpine:3.7
+
+COPY --from=build /go/bin/hugo /bin/hugo
+
 RUN \
   adduser -h /site -s /sbin/nologin -u 1000 -D hugo && \
-  apk add --no-cache \
-    dumb-init
-COPY --from=build /go/bin/hugo /bin/hugo
+  apk add --no-cache dumb-init
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,23 +1,35 @@
...
-  adduser -h /site -s /sbin/nologin -u 1000 -D hugo && \
   apk add --no-cache \
-    dumb-init
+    git \
+    musl-dev && \
+  go get github.com/golang/dep/cmd/dep && \
+  go get github.com/kardianos/govendor && \
+  govendor get github.com/gohugoio/hugo && \
+  cd /go/src/github.com/gohugoio/hugo && \
+  dep ensure && \
```
tjamet: Although you consider it breaks docker best practices, you may consider that doing this requires re-building
and re-downloading everything for every single `docker build`.

It is okay when the build step is built by a remote server that always discards cache, but might be interesting when iterating on `docker build` locally. Docker has even edited its [best practices](https://github.com/docker/docker.github.io/blob/master/engine/userguide/eng-image/dockerfile_best-practices.md#minimize-the-number-of-layers) explaining latest versions of docker have "have mitigated the need" of minimizing the number of layers.

As suggested in #4077 distinguishing installing dependent tools, dependencies and actual binary generation allows to benefit the docker build cache while building the image.
Having a dockerfile of (as set in #4077 ):

```
RUN apk add --no-cache \
    git \
    musl-dev \
 && go get github.com/golang/dep/cmd/dep

COPY Gopkg.lock Gopkg.toml /go/src/github.com/gohugoio/hugo/
RUN dep ensure -vendor-only
COPY . /go/src/github.com/gohugoio/hugo/
RUN go install -ldflags '-s -w'
```

considers that:
- `dep`, `git`, `musl-dev` are external tools that change the less, without a strong dependency on the exact version
- `go` dependencies are changing slower than the code itself, and quite [long to install](https://github.com/golang/dep/blob/master/docs/FAQ.md#why-is-dep-slow) that are worth caching in local developments, and needs to be re-built every time `Gopkg.*` changes
- `hugo` code changes the most often and thus needs to be re-build every time a file is changed

ellerbrock:
@tjamet,

thanks for the explanation from the developer view working on localhost.
makes totally sense and i made a new pr with the requested changes https://github.com/gohugoio/hugo/pull/4157

that's why working with other smart people on open source makes so much fun.

cheers maik

tjamet:
Thanks for your consideration @ellerbrock I have proposed kind of the same paragraph on [best practices](https://github.com/docker/docker.github.io/pull/5481) :)

ellerbrock:
yeah i saw that one, nice idea to rewrite the docker best practices :D

ellerbrock:
@tjamet any thoughts in merging it?
and like mentioned here: https://github.com/gohugoio/hugo/pull/4155 i already created a hugoio account for you guys with automated builds. i can give it over to you guys ...

cheers maik

tjamet:
If I had to choose, I think I would go for merging #4077 that has the advantage of benefiting build cache for who is using it locally while keeping the same final image
But I am a contributor, I can't decide what to merge or not

PR #4157, opened 2017-12-11
build(docker-build): optimize the docker build process

ellerbrock:
hello @tjamet,

like discussed in https://github.com/gohugoio/hugo/pull/4154 i updated the dockerfile for faster build times.

_it's always a pleasure to work together with other smart people on open source project and end up with something better then before._

smart

cheers maik

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+*.md
+*.log
+*.txt
+.git
+.github
+.circleci
+docs
+examples
+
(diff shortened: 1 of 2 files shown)


ellerbrock:
i added a `.dockerignore` file like mentioned in [docker best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#use-a-dockerignore-file) and excluded some folders to optimize the build time a bit.

anthonyfok:
This pull request from December 2017 was superseded by PR #4518 in March 2018.  Many thanks to @ellerbrock and @skoblenick for your group effort in improving Dockerfile!

PR #4160, opened 2017-12-12
add Title to MenuEntry

rmetzler:
This PR allows to set a title for a MenuEntry in the site config menus.
This allows to render anchor links in menus which have a title attribute:

```
{{ range .Site.Menus.main }}
<!-- ... -->
            <a href="{{ .URL }}" title="{{ .Title }}">
                {{ .Pre }}
                <span>{{ .Name }}</span>
            </a>
<!-- ... -->
{{ end }}
``` 

This feature is useful for screenreaders and SEO.

I didn't implement the same in the frontmatter menu field, because the attribute `title` is too generic. `linkTitle` or `anchortitle` might be better names.

I would love if you could merge this and we figure out if the title attribute is something people need in frontmatter menues. I only use site config menues yet.

### DIFF
--- a/hugolib/menu.go
+++ b/hugolib/menu.go
@@ -25,7 +25,7 @@ import (
 // or in the site config.
 type MenuEntry struct {
 	URL        string
-	Page       *Page // probably only set when called from Page front matter
+	Page       *Page
 	Name       string
 	Menu       string
 	Identifier string
@@ -219,10 +219,13 @@ func (m Menu) Reverse() Menu {
 }
 
 func (m *MenuEntry) Title() string {
-	if m.title == "" {
-		if m.Page != nil {
-			return m.Page.LinkTitle()
-		}
+	if m.title != "" {
+		return m.title
 	}
-	return m.title
+
+	if m.Page != nil {
+		return m.Page.LinkTitle()
+	}
+
+	return ""
 }
--- a/hugolib/menu.go
+++ b/hugolib/menu.go
@@ -28,6 +28,7 @@ type MenuEntry struct {
 	Name       string
 	Menu       string
 	Identifier string
+	Title      string
 	Pre        template.HTML
 	Post       template.HTML
 	Weight     int
@@ -95,6 +96,8 @@ func (m *MenuEntry) marshallMap(ime map[string]interface{}) {
 			m.Weight = cast.ToInt(v)
 		case "name":
 			m.Name = cast.ToString(v)
+		case "title":
+			m.Title = cast.ToString(v)
 		case "pre":
 			m.Pre = template.HTML(cast.ToString(v))
 		case "post":
--- a/hugolib/menu.go
+++ b/hugolib/menu.go
@@ -25,10 +25,11 @@ import (
 // or in the site config.
 type MenuEntry struct {
 	URL        string
+	Page       *Page // probably only set when called from Page front matter
 	Name       string
 	Menu       string
 	Identifier string
-	Title      string
+	title      string
 	Pre        template.HTML
 	Post       template.HTML
 	Weight     int
@@ -97,7 +98,7 @@ func (m *MenuEntry) marshallMap(ime map[string]interface{}) {
 		case "name":
 			m.Name = cast.ToString(v)
 		case "title":
-			m.Title = cast.ToString(v)
+			m.title = cast.ToString(v)
 		case "pre":
 			m.Pre = template.HTML(cast.ToString(v))
 		case "post":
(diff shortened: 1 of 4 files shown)


bep:
Hey and thanks for this,

We need to have this on all menu items. I know you say you only need "config menus", which is why I'm insisting on getting both right in one go. Because if I merge this, you will probably never come back to complete the task.

There is another issue asking for a reference to the `*Page` on a given menu item (which would solve the title case for those).

So:

* Add `Page` to `MenuItem` and set it when relevant.
* Add a `Title` **method** on `MenuItem` - use the local title (from the config) if set, fall back to `Page.Title` if page is set.

Also, move the test to a standalone test in `menu_test.go` The setupMenuTests harnish is about to be deleted in a branch I'm working on.

rmetzler:
At first I wanted to add my test to `menu_test.go`, but I didn't understand the code. It was far easier to use the old code.
I also never understood the menu attribute in the frontmatter, that's why I initially decided against trying to implement anything there.

You say `MenuItem`, do you mean `MenuEntry`?

rmetzler:
Looks like this is the relevant code https://github.com/gohugoio/hugo/blob/master/hugolib/page.go#L1365-L1419

But the test is in `menu_test.go` and `menu_old_test.go`.

bep:
>You say MenuItem, do you mean MenuEntry?

Yes, that is correct. And you found the correct spot. So the MenuEntry needs a `Page` which can be nill in some cases.

The `menu_old_test.go` does not exist anymore in my current branch, which means that this PR will be hard for me to manage in the near future. It has some valid tests not inside the new file, but it is pretty fragile, so instead of keeping on fiddling with it, I just deleted it. 

From this PR you need some "smoke tests", i.e. a site with a bit of both and check for page + title.

rmetzler:
Ok.

One thing, I find irritating is the [`LinkTitle()` func](https://github.com/gohugoio/hugo/blob/master/hugolib/page.go#L903-L908). From [the docs](https://gohugo.io/templates/lists/#by-link-title) this function and variable would be better named [`anchorText`](https://en.wikipedia.org/wiki/Anchor_text) or `linkText`.

The name `linkText` creates a collision with what I thought the new vairable / function should be named. How should I name the new variable / function? `anchorTitle`, `aTitle` or something different?

bep:
>One thing, I find irritating is the LinkTitle() func. From the docs this function and variable would be better named anchorText or linkText.

I'm not sure what that method has to do with this. This PR does not involve making any changes to `Page`.

`MenuEntry` should get:

1. A `Page` pointer.
2. A `Title()` method, which gets its value from the menu config, if set, or the `Page.LinkTitle()` if page is set.

rmetzler:
I'm not a native English speaker, so please excuse any confusion.

The relevant code you want me to change is part of `page.go`, that's why I'm asking.
And there are a few client calls to `p.LinkTitle()` which doesn't exactly do what the name says. The docs hint it should hold anchor text. I just found the naming confusing, because pages have a `title` attribute, a `linkTitle`, and now get a `title` in the menu.

I added to the test and some of the functionality is already there, because of this line: https://github.com/gohugoio/hugo/blob/master/hugolib/page.go#L1410
Which is very interesting, because as a user, you are free to use any of the available keys in this map.

Also [this comment](https://github.com/gohugoio/hugo/blob/master/hugolib/menu.go#L24-L25) hints in the direction that `MenuEntry` might be better of as an interface with `SiteConfigMenuEntry` and `PageMenuEntry` as the implementations. But I guess this is out of scope.

hugolib/menu.go
```diff
@@ -213,3 +217,12 @@ func (m Menu) Reverse() Menu {
 
 	return m
 }
+
+func (m *MenuEntry) Title() string {
+	if m.title == "" {
```
bep: This is a little bit matter of taste, but I always prefer to return early if I can, and it makes the code less nested and easier to read, .e.g

```go
if m.title != "" {
   return m.title
}

...
```

hugolib/menu.go
```diff
@@ -25,9 +25,11 @@ import (
 // or in the site config.
 type MenuEntry struct {
 	URL        string
+	Page       *Page // probably only set when called from Page front matter
```
bep: Remove this comment. The "probably" makes me start to think that we may have missed something, which I don't think, so this will just confuse the next person looking at this code.

hugolib/menu.go
rmetzler: fixed

hugolib/menu.go
rmetzler: fixed

PR #4167, opened 2017-12-15
tpl: Add reflect namespace

### LINKED ISSUE #4081: Add reflection-related template functions
See https://discourse.gohugo.io/t/how-to-test-if-a-variable-is-of-a-certain-type/9207.

Related to #2427.

moorereason:
Adds KindIs, KindOf, TypeIs, TypeIsLike, and TypeOf funcs in a reflect
namespace.

Fixes #4081

### DIFF
--- a/docs/content/functions/reflect.KindIs.md
+++ b/docs/content/functions/reflect.KindIs.md
@@ -1,0 +1,56 @@
+---
+title: reflect.KindIs
+description: Reports whether a value is of a kind.
+godocref:
+date: 2017-12-15
+publishdate: 2017-12-15
+lastmod: 2017-12-15
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [reflect, reflection, kind]
+signature: ["reflect.KindIs KIND VALUE"]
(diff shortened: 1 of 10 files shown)


kaushalmodi:
Thank you! Can you please also add the docs for these functions?

bep:
Hold on with the docs until we decide on the API.

bep:
As I said in the other PR, we need to be restrictive about adding new template funcs until we fix the docs setup for it.

But we cannot build an API where the user has to set up conditionals with this hefty list:

```bash
* array
 * bool
 * chan
 * complex128
 * complex64
 * float32
 * float64
 * func
 * int
 * int16
 * int32
 * int64
 * int8
 * interface
 * invalid
 * map
 * ptr
 * slice
 * string
 * struct
 * uint
 * uint16
 * uint32
 * uint64
 * uint8
 * uintptr
 * unsafePointer
```

We need to discuss a simpler API. But that will have to wait.

bep:
@moorereason please create a reflect proposal issue so we can agree on the design.

PR #4168, opened 2017-12-16
tpl: Add filepath namespace

### LINKED ISSUE #2394: Add ToSlash template func

### LINKED ISSUE #3187: tpl: Add filepath.FromSlash template func
Hi!

I just want to let you know that `.File.Dir` can't be used as `where` value on Windows systems due to the different slash.

```
{{ with where $.Site.Pages ".File.Dir" "_home/versions/" }}
```

On Windows it needs to be:

```
 {{ with where $.Site.Pages ".File.Dir" "_home\versions\" }}
```

This is a problem when developing on multiple operational systems.

Thanks!

moorereason:
Add Base, Clean, Dir, Ext, FromSlash, Separator, Split, and ToSlash funcs.

Fixes #2394
Fixes #3187

### DIFF
--- a/docs/content/functions/filepath.Base.md
+++ b/docs/content/functions/filepath.Base.md
@@ -1,0 +1,29 @@
+---
+title: filepath.Base
+description: Returns the last element of a path.
+godocref:
+date: 2017-12-15
+publishdate: 2017-12-15
+lastmod: 2017-12-15
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [filepath, base]
+signature: ["filepath.Base PATH"]
+workson: []
+hugoversion:
+relatedfuncs: []
+deprecated: false
+aliases: []
+---
+
+`filepath.Base` returns the last element of `PATH`.
+Trailing path separators are removed before extracting the last element.
+If the path is empty, `filepath.Base` returns "`.`".
+If the path consists entirely of separators, `filepath.Base` returns a single separator.
+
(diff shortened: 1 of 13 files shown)


bep:
I know I said we could be more liberal with template func additions once we got namespaces, but it still has to be maintained/explained to people/documented, so we're not adding the full Go stdlib, so to speak.

I will look at your related PRs once we get the following in place

* https://github.com/gohugoio/hugoDocs/pull/105
* A docs integration with the docs helper JSON for the template funcs

But in general; this PR contains too much new API for no obvious good reason. But as I said, that will have to wait.

bep:
@moorereason please create a `filepath` proposal issue so we can agree on the design.

PR #4195, opened 2017-12-29
docs: Remove defaultExtension from configuration doc

satotake:
As far as I tested, `defaultExtension` in config has no effect with v0.31.1.
It might be deprecated in https://github.com/gohugoio/hugo/commit/ee75e2999b66bd9f258a241c487b6677cf2fa071.
It is misleading to leave the descriptions about `defaultExtensions` in docs.
Removed them.


bep:

Issue #4003, opened 2017-10-24
Mage failing with "undefined: mg.Context" error 


vassudanagunta:
All mage commands (except mage -h) in the hugo source directory fail with the exact same error output:

```bash
mage hugo
# command-line-arguments
./mage_output_file.go:330:17: undefined: mg.Context
Error: error compiling magefiles

mage install
# command-line-arguments
./mage_output_file.go:330:17: undefined: mg.Context
Error: error compiling magefiles

mage -l
# command-line-arguments
./mage_output_file.go:330:17: undefined: mg.Context
Error: error compiling magefiles
```

I even tried with a clean GO environment, following all the steps in `CONTRIBUTING.md` exactly, with the same result. 

I ran `mage -keep` to diagnose the problem.  The resulting [mage_output_file.go](https://github.com/gohugoio/hugo/files/1409337/mage_output_file.go.txt) is attached.  Line 330 is `ctx, cancel := mg.Context()`, indicating that for some reason the Context func in the mg package can't be linked even though I confirmed it was there. 

I can't tell if this is a flaw with mage or the hugo source. I was able to run `mage -l` and `mage build` from the `~/go/src/github.com/magefile/` directory, so I'm pretty sure my mage installation is sound.

vassudanagunta:
@natefinch Any ideas on how to isolate?

moorereason:
You likely need to install a new `mage` (using `...` in your `go get` command).

```
go get -u github.com/magefile/mage/...
```

vassudanagunta:
Trust me, I tried that and everything. Has anyone tried doing building hugo from a completely clean (from scratch) go environment?

natefinch:
do you have a vendor directory under hugo's root directory?  It's probably caching the wrong version there.

kevingimbel:
I just tried to run `mage hugo` before updating mage but _after_ pulling the latest changes from master. 

Without updating `mage` it works, after running `go get -u github.com/magefile/mage/...` I am getting the same error as @vassudanagunta.

Before Updating `mage` (with latest source from `master` branch): 
```sh
~GOPATH/src/github.com/gohugoio/hugo master
λ mage hugo

~GOPATH/src/github.com/gohugoio/hugo master
λ mage -version
Mage Build Tool
Build Date: <not set>
Commit: <not set>

~GOPATH/src/github.com/gohugoio/hugo master
λ ./hugo version
[...]

Issue #3997, opened 2017-10-23
Can no longer keep Hugo up-to-date from source!?


galopin:
Mage now fails to update vendored packages ⁉️ 

```
$ go get -u github.com/magefile/mage
$ cd ~/Go/src/github.com/gohugoio/hugo
$ git pull
$ mage install
# command-line-arguments
.\mage_output_file.go:330:17: undefined: mg.Context
Error: error compiling magefiles
```

What's going on?

It seems you have made the [move](https://github.com/gohugoio/hugo/commit/f711b170e4bb286ffa544dfdb48cff4e8f74497f) from `govendor` to the standard Go dependency management tool. But I cannot infer this change is the cause of the problem. Looks like it is a [Mage](/magefile/mage)-only problem…

galopin:
@bep I'm afraid that https://github.com/gohugoio/hugo/commit/45ad8ac381ab71aec6f62bc0842934b8bdecd9d7 does not appear to solve the above issue 🐛  

```
$ cd ~/Go/src/github.com/gohugoio/hugo
$ git pull
$ git log --oneline
45ad8ac vendor: Lock mage to v1
133cfd1 mage: Skip Check on Go 1.8
[…]
```

`mage install` yields the exact same cryptic error message.

bep:
Please use https://discourse.gohugo.io/ for questions/troubleshooting. Also see  [Hugo Documentation](http://gohugo.io/overview/introduction/).

galopin:
Understood but this used to work as described above no later than Friday, i.e. before the https://github.com/gohugoio/hugo/commit/ba0b22ef4491fa70b9f4fc9b091a5192a7eed59f commit… and I also read the friendly manual (#3984). Thank you.

Eventually I was able to rebuild and install Hugo from source by cleaning the whole slate using `go clean -i …` command for both `github.com/kardianos/govendor` and `github.com/gohugoio/hugo` packages — and this time, from the Windows command prompt instead of the Portable Git prompt! 😄 

`Hugo Static Site Generator v0.31-DEV-45AD8AC windows/386 BuildDate: 2017-10-23T16:44:26+02:00`

Kudos for switching to `dep` 🚀

bep:
@galopin the only way this issue tracker is manageable is for us to route everything that resembles a question to the discussion forum.

vassudanagunta:
I'm getting the exact same error: #4003.

PR #4199, opened 2017-12-30
include update flag to mage installation;

yarko:
I had problems with mage which required an update.
After searching around for a while, simply `go get -u...` solved the issue,
so seems reasonable to just include that flag in the basic documentation,
as it is elsewhere on this page.


yarko:
...this should address problems such as #3997 and #4003 - which I also experienced, before realizing an update of mage would solve things...

PR #4203, opened 2018-01-01
Support offline builds

vassudanagunta:
Currently Hugo's mage build targets depend on the Vendor target, meaning that `go get -u github.com/golang/dep/cmd/dep` is executed each and every time. Not only does this unnecessarily slow builds, it prevents builds when one's machine is offline (no internet access). 

Since Mage now supports invoking multiple build targets, there is no longer a need to hardcode the dependency on the Vendor target. For example:

- offline build: `mage Hugo`
- online build, with updated dependencies: `mage Vendor Hugo`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,42 +88,42 @@ new default function more useful for Hugo users.
 Fixes #1949
 ```
 
-### Vendored Dependencies
-
-Hugo uses [Go Dep](https://github.com/golang/dep) to vendor dependencies, but we don't commit the vendored packages themselves to the Hugo git repository.
-Therefore, a simple `go get` is not supported since `go get` is not vendor-aware.
-
-You **must use Go Dep** to fetch and manage Hugo's dependencies.
-
-###  Fetch the Sources From GitHub
+###  Fetching the Sources From GitHub
 
 Due to the way Go handles package imports, the best approach for working on a
 Hugo fork is to use Git Remotes.  Here's a simple walk-through for getting
 started:
 
-1. Install Go Dep and get the Hugo source:
+1. Get the Hugo source:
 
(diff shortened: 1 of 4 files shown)


bep:
Thanks for this. A couple of things:

* Remove the console prints. They were not there originally for a reason, and they're not relevant to this issue.
* Get your commit message(s) in line with the rest

But most importantly:

* Is all use of `mage` in this repo updated to handle the new build targets?

vassudanagunta:
@bep 

I did an exhaustive search for all mage calls and got everything I could find. I just checked again and it doesn't look like I missed anything, though i went ahead and merge two mage calls in `smapcraft.yaml` for consistency. 

I removed the console prints. 

I also updated CONTRIBUTING.md to take advantage of the mage targets and simplify it some. I can simplify it more, but I'll do that in a separate pull request when I get the chance.

bep:
Good. I will probably have to do one more patch release for 0.32 (later today), and prefer to do that from master if I can. So I will merge this after.

vassudanagunta:
@bep README.md

```
To update Hugo’s dependencies, use `go get` with the `-u` option.

    go get -u -v github.com/gohugoio/hugo
```

vassudanagunta:
@bep 

README.md currently says:
```
To update Hugo’s dependencies, use `go get` with the `-u` option.

    go get -u -v github.com/gohugoio/hugo
```

This seems to contradict CONTRIBUTING.md (both before and after the changes in this pull request). Will the above still "build and install the binaries" as the README.md says, or will users have to install and run Mage as CONTRIBUTING.md says? If the latter, I will update the README.

vassudanagunta:
@bep ping... I think you missed my last message above, re README.md...

PR #4208, opened 2018-01-01
tplimpl: Add gtag.js template that is new google analytics tracking code.

nishim:
This PR add new internal template gtag.js template that is new google analytics tracking code.

Currently, [Google Analytics](https://analytics.google.com/) show only gtag.js for tracking code.
I think that should replace `google_analytics.html` implementation in the future. But Google will support both analytics.js and gtag.js tracking code for the while, so I add new template at the moment.

FYI

### DIFF
--- a/docs/content/templates/internal.md
+++ b/docs/content/templates/internal.md
@@ -27,7 +27,7 @@ While the following internal templates are called similar to partials, they do *
 
 ## Google Analytics
 
-Hugo ships with internal templates for Google Analytics tracking, including both synchronous and asynchronous tracking codes.
+Hugo ships with internal templates for Google Analytics tracking, including both [analytics.js](https://developers.google.com/analytics/devguides/collection/analyticsjs/) and [gtag.js](https://developers.google.com/analytics/devguides/collection/gtagjs/) tracking codes.
 
 ### Configure Google Analytics
 
@@ -45,6 +45,7 @@ googleAnalytics: "UA-123-45"
 
 You can then include the Google Analytics internal template:
 
+#### analytics.js
 ```
 {{ template "_internal/google_analytics.html" . }}
 ```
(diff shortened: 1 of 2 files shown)


onedrawingperday:
I just want to note that if this implemented, there can be no options in Hugo's Privacy Config for the gtag.js

> In short, if the admin is using GTM as a one-stop-shop for events and setup of the Universal Analytics Code, the compliance with GDPR falls on his shoulders. He needs to configure it on the GTM workspace. The default code for GTM does not break GDRP, as far as I know.

The quote is from @brunoamaral 's comment over [here](https://github.com/gohugoio/hugo/issues/4616#issuecomment-382089659)

coliff:
@onedrawingperday -its true that gtag.js is designed so that it can be configured using google tag manager site - but it's also possible to add configuration options to the implementation script.

So if this was implemented to Hugo, it'd be possible still to have an `anonymizeIP: true` option in the privacy config for example.
REF: https://developers.google.com/analytics/devguides/collection/gtagjs/ip-anonymization

onedrawingperday:
@coliff Pull Requests are always welcome. 

Personally I'm done with the Hugo simple variants for internal templates, except for the internal shortcodes Speaker Deck and possibly YouTube.

bep:
Sorry for coming late to the party. So this PR in relation to the "privacy settings" we have for GA (as in: is it possible to add for gtag, too):

* [ ] UseSessionStorage  -- not sure if we can disable cookies in gtag ?
* [x] RespectDoNotTrack 
* [x] AnonymizeIP

awhitford:
Google is sending me email to upgrade `analytics.js` to `gtag.js`, but I am using the `_internal/google_analytics.html` template.  So, is the plan to revise the template?

djatwood:
With the merge of #8287 is this PR still necessary?

coliff:
> With the merge of #8287 is this PR still necessary?

I think this can be closed now

PR #4215, opened 2018-01-02
Fix symlink handling in templates

AMDmi3:
There are two logic errors here:
- First, when the template currently being processed is a symlink,
symlink target is retreived from the wrong path - the root path of
template directory
- Second, if the template is really a symlink, regardless of whether
it's target is a directory or a regular file, the processing is
finished. Fix it to only happen if the symlink target is not a regular
file

### DIFF
--- a/tpl/tplimpl/template.go
+++ b/tpl/tplimpl/template.go
@@ -413,9 +413,9 @@ func (t *templateHandler) loadTemplates(absPath string, prefix string) {
 
 		t.Log.DEBUG.Println("Template path", path)
 		if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
-			link, err := filepath.EvalSymlinks(absPath)
+			link, err := filepath.EvalSymlinks(path)
 			if err != nil {
-				t.Log.ERROR.Printf("Cannot read symbolic link '%s', error was: %s", absPath, err)
+				t.Log.ERROR.Printf("Cannot read symbolic link '%s', error was: %s", path, err)
 				return nil
 			}
 
@@ -427,8 +427,8 @@ func (t *templateHandler) loadTemplates(absPath string, prefix string) {
 
 			if !linkfi.Mode().IsRegular() {
 				t.Log.ERROR.Printf("Symbolic links for directories not supported, skipping '%s'", absPath)
+				return nil
 			}
-			return nil
 		}
 
 		if !fi.IsDir() {


AMDmi3:
Ping? IMO this is fairy serious problem and the fix is trivial.

bep:
>Ping? IMO this is fairy serious problem and the fix is trivial.

We need a test to demonstrate the problem.

AMDmi3:
Sure, here's a test repository: https://github.com/AMDmi3/hugo-symlink-bug, it features a valid symlink `layouts/index.html` to `layout/posts/list.html`. The intent is to use posts list as an index page.

### Before fix:

```
Building sites … ERROR 2018/01/15 17:48:57 Symbolic links for directories not supported, skipping '/home/amdmi3/projects/hugo-symlink-bug/layouts'
  
                   | EN  
+------------------+----+
  Pages            |  7  
  Paginator pages  |  0  
  Non-page files   |  0  
  Static files     |  0  
  Processed images |  0  
  Aliases          |  0  
  Sitemaps         |  1  
  Cleaned          |  0  

Total in 23 ms
```

Index page answers with `I am default/list.html` (incorrect).

### After fix:

```
Building sites …
                   | EN  
+------------------+----+
  Pages            |  7  
  Paginator pages  |  0  
  Non-page files   |  0  
  Static files     |  0  
  Processed images |  0  
  Aliases          |  0  
  Sitemaps         |  1  
  Cleaned          |  0  

Total in 20 ms
```

Index page answers with `I am posts/list.html` (as intended).

bep:
>Sure, here's a test repository

Sorry, I meant an **automated test** as part of this PR; we need that to prevent it from breaking again in the future etc. There should be examples to build on.

AMDmi3:
Oh, sorry I misunderstood. On it.

AMDmi3:
Hm, it could be trivial, but tests use [afero](https://github.com/spf13/afero) which doesn't support symlinks, while symlink support is obviously required to do the test.

Should I implement symlink support for afero? It doesn't look too hard, however I've never written in Go before.

AMDmi3:
Ping?

bep:
See https://github.com/gohugoio/hugo/blob/master/hugolib/page_bundler_capture_test.go#L84

bep:
Any update on the testing?

AMDmi3:
Not yet

bep:
I'm sorry for the delay, but this problem should be fixed now.

PR #4216, opened 2018-01-02
Fix index typo

lognaturel:
This could either be "check out Hugo's themes" or "check out the Hugo themes". I think the latter reads better.


lognaturel:
This is a simple typo fix. Looks like I missed the CLA bot message and that might be why it's been sitting, sorry!

PR #4222, opened 2018-01-04
add telegram shortcode

kropp:


docs/content/content-management/shortcodes.md
```diff
@@ -300,6 +300,36 @@ Using the preceding `tweet` example, the following simulates the displayed exper
 
 {{< tweet 877500564405444608 >}}
 
+### `telegram`
+
+To embed a post from public Telegram channel you need just its URL:
```
cosmonawt: This also works with public groups, not only channels.

kropp: Thanks! Updated documentation to reflect this.

Hanzei:
What is the status on this PR?

onedrawingperday:
Just my two cents:

I would think twice or even thrice, before adding more shortcodes to Hugo core.

We already had a very hard time adding Privacy Options to the existing ones and we're still not done.

Telegram is a messaging service. I really cannot see how this could be used widely. I mean, who posts their private messages in public online?

Also if you think that this shortcode is really needed please consider adding Privacy Options for it. See the new [Privacy Config](https://gohugo.io/about/hugo-and-gdpr/) and recent commits at Hugo Master.

jhabdas:
> Telegram is a messaging service.

I put a thumb up here as gists, tweets and instagrams are all also social. From a privacy perspective, I feel confident telegram wouldn't allow _anything, whatsoever_ to be publicly displayed if it were private (such as in the case of a secret chat).

jhabdas:
> Telegram is a messaging service.

I put a thumb up on the addition as gists, tweets and instagrams are all also social. From a privacy perspective, I feel confident telegram wouldn't allow _anything, whatsoever_ to be publicly displayed if it were private (such as in the case of a secret chat).

jhabdas:
https://rawgit.com/theel0ja/e5c0c4f20a02478330be49b4fc6a2214/raw/ceabcfa2980fe7b818f18b0bc361e1f6ef6806da/telegram-widget-demo.html for live examples of the widget output to a page, link borrowed from a [privacybadger issue](https://github.com/EFForg/privacybadger/issues/2029).

Looking at the network waterfall the widget managed to sneak a google analytics tracking request into my browser even though I've explicitly blocked calls to google's tracking domains in my hosts file. This was sneaky and, therefore, I agree with @onedrawingperday Re: the privacy concerns of adding this.

theel0ja:
@onedrawingperday The feature is meant for Telegram channels.

Demo: https://rawgit.com/theel0ja/e5c0c4f20a02478330be49b4fc6a2214/raw/ceabcfa2980fe7b818f18b0bc361e1f6ef6806da/telegram-widget-demo.html

PR #4254, opened 2018-01-10
docs: Add documentation for the .HasShortcode function

rivo:
This function was sorely needed and I was surprised to find out it existed but was not documented. I've added a section in "Creating Your Own Shortcodes" because it's needed when creating shortcodes. I would have added a dedicated method description, too, but I'm not sure where. I found only documentation for variables of pages, not methods of pages.


bep:

rivo:
Ok, thanks. I'll reopen it there. Is there a reason the old docs are still in this repo? I assume they are outdated then?

rivo:
gohugoio/hugoDocs#297

Issue #4268, opened 2018-01-12
Proposal: asset function


nfisher:
It would be nice if there were a function that managed external assets providing the following signature;

```
script  DEST INC_SRI IMMUTABLE [ATTR] SRC...
style  DEST INC_SRI IMMUTABLE [ATTR] SRC...
```
The arguments are as follows;

- **DEST** (string) - base URL relative path.
- **INC_SRI** (boolean) - adds integrity tag with SHA256 sum.
- **IMMUTABLE** (boolean) - adds BASE62(SHA256 sum) as a postfix of the filename prior to the extension.
- **ATTR** - is a map which adds html additional attributes to the tag.
- **SRC** ([]string) - source path for the file(s), multiple will be concatenated.

This would enable the following things;

1. Allow the use of Content-Security-Policy: require-sri-for script style.
2. Fix issues with CDN's that don't cache resources with query params.
3. Allow the assets to be referenced as immutable or at least a very long cache time.
4. Simplify cache busting with one function (e.g. no need to use sha256 and readFile).
5. For pages that are cached the static assets can be made permanently available.

Example Scenario

```Gherkin
Given the user provides the following template sequence:
  {{ script "/js/postman.js", true, true, "static/js/bootstrap.js" }}
When it is processed
Then it should copy the source to the destination folder as named
And it should generate the following tag in the HTML:
  <script type="text/javascript" 
               src="/js/postman-3PVPh0f01T2hWni0XcU5x0jU8Kqdgb64v1avoKGC4G2O.js" 
               integrity="sha256-3vw5dArBhZ2OJ4XtRzIIQJYn6Hrd1fePLeqsuToS1R0="></script>
```

See #621 for discussion on cache-busting.

moorereason:
Please don't put multiple issues into one GH issue.  Rewrite this issue to focus on SRI.

The cache busting enhancement request is #621.

PR #4278, opened 2018-01-15
tpl/tags: Add Script and Style function

nfisher:
Currently a draft idea to add SRI and immutable filenames to
script and style resources.

See #4268


nfisher:
@bep thoughts on this? The failures are the unimplemented functionality. If you're ok with it I'll proceed to complete the functionality. If there's some changes you'd like to see let me know. Otherwise I'll close the related issue and this PR if it doesn't fit inline with the direction you'd like to take Hugo.

bep:
Will look at the corresponding issue, eventually.

nfisher:
After rubber-ducking this API with a former colleague that uses Hugo I think a simplification of the API I initially outlined is worth while.

In this implementation I've removed INC_SRI and IMMUTABLE. If someone doesn't want the SRI I expect they'd use a simple HTML script/link tag to achieve their goal. Will look at a global configuration for the immutable asset postfix.

I could see enhancing this with the following features;

 1. allow custom global version specified from a flag (e.g. git or build number).
 2. disable immutable asset entirely only adding the SRI.

So adding assets is as simple as;

```
{{ script "/js/out.js" nil "static/js/in.js" }}
{{ style "/css/out.css" nil "static/css/in.css" }}
```

And if you want to concatenate multiple files into one;

```
{{ script "/js/out.js" nil "static/js/in0.js" "static/js/in1.js" }}
{{ style "/css/out.css" nil "static/css/in0.css" "static/css/in1.css" }}
```

tpl/tags/tags.go
```diff
@@ -0,0 +1,95 @@
...
+	asset := NewAsset(dest, bytes)
+	url := asset.Url(true)
+
+	// TODO (nfisher 2018-01-25): Might need to introduce a write lock to prevent partial/interleaved writes?
+	err = Prepare(destinationFs, ns.PublishDir, dest)
+	if err != nil {
+		return html, err
+	}
+
+	err = Write(destinationFs, ns.PublishDir, url, bytes)
```
nfisher: @bep what kind of concurrency do I have to worry about here with Hugo? Does afero provide "file locks"? I could push this on a channel and have a single go routine that writes files out and dedups where required.

bep: I think we need to discuss the issue before discussing implementation.

nfisher: Quick background these tags;

 a) generate new files (CSS or JS) via concatenation where multiple SRC are specified so that the number of requests can be reduced where desired.
 b) copy and rename the file contents using DEST as a base and appending the shasum before the file extension for cache-busting purposes.
 c) add a SRI integrity attribute to the resources HTML tag for some measure of increased security where CDN's are used.

The primary motivation for this PR is to minimise hand cranking the SRI and cache-busting tags for every change which are both based on the files shasum/identity. If an integrity attributes shasum doesn't match the shasum of it's contents modern browsers will not load that file. (It's typically employed in combination with CDN's as a security measure.)

I'm currently trying to understanding how Hugo processes each "post" and it's associated templates. I'm assuming it's in parallel. As a result I suspect theres potential data races for who's seen the asset first. That race involves two elements that would require some form of synchronisation/mutual exclusion to ensure correct behaviour;

1. caching shasum calculations.
2. writing to DEST.

The issue I'm concerned with is how to handle one of these functions being placed in a baseof template as an example. If I have a site generating 10s or 100s of posts presumably the baseof template will be evaluated for each post. Cached shasums is easy to handle, stuff it in a lockable map. The bit I'm less certain about is writing the files to disk, currently I'm naively writing the files on every invocation of these functions. I don't see this as optimal because it's inefficient and could lead to data corruption assuming it's possible to have the same file opened by multiple writers, potentially resulting in interleaved writes and corrupt data.

PR #4308, opened 2018-01-21
Tighten page kind logic, introduce tests

vassudanagunta:
`Site.kindFromFilename()` was a bit loose in its logic and inconsistent with `FileInfo.TranslationBaseName()`, which only recognizes language codes explicitly configured for the site. This is illustrated by running the tests I added with the original `page.go` and looking at the failures. It was also a bit inefficient, reparsing file path that had already been parsed before by `FileInfo`. 

I replaced `Site.kindFromFilename(filename string)` with `Site.kindFromFileInfo(fi *fileInfo)` with tighter logic and introduced tests (there was no testing of this logic before).

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -759,7 +759,7 @@ func (s *Site) newPage(filename string) *Page {
 func (s *Site) newPageFromFile(fi *fileInfo) *Page {
 	return &Page{
 		pageInit:    &pageInit{},
-		Kind:        kindFromFilename(fi.Path()),
+		Kind:        kindFromFileInfo(fi),
 		contentType: "",
 		Source:      Source{File: fi},
 		Keywords:    []string{}, Sitemap: Sitemap{Priority: -1},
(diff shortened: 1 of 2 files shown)


vassudanagunta:
I discovered this as I was working on an idea that needs to update the page kind logic.

hugolib/page.go
```diff
@@ -2007,24 +2007,16 @@ func sectionsFromDir(dirname string) []string {
...
-
-func kindFromFilename(filename string) string {
-	if !strings.HasPrefix(filename, regularPageFileNameDoesNotStartWith) && !strings.Contains(filename, regularPageFileNameDoesNotContain) {
-		return KindPage
-	}
-
-	if strings.HasPrefix(filename, "_index") {
+func kindFromFileInfo(fi *fileInfo) string {
+	if fi.TranslationBaseName() == "_index" {
+		if len(fi.Dir()) > 0 {
```
bep: This patch cleans it up nicely, but to check for non-empty strings I think `if fi.Dir() != ""` is better. Looking at your code, while correct, makes me start to think about "what is the importance of the length being greater than something?" `if len(fi.Dir()) != 0` is my second choice, but I prefer the first one.

vassudanagunta: I agree! Coming right up!

PR #4310, opened 2018-01-22
Add external renderer configuration

### LINKED ISSUE #4082: Allow defining asciidoc attributes to pass to invocation of asciidoctor
So asciidoctor has attributes you can set on the command line to set globally

i.e.

:icons: font
:iconsdir: /images/icons

which can be set by in CLI usage like:

asciidoctor -a icons=font -a iconsdir=/images/icons

So instead of putting in the two attributes in every file, it can just be globally set.

Looking at https://github.com/gohugoio/hugo/blob/8717a60cc030f4310c1779c0cdd51db37ad636cd/helpers/content.go#L601

I don't see any way to pass along extra arguments

It would be nice if there was some sort of config level option in hugo to pass extra parameters to external helpers.

unguiculus:
This introduces the option to configure path and arguments for external
renderers (asciidoc, pandoc, and rst). The code makes sure that the
default arguments are always added. 'config.Provider' needs to implements
the additional method 'GetStringSlice' to read the args list from the config.

Fixes #4082

### DIFF
--- a/config/configProvider.go
+++ b/config/configProvider.go
@@ -20,6 +20,7 @@ type Provider interface {
 	GetBool(key string) bool
 	GetStringMap(key string) map[string]interface{}
 	GetStringMapString(key string) map[string]string
+	GetStringSlice(key string) []string
 	Get(key string) interface{}
 	Set(key string, value interface{})
 	IsSet(key string) bool
--- a/docs/content/content-management/formats.md
+++ b/docs/content/content-management/formats.md
@@ -201,12 +201,26 @@ For example, for Asciidoc files, Hugo will try to call the `asciidoctor` or `asc
 
 To use these formats, just use the standard extension and the front matter exactly as you would do with natively supported `.md` files.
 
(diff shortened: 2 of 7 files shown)


bep:
Thanks for this, but we're not merging this. This has been discussed before, the discussion here: https://github.com/gohugoio/hugo/pull/4060

dreamcat4:
Documentation here:

https://gohugo.io/content-management/formats/#additional-formats-through-external-helpers

Requires Hugo version 0.32 or newer.

RMStoica-zivver:
> Thanks for this, but we're not merging this. This has been discussed before, the discussion here: #4060

pandoc has nothing to do with anything. please re-open this

moorereason:
@RMStoica-zivver, the resolution was "we're not merging this."

RMStoica-zivver:
> @RMStoica-zivver, the resolution was "we're not merging this."

With no indication as to why.

Issue #4320, opened 2018-01-24
Zero front matter should be treated identically to Empty front matter 


vassudanagunta:
Currently a file with the contents:
 
    ```
    
    ```
    page with empty front matter
 
is not treated identically to a file with the contents:

    page with no front matter

Specifically, the logic in Page.update() is executed for the former case, but skipped for the latter. At the moment, the only impact is that the `useModTimeAsFallback` config setting is ignored for the page with no front matter. This may be minor, but if proposed new functionality for additional default logic for Date and Title are implemented, it will matter a lot more.

PR #4321, opened 2018-01-24
Add page metadata dates tests

vassudanagunta:
Adds combinatorial tests for the various kinds of dates in Page metadata. Provide:

- needed coverage
- a way to visually review the logic (or illogic) of the current logic
- a way to expose an existing bug #4320
- a framework to easily add tests to cover proposed new
   metadata date sources, and to do so combinatorially with 
   existing sources

Page.go was updated to use a different method to access a file's modification timestamp. This was necessary for the tests to work, but is a better way to access the value anyway.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -1343,11 +1343,8 @@ func (p *Page) update(f interface{}) error {
 	}
 
 	if p.Date.IsZero() && p.s.Cfg.GetBool("useModTimeAsFallback") {
-		fi, err := p.s.Fs.Source.Stat(filepath.Join(p.s.PathSpec.AbsPathify(p.s.Cfg.GetString("contentDir")), p.File.Path()))
-		if err == nil {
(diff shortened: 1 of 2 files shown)


vassudanagunta:
rebased on master

bep:
I assume this is a duplicate PR?

vassudanagunta:
The one you just merged was a child of this one, so GitHub includes the commit from this one in that one (the SHA remains the same). I set it up so you could merge this one first (since the commit is valuable and could stand on its own regardless of whether you accepted the child PR. 

But now it is redundant. I guess GitHub doesn't try to figure that out.

bep:
The correct thing would be to cherry pick this commit into the other PR.

vassudanagunta:
That would take the commit from this PR branch and produce a new commit (new SHA) on the other PR's branch, whereas what I did was branch that PR off of the head of this one. Are you sure cherry picking would do something better?

PR #4322, opened 2018-01-24
Correctly handle page without front matter, fix metadata date bugs

vassudanagunta:
This fixes Issues #4320 and #4323 and adds a nice test function that sets us up well for future additions to metadata logic. 

It depends on Pull Request #4321, which should be merged first.

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -1108,19 +1108,18 @@ func (p *Page) prepareForRender(cfg *BuildCfg) error {
 
 var ErrHasDraftAndPublished = errors.New("both draft and published parameters were found in page's frontmatter")
 
-func (p *Page) update(f interface{}) error {
-	if f == nil {
-		return errors.New("no metadata found")
+func (p *Page) update(frontmatter map[string]interface{}) error {
+	if frontmatter == nil {
+		return errors.New("missing frontmatter data")
 	}
-	m := f.(map[string]interface{})
 	// Needed for case insensitive fetching of params values
-	helpers.ToLowerMap(m)
+	helpers.ToLowerMap(frontmatter)
 
 	var modified time.Time
 
 	var err error
 	var draft, published, isCJKLanguage *bool
-	for k, v := range m {
+	for k, v := range frontmatter {
 		loki := strings.ToLower(k)
 		switch loki {
 		case "title":
@@ -1371,7 +1370,6 @@ func (p *Page) update(f interface{}) error {
 	p.params["iscjklanguage"] = p.isCJKLanguage
 
 	return nil
-
 }
 
 func (p *Page) GetParam(key string) interface{} {
@@ -1614,14 +1612,12 @@ func (p *Page) parse(reader io.Reader) error {
 	if err != nil {
 		return fmt.Errorf("failed to parse page metadata for %q: %s", p.File.Path(), err)
 	}
-
-	if meta != nil {
-		if err = p.update(meta); err != nil {
-			return err
-		}
+	if meta == nil {
+		// missing frontmatter equivalent to empty frontmatter
+		meta = map[string]interface{}{}
 	}
 
-	return nil
+	return p.update(meta)
 }
 
 func (p *Page) RawContent() string {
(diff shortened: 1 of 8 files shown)


hugolib/page.go
```diff
@@ -1115,215 +1115,213 @@ func (p *Page) prepareForRender(cfg *BuildCfg) error {
...
-	// Needed for case insensitive fetching of params values
-	helpers.ToLowerMap(m)
-
 	var modified time.Time
-
-	var err error
 	var draft, published, isCJKLanguage *bool
-	for k, v := range m {
-		loki := strings.ToLower(k)
-		switch loki {
```
bep: This switch needs to be on the same indentation level as before (first level). So, either return early (preferred), or if that is not possible for some reason: Extract method or func.

bep: Other than that it looks good -- but see the commit message guidelines.

vassudanagunta: Returning early defeats the purpose, as some of the logic at the end of the function must execute whether or not there is front matter. So I tried refactoring this function into two: processFrontmatterMetadata and processDerivedMetadata. But the problem is that the CJKLanguage detection logic depends on knowledge of what happened up top (hence the special treatment with the *isCJKLanguage var).

```
	if isCJKLanguage != nil {
		p.isCJKLanguage = *isCJKLanguage
	} else if p.s.Cfg.GetBool("hasCJKLanguage") {
		if cjk.Match(p.rawContent) {
			p.isCJKLanguage = true
		} else {
			p.isCJKLanguage = false
		}
	}
	p.params["iscjklanguage"] = p.isCJKLanguage
```

So the extraction is possible, but quite ugly.

vassudanagunta: Ok I just realized there is another approach: always pass in a frontmatter map to this method, an empty one in the case that there is no front matter. Does that work for you? Essentially, which do you prefer:

1. Keep the big monolithic function we have now (switch restored to first level), which more easily allows for coupled logic as we have now, or,
2. Refactor it into processFrontmatterMetadata and processDerivedMetadata functions, but with some strange (looking) args such as `didFrontMatterSpecifyIsCJK`.

bep: Empty map in works, and kind of matches the method signuature.

vassudanagunta:
@bep good to go.

hugolib/page.go
```diff
@@ -1114,11 +1114,11 @@ func (p *Page) prepareForRender(cfg *BuildCfg) error {
 
 var ErrHasDraftAndPublished = errors.New("both draft and published parameters were found in page's frontmatter")
 
-func (p *Page) update(f interface{}) error {
-	if f == nil {
-		return errors.New("no metadata found")
+func (p *Page) update(frontmatter interface{}) error {
```
bep: why is frontmatter an `interface{}` and not `map[string]interface{}`?

vassudanagunta: That's what it was before / I didn't change it. I'm a novice go programmer, so I don't change things I don't need to yet :)

vassudanagunta: Want me to change it now?

bep: OK, no just leave it.

bep: But you need to do something about the commit messages. 5 commits is maybe too much, too, but the important part is the commit message title (we reference the issues in the body with "Fixes #<some-id>, See #<some-id>.

moorereason: I'd recommend updating the function signature to use `map[string]interface{}` instead of `interface{}`.  We type-assert as `map[string]interface{}` at L1121 and don't handle any assertion failures there.

bep: >I'd recommend updating the function signature to use map[string]interface{} instead of interface{}

We will consider that for the future, not this PR.

vassudanagunta: How about I squash it down to 2 commits (add test coverage + fix)? And since I'm doing that anyway I'll go ahead and change the method signature.

vassudanagunta: So ditch Pull Request #4321 (adds pure test coverage)? If we keep it, then I can get the total (counting #4321) down to 3 commits. I don't care either way.

bep: No, keep the code, tests included. This is not a "squash everything" game;  just get the commit messages in line with the guidelines (re my comment above).

vassudanagunta:
@bep Reabased on master and good to go. Changing the method signature cascaded to many other files. But a good thing as type safety is improved. I just had one question of curiousity asked in code review above.

Issue #3805, opened 2017-08-15
Use filename as title when it’s not defined in front matter


idealhack:
Note: I'm new to Hugo, so If there's a better way about this fell free to correct me.

For now I use `{{.File.TranslationBaseName}}` in template cause `{{.Title}}` is empty string. I thought it's better to store it in the title variable, as themes out there are already using it. Also we could use file dates for `date` and `lastmod`.

I thought this is a nice feature especially when there're a lot of content files imported from another system.

For myself, I use [nvALT](http://brettterpstra.com/projects/nvalt/) to take markdown notes. nvALT has already handle the title and date in this way, now I want to use Hugo for building the website. But maintaining notes metadata by hand is not a good practice IMHO.

What do you think?

> This was originally posted at https://discourse.gohugo.io/t/use-filename-as-title-when-its-not-defined-in-front-matter/7965

maxandersen:
I got a patch that makes this happen, but dependent on the changes in https://github.com/gohugoio/hugo/pull/3762 that in addition to deriving date from filename also fixes how metadata defaults are calculated which is needed to fix title handling. If that PR gets merged in I can easily make a PR for this issue too.

maxandersen:
note I think its worth going a level further and try detect if content has a top level header as defined by markdown, asciidoc or orgmode and use that header if present, if not use filename.

bep:
>note I think its worth going a level further and try detect if content has a top level header as defined by markdown, asciidoc or orgmode and use that header if present,

No, that is too hacky and not worth it.

maxandersen:
> No, that is too hacky and not worth it.

too hacky? as compared to try derive from filename ? 

It works pretty well in awestruct and others and it has the big plus your blog posts actually render with proper title when shown "raw" in github.

maxandersen:
btw. made the patch before I saw your comment so you can see the idea at https://github.com/gohugoio/hugo/pull/3859

maxandersen:
note, the idea is similar to extract toc and summary that tries to parse raw and generated content to derive content.

bep:
Note that the current TOC solution is also considered "hacky and slow" in my book, and will be fixed once Blackfriday 2 is out. We're not doing any regexp magic looking for titles.

maxandersen:
> Also, I'm not a big fan of pull requests that try to do more than the issue they describe. Do one thing well.

okey - I've updated patch to just to title based on filename/slug.

[...]

Issue #3977, opened 2017-10-17
Future content issue with `date` and `publishdate` params


MunifTanjim:
**Hugo version**: 0.30

From the [Hugo CLI documentation](https://gohugo.io/commands/hugo/):
> -F,  --buildFuture             include content with publishdate in the future

## Previous Behavior

- Hugo will render content with `date` param set to a future date by default.
- Hugo will render content with `publishdate` param set to a future date only with `-F` or `--buildFuture` flag enabled.

## Current Behavior

But with the 6a30874f19610a38e846e120aac03c68e12f9b7b commit, the value of `date` & `publishdate` params will be same if one is absent in the front-matter.

So, content with `date` param set to a future date will also be considered as future content and will not be rendered by default.

## Issue

Clearly one of the mentioned behavior is/was unintended (probably the current behavior, as I saw that the above-mentioned commit fixes an unrelated problem (#3854).

What if someone wants to publish a content with a future date before the date has arrived?

bep:
>What if someone wants to publish a content with a future date before the date has arrived?

I made this change, and I may have stepped on someone's toes here. But the front matter "date logic" is pretty elaborate, and there were several holes in it.

So I still think 0.30 does it right.

So, 

* `date` is like the created or authored timestamp.
* 'publishdate` is when it should be published.

To publish content with a future author date:

*  Current date is `2017-17-10`
* You want to publish an article today with author-date `2017-10-20` (not sure why)
* You should be able to set  `date=2017-10-20` **and** `publishDate=2017-10-17`

So, the change in 0.30 meant that you for the use case above may have to set 2 dates and not just one, which makes sense in my head. Less magic.  The important part is that Hugo behaves oddly without a date ... and now it is more likely that we get one.

git001:
We have always such a combination on frontmatter.

```
...
publishdate: 2017-02-14
date: 2017-02-14
...
```

and in the ` layouts/shortcodes/lastmod.html ` this code

```
{{- if $.Page.Params.LastMod -}}
{{- dateFormat ( or $.Site.Params.dateFormat "2006, Jan 02" ) $.Page.Params.Lastmod -}}
{{- else -}}
{{- dateFormat ( or $.Site.Params.dateFormat "2006, Jan 02" ) $.Page.Params.PublishDate -}}
{{- end -}}
```

We choose the combo above because the `LastMod` was not available everytime.

@MunifTanjim please correct me if i'm wrong

jcw:
[...]

PR #4340, opened 2018-01-28
WIP: discuss and cleanup date handling

vassudanagunta:
I opened this PR to expose and support discussion about peculiarities in the default date metadata logic. 

The PR initial commit includes improved test code that:
- makes it easier to examine current behavior
- adds more coverage both to expose more of the current behavior as well as to help Hugo stability (avoid unintentional behavioral changes).
- lays the groundwork for testing new options for default date values

## date peculiarities

I believe as Hugo heads toward v1.0 it needs to get tighter with respect to both coherence and stability of behavior. This cases this issue covers is a good example of where Hugo could use some tightening up. 

Take a look at [the test case table](https://github.com/vassudanagunta/hugo/blob/1cea4d37e851c04bc47c329b2aa6806665275a3b/hugolib/page_frontmatter_test.go#L207-L288), which includes instructions on how to read it. Each of the test cases pass for the current Hugo release and thus represent current Hugo behavior.

1. The default value for four dates is ''Jan 1, 0001", and for the fifth its `nil`.  I'd think `nil` makes most sense, but in any case it should be consistent. (See test cases [`N1` and `N2`](https://github.com/vassudanagunta/hugo/blob/ec2cdb95214482b7a46956ba21ddfb34cf67ecb1/hugolib/page_metadata_test.go#L243-L244))
2. If the frontmatter specifies `modified` , it fills both `LastMod` and `Modified` page values. But if the frontmatter specifies `lastmod`, it fills only to the `LastMod` page value. ([`1L`, `1M`](https://github.com/vassudanagunta/hugo/blob/ec2cdb95214482b7a46956ba21ddfb34cf67ecb1/hugolib/page_metadata_test.go#L248-L249), [`4M` and `4E`](https://github.com/vassudanagunta/hugo/blob/ec2cdb95214482b7a46956ba21ddfb34cf67ecb1/hugolib/page_metadata_test.go#L259-L260))
3. The `useModTimeAsFallback` config setting affects `Date` and `LastMod`, but not `Modified`. Do we really want `Date` to be set by something as volatile as the file timestamp? Git, for example, doesn't preserve them. Doesn't it make more sense for user templates to make such determination? As long as  `useModTimeAsFallback`  sets `LastMod`, templates have the option to use it when `Date` is not specified.  ([`M0`](https://github.com/vassudanagunta/hugo/blob/ec2cdb95214482b7a46956ba21ddfb34cf67ecb1/hugolib/page_metadata_test.go#L265))
4. Normally when `Date` has a value but `PublishDate` doesn't, `PublishDate` defaults to `Date`'s value. But when `Date` is set by `useModTimeAsFallback`, this doesn't happen. I think this makes sense because `PublishDate` shouldn't be set by something as volatile as the file timestamp. But in the same vein why does `useModTimeAsFallback` beat out a `lastmod` explicitly set in the front matter as a default value for `Date`? ([`ML`](url))  The right thing happens in the other direction: frontmattter `date` overrides `useModTimeAsFallback`  to fill out `lastMod`. ([`MD`](https://github.com/vassudanagunta/hugo/blob/ec2cdb95214482b7a46956ba21ddfb34cf67ecb1/hugolib/page_metadata_test.go#L268)) 
5. Maybe my Spidey Sense is wrong, but the way either one of `date` and `publishdate` front matter will fill three page date values feels haphazard. ([`1D` and `1P`](https://github.com/vassudanagunta/hugo/blob/ec2cdb95214482b7a46956ba21ddfb34cf67ecb1/hugolib/page_metadata_test.go#L246-L247), [`21`-`23`](https://github.com/vassudanagunta/hugo/blob/ec2cdb95214482b7a46956ba21ddfb34cf67ecb1/hugolib/page_metadata_test.go#L252-L254)) I can see that it is helpful for some people, for lazy templates, but should this be hardcoded Hugo behavior? There's a reason we have three different fields, right? Wouldn't it be better for templates to be explicit about such things? Wouldn't it be better to support something like {{ .LastMod|.Date|.PublishDate }} in templates to support the exact same precedence logic or custom one rather than hardcoding it in Hugo? My observation is that someone opens an issue requesting that such and such date default to some other date, it sounds reasonable, and in the logic goes. But over time, these changes that feel right on their own don't make for a clear bigger picture.

### solutions

We could update the logic right here in this PR. But my sense is that there will be no way to please everybody. But there are at least a couple of other options to consider:

-  Something like what I propose above: a way for templates to say: here is a list of variables in precedence order. Give me the value of the first non-nil one. 
-  [Design Proposal: defaultDate & defaultTitle](https://github.com/gohugoio/hugo/issues/4341). It addresses #285 and #3805 and provides a way out of the above wonkiness.

But if the decision is to leave things as they are, I'd still like this PR merged because the tests are much better. Not now though. Let's have a test code review-based discussion.


bep:
Hugo is in some sense legacy software. Code with a long history will have parts that when looking at it now, have room for improvement. But don't use the word "wonky" and other negative loaded words to describe other people's work if you expect to start a serious discussion about improving it.

bep:
As to the PR itself:

* The way of testing this, with front matter variations and then create pages from it, was probably convenient at some point.
* But if you want full test coverage for all of these front matter variations, this gets wordy very fast
* And assuming that we have plenty of front matter unmarshal tests in other places, we should try to get proper unit tests for this

But that requires refactoring of the code and not just adding more tests.

bep:
As to the PR itself:

* The way of testing this, with front matter variations and then create pages from it, was probably convenient at some point.
* But if you want full test coverage for all of these front matter variations, this gets wordy very fast
* And assuming that we have plenty of front matter unmarshal tests in other places, we should try to get proper unit tests for this

But that requires refactoring of the code and not just adding more tests.

If you want to clean up the code/tests, that would be appreciated. I have looked at your numbered list, and in general, I don't agree.

bep:
>I have looked at your numbered list, and in general, I don't agree.

Some details on the above:

* I'm a little confused about your comments about values for Page date fields that do not exist, e.g. `Modified`, but I guess you refer to how we pass on values to the `.Params` object. 
* Dates are really important to Hugo, so we have date adapters for people coming from Jekyll etc. We need to use those dates during the build (for sorting, expiry date, publish date), so we cannot push that logic to the templates. Besides, that template logic would be ... not pretty.
* You could probably argue that we should ask people to get their data correct and just use those values literally (date, lastMod etc.), but after enough of the same questions on "why my site doesn't work as expected", you kind of give in to some compromises.
* I agree about the not-so-usefulness of `useModTimeAsFallback`, but it is there, and some people are using it, so removing it would create some noise. Not sure it's worth it. If `useModTimeAsFallback` beats `lastMod` in front matter, that is an obvious bug and should be fixed.

bep:
But thinking about it, `useModTimeAsFallback` does work as originally designed. But see https://github.com/gohugoio/hugo/issues/4351

vassudanagunta:
I apologize if I offended by use of the word "wonky". But in America at least it simply means "strange" or "feels off". I am not criticizing the project or anyone's work, just giving my honest observation of the current state of something. I don't have all the knowledge that you do about the history of Hugo and how various behavior evolved. But in my experience feedback from the fresh eyes of a newcomer can be valuable. And easy to discard if wrong.  

The funny thing is that my recent PRs and Issues are really just offshoots of the real work I'm doing. What I'm after is for Hugo to **properly support content without front matter**. 

That in turn is an offshoot of [a standards proposal I'm working on](https://github.com/Markaround/markaround), which I plan to demo with a version of Hugo I've been implementing in a fork. Whenever I come across something that seems like it has universal benefit to Hugo, I think to offer it upstream. Hence PR #4322, which I preceded with PR #4321 to be a good citizen. Producing those tests revealed two bugs which I fixed in #4322, but also triggered my observations in this PR. 

With regards to my observations, I personaly don't have a dog in this race. I simply am offering them up for consideration. I believe in simple, coherent and intuitive data models and software interfaces. But I also know I am ignorant of Hugo's history and use cases, and I know current logic is an artifact of compromise, that maybe these compromises are now locked in, and maybe they are not even a big deal. In that case ignore my observations or file them away, but please merge this pull request because the tests provide better coverage and are easier to understand, and they provide a form of documentation of these compromises. They guards against unintentional changes.

vassudanagunta:
> Deprecate useModTimeAsFallback #4351

Are you sure you want to do this? See #2239.

vassudanagunta:
See also #3977, which just fell of the map. @bep your last comment there (3 months ago) implied a logic different from what we have today.

Issue #4298, opened 2018-01-19
Avoid panic when defaultContentLanguage does not exist


bep:
See https://discourse.gohugo.io/t/problem-when-updating-from-0-31-to-0-32-and-newer-panic-proc-not-found/10083

```
defaultContentLanguage = "en"
```

But there is no language definition for `en` in the project.

We should add a validation for this and give a clear and exit with an ERROR message.

The workaround for the thread starter is to change the above to `pl` (a valid and existing language code).

PR #4342, opened 2018-01-28
Lowercase default lang when checking if it exists

kyoshino:
Fix a regression from #4298. `lang.Lang` here is lowercased so the matching doesn’t work when the `defaultContentLanguage` config is written in the `ab-CD` style.

---

This is my first contribution to Hugo and first Go lang coding so I might be wrong. My site can't be built with the current version 0.34 due to the `en-CA` language code ([config](https://github.com/fxsitecompat/www.fxsitecompat.com/blob/master/config.yaml)). Actually Hugo has more issues on handing of the `ab-CD` style language code (see [hacks in my templates](https://github.com/fxsitecompat/www.fxsitecompat.com/search?q=replace+en-ca)) but those are separate issues. I'd like to see this regression fixed first. Thanks!

### DIFF
--- a/hugolib/config.go
+++ b/hugolib/config.go
@@ -146,9 +146,10 @@ func loadLanguageSettings(cfg config.Provider, oldLangs helpers.Languages) error
 
 	// The defaultContentLanguage is something the user has to decide, but it needs
 	// to match a language in the language definition list.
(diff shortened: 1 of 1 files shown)


kyoshino:
Ugh, sorry but I can't sign the Agreement. Will file an issue instead.

PR #4346, opened 2018-01-28
[Fix #4339] Render changed files in Fast Render Mode.

### LINKED ISSUE #4339: Render new content in "Fast Render Mode"
Fast Render Mode is a great feature, but I faced with some usability problems with it: when you have a server running and create a new content page, it is added to index (as far as index is always rerendered), but if you try to navigate to this new page, you got 404. I understand the idea behind the Fast Render Mode to render only last visited pages, but there is no way to visit new page, because it's just created.

Even more confusion if you run dev server with `--navigateToChanged` flag, than you instantly redirected to 404 when creating new content page.

So my suggestion is: maybe it makes sense to add just created pages to filter of "Recently Visited" and render them too?

If you think it's a good idea, I can prepare pull request for this feature.

yanzay:


bep:
Thanks for this, much appreciated. This has been bugging many (me included). I have merged this manually, and added a couple of commits on my own. See the commit log in master.

Issue #4341, opened 2018-01-28
Design Proposal: defaultDate & defaultTitle


vassudanagunta:
I changed my mind. I now like @maxandersen's improvement (#3762) of @devjack's implementation (#3310) of @xaprb's now nearly 4 year old request (#285). A single config value. A simple bool that enables "smart" date detection (that could get smarter over time), rather than asking the user to figure out a regex:

```
 defaultLayout                        string
 useModTimeAsFallback                 bool  <--- existing, by #2239
 useFilenameDateAsFallback            bool  <--- proposed by #3762
 useFilenameOrHeadingAsTitleFallback  bool  <--- implied by #3805 
 defaultContentLanguage               string
 defaultContentLanguageInSubdir       bool
```

## concerns

-  All the other settings that control defaults are prefixed with "default". These two (three) aren't.
-  What if `useModTimeAsFallback` and `useFilenameDateAsFallback` are both set to TRUE?
-  Do we always extract it (so it is not part of slug) or will some people want it to remain? What if I have dates in all my filenames, but in some cases have the date in the front matter? The result will be some slugs will have dates and others won't. Is the solution then to add another config item, `keepDateInSlug`? Hmmm.
-  What if it should be PublishDate, not Date that comes from the filename? 

## an alternate solution

This proposal address #285, #3805 and #4340 and all of the concerns above. Compare the set of settings above to this:

```
defaultLayout                  string
defaultDate                    string  <--- for #285
defaultTitle                   string  <--- for #3805
defaultContentLanguage         string
defaultContentLanguageInSubdir bool
```

### defaultDate

`defaultDate` has the following symbolic values (final identifiers may be different):

-  `none`
-  `in filename`    &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;  same as #3762
-  `remove from filename` &nbsp;&nbsp;   same as `in filename`, but removes date from slug
-  `file mod time`   &nbsp;&nbsp; &nbsp;&nbsp;&nbsp;   same as `useModTimeAsFallback`, which we can deprecate
-  `file creation time`   &nbsp;&nbsp;   future, if needed

In the future we can also distinguish between Date and PublishDate if necessary:

```
defaultDate="file mod time"
defaultPublishDate="in filename"
```

We can also use this setting to address #4340, replacing the currently hardcoded logic for defaulting Date, PubDate and LastMod to each other (which I think is problematic). We simply add a couple more symbolic values for the setting:

```
[...]

PR #4359, opened 2018-01-31
WIP: Use top-level markdown heading as default page Title

vassudanagunta:
Supports content where the title is declared as the top level heading within the markdown itself rather than in front matter. Sets the Page's title variable only if it was not set via front matter. See [this comment](https://github.com/gohugoio/hugo/issues/4341#issuecomment-361804082) under #4341.

This is a POC/WIP. It is fully functional. It is implemented not by a regex scan of the content, but via a hook into the BlackFriday renderer. As I am still new to the Hugo codebase, I'm not sure if I picked the right place to code the hook and pass it into BlackFriday. 

Since the hook is intercepting all BlackFriday heading render calls, it can also form the basis of a better TOC solution, with a proper data structure (I know @bep wants this!)

There are not yet config settings or tests, but need confirmation that this is indeed desired functionality first.

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -445,6 +445,9 @@ type RenderingContext struct {
 	Config       *BlackFriday
 	RenderTOC    bool
 	Cfg          config.Provider
+	// handle the first level-one header,
+	// return whether or not to remove it from the page body
+	HandleTitle func(title string) bool
 }
 
 // RenderBytes renders a []byte.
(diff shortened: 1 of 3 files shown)


bep:
The idea is good, but doing this in Blackfriday 1 creates work down the line when we're going for v2, which is a daunting task on its own.

vassudanagunta:
Makes sense. If Blackfriday 2 gets us a nice TOC data structure, we could just pull the title out of that.

bep:
>If Blackfriday 2 gets us a nice TOC data structure, we could just pull the title out of that.

It doesn't out of the box, but the API is totally different, so this PR will have to be somehow adapted/rewritten.

vassudanagunta:
Ok, I'll close it. It will still exist for reference by the other discussions on metadata defaults.

PR #4371, opened 2018-02-01
Update "Contributors" link in site-footer.html

celtic-coder:
The "Contributors" link is correct in the Release Notes documents, for example, but is incorrect in the site footer.

### DIFF
--- a/docs/themes/gohugoioTheme/layouts/partials/site-footer.html
+++ b/docs/themes/gohugoioTheme/layouts/partials/site-footer.html
@@ -3,7 +3,7 @@
       <div class="pb3 pt4 w-100 w-50-ns">
       
          <div class="b f3  light-gray mb3 nested-links tc">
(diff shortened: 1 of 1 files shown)


vassudanagunta:
Does it matter? The two links go to the same place.

celtic-coder:
If they both will have the same destination going forward, then it is not a big deal. However, it is always best to keep links consistent in order to avoid possible broken links in the future.

In this case, it looks like the Release version is the most up-to-date as it is also used in other places in GitHub. The lo

bep:
Sorry for the delay on this. I would prefer to get theme fixes in the theme itself. Which would remove the need of me having to forward this patch to the theme.

https://github.com/gohugoio/gohugoioTheme

celtic-coder:
Hi Bjørn, thanks for the suggestion! I will take a look at adding the patch.

Issue #4366, opened 2018-02-01
Array data source in the path of another can result in golang Panic


vassudanagunta:
When two source data files at the same level in the Site.Data tree contain array data at the top level, such as the following:

`themes/mytheme/data/a/b1.json`:
```json
["x", "y", "z"]
```

`data/a/b1.json`:
```json
["1", "2", "3"]
```

...the following error occurs: 

> PANIC: interface conversion: interface {} is []interface {}, not map[string]interface {}

The expected behavior is that `Site.Data.a.b1` = `["1", "2", "3"]`

The culprit is the `Site.handleDataFile(...)` in `site.go`, in which the following section of code checks for pre-existing data at a node in the tree and incorrectly assumes it is a map node: https://github.com/gohugoio/hugo/blob/19e26cb4ebdc17ad190a0058b347ac46ed75f4c8/hugolib/site.go#L815-L817

This is a pre-existing bug that goes back to at least 2015. 

Add the following test to `datafiles_test.go` to uncover (and cover) this case:

```golang
func TestDataDirMultipleSourcesArrayFile(t *testing.T) {
	t.Parallel()

	sources := [][2]string{
		{filepath.FromSlash("themes/mytheme/data/a/b1.json"), `["x", "y", "z"]`},
		{filepath.FromSlash("data/a/b1.json"), `["1", "2", "3"]`},
	}

	// Per handleDataFile() comment:
	// 1. A theme uses the same key; the main data folder wins
	// 2. A sub folder uses the same key: the sub folder wins
	expected, err := parser.HandleJSONMetaData([]byte(`
{ "a" : {
        "b1": ["1", "2", "3"]
        }
}`))

	if err != nil {
		t.Errorf("Error %s", err)
		return
	}

	doTestDataDir(t, expected, sources,
		"theme", "mytheme")
}
```

(bug uncovered while investigating #4361.

Issue #4361, opened 2018-01-31
regression: Fail to unmarshal JSON arrays


JoshStrobl:
As of Hugo 0.35, I am no longer able to parse a valid JSON file for use in `.Site.Data.var`. Running `--debug` reports:

``` bash
WARN 2018/01/31 17:26:12 Failed to read data from mirrors.json/mirrors.json: json: cannot unmarshal array into Go value of type map[string]interface {}                                                
WARN 2018/01/31 17:26:12 Failed to read data from sws.json/sws.json: json: cannot unmarshal array into Go value of type map[string]interface {}
```

`mirrors.json` is saved in `data/`, with the contents of:

``` JSON
[
	{
		"Latitude": 52.3702,
		"Location": "Amsterdam",
		"Longitude": 4.8952,
		"Type": "official",
		"Url": "https://stroblindustries.com/isos/"
	},
	{
		"Latitude": 33.753746,
		"Location": "Atlanta",
		"Longitude": -84.386330,
		"Type": "official",
		"Url": "http://us.justin.id.au/iso/"
	},
	{
		"Latitude": 52.5200,
		"Location": "Germany",
		"Longitude": 13.4050,
		"Type": "community",
		"Url": "http://solus.veatnet.de/iso/"
	},
	{
		"Latitude": 51.507351,
		"Location": "London",
		"Longitude": -0.127758,
		"Type": "official",
		"Url": "https://soluslond1iso.stroblindustries.com/"
	},
	{
		"Latitude": 40.357298,
		"Location": "New Jersey",
		"Longitude": -74.667223,
		"Type": "community",
		"Url": "http://mirror.math.princeton.edu/pub/solus-iso/"
	},
	{
		"Latitude": 43.086002,
		"Location": "Rochester",
		"Longitude": -77.674538,
		"Type": "community",
		"Url": "https://mirrors.rit.edu/solus/images/"
	},
	{
		"Latitude": 37.7749,
		"Location": "San Francisco",
		"Longitude": -122.4194,
		"Type": "official",
		"Url": "https://solussf1iso.stroblindustries.com/"
	}
]
```

`sws.json` is saved in `data/`, with the contents of:

``` JSON
[
	{
		"episode": "10",
		"description": "The team talks about work done since the snapshot 2017.04.18.0, ferryd, cuppa, plans for the new site, and more.",
		"date": "Fri, 14 Apr 2017 01:33:12 +0300",
		"length": [
			"21474938",
			"43046701"
		],
[...]

Issue #3890, opened 2017-09-18
Allow other types than maps in .Site.Data


felicianotech:
The following example YAML file cannot currently be used with Hugo:

```
# ./data/books.yml
- title: "Harry Potter and the Order of the Phoenix"
  author: "J.K. Rowlings"
  pages: 870
- title: "A Wrinkle in Time"
  author: "Madeleine L'Engle"
  pages: 240
- title: "My Side of the Mountain"
  author: "Jean Craighead George"
  pages: 175
```

This is because the root node here would be a YAML sequence (a.k.a a list), which is valid YAML but Hugo doesn't currently support.

This Issue originated in [Hugo Discussion](https://discourse.gohugo.io/t/using-a-list-as-the-root-node-of-a-yaml-data-template/8401).

zealic:
I has tested data file format support for array type:

- [x] JSON
- [ ] YAML
- [ ] TOML

PR #4372, opened 2018-02-02
Increase data directory test coverage

vassudanagunta:
Created these tests as I was trying to solve a number of related issues (see below). A comprehensive fix is not trivial because Hugo uses three different three unmarshalers that behave differently for the same data structures. I will likely revert to fixing the regression first before making another attempt at the big picture.  To keep changes incremental and easier to review, I'm pushing these test changes first, and **I'm initially basing them on the v0.34 branch** to demonstrate that the new tests all pass for that release. 

@bep or @moorereason, please review this test code. You can merge it onto v0.34 if you wish, but not onto v0.35, as a couple of the new tests will fail there. Otherwise I'll include this commit along with a fix on a separate PR to master. 

In this commit:
*  Adds retro-coverage for #4361 (regression introduced in v.035)
*  Verifies open issues #4138, #3890, #4366, #4083
*  Removes test reliance on the very code it is testing (hugo/parser package).
   Expected results are now all built manually / are more precise.
   Tests can run against different versions (no linkage errs)

### DIFF
--- a/hugolib/datafiles_test.go
+++ b/hugolib/datafiles_test.go
@@ -26,8 +26,9 @@ import (
 	"github.com/gohugoio/hugo/deps"
 	jww "github.com/spf13/jwalterweatherman"
 
-	"github.com/gohugoio/hugo/parser"
+	"fmt"
 	"github.com/stretchr/testify/require"
+	"runtime"
 )
 
 func TestDataDirJSON(t *testing.T) {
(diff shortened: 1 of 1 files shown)


vassudanagunta:
@bep Since this PR was just to demo the new tests against v0.34, and since #4373 includes these new tests plus a v0.35 fix, you can just close the PR if and when you merge that one.

Issue #2481, opened 2016-09-19
Add anchor in place of <!--more-->


edio:
Articles generated by Hugo miss anchor to scroll to the place of the article where summary ends.

Use-case: I read article on the main page, click lint to read more.
I expect to continue reading from the place where summary ended.
But instead article is opened from the beginning.

I'm thinking about fixing this by generating html in place of  tag
`
<a name="_summary-end"></a>
`
The name is yet to discuss. It could be `_cut` , `_readmore`, or something like that.

Then in index page it'll be
`
<a href="{{ .RelPermalink }}#_summary-end">Read More…</a>
`

Or it can even be a new variable `{{ .ReadMoreLink }}`.

I'd be willing to implement, but at first, what do you think of that feature?

bep:
I think this should be solved by adding a `ContentAfterSummary` field on `Page`. There is a issue about that somewhere -- there was even a PR at some point, but it never finished. After my latest summary refactorings, such a field should be trivial to add.

Please shout if there are issues here I'm not getting, but in my head this is a template responsibility, son in `single.html`:

```html
{{ .Summary }}
 <a name="_summary-end"></a>
{{.ContentAfterSummary }}
```

bep:
This issue has been automatically marked as stale because it has not been commented on for at least four months.

The resources of the Hugo team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still valuable, please open a proposal at https://discuss.gohugo.io/.

This issue will automatically be closed in four months if no further activity occurs. Thank you for all your contributions.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

PR #4385, opened 2018-02-07
Add ContentAfterSummary property to Page

### LINKED ISSUE #716: A way to access content after <!--more--> marker

edio:
Fixes #716 #2481

I'm not sure at all about `rst` markup, I just made pass existing tests without true comprehension. Please pay attention to that part during review

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -127,9 +127,10 @@ type Page struct {
 	params map[string]interface{}
 
 	// Content sections
-	Content         template.HTML
-	Summary         template.HTML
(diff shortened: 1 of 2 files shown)


edio:
In case if anyone is interested in demo, I generated content of my blog (link in profile) with the code form this branch.

PR #4388, opened 2018-02-08
Fix handling of taxonomy terms containing slashes

### LINKED ISSUE #3577: removePathAccents true leave dot and forward slash in permalinks
I have this permalink set up for posts.

    permalinks:
      post: "/post/:year/:month/:day/:slug"

If no slug is preset the post title will be used instead, and that is so for more or less all my posts.

Together with "removePathAccents: true" I get a problem with characters like a dot or forward slash in the posts title that does not get replaced in the generated permalink for the post. I have solved it for now by adding a slug to the post with problematic titles.

This example title

    Lörem ipsum dölor sit amet/consectetur do example.org

generates the following permalink

    /post/2006/02/02/lorem-ipsum-dolor-sit-amet/consectetur-do-example.org/

I had hoped it would become

    /post/2006/02/02/lorem-ipsum-dolor-sit-amet-consectetur-do-example-org/

If I set "removePathAccents: false" I get

    /post/2006/02/02/lörem-ipsum-dölor-sit-amet-consectetur-do-example-org/

but I would prefer to have åäö replaced with ago in permalink.

### LINKED ISSUE #4090: Slashes in a taxonomy term break path resolution

moorereason:
This PR attempts to resolve #4090.  The first commit contains tests that demonstrate the problem.

Special thanks to @tsuereth for his initial work on #4092, which laid the groundwork for this PR.

This PR may fix the title issues in #3577, but I haven't tested it yet.

### DIFF
--- a/helpers/path.go
+++ b/helpers/path.go
@@ -74,6 +74,35 @@ func (filepathBridge) Separator() string {
 
 var fpb filepathBridge
 
+// segmentReplacer replaces some URI-reserved characters in a path segments.
+var segmentReplacer = strings.NewReplacer("/", "-", "#", "-")
+
+// MakeSegment returns a copy of string s that is appropriate for a path
+// segment.  MakeSegment is similar to MakePath but disallows the '/' and
+// '#' characters because of their reserved meaning in URIs.
+func (p *PathSpec) MakeSegment(s string) string {
+	s = p.MakePathSanitized(strings.Trim(segmentReplacer.Replace(s), "- "))
+
+	var pos int
+	var last byte
+	b := make([]byte, len(s))
+
+	for i := 0; i < len(s); i++ {
+		// consolidate dashes
+		if s[i] == '-' && last == '-' {
+			continue
+		}
+
+		b[pos], last = s[i], s[i]
+		pos++
+	}
+
+	if p.disablePathToLower {
+		return string(b[:pos])
+	}
+	return strings.ToLower(string(b[:pos]))
+}
+
 // MakePath takes a string with any characters and replace it
 // so the string could be used in a path.
 // It does so by creating a Unicode-sanitized string, with the spaces replaced,
--- a/helpers/path.go
+++ b/helpers/path.go
@@ -97,7 +97,7 @@ func (p *PathSpec) MakeSegment(s string) string {
 		pos++
 	}
 
-	if p.disablePathToLower {
+	if p.DisablePathToLower {
 		return string(b[:pos])
 	}
 	return strings.ToLower(string(b[:pos]))
--- a/helpers/path_test.go
+++ b/helpers/path_test.go
@@ -54,6 +54,8 @@ func TestMakeSegment(t *testing.T) {
 		{"Say What??", "say-what"},
 		{"Your #1 Fan", "your-1-fan"},
 		{"Red & Blue", "red-blue"},
+		{"double//slash", "double-slash"},
+		{"My // Taxonomy", "my-taxonomy"},
 	}
 
 	for _, test := range tests {
--- a/helpers/path_test.go
+++ b/helpers/path_test.go
@@ -37,10 +37,6 @@ import (
 )
 
 func TestMakeSegment(t *testing.T) {
-	v := viper.New()
-	l := NewDefaultLanguage(v)
-	p, _ := NewPathSpec(hugofs.NewMem(v), l)
-
 	tests := []struct {
 		input    string
 		expected string
@@ -59,6 +55,12 @@ func TestMakeSegment(t *testing.T) {
 	}
 
 	for _, test := range tests {
+		v := newTestCfg()
+
+		l := langs.NewDefaultLanguage(v)
+		p, err := NewPathSpec(hugofs.NewMem(v), l)
+		require.NoError(t, err)
+
 		output := p.MakeSegment(test.input)
 		if output != test.expected {
 			t.Errorf("Expected %#v, got %#v\n", test.expected, output)
--- a/helpers/path_test.go
+++ b/helpers/path_test.go
@@ -36,6 +36,34 @@ import (
 	"github.com/spf13/viper"
 )
 
+func TestMakeSegment(t *testing.T) {
+	v := viper.New()
+	l := NewDefaultLanguage(v)
+	p, _ := NewPathSpec(hugofs.NewMem(v), l)
+
+	tests := []struct {
+		input    string
+		expected string
+	}{
+		{"  FOO bar  ", "foo-bar"},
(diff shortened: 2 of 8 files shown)


moorereason:
@bep,
Can you take a look at this PR?  I'm getting random test results when the race detector is enabled (on linux), which matches what Travis is showing (plus tip failed on darwin).  Without the race detector, I can't get it to fail.  For the life of me, I don't see what could be causing the failures.

I've confirmed the failures with Go 1.9.3, 1.9.4, 1.10rc1, and 1.10rc2.

The following will bypass the test cache in Go 1.10:
```
sh$ cd gohugoio/hugo/hugolib
sh$ while :; do echo `date` && go1.10rc2 test -race -run=TestTaxonomiesWithAndWithoutContentFile  ; done
Thu Feb 8 15:52:42 CST 2018
PASS
ok      github.com/gohugoio/hugo/hugolib        1.480s
Thu Feb 8 15:52:49 CST 2018
--- FAIL: TestTaxonomiesWithAndWithoutContentFile (0.00s)
    --- FAIL: TestTaxonomiesWithAndWithoutContentFile/uglyURLs=false,preserveTaxonomyNames=true (0.10s)
        Error Trace:    testhelpers_test.go:41
                        taxonomy_test.go:136
                        taxonomy_test.go:60
        Error:          Should be true
        Test:           TestTaxonomiesWithAndWithoutContentFile/uglyURLs=false,preserveTaxonomyNames=true
        Messages:       File no match for
                        "Cat1" in
                        "public/categories/cat1/index.html":
                        List|CAt1|
FAIL
exit status 1
FAIL    github.com/gohugoio/hugo/hugolib        0.551s
Thu Feb 8 15:52:51 CST 2018
--- FAIL: TestTaxonomiesWithAndWithoutContentFile (0.00s)
    --- FAIL: TestTaxonomiesWithAndWithoutContentFile/uglyURLs=true,preserveTaxonomyNames=true (0.10s)
        Error Trace:    testhelpers_test.go:41
                        taxonomy_test.go:136
                        taxonomy_test.go:60
        Error:          Should be true
        Test:           TestTaxonomiesWithAndWithoutContentFile/uglyURLs=true,preserveTaxonomyNames=true
        Messages:       File no match for
                        "Cat1" in
                        "public/categories/cat1.html":
                        List|CAt1|
    --- FAIL: TestTaxonomiesWithAndWithoutContentFile/uglyURLs=false,preserveTaxonomyNames=true (0.10s)
        Error Trace:    testhelpers_test.go:41
                        taxonomy_test.go:136
                        taxonomy_test.go:60
        Error:          Should be true
        Test:           TestTaxonomiesWithAndWithoutContentFile/uglyURLs=false,preserveTaxonomyNames=true
        Messages:       File no match for
                        "Cat1" in
                        "public/categories/cat1/index.html":
                        List|CAt1|
FAIL
exit status 1
FAIL    github.com/gohugoio/hugo/hugolib        0.468s
```

moorereason:
Correction: I'm able to reproduce the anomaly in Go 1.10rc2 without the race detector.

moorereason:
Fixed the issue.  PTAL

helpers/path.go
```diff
@@ -75,6 +75,35 @@ func (filepathBridge) Separator() string {
 
 var fpb filepathBridge
 
+// segmentReplacer replaces some URI-reserved characters in a path segments.
+var segmentReplacer = strings.NewReplacer("/", "-", "#", "-")
```
bep: Why do you replace "-" with "-"? (the same character).

helpers/path.go
```diff
@@ -75,6 +75,35 @@ func (filepathBridge) Separator() string {
...
+// segment.  MakeSegment is similar to MakePath but disallows the '/' and
+// '#' characters because of their reserved meaning in URIs.
+func (p *PathSpec) MakeSegment(s string) string {
+	s = p.MakePathSanitized(strings.Trim(segmentReplacer.Replace(s), "- "))
+
+	var pos int
+	var last byte
+	b := make([]byte, len(s))
+
+	for i := 0; i < len(s); i++ {
```
bep: If I read this correctly this removes double-slashes. I assume that come from "My // Taxonomy" cases. Which is fine as long as this isn't used in the "general case" (which would break some URLs and make people grumpy). But you should add a test variant for it below.

helpers/path.go
moorereason: NewReplacer uses "old-new" pairs.  This replaces '/' and '#' with '-'.

helpers/path.go
moorereason: Yes, we would replace both slashes with `-` using `segmentReplacer` (other interesting characters like a space and `?` are replaced by `MakePathSanitized` used above), and this loop removes consecutive dashes.

So, `My // Taxonomy` would become `my-taxonomy`.

moorereason: Fixed.

moorereason:
@bep,
It's been a while since I submitted this PR, but it's held up by your requested changes that I believe I've made.  Can you add this to a future milestone for review?  Thanks

helpers/path.go
Zebradil: What if there is a `?` character? Is it possible to use a common url encoding here?

moorereason: See the tests below.  The `?` is removed.  URL encoding is beyond the scope of this PR.

onedrawingperday:
@moorereason 
Does your PR apply also to Taxonomies with custom permalinks configuration?

Let's say that I have a taxonomy called `color` and its permalinks are configured like so: `colour = "/:slug/"` 

Then the taxonomy is assigned to content files with: `colour = "handbags/backpacks/gold"`

With the above Hugo is able to nest a taxonomy under a Section or another Taxonomy. 

Also [here](https://discourse.gohugo.io/t/how-to-generate-chronological-blog-archives-in-hugo/13491/6?u=onedrawingperday) is another use case  that makes use of the above technique to generate Chronological Blog Archives.

As you understand in both use cases the forward slashes are desirable.

moorereason:
@onedrawingperday, I think it will work as you expect, but I'd be more than happy to test a sample site if you have one available.

onedrawingperday:
@moorereason Here is a sample repo with Nested Taxonomies for you test. 
https://github.com/onedrawingperday/hugo-test

I have included both of the Nested Taxonomies use cases that I described above.
If you have a question let me know.

Thanks!

moorereason:
Thanks, @onedrawingperday.  I tested your site with my original PR branch and also rebased with the latest master.  In both cases, your test site comes out exactly the same. 👍

Issue #4393, opened 2018-02-09
Consider YAML vs datadir vs config vs non-string keys


bep:

bep:
@vassudanagunta let us handle this in this issue. I have disabled some YAML tests for now.

There are many arguments for keeping it as it is now (string keys all over).

The question is: 

Do we break something? Or: Maybe we fix more than we break?

vassudanagunta:
Roger. Just please leverage the significant time I already put into #4379. Pretty please 😉 .

bep:
@vassudanagunta #4379 looks like a bigger issue, and I don't have time to consider that (at least now). Remember: I did not implement the front matter handling nor the data handling, and I don't think it's fair to ask me to dig myself into a "rabbit hole" (your words) dug by others.

But I will try to fix **this particular issue**. And my thoughts are this:

* Front matter and /data are meant to be portable across formats (which was the reason why I closed my eyes and merged the related PR)
* We even have commands to do conversion between them that kind of does not work because of this
* People should be able to mix-and-match YAML and TOML etc. content 
* Theme author should be able to write portable templates 
* Having YAML keys as `interface{}` makes sorting random. The lexicographically sort you get with string keys have its own problems, but it is better than random.

So, here is how I **suggest** we do this:

* Keep the implementation as is now (adjust tests)
* Add a check for non-string keys and print a WARNING to the console.

vassudanagunta:
I thought it was fair to ask you given that you are the current steward of Hugo's vision and evolution toward 1.0. 😀  And to be fair(er), I did most of the work of laying things out very clearly. All that was needed was a decision by people who know Hugo's vision and have been around long enough to know what its users expect (not me!). 

And above, you did just that! You definitively answered one of the two main issues I raised #4379 with a clear vision. So thanks! ❤️  

I will update the tests as well as #4379 so that when you and the other Hugo maintainers have time, you can take a look at the remaining issue (overlapping data). All that is needed is a decision, including the decision to punt, in which case just merge #4379 as-is as it codifies in tests Hugo's current behavior, guarding against unintentional changes as well as clarifying it for anyone who decides something should change.

vassudanagunta:
Alright I implemented everything above in #4402 except:

> Add a check for non-string keys and print a WARNING to the console.

@bep you accepted my PR before I got to it (bedtime!). Should I assume that you decided to do it or should I continue?

PR #4402, opened 2018-02-12
JSON, YAML and TOML data equivalency, Better array data support

vassudanagunta:
Proposed resolution of #4393, per @bep's comment under that issue:

> And my thoughts are this:
>
> - Front matter and /data are meant to be portable across formats (which was the reason why I closed my eyes and merged the related PR)
> - We even have commands to do conversion between them that kind of does not work because of this
> - People should be able to mix-and-match YAML and TOML etc. content
> - Theme author should be able to write portable templates
> - Having YAML keys as `interface{}` makes sorting random. The lexicographically sort you get with string keys have its own problems, but it is better than random.
>
> So, here is how I **suggest** we do this:
>
> - Keep the implementation as is now (adjust tests)
> - Add a check for non-string keys and print a WARNING to the console.

Four commits with isolated pushes to isolate changes for easier diff review and isolated CI tests results:  

1. The Re-enable YAML data tests disabled in f554503f in isolated committed to prove production code passes tests as-is. Resolves test breakage by #4138.
2. Refactoring of test code also made in isolated commit and push for same reason.
3. #4366 fixed with test coverage. 
4. #3890 fixed with test coverage. 

See commit comments for details.

If this PR is accepted, I can reduce the scope of #4379.

### DIFF
--- a/hugolib/datafiles_test.go
+++ b/hugolib/datafiles_test.go
@@ -95,25 +95,23 @@ func TestDataDirBoolean(t *testing.T) {
 
 func TestDataDirTwoFiles(t *testing.T) {
 	t.Parallel()
-	equivDataDirs := make([]dataDir, 2)
+	equivDataDirs := make([]dataDir, 3)
 
 	equivDataDirs[0].addSource("data/test/foo.json", `{ "bar": "foofoo"  }`)
-	equivDataDirs[0].addSource("data/test.json", `{ "hello": [ { "world": "foo" } ] }`)
+	equivDataDirs[0].addSource("data/test.json", `{ "hello": [ "world", "foo" ] }`)
 
 	equivDataDirs[1].addSource("data/test/foo.yaml", "bar: foofoo")
-	equivDataDirs[1].addSource("data/test.yaml", "hello:\n- world: foo")
+	equivDataDirs[1].addSource("data/test.yaml", "hello:\n- world\n- foo")
 
-	// TODO Unresolved Issue #3890
-	/*
-		equivDataDirs[2].addSource("data/test/foo.toml", "bar = \"foofoo\"")
-		equivDataDirs[2].addSource("data/test.toml", "[[hello]]\nworld = \"foo\"")
-	*/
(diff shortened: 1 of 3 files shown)


bep:
>Add a check for non-string keys and print a WARNING to the console.

Could you add a warning for the above case. Use the DistinctWarningLogger in helper so the user only gets 1 warning.

Also, what I see this commit:

https://github.com/gohugoio/hugo/pull/4402/commits/1e74e70c0cd7c87be122af82f70bfec479a2a2e1

About YAML array support. I don't even know how to express that in YAML so it would be cool to see a test.

vassudanagunta:
@bep 

>> Add a check for non-string keys and print a WARNING to the console.
>
> Could you add a warning for the above case. Use the DistinctWarningLogger in helper so the user only gets 1 warning.

I will have to introduce the check in the `stringifyYAMLMapKeys` function introduced in 16a5c74. So I would use a call to `helpers.DistinctWarnLog.Println(err)`, right?

bep:
>So I would use a call to helpers.DistinctWarnLog.Println(err), right?

Yes, but don't bee to specific in the error message (to avoid spamming).

vassudanagunta:
https://github.com/gohugoio/hugo/blob/b5cb8866e9797efa28072e4eac68abd8a1b9c5d3/parser/frontmatter.go#L225-L242

@bep About to add the check and warning... I'm truly a Go newb (all the Go code I've ever written has been my small contributions to Hugo to-date), but this method copies the entire data structure, however broad and however deep, rather than making in situ modifications, even if only one or even no element out of thousands needs to be stringified. This will be both memory and time expensive for large inputs. See #1065. Am I correct?

bep:
>res[fmt.Sprintf("%v", k)]

@vassudanagunta the above should use `cast.ToString` to make it consistent with the rest of the code (and faster/cheaper).

As to the effectiveness of the code. There are some optimisations 

1. no need to re-create the slices
2. no need to update map if not changed

But is that the scope of this particular issue?

PR #4405, opened 2018-02-13
Render Markdown in figure sc generated figcaption caption and attr

kaushalmodi:
- The Markdown emphasis characters if present in the "caption" and "attr"
  parameters for the figure shortcode will be rendered into HTML, and used in
  the shortcode-generated figcaption tag.
- The "caption" parameter is reused in the "alt" attribute in the 
  tag. Though, the alt text should contain only plain text. So emphasis tags are
  removed when "caption" is reused in the "alt" attribute.
- Add test for this.

### DIFF
--- a/hugolib/shortcode_test.go
+++ b/hugolib/shortcode_test.go
@@ -298,6 +298,11 @@ func TestFigureOnlySrc(t *testing.T) {
 	CheckShortCodeMatch(t, `{{< figure src="/found/here" >}}`, "\n<figure>\n    \n        <img src=\"/found/here\" />\n    \n    \n</figure>\n", nil)
 }
 
+func TestFigureCaptionWithMarkdown(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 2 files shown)


bep:
For changes like this, we need an issue. I'm not sure about this. If we want to add this, we want to be sure that it does not break sites. And for that, we need an issue.

bep:
Reissue this when/if we agree that we want to do this.

DirtyF:
Thanks @kaushalmodi for the custom `figure` shortcode to support Markdown. 🙇

kaushalmodi:
@DirtyF Glad you found it useful. The proposal to bake this into hugo is made at https://github.com/gohugoio/hugo/issues/4406. May be if you comment/vote for addition of this feature.. So far, looks like I am the only one pushing for this :(

DirtyF:
@kaushalmodi AFAIC I'm fine with a custom shortcode - though I wouldn't mind if this was supported out-of-the-box. I have no strong opinion on this and FYI this isn't supported by default in Jekyll, either. I guess you have to love to hack your way to get into SSGs. 

I just had to pay attention and escape double quotes in my figure caption. 
 The error message `got positional parameter 'hidden'. Cannot mix named and positional parameters` confused me a bit.

lorio:
Also, the generated `<figure>` element does not contain the `<figcaption>` but it is rendered as a `p` and inserted after the `<figure>` in 0.40.   `<figcaption>` should be inside `<figure>`

PR #4509, opened 2018-03-16
Add Reset method to delete key from Scratch

paulcmal:
Following a [discussion](https://discourse.gohugo.io/t/delimit-and-create-array/11065/11?u=cmal) in the forums, I'm proposing a `Reset` method for Scratch, to remove a Scratch entry. It could be renamed Delete if you prefer.

This is my first time writing some golang so sorry if I'm doing some things wrong. Unfortunately, I'm running golang 1.8.1 from Debian stable so I cannot run the tests. I tried it with the following snippet and it appears to work like a charm:

```
{{ $.Scratch.Set "test" "lol" }}
{{ $.Scratch.Reset "test" }}
{{ $.Scratch.Reset "anothertest" }}
{{ $.Scratch.Add "test" "Bonjour" }}
{{ $.Scratch.Get "test" }}
```

It printed "Bonjour" like expected and did not raise an error when attempting to delete a key that did not exist (I believe this is the desired behaviour).

I read the contribution guidelines and tried to give meaning to my commit message. Don't hesitate to amend my commit :)

### DIFF
--- a/docs/content/functions/scratch.md
+++ b/docs/content/functions/scratch.md
@@ -32,6 +32,7 @@ See [this Go issue](https://github.com/golang/go/issues/10608) for the main moti
 * `Get` returns the `value` for the `key` given.
 * `SetInMap` takes a `key`, `mapKey` and `value`
 * `GetSortedMapValues` returns array of values from `key` sorted by `mapKey`
+* `Delete` takes a `key` to remove
 
 `Set` and `SetInMap` can store values of any type.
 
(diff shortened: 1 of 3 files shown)


bep:
Great, we can add this if you:

1. Rename it to "Delete". This is more in line with what it does.
2. Sign the CLA above (if it does not work, check your commit email address vs the addresses you have registered in GitHub)
3. Update the Scratch docs. Do it in this repository.

paulcmal:
Thank you for the quick answer @bep!

I added an example in the docs, renamed `Reset` to `Delete`. Unfortunately, I can't figure out why the tests are failing, since I can't run them locally.

I tried to upgrade to golang 1.10 manually. That worked (`go version go1.10 linux/amd64`) but even after cleaning/redownloading/recompiling everything, I still get `Skip Check on go1.8.1`. When I compile my own small hello world, I get runtime version 1.10 as expected. Do you have any idea what could be causing this?

bep:
You need to run `gofmt` on your code.

bep:
@paulcmal give LiteIDE a try. It does gofmt on save -- it is really fast and easy to use and my fav editor for Go code. And it is free ...

PR #4155, opened 2017-12-10
feat(docker): hugo from scratch + dockhub & quay account

ellerbrock:
Hello Hugo's,

i optimized our Hugo Docker Image to run [FROM scratch](https://docs.docker.com/engine/userguide/eng-image/baseimages/#create-a-simple-parent-image-using-scratch) instead of [Alpine Linux](https://hub.docker.com/r/_/alpine/) and with that another optimization from 27MB to [6MB](https://hub.docker.com/r/gohugoio/hugo/tags/). Pretty sweet optimization when thinking that we started with [428MB](https://github.com/gohugoio/hugo/pull/3674)!

The good thing with that is we don't need to worry about security patches and image updates in the future anymore. Looking at the latest Version 3.7 of [Alpine](https://hub.docker.com/r/library/alpine/tags/3.7/) there are 3  unpatched [vulnerabilities](https://hub.docker.com/r/library/alpine/tags/) mentioned with the root cause in the underlaying Busybox:

- [CVE-2017-16544](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-16544)
- [CVE-2017-15873](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-15873)
- [CVE-2017-15874](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-15874)

**Hugo DockerHub**

I ask on [26 Jul](https://github.com/gohugoio/hugo/pull/3738#issuecomment-318174847) about the topic to configure DockerHub with [Automated Builds](https://docs.docker.com/docker-cloud/builds/automated-build/) for the Hugo Project. Since there is still no official Hugo Repo i was so free and created one myself.

DockerHub Account: [gohugoio](https://hub.docker.com/u/gohugoio/)
Repository: [hugo](https://hub.docker.com/r/gohugoio/hugo)
Pull Command: `docker pull gohugoio/hugo`

**Quay Account**

Quay Account: [gohugoio](https://quay.io/user/gohugoio)
Repository: [hugo](https://quay.io/repository/gohugoio/hugo)
Pull Command: `docker pull quay.io/gohugoio/hugo`

The nice thing on Quay is that they run under the hood [Clair](https://github.com/coreos/clair) (Vulnerability Static Analysis for Containers) and we can define to trigger notifications when problems are found.

**Current Status**

[...]

PR #4518, opened 2018-03-18
Update Dockerfile to a multi-stage build

skoblenick:
These changes are the sum of the work done by @ellerbrock in 
previous pull requests. As those branches are stale and have 
conflicts this should bring the work up-to-date.

- Hugo container is based on SCRATCH to further reduce the 
  footprint and the vulnerability surface
- Update Alpine image to 3.7 in the build container
- Update Go Lang to 1.10 in the build container
- Add .dockerignore file per the Docker best practices

Closes #4154, #4155, #4157

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,27 @@
-FROM golang:1.9.0-alpine3.6 AS build
+# GitHub:       https://github.com/gohugoio
+# Twitter:      https://twitter.com/gohugoio
+# Website:      https://gohugo.io/
 
-RUN apk add --no-cache --virtual git musl-dev
-RUN go get github.com/golang/dep/cmd/dep
+FROM golang:1.10.0-alpine3.7 AS build
+
+ENV CGO_ENABLED=0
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,23 +1,29 @@
-FROM golang:1.9.0-alpine3.6 AS build
+# GitHub:       https://github.com/gohugoio/
+# DockerHub:    https://hub.docker.com/r/gohugoio/
+# Quay:         https://quay.io/user/gohugoio
```
bep: I see some accounts above that I don't recognize. I think it is better to just leave the official 2.

ellerbrock: hi @bep,

like mentioned a while ago [here](https://github.com/gohugoio/hugo/pull/4155#issuecomment-350538197) i created those accounts to be consistent with the naming. just let me know when you consider to use them and i will hand them over to you ...

cheers maik

skoblenick: @bep Updated to only official

PR #4527, opened 2018-03-23
Update "Advanced Install" in README.md

thedodobird2:
I discovered an inconsistency between the website documentation (https://gohugo.io/getting-started/installing/#source) and this README, for installing from source. I checked them both and the website seems to be newer so I would like to propose updating this README to install Hugo using mage (unless it doesn't really matter).


bep:
>Okay, but at least both documentations should be consistent, right? 

Correct, and thanks.

PR #4541, opened 2018-03-27
Add .Site.Info.IsServer flag when using built-in Hugo server.

### LINKED ISSUE #4478: Add .Site.IsServer to determine if the site was built in server mode
As per the following Discourse discussion:

https://discourse.gohugo.io/t/disable-portions-of-content-while-developing/10860

It would be helpful to put a condition into a template that can determine whether the page is being built/server by `hugo server` (and not necessarily with the `-D` flag, for which there is already a variable).

This would be helpful for the case where you want to test the site in its final form (no drafts) but want to switch off, e.g. analytics or ads.

felicianotech:
Fixes #4478.

Docs PR coming shortly.

Update:

Docs PR: gohugoio/hugoDocs#430

### DIFF
--- a/docs/content/variables/site.md
+++ b/docs/content/variables/site.md
@@ -52,6 +52,9 @@ The following is a list of site-level (aka "global") variables. Many of these va
 .Site.IsMultiLingual
 : whether there are more than one language in this site. See [Multilingual](/content-management/multilingual/) for more information.
 
+.Site.IsServer
+: a boolean to indicate if the site is being served with Hugo's built-in server. See [`hugo server`](/commands/hugo_server/) for more information.
+
 .Site.Language.Lang
 : the language code of the current locale (e.g., `en`).
 
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -380,7 +380,6 @@ type SiteInfo struct {
 	BuildDrafts           bool
 	canonifyURLs          bool
 	relativeURLs          bool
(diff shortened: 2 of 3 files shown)


hugolib/site.go
```diff
@@ -380,6 +380,7 @@ type SiteInfo struct {
 	BuildDrafts           bool
 	canonifyURLs          bool
 	relativeURLs          bool
+	IsServer              bool
```
bep: I think what we want to do is just add a method that returns the value of this{

https://github.com/gohugoio/hugo/blob/3cdf19e9b7e46c57a9bb43ff02199177feb55768/deps/deps.go#L206

bep: Or more specifc this value:

felicianotech: Ah gotcha. Will rework.

docs/content/variables/site.md
```diff
@@ -122,4 +125,4 @@ You can use `.Site.Params` in a [partial template](/templates/partials/) to call
 <meta name="description" content="{{if .IsHome}}{{ $.Site.Params.description }}{{else}}{{.Description}}{{end}}" />
 {{< /code >}}
 
-[config]: /getting-started/configuration/
\ No newline at end of file
+[config]: /getting-started/configuration/
```
felicianotech: I think this is just that Vim added a newline at the end.

PR #4564, opened 2018-04-04
Add support for mermaid diagram rendering in Markdown codeblock

Honghe:
So we can add [mermaid](https://mermaidjs.github.io/) diagram simply by 2 step:
* indicate the code language to `mermaid` in Markdown
* add mermaid.js to layout/theme.

    ```mermaid
    sequenceDiagram
        Alice->>John: Hello John, how are you?
        John-->>Alice: Great!
    ```

image

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -1594,7 +1594,11 @@ func (p *Page) parse(reader io.Reader) error {
 
 	p.renderable = psr.IsRenderable()
 	p.frontmatter = psr.FrontMatter()
-	p.rawContent = psr.Content()
+	// replace for mermaid
+	p.rawContent, err = replaceMermaidTokens(psr.Content())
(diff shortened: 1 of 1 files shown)


bep:
Thank you for your contribution, but this seem to be a little too narrow a use case for us to include into core like this. We may revisit this in the future once we get plugins into place.

bep:
Also note that you can probably do this with shortcodes.

Honghe:
@bep Thx. Yes, `Shortcodes` can do this, but the grammer `{{%%}}` or `{{<>}}` is a little inconvennient with most Markdown eco Tools, such as vscode or Typora，cause they use `\```mermaid` at codeblock.  
Agree with you, Hugo opens plugin API maybe is a better solution :)

PR #4568, opened 2018-04-04
Adds .gitattributes to force Go files to LF

neurocline:
Even on Windows machines, we want *.go files to have LF line
endings (go fmt insists). This leaves other files as auto, but
makes *.go files always be LF on Windows machines. With this
change, running go fmt will be a pleasant operation, not a
painful one.

Note that changing .gitattributes only affects files at initial
checkout. To update an entire working directory after making a
change like this, do this (but only after committing all work in
progress, this will destroy uncommitted changes):

$ git rm --cached -r .
$ git reset --hard

Again - do not do this to a working directory with uncommitted
work in it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Text files have auto line endings
+* text=auto
+
+# Go source files always have LF line endings
+*.go text eol=lf


neurocline:
My checking comment has a bug. It should have been

$ git rm --cached -r .

bep:
This is probably fine -- bit someone with a little more Git knowledge in this department should OK it... I normally do not work on Windows (I have a VM that I spin up from time to time) and have not felt this pain.

moorereason:
@neurocline,
Can you explain why this is necessary?  How has this issue manifested itself in your environment?  What editor are you using?

I do most of my Go work in Linux, but I've done some in Windows.  I've never had issues with line endings or gofmt.  I've never seen a Go project need this.  I use vim or Visual Studio Code when in Windows.

anthonyfok:
I don't know the answer either.  I _rarely_ use Windows for Hugo maintenance except for the occasional debug and testing.

Based on my limited experience, I just know that, when installing Git for Windows, it will ask the end user whether to keep the LF as is, or to convert them to CRLF when pulling, and to convert them back to LF when pushing.  I guess I haven't tried running `go fmt` with those CRLF line endings yet.

Since this problem would affect other Windows users too, I decided to look at Go itself, and indeed, found a `.gitattributes` file there for the exact same purpose, but slightly different implementation.  From https://github.com/golang/go/blob/master/.gitattributes:

```
# Treat all files in the Go repo as binary, with no git magic updating
# line endings. Windows users contributing to Go will need to use a
# modern version of git and editors capable of LF line endings.
#
# We'll prevent accidental CRLF line endings from entering the repo
# via the git-review gofmt checks.
#
# See golang.org/issue/9281

* -text
```

(whereas @neurocline has `* text=auto` and `*.go text eol=lf`.)

I don't know, which one is better?  :-D

neurocline:
The proximate reason for this pull request is that most Windows Git users have "core.autocrlf=true" in their global Git settings (you can see where a Git config value is stored with "git config --show-origin --list"), because the Git for Windows installer suggests you do it this way. This means that all text files checked out to a working directory are in CRLF form, and gofmt only writes files in LF line endings. I felt that we shouldn't rely on how users set up their Git systems, and you want to encourage people to run gofmt on their Hugo changes all the time. Now, this may change - over the past 5 years, most of the major Windows text editing tools have become friendly to LF line endings, and maybe we can all ditch the idea of CRLF files. But until all Windows Git users stop having core.autocrlf=true, I think we need a .gitattributes in repos with Go source files.

@morereason, if you have "core.autocrlf=true" on your Windows system, you will get Go files with CRLF line endings and gofmt will rewrite all of them to LF line endings (and, more specific to Hugo, "mage check" will fail). You may be one of the people who installed Git on Windows without that setting. But most Windows Git developers have core.autocrlf=true.

I had an extraneous setting "text=auto" in my proposed .gitattributes, and I will remove it. That was too pedagogical, and could conflict with someone else's setting. It's also not necessary to the issue at hand.

The Go team's .gitattributes is interesting. On the one hand, ".go text eol=lf" and "* -text" would do exactly the same thing; not alter line endings. On the other hand, their change is a little over the top, since it declares all Go files as binary files, and there are parts of Git that treat binary files differently than text files. Maybe all those have changed over the years, and it's now synonymous. It still makes me feel a little dirty doing so. I'd prefer the explicit "Go files are text files with LF line endings".

I'm going to change my repo to the "* -text" setting and do a little work and see what happens.

neurocline:
Now I remember why I had "text=auto". That's a call to make sure text file line endings are normalized to LF in the repo. So I think there are only two sane choices here.

1. The Go team approach

* -text

2. The targeted approach

* text=auto
*.go text eol=lf

In the Go team approach, no files have line endings normalized. Git still decides what is text vs binary by looking at file contents, so "git diff" still displays individual line changes. This relies on a code review tool preventing CRLF files from being committed; as long as you run gofmt religiously, at least the .go files will be correct.

In the targeted approach, text files have line endings normalized so only LF line endings get into the repo, and .Go files are checked out with LF line endings. This way, no CRLF files ever get into the repo, and .Go files are LF in the working directory, making gofmt happy.

Sorry, this is a lot of commentary for such a small change, but it is a change to workflow, so I guess worth taking care with.

PR #4570, opened 2018-04-05
Adds abort error handling

neurocline:
Removes internal os.Exit calls in favor of an abort channel;
sending an error on the abort channel tells top-level code that
it should exit. At the moment, this does not unwind all that
quickly, it just means that a failed build operation will result
in the outer "hugo server" or "hugo --watch" exiting.

This also removes utils.StopOnErr - it was only used in one
place and it had an os.Exit call in it.

There are remaining os.Exit calls but they are being used to
return non-zero status from main. Ideal future code would return
exit status to main and it would use os.Exit if the exit status
is non-zero.

### DIFF
--- a/commands/commandeer.go
+++ b/commands/commandeer.go
@@ -51,6 +51,12 @@ type commandeer struct {
 	// We can do this only once.
 	fsCreate sync.Once
 
+	// Routines send to abort to say "stop operation"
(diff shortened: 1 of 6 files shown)


neurocline:
Pull request race condition :) I'll rebase and resubmit.

PR #4572, opened 2018-04-05
Adds --source, --theme, --themesDir to "hugo config"

neurocline:
Add extra command-line flags so "hugo config" can work on site
dirs that have themes located elsewhere, or so that we can do
"hugo config" on a site that's not the current working directory.

### DIFF
--- a/commands/list_config.go
+++ b/commands/list_config.go
@@ -29,6 +29,10 @@ var configCmd = &cobra.Command{
 }
 
 func init() {
+	configCmd.Flags().StringVarP(&source, "source", "s", "", "filesystem path to read files relative from")
+	configCmd.Flags().StringVarP(&theme, "theme", "t", "", "theme to use (located in /themes/THEMENAME/)")
+	configCmd.Flags().StringVar(&themesDir, "themesDir", "", "filesystem path to themes directory")
+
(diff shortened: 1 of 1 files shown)


commands/list_config.go
```diff
@@ -29,6 +29,10 @@ var configCmd = &cobra.Command{
 }
 
 func init() {
+	configCmd.Flags().StringVarP(&source, "source", "s", "", "filesystem path to read files relative from")
+	configCmd.Flags().StringVarP(&theme, "theme", "t", "", "theme to use (located in /themes/THEMENAME/)")
+	configCmd.Flags().StringVar(&themesDir, "themesDir", "", "filesystem path to themes directory")
+
```
bep: I agree about the change itself, but it would be even better if you could somehow reuse the same lines in the other variant. I think we have a hugoBuildFlags method or something.

neurocline:
The reason I didn't use initHugoBuildCommonFlags is because that function adds 25 other build flags besides the --source, --theme, and --themesDir flags (most of which make no sense for "hugo config"), and I didn't feel that it was worth adding another helper function to split those out.

However, if you prefer that approach, I can make a new helper function to split those flags out into.

PR #4640, opened 2018-04-17
Peertube shortcode

paulcmal:
Inspired by youtube and vimeo shortcode, here's a PR for a peertube shortcode. Peertube is a very interesting federated video streaming platform based on Webtorrent.

I would love to see Hugo support Peertube embeds out-of-the-box :)

Don't hesitate to let me know if there are ways in which I can improve my proposal!


onedrawingperday:
I am all for federated alternatives to Social Media Services. But Peertube at its [about](https://peertube.cpy.re/about) page states the following:

> your public IP address is stored in the public BitTorrent tracker of the video PeerTube instance as long as you're watching the video. If you want to keep your public IP address private, please use a VPN or Tor

The above just doesn't seem good enough with the new privacy regulation in Europe.

I'm sorry but I don't think that this shortcode should be added to Hugo core.

paulcmal:
> The above just doesn't seem good enough with the new privacy regulation in Europe.

Are you saying P2P services like Bittorrent, Bitcoin & others are not in compliance with [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) ? I think that's a fair misunderstanding of these regulations which concern organizations storing personal information, not technical systems relying on personal information (such as the IP address) to provide some functions.

GDPR does affect the Peertube federated servers, though. There's [a ticket](https://github.com/Chocobozzz/PeerTube/issues/226) on this. Would you reconsider merging this once these issues are addressed on their side?

paulcmal:
> I still don't think that this should be added to Hugo

I understand. However i think it has its place for the single reason that's there's a default Youtube shortcode provided out-of-the-box. I don't see a reason to enforce Google's monopoly without providing alternatives as well (although i personally don't see a reason for such shortcodes to be provided out-of-the-box anyway).

The peertube ecosystem has matured a lot in the past year and gained a lot of traction. If some people think this is worth reconsidering i'm willing to rework my PR with the latest Peertube shenanigans :)

PR #4673, opened 2018-04-26
hugolib: Replace deprecated {Get,}ByPrefix with {Get,}Match

anthonyfok:
Oops, sorry, I should have tested more thoroughly before I pushed.  I could reproduce the CI build failure locally.


anthonyfok:
Found the problem in my previous push: I should have used `**` instead of just `*` in order to match the deep directories.  Re-pushed.

Issue #4672, opened 2018-04-26
[armel port] TestPageBundlerSiteRegular fails with v0.40 and v0.40.1


anthonyfok:
Hugo v0.40 and v0.40.1 fails to build on armel (ARM EABI with soft float) and only on armel on Debian build daemon:

* https://buildd.debian.org/status/package.php?p=hugo
* https://buildd.debian.org/status/fetch.php?pkg=hugo&arch=armel&ver=0.40.1-1&stamp=1524683104&raw=0

> Note: The same test for `mips` build took over 10 minutes and go test timed out, but increasing the go test timeout value should fix it.

With the help of `git bisect run` on a Debian armel test machine, it was found that the first commit that causes TestPageBundlerSiteRegular to fail on armel is 4d26ab33dcef704086f43828d1dfb4b8beae2593:

```
(sid_armel-dchroot)foka@abel:~/go/src/github.com/gohugoio/hugo$ time go test -v -run TestPageBundlerSiteRegular ./hugolib/ 
=== RUN   TestPageBundlerSiteRegular
=== PAUSE TestPageBundlerSiteRegular
=== CONT  TestPageBundlerSiteRegular
=== RUN   TestPageBundlerSiteRegular/ugly=false
ERROR 2018/04/26 14:41:26 Resources's GetByPrefix is deprecated and will be removed in Hugo 0.42. We have added the more flexible Resources.GetMatch (find one) and Resources.Match (many) to replace the "prefix" methods. 

These matches by a given globbing pattern, e.g. "*.jpg".

Some examples:

* To find all resources by its prefix in the root dir of the bundle: .Match image*
* To find one resource by its prefix in the root dir of the bundle: .GetMatch image*
* To find all JPEG images anywhere in the bundle: .Match **.jpg
=== RUN   TestPageBundlerSiteRegular/ugly=true
--- FAIL: TestPageBundlerSiteRegular (356.79s)
    --- FAIL: TestPageBundlerSiteRegular/ugly=false (179.46s)
        Error Trace:    testhelpers_test.go:440
    	            	page_bundler_test.go:194
    	Error:      	Should be true
    	Test:       	TestPageBundlerSiteRegular/ugly=false
    	Messages:   	File no match for
    	            	"TheContent" in
    	            	"/work/public/2017/pageslug/index.html":
    	            	
    	            	Single Title: Bundle Galore
    	            	Content: 
    	            	
    	            	
[...]

PR #4674, opened 2018-04-26
hugolib: Use double quotes instead of back quotes in timeout warning

anthonyfok:
A very trivial pull request: I was seeing an extra `\n` at the end of the warning message after adding `Logger: newWarningLogger()` (an excellent suggestion, by the way, thanks @ bep!   See #4672)

```
=== RUN   TestPageBundlerSiteRegular/ugly=false
...
WARN 2018/04/26 20:10:29 WARNING: Timed out creating content for page
"b/my-bundle/index.md" (.Content will be empty). This is most likely a circular
shortcode content loop that should be fixed. If this is just a shortcode calling
a slow remote service, try to set "timeout=20000" (or higher, value is in milliseconds)
in config.toml.\n
...
```

Thanks!

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -305,7 +305,7 @@ func (p *Page) initContent() {
 
 		select {
 		case <-ctx.Done():
-			p.s.Log.WARN.Printf(`WARNING: Timed out creating content for page %q (.Content will be empty). This is most likely a circular shortcode content loop that should be fixed. If this is just a shortcode calling a slow remote service, try to set "timeout=20000" (or higher, value is in milliseconds) in config.toml.\n`, p.pathOrTitle())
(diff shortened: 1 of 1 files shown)


moorereason:
I'm assuming we need the newline with Printf.  I'd suggest using a regular double-quoted string and escape the internal double-quotes around `timeout=20000`.

anthonyfok:
@moorereason Thanks for the review!

Yes, I was wondering about that too, but in the same hugolib/page.go file, there were many cases of other p.s.Log.*.Printf() calls where no trailing `\n` was added.  I suppose the p.s.Log, which uses spf13's jWalterWeatherman.Notepad, uses the standard Go log.Logger, of which its documentation https://golang.org/pkg/log/ says:

> Every log message is output on a separate line: if the message being printed does not end in a newline, the logger will add one. 

Anyhow, I do like your suggestion because @ bep did add the `\n`, so let's keep it.  I'll make a new push soon.

PR #4697, opened 2018-05-04
Update README.md

EOussama:
Centered the logo because it looks visually better?


RealOrangeOne:
Looks like some of your markup is invalid. You can't mix `p` tags and markdown quite like that.

PR #4705, opened 2018-05-07
Render *.xml files as XML (with <?xml … ?> header)

stapelberg:
This is required to generate XML files with proper UTF-8 encoding.


stapelberg:
Still relevant AFAICT

PR #4720, opened 2018-05-14
Add configFile(s) back to the watch list after REMOVE event

### LINKED ISSUE #4701: "hugo server -w" ignores config.toml changes after REMOVE event since 0.38
... and after 1 save in gedit (GNOME Text Editor), as tested with Hugo 0.40.2 on Debian sid (buster, testing) on amd64.

When I try to save config.toml repeatedly every few seconds:

* For Vim, Hugo receives a RENAME event the first save (and rebuilds), and CHMOD+REMOVE or just REMOVE event the second save (and rebuilds), and after that, Vim seems to forget about config.toml altogether, and does not receive further events from changes to config.toml, unless I press Ctrl+C and restart `hugo server`.

* For gedit, Hugo receives CHMOD+REMOVE for the first save (and rebuilds), and ignores config.toml changes from then on unless `hugo server` is restarted.

Maybe it has to do with how currently Hugo handles the REMOVE event?  Or something else altogether?  (I haven't checked the code yet at all.)

Thanks!

anthonyfok:
Fixes #4701

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -705,6 +705,18 @@ func (c *commandeer) newWatcher(dirList ...string) (*watcher.Batcher, error) {
 						if ev.Op&fsnotify.Chmod == fsnotify.Chmod {
 							continue
 						}
+						if ev.Op&fsnotify.Remove == fsnotify.Remove {
+							for _, configFile := range c.configFiles {
+								counter := 0
+								for watcher.Add(configFile) != nil {
+									counter++
+									if counter >= 100 {
+										break
+									}
+									time.Sleep(100 * time.Millisecond)
+								}
+							}
+						}
 						// Config file changed. Need full rebuild.
 						c.fullRebuild()
 						break


commands/hugo.go
```diff
@@ -781,6 +781,15 @@ func (c *commandeer) newWatcher(dirList ...string) (*watcher.Batcher, error) {
 						if ev.Op&fsnotify.Chmod == fsnotify.Chmod {
 							continue
 						}
+						if ev.Op&fsnotify.Remove == fsnotify.Remove {
+							for _, configFile := range c.configFiles {
+								c.Logger.FEEDBACK.Println(configFile, "disappeared, waiting for it to come back...")
+								for watcher.Add(configFile) != nil {
+									time.Sleep(100 * time.Millisecond)
+								}
+								c.Logger.FEEDBACK.Println(configFile, "is back!")
```
bep: I'm not a fan of these funny log messages, esp since you're not really "waiting for it to come back" (which is fine)

I dont' see the full code in this diff, but this event is the config file only, right? We're not adding a sleep fore every remove event?

anthonyfok: @bep Thanks for the review!

> I'm not a fan of these funny log messages, esp since you're not really "waiting for it to come back" (which is fine)

Haha, true!  The log messages are not really needed, but just for my debugging.
I'll push this commit again with the funny log messages removed.

> I dont' see the full code in this diff, but this event is the config file only, right? We're not adding a sleep fore every remove event?

Yes, this event is for the config file only, being within the existing if-block for config files:

```go
    if configSet[ev.Name] {
        // if Chmod event, skip (continue)
        // if Remove event, sleep and add it back
        // Config file changed. Need full rebuild.
    }
```

anthonyfok:
Funny log messages removed, and a commit force-pushed.  ;-)
Thanks!

>> I'm not a fan of these funny log messages, esp since you're not really "waiting for it to come back" (which is fine)
>
>Haha, true!  The log messages are not really needed, but just for my debugging.
>I'll push this commit again with the funny log messages removed.
>
>> I dont' see the full code in this diff, but this event is the config file only, right? We're not adding a sleep fore every remove event?
>
>Yes, this event is for the config file only, being within the existing if-block for config files:
>
>```go
>    if configSet[ev.Name] {
>        // if Chmod event, skip (continue)
>        // if Remove event, sleep and add it back
>        // Finally, full rebuild because of config change
>    }
>```

commands/hugo.go
```diff
@@ -781,6 +781,13 @@ func (c *commandeer) newWatcher(dirList ...string) (*watcher.Batcher, error) {
 						if ev.Op&fsnotify.Chmod == fsnotify.Chmod {
 							continue
 						}
+						if ev.Op&fsnotify.Remove == fsnotify.Remove {
+							for _, configFile := range c.configFiles {
+								for watcher.Add(configFile) != nil {
```
bep: I don't remember this for loop in the original PR. What is the motivation behind this? If it errors once if will continue to error forever, I assume.

anthonyfok: Thank you again for the code review.

> I don't remember this for loop in the original PR.

I have kept the original commit for this PR is here: https://github.com/anthonyfok/hugo/commit/b73d8107fe8f654f3915f30a14185bdd9417b968

> What is the motivation behind this?

watcher.Add(configFile)  returns non-nil when the configFile does not exist.

This for loop is to ensure the configFile is indeed added to the fsnotify watcher.
On a slow/busy system, it may take longer than 100ms for the removed file to reappear, sometimes over 1 second, so it went through the loop like 10 times until config.toml re-appear.

> If it errors once if will continue to error forever, I assume.

Not in my local testing, when I kept saving config.toml from within Vim or GNOME gedit.

However, I think you correctly point out a potential danger of getting stuck in this loop, as in: what if watcher.Add() fails not because config.toml does not exist, but for some other reasons, e.g. hitting the descriptor limit somehow?

So, perhaps there should be a way out of the loop, e.g. a time out, and/or running this loop in a separate thread?

What would your recommend?  :-)

bep: I would recommend a "if counter > 100" then break. The value does not matter. I assume this will work fine as is in 99.9 percent of the cases, but other people will look at the code and wonder.

bep: @anthonyfok bump; I will merge this but you need to put in a break in the loop so it can exit in those special cases.

anthonyfok: Hi!  Sorry, I had a very busy week (preparing and working at a conference), and then I got lazy.  :-D 
 Thank you for the reminder!  I'd better get on it and add the counter.  Will update this pull request within an hour.

anthonyfok:
Hi!  I am sorry, I failed to keep my promise and finished it up last week.

I have added a counter to my local test code, but then realized my code did not work with neovim (unrelated to the counter), so I would like to refine this pull request some more when I have time to test more thoroughly.

bep:
@anthonyfok what is the status of this? I think this issue is annoying enough for many so we could probably merge this and fix neovim in another.

anthonyfok:
@bep Thanks for the reminder!  I have finally added a counter as you suggested just there should be no chance of it getting stuck forever.  Please check and merge it if it looks OK to you.

I just noticed something that perplexes me a bit.  I just noticed that Hugo 0.46 no longer prints any message when config.toml is changed, but it does silently rebuild the site and live-reloads it.  But yeah, that's a cosmetic and unrelated issue for another PR.

bep:
OK, so my recent fixes made this one conflict ... Could you have a look at it? I'm a bit "done with Hugo" for today ...

PR #4730, opened 2018-05-17
[Docs] Fix misspelled setting for using classic Pygments

mbaeuerle:
I think this is the right setting for using Pygments.


mbaeuerle:
@eyoder the `pygmentsUseClasses=true` option does exist and in your cited case I think it is correctly used.

kaushalmodi:
@mbaeuerle Thanks for catching this. PRs for doc site need to go to https://github.com/gohugoio/hugoDocs. As this was a quick minor fix, I did it myself in https://github.com/gohugoio/hugoDocs/commit/0fd7b416997bf76cd99e3ae2dfae406cb8b6724a. 

So this PR can be closed.

PR #4731, opened 2018-05-17
Use HTML5 <meta charset="utf-8">

coliff:
Use: 
`<meta charset="utf-8">`
instead of the XHTML version:
`<meta http-equiv="content-type" content="text/html; charset=utf-8" />`
less markup :-)

### DIFF
--- a/hugolib/alias.go
+++ b/hugolib/alias.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


coliff:
thanks for the feedback and suggestions @XhmikosR !  

@bep  could we get this one merged please? or should I add the default lang and lowercase doctype as suggested by @XhmikosR  in this same PR?

Issue #4755, opened 2018-05-22
Improve image handling in youtube_simple


bep:
Seems it is not possible to guess the URL to the best preview image version of a given YouTube video.

Some options to fix this is:

* Adding a positional param isn't easy to do in a non-breaking way (someone has been clever enough to add "class" as a second parameter). Also, this should ideally be a configuration toggle. Having to manually inspect and modify a thousand videos on a site does not sound ideal.
* Find another way to detect image size/whatever
* Integrate with the YouTube API, which probably requires API key handling and sound like work. 

If someone can come up with a simple approach that is also good, we can do it. But we may have to remove the simple variant for now.

/cc @onedrawingperday

onedrawingperday:
@bep 
It seems that the loading of the YouTube `maxresfefault.jpg` thumbnail is relevant to the video aspect ratio.

For example [this video](https://www.youtube.com/watch?v=TJAfLE39ZZ8) is low resolution and yet it has the [correct max res thumbnail](https://i.ytimg.com/vi/TJAfLE39ZZ8/maxresdefault.jpg)
[Here is the JSON response of the video](https://www.youtube.com/oembed?url=https://www.youtube.com/watch?v=TJAfLE39ZZ8&format=json)

While [this video](https://www.youtube.com/watch?v=d_EADBnXjXc) that is also low resolution has [the wrong thumbnail](https://i.ytimg.com/vi/EADBnXjXc/maxresdefault.jpg)
[Here is the JSON response of the video](https://www.youtube.com/oembed?url=https://www.youtube.com/watch?v=d_EADBnXjXc&format=json)

The low resolution videos with the wrong maxresdefault thumbnail have this value:
`"width": 459`

While videos with the correct thumbnail have: `"width": 480`

I have already tested this in a bunch of videos and I will test it some more to be 100% sure that this is the case.

But if what I found is valid for all cases then we could probably do the thumbnail detection.

If you don't like this approach. Then maybe we should just load the `hqdefault` thumb. This low res image exists for all YouTube videos. Of course it will look bad in high-res. But it's better than nothing.

Anyway let me know what you think and I could PR this later today.

P.S. Again my apologies for the erroneous PR . I shouldn't be doing PRs at 11 o' clock at night after a long day.

bep:
I assume that most newer videos is 720p+ and would have a high-res image. We **really want to show that high res image**. I would even say that it would be better with some false negatives.

[...]

PR #4756, opened 2018-05-22
Detect which thumb is loaded in youtube_simple

onedrawingperday:
Added the `base64Encode` func to detect the contents of `maxresdefault.jpg` because of the way YouTube handles thumbnails for low res videos.

If the base64 value is `Ly9pLnl0aW1nLmNvbS92aS85TXpuSmR5T2xfRS9tYXhyZXNkZWZhdWx0LmpwZw==` load the low res `hqdefault.jpg` thumb instead. 

~~~Also edited the CSS to include the styles for the low res videos.~~~

Fixes https://github.com/gohugoio/hugo/issues/4755

### DIFF
--- a/tpl/tplimpl/embedded/templates/shortcodes/__h_simple_assets.html
+++ b/tpl/tplimpl/embedded/templates/shortcodes/__h_simple_assets.html
@@ -3,19 +3,19 @@
 {{/* Only include once */}}
 {{  .Page.Scratch.Set "__h_simple_css" true }}
 <style>
-.__h_youtube, .s_youtube_simple {
+.__h_youtube {
    position: relative;padding-bottom: 56.23%;height: 0;
    overflow: hidden;
    max-width: 100%;
    background: #000;
    margin: 5px;
 }
-.__h_youtube img, .s_youtube_simple img {
+.__h_youtube img {
    min-width:100%;
    height:auto;
    color: #000;
 }
-.__h_youtube .play, .s_youtube_simple .play {
+.__h_youtube .play {
    height: 72px;
    width: 72px;
    left: 50%;
--- a/tpl/tplimpl/embedded/templates/shortcodes/__h_simple_assets.html
+++ b/tpl/tplimpl/embedded/templates/shortcodes/__h_simple_assets.html
@@ -3,19 +3,19 @@
 {{/* Only include once */}}
 {{  .Page.Scratch.Set "__h_simple_css" true }}
 <style>
-.__h_youtube {
+.__h_youtube, .s_youtube_simple {
    position: relative;padding-bottom: 56.23%;height: 0;
    overflow: hidden;
    max-width: 100%;
    background: #000;
    margin: 5px;
 }
-.__h_youtube img {
+.__h_youtube img, .s_youtube_simple img {
    min-width:100%;
    height:auto;
    color: #000;
 }
-.__h_youtube .play {
+.__h_youtube .play, .s_youtube_simple .play {
    height: 72px;
    width: 72px;
    left: 50%;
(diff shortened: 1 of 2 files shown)


tpl/tplimpl/embedded/templates/shortcodes/__h_simple_assets.html
```diff
@@ -3,19 +3,19 @@
...
-.__h_youtube {
+.__h_youtube, .s_youtube_simple {
    position: relative;padding-bottom: 56.23%;height: 0;
    overflow: hidden;
    max-width: 100%;
    background: #000;
    margin: 5px;
 }
-.__h_youtube img {
+.__h_youtube img, .s_youtube_simple img {
```
bep: The idea (which may or not be a good one) was to give users a stable class to match with their external stylesheets. It does not make sense to use it in the inline styles. And not in this PR, anyhow.

tpl/tplimpl/embedded/templates/shortcodes/youtube_simple.html
```diff
@@ -8,8 +8,12 @@
 {{ end }}
 {{ $secondClass := "s_youtube_simple" }}
 {{ $divID := printf "%s_%d" $secondClass (add $.Ordinal 1) }}
+{{ $tb := printf "//i.ytimg.com/vi/%s/" $id }}
+{{ $mx := printf "%smaxresdefault.jpg" $tb }}
+{{ $hq := printf "%shqdefault.jpg" $tb }}
+{{ $dt := $mx | base64Encode }}
```
bep: I'm not sure about what this is. This is an encoding of the URL? I cannot see how that can give us info about the image itself (size etc.)

tpl/tplimpl/embedded/templates/shortcodes/__h_simple_assets.html
onedrawingperday: Ok.

tpl/tplimpl/embedded/templates/shortcodes/youtube_simple.html
onedrawingperday: It gets the base64 value of [this](http://i.ytimg.com/vi/9MznJdyOl_E/maxresdefault.jpg) image that YT loads when there is no actual `maxresdefault.jpg`

The base64 value of this image is a given so if its base64 is present we will load the `hqdefault.jpg`

bep:
```bash
{{ $dt := $mx | base64Encode }}
```

Are you saying that mx contains the image? And not the URL to the image? I may need a good nights sleep ...

onedrawingperday: I have tested this locally and it fetches the correct thumb for the low-res videos.

Also I removed the CSS edits from this PR.

onedrawingperday: @bep 

`$dt` stands for detect 
`$mx` stands for `maxresdefault.jpg` 

The base64 value of [this](http://i.ytimg.com/vi/9MznJdyOl_E/maxresdefault.jpg) image is `Ly9pLnl0aW1nLmNvbS92aS85TXpuSmR5T2xfRS9tYXhyZXNkZWZhdWx0LmpwZw==`

So if that value is present then we skip `maxresdefault.jpg` and load `hqdefault.jpg`

bep: >I have tested this locally and it fetches the correct thumb for the low-res videos.

But does it work the other way around?

onedrawingperday: Why should it work the other way around?

YouTube loads this image when an actual maxresdefault.jpg is not available for low-resolution videos.

onedrawingperday: Ok. I just figured out why I was getting the correct thumbnail locally.

The decoded base64 value is for the url `//i.ytimg.com/vi/9MznJdyOl_E/maxresdefault.jpg`

And it's for that specific video. Doh!

I'm closing this PR. Sorry to waste your time. I'll look into this a bit more with a clear head tomorrow.

PR #4760, opened 2018-05-23
Remove unneeded 'chrome=1' X-UA-Compatible meta tag

coliff:
This was only ever needed for a really old Chromeframe plugin for old IE and was discontinued by Google long ago.


bep:
Thanks for these, but at least one of these are coming from other modules (/dcos is a Git sub tree), and it would be too much work to push forward the changes. I would appreciate if patches to the theme is added to https://github.com/gohugoio/gohugoioTheme etc.

Issue #4763, opened 2018-05-24
Icons for Simple mode Hugo shortcodes


onedrawingperday:
Related PR: #4747 
Related Issues: #4749 #4764

**Proposal A**
I designed these control icons for use in the Simple Mode Hugo shortcodes inspired by the hexagon that is used in the Hugo logo. 

![hugo-icons](https://user-images.githubusercontent.com/7918298/40473651-755bb0ea-5f45-11e8-9956-62bfca4934a1.png)

I have made 2 variations of the play button for you to choose what you prefer.

**Example display for YouTube/Vimeo with the first play button**

![yt-vimeo-hugo-1](https://user-images.githubusercontent.com/7918298/40473684-8f5a5a46-5f45-11e8-80e9-bec579444e92.png)

**Example display for YouTube/Vimeo with the second play button**

![yt-vimeo-hugo-2](https://user-images.githubusercontent.com/7918298/40473727-aa116bcc-5f45-11e8-9260-8030a03c9338.png)

**Example display for the Speaker Deck Simple Mode Shortcode**

![speaker-deck-hugo](https://user-images.githubusercontent.com/7918298/40473788-d386f742-5f45-11e8-85a7-abb204b9c497.png)

**Proposal B**
As an alternative the following Fork Awesome svg icons can be used in the Hugo Simple Mode shortcodes. The licence for these icons is fully Open Source and is available [here](https://forkawesome.github.io/Fork-Awesome/license/)

![fork-awesome-icons](https://user-images.githubusercontent.com/7918298/40473922-4880377a-5f46-11e8-8e6f-a00e84b173b6.png)

**Example display for YouTube/Vimeo with Fork Awesome play button**

![yt-vimeo-fork-awesome](https://user-images.githubusercontent.com/7918298/40474096-c25d8278-5f46-11e8-9c9a-2508ace50200.png)

**Example display for the Speaker Deck Simple Mode Shortcode with Fork Awesome Icons**

![speaker-deck-fork-awesome](https://user-images.githubusercontent.com/7918298/40473990-77b7014a-5f46-11e8-99dc-320fae852f32.png)

CC/ @bep and anyone else

bep:
I think we should go with the traditional round play button.

onedrawingperday:
Ok. Which play button mine of the Fork Awesome one?

What about Speaker Deck? Is the proposed display what you have in mind for this shortcode?

bep:
[...]

PR #4771, opened 2018-05-25
youtube_simple: play button + CSS fixes

onedrawingperday:
Reference #4763 

Added a SVG play button that I designed and release it under the [Apache 2.0 Licence](https://www.apache.org/licenses/LICENSE-2.0)

Also fixed the CSS after some more tests.


bep:
There is a test failure:

```bash
ERROR 2018/05/25 00:49:46 error processing shortcode "_internal/shortcodes/youtube.html" for page "simple.md": html/template:_internal/shortcodes/youtube.html: "\"" in attribute name: "1\" xmlns=\"http://www.w3.org/2000"
```

PR #4786, opened 2018-05-27
Internal Shortcodes: vimeo_simple + vimeo (default)

onedrawingperday:
Based on the work of @bep see: https://github.com/gohugoio/hugo/commit/448081b840db4a23c0c49c2d869ac207dcb6ac40
Related issue: #4749

Added the following:
- A simple variant for Vimeo 
- Modified the Privacy Config to include simple mode for Vimeo
- Added again `__h_simple_assets.html` and changed the classes prefix to `__h_video` since this is meant to be used in both Vimeo and (hopefully) YouTube simple shortcodes.
- Added a SVG play button that I designed and release it under the [Apache 2.0 Licence](https://www.apache.org/licenses/LICENSE-2.0)
- Also updated the standard Vimeo shortcode. 
- Removed the frameborder attribute from the iframe because it is [obsolete](https://www.w3schools.com/tags/att_iframe_frameborder.asp), added `border: 0;` and removed the 30px top-margin from the inline style.

**N.B.** 
I was unable to have these changes reflected in a compiled binary. I think I'm missing something. I was able to run `mage generate` but still I didn't see anything new when I run `go build`. 

I was only able to test these shortcodes locally when I included them in a Hugo project.

So this PR needs to be tested some more.

### DIFF
--- a/config/privacy/privacyConfig.go
+++ b/config/privacy/privacyConfig.go
@@ -82,6 +82,11 @@ type Twitter struct {
 // Vimeo holds the privacy configuration settingsrelated to the Vimeo shortcode.
 type Vimeo struct {
 	Service `mapstructure:",squash"`
+
+	// If simple mode is enabled, only a thumbnail is fetched from i.vimeocdn.com and
+ 	// shown with a play button overlaid. If a user clicks the button, he/she will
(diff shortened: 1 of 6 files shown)


onedrawingperday:
Needs. more work. There are errors and I am closing this until I sort them out

PR #4788, opened 2018-05-27
Add a BlackFriday option for rel="nofollow" on external links

### LINKED ISSUE #4722: Add a BlackFriday option for rel="nofollow" on external links
BlackFriday can already be [configured](https://gohugo.io/getting-started/configuration/#configure-blackfriday) to open external links in a new tab. It would be quite helpful if we can also specify the `rel=nofollow` tag on external links.

Luckily, BlackFriday already [seems to](https://godoc.org/github.com/russross/blackfriday) have a feature for that: the `HTML_NOFOLLOW_LINKS` constant flag. That GoDoc link doesn't state it, but it seems to be [implemented](https://github.com/russross/blackfriday/blob/master/html.go#L442) in such a way that the `rel=nofollow` tag is added to external links only. That would mean this feature is a good solution for this feature request, without adjustments or modifications on the part of Hugo's code base.

---

It might probably also useful to expose the `HTML_NOREFERRER_LINKS` BlackFriday setting at the same time, since it's a related link setting and might save work in having to revisit in the future. But I don't want to make the scope of this issue too big, so this is just a suggestion for the developer.

---

The forum discussion of this feature is here: [Setting rel=“nofollow” for external links in all articles](https://discourse.gohugo.io/t/setting-rel-nofollow-for-external-links-in-all-articles/10990).

stefanneuhaus:
Add a configuration option "nofollowLinks". When set to "true" the "HTML_NOFOLLOW_LINKS" flag is being passed to Blackfriday. Thereby all *absolute* links will get a "nofollow" value for the "rel" attribute.

Fixes #4722


Jos512:
Thank you for making this contribution!

PR #4795, opened 2018-05-29
Add PreprocessedContent

theory:
I wanted a way to get out the raw markdown after it was preprocessed for shortcakes and Emoji and before it was converted to HTML. This will allow me to publish plain text content as the raw markdown, but with the shortcodes removed and replaced with content appropriate to a plain-text output. So I poked at the internals and figured out how to get it to do it, though I'm sure it's not as efficient or Hugo-ish as one might want. LMK if you'd like it done differently, and pointers for adding tests would also be appreciated.

### DIFF
--- a/docs/content/en/templates/output-formats.md
+++ b/docs/content/en/templates/output-formats.md
@@ -68,7 +68,7 @@ Given a media type and some additional configuration, you get an **Output Format
 
 This is the full set of Hugo's built-in output formats:
 
-{{< datatable "output" "formats" "name" "mediaType" "path" "baseName" "rel" "protocol" "isPlainText" "isHTML" "noUgly">}}
+{{< datatable "output" "formats" "name" "mediaType" "path" "baseName" "rel" "protocol" "isPlainText" "isHTML" "noUgly" "noRender">}}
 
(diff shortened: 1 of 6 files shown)


theory:
Hrm, now that I'm playing with this patch, trying to add a shortcode with two templates --- on HTML, one plain text --- I'm finding there's no way to have different shortcode outputs for different output formats. :-( I suspect adding support for that would be a lot more work…

bep:
I think this needs to wait. We will eventually have to upgrade to v2 of Blackfriday and there are several other shortcode related issues in the pipeline that I'm not sure how is going to work re. this.

theory:
Any suggestion how to get this thing hooked up so that we can have output formats with content not processed by Blackfriday or whatever, @bep? Or would such a thing really just have to wait for BF2 integration?

bep:
I understand the need, but I just don't want to spend time on forcing this in now when we need to revisit this later. The BF2 upgrade is depending on a move to vgo as dependency manager, and this particular issue falls into the part where we decide how to model multiple rendering formats. There is an issue about a latex renderer, and I guess a "no renderer" would also fit your use case. In any case, we cannot do this in the form of hard-coded if-else constructs. And we need to wait for the other building blocks to fall in place.

theory:
Okay. I'll just keep my own copy of this for now so I can keep generating plain text output with raw shortcodes for now, [like this](https://justatheory.com/2015/04/the-watch-is-you/copy.text), and will keep my eye out for BF2 and renderer flexibility stuff to land before looking at it again. Thanks.

bep:
An added note: I think this should be a property of an output format; i.e. you can define a "content renderer" per output format (default HTML).

Issue #4728, opened 2018-05-15
Option to support unix-style path semantics


vassudanagunta:
Unix-style path semantics, in particular support for both absolute and relative paths, are well understood and have become the path "lingua franca", from file systems to HTML links. Hugo's `.Site.GetPage` method and `ref`/`relref` shortcodes go against these conventions:

- relative paths are not supported. (though there are unused artifacts in the code indicating that it was once the intent).
- leading "/"s are ignored. What looks like a relative path is actually absolute.

The latter adds unnecessary confusion for new users. The former is a significant functional disadvantage. Relative paths allow sections (directory branches) in the content tree to be self-contained, and facilitate moving or renaming them without breaking internal links. It is one of the main reasons HTML follows unix-style path semantics (and given that HTML is Hugo's most common target, it would be nice to have the same semantics). 

### relative path vs shorthand references 

Hugo supports what I'm calling "shorthand" refs for the `.Site.GetPage` method and the `ref`/`relref` shortcodes. [I'm not sure it was a good idea](https://github.com/gohugoio/hugo/issues/4727).

Looking at the discussion, shorthand refs came about to alleviate a pain point: One had to type the fully-qualified, absolute (from the content root) path to the target, no matter how deep, and even if one page was referring to another page in the same directory.

Supporting relative paths would have alleviated that pain without introducing ambiguity, and in a way that is arguably more powerful. 

To re-use the example from #4727:

```
content/
├── _index.md
├── help.md
├── blog
│   └── lessons.md
│   ├── lessons.nn.md
└── lessons
    ├── _index.md
    ├── links
    │   ├── _index.md
    │   └── help.md
    └── tags
        ├── _index.md
        └── help.md
```

- The most common cases of reference are very short:
  - sibling items: The Links lesson references its help page as `help.md`
  - subsection: The Lessons home refers to its various lessons simply: `links` and `tags`.
- The next most common cases are only slightly longer. But that extra length is *meaningful*:
  - items in the root: the Links lesson references the site-wide help as `/help.md` and the homepage as `/`.  The "/" prefix says with a single char: *I am referring to a root item!*
  - parent items: The Links lesson refers to the Lessons home page with a simple `../`. That three character prefix says *I'm pointing back out to my containing section*, and it will work even if Lessons is renamed to Tutorials.
  - subsection items:  If for any reason the Lessons home page want to refer the help pages for each of the lessons, the added prefix makes this appropriatel explicit: `links/help.md` and `tags/help.md`.
[...]

Issue #4727, opened 2018-05-15
GetPage, rel and relref accept ambiguous refs without warning


vassudanagunta:
Hugo supports "shorthand refs" when looking up or linking to other pages via the `.Site.GetPage` method and the `ref`/`relref` shortcodes. It makes authoring links a little easier, but it introduced number of problems:

- It is too easy for links to point to the wrong page. The author may not even be aware of the ambiguity, so may not now to invoke [the workaround](https://github.com/gohugoio/hugo/issues/4312#issue-290525225).
- Top-level pages and sections do not have a guaranteed-to-be-unique reference, so the workaround is not even available for such pages.
- It is far too easy for a once correct link to suddenly point to the wrong page on subsequent site builds, without warning.

The last one is the most dangerous.

### the ambiguity, illustrated

Take the following content directory:

```
content/
├── _index.md
├── help.md
├── blog
│   └── lessons.md
│   ├── lessons.nn.md
└── lessons
    ├── _index.md
    ├── links
    │   ├── _index.md
    │   └── help.md
    └── tags
        ├── _index.md
        └── help.md
```

The following table shows all the supported ways one can reference each page in calls to `.Site.GetPage`, `ref` or `relref`.  Ambiguous refs are marked with :warning:. Notice that two nodes do not have an unambiguous option. 

| path within `content/`  | "full" ref             | language independent "full" ref | shorthand ref                  |
| ----------------------- | ---------------------- | ----------------------------- | ------------------------------ |
| /_index.md              | (empty string)         |                               |                                |
| /help.md               | **help.md** :warning: | **help** :warning:    |                                |
| /blog                   | blog                   |                               |                                |
| /blog/lessons.en.md     | blog/lessons.md        | blog/lessons                  | lessons.md<br />**lessons** :warning:    |
| /blog/lessons.nn.md     | blog/lessons.nn.md     | blog/lessons                  | lessons.nn.md<br />**lessons** :warning: |
| /lessons/_index.md      | **lessons** :warning: |                               |                                |
| /lessons/links/_index.md | lessons/links |           |   |
| /lessons/links/help.md | lessons/links/help.md | lessons/links/help           | **help.md** :warning:<br />**help** :warning: |
| /lessons/tags/_index.md | lessons/tags |            |   |
| /lessons/tags/help.md  | lessons/tags/help.md  | lessons/tags/help            | **help.md** :warning:<br />**help** :warning:        |

### how a link gets modified without warning on site or Hugo updates  

[...]

Issue #4726, opened 2018-05-14
GetPage `kind` param is neither fully respected nor really needed


vassudanagunta:
`.Site.GetPage` takes a `kind` arg, but its implementation only partially respects it. For kind `"page"`, it looks up only *page* Pages, as expected. But for any of the four node-kinds, *home*, *section*, *taxonomy*, and *taxonomyTerm*, it doesn't similarly limit itself.

For example:

- `.Site.GetPage("home", "posts")` doesn't return the one and only page of kind "home", but instead returns the "posts" section page (it's a coincidence that the returned page is the "home" page of the section).
- `.Site.GetPage("section", "categories")` returns a page of kind taxonomyTerm if one or more pages have a category value in their front matter.
- `.Site.GetPage("taxonomy" "funk")` won't return the taxonomy page for "music", but will either fail or return a "music" section or taxonomyTerm page if one existed. Taxonomy GetPage calls must use the fully qualified name, e.g. `.Site.GetPage("taxonomy", "tags/funk")`. But then the "taxonomy" arg is redundant, as the "tags/" prefix already identifies this as a request for a taxonomy page as well as specifying which taxonomy ("tags").

This strange behavior is mostly innocuous, as the above are strange corner cases, and as the paths for any two node kinds, regardless of type, can never be the same.

But if the `kind` arg is eliminated:

- eliminates the potential for confusion/surprise
- simplifies the API
- gives `.Site.GetPage` a signature consistent with the `ref`/`relref` shortcodes, which perform essentially the same function.  

Pull request forthcoming...

kaushalmodi:
This was suggested a while ago here too: https://github.com/gohugoio/hugo/issues/4652.

.. but I didn't know that the behavior was inconsistent (as I never tried putting incorrect Kinds). 

I'd be interested in seeing what you propose in the PR. Would it break all the existing GetPages?

bep:
It think it will save time in the long run if you bring these issues you open to some kind of conclusion before you arrive with a PR.

The big question, in this case, is if it can be done without breaking existing code.

vassudanagunta:
@bep What are the benchmarks I should run for this?

Issue #4652, opened 2018-04-20
Make .Site.GetPage work with 1 parameter


bep:
Not sure why I insisted on the kind as a first param ... I don't see any ambiguous cases, and it should be possible to do in a non-breaking way, I think ...

kaushalmodi:
What is your use case for this?

This will just confuse the users and more so add a maintenance burden on your shoulders.

At the moment, the syntax is quite simple, first the Kind, then the Page name. 

If the first argument becomes (Kind or Page name), helping people understand that and debug that will be a pain. 

Also the current syntax where the Kind is mandated is very clean.. someone reading the code knows instantly of the section page is getting fetched or the "page" page, or ..

kaushalmodi:
If the motivation is that people don't have to understand Page Kinds, then that would be bad because knowing the Kind is useful in many places like output formats, and mainly grasping the layout lookup order.

bep:
I was writing some test-shortcodes for my work on the shortcodes, and if you want to write a generic "get-page-content" shortcode ... you either have to know what it is or try several.

So: `{{< get-page-content "section" "foo/mydirectory" >}}` would fail for no good reason if I change that dir into a ... leaf bundle.

bep:
But I'm not totally convinced...

vassudanagunta:
@bep, I'm still a Go newb, but I don't think it is possible to remove the typ param from:

```go
func (s *SiteInfo) GetPage(typ string, path ...string)`
```

because the first param ("page", "section", etc) in existing calls to `.Site.GetPage` would then come across as part of the variadic path arg, right?

I was going to propose a combined solution to this and #4728 by adding `.GetPage` (i.e. Page) and deprecating `.Site.GetPage`, which would allow the method to support relative paths:

```go
func (p *Page) GetPage(path ...string)
```

[Though I was wondering](https://discourse.gohugo.io/t/why-does-getpage-support-paths-represented-as-multiple-strings/11989) if the new method can dispense support for passing the path in broken up into separate strings:

```go
func (p *Page) GetPage(path string)
```

The nice thing about this is that `.GetPage` would take the same arg and have the same semantics as the `ref`/`relref` shortcodes. Consistency makes for a less confusing API.

PR #4796, opened 2018-05-30
GetPage, ref and relref improvements

vassudanagunta:
Below are a number of issues related to how one looks up or links to pages in Hugo.

- Problems:
  - [#4727](https://github.com/gohugoio/hugo/issues/4727): GetPage, rel and relref accept ambiguous refs without warning.
  - [#4147](https://github.com/gohugoio/hugo/issues/4147) and [#4130](https://github.com/gohugoio/hugo/issues/4130): ref and relref need to support all kinds of pages.
    Changes introduced in v0.33 [make this a must](https://github.com/gohugoio/hugo/issues/4147#issuecomment-389065470) as the use of refs rather than hardcoding is necessary for dynamic handling of changed URLs.
  - [#4726](https://github.com/gohugoio/hugo/issues/4726): GetPage neither respects nor needs its `kind` arg. [#4652](https://github.com/gohugoio/hugo/issues/4652) points out it isn't needed.
- Nice-to-haves:
  - [#4728](https://github.com/gohugoio/hugo/issues/4728): Option to support unix-style path semantics.

There was [some discussion of this](https://discourse.gohugo.io/t/getpage-ref-and-relref-renovations/12103/4) on the Hugo forum. 

## These issues *beg* to be resolved together rather than piecemeal

If you start working on the above issues, you quickly discover how much the solutions overlap or have dependencies with each other. 

This is an opportunity to provide authors an improvement in how they author links and dependencies between content:

- **simple** and **consistent**: `Page.GetPage`, `Page.Ref` and `Page.RelRef`, `ref` and `relref` should all take the same single arg with the same semantics.
- **natural**: unix path semantics are well understood. The web is built on it. It is the most natural way to reference hierarchically organized content, but also works equally well for flat content with the use of relative paths. In fact, within a section, everything looks flat with relative paths, no matter how deeply nested the section is.
- **reliable**: Authors manually check links before publishing. Once it is published, they will not check again. Authors need confidence that link targets don't change out from under them. Relative links are perfectly succinct while being unambiguous. But if ambiguous and "mostly stable" shorthand or even wiki-links are to be supported, authors should be warned when a link is ambiguous or when a change is made that causes a link to become ambiguous. The ambiguity detection in this PR provides that.  
- **self-contained sections**: Links *between* elements within a section of self-contained content should not have any coupling or dependencies on anything outside the section, including the location of the publish context of the section itself. Relative links provide that. Complex websites rely heavily on this principle, as does complex code (encapsulation).

### Summary of Changes

As always i've broken down the changes into semantically discrete commits that are human parsable and can be reasoned about individually, and submitted them individually so tests are run against each.

Please see the comments on the individual commits and the linked issues for details. But here is a quick summary of the outward changes:

- `Site.GetPage`, `ref` and `relref` now work across all pages. The `kind` is ignored.

- There is a new `Page.GetPage` that, like `ref` and `relref` , takes a single `ref` arg. All three operate on a Page context and interpret the ref arg on that context in exactly the same way.  `Site.GetPage` is supported for backward compatibility, but `Page.GetPage` should be used instead.

- In addition to all of the previous types of refs you could pass to these methods, you can now pass absolute and relative refs with unix path semantics. e.g.:
  - `/section/subsection/page.md`
  - `sibling-page.md`
  - `..`  (parent section)
  - `../page-in-parent-section.md`
  
  Absolute or relative refs are a way, and in some cases the only way, to avoid ambiguity.  For example, if there are multiple content files named `sibling-page.md`, because of the continued support for shortcut refs you have to use a unix absolute path, or if you need a relative path, force the ref to be interpreted as such and not as a ambiguous shortcut ref with `./sibling-page.md`.
  
  `Site.GetPage` has no page context so does not support relative refs; Use `Page.GetPage`.

- If called with an ambiguous ref, an error with a helpful message will be generated, e.g.: 
    ```
    The reference "help.md" in "/tutorial/_index.md" resolves to more than one page.
    Use either an absolute path (begins with "/") or relative path to the content
    directory target.
    ```

I ran benchmarks at three points (see commit comments). Performance differences swung both ways and are probably within the margin of benchmark error.

## TODO

All the existing tests pass. That's pretty solid coverage. But I will add some additional tests once there is buy-off of this PR.

### DIFF
--- a/hugolib/disableKinds_test.go
+++ b/hugolib/disableKinds_test.go
@@ -130,7 +130,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindPage, "public/sect/p1/index.html", "Single|P1")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p := s.getPage(KindHome)
+			p, _ := s.getPage("/")
 			if isDisabled {
 				return p == nil
 			}
@@ -138,7 +138,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindHome, "public/index.html", "Home")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p := s.getPage(KindSection, "sect")
+			p, _ := s.getPage("sect")
 			if isDisabled {
 				return p == nil
 			}
@@ -146,7 +146,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindSection, "public/sect/index.html", "Sects")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p := s.getPage(KindTaxonomy, "tags", "tag1")
+			p, _ := s.getPage("tags/tag1")
 
 			if isDisabled {
 				return p == nil
@@ -156,7 +156,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindTaxonomy, "public/tags/tag1/index.html", "Tag1")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p := s.getPage(KindTaxonomyTerm, "tags")
+			p, _ := s.getPage("tags")
 			if isDisabled {
 				return p == nil
 			}
@@ -165,7 +165,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindTaxonomyTerm, "public/tags/index.html", "Tags")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p := s.getPage(KindTaxonomyTerm, "categories")
+			p, _ := s.getPage("categories")
 
 			if isDisabled {
 				return p == nil
@@ -175,7 +175,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindTaxonomyTerm, "public/categories/index.html", "Category Terms")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p := s.getPage(KindTaxonomy, "categories", "hugo")
+			p, _ := s.getPage("categories/hugo")
 			if isDisabled {
 				return p == nil
 			}
--- a/hugolib/disableKinds_test.go
+++ b/hugolib/disableKinds_test.go
@@ -130,7 +130,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindPage, "public/sect/p1/index.html", "Single|P1")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p, _ := s.getPage("/")
+			p, _ := s.getPage(nil, "/")
 			if isDisabled {
 				return p == nil
 			}
@@ -138,7 +138,7 @@ func assertDisabledKinds(th testHelper, s *Site, disabled ...string) {
 		}, disabled, KindHome, "public/index.html", "Home")
 	assertDisabledKind(th,
 		func(isDisabled bool) bool {
-			p, _ := s.getPage("sect")
+			p, _ := s.getPage(nil, "sect")
 			if isDisabled {
 				return p == nil
 			}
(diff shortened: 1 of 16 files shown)


bep:
I have one important comment: Site.getPage needs to work as before, so adjusting million tests to some new format is counter-productive (it is much better that they stay as is for now to confirm the existing behaviour).

vassudanagunta:
@bep There are *exactly* two callers to`PageCollections.getPage` outside of the tests: Site.GetPage and Site.refLink (used by ref and relref). 

I went with that approach first.... Not change PageCollections.getPage

vassudanagunta:
The test failure is bad test code (prod code is fine) that I added that doesn't work for Windows. Now I have to do another `rebase -i`...

bep:
OK, I will wait, but I will repeat myself (because I think it's important):
 
This PR is (I think) about adding `.Page.GetPage`. It is not about changing the signature of `.Site.getPage` or `.Site.GetPage`. So to me, it would make sense to (at least now) keep the tests for `.Site.getPage` as-is (with kind argument). 

As it is now you will need to write new tests for the tests you rewrite. Which does not sound optimal.

But up to you.

vassudanagunta:
@bep I'm busy right now will get back to address your points/questions in a day or two.

vassudanagunta:
@bep, The new `Page.GetPage` may have been your focus, but it wasn't mine.  That method is rather a small natural consequence of the things I think are important: *a **simple**, **consistent**, **natural**, **reliable** way to author links and dependencies between content*, in a way that allows sections of content to **be self-contained**. 

Please read the ***These issues beg to be resolved together rather than piecemeal*** section that I just added to the PR description above. If you agree with that section, then we can proceed with a code review where I address your implementation concerns. Please let me know.

bep:
>All the existing tests pass. That's pretty solid coverage.

I see a lot of test adjustments related to this, which in my head isn't the same as having the existing tests pass.

My number one concern about this is breaking sites in the wild. Are there known situations where the "old" will not work? I.e. you would have to adjust your content/templates to get it right.

bep:
Note that the implementation itself looks solid (I have only quickly read through it), but it would be much easier to get an opinion on the side effects if you kept the old tests as is. I.e. created a new `.Site.getNewPage` or something.

vassudanagunta:
@bep https://github.com/upend/IF_MS_BUYS_GITHUB_IMMA_OUT and it's popularity has been taking up my free time. Would you mind reading my last comment and making a decision. I'll do what you decide if I have the time. I'd like to do it before I leave GitHub.

bep:
@vassudanagunta I'm on vacation where there is sun, so this issue is not on the top of my list ... And please stop posting that GitHub vs MS propaganda in the Hugo forums.

vassudanagunta:
@bep, why do you have to be so unkind?  I posted it once, and said I'd understand if it was deemed off-topic, and didn't protest or repost when it was shut down. Rather than respecting a difference of opinion, you dismiss mine as being "dramatic" and essentially accuse me of being dishonest. Why do I bother contributing? Do you realize there is *nothing* in it for me? I don't need the hassle or the unkindness.

bep:
@vassudanagunta what is the status of this and the other WIP pull request you have open?

vassudanagunta:
@bep I spent a lot of time explaining this PR in its description and then in [a reply to your concerns](https://github.com/gohugoio/hugo/pull/4796#issuecomment-394044573). Likewise on description of [the other one](https://github.com/gohugoio/hugo/pull/4379). What more do you want me to say?

I did neither PR or any of the others I've submitted for my own benefit. If they are not valuable please just say so. I don't want to waste any more time. It is so much cheaper and more efficient to just use my own fork.

bep:
No, just wanted to chime in and see if you are still active on GitHub,

vassudanagunta:
I'm minimizing my activity here. If there is a project that I'd like to help, and I have no other option, I'll do it through here. My own projects will move elsewhere.

stp-ip:
Thanks for all the effort. This would be a great addition. I just wanted to jump in and give my appreciation for the efforts and looking forward to see this integrated.

bep:
I will get back to this once I get some other major work out of the way ... should not be too long. This in general looks very good.

vassudanagunta:
git rebase; push -f 2-month old commits

bep:
Will finish this in #4955

PR #4797, opened 2018-05-30
Fix vimeo_simple  thumb scaling

onedrawingperday:
As the greek saying goes _το δις εξαμαρτείν ουκ ανδρός σοφού_ (making the same mistake twice does not befit the wise) 

-but I did and here is the fix.-


bep:
Can you fix the commit message? I fixed it in the PR yesterday, but it would save time if you could get it right yourself. I know I'm a little pedantic in this area, but I think it's important.

It should be mentioned in the guidelines.

onedrawingperday:
@bep 
Changed the commit message to `Fix vimeo_simple thumb scaling`
Better?

PR #4799, opened 2018-05-30
tpl: Remove frameborder attr YT iframe + CSS fixes

onedrawingperday:
This PR contains the following:
- Removed the frameborder attribute from the iframe because it is [obsolete](https://www.w3schools.com/tags/att_iframe_frameborder.asp) and added `border: 0;` in the inline style
- Also removed frameborder from the test because the test fails if the attribute is not removed ([catch-22](https://www.merriam-webster.com/dictionary/catch-22)).
- On the container div I removed the `padding-top: 30px;`  from the inline style. 

What the `padding-top` CSS property does can be seen at the Hugo Docs [YT display](https://gohugo.io/content-management/shortcodes/#example-youtube-display). 

Basically it expands the `iframe` height and there are 2 scenarios depending on the browser:

1. The iframe will have black bars in the top & bottom
2. The iframe will expand to fill the `div `height and parts of the video from the left & right will be chopped off from the display.

And that's why it needs to go.

### DIFF
--- a/hugolib/embedded_shortcodes_test.go
+++ b/hugolib/embedded_shortcodes_test.go
@@ -197,17 +197,17 @@ func TestShortcodeYoutube(t *testing.T) {
 	}{
 		{
 			`{{< youtube w7Ft2ymGmfc >}}`,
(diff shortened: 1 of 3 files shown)


onedrawingperday:
I will be unavailable for the rest of the day. If something is wrong with this PR I will see to it tomorrow.

PR #4800, opened 2018-05-30
WIP: Bibliography with Pandoc

bilke:
It would be very handy to generate bibliographies with Hugo, see #4060.

This PR adds two possible implementations (see each commit):

1. Pass frontmatter into the external helper and check for the `bibliography`-key and pass its value to Pandoc, passes CI ✅ 
2. Simply add the frontmatter into the processing of the content on the external helper. This allows for even more functionality such as creating full bibliographies without citing every entry (in frontmatter add: `nocite: '@*'`) but is obviously more intrusive; another option would be to create a dedicated frontmatter entry (e.g. `external-frontmatter`) and to pass just this. Does not pass CI at the moment 🔴 

What do you think?

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -455,14 +455,15 @@ func ExtractTOC(content []byte) (newcontent []byte, toc []byte) {
 // for a given content rendering.
 // By creating you must set the Config, otherwise it will panic.
 type RenderingContext struct {
-	Frontmatter  map[string]interface{}
-	Content      []byte
-	PageFmt      string
-	DocumentID   string
-	DocumentName string
-	Config       *BlackFriday
-	RenderTOC    bool
-	Cfg          config.Provider
+	Frontmatter    map[string]interface{}
+	FrontmatterRaw []byte
+	Content        []byte
+	PageFmt        string
+	DocumentID     string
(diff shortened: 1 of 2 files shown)


lyndondrake:
I'd be very keen to test this, if it's still being progressed?

bilke:
Still waiting for feedback. In the meantime I just rebased on master.

bep:
@bilke sorry for the late feedback on this. I can see how this would be useful and the implementation looks good. 

But I will have to wait with this a little. Now with Go 1.11 coming out I can finally move this project to Blackfriday v2. And with that, I plan to do a redesign of the whole "content rendering" thing, including how PanDoc is set up. One of those items is configuration. I'm not a fan of adding top-level front matter flags (`bibliography`) just for PanDoc. But I would have to think a little how to do this.

So, this feature itself is great, just that I have to wait a little before we marry into the API.

ankitp94:
Any chance this will be integrated into the official build?

bwc69151:
@jmooring still relevant?

jmooring:
Closing. See https://github.com/gohugoio/hugo/issues/15062.

PR #4802, opened 2018-05-30
Add twitter_simple

onedrawingperday:
This PR contains the following:
- Simple variant for the internal Twitter shortcode with a basic set of inline styles (see attached screenshot of proposed display below)

- The CSS classes of the `blockquote` and its contents are provided by the Twitter oEmbed API.

- If the user wishes to disable the inline styles, it is possible to do so with the following configuration:
```
[services]
[services.twitter]
disableInlineCSS = true
```
twitter_simple


bep:
This looks solid, thanks.

Issue #4562, opened 2018-04-03
Figure shortcode with baseURL


qw3ry:
Currently, the `figure` shortcode does not work for [sites with a non-standard BaseURL](https://discourse.gohugo.io/t/solved-how-to-add-relurl-to-figure-src/9932).

Because I need this for my current project I coded a custom shortcode prefixing image URLs starting with a `/` with the BaseURL. Basically I fix this by replacing `<img src="{{ .Get "src" }}" … />` with `<img src="{{ .Get "src" | relUrl }}" … />`

I'd be happy to create a PR if this is a wanted addition/change.

kaushalmodi:
> does not work for sites with a non-standard BaseURL.

It wasn't clear from the linked page, but I believe you refer to BaseURL's like `https://example.com/foo/` as non-standard BaseURL. If so, I agree!

This has been a known problem and has appeared on the [Discourse forum][2] too.

> Basically I fix this by replacing <img src="{{ .Get "src" }}" … /> with <img src="{{ .Get "src" | relUrl }}" … />

That, IMO, is not a complete fix. It does not handle the case when images are in one of the content dirs.

I ended up with [this custom `figure` shortcode][1], but here is the relevant piece:

```html
{{ $permalink := $.Page.Permalink }}
{{ $image := .Get "src" }}
{{ $image_link_absolute := (findRE "^/" $image) }}
<figure{{ with .Get "class" }} class="{{ . }}"{{ end }}>
    <!-- snip -->
        {{ if $image_link_absolute }}
            <img src="{{ $image | absURL }}"
        {{ else }}
            <img src="{{ (printf "%s%s" $permalink $image) }}"
        {{ end }}
    <!-- snip -->
```

- I am simply using `absURL` instead of `relURL` in your case. But that's just 1/2 of the fix. That deals with only `src` values that begin with `/`. It is safe to assume that if a user starts `src` value with `/`, they mean to have that path start from the site's BaseURL.
- The other 1/2 of the solution is when `src` does **not** begin with `/`. In that case, I assume that user wants to refer to that `src` in the "current dir" i.e. from the current Page's Permalink. So I prefix the `src` with `$.Page.Permalink` in that case.


[1]: https://gitlab.com/kaushalmodi/hugo-theme-refined/blob/master/layouts/shortcodes/figure.html
[2]: https://discourse.gohugo.io/t/embedding-page-resources-into-posts/11120

tomnis:
[...]

PR #4801, opened 2018-05-30
Make figure also insert absolute links in the src attribute

kaushalmodi:
- Fixes https://github.com/gohugoio/hugo/issues/4562
  - So now the figure shortcode inserted images work fine on list pages and on
    sites with baseURL containing subdir too.

Other:

- Consistency edits: {{.}} -> {{ . }}, {{else}} -> {{ else }}, and similar

Issue #4574, opened 2018-04-05
Add a Markdown URL rewriter


bep:
So, we have the `relref` and `ref` shortcodes to help with linking in content files, and we have page
 bundles which enables page-relative linking of image resources etc.

But, given this example:

```markdown
[I'm an inline-style link](/posts/my-post)
![alt text](/images/logo.png
```

All of the above are relative to the host (server) root. Which works fine ... most of the time.

But it gets challenging for sites with many languages ... 

```markdown
[I'm an inline-style link](/jp/posts/my-post)
```

Esp. if the translation method is to just copy the English content tree and start translating.

1. It is easy to do transformations of the links above.
2. It is, however, hard to find a "good rule that fits all".

So, I suggest that we add a `UrlRewriter` (for content only for starters), configurable either globally or per language:

```toml
[languages.en]
languageName = "English"
weight = 1

[languages.nn]
languageName = "Nynorsk"
weight = 2
contentDir = "content_nn"
[[contentURLWriter.nn]]
type = "link" # or image (or all)
pattern = ".*"
rewrite = `{{ .URL | relLangURL }}`
[[contentURLWriter.nn]]
type = "image" # or images (or all)
pattern = ".png"
rewrite = `{{ .URL | upper }}`
```

The above example is quickly typed and the API can be improved, but my thoughts are:

* We pick the first matching regexp for a given link
* We treat the `rewrite` rules as Go templates with both a `.Site` `.URL` and all of Hugo's template funcs.

RealOrangeOne:
Suggestion, depending on the implementation complexity: Adding `.Page` to the context?

bep:
>Adding .Page to the context?

A big maybe. It will eventually lead to questions about "I cannot access the content" ... which is of course not possible.

bep:
Thinking about this ... when I get my lazy self out of the couch and upgrade to Blackfriday 2, we should be able to do more in this department, so possible a more general "filter syntax" would be good. Suggestions welcome.

jmooring:
This is supported via markdown render hooks.

PR #4803, opened 2018-05-30
Make figure SC insert absolute links in the src attribute

kaushalmodi:
- Fixes https://github.com/gohugoio/hugo/issues/4562
  - So now the figure shortcode inserted images work fine on list pages and on
    sites with baseURL containing subdir too.
- Support figure SC src attribute with protocol or two leading slashes
- Update the figure shortcode + src attr tests to use BaseURL
- Add tests for all of the above.

Other:

- White-space change syncup between figure SC and tests
- Consistency edits: {{.}} -> {{ . }}, {{else}} -> {{ else }}, and similar

### DIFF
--- a/hugolib/embedded_shortcodes_test.go
+++ b/hugolib/embedded_shortcodes_test.go
@@ -123,22 +123,22 @@ func TestShortcodeFigure(t *testing.T) {
 	}{
 		{
 			`{{< figure src="/img/hugo-logo.png" >}}`,
-			"(?s)\n<figure>.*?<img src=\"/img/hugo-logo.png\" />.*?</figure>\n",
+			"(?s)\n<figure>.*?<img src=\"/img/hugo-logo.png\"/>.*?</figure>\n",
 		},
 		{
 			// set alt
 			`{{< figure src="/img/hugo-logo.png" alt="Hugo logo" >}}`,
-			"(?s)\n<figure>.*?<img src=\"/img/hugo-logo.png\" alt=\"Hugo logo\" />.*?</figure>\n",
+			"(?s)\n<figure>.*?<img src=\"/img/hugo-logo.png\" alt=\"Hugo logo\"/>.*?</figure>\n",
 		},
 		// set title
 		{
 			`{{< figure src="/img/hugo-logo.png" title="Hugo logo" >}}`,
-			"(?s)\n<figure>.*?<img src=\"/img/hugo-logo.png\" />.*?<figcaption>.*?<h4>Hugo logo</h4>.*?</figcaption>.*?</figure>\n",
+			"(?s)\n<figure>.*?<img src=\"/img/hugo-logo.png\"/>.*?<figcaption>.*?<h4>Hugo logo</h4>.*?</figcaption>.*?</figure>\n",
 		},
 		// set attr and attrlink
 		{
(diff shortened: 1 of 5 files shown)


kaushalmodi:
@theory This is a continuation of https://github.com/gohugoio/hugo/pull/4801.

That PR got auto-closed in my attempt to squash all the commits in there.

theory:
[AbsURL](https://github.com/gohugoio/hugo/blob/master/helpers/url.go#L160) is so close already; would be nice if the `absURL` template function (and `relURL`) could be updated to be aware of thee current page path. Would save a lot of jiggery pokery in templates like this.

bep:
When I see constructs like the below I think it is time to stop and try to improve:

1)

```
  {{- /* Cannot use absURL below because it doesn't work as expected if baseURL has a subdir. */ -}}
```

2)

```
 {{- $permalink_pretty := $.Page.Permalink | replaceRE "\\.html$" "/" -}}
-        {{ if .Get "attrlink"}}</a> {{ end }}	+            <img src="{{ (printf "%s%s" $permalink_pretty $image) }}"
```

1) If there is a problem with `absURL`, then `absURL` needs to be fixed. We cannot create elaborate workaround in other parts of Hugo just to get this kind of working.
2) This probably will work in most cases, but it does not look optimal and you will get dizzy testing creating all the shortcode test-variations.

So, if we skip the `absURL` debate for now ... I think what we really need is a well understood and specified `.Page.AbsURL <some-path>`, which, when provided a path without a leading slash will give you the ´Page`-relative URL, else the "base-URL relative" URL (we need to agree on this). And make it clear that it does not respect the `canonifyURLs` setting (we have another issue for that).

We probably need a better name for it. But method like that would make this a lot more solid.

Also, see #4574

theory:
So should there be a method on the Page object to which one can pass the name of a file in the bundle and get back a URL?

I think the image shortcode needs to be able to handle all the possibilities, since one can't tell in advance whether the `src` param will be a full URL, a full path, a relative path, or the name of a bundle file. Unless there are different params for each case. Or else it shouldn't care at all, in which case the author would need to call a function (like the above-described `Page.AbsURL`) as appropriate given the image they actually want to reference.

kaushalmodi:
I don't know how `Page.AbsURL` solves this problem.. unless it takes in an argument that needs to be urlized with respect to that Page.

As @theory mentioned, it would need to handle all those cases at the `figure` SC handles in this PR.

But rather than having `absURL` behave differently than this new `Page.AbsURL` (or whichever name we pick), it would be nice to fix `absURL` instead.. i.e. make it behave as the `figure` SC fix behaves in this PR.

theory:
I looked at [the `PathSpec` source](/gohugoio/hugo/blob/master/helpers/pathspec.go), and it doesn't look like [`AbsURL`](/gohugoio/hugo/blob/master/helpers/url.go#L160) has access to or is aware of the presence of the current `Page` object. Which would make it difficult to prepend the Page bundle path to the URL.

kaushalmodi:
@theory Your edit in https://github.com/gohugoio/hugo/pull/4803#issuecomment-393529574 looks nice. Can you open a new issue proposing this new function.

We can discuss further over there, including discussing a name for that function.. `AbsURL` would be too confusing.

And I will close this PR for the sake of the "right thing to do" i.e. have a "Page.AbsURL" :)

theory:
Done in #4804.

PR #4805, opened 2018-05-31
Add repeat function.

theory:
Tempted to add `join`, too.


theory:
Plain text output. I'm using it to build tables and whatnot.

theory:
Gonna be looking to make something to wrap text soon, too, but all this is making me think there needs to be an easier way to write, like, plugins or something. Dunno how to do what when the output is a static binary, though.

bep:
I was going to just merge this, but I will just chime in with a thought. The idea with the namespace/alias setup was to avoid putting everything into one global namespaces -- and possibly get some help in the docs department (the dochelper JSON for template funcs is currently not in use).

So, the alias is meant for _important_ core functions (i.e. `where` etc.). We have some that does not match the important category, but we cannot remove the.

This is, hower, not a core function, it is a strings utility function. One might imaging we get a repeat function for some other "thing", and it would be bad to have an alias for this.

I'm happy to add these template functions, but could you nil out the alias field and put the docs in `strings.Repeat` etc.

Yes, I know that we should somehow autogenerate the docs for this ...

bep:
Also maybe add add some upper limit to avoid out of mem pranks...

theory:
Maybe just use int16 to limit it to `(-32768 to 32767)`?

PR #4807, opened 2018-06-01
Add runelen function

theory:
Unlike RuneCount, it does not strip HTML or whitespace before counting runes. Useful
for formatting plain text.

### DIFF
--- a/docs/content/en/functions/runelen.md
+++ b/docs/content/en/functions/runelen.md
@@ -1,0 +1,28 @@
+---
+title: runelen
+description: Determines the number of runes in a string.
+godocref:
+date: 2018-06-01
+publishdate: 2018-06-01
+lastmod: 2018-06-01
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [counting, character count, length, rune length, rune count]
+signature: ["runelen INPUT"]
+workson: []
+hugoversion:
+relatedfuncs: []
+deprecated: false
+aliases: [/functions/runelen/]
+---
+
+In contrast with `countrunes` function, which strips HTML and whitespace before counting runes, `runelen` simply counts all the runes in a string. It relies on the Go [`utf8.RuneCount`](https://golang.org/pkg/unicode/utf8/#RuneCount) function.
+
+```
+{{ "Hello, 世界" | runelen }}
+<!-- outputs a content length of 9 runes. -->
+```
+
+[pagevars]: /variables/page/
--- a/docs/content/en/functions/runelen.md
+++ b/docs/content/en/functions/runelen.md
@@ -18,11 +18,11 @@ deprecated: false
 aliases: [/functions/runelen/]
 ---
 
-In contrast with `countrunes` function, which strips HTML and whitespace before counting runes, `runelen` simply counts all the runes in a string. It relies on the Go [`utf8.RuneCount`](https://golang.org/pkg/unicode/utf8/#RuneCount) function.
+In contrast with `countrunes` function, which strips HTML and whitespace before counting runes, `runelen` simply counts all the runes in a string. It relies on the Go [`utf8.RuneCount`] function.
 
 ```
 {{ "Hello, 世界" | runelen }}
 <!-- outputs a content length of 9 runes. -->
 ```
 
-[pagevars]: /variables/page/
+[`utf8.RuneCount`]: https://golang.org/pkg/unicode/utf8/#RuneCount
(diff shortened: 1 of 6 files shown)


docs/content/en/functions/runelen.md
```diff
@@ -0,0 +1,28 @@
+---
+title: runelen
```
bep: Now the documentation does not match the name of the function. And maybe we should call it `strings.RuneCount`; I know it is slightly confusing with the `strings.CountRunes` -- but it matches the way we do other "delegate to stdlib" situations?

theory: Yeah, I was trying to make it a closer parallel to `len`, thinking to leave `*count*` as a category where Hugo might get more clever (words, runes without markup or whitespace, etc.). Functions with `len` in them might be expected to be much ore straightforward. At least, that's what I was thinking. I think it would be smart to keep the name `CountRunes` in the `stings` namespace, though, so it's the same as the Go library, and with that thought that one day there would be some sort of tool to generate a complete export of a package's functions (and docs) by using introspection or something, rather than the ad-hoc exports made today by folks like me who get the itch.

No, wait, `strings.CountRunes` already means something else. Grrrr. Maybe no one uses that, in favor of the `countrunes` alias?

bep: >strings.CountRunes 

Yes, but `strings.RuneCount` does not, which I will admit is confusing, but I don't think we should make up new names if we can avoid it.

theory: Yeah, the point I was trying to make is that, ideally, these things would have the same names and semantics as in the Go strings package, on the assumption that, at some point, these libraries of utility functions could be autogenerated. But perhaps such an would have to break things, anyway, and it just becomes Hugo 2?

I'm not such a fan of `RuneCount` because it sounds more like a noun (field) than a verb (function). We could make it `CountChars` or `CountPoints` or something. I think a new name is less confusing than two names that are so similar. As Larry Wall [has said](https://www.computerworld.com.au/article/270267/a-z_programming_languages_perl/?pp=4), ”Similar things should look similar but similar things should also look different, and how you trade those things off is an interesting design principle.” Which is why I thought using "Len" in the name would be useful. But then the thing I find myself frequently say is “naming things is hard.”

Anyway, gimme a ruling and I'll change it.

bep: The ruling is `strings.CountRunes`.

I have no big interest in "auto generating" the template funcs themselves. That implies a need for many/all of them. I want to keep the count down and only add the ones "we really need". I want to do something about auto-docs, but that is a little beside the point.

theory: Er, `strings.CountRunes` already exists. Do you want me to rename it? Or did you mean `RuneCount`?

See, we're already getting confused! :-(

theory: Oh wait, I'm using `utf8.CountRunes`. Maybe we should put it in a `utf8` namespace?

docs/content/en/functions/strings.RuneCount.md
```diff
@@ -0,0 +1,28 @@
...
+godocref:
+date: 2018-06-01
+publishdate: 2018-06-01
+lastmod: 2018-06-01
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [counting, character count, length, rune length, rune count]
+signature: ["strings.CountRunes INPUT"]
```
bep: RuneCount

theory: Bah, was halfway through making it `CountRunes` before I remembered the conflict. Fixed.

rdwatters:
Why not “strings.TextLen” or “strings.RunesLen” or “strings.CountAllRunes”? Apologies for the late hop into the conversation. Very likely I’m missing something important, but sometimes naming benefits from outside perspective :bow:

docs/content/en/functions/strings.RuneCount.md
```diff
@@ -0,0 +1,28 @@
...
+workson: []
+hugoversion:
+relatedfuncs: ["len", "countrunes"]
+deprecated: false
+aliases: []
+---
+
+In contrast with `countrunes` function, which strips HTML and whitespace before counting runes, `RuneCount` simply counts all the runes in a string. It relies on the Go [`utf8.RuneCount`] function.
```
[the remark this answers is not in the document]
theory: Done.

theory:
`strings.CountAllRunes` makes sense to me, probably less confusing.

theory:
Or `RuneCountInString`, since that's the name in the `utf8` package.

PR #4824, opened 2018-06-06
Small typo

SvenKortekaas:
Fixed a small typo in the privacyConfig_test.go file


bep:
Thanks for this, but it isn't actually a typo. It is just there to test for potential case issues (the test is even named `TestDecodeConfigFromTOMLCaseInsensitive`).

PR #4826, opened 2018-06-06
Always load script over HTTPS

coliff:
This is potentially safer, faster and recommended by Google
REF: https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet

### DIFF
--- a/hugolib/embedded_templates_test.go
+++ b/hugolib/embedded_templates_test.go
@@ -50,7 +50,7 @@ Disqus:
 	b.AssertFileContent("public/index.html",
 		"'anonymizeIp', true",
 		"'script','https://www.google-analytics.com/analytics.js','ga');\n\tga('create', 'ga_id', 'auto')",
-		"<script async src='//www.google-analytics.com/analytics.js'>")
+			    "<script async src='https://www.google-analytics.com/analytics.js'>")
 
 	// Disqus
 	b.AssertFileContent("public/index.html", "\"disqus_shortname\" + '.disqus.com/embed.js';")
--- a/hugolib/embedded_templates_test.go
+++ b/hugolib/embedded_templates_test.go
@@ -50,7 +50,7 @@ Disqus:
 	b.AssertFileContent("public/index.html",
 		"'anonymizeIp', true",
 		"'script','https://www.google-analytics.com/analytics.js','ga');\n\tga('create', 'ga_id', 'auto')",
-			    "<script async src='https://www.google-analytics.com/analytics.js'>")
+		"<script async src='https://www.google-analytics.com/analytics.js'>")
 
 	// Disqus
 	b.AssertFileContent("public/index.html", "\"disqus_shortname\" + '.disqus.com/embed.js';")
--- a/tpl/tplimpl/embedded/templates.autogen.go
+++ b/tpl/tplimpl/embedded/templates.autogen.go
@@ -163,7 +163,7 @@ if (!doNotTrack) {
 	ga('send', 'pageview');
 }
 </script>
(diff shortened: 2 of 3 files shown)


onedrawingperday:
Well... the last build failed because I edited `embedded_templates_test.go` and `autogen.go` directly through GitHub and they're not `gofmt'ed`. (Can't clone the PR branch right now and `gofmt` locally.) Maybe @coliff can fix this.

About the HTTPS loading this is an old discussion see here: 
https://stackoverflow.com/questions/20516584/is-loading-scripts-or-other-resources-via-https-on-a-http-page-problematic#31100809

Also about the schemaless URLs that Hugo currently uses this is the gist of the argument against them:

> If you've been referencing cross-origin resources with the protocol-relative URL scheme, it's time to stop. Especially if you're doing so from an insecure origin. You're likely missing out on the performance benefits of HTTP/2, and maybe missing out on Brotli compressed content. If CDNs decide to stop using HTTP altogether, your requests will be burdened by higher latency as redirects to secure resources occur.

[Link](https://jeremywagner.me/blog/stop-using-the-protocol-relative-url/)

Bottom line browsers complain about mixed content when loading HTTP resources on a HTTPS` page.
The other way around (loading HTTPS resources on a HTTP page) does not cause browser warnings.

Also these days everyone is moving to HTTPS. So this PR makes sense.

bep:
>Also these days everyone is moving to HTTPS. So this PR makes sense.

With a site running HTTPS, does this PR have any relevance?

jloh:
The domain `www.google-analytics.com` sends a HSTS header along with any requests and is on the [preload list](https://hstspreload.org/?domain=www.google-analytics.com), so any requests are going to either be 307'd (internal browser redirect) or redirected to https after the first non-HTTPs request anyway. It seems like this change is going to save some people the additional request in the worst case scenario so I'd say this change makes sense (once tests pass)!

```
$ curl -vso/dev/null www.google-analytics.com/analytics.js
*   Trying 216.58.200.110...
* Connected to www.google-analytics.com (216.58.200.110) port 80 (#0)
> GET /analytics.js HTTP/1.1
> Host: www.google-analytics.com
> User-Agent: curl/7.43.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Strict-Transport-Security: max-age=10886400; includeSubDomains; preload
< Timing-Allow-Origin: *
< Date: Wed, 06 Jun 2018 23:22:11 GMT
< Expires: Thu, 07 Jun 2018 01:22:11 GMT
< Last-Modified: Fri, 18 May 2018 01:10:24 GMT
< X-Content-Type-Options: nosniff
< Content-Type: text/javascript
< Vary: Accept-Encoding
< Server: Golfe2
< Cache-Control: public, max-age=7200
< Age: 3956
< Accept-Ranges: none
< Transfer-Encoding: chunked
<
{ [2397 bytes data]
* Connection #0 to host www.google-analytics.com left intact
```

onedrawingperday:
@coliff Ok. I fixed the build failure and the commit message. 

@jloh Thanks for posting the curl requests log.

@bep it's entirely up to you whether this PR should be merged or not. 

I did say above that people are moving to HTTPS these days but what I didn't say was that some people still haven't done this and this PR might save a bit of time.

coliff:
thanks for merging this @bep - and thanks to @onedrawingperday for fixing the the build failure, commit message and giving the reasons why protocolless links are an [anti-pattern](https://www.paulirish.com/2010/the-protocol-relative-url/).

There are some other protocolless linked scripts in Hugo which I'd like to HTTPS for improved security and performance and will open a seperate PR for those.

PR #4849, opened 2018-06-13
tpl: Support text/template/parse API change in go1.11

### LINKED ISSUE #4784: Make Hugo work with Go 1.11 (and Go 1.10)
I'm on the Go master branch, Ubuntu 16.04, Linux 4.4.0-127-generic:

```
% go version
go version devel +65c365b Wed May 23 23:51:30 2018 +0000 linux/amd64
```

The following (gotten from [getting started](https://gohugo.io/getting-started/installing#fetch-from-github)) runs as expected:

```
go get github.com/magefile/mage
go get -d github.com/gohugoio/hugo
cd ${GOPATH:-$HOME/go}/src/github.com/gohugoio/hugo
mage vendor
```

However, running install fails:

```
% mage install
# github.com/gohugoio/hugo/tpl/tplimpl
tpl/tplimpl/template_ast_transformers.go:116:18: cannot range over x.Decl (type bool)
tpl/tplimpl/template_ast_transformers.go:132:10: undefined: parse.VariableNode
tpl/tplimpl/template_ast_transformers.go:133:30: an.Ident undefined (type parse.Node has no field or method Ident)
Error: running "go install -ldflags -X github.com/gohugoio/hugo/hugolib.CommitHash=d68367c -X github.com/gohugoio/hugo/hugolib.BuildDate=2018-05-27T08:47:25-0400 github.com/gohugoio/hugo" failed with exit code 2
```

It appears hugo depends on "text/template/parse" which got an API change on [Dec 17](https://github.com/golang/go/commit/28c1ad9d35f27b3b57afff4ee78faac746a8ed0a#diff-e04f7a5dd779d944dd776f82dcf35268) which was merged to master in April:

```
To avoid confusion, rename PipeNode.Decl to PipeNode.Vars, as the
variables may not always be declared after this change.
```

Renaming `Decl` to `Vars` and `VariableNode` to `AssignNode` in "template_ast_transformers.go" allowed Hugo to install:

```
% hugo version
Hugo Static Site Generator v0.42-DEV-D68367C linux/amd64 BuildDate: 2018-05-27T09:03:03-0400
```

It seems out of the ordinary for Go to have a breaking change. It would be nice to be able to install hugo with Go master.

anthonyfok:
Thanks to @rjoleary for the investigation.

Fixes #4784

### DIFF
--- a/tpl/tplimpl/template_ast_transformers.go
+++ b/tpl/tplimpl/template_ast_transformers.go
@@ -91,54 +91,6 @@ func applyTemplateTransformers(templ *parse.Tree, lookupFn func(name string) *pa
 	return nil
 }
 
-// paramsKeysToLower is made purposely non-generic to make it not so tempting
-// to do more of these hard-to-maintain AST transformations.
-func (c *templateContext) paramsKeysToLower(n parse.Node) {
-	switch x := n.(type) {
-	case *parse.ListNode:
-		if x != nil {
-			c.paramsKeysToLowerForNodes(x.Nodes...)
-		}
-	case *parse.ActionNode:
-		c.paramsKeysToLowerForNodes(x.Pipe)
-	case *parse.IfNode:
-		c.paramsKeysToLowerForNodes(x.Pipe, x.List, x.ElseList)
-	case *parse.WithNode:
-		c.paramsKeysToLowerForNodes(x.Pipe, x.List, x.ElseList)
-	case *parse.RangeNode:
-		c.paramsKeysToLowerForNodes(x.Pipe, x.List, x.ElseList)
-	case *parse.TemplateNode:
-		subTempl := c.getIfNotVisited(x.Name)
-		if subTempl != nil {
-			c.paramsKeysToLowerForNodes(subTempl.Root)
-		}
-	case *parse.PipeNode:
-		for i, elem := range x.Decl {
-			if len(x.Cmds) > i {
-				// maps $site => .Site etc.
-				c.decl[elem.Ident[0]] = x.Cmds[i].String()
-			}
-		}
-
-		for _, cmd := range x.Cmds {
-			c.paramsKeysToLower(cmd)
-		}
-
-	case *parse.CommandNode:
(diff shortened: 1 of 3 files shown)


anthonyfok:
Good news: Yay!  It compiled!

Bad news: `go test ./...` fails at TestShortcodePreserveOrder?

Compare:
* Job #9196.5 (master): https://travis-ci.org/gohugoio/hugo/jobs/391685137
* Job #9199.5 (this PR): https://travis-ci.org/gohugoio/hugo/jobs/391820859

bep:
>  <p>v1: 0 sgo: &lt;nil&gt;|v2: 1 sgo: 0|v3: 2 sgo: 1|v4: 3 sgo: 2|v5: 4 sgo: 3|</p>

So, in the above it seems that Go 11 prints `nil` as ... "nil", which is a pretty breaking change. I will try to look into this and create an issue on their tracker.

If this is the only breaking test, you can just remove "v1: 0 sgo: " from the assert -- it is not important for this particular test.

anthonyfok:
> So, in the above it seems that Go 11 prints `nil` as ... "nil", which is a pretty breaking change. I will try to look into this and create an issue on their tracker.

Thank you so much for looking into this and filing golang/go#25875!  You are amazing as always!

> If this is the only breaking test, you can just remove "v1: 0 sgo: " from the assert -- it is not important for this particular test.

Yes, it is indeed the only breaking test, as the following one-time experiment shows:

* https://travis-ci.org/gohugoio/hugo/builds/392039772

Sidenote: The first build on Linux with go1.9.5 on Travis CI actually produced a race condition!  Unfortunately, I pressed "Rebuild" for that job, which now passes, but it seems to be no way to get at the old build log.

Anyhow, seeing how golang/go#25875 is labelled "NeedsFix" and "release-blocker" for go1.11, and how the two issues (API change vs `<nil>`) are not directly related, I think I'd better keep hugolib/shortcode_test.go as-is, and not remove `"v1: 0 sgo: "`.

Issue #4865, opened 2018-06-21
Relaxing Paginate argument


vdanjean:
As discussed in https://discourse.gohugo.io/t/paginate-from-slice-of-page/12524 , Paginate currently do not accept a (plain) slice of Page, nor a Dict to used as PagesGroup. it would be a useful extension when specific ordering or grouping have to be used in templates

bep:
See my comment and the rest of the discussion in https://github.com/gohugoio/hugo/pull/4866#issuecomment-399762200

vdanjean:
I tried to implement in #5132 what have been proposed in #4866 comments. The API is indeed really easier to use.

PR #4866, opened 2018-06-21
Relaxing Paginate argument

vdanjean:
Hi,

This is a proof-of-concept for issue #4865 
This is my very first code in go language. Feel free to fully rewrite it if needed.
No tests are provided as I do not know how to put them in the sources, but it allows me to correctly build my site with a template that manually reorganize and group the various pages.

Regards,
Vincent

### DIFF
--- a/hugolib/pagination.go
+++ b/hugolib/pagination.go
@@ -391,7 +391,7 @@ func resolvePagerSize(cfg config.Provider, options ...interface{}) (int, error) 
 }
 
 func groupsFromDict(dict map[string]interface{}) (PagesGroup, error) {
-	var keys []string;
+	var keys []string
 
 	iikeys, haskey := dict["SortedKeys"]
 	if haskey {
--- a/hugolib/pagination.go
+++ b/hugolib/pagination.go
@@ -424,6 +424,8 @@ func toPages(seq interface{}) (Pages, error) {
 		return seq.(Pages), nil
 	case *Pages:
 		return *(seq.(*Pages)), nil
+	case []*Page:
+		return Pages(seq.([]*Page)), nil
 	case WeightedPages:
 		return (seq.(WeightedPages)).Pages(), nil
 	case PageGroup:
--- a/hugolib/pagination.go
+++ b/hugolib/pagination.go
@@ -19,6 +19,7 @@ import (
 	"html/template"
 	"math"
 	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/gohugoio/hugo/config"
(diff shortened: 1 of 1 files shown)


bep:
* This `SortedKeys` dict, is that a construct of yours or is it something we actually contstructs in Hugo?
* `[]*Page` is that the union case? I was guessing it was `[]interface{}`

vdanjean:
The 'SortedKeys' is just a special key in the Dict that I choose. Here is a part of my template:

                    {{ $s.Set "default" (where $pagescol "Title" "__") }}
                    {{ $s.Set "cours" (where $pagescol "Title" "__") }}
                    {{ range .Data.Pages }}
                    // note: the code/page.html partial fill the provided .Scratch after analysing several params of the provided Page, in particular, it sets .Scratch.Get "pinfo"
                    {{   partial "code/page.html" (dict "Scratch" .Scratch "VarName" "pinfo" "Page" .) }}
                    {{   $type := (.Scratch.Get "pinfo").type }}
                    {{   $s.Set $type (union ($s.Get $type) (where $pagescol "Title" .Title)) }}
                    {{ end }}
                    // note: "toto" is just here to test with a non-existant group of pages
                    {{ $s.Set "pages2" (dict "SortedKeys" (slice "cours" "default" "toto") "cours" (sort ($s.Get "cours") "Params.  rating") "default" ($s.Get "default")) }}
                        {{ $paginator := .Page.Paginate (($s.Get "pages2")) }}
                          {{ range $paginator.PageGroups }}
                          {{ markdownify (printf "## %s" .Key) }}
                          {{ range .Pages }}
                          {{   partial "lists/partials/blogentry.html" (dict "Page" .) }}
                          {{ end }}
                          {{ end }}

vdanjean:
The '[]*Page' case in the first patch is what allows one to give a slice of Page to Paginate.

It is required for ' $paginator := .Paginate (sort .Data.Pages "Params.rating") ' to succeeds.
I'm too new to the go language to understand why the 'Pages' case is not enought to catch this situation.

vdanjean:
About 'SortedKeys', perhaps 'SortedGroupsNames' would be better.

And you might find more convenient to expose (with a new hugo construct) the transformation of a Dict into a PagesGroup. For now, with my patch, it is implicit when providing the Dict to Paginate, but perhaps it can be useful in other situations. But I did not look for how to do that (ie adding a new hugo template keyword)

bep:
>About 'SortedKeys', perhaps 'SortedGroupsNames' would be better.

I really like this PR. The dict name is a little magic, but I see no way around it. But we should agree on a name.

We could consider creating a "overloaded dict" func with this special meaning: Creating page groups. `pageGroups` ... or something.

bep:
Also, I have disabled the failing asciidoctor in the Windows build for now, so if you rebase against master, this build should also go green.

regisphilibert:
So, I've read the discourse thread and now this issue in order to understand what this PR does.
What I understand is that the initial intention was to be able to use paginate on a user constructed slice/array of page objects, rather than the conventional output of a page collection (that being the result of a `were` or plain old `.Pages`).

Am I understanding the purpose of this PR correctly?

> We could consider creating a "overloaded dict" func with this special meaning: Creating page groups. pageGroups ... or something.

Now I'm lost, what do you mean by an "overloaded dict"?

bep:
>Am I understanding the purpose of this PR correctly?

So, the intention of the pagination implementation was to be really flexible; i.e you should be able to paginate any **slice of Page** ... But with Go's type system and the myriads of funcs and methods working on pages, the current implementation does not handle everything ...

To understand the main part of this PR talks about paginating ad hoc **page groups**. Page groups are built-into Hugo via the `.GroupByDate` and similar functions. These are used to create "archive type" of list pages grouped by month etc. These can be paginated.

So, this pr uses the `dict` func to create what I would call a ad-hoc based group of pages. This is a really good idea and very useful, but the current syntax is a little magical.

Maybe something like this would be better:

```
{{ $cool := where .Site.RegularPages "Params.cool" true }}
{{ $blue := where .Site.RegularPages "Params.blue" true }}

{{ $groups :=  slice (pageGroup "cool"  $cool ) (pageGroup "blue"  $blue ) }}

{{ range (.Paginate $groups).PageGroups  }}
```

Doing it this way would also create the proper object, making the paginator change in this PR simpler.

The example above with the current PR would look like this:

```
{{ $cool := where .Site.RegularPages "Params.cool" true }}
{{ $blue := where .Site.RegularPages "Params.blue" true }}

{{ $groups :=  dict "SortedKeys" (slice (pageGroup "cool"  $cool ) (pageGroup "blue"  $blue )) }}

{{ range (.Paginate $groups).PageGroups  }}
```

bep:
@vdanjean I'm going to close this and follow this up in #4865 -- your implementation looks fine, but I think we need to go an extra mile on this before we commit to some API.

PR #4877, opened 2018-06-25
Improve new theme generator

openscript:
This pull request implements two little improvements to the theme generator. As discussed [here](https://discourse.gohugo.io/t/incomplete-generation-of-a-new-theme/12577/4) the generation of new themes seems to be incomplete to me.

After the generation of a new theme a `head.html` needs to be added, because the `baseof.html` which is generated references this partial. If the partial is missing an error is thrown. This might cause confusion for new users.

Also a doctype is useful in the `baseof.html` template, which was added with this pull request as well.

### DIFF
--- a/commands/new_theme.go
+++ b/commands/new_theme.go
@@ -97,6 +97,7 @@ func (n *newThemeCmd) newTheme(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


anthonyfok:
Thank you @openscript!
Yes, I noticed the same problems myself after the v0.42.1 release, but haven't gotten around fixing it yet, so I am glad you stepped in and made the improvement!  Much appreciated!

PR #4901, opened 2018-07-04
Add flag to allow modifying task list checkboxes

### LINKED ISSUE #4899: Modifiable TODO lists in Blackfriday
Currently, the task list item generates

```
<input type="checkbox" checked disabled class="task-list-item">
```

However, I'd like to have a task list that my users can check off (e.g. bring up a page with a flight packing list and check things off as they go). Could you please add a flag which removes the disabled flag?

psa:
Adds a new Blackfriday flag "TaskListsEditable", which creates task list
checkboxes without the "disabled" flag, allowing those viewing the task
list to check and uncheck boxes in their view.

Fixes #4899


bep:
See my comment in #4899

Issue #3811, opened 2017-08-17
File watcher does not respect ignoreFiles setting


yihui:
I'm using Hugo 0.26. Here is my list of `ignoreFiles` in config.toml:

```toml
ignoreFiles = ["\\.Rmd$", "\\.Rmarkdown$", "_files$", "_cache$"]
```

When I edit a `.Rmarkdown` file, `hugo server` still rebuilds the site:

```
Change detected, rebuilding site
2017-08-17 12:02 -0500
Source changed "..../content/post/2017-08-09-test-rmarkdown.Rmarkdown": WRITE
Built site for language en:
0 draft content
0 future content
0 expired content
6 regular pages created
24 other pages created
1 non-page files copied
0 paginator pages created
5 tags created
3 categories created
total in 12 ms
```

It seems to be a regression from #1189.

davidgomes:
I can also confirm this is still happening. Hugo Serve, on Hugo 0.26, ignores the ignoreFiles options.

robsonsobral:
This still happens. Tested using:

```yaml
ignoreFiles: [ "\\.tmp$", "\\.tmp\\-browserify" ]
```

moorereason:
The `ignoreFiles` setting instructs Hugo to skip building content files during the "build" stage.  The file watcher used to detect content changes does not currently use the `ignoreFiles` setting.

robsonsobral:
In the case of browserify, this is annoying because the site reloads on the tmp file change, but not after the assets is compiled.

Thank you, @moorereason .

robsonsobral:
The issue remains:

```
Syncing assets/scripts/index.js.tmp-browserify-42993960592106339469 to /
```

paulswanson:
I'm having this problem, I think. Every time a make an edit in Vim a swap file is updated and then Hugo (running dev mode) reloads the page in the browser. I've added ignore statements but they appear to not work. The page should obviously only load when I save the file in question.

yihui:
I don't think this issue has been fixed yet.

mohkale:
Can confirm I'm still having the same issue 😢.

mohkale:
[...]

PR #4909, opened 2018-07-06
Honor ignoreFiles in the FS watcher

johnewart:
I have been using blogdown and whenever temporary files are generated or Rmd files are updated Hugo will rebuild numerous times (and ping the browser to live-reload) so I added functionality to filter out files matching any of the `ignoreFiles` regexes from the list of content files to determine if a rebuild should occur. 

It felt like assets shouldn't have a filter applied (though maybe they should and this filter should apply to both types... easy enough to change) so it only applies to content in the partitioned list; if this should be updated to filter both types let me know. 

Sorry the diff is a bit visually jarring -- it wraps a large chunk of the `newWatcher` method...

Addresses #3811

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os/signal"
+	"regexp"
 	"sort"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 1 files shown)


dmgawel:
The AppVeyor build fail looks not related to the PR changes. @johnewart could you try to force a rebuild? IMO rebasing your branch over current master and push -f should do the job and make sure it's up to date.

I wait for this change to be merged :-)

mdbraber:
Would be great if this could be merged (having the issues as mentioned in #3811)

SchumacherFM:
Happy new year! There is a merge conflict which needs to solved ;-)

Issue #4908, opened 2018-07-06
Build related 0.43 tasks


bep:
* Update Goreleaser with a new set of binaries for Win, MacOs and Linux (for now)
* Finish the Docker image for the above
* Snap? I assume we leave it as is for now. @anthonyfok 
* Talk to the Brew people.

anthonyfok:
> Snap? I assume we leave it as is for now. @anthonyfok

I have just started tweaking snapcraft.yaml and experimenting with the `extended-snap` branch, which https://launchpad.net/~gohugoio/+snap/hugo-dev is now temporarily tracking and making test builds for me.  :-)

bep:
Re. snap, is it possible to deliver 2 versions? I assume there will be people wanting a "pure Go" Hugo version and does not need the SCSS/SASS.

anthonyfok:
> Re. snap, is it possible to deliver 2 versions? I assume there will be people wanting a "pure Go" Hugo version and does not need the SCSS/SASS.

I think it is possible.  Would you like two separate executables in the same snap (easier to do), or two entirely separate snaps?

Assuming the first case, how would you like to name the executables?  How about `hugo` and `hugo.unextended`?

bep:
>hugo and hugo.unextended?

... Unextended?


No, I was more thinking some "snap target" (I'm not sure what that is) ... like dev, extended.

If two binaries, then the default should be the one without change, ie. hugo and then hugo-extended.

bep:
Note that the binaries on the release page will be named "hugo" and "hugo_extended"; I think I want them to be named the same, but that will have to wait.

anthonyfok:
> No, I was more thinking some "snap target" (I'm not sure what that is) ... like dev, extended.

Snaps can be released into 4 channels, namely "stable", "candidate", "beta" and "edge".  IIRC, `hugo-dev` is just a random name I picked for the build daemon (?) that tracks the hugo master branch for the most cutting-edge builds, and have them uploaded and published directly into the "edge" channel, but the name `hugo-dev` itself doesn't actually appear anywhere in the snap itself, and snap has no knowledge of that name.

So, to install what I named `hugo-dev` on Launchpad, the end user would actually use `snap install hugo --edge`, and not `snap install hugo-dev`.

While we can create an entirely new snap called `hugo-extended`, that seems a bit too complicated to me, especially we would have to maintain an extra branch with a special snapcraft.yaml that says `name: hugo-extended` on its the first line.

[...]

PR #4916, opened 2018-07-08
WORK IN PROGRESS: Extend the GoReleaser to support SCSS

bep:
See #4908

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -1,6 +1,5 @@
 builds:
-- main: main.go
-  binary: hugo
+- binary: hugo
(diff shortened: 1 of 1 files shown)


bep:
@caarlos0 can I ask you a quick question. So, I configured multiple binary sections and I kind of expected to also get multiple archives, but both `hugo` and `hugo_extended` gets bundles into the same. Is there a way around this?

bep:
/cc @anthonyfok I'm not sure this is a big problem ... initially. Will sleep on it.

Ideally I think we should have separate archives but the same binary name (hugo) for these.

bep:
OK, I will just do a "second manual Go releaser build" when doing the release tomorrow with a different archive name for the "extended" variants, then we can think of a long term solution later.

Issue #4945, opened 2018-07-14
[Feature-request] Get the date of the next scheduled change


lacqui:
When a change is pushed through git, my site is built immediately. However, if I have something scheduled for future publication or expiry, there is no way to handle this.

I propose a method of outputting the date of the next scheduled change, in a manner that can be put into a scheduler for a future automatic rebuild.

PR #4946, opened 2018-07-14
Added ability to output next change date

lacqui:
Proposed initial solution for #4945


lacqui:
This is very basic, and only outputs the date of the next build if requested. I don't know how to properly handle a configuration, in order to output in a different format, such as may be required by scheduler commands.

PR #4954, opened 2018-07-16
Fix addkit link to account for i18n

hairmare:
I did this change through GitHub's WebUI. I'm not sure why GitHub wants to change the last line in the file but the change is inconsequential.


hairmare:
Wow, that was fast. Thanks!

Issue #4861, opened 2018-06-20
Render in Page bundles not working - was working in 0.40.2


jrisch:
I've used the Render function inside Page bundles extensively throughout the month of april without problems. After having a month long break from coding, I started up again in june, but with the newest versions of Hugo, the `hugo server` command fails with:

`Failed to render "_default/page.html": runtime error: invalid memory address or nil pointer dereference`

You can test with this test site: https://github.com/jrisch/hugo-test

Building the site or running `hugo server` with version 0.40.2 works fine, but newer versions does not.

reber0:
@jrisch I met the same problem, did you solve it?

nonumeros:
@jrisch  I agree with the non-working versions part, but it can't possibly be solely the `Render` function because if you were to have e.g.
      
     {{ define "main" }}
      {{ range .Data.Pages }}
       {{ .Render "codeexample"}}
      {{ end }}
     {{ end }}


instead of 

    {{ define "main" }}
    {{- with .Resources.Match "codeexample.md" }}
    {{- range . }}
    {{ .Render "codeexample" }}
     {{- end }}
     {{- end }}
    {{ end }}

on page.html then you'd notice it works under the latest version

jrisch:
@nonumeros You're correct that the `hugo serve` command works when using your code, but the product page renders empty. :-(

jrisch:
The newest hugo version (0.46) throws a stack trace - don't know if that is helpful:

```
Building sites … ERROR 2018/08/07 11:13:16 Failed to render "_default/page.html": programming error: no mainPageOutput for "products/myproduct/codeexample.md"
ERROR 2018/08/07 11:13:16 Stack Trace:
goroutine 57 [running]:
github.com/gohugoio/hugo/hugolib.stackTrace(0x4b0, 0x4c95113, 0x17)
	/private/tmp/hugo-20180801-41150-1lexkxf/hugo-0.46/src/github.com/gohugoio/hugo/hugolib/page.go:280 +0x76
github.com/gohugoio/hugo/hugolib.(*Site).renderForLayouts.func1(0xc4207ff8a0, 0xc42025a300)
	/private/tmp/hugo-20180801-41150-1lexkxf/hugo-0.46/src/github.com/gohugoio/hugo/hugolib/site.go:1742 +0x135
panic(0x4b12440, 0xc4204de420)
	/usr/local/Cellar/go/1.10.3/libexec/src/runtime/panic.go:502 +0x229
[...]

PR #4960, opened 2018-07-18
Fix rendering of resource pages in leaf bundles

hpoul:
When having a leave bundle with nested pages calling `.Render` currently results in an error message. I do not think this is supposed to happen. This PR provides a very simple fix. See the test for an example.

Also fixes https://github.com/gohugoio/hugo/issues/4861 and hence the referenced example repository https://github.com/jrisch/hugo-test can be taken as an example use case. (without this PR rendering hugo-test breaks with `Failed to render "_default/page.html": programming error: no mainPageOutput for "products/myproduct/codeexample.md"`)

### DIFF
--- a/hugolib/page_bundler_test.go
+++ b/hugolib/page_bundler_test.go
@@ -345,6 +345,34 @@ func TestPageBundlerSiteWitSymbolicLinksInContent(t *testing.T) {
 
 }
 
+func TestPageBundlerRenderResource(t *testing.T) {
+	cfg, fs := newTestCfg()
+
+	assert := require.New(t)
+
+	workDir := "/work"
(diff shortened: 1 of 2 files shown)


bep:
>I do not think this is supposed to happen. This PR provides a very simple fix. See the test for an example.

The error message is not good, but it was never intended that the pages inside a bundle should be `.Render`. I'm not merging this PR at this moment. This needs some thought and, if implemented, a less hardcoded implementation.

bep:
I'm closing this now. We have at least one issue to track this. We need to figure out a more general solution to this.

gpoul:
@bep the issue you're referring to is #4861 or another one?

PR #4970, opened 2018-07-19
Add test coverage for recent ref overhaul

vassudanagunta:
These are the tests I intended to write to include in and give thorough coverage to #4796 once I got word that the work I had done was approved (I didn't want to waste my time writing tests for a PR that wasn't going anywhere). But my commits from that PR suddenly got merged via #4955.

A logic change made in the latter PR, in commit 3eb313fe, introduced a flaw in the handling of absolute refs. These tests demonstrate that flaw. 

They also show some corner-case issues resulting from the co-existance of unix-path semantics and "ambiguous ref" semantics (They will necessarily conflict!). Whether or not the issues are addressed, it is good they are known. 

All the above issues are flagged in the tests with the text "ISSUE:".

### DIFF
--- a/hugolib/page_collections_test.go
+++ b/hugolib/page_collections_test.go
@@ -96,6 +96,29 @@ func BenchmarkGetPageRegular(b *testing.B) {
 	}
 }
 
+type testCase struct {
+	kind          string
+	context       *Page
+	path          []string
+	expectedTitle string
+}
+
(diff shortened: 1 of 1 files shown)


bep:
Merge in 2bac3715448e90e197ada7cc73c87f696c19def6 ... and the test passes ...

PR #4975, opened 2018-07-19
Increase refLinker test coverage

vassudanagunta:
@bep These are tests from my fork that I missed including in my PR for the ref logic changes. It may seem a little redundant with the tests for GetPage, but refLinker does more than simply look up a page based on a ref: 
- It first parses the ref to separate the file path from the fragment identifier (either one, both or neither can be supplied). The current implementation uses a URL parser to parse a file path, so we cannot assume it is correct.
- Instead of returning the Page, it returns its permalink or relative permalink 
- It appends the fragment identifier onto that (tested), possible making sure it is unique (not tested yet)

So I think these tests are worth having. Your call of course.


bep:
The tests looks good, but Travis complains about gofmt.

vassudanagunta:
@bep sorry. single extra newline. fixed.

PR #4976, opened 2018-07-20
Add go.mod and go.sum

leighmcculloch:
### What
Add `go.mod` and `go.sum`.

### Why
Hugo appears to work with `go mod` out of the box. For users with go tip, or 1.11+ once it has been released, wanting to install or dev with source, using the standard `go install` is simpler than the current set of instructions to install dep and run a couple mage commands.


bep:
I agree, and we will get this done, but as this involves some testing and QA about which dependencies are picked etc., it must be done with someone who can give more insight than "it appears to work". I have an issue to track this. Thanks, but this will have to wait a little.

leighmcculloch:
Could you share the issue that's tracking this? Before I opened the PR I searched for issues and PRs relating to it, but GitHub search returned none referencing `vgo` or `go.mod`.

kaushalmodi:
@leighmcculloch https://github.com/gohugoio/hugo/pull/4754

bep:
I can add to this that there are some unresolved issues with Mage (it is going to be fixed) and modules, so I think we are going to wait for Go 1.11 until we do the switch. Which also involves getting rid of Go Dep. But we are doing it, and I'm very exited about it for many reasons.

leighmcculloch:
@bep Does it make sense to continue to use mage once go.mod is out?

bep:
>@bep Does it make sense to continue to use mage once go.mod is out?

Yes.

PR #5013, opened 2018-07-31
Fix file paths for uncached transformed images

### LINKED ISSUE #5012: Resized images's `.Content` represents source image on first build
Just discovered an odd bug which ended up nuking a Netlify site of mine.

Repro steps:

* Resize an image with Hugo to 16x16
* Load its `.Content` into a data URI
* Embed the in an image tag

Expectation:

* A 16px wide image is embedded onto the page

Result:

* First build: The full-sized image is embedded
* Subsequent builds (warm `resources/_gen`):  A 16px wide image is embedded onto the page

## Demonstration

https://image-resize-bug.netlify.com/
https://with-resources-folder--image-resize-bug.netlify.com/

https://github.com/RJacksonm1/hugo-resize-bug/commit/c85af401662a0b64e0304d3ca5cf1e0f558d3a45

RJacksonm1:
When resizing an image the `sourceFilename` (and thus its `.Contents`) represented the original file _if_ the resized image was not already within Hugo's resource cache. If the resized file was already in the resource cache, `sourceFilename` (and `.Contents`) were set as expected.

This PR defers setting `sourceFilename` until we already have an image, whether it was loaded from cache or created anew.

Fixes #5012

### DIFF
--- a/resource/image_cache.go
+++ b/resource/image_cache.go
@@ -100,16 +100,16 @@ func (c *imageCache) getOrCreate(
 
 	if exists {
 		img = parent.clone()
-		img.relTargetDirFile.file = relTarget.file
-		img.sourceFilename = cacheFilename
-		// We have to look in the resources file system for this.
-		img.overriddenSourceFs = img.spec.BaseFs.Resources.Fs
 	} else {
 		img, err = create(cacheFilename)
 		if err != nil {
 			return nil, err
 		}
 	}
+	img.relTargetDirFile.file = relTarget.file
+	img.sourceFilename = cacheFilename
+	// We have to look in the resources file system for this.
+	img.overriddenSourceFs = img.spec.BaseFs.Resources.Fs
 
 	c.mu.Lock()
 	if img2, found := c.store[key]; found {
--- a/resource/image_test.go
+++ b/resource/image_test.go
@@ -73,6 +73,7 @@ func TestImageTransformBasic(t *testing.T) {
 	assert.NoError(err)
 	assert.True(image != resized)
 	assert.True(image.genericResource != resized.genericResource)
+	assert.True(image.sourceFilename != resized.sourceFilename)
 
 	resized0x, err := image.Resize("x200")
 	assert.NoError(err)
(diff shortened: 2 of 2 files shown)


bep:
@RJacksonm1 I suspect that the test that fails the CI build is actually a test for this particular issue, but where the assertion is wrong.

RJacksonm1:
@bep Ah cheers. I'm having trouble getting my tests running at the minute (`use of internal package not allowed`, despite having my code in `/Users/rjackson/Go/src/github.com/gohugoio/hugo` 🤷‍♂️ ).

Once I get the tests working, I'll have a crack at making sure this issue is properly covered.

bep:
>use of internal package not allowed

Hmm... I've been in that boat for other projects. A vendoring issue? If you use `mage vendor` from the `hugo` folder it should work.

RJacksonm1:
@bep No luck with that alone. Looks like I've solved it by nuking everything and re-downloading the source as per the [contributing guide](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#fetching-the-sources-from-github). I was previously following the README's instructions (though I'd imagine the fault is more likely on my side regardless)

I now have the failing `TestImageTransformBasic` test 🙌

RJacksonm1:
Alright, tests updated.

I've rebased the change so my new test for the correct `sourceFilename` behaviour runs before updating the code, so we can compare before/after the fix:

```
git rebase --exec 'go test github.com/gohugoio/hugo/resource -run "^TestImageTransformBasic$" || true' origin/master

Executing: go test github.com/gohugoio/hugo/resource -run "^TestImageTransformBasic$" || true
     sunset.jpg	en	Filename: content/sunset.jpg	Base: content
--- FAIL: TestImageTransformBasic (0.03s)
        Error Trace:    image_test.go:76
	Error:      	Should be true
	Test:       	TestImageTransformBasic
FAIL
FAIL	github.com/gohugoio/hugo/resource	0.059s

Executing: go test github.com/gohugoio/hugo/resource -run "^TestImageTransformBasic$" || true
ok  	github.com/gohugoio/hugo/resource	(cached)
```

As for the test that broke earlier, that was actually a fault in the original test data. I fixed that and squashed it into the fix commit. My notes from the commit message:

>Fix image resize-then-fit test
>
>This test was written based on erroneous behaviour. We resize the image to 300x200px, and are now trying to fit it within a 50px square. The longest edge is 300 pixels, so we need to divide it by 6 (300 / 50 == 6). And then scale the shortest edge with the same proportion (200 / 6 == 33.33).
>
>The original test was transforming the original source image, hence the previous values:
>
>900 x 562
>900 / 50 == 18
>562 / 18 == 31.22

bep:
Thanks a lot for this fix, I really like ready-to-pull PRs.

PR #5036, opened 2018-08-05
Add fileStat to tpl/os/os

satotake:
#### Summary
Add `fileStat` method to `os`
`fileStat PATH` returns `os.FileInfo`

#### Use Case
Method `readFile` has the size limitation (1MB).
To avoid throwing the error and make a workaround, the size must be got before executing `readFile`.
Under the current condition, this can be done via `[]os.FileInfo` as a result of `readDir`.
However, it is a little complicated because you have to handle an array even though you need just a single `os.FileInfo`.
`fileStat` is useful in such situations.

### DIFF
--- a/tpl/os/init.go
+++ b/tpl/os/init.go
@@ -55,7 +55,7 @@ func init() {
 			},
 		)
 
-		ns.AddMethodMapping(ctx.FileStat,
+		ns.AddMethodMapping(ctx.Stat,
 			[]string{"fileStat"},
 			[][2]string{
 				{`{{ (fileStat "files/README.txt").Size }}`, `11`},
--- a/tpl/os/init.go
+++ b/tpl/os/init.go
@@ -58,7 +58,7 @@ func init() {
 		ns.AddMethodMapping(ctx.FileStat,
 			[]string{"fileStat"},
 			[][2]string{
-				{`{{ (fileExists "files/README.txt").Size }}`, `11`},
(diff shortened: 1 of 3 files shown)


tpl/os/os.go
```diff
@@ -130,3 +130,22 @@ func (ns *Namespace) FileExists(i interface{}) (bool, error) {
 
 	return status, nil
 }
+
+// FileStat Stat returns the os.FileInfo structure describing file.
+func (ns *Namespace) FileStat(i interface{}) (_os.FileInfo, error) {
```
bep: I think we should name it simply `os.Stat`. I think we (at least has tried) to name the other funcs that more or less wraps the std lib variant with the same name as the original. In this case I think that makes sense.

satotake: I see.
I have renamed the name.

tpl/os/init.go
```diff
@@ -55,6 +55,13 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.FileStat,
+			[]string{"fileStat"},
```
satotake: Is it better to rename here?, @bep
In my opinion, `fileStat` will be better than `Stat` because  we cannot get what kind of stat. if `Stat` is used in Hugo template.

bep: I think we should remove it. We don't need an alias for this, as `os.Stat` is already short enough.

PR #5092, opened 2018-08-17
Changed postcss to check for local installation under node_modules/.bin

### LINKED ISSUE #5091: Local installation of postcss-cli doesn't work on Windows
# Problem

Using `postcss-cli` from a local `node_modules` folder within a project currently doesn't work on Windows. The reason is that `postcss` found under `node_modules/postcss-cli/bin` is not executable on Windows.

https://github.com/gohugoio/hugo/blob/6746826689faca5ca09f46725501dceff1a0ebbc/resource/postcss/postcss.go#L101

# Proposed fix

I propose to always try to find the local installation under `node_modules/.bin`. This should work on other environments as well.

P.S. I'm using `Hugo v0.47`

ericselin:
Fixes #5091

### DIFF
--- a/resource/postcss/postcss.go
+++ b/resource/postcss/postcss.go
@@ -98,7 +98,7 @@ func (t *postcssTransformation) Key() resource.ResourceTransformationKey {
 // npm install -g autoprefixer
 func (t *postcssTransformation) Transform(ctx *resource.ResourceTransformationCtx) error {
 
-	const localPostCSSPath = "node_modules/postcss-cli/bin/"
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
As this value is now generic, it may work better being moved into a more generic location. I'm not familiar with Go's shell execution framework, but if it's possible to mutate the `$PATH` for all calls, that could be an interesting implementation too.

bep:
Merged manually, thanks for this and sorry for the delay.

PR #5102, opened 2018-08-20
fix: build on armv7

### LINKED ISSUE #5101: Check rasberry build
See https://discourse.gohugo.io/t/error-installing-47-on-raspbian/13663

>dpkg: error processing archive ./hugo_0.47_Linux-ARM.deb (--install):
 package architecture (armel) does not match system (armhf)

In Hugo 0.47 we updated Goreleaser to 0.79 (from something very old, but the config should be more or less the same, see: https://github.com/gohugoio/hugo/commit/20148345aa667bfbf4d76b6fa8349f18f80028c2

Does the above ring any bells, @caarlos0 ?

caarlos0:
closes #5101

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -18,6 +18,8 @@ build:
     - 386
     - arm
     - arm64
+  goarm:
+    - 7
   ignore:
     - goos: openbsd
       goarch: arm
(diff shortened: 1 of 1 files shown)


caarlos0:
Why is openbsd arm 6 ignored? Should I ignore arm 7?

bep:
>Why is openbsd arm 6 ignored?

Because it failed at one point in time. But the Go packages in this department have moved since then, I guess.

PR #5114, opened 2018-08-25
resource: Add instruction to install PostCSS when missing

### LINKED ISSUE #5111: Differentiate error messages between "PostCSS not found" and "libSass not compiled in"
As of 0.47.1, according to Transform() in resource/postcss/postcss.go, when Hugo cannot find the executable "postcss" in "node_modules/postcss-cli/bin/" (fetched via `npm install postcss-cli` without the `-g` flag) or in the PATH, it would show errors.FeatureNotAvailableErr, i.e.:

    this feature is not available in your current Hugo version

resulting in, for example:

> ERROR 2018/08/24 00:46:39 error: failed to transform resource: POSTCSS: failed to transform "styles/main.css" (text/css): this feature is not available in your current Hugo version

which is slightly misleading for the end user because the correct solution is not to look for another Hugo version (extended or not), but to have PostCSS (postcss-cli) installed properly so that Hugo can find it.

I hesitate to change the error message directly because errors.FeatureNotAvailableErr is also used, accurately, when libSass is not compiled in.  I thought of creating another error message altogether, but hesitated when I realized that in resource/transform.go errors.FeatureNotAvailableErr is used as a check.

So, yes, I thought I should ask you (bep) first how to best approach this issue.  :-)

Admittedly, a very minor issue, and non-urgent.  Thanks in advance for looking into this!

P.S.  I came across this when I finally tested my recent changes to snapcraft.yaml going into 0.47.1, and discovered that my fix didn't work.  (Oops, sorry!)  First of all, I inadvertently deleted the "usr/bin/nodejs" binary in the clean-up, and secondly, it was a rather old v4 version that got installed as "usr/bin/nodejs" instead of "usr/bin/node".  A fix for that will be ready soon, by using the "nodejs" plugin in Snap.

anthonyfok:
Fixes #5111 

@ bep  Thank you for your great advice!

### DIFF
--- a/resource/transform.go
+++ b/resource/transform.go
@@ -391,7 +391,11 @@ func (r *transformedResource) transform(setContent bool) (err error) {
 				// If a prepared bundle for this transformation chain is available, use that.
 				f := r.tryTransformedFileCache(key)
 				if f == nil {
-					return fmt.Errorf("%s: failed to transform %q (%s): %s", strings.ToUpper(tr.transformation.Key().name), tctx.InPath, tctx.InMediaType.Type(), err)
+					errMsg := err.Error()
+					if tr.transformation.Key().name == "postcss" {
+						errMsg = "PostCSS not found; install with \"npm install postcss-cli\". See https://gohugo.io/hugo-pipes/postcss/"
+					}
+					return fmt.Errorf("%s: failed to transform %q (%s): %s", strings.ToUpper(tr.transformation.Key().name), tctx.InPath, tctx.InMediaType.Type(), errMsg)
 				}
 				transformedContentr = f
 				defer f.Close()


resource/transform.go
```diff
@@ -391,7 +391,11 @@ func (r *transformedResource) transform(setContent bool) (err error) {
 				// If a prepared bundle for this transformation chain is available, use that.
 				f := r.tryTransformedFileCache(key)
 				if f == nil {
-					return fmt.Errorf("%s: failed to transform %q (%s): %s", strings.ToUpper(tr.transformation.Key().name), tctx.InPath, tctx.InMediaType.Type(), err)
+					errMsg := err.Error()
+					if tr.transformation.Key().name == "postcss" {
```
bep: You should also check that this is a FeatureNotAvailableErr -- you may get other errors where this error message may confuse more than it helps

anthonyfok: Thank you for reviewing!

> You should also check that this is a FeatureNotAvailableErr -- you may get other errors where this error message may confuse more than it helps

If I understand correctly, this check for FeatureNotAvailableErr is already done a few lines back.  My changes are with an `if err == errors.FeatureNotAvailableErr { }` block which already exists in the code, see https://github.com/gohugoio/hugo/blob/94d6d6780fac78e9ed5ed58ecdb9821ad8f0f27c/resource/transform.go#L387-L405

Seeing that `err` is not redefined within this `if` block, I suppose I do not need to check it again?  :-)

Issue #4952, opened 2018-07-15
Use local installation of Postcss-cli


carlmjohnson:
To use the new postCSS resource feature in Hugo 0.43, you must have postcss-cli installed on your path. However, ideally, it should be enough to install it into `node_modules/.bin/`. That way your project can document all its requirements in a project specific `package.json` and not need a separate global install. It also eases dealing with multiple projects with different versions of postcss-cli in use.

numbfall:
Local install of postcss-cli is still not working. I get the following error unless I install postcss-cli globally.

```
Building sites … ERROR 2018/08/05 17:46:48 error: failed to transform resource: POSTCSS: failed to transform "css/tachyons.css" (text/css): this feature is not available in your current Hugo version
```

I have postcss in my `node_modules/.bin/` and `node_modules/postcss-cli/bin/`

This is my html

```
{{ $style := resources.Get "css/tachyons.css" | resources.PostCSS }}
```

You need to reopen this issue.

tech4him1:
@numbfall Please be sure you are using Hugo v0.45 or newer -- that was when the fix was added.

TheBaronHimself:
I think it would be useful to have this functionality for themes, too. I.e. Hugo using postcss-cli and any plugins contained in `themes/sometheme/node_modules/.bin`. This way a theme's dependencies could be installed more easily. Or am I being silly?

TheBaronHimself:
@onedrawingperday 

That makes lots of sense, actually. Thanks.

inwardmovement:
Local `postcss-cli` seems to work [on Mac](https://discourse.gohugo.io/t/why-postcss-has-to-be-installed-globally/19680), it should work on Windows too. Can we reopen?

PR #5121, opened 2018-08-26
Remove incorrect installation instructions

RealOrangeOne:
Installation needs to be done with `-g` to correctly install. I voted to remove instructions entirely and just reference the docs to save duplication.

### DIFF
--- a/resource/transform.go
+++ b/resource/transform.go
@@ -393,7 +393,7 @@ func (r *transformedResource) transform(setContent bool) (err error) {
 				if f == nil {
 					errMsg := err.Error()
 					if tr.transformation.Key().name == "postcss" {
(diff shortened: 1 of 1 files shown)


StephenBrown2:
Local installation works (and is checked first) as of 0.45: https://github.com/gohugoio/hugo/issues/4952

But referencing the docs is good!

bep:
Local installation works fine and is probably the recommended way.

```bash
 npm install -D --save postcss-cli
```

numbfall:
> 
> 
> Local installation works fine and is probably the recommended way.
> 
> ```shell
>  npm install -D --save postcss-cli
> ```

I am shy to report this again, you did close the issue previously. I am on version 0.49 now. postcss-cli still does not work unless I install it globally on windows 10. 

I have tested it on different windows 10 machines with same results. css simply does not build on change. and removing the resources folder giver a 'missing postcss-cli' error. 

should I tell hugo somehow to use the local modules? what am I missing?

PR #5132, opened 2018-08-28
New proposal for relaxing Paginate argument

vdanjean:
This proposal is discussed in issue #4865 (with numerous comments in closed PR #4866)

### DIFF
--- a/hugolib/pageGroup.go
+++ b/hugolib/pageGroup.go
@@ -296,3 +296,10 @@ func (p Pages) GroupByParamDate(key string, format string, order ...string) (Pag
 	}
 	return p.groupByDateField(sorter, formatter, order...)
 }
+
+// Group creates a PageGroup from a key and a Pages object
+func (p *Page) Group(key interface{}, pages Pages) (PageGroup, error) {
+	pageGroup := PageGroup{Key: key, Pages: pages}
+
+	return pageGroup, nil
+}
--- a/hugolib/pagination.go
+++ b/hugolib/pagination.go
@@ -399,7 +399,11 @@ func paginatePages(td targetPathDescriptor, seq interface{}, pagerSize int) (pag
 
 	var paginator *paginator
 
-	if groups, ok := seq.(PagesGroup); ok {
+	groups, err := toPagesGroup(seq)
+	if err != nil {
+		return nil, err
+	}
+	if groups != nil {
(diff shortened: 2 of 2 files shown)


kaushalmodi:
My only comment would be to include doc updates in this PR too, if there are plans to merge it.

hugolib/pagination.go
```diff
@@ -414,6 +418,39 @@ func paginatePages(td targetPathDescriptor, seq interface{}, pagerSize int) (pag
...
+		return seq.(PagesGroup), nil
+	case []PageGroup:
+		return PagesGroup(seq.([]PageGroup)), nil
+	case []interface{}:
+		list := seq.([]interface{})
+		l := len(list)
+		if l == 0 {
+			break
+		}
+		switch list[0].(type) {
```
bep: `switch v := list[0].(type)` same comment as above -- avoids casting inside the switches.

hugolib/pagination.go
```diff
@@ -414,6 +418,39 @@ func paginatePages(td targetPathDescriptor, seq interface{}, pagerSize int) (pag
...
+		case PageGroup:
+			pagesGroup := make([]PageGroup, len(list))
```
bep: ` make(PagesGroup, len(list))` is better.

hugolib/pageGroup.go
```diff
@@ -296,3 +296,10 @@ func (p Pages) GroupByParamDate(key string, format string, order ...string) (Pag
 	}
 	return p.groupByDateField(sorter, formatter, order...)
 }
+
+// Group creates a PageGroup from a key and a Pages object
+func (p *Page) Group(key interface{}, pages interface{}) (PageGroup, error) {
```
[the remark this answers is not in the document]
vdanjean: Done
But I do not really understand when using interface{} or not (when the argument can have various real types?)

hugolib/pagination.go
vdanjean: Not done as the value of the first element in the array is not important. Perhaps the check is not the good one.
There is a loop just after, iterating over all elements, to cast them one by one. The initial switch is here to handle type errors but perhaps it should be done for all elements.

If there is a better way to do it in go, tell me or fix it directly.

hugolib/pagination.go
vdanjean: Done

PR #5137, opened 2018-08-30
Fixing link for article which moved to a new domain for hosting hugo on firebase

TheAndruu:
The article remains valid and accurate, just moved to a new domain.


TheAndruu:
Ok, thanks

PR #5159, opened 2018-09-03
Add CODE_OF_CONDUCT.md

coliff:
It's nice to have this for projects.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable behavior and are expected to take appropriate and fair corrective action in response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, or to ban temporarily or permanently any contributor for other behaviors that they deem inappropriate, threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces when an individual is representing the project or its community. Examples of representing a project or community include using an official project e-mail address, posting via an official social media account, or acting as an appointed representative at an online or offline event. Representation of a project may be further defined and clarified by project maintainers.
+
+## Enforcement
+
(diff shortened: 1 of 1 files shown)


onedrawingperday:
I saw this PR when it was submitted, I also think that it is not for Hugo yet, maybe if the project has financial backing or way more people behind it.

Who is going to enforce all that? I know that I'm not.

Also I noticed that @bep 's personal email is posted under as the Enforcement Section of this PR that doesn't seem right. 

If this is agreed upo a more generic email should be created and use for something like this.

coliff:
There's an alternative Code of Conduct template for larger projects:
https://github.com/gohugoio/hugo/community/code-of-conduct/new?template=citizen-code-of-conduct
You could choose to remove the email address if desired.

onedrawingperday:
I see @coliff 

I have opened a Topic in the Forum for this PR so that more people can vote.

https://discourse.gohugo.io/t/vote-for-a-new-code-of-conduct-for-the-hugo-project/14207

RickCogley:
I voted no. The CoC talks about not violating privacy yet has @bep's email... _smh_ because it's not like he is busy or anything.

There are a lot of reasons I think this is not needed. A couple: 

* I dislike the idea that people can't self-govern enough to be decent to one another, especially since this is a volunteer project. 
* It formalizes enforcement, yet who will arbitrate that? 
* It's entirely too PC for me; so much so I'm getting a rash thinking about it. Each person should choose _not_ to be offended by things, in my opinion. Being offended, and the offender's _intent_ to cause offense are really two very different things. 
* Why would the maintainers _not_ discriminate based on experience, with regard to who they give commit access to. That's just asking for trouble. 

So far there have been _very_ few issues. This is a great project as it is, and doesn't need a "nanny". 

In fact the CoC could just be this: 

> Don't be a dick. Don't waste other peoples' time. Assume good intent.

onedrawingperday:
>It's entirely too PC for me; so much so I'm getting a rash thinking about it.

It's _too_ politically correct in my eyes also.

>So far there have been very few issues.

True that.

However this is still up for a vote and I promoted it in the Forum to get more people to vote.

If it's incorporated maybe a volunteer could join the Hugo team for handling the Code of Conduct issues and maybe also keeping an eye on GitHub SPAM.

In the Forum we have great filters SPAM but on GitHub none really. Just yesterday I had to report to GitHub and close 10 issues on the main Hugo Bugs tracker.

As things are in the near future I may not be able to _even_ do the things I'm already doing around here due to lack of time.

So I will certainly not have the time to moderate content posted through the Hugo Project's channels for issues that pertain to the proposed Code of Conduct.

holehan:
I voted yes because it's indeed "nice to have this for projects". A Code of Conduct shows a project cares.

bep:
I voted no. I know they have a similar on the Go project, which with its size it may make sense, but even there I have seen some odd enforcement.

The 2 reasons why I don't think it fits in here is 1) Most of this should go without saying and 2) I suspect it will add work/noise. The Hugo community is a small, but great one; with some formalized rules there will be pedantic internet trolls that will claim that "some comment on the forum is in violation with #34 in the CoC" ... And I would get that mail; I would possibly not agree with that person, but I would then have to do something about it, but what?

I think this is mainly an issue about the forum, and there we have established some ground rules that I think work mostly fine.

martignoni:
I voted no, mainly because of the implied workload for project maintainers. If needed, this can be added later.

kaushalmodi:
I never get the point of *CoC*. It's as if the folks about to do misbehaviour are going to read the whole CoC, and then not do that misbehaviour on reading it. 

And about controlling such trolls, I think we already doing that on the Discourse forum. 

As @RickCogley said:

> Don't be a dick. Don't waste other peoples' time. Assume good intent.

This applies in general to anything. It's common sense. 

I think this CoC thing is just another gimmick.

regisphilibert:
> Don't be a dick. Don't waste other peoples' time. Assume good intent.

Now that's a _DoC_ I'd vote for.

moorereason:
I vote No.  @bep & @RickCogley nailed it.

kevingimbel:
I voted No. 

I've been reading almost every issue report that comes in over the last year and most of the time things work out quite well. A Code of Conduct wouldn't benefit the general vibe this community has - which is a good vibe!
As others said, I too think Hugo is too small "staff"-wise to enforce the CoC or have a dedicated person who deals with it.

jgreely:
Solution in search of a problem.

bep:
I will close this. The reason I opened this poll is that it would look odd if I closed this without any discussion.

felicianotech:
Poll closed so quickly. 😐

Community is something I spend  a lot  of time thinking about. Code of Conducts (CoC) are very important for communities. "Don't be a dick" makes sense, and as some mentioned, common sense. The problem is, there's an age old saying, common sense isn't that common.

People from various backgrounds with regard to race, finance, language, and other cultural differences may have a different idea of common sense.

CoC should be endorsed and enforced by some sort of governance, which Hugo doesn't currently have. So I agree this is premature.

Excitedly, you all are doing a fantastic job with Hugo and it's growing fast. Don't expect to be so small for long. 😊

PR #5176, opened 2018-09-06
WIP: Fix all golint issues

moorereason:
I'm planning to do one commit per root package.  If a more substantial change is required (like renaming an important item), I'll likely do a separate commit for that.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -62,6 +62,8 @@ type Response struct {
 	Cmd *cobra.Command
 }
 
+// IsUserError returns true is the Response error is a user error rather than a
+// system error.
 func (r Response) IsUserError() bool {
 	return r.Err != nil && isUserError(r.Err)
 }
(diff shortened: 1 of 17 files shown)


output/outputFormat.go
```diff
@@ -330,10 +331,12 @@ func decode(mediaTypes media.Types, input, output interface{}) error {
 	return decoder.Decode(input)
 }
 
+// BaseFilename returns the base filename of formats.
```
bep: I much prefer a lint warning before comments like the above.

moorereason: Agreed.  This is super tedious, and I started getting tired and hasty.  I'll go back through these and double-check my work.

moorereason:
Okay, that's it for this PR.  There's many more issues in `hugolib` and `resource`, but there's possibly too much in this PR already.  And my brain hurts.

Please review and make suggestions for improvements.  If we're going to merge this, I'd like to do it soon before merge conflicts start creeping in.

PR #5178, opened 2018-09-06
Make JSON minification more generic

zinefer:
[JSON-LD](https://json-ld.org/) is is a standardized format for providing information about a page and classifying the page content. In my use case I am using it for [Google SEO](https://developers.google.com/search/docs/guides/intro-structured-data).

In this pull request I've changed the minifier `mediaTypes` regex to allow for `application/ld+json`.

I wasn't able to find any tests for this section of code and I'm still pretty new to go so I'm unsure of how to add one.


bep:
1) I suspect you meant to attach this to the JSON minifier?
2) You can create application/ld+json as a custom mime type in your project and it will be minified. We may add a regexp to catch more JSON variants out of the box, but I think that needs to be less ad-hoc and with tests.

PR #5219, opened 2018-09-15
Simplify travis-ci build

wfscheper:
This change avoids the need to copy the source out of the GOPATH by
applying GO111MODULE=on as an environment variable. Additionally, avoids
issues with parallel module downloads (see golang/go#26794 and
golang/go#27372) by running 'go mod download' during the install step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,15 +11,14 @@ go:
 os:
   - linux
   - osx
+env:
+- GO111MODULE=on
 matrix:
   allow_failures:
     - go: tip
   fast_finish: true
 install:
-  - mkdir -p $HOME/src
-  - mv $HOME/gopath/src/github.com/gohugoio/hugo $HOME/src
-  - export TRAVIS_BUILD_DIR=$HOME/src/hugo
-  - cd $HOME/src/hugo
+  - go mod download
   - go get github.com/magefile/mage
 script:
 - mage -v test


wfscheper:
I used hugo's magefile as an example for converting one of my own projects from make, and ran into many of the same issues with travis, go modules, and mage that you experienced. I thought you might be interested in this alternative solution.

bep:
There are no issues in with parallel builds in the current setup. This is handled in the Mage script (where it belongs). Also, are you really sure that running with `GO111MODULE=on` is exactly the same as the existing setup? I say this because I much prefer a correct but a bit longer Travis script than a short variant with different behaviour.

wfscheper:
Ah, I didn't see where you addressed the parallel build problem in the
magefile. I'll take another look. My issue was cropping up intermittently
due to collisions between go fmt and go vet fetching the same modules.

My understanding is that setting GO111MODULE to 'on' makes go use the
module behavior regardless of whether it is run in the GOPATH. I can double
check that and link to the relevant docs later tonight.

bep:
>My understanding is that setting GO111MODULE to 'on' makes go use the
module behavior regardless of whether it is run in the GOPATH.

Yes, I have read the docs on this. Which is why this variable is also set in the `mage` file. But I have spent a stupid amount of time testing different variations of this, and my conclusion was that the only way to get the correct version of mage via go get (i.e. the version in `go.mod`) was to have the project live outside of GOPATH.

wfscheper:
Yes, but setting GO111MODULE in the magefile only applies when running
mage, not when fetching mage with go get prior to the build.

Issue #5223, opened 2018-09-18
Commit 555a561: broke Taxonomies with custom permalinks configuration


onedrawingperday:
CC / @moorereason 

It seems that commit https://github.com/gohugoio/hugo/commit/06976ebb8778e08fea89dbd9fcb875f6c8d650d4 breaks Taxonomies with Custom Permalinks

In the previous related PR I asked for a test of this [repo](https://github.com/gohugoio/hugo/pull/4388#issuecomment-414379228) and it was OK.

But now with the same unchanged test repo Taxonomies with Custom Permalinks no longer work 100%. My custom Year Taxonomy i.e `2018` works. But monthly Archives i.e `2018/08` and Color Taxonomy i.e `drawing/black` no longer generate.

Also there is no error in the console.

----

I am 100% sure that the problem is caused by this specific commit because I went temporarily to the previous commit of this one by executing `git checkout 555a561` and then compiled from source, by doing that Taxonomies with Custom Permalinks began working again.

You can use my [repo](https://github.com/gohugoio/hugo/pull/4388#issuecomment-414379228) again for testing.

Thanks.

moorereason:
Thanks for the report.  I know I tested your test site with my changes, but I clearly missed something.  I'll look at it again.

PR #5245, opened 2018-09-21
hugolib: Normalize permalink path segments

### LINKED ISSUE #4926: `#` not escaped in Permalink when in title value
Initially posed a question in the forum: https://discourse.gohugo.io/t/hash-pound-sign-not-escaped-in-title-permalink-breaking-urls/12825

This seems really strange to me as a default behavior, and feels like a bug. Allow me to asplain? 

I have a post title, like `tcg's series of posts #1`. 

The [Permalinks configuration examples](http://gohugo.io/content-management/urls/#permalinks-configuration-example) in the docs contain the following example: 

`post = "/:year/:month/:title/"`

... which lead me to believe that this would pretty much automatically work for my not-unreasonable post titles. ("work" here meaning "escape the URL for `.Permalink` and filename value automatically). 

In the forum, [maiki](https://discourse.gohugo.io/t/hash-pound-sign-not-escaped-in-title-permalink-breaking-urls/12825/2) suggested that I might use `:slug` as a workaround, meaning that I would need to always manually provide `:slug`. 

Also, the aforementioned docs says this about `:slug`: 

> the content’s slug (or title if no slug is provided in the front matter)

Meaning that it will always fall back to `:title` if I fail to provide a slug. Which seems to me like **I always have to be sure to manually provide a URL-friendly value... as a human**, writing a document. That doesn't feel right. 

I think this is a bug. (Sadly, I am not yet familiar enough with Go to find a spot and try to offer a PR for the expected behavior).

I am willing to be wrong, and understand I may be Doing It Wrong. 

EDIT: 

Using versions **`0.42.2`** and **`0.43`**, on macOS

moorereason:
When constructing permalinks, ensure that most inputs used as path
segments are normalized with PathSpec.MakeSegment instead of
PathSpec.URLize.

The primary exception to that rule is with taxonomy titles in
pageToPermalinkTitle(). The approach taken here is to use URLize for
taxonomy pages. Everything else will use MakeSegment. The reason for
this exception is that people use taxonomies such as "s1/p1" to generate
URLs precisely they way they wish (see #5223). Tests have been added to
check for this case.

Fixes #4926

### DIFF
--- a/hugolib/permalinks.go
+++ b/hugolib/permalinks.go
@@ -152,9 +152,13 @@ func pageToPermalinkDate(p *Page, dateField string) (string, error) {
 
 // pageToPermalinkTitle returns the URL-safe form of the title
 func pageToPermalinkTitle(p *Page, _ string) (string, error) {
-	// Page contains Node which has Title
-	// (also contains URLPath which has Slug, sometimes)
-	return p.s.PathSpec.URLize(p.title), nil
+	if p.Kind == "taxonomy" {
+		// Taxonomies are allowed to have '/' characters, so don't normalize
+		// them with MakeSegment.
+		return p.s.PathSpec.URLize(p.title), nil
+	}
+
+	return p.s.PathSpec.MakeSegment(p.title), nil
 }
 
 // pageToPermalinkFilename returns the URL-safe form of the filename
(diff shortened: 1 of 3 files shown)


onedrawingperday:
Just tested the merged commit and everything looks fine on my end.

Thank you! 👍

PR #5247, opened 2018-09-22
Allow first function to return an empty array.

### LINKED ISSUE #5235: Support "first 0 .Pages"
```bash
<first 0 .Pages>: error calling first: can't return negative/empty count of items from sequence
```

There is currently no (known) way of creating an empty `Pages` collection. Doing `first 0 .Pages` is the best I could think of, but that fails. I don't see why this should not work.

felicianotech:
Fixes #5235.

### DIFF
--- a/docs/content/en/functions/first.md
+++ b/docs/content/en/functions/first.md
@@ -25,3 +25,5 @@ aliases: []
     {{ .Render "summary" }}
 {{ end }}
 ```
+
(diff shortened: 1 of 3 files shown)


moorereason:
Can you write a test case for this change?

felicianotech:
@moorereason Good point. Incoming.

bep:
For the future, please note the commit message guidelines.

PR #5249, opened 2018-09-23
Build on CircleCI outside of GOPATH

### LINKED ISSUE #5135: circleci: Build project outside of GOPATH
Current is:

```
working_directory: /go/src/github.com/gohugoio
```

This works, but means that we need to set `GO111MODULE=on` in env. We should just move the build outside of GOPATH to make this explicit.

felicianotech:
Fixes #5135 

The default working directory on CircleCI is `~/project` which can be seen expanded when the workspace is attached.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,7 @@
 defaults: &defaults
-  working_directory: /go/src/github.com/gohugoio
   docker:
       - image: bepsays/ci-goreleaser:1.11-2
   environment:
-    GO111MODULE: "on"
     CGO_ENABLED: "0"
 
 version: 2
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -27,7 +26,7 @@ jobs:
     <<: *defaults
     steps:
       - attach_workspace:
-          at: /go/src/github.com/gohugoio
```
bep: This now does not match the working_directory setting above.

felicianotech: Did too many branches last night and cherry picked the wrong thing. The PR is updated to remove the `working_directory` key from the top.

PR #5252, opened 2018-09-25
hugolib: Introduce Page.NextPage and Page.PrevPage

### LINKED ISSUE #1061: .Next vs .NextInSection goes in opposite directions

felicianotech:
Introduce new page position variables in order to fix the ordering issue
of `.Next` and `.Prev` while also allowing an upgrade path via
deprecation.

`.NextInSection` becomes `.NextPageInSection`.
`.PrevInSection` becomes `.PrevPageInSection`.

`.Next` becomes a function returning `.PrevPage`.
`.Prev` becomes a function returning `.NextPage`.

Fixes #1061


bep:
Yea, this looks like a good compromise. To be clear: Is the Next "newer" in both cases now?

@kaushalmodi what do you think about this?

kaushalmodi:
Minor comment on the displayed warning.

Right now it shows:

> WARNING: Page's .Next is deprecated and will be removed in a future release. Use .PrevPage

To a casual user, that might look like a typo. So may be make it more explicit?:

> WARNING: Page's .Next is deprecated and will be removed in a future release. Use .PrevPage (yes, not .NextPage) instead of .Next.

kaushalmodi:
While try to refactor my theme, I noticed that the `.Paginator` has `.Next` and `.Prev` too. Do they need to be refactored too?

kaushalmodi:
Other than those comments, the PR looks great functionally!

Summary of suggested changes:

1. Understand if the `.Next`/`.Prev` in `.Paginator` should be refactored too? Or at least have `.NextPage` and `.PrevPage` as aliases to them (no warning).
2. Make the warning message more explicit.
3. Add file path and line number to the warning.

felicianotech:
@bep 

> To be clear: Is the Next "newer" in both cases now?

Yes. In following with the "default sort", if page weight is the same, `.NextPage` and `.NextPageInSection ` will point to the page that is newer by date. The other two will point to the older page.

@kaushalmodi 

> To a casual user, that might look like a typo. So may be make it more explicit?

I agree with you completely. Problem is, that message is mostly generated from a helper function however I might be able to squeeze something like that in there. I'll check.

> I noticed that the .Paginator has .Next and .Prev too. Do they need to be refactored too in this PR?

I don't know. I honestly did not take Paginator into consideration for this PR. Aside from the naming, do they go in the right direction already? I can take a look at that code and see what's up.

> Is it possible to also display the layout file path and line number in the warning. That way, the user know where exactly the refactoring needs to happen.

Not sure, but that's likely a different PR. As  I mentioned above, Hugo has a helper function for deprecating things that I'm using. I provide what wa deprecated and what to use instead, and it generates that warning message by itself. So maybe a PR enhancing that function is in order?

---

Thanks for the feedback. I'll follow up about Paginator soon.

bep:
>I noticed that the .Paginator has .Next and .Prev too. Do they need to be refactored too in this PR?

No. That is certainly not `.NextPage` -- `.NetxPager` would be more correct, but that just shows that it must be left alone.

bep:
Also note that we may "switch back" to using "Next/Prev" in a future version. This "stunt" is mainly a way to handle the change of direction.

kaushalmodi:
> Also note that we may "switch back" to using "Next/Prev" in a future version. 

+1. I like that plan.

bep:
One question: Which one of the two has now "changed direction"?

kaushalmodi:
> Which one of the two has now "changed direction"?

`.Next`

bep:
>.Next

OK, then I suggest a slightly different approach to this:

* We introduce NextPage and PrevPage and deprecate Next/Prev (like we did here)
* We leave Next/PrevInSection as is, that is: We do not do any deprecation and do not add any NextPageInSection etc.

That will leave us with slightly mismatched names for some versions. But then we can in, say Hugo 0.53 (or something), route Next/Prev to NextPage and just adjust the documentation.

felicianotech:
I pushed a commit to update the deprecation message (thanks @kaushalmodi). I'll rebase before this PR gets merged.

So leave Paginator alone?

> We leave Next/PrevInSection as is

Okay, I kind of liked have "page" in the variable name so that the purpose is more explicit than just saying "next". The upside to making this temporary though is that not including the word "page" makes for shorter variables I guess.

I guess I'll get another commit ready to revert the changes to the InSection variables.

bep:
>The upside to making this temporary though is that not including the word "page" makes for shorter variables I guess.es, 

There are several reasons to leave it as is.

1. It is a method so it has a receiver. `Page.Next`, `Paginator.Next` ... i.e. give me the next of "this", so yes, leave the paginator as is.
2. And it's shorter.

felicianotech:
Okay I pushed a new commit that undoes the deprecation of `.NextInSection` and `.PrevInSection`. I left the doc description changes though as it's more informative to users.

If we like this, I can rebase my commits but also @bep, GitHub can do merge rebase commits built-in. Whichever you prefer.

kaushalmodi:
@felicianotech This PR looks good for merging from my testing. Thanks for working on this.

bep:
>If you keep the two new names consistent per @felicianotech ‘s recommendation, you wouldn’t have to do anything with these in the wild, right?

Not in my head. I would have to update my sites to use `NextPageInSection` and `PrevPageInSection`. Which is making me do work for nothing. And it would also move away from a **consistent** set of other `Next` and `Prev` **methods**.

felicianotech:
Okay I've rebased on master. Once CI tests pass, this PR is ready to go on my end.

Issue #5254, opened 2018-09-25
Multilingual example is broken (pages 404)


robinwassen:
The multilingual example that resides in /hugo/examples/multilingual/ is broken, changing language on the home page works. But not accessing any pages in any language.

### Steps to reproduce

1. Check out the https://github.com/gohugoio/hugo/ repo (latest / b7706d6)
2. Serve the examples/multilingual/ with `hugo server -D`
3. Open the site in your browser
4. Navigate to the About page in any language

### Expected results

You should see the about page

### Actual results

You get a 404

### Hugo version

```
robin@rmbp multilingual (master)$ hugo version
Hugo Static Site Generator v0.49/extended darwin/amd64 BuildDate: unknown
```

robinwassen:
The output suggests that there is a layout missing for taxonomyTerm:

```
robin@rmbp multilingual (master)$ hugo server -v
INFO 2018/09/25 09:55:19 Using config file: /Users/robin/dev/hugo/examples/multilingual/config.toml
Building sites … INFO 2018/09/25 09:55:19 syncing static files to /
INFO 2018/09/25 09:55:19 found taxonomies: map[string]string{"group":"groups"}
INFO 2018/09/25 09:55:19 found taxonomies: map[string]string{"group":"groups"}
INFO 2018/09/25 09:55:19 Alias "en" translated to "en/index.html"
[...]

PR #5256, opened 2018-09-25
docs: Fixed multilingual example compatibility with latest version

robinwassen:
Fixes: #5254 

This is my first time poking around with Hugo, so I am not sure if these changes are perfect. But at least it works to navigate to all pages.


examples/multilingual/content/uudis/_index.et.md
```diff
@@ -0,0 +1,7 @@
++++
+title = "Uudised"
+url = "/uudised"
```
bep: I think we should remove what I consider superflous front matter in this example. In this particular file that would be "url" and "listing". If you could also remove "url" from all other content files. We should encourage using the URLs that Hugo provides out of the box.

robinwassen: I have removed the front matter now. Also organized the content files to get a more coherent structure of the example site.

PR #5259, opened 2018-09-27
Enable page exclusion from sitemap.xml

### LINKED ISSUE #653: Means to exclude content from sitemap.xml
I've looked through the existing documentation on the "Front Matter" and have not been able to find an option to exclude content from appearing in the sitemap.

This use case arises when we want content to be included in "master" pages. For example, our careers page lists all careers by ranging over `.Data.Pages` and displaying the title and content (which amounts to a brief description), though we technically do not want each individual career to have its own landing page.

Normally this isn't an issue, but it becomes one on the sitemap.xml as the pages, which shouldn't really have their own URL, are included in the sitemap. As these pieces of content don't have dates associated with them, Google Webmaster Tools reports an invalid date. It also produces a fairly pointless URL with negligible content.

felicianotech:
This PR adds a new key to .Sitemap called `Exclude`. `Exclude` is a
boolean that is set to false by default meaning all available pages are
included. This is the behavior currently. When `Exclude` is set to true,
it will not appear in any `sitemap.xml` files that Hugo may generate.

`Exclude` can be set to true in the Hugo config turning `sitemap.xml`
into an opt-in rather than an opt-out.

Fixes #653

### DIFF
--- a/docs/content/en/templates/sitemap-template.md
+++ b/docs/content/en/templates/sitemap-template.md
@@ -33,11 +33,14 @@ A sitemap is a `Page` and therefore has all the [page variables][pagevars] avail
 `.Sitemap.Filename`
 : The sitemap filename
 
+.Sitemap.Exclude
+: Whether or not to exclude the page from the sitemap. A boolean that defaults to false.
+
 If provided, Hugo will use `/layouts/sitemap.xml` instead of the internal `sitemap.xml` template that ships with Hugo.
 
 ## Sitemap Templates
 
-Hugo has built-on Sitemap templates, but you can provide your own if needed, in either `layouts/sitemap.xml` or `layouts/_default/sitemap.xml`.
+Hugo has built-in Sitemap templates, but you can provide your own if needed, in either `layouts/sitemap.xml` or `layouts/_default/sitemap.xml`.
 
 For multilingual sites, we also create a Sitemap index. You can provide a custom layout for that in either `layouts/sitemapindex.xml` or `layouts/_default/sitemapindex.xml`.
 
@@ -48,7 +51,7 @@ This template respects the version 0.9 of the [Sitemap Protocol](http://www.site
 ```xml
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
   xmlns:xhtml="http://www.w3.org/1999/xhtml">
-  {{ range .Data.Pages }}
+  {{ range where .Pages ".Sitemap.Exclude" false }}
   <url>
     <loc>{{ .Permalink }}</loc>{{ if not .Lastmod.IsZero }}
     <lastmod>{{ safeHTML ( .Lastmod.Format "2006-01-02T15:04:05-07:00" ) }}</lastmod>{{ end }}{{ with .Sitemap.ChangeFreq }}
@@ -102,6 +105,7 @@ Defaults for `<changefreq>`, `<priority>` and `filename` values can be set in th
   changefreq = "monthly"
   priority = 0.5
   filename = "sitemap.xml"
+  exclude = true
 {{</ code-toggle >}}
 
 The same fields can be specified in an individual content file's front matter in order to override the value assigned to that piece of content at render time.
--- a/hugolib/site_render.go
+++ b/hugolib/site_render.go
@@ -326,6 +326,7 @@ func (s *Site) renderSitemap() error {
 	page.Sitemap.ChangeFreq = sitemapDefault.ChangeFreq
 	page.Sitemap.Priority = sitemapDefault.Priority
 	page.Sitemap.Filename = sitemapDefault.Filename
+	page.Sitemap.Exclude = sitemapDefault.Exclude
 
 	n.data["Pages"] = pages
 	n.Pages = pages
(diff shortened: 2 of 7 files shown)


docs/content/en/templates/sitemap-template.md
```diff
@@ -48,7 +51,7 @@ This template respects the version 0.9 of the [Sitemap Protocol](http://www.site
 ```xml
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
   xmlns:xhtml="http://www.w3.org/1999/xhtml">
-  {{ range .Data.Pages }}
+  {{ range where .Pages ".Sitemap.Exclude" false }}
```
felicianotech: Went with `.Pages` instead of `.Data.Pages` because Hugo Docs says they are the same but  the former is the preferred convention.

hugolib/sitemap_test.go
```diff
@@ -86,11 +86,12 @@ func doTestSitemapOutput(t *testing.T, internal bool) {
 
 func TestParseSitemap(t *testing.T) {
 	t.Parallel()
-	expected := Sitemap{Priority: 3.0, Filename: "doo.xml", ChangeFreq: "3"}
+	expected := Sitemap{ChangeFreq: "3", Priority: 3.0, Filename: "doo.xml", Exclude: true}
```
felicianotech: Exclude was added here but I also changed the order of the properties to match the order they appear below and in Docs.

hugolib/site_render.go
```diff
@@ -324,6 +324,7 @@ func (s *Site) renderSitemap() error {
 	page.Sitemap.ChangeFreq = sitemapDefault.ChangeFreq
 	page.Sitemap.Priority = sitemapDefault.Priority
 	page.Sitemap.Filename = sitemapDefault.Filename
+	page.Sitemap.Exclude = sitemapDefault.Exclude
```
moorereason: I'm confused by this.  `page` is the sitemap page itself.  Is it necessary to exclude/include the sitemap from itself?

Actually, this whole function is confusing.  We create a new sitemap page, `n`, on L314, then create another new sitemap page, `page`, on L319.  We make changes to both nodes (L327, L329), then range over `pages` which _shadows_ the `page` variable (not a good practice).  We finally use `n` in the return statement but only use `page.Sitemap.Filename` there.  It appears like the `page` variable here in line 327 could be merged with `n`.  @bep, can you explain what's happening here?  Should we clean this up while we're here?

hugolib/site_render.go
```diff
@@ -346,6 +347,7 @@ func (s *Site) renderSitemap() error {
 		if page.Sitemap.Filename == "" {
 			page.Sitemap.Filename = sitemapDefault.Filename
 		}
+		// Sitemap.Exclude isn't listed here because bools default to false, so it's done for us.
```
moorereason: If we're going to add `sitemapDefault.Exclude`, we should handle the scenario of `sitemapDefault.Exclude = true`.  However, I don't see how we make that work without `page.Sitemap.Exclude` being of type `*bool`.  If it's `nil`, we know the user didn't set the value.  If it's `false`, we know they mean to override the default.

A couple questions here:  Can we store front matter values into a pointer?  Do we need to add `cast.ToBoolP` to the cast library?

Or am I overthinking everything?  :wink:

hugolib/config.go
```diff
@@ -436,7 +436,7 @@ func loadDefaultSettingsFor(v *viper.Viper) error {
 	v.SetDefault("titleCaseStyle", "AP")
 	v.SetDefault("taxonomies", map[string]string{"tag": "tags", "category": "categories"})
 	v.SetDefault("permalinks", make(PermalinkOverrides, 0))
-	v.SetDefault("sitemap", Sitemap{Priority: -1, Filename: "sitemap.xml"})
+	v.SetDefault("sitemap", Sitemap{Priority: -1, Filename: "sitemap.xml", Exclude: false})
```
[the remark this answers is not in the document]
felicianotech: True. I "assumed" that somewhere else so I should probably do that here.

hugolib/site_render.go
felicianotech: It confused me too. I added this line to follow the convention I saw but I agree, it was confusing. I'll wait to see what bep says.

hugolib/site_render.go
felicianotech: It's already set to a boolean.

hugolib/config.go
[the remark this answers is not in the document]
felicianotech: Addressed in the latest commit.

felicianotech:
I've addressed the comments I could and rebased with the latest master.

DirtyF:
It would be nice if Hugo could offer better SEO by default, thanks for considering reviewing this @bep 🙇

XhmikosR:
@felicianotech: oh, no :/ This was a pretty useful feature. Please consider making a new PR. @bep has so many things to do, hence the delay/lack of replies. But it shouldn't put you off, even though I do feel you.

gavinr:
@felicianotech @bep Why was this closed? I agree, this feature should be added - can we re-open this PR?

felicianotech:
I don't remember. I think I closed it because it was just sitting there and went out-of-date yet again.

I could maybe open this as a new PR but so much has changed in Hugo it'll likely need some re-working.

felicianotech:
Opened a new PR: #6370

Issue #4558, opened 2018-04-02
Page resources not working (as expected) with extract date and slug from filename funtionality


Brixy:
Hi all!

Thanks for the brilliant _extract date and slug from filename_ funtionality.

Page resources does not seem to work in such a case:

```
.
└── blog
    ├── any-folder-name
    │   └── 2018-03-02-file-name.md
    │   └── file.pdf
```

When building, file.pdf is located in `blog/2018-03-02-file-name`.

Yet, I would like to link to file: `<a href="file.pdf">Download</a>`.

That is, the file should be located depending on config settings, e. g.:

```
[permalinks]
blog = "/:slug/"
```

This would be even simpler:

```
.
└── blog
    ├── 2018-03-02-FILE-name
    │   └── index.md
    │   └── file.pdf
```

Thank you!

Brixy:
Would this simple concept make sense:

If filename is `index.md` or `_index.md` get date and slug from parent folder name.

Slug sould be treated as if in front matter. Folder `2018-04-13-name` ist not generated, but e. g. `/2018/name`—including all page resources + index.html.

(Should other md files in the folder also be published to `/2018/name/`?? Would make sense in my setups.)

Japanuspus:
Just wanted to comment that [the duplicate issue](https://github.com/gohugoio/hugo/issues/4609) opened by @onedrawingperday  was more clearly stated.

The date from filename implementation should treat files and leaf page bundles equally, so that the following input file structures should result in the same site:

Input as leaf bundle
```
.
└── blog
    └── 2018-03-02-FILE-name
        └── index.md
``` 
Input as single file
```
.
└── blog
    └── 2018-03-02-FILE-name.md
```

onedrawingperday:
Begone stale bot.

This issue has merit.

Japanuspus:
I have a [proposed fix](https://github.com/Japanuspus/hugo/commit/220368a94ae8236ff4d4a43a3557647e272b530f), and is starting to read through the [contribution guidelines](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md). Is anybody who is already following this issue able to review, or should I ping someone?

This is my first go-code ever, and I must say that the toolchain experience was very nice. From no tools and no language knowledge to an implemented fix and a working binary in the course of a few hours is hard to beat!

## Details of fix
[...]

PR #5267, opened 2018-09-30
Date and slug from filename for leaf bundles

Japanuspus:
The definition of the `BaseFilename` field in the `FrontMatterDescriptor` structure provided to the frontmatter handlers has been changed, so that in the case of a leaf bundle `index.`-file, the name provided is that of the containing bundle folder.
Since the use of the `FrontMatterDescriptor` is localized to the frontmatter handler, this should not have external effects.

To compute the new value of `BaseFilename`, I have added a `ContentBaseName` method to the `File` interface. The `FileInfo` implementation of this interface has been modified to compute the value as part of `lazyinit`. 

This is my first ever `go`-code, so feel free to rewrite as you wish, or let me know if I should implement in another way.

### DIFF
--- a/source/fileInfo.go
+++ b/source/fileInfo.go
@@ -191,11 +191,11 @@ func (fi *FileInfo) init() {
 		}
 		fi.section = section
 
-		var contentBaseName = fi.translationBaseName
 		if fi.isLeafBundle && len(parts) > 0 {
-			contentBaseName = parts[len(parts)-1]
+			fi.contentBaseName = parts[len(parts)-1]
+		} else {
+			fi.contentBaseName = fi.translationBaseName
 		}
-		fi.contentBaseName = contentBaseName
 
 		fi.uniqueID = helpers.MD5String(filepath.ToSlash(fi.relPath))
 	})
--- a/source/fileInfo.go
+++ b/source/fileInfo.go
@@ -70,6 +70,10 @@ type File interface {
 	// not even the optional language extension part.
 	TranslationBaseName() string
 
+	// ContentBaseName is a either TranslationBaseName or name of containing folder
+	// if file is a leaf bundle.
+	ContentBaseName() string
+
 	// UniqueID is the MD5 hash of the file's path and is for most practical applications,
 	// Hugo content files being one of them, considered to be unique.
 	UniqueID() string
@@ -106,6 +110,7 @@ type FileInfo struct {
 	relPath             string
 	baseName            string
 	translationBaseName string
+	contentBaseName     string
 	section             string
 	isLeafBundle        bool
 
@@ -144,6 +149,13 @@ func (fi *FileInfo) BaseFileName() string { return fi.baseName }
 // language segement (ie. "page").
 func (fi *FileInfo) TranslationBaseName() string { return fi.translationBaseName }
 
+// ContentBaseName is a either TranslationBaseName or name of containing folder
+// if file is a leaf bundle.
+func (fi *FileInfo) ContentBaseName() string {
+	fi.init()
+	return fi.contentBaseName
+}
+
 // Section returns a file's section.
 func (fi *FileInfo) Section() string {
 	fi.init()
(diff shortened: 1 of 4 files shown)


Japanuspus:
Due to bad github-foo, I am missing a link to the issue: #4558

bep:
@Japanuspus just a note that I have not forgotten this PR, I will get to it in one of the next few days ...

moorereason:
@Japanuspus 
If you're willing and able, you can update your commit message to [follow our guidelines](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines).  We :heart: clean commit messages.  😃

bep:
Other than the before-mentioned commit message, I think this looks fine. I don't love the name of the `ContentBaseName` method, but I cannot come up with anything better. That interface is (from historical reasons) a little too big and slightly confusing, but that is out of scope for this.

@moorereason what do you think?

source/fileInfo.go
```diff
@@ -177,11 +189,15 @@ func (fi *FileInfo) init() {
 		if (!fi.isLeafBundle && len(parts) == 1) || len(parts) > 1 {
 			section = parts[0]
 		}
-
 		fi.section = section
 
-		fi.uniqueID = helpers.MD5String(filepath.ToSlash(fi.relPath))
+		var contentBaseName = fi.translationBaseName
```
moorereason: Simplify this section to avoid writing to `contentBaseName` twice on leaves.

```go
if fi.isLeafBundle && len(parts) > 0 {
    fi.contentBaseName = parts[len(parts)-1]
} else {
    fi.contentBaseName = fi.translationBaseName
}
```

source/fileInfo.go
```diff
@@ -177,11 +189,15 @@ func (fi *FileInfo) init() {
...
+		if fi.isLeafBundle && len(parts) > 0 {
+			contentBaseName = parts[len(parts)-1]
+		}
+		fi.contentBaseName = contentBaseName
 
+		fi.uniqueID = helpers.MD5String(filepath.ToSlash(fi.relPath))
```
moorereason: Unrelated change?

source/fileInfo_test.go
```diff
@@ -94,4 +94,14 @@ func TestFileInfoLanguage(t *testing.T) {
...
+
+	// test contentBaseName implementation
+	fi1 := s.NewFileInfo("", "2018-10-01-contentbasename.md", false, nil)
+	fi2 := s.NewFileInfo("", "2018-10-01-contentbasename.en.md", false, nil)
+	fi3 := s.NewFileInfo("", "2018-10-01-contentbasename"+helpers.FilePathSeparator+"index.en.md",
+		true, nil)
+
+	assert.Equal("2018-10-01-contentbasename", fi1.ContentBaseName())
+	assert.Equal("2018-10-01-contentbasename", fi2.ContentBaseName())
+	assert.Equal("2018-10-01-contentbasename", fi3.ContentBaseName())
```
moorereason: Need a test that ContentBaseName equals TranslationBaseName when not in a bundle.

source/fileInfo_test.go
```diff
@@ -94,4 +94,14 @@ func TestFileInfoLanguage(t *testing.T) {
 
 	assert.Equal("sv", fiSv.Lang())
 	assert.Equal("en", fiEn.Lang())
+
+	// test contentBaseName implementation
+	fi1 := s.NewFileInfo("", "2018-10-01-contentbasename.md", false, nil)
+	fi2 := s.NewFileInfo("", "2018-10-01-contentbasename.en.md", false, nil)
+	fi3 := s.NewFileInfo("", "2018-10-01-contentbasename"+helpers.FilePathSeparator+"index.en.md",
```
moorereason: Use [`filepath.Join`](https://golang.org/pkg/path/filepath/#Join).

source/fileInfo.go
Japanuspus: My initial implementation actually used direct assignment, but then I spotted this local variable pattern in the lines just above (implementing `section`), and thought it might be a golang idiom...

I'll change it back to direct assignment which I also find easier to read.

source/fileInfo.go
Japanuspus: Seems to be just a bad diff -- this line was right below my insert. I haven't been able to find out why git is marking this as delete/add.

Japanuspus:
Thanks, @bep and @moorereason. 
I don't quite like the `contentBaseName` either, but `leafName` would obviously not make sense to use in the non-leaf case, so I gave up finding a better alternative.

With respect to the implementation, another approach would be to put the logic in `page.go`, but that seemed much harder to test.

Have just pushed a new commit to this branch without squashing, hope that is ok. New commit message should be usable for PR -- but should I amend the original commit?

vassudanagunta:
The logical name is LogicalName, but that's taken.

bep:
@moorereason is this good to go now?

PR #5271, opened 2018-10-02
tpl: Add a delimiter parameter to lang.NumFmt

### LINKED ISSUE #5260: use space in lang.NumFmt options
With French typographic rules, prices are supposed to be printed that way:

`1 000,23€`

But I can't get `lang.NumFmt` to print a space instead of a dot... I tried `lang.NumFmt 2 (.Get 0) "- ,  "` and `lang.NumFmt 2 (.Get 0) "- ,  "` (nbsp), but both are replaced by no space at all. `lang.NumFmt 2 (.Get 0) "- , &nbsp;"` gave `1&nbsp;000,23` and `lang.NumFmt 2 (.Get 0) "- , \ "` nothing at all.

moorereason:
The original implementation of NumFmt did not take into account that the
options delimiter (a space) could be a valid option.  Adding a delim
parameter seems like the simplest, safest, and most flexible way to
solve this oversight in a backwards-compatible way.

Fixes #5260


kaushalmodi:
Can you also update the docs that go along with this PR?

moorereason:
I was waiting to see if this approach would be acceptable.  It looks like @bep agrees with your comment, so I'll add docs to this PR.  👍

PR #5272, opened 2018-10-02
Use stdlib context package

GregorioMartinez:
Hugo now depends on go 1.11
With that bump we no longer need to use `golang.org/x/net/context` and can instead use the one found in stdlib

### DIFF
--- a/hugolib/page_bundler.go
+++ b/hugolib/page_bundler.go
@@ -14,13 +14,11 @@
 package hugolib
 
 import (
+	"context"
 	"fmt"
 	"math"
 	"runtime"
 
-	// Use this until errgroup gets ported to context
-	// See https://github.com/golang/go/issues/19781
-	"golang.org/x/net/context"
 	"golang.org/x/sync/errgroup"
 )
 
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -14,6 +14,7 @@
 package hugolib
 
 import (
+	"context"
 	"errors"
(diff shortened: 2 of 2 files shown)


bep:
The Hugo motivation for using `golang.org/x/net/context` is this issue https://github.com/golang/go/issues/19781 -- that may or not be a good reason, but it has never been related to Go 1.11 (`net/context` was added in Go 1.8 or something).

GregorioMartinez:
My mistake. Should have checked in about why it was still there instead of assuming it was an oversight from before `context` was included in the stdlib.

Reading that thread it sounds like the Go team is going to be using `golang.org/x/net/context` in `golang.org/x/sync/errgroup/`to continue supporting Go 1.6 users for a bit longer.

Does it make sense to update Hugo's dependency and allow `/golang.org/x/sync/errgroup` to specify it's own? Is there a reason they should be the same?

hugolib/page_bundler.go
```diff
@@ -14,13 +14,13 @@
 package hugolib
 
 import (
+	"context"
 	"fmt"
 	"math"
 	"runtime"
 
 	// Use this until errgroup gets ported to context
 	// See https://github.com/golang/go/issues/19781
```
bep: If you could also remove the now superflous comments above, I'll merge.

GregorioMartinez: Done and combined into one commit.

PR #5289, opened 2018-10-06
docs: Fix typo in quick-start.md

twlz0ne:
The original command put the article into `content/posts`, then there is nothing display on the home page.

It does give the First-time users like me a little trouble.


bep:

twlz0ne:
@bep https://github.com/gohugoio/hugoDocs/pull/628

PR #5306, opened 2018-10-11
Extend youtube shortcode to support playlists

Kissaki:
The player will load the first video of the playlist by default.
The playlist is available in the YouTube player menu (top left).

I’m not sure if this is the best way to code the template.
I declare `$path`, and then determine the subpath in an if else block, and append the adequate autoplay query parameter if necessary.

### DIFF
--- a/tpl/tplimpl/embedded/templates/shortcodes/youtube.html
+++ b/tpl/tplimpl/embedded/templates/shortcodes/youtube.html
@@ -2,8 +2,16 @@
 {{- if not $pc.Disable -}}
 {{- $ytHost := cond $pc.PrivacyEnhanced  "www.youtube-nocookie.com" "www.youtube.com" -}}
 {{- $id := .Get "id" | default (.Get 0) -}}
+{{- $path := "" -}}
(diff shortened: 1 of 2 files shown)


tpl/tplimpl/embedded/templates/shortcodes/youtube.html
```diff
@@ -2,8 +2,16 @@
 {{- if not $pc.Disable -}}
 {{- $ytHost := cond $pc.PrivacyEnhanced  "www.youtube-nocookie.com" "www.youtube.com" -}}
 {{- $id := .Get "id" | default (.Get 0) -}}
+{{- $path := "" -}}
+{{- if hasPrefix $id "PL" -}}
+{{- $path = printf "%s%s" "?listType=playlist&list=" $id -}}
```
moorereason: Your `printf` statements can simplified (in my eyes) to only one substitution:

```
{{- $path = printf "?listType=playlist&list=%s" $id -}}
```

moorereason:
@Kissaki,
Can you provide documentation from Youtube that stand-alone videos will never has an ID that starts with `PL`?

PR #5316, opened 2018-10-15
Try out Windows builds in Travis

emirb:
Trying out [early Windows release](https://blog.travis-ci.com/2018-10-11-windows-early-release)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ go:
 os:
   - linux
   - osx
+  - windows
 matrix:
   allow_failures:
     - go: tip
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
 git:
   depth: false
 go:
-  - 1.11.x
+  - 1.11
   - tip
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ matrix:
   allow_failures:
     - go: tip
   fast_finish: true
+before_install:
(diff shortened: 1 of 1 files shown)


bep:
It would be cool if we could use Travis only, but with the extended build, our build setup is not trivial.

moorereason:
Travis is working on the ["There is not enough space on the disk" issue](https://travis-ci.community/t/there-is-not-enough-space-on-the-disk/338).

moorereason:
Windows builds on Travis are green ([except for tip](https://travis-ci.community/t/go-command-not-found/340))! :tada:

This is big.

bep:
Yes, this is pretty great. The "Go tip" build is a way to get "early warnings" about API builds etc., so that is fine if we could turn it off. That said, we usually build for 2 Go versions, so we should think about how to set it up once Go 1.12 is out etc.

bep:
@emirb what is the status of this? Is there a way to configure "the matrix" re. the discussion above?

emirb:
@bep If I understood your intentions correctly, it is already set up like that.
Travis runs first matrix of tests with 1.11 and another with tip. Failures are allowed for tip, so we can get early warnings, however, they won't break the build.

bep:
>@bep If I understood your intentions correctly, it is already set up like that.

My intentions are to get a setup where I can exclude a Go version for an OS. That failures on tip "is allowed" doesn't mean that we want to run it for every commit, that is a waste and confuses people who look at the build status.

emirb:
@bep Got you; the latest commit does just that.

Issue #5315, opened 2018-10-15
Upgrade 'Error: Unable to locate Config file' message


marekkowalczyk:
Consider modifying the error message **‘Error: Unable to locate Config file. Perhaps you need to create a new site. Run hugo help new for details.’** issued in [this usecase](https://discourse.gohugo.io/t/hugo-new-test-md-throws-unable-to-locate-config/14708?u=marekkowalczyk) to include the suggestion: **ou need to call hugo new test.md from the root of the project.**. 

HT [pointyfar](https://discourse.gohugo.io/u/pointyfar/summary).

moorereason:
Fixed by #5317

PR #5317, opened 2018-10-15
Added helpful line to new command reminding people to run within the …

elliotforbes:
…root directory of their site


elliotforbes:
Attempts to address - https://github.com/gohugoio/hugo/issues/5315#issue-370063994

elliotforbes:
Apologies, I've run gofmt on the file in question. Looks like it's fixed and I absolutely agree this formatting needs to be done to keep things consistent so no worries! :)

marekkowalczyk:
Thanks you for the prompt fix! It will save uses many hours of trial and error.

PR #5319, opened 2018-10-16
Generate .gitignore when creating a new site

napei:
I thought it would be a useful addition to add a `.gitignore` file when the site is first generated using `hugo new site [dir]`.

At present the file contains a single listing for the `public/` directory. This doesn't necessarily need to be committed every time there's a change as most people I've seen add a gitignore file anyway for that directory.

I'm not sure about any other directories/files that should be added to this file. Perhaps the `resources/` directory also as it's generated by a `hugo` command also.

Let me know what you think.

### DIFF
--- a/commands/new_site.go
+++ b/commands/new_site.go
@@ -97,6 +97,7 @@ func (n *newSiteCmd) doNewSite(fs *hugofs.Fs, basepath string, force bool) error
 	}
 
 	createConfig(fs, basepath, n.configFormat)
+	createGitignore(fs, basepath)
 
 	// Create a default archetype file.
 	helpers.SafeWriteToDisk(filepath.Join(archeTypePath, "default.md"),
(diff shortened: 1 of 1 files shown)


onedrawingperday:
No. The resources directory is very much needed in Git.

Other than that this seems like an interesting addition.

napei:
@bep not to throw a spanner in the works, but I just tested this on windows and it doesn't work as windows prohibits creating dotfiles (technically no name). I don't quite know enough about Go to troubleshoot this.

I can confirm that this works on linux however, as that's my main working environment.

bep:
@napei OK, thanks for the heads up. I have reverted the relevant change.

We could maybe add a `if runtime.GOOS != "windows"` or something, but now we probably need a test for it to verify.

napei:
@bep Oops disregard my stupidity. Can't run a compiled hugo if it's already in your path somewhere else.

I changed output to a *.exe file and all works well. Sorry for the mixup. This PR is good on windows.

PR #5323, opened 2018-10-16
resource: Optimize integrity string generation

moorereason:
Remove use of fmt.Sprintf for simple string concatenation.  A simple
change for a small perf boost.

```
name         old time/op    new time/op    delta
Integrity-4     525ns ± 2%     268ns ± 2%  -48.92%  (p=0.000 n=10+10)

name         old alloc/op   new alloc/op   delta
Integrity-4      144B ± 0%      112B ± 0%  -22.22%  (p=0.000 n=10+10)

name         old allocs/op  new allocs/op  delta
Integrity-4      5.00 ± 0%      3.00 ± 0%  -40.00%  (p=0.000 n=10+10)
```


bep:
Sure, but note that a 250ns save like this is only worth while in a very hot loop, and in most cases I would keep the fmt version for clarity. But in this case, the new code actually looks better.

PR #5336, opened 2018-10-20
hugolib: Extract title from filename

### LINKED ISSUE #3805: Use filename as title when it’s not defined in front matter

maxandersen:
Why:

 * Would be nice to just add content without a frontmatter.
   title is at the moment the only required field in frontmatter for
   hugo.
 * Like to migrate existing content more easily to frontmatter.
   Existing static generators can derive title from filename or content
   markup.

This change addreses the need by:

 * use the same mechanics added for handling date fields in
   commit: 68bf1511f2b to allow configure how title is configured.

   The default for doing title is simply:

   ```toml
   [frontmatter]
   title = [ "title" ]
   ```

   So, if you want to use a different field than `title` this works:

   ```toml
   [frontmatter]
   title = [ "mytitle", ":default" ]
   ```

   This commit adds support for deriving title from filename and possible
   to combine with filename suppor for dates.

   A good configuration for deriving date and titles from filenames while still
   honoring values found in frontmatter are:

   ```toml
   [frontmatter]
   date = [ ":default", "filename" ]
   title = [ ":default", "filename" ]
   ```

   Fixes #3805

### DIFF
--- a/docs/content/en/getting-started/configuration.md
+++ b/docs/content/en/getting-started/configuration.md
@@ -373,6 +373,23 @@ The above will try first to extract the value for `.Date` from the filename, the
 `:git`
 : This is the Git author date for the last revision of this content file. This will only be set if `--enableGitInfo` is set or `enableGitInfo = true` is set in site config.
 
+## Configure Title
+
+It is possible to enable deriving title from the filename/slug. 
+
+`:filename`
+
+: Fetches the title from the content file's filename. For example, `my_page.md` will extract the title `My Page`. 
+If you use `:filename` for date too then `2018-02-22-my_page.md` will extract both date and title from the filename.
+
+An example:
+
+```toml
+[frontmatter]
+date  = [":filename", ":default"]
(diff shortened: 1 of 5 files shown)


maxandersen:
whoops - bad commit; redoing.

Issue #5195, opened 2018-09-10
.IsTranslated is set to `true` although there is no translated content


nop33:
Here's a minimal example to reproduce the issue: https://github.com/nop33/hugo-minimal-setup

This is the project structure
```
project/
└── content/
    ├── subpage/
    │   ├── _index.fr.md
    │   └── _index.md
    └── _index.md
```

And these are my language settings in the `.config.yml`:
``` yaml
languageCode: "en-us"
defaultContentLanguage: "en"
languages:
  en:
    languageName: "English"
    weight: 1
  fr:
    languageName: "French"
    weight: 2
```

Now, when I check for `{{ .IsTranslated }}` in the template where the home page is rendered, it is set to `true`, even though a `content/_index.fr.md` file doesn't exist.

Since a translation of the `content/_index.md` has not been defined, why is Hugo creating an empty page for the French translation, in the particular case, the French home page?

```
Hugo Static Site Generator v0.48 darwin/amd64 BuildDate: 2018-08-29T06:33:08Z
GOOS="darwin"
GOARCH="amd64"
GOVERSION="go1.11"
```

I asked for help from the community, but haven't received any replies yet: https://discourse.gohugo.io/t/how-to-translate-only-a-subset-of-the-pages/14083

I also found these 2 issues, but they didn't help:
* https://github.com/gohugoio/hugo/issues/3470
* https://github.com/gohugoio/hugo/issues/2699

bep:
Hugo will create default variants of the home page and any root section (/blog). This has always been the case, even before multiple languages etc. You may argue that `.IsTranslated` may be a little misleading in this case, but we're not changing it.

nop33:
Well, it is not only misleading, Hugo generates the `public/fr/index.html` page whereas the `domain.name/fr` page should give a 404.

So to rephrase, one cannot translate a subpage of the website without translating the home page as well. Correct?

nop33:
I don't understand why you close this issue, since it is an unexpected behavior that is not documented anywhere. I would understand the closure of the issue if this behavior was expected and documented somewhere.

nop33:
> Hugo will create default variants of the home page and any root section (/blog).

@bep, can you please point me to the place in the docs where this is described? Thanks!

PR #5338, opened 2018-10-20
Add note regarding missing translation files

nop33:
Based on [this discussion](https://github.com/gohugoio/hugo/issues/5195#issuecomment-419834325) I believe it is important to mention in the documentation that Hugo will generate root section page translation files in the case of existing subpage translation files.


bep:
Thanks, and sorry for the delay, but we do documentation changes in its own repository:  https://github.com/gohugoio/hugoDocs/pulls

PR #5339, opened 2018-10-21
Allow "jsonify" to pretty print JSON

### LINKED ISSUE #5040: Allow "jsonify" to pretty print JSON

SeanPrashad:
Fixes #5040: Allow "jsonify" to pretty print JSON

This PR replaces `json.Marshal` with `json.MarshalIndent`, to return pretty-printed output instead of compact. A sensible spacing of two spaces has chosen for the indentation.

### DIFF
--- a/tpl/encoding/encoding.go
+++ b/tpl/encoding/encoding.go
@@ -50,9 +50,9 @@ func (ns *Namespace) Base64Encode(content interface{}) (string, error) {
 	return base64.StdEncoding.EncodeToString([]byte(conv)), nil
 }
 
-// Jsonify encodes a given object to JSON.
+// Jsonify encodes a given object to JSON, returning pretty printed output.
 func (ns *Namespace) Jsonify(v interface{}) (template.HTML, error) {
-	b, err := json.Marshal(v)
+	b, err := json.MarshalIndent(v, "", "  ")
 	if err != nil {
 		return "", err
 	}
--- a/docs/content/en/functions/jsonify.md
+++ b/docs/content/en/functions/jsonify.md
@@ -1,7 +1,7 @@
 ---
 title: jsonify
 linktitle: jsonify
-description: Encodes a given object to JSON.
+description: Encodes a given object to JSON, returning pretty printed output.
 godocref:
 date: 2017-02-01
 publishdate: 2017-02-01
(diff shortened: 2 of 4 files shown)


tpl/encoding/encoding.go
```diff
@@ -50,9 +50,13 @@ func (ns *Namespace) Base64Encode(content interface{}) (string, error) {
...
-// Jsonify encodes a given object to JSON.
-func (ns *Namespace) Jsonify(v interface{}) (template.HTML, error) {
-	b, err := json.Marshal(v)
+// Jsonify encodes a given object to JSON, returning pretty printed output.
+func (ns *Namespace) Jsonify(v interface{}, indentSpacing string) (template.HTML, error) {
+	if indentSpacing == "" {
+		indentSpacing = "  "
+	}
+
+	b, err := json.MarshalIndent(v, "", indentSpacing)
```
SeanPrashad: Getting some very weird behaviour here - when I switched to `json.MarshalIndent` and updated the respective table test, there was a panic occurring. I was looking into the docs for `template.HTML` but everything looked ok at first glance.

SeanPrashad:
Small snippet for how json.marshalIndent() works:

```go
package main

import (
	"encoding/json"
	"fmt"
	"log"
)

func main() {

	data := []string{"A", "B", "C"}

	json, err := json.MarshalIndent(data, "", "  ")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(string(json))
}
```

prints:
```
[
  "A",
  "B",
  "C"
]
```

tpl/encoding/encoding.go
SeanPrashad: Respective logs from Travis: https://travis-ci.org/gohugoio/hugo/jobs/444259677#L759

SeanPrashad:
Currently running on `go version go1.11.1 darwin/amd64` if that helps

tpl/encoding/encoding.go
SeanPrashad: Currently running on go version go1.11.1 darwin/amd64 if that helps

SeanPrashad: **Update**: `template.HTML("[\n  \"a\",\n  \"b\"\n]")` was the answer - I had to escape quotes

bep:
Thanks for this, and a special thanks for keeping the commit message in line ... so I could just click the button

PR #5366, opened 2018-10-27
Bump Go versions and use 1.n.x to get latest minor versions

HaraldNordgren:


bep:
The versioning in Hugo is a coordinated dance, and we cannot do "x" versions.

PR #5157, opened 2018-09-03
Add support for featured and global image to OpenGraph template

max-arnold:
Make `opengraph.html` image tag behave in the same way as `twitter_card.html`. Now it can pick an image like `*cover*` or `*thumbnail*` from page bundle, plus also respects `.Site.Params.images`.

PR #5378, opened 2018-10-30
Add support for featured and global image to OpenGraph template

max-arnold:
Make `opengraph.html` image tag behave in the same way as `twitter_card.html`. Now it can pick an image like `*cover*` or `*thumbnail*` from page bundle, plus also respects `.Site.Params.images`

This is a copy of #5157 which I deleted accidentally while rebasing my branch to the latest master.
See this comment for additional info on how I tested it https://github.com/gohugoio/hugo/pull/5157#issuecomment-419701217

### DIFF
--- a/tpl/tplimpl/embedded/templates/opengraph.html
+++ b/tpl/tplimpl/embedded/templates/opengraph.html
@@ -2,9 +2,18 @@
 <meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
 <meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}" />
 <meta property="og:url" content="{{ .Permalink }}" />
-{{ with $.Param "images" }}{{ range first 6 . }}
+{{ with $.Params.images }}{{ range first 6 . -}}
 <meta property="og:image" content="{{ . | absURL }}" />
-{{ end }}{{ end }}
+{{ end }}{{ else -}}
+{{- $images := $.Resources.ByType "image" -}}
+{{- $featured := $images.GetMatch "*feature*" -}}
+{{- if not $featured }}{{ $featured = $images.GetMatch "{*cover*,*thumbnail*}" }}{{ end -}}
+{{- with $featured -}}
+<meta property="og:image" content="{{ $featured.Permalink }}"/>
+{{ else -}}
+{{- with $.Site.Params.images -}}
+<meta property="og:image" content="{{ index . 0 | absURL }}"/>
+{{ end }}{{ end }}{{ end }}
 
 {{- $iso8601 := "2006-01-02T15:04:05-07:00" -}}
 {{- if .IsPage }}
@@ -13,8 +22,7 @@
 {{ end }}
 {{- if not .Lastmod.IsZero }}<meta property="article:modified_time" {{ .Lastmod.Format $iso8601 | printf "content=%q" | safeHTMLAttr }} />{{ end }}
 {{- else }}
-{{- if not .Date.IsZero }}
-<meta property="og:updated_time" {{ .Date.Format $iso8601 | printf "content=%q" | safeHTMLAttr }} />
+{{- if not .Date.IsZero }}<meta property="og:updated_time" {{ .Date.Format $iso8601 | printf "content=%q" | safeHTMLAttr }} />
 {{- end }}
 {{- end }}{{/* .IsPage */}}
 
--- a/tpl/tplimpl/embedded/templates.autogen.go
+++ b/tpl/tplimpl/embedded/templates.autogen.go
@@ -185,9 +185,18 @@ if (!doNotTrack) {
 <meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
 <meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}" />
 <meta property="og:url" content="{{ .Permalink }}" />
-{{ with $.Param "images" }}{{ range first 6 . }}
+{{ with $.Params.images }}{{ range first 6 . -}}
 <meta property="og:image" content="{{ . | absURL }}" />
-{{ end }}{{ end }}
+{{ end }}{{ else -}}
+{{- $images := $.Resources.ByType "image" -}}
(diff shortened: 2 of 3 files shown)


tpl/tplimpl/embedded/templates/opengraph.html
```diff
@@ -2,18 +2,26 @@
...
 <meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}" />
 <meta property="og:url" content="{{ .Permalink }}" />
-{{- with .Params.images }}{{ range first 6 . }}
+{{ with .Params.images }}{{ range first 6 . -}}
 <meta property="og:image" content="{{ . | absURL }}" />
-{{ end }}{{ end }}
+{{ end }}{{ else -}}
+{{- $images := $.Resources.ByType "image" -}}
+{{- $featured := $images.GetMatch "*feature*" -}}
+{{- $featured := cond (ne $featured nil) $featured ($images.GetMatch "{*cover*,*thumbnail*}") -}}
```
bep: This will be slightly faster if you do:

```
{{ $if not $featured }}
{{- $featured = $images.GetMatch "{*cover*,*thumbnail*}" -}}
```

bep:
This looks very good, the reason I've been holding off on this a little is the "what names do we choose"?

This PR checks images with names matching:

1. "*feature*"
2. "*cover*"
3. "*thumb*"

Which I guess matches some implementations in the wild.

If we can agree on the above, this can be merged. I think the names above makes sense, but we should consider using a double asterisk, e.g. "**feature*" to allow "images/featured.jpg"

@kaushalmodi @budparr @regisphilibert @digitalcraftsman @onedrawingperday and gang

max-arnold:
@bep The naming and most of the logic was inherited from the existing `twitter_cards.html` for consistency.

max-arnold:
Do you want me to implement the suggested improvement for the `twitter_cards` template too?

```
{{ $if not $featured }}
{{- $featured = $images.GetMatch "{*cover*,*thumbnail*}" -}}
```

https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/twitter_cards.html#L7

bep:
>Do you want me to implement the suggested improvement f

Let us wait for some of the people I named above to chime in with their opinions.

regisphilibert:
>  I think the names above makes sense, but we should consider using a double asterisk, e.g. **feature* to allow images/featured.jpg

Not sure.

Double asterisks means any images wherever in the bundle with a filepath containing any of the 3 keywords wins. 
`/covers/song_image_3.jpg`
`/gallery_thumbnails/too-small-for-facebook-01.jpg`
etc...

This may make unrelevant images more likely to be defined as featured when user simple does not want any for this particular page.

Maybe some noise in the discourse upon introduction. 

My own two cents.

bep:
```
/covers/song_image_3.jpg
/gallery_thumbnails/too-small-for-facebook-01.jpg
```

Good point, but that could be handled with:

```bash
{{- $featured := $images.GetMatch "{*feature*,**/feature*}" -}}
```

But it's not important. It is probably best to take the simplest road with less potential friction. Just thought we should discuss it a little.

max-arnold:
This was partially implemented in #5636

tpl/tplimpl/embedded/templates/opengraph.html
max-arnold: @bep Did you mean the following change (or `$if` is a thing in Hugo)?

```diff
-{{- $featured := cond (ne $featured nil) $featured ($images.GetMatch "{*cover*,*thumbnail*}") -}}
+{{- if not $featured -}}
+{{- $featured = $images.GetMatch "{*cover*,*thumbnail*}" -}}
+{{- end -}}
```

max-arnold:
Stale bot, please go away...

tpl/tplimpl/embedded/templates/opengraph.html
bep: OK, first: I'm sorry that I let this hanging. We need to add this to Hugo.

What I meant above was:

```bash
{{ $if not $featured }}
{{- $featured = $images.GetMatch "{*cover*,*thumbnail*}" -}}
```

bep:
@max-arnold what is the status of this re my comment(s)? I would love to merge this when it's done.

max-arnold:
@bep Sorry, I completely forgot about that. I'll update the PR tomorrow.

tpl/tplimpl/embedded/templates/opengraph.html
```diff
@@ -2,9 +2,18 @@
 <meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
 <meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}" />
 <meta property="og:url" content="{{ .Permalink }}" />
-{{ with $.Param "images" }}{{ range first 6 . }}
+{{ with $.Params.images }}{{ range first 6 . -}}
```
bep: The above isn't correct (or: it's a breaking change that is not related to this)

bep: Other than that this looks great. Could you run `mage generate` to regenerate the Go code from the templates=

max-arnold:
@bep I changed the if statement to address your comment and rebased the PR.

Also I changed the same if statement in `twitter_cards.html` to make both templates consistent:

From

`{{- $featured := cond (ne $featured nil) $featured ($images.GetMatch "{*cover*,*thumbnail*}") -}}`

To

`{{- if not $featured }}{{ $featured = $images.GetMatch "{*cover*,*thumbnail*}" }}{{ end -}}`

tpl/tplimpl/embedded/templates/opengraph.html
max-arnold: @bep I'm not very familiar with Go, and unfortunately, the `mage generate` command failed (I ran `go install` before it):

```
% mage generate

Error parsing magefiles: failed to check types in directory: magefile.go:6:2: could not import bytes (type-checking package "bytes" failed (/opt/local/lib/go/src/bytes/buffer.go:10:2: could not import errors (type-checking package "errors" failed (/opt/local/lib/go/src/errors/wrap.go:8:2: could not import internal/reflectlite (type-checking package "internal/reflectlite" failed (/opt/local/lib/go/src/internal/reflectlite/value.go:8:2: could not import runtime (type-checking package "runtime" failed (/opt/local/lib/go/src/runtime/mem_darwin.go:14:6: sysAlloc redeclared in this block))))))))
```

max-arnold: Reinstalled the mage binary and the command succeeded

Issue #4462, opened 2018-02-28
Hugo outputs weird characters on Windows Console(s)


seshaljain:
Firstly, `hugo version`:
Hugo Static Site Generator v0.36.1 windows/386 BuildDate: 2018-02-15T09:08:18Z

I started using Hugo recently and saw `←[K25lBuilding sites . ←[?25h` while running hugo from Git Bash, CMD and bash via integrated terminal in VS Code.

I think they are formatting characters, but could not find any issues or threads regarding them so created a new one.

(CMD output):

```
C:\Users\HP\Documents>hugo new site test-hugo
Congratulations! Your new Hugo site is created in C:\Users\HP\Documents\test-hug
o.

Just a few more steps and you're ready to go:

1. Download a theme into the same-named folder.
   Choose a theme from https://themes.gohugo.io/, or
   create your own with the "hugo new theme <THEMENAME>" command.
2. Perhaps you want to add some content. You can add single files
   with "hugo new <SECTIONNAME>\<FILENAME>.<FORMAT>".
3. Start the built-in live server via "hugo server".

Visit https://gohugo.io/ for quickstart guide and full documentation.

C:\Users\HP\Documents>cd test-hugo

C:\Users\HP\Documents\test-hugo>hugo
←[K25lBuilding sites . ←[?25h
                   | EN
+------------------+----+
  Pages            |  3
  Paginator pages  |  0
  Non-page files   |  0
  Static files     |  0
  Processed images |  0
  Aliases          |  0
  Sitemaps         |  1
  Cleaned          |  0

Total in 66 ms
```

PR #5384, opened 2018-10-31
Fixes ANSI character output regression on Windows

LorenzCK:
Fixes a regression introduced in Hugo version 0.50 by commit df021317a964a482cd1cd579de5a12d50faf0d08, related to issue #4462.

Previously this was fixed in 16b1a4148ad05a40cd748e373b5dc18447285837 by replacing the ANSI characters in `commands/helpers.go` with empty strings. This fix instead updates package `common/terminal`, thus disabling all ANSI character output on Windows (including output colorization).


common/terminal/colors_others.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+	"os"
+
+	isatty "github.com/mattn/go-isatty"
+)
+
+// IsTerminal return true if the file descriptor is terminal and the TERM
+// environment variable isn't a dumb one.
+func IsTerminal(f *os.File) bool {
+	fd := f.Fd()
```
bep: Just return false if `runtime.GOOS == "windows", which is a smaller and better change.

LorenzCK: OK, changes applied. Please squash if this is OK 😊

common/terminal/colors.go
```diff
@@ -52,6 +57,10 @@ func Warning(s string) string {
 
 // colorize s in color.
 func colorize(s, color string) string {
+	if runtime.GOOS == "windows" {
```
bep: This check is superflous. The `IsTerminal` is always called first. We will eventually try to get "some Windows support", so having the check in one place only would make that easier.

LorenzCK: OK, done.

bep:
Merged and thanks! For future PRs, see my slightly modified commit message (we have some guidelines somewhere).

LorenzCK:
Great, thanks @bep, will check out the guidelines

PR #5396, opened 2018-11-03
[WIP] Allow "Jsonify" the option to pretty print JSON

### LINKED ISSUE #5040: Allow "jsonify" to pretty print JSON

SeanPrashad:
Fixes #5040: Allow "Jsonify" to pretty print JSON

This PR allows Jsonify the option to return pretty-printed output, where the spacing is specified.

### DIFF
--- a/tpl/encoding/encoding_test.go
+++ b/tpl/encoding/encoding_test.go
@@ -85,18 +85,21 @@ func TestJsonify(t *testing.T) {
 	ns := New()
 
 	for i, test := range []struct {
-		v      interface{}
-		expect interface{}
+		v       interface{}
+		spacing string
+		expect  interface{}
 	}{
-		{[]string{"a", "b"}, template.HTML(`["a","b"]`)},
-		{tstNoStringer{}, template.HTML("{}")},
-		{nil, template.HTML("null")},
+		{[]string{"a", "b"}, nil, template.HTML("[\n\"a\",\n\"b\"\n]")},
+		{[]string{"a", "b"}, "", template.HTML("[\n\"a\",\n\"b\"\n]")},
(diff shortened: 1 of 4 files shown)


SeanPrashad:
As Go doesn't have optional parameters, I modified `Jsonify` to be a [variadic function](https://gobyexample.com/variadic-functions), however I'm not sure if this is the most elegant and maintainable way to do it.

Any and all feedback greatly appreciated!

tpl/encoding/encoding_test.go
```diff
@@ -85,18 +85,21 @@ func TestJsonify(t *testing.T) {
...
-		v      interface{}
-		expect interface{}
+		v       interface{}
+		spacing string
+		expect  interface{}
 	}{
-		{[]string{"a", "b"}, template.HTML(`["a","b"]`)},
-		{tstNoStringer{}, template.HTML("{}")},
-		{nil, template.HTML("null")},
+		{[]string{"a", "b"}, nil, template.HTML("[\n\"a\",\n\"b\"\n]")},
```
SeanPrashad: This test should represent the current use case when a user does not specify the spacing **at all** .

bep:
>Any and all feedback greatly appreciated!

Yes. String options are not optimal -- if we would get more options in the future ...

As I see it it is either that or a `map` (which you could then say `"asdf" | jsonify (dict "delimiter" "   ")`

This is how I did it for `resources.ToCSS` etc. with many options.

bep:
>Thanks for the feedback @bep - just wanted to let you know that I'm hoping to get another crack at this in the coming weeks after exams for school are done!

Great! Good luck with your exams!

Issue #5078, opened 2018-08-15
Use page resources for OpenGraph and Twitter Card images


tashian:
For my site I'm using page resources to transform and display cover images for my stories. It's working super well.

But the internal OpenGraph and Twitter Card templates use the "images" param:

https://github.com/gohugoio/hugo/blob/488631fe0abc3667355345c7eb98ba7a2204deb5/tpl/tplimpl/template_embedded.go#L194

Page resources seem like the ideal way to do OpenGraph and Twitter meta tags, too. Does it make sense to migrate the internal templates?

For now I have created my own partials. The image tag looks like this:

    {{ with (.Resources.ByType "image").Match "cover" }}{{ range first 6 . }}
      <meta property="og:image" content="{{ .RelPermalink | absURL }}" />
    {{ end }}{{ end }}

This depends on me having a "cover" resource, but a more generic version of this would just choose the first image resource on the page.

I think the only challenge here is making the change in a non-breaking way.

MichaelHierweck:
Would this be fixed by #5157?

max-arnold:
I believe that right now Twitter card template already does what you want, and indeed #5157 will bring the same behavior to the OpenGraph template. The lookup order is:

1. Explicit `images` param in a frontmatter
2. Page bundle images which match `*cover*` or `*thumbnail*`
3. Global `Site.Params.images`

spech66:
Seems like the Schema template is missing this too.

max-arnold:
@spech66 I'll try to submit a PR with something similar for `schema.html` in a few days.

max-arnold:
@spech66 Done!

TomPichaud:
Could you also update the OpenGraph template please ?

It would be nice to get some kind of cohesion between these 3 templates.

max-arnold:
@TomPichaud I did exactly that! The #5378 will bring this behavior to OpenGraph template, and #5415 to Schema.

Hopefully, someone from maintainers will merge these PRs.

MichaelHierweck:
Please get this PR merged. (I really don't understand why this bug/PR is still open.)

robshep:
Please merge this.  I've locally patched for now.

For me it's not convenience, but to workaround a bug. 

`absURL` from `images` **doesn't** work on page-bundled images. 

`.Permalink` from `resources` **does** work on page-bundled images

hacdias:
This would be really nice to have,

/cc @bep

max-arnold:
[...]

PR #5415, opened 2018-11-07
Add support for featured and global image to Schema template

max-arnold:
Make schema.html image tag behave in the same way as twitter_card.html. Now it can pick an image like *cover* or *thumbnail* from page bundle, plus also respects .Site.Params.images

Should nicely complement the #5378, as requested in https://github.com/gohugoio/hugo/issues/5078#issuecomment-434188426


max-arnold:
Bump

max-arnold:
Bump

PR #5276, opened 2018-10-02
Render markup in figure sc caption and attr

kaushalmodi:
1. First commit cleans up the shortcode to make it more readable, and ready it for the second commit. This commit also updates the tests to pass for the white-space changes made in this commit.
2. Second commit renders the Markdown in `caption` and `attr` parameters of the `figure` shortcode. 2 test cases are added for this change too. None of the existing tests (after the prior whitespace fix commit) needed to be modified for this second commit.

Fixes https://github.com/gohugoio/hugo/issues/4406.

---

I found these 13 tests to fail locally before/after building with this PR:

```
--- FAIL: TestPageWithMoreTagOnlySummary (0.43s)
--- FAIL: TestSummaryWithHTMLTagsOnNextLine (0.34s)
--- FAIL: TestWordCountWithIsCJKLanguageFalse (0.48s)
--- FAIL: TestWordCount (0.53s)
--- FAIL: TestWordCountWithMainEnglishWithCJKRunes (0.53s)
--- FAIL: TestWordCountWithAllCJKRunesWithoutHasCJKLanguage (0.62s)
--- FAIL: TestWordCountWithAllCJKRunesHasCJKLanguage (0.67s)
--- FAIL: TestPageWithMoreTag (0.54s)
--- FAIL: TestPageWithEmbeddedScriptTag (0.65s)
--- FAIL: TestPageWithShortCodeInSummary (0.66s)
--- FAIL: TestPageWithDateFields (0.62s)
--- FAIL: TestPageWithDelimiter (0.59s)
--- FAIL: TestCreateNewPage (0.48s)
--- FAIL: TestGetJSON (4.00s)
````

PR #5449, opened 2018-11-18
Fix figure whitespace issue, introduced in PR #5276

max-arnold:
As a result of #5276, there is no space between the caption and the attribution link of the `figure` shortcode. I.e. this:

`<p>Image by<a href="http://example.com">Author</a></p>`

renders in browser as:

`Image byAuthor`.

This patch should fix it.


thewebmastercom:
I think this is the wrong approach.

You have deleted this line: 

`{{- with .Get "attrlink" -}}`

Instead of doing that, try changing that line to:

`{{- with .Get "attrlink" }}`

When I did that, the issue resolved itself. Maybe you can verify that on your end @max-arnold ?

PR #5452, opened 2018-11-20
[RFC] - Dump template path to page if rendering to memory

raghur:
for ticket - #5453 

For now tests etc do not work but I'm just starting this so that I can get some inputs.

### DIFF
--- a/tpl/template.go
+++ b/tpl/template.go
@@ -135,6 +135,7 @@ type TemplateAdapter struct {
 }
 
 var baseOfRe = regexp.MustCompile("template: (.*?):")
+var IncludeTemplateRef = false
 
 func extractBaseOf(err string) string {
 	m := baseOfRe.FindStringSubmatch(err)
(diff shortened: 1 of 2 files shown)


tpl/template.go
```diff
@@ -125,9 +126,20 @@ func (t *TemplateAdapter) Execute(w io.Writer, data interface{}) (execErr error)
 		defer t.Metrics.MeasureSince(t.Name(), time.Now())
 	}
 
+	templatePath := ""
+	if IncludeTemplateRef {
+		if !strings.HasPrefix(t.Name(), "_internal") && t.Name() != "alias" {
```
raghur: If I enter the block when template name is `alias`, then `fileAndFilename` panics on the stat line. Any help appreciated.

bep:
How does this handle base templates (i.e. `baseof.html`)?

raghur:
Is that one of the `_internal` templates? If so, right now, those are all skipped altogether.

PR #5474, opened 2018-11-27
tpl/path: Add tests

moorereason:
Coverage of path.go goes from 2.6% to 100%.


bep:
There are some Windows test failures.

moorereason:
Fixed.

PR #5483, opened 2018-11-28
Add reflect template namespace

moorereason:


tpl/reflect/reflect_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+func TestIsSlice(t *testing.T) {
+	for i, test := range []struct {
+		v      interface{}
+		expect interface{}
+	}{
+		{[]int{1, 2}, true},
+		{[]tstNoStringer{{}, {}}, true},
+		// failures
+		{nil, false},
```
bep: I have only one comment to this PR. The failures section:

1. It's not failures. There is no error involved.
2. It looks copy-pasted from somewhere else. I think it's fair to assume that Go's reflect package is well tested, so any tests in here should either test the integration and possibly also document the usage. So I would say that 1 slice (true) and 2 falses (nil + some other) would be enough and easier to understand. As it is now, it looks way more complex than it is. This comment also relates to the map variant.

moorereason: PTAL

PR #5521, opened 2018-12-13
POC: implement starlight plugins

natefinch:
This is a proof of concept of how hugo could leverage [starlight](https://github.com/starlight-go/starlight) to run plugins written in the [starlark](https://github.com/google/starlark-go) python dialect.

Example:

Template:
```
{{plugins.Starlight "hello.star" "page" .Page}}
```

plugins/hello.star:
```
output("hello " + page.Title)
```

### DIFF
--- a/tpl/plugins/init.go
+++ b/tpl/plugins/init.go
@@ -30,12 +30,7 @@ func init() {
 			Context: func(args ...interface{}) interface{} { return ctx },
 		}
 
-		ns.AddMethodMapping(ctx.Starlight,
-			[]string{"starlight"},
-			[][2]string{
-				{`{{plugins.starlight "uppercase.star" .Page.Title }}`, `Blockhead`},
-			},
-		)
+		ns.AddMethodMapping(ctx.Starlight, nil, nil)
 		return ns
 	}
 
--- a/tpl/plugins/init.go
+++ b/tpl/plugins/init.go
@@ -1,0 +1,43 @@
+// Copyright 2018 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package plugins implements plugins of various sorts for Hugo templates.
+package plugins
+
+import (
+	"github.com/gohugoio/hugo/deps"
+	"github.com/gohugoio/hugo/tpl/internal"
+)
+
+const name = "plugins"
+
+func init() {
+	f := func(d *deps.Deps) *internal.TemplateFuncsNamespace {
+		ctx := New(d)
+
+		ns := &internal.TemplateFuncsNamespace{
+			Name:    name,
+			Context: func(args ...interface{}) interface{} { return ctx },
+		}
+
+		ns.AddMethodMapping(ctx.Starlight,
+			[]string{"starlight"},
+			[][2]string{
+				{`{{plugins.starlight "uppercase.star" .Page.Title }}`, `Blockhead`},
+			},
+		)
+		return ns
+	}
+
+	internal.AddTemplateFuncsNamespace(f)
+}
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ dist
 
 
 vendor
+
+# visual studio code settings files
+.vscode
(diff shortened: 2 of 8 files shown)


natefinch:
```
template: test:1:2: executing "test" at <plugins>: can't evaluate field starlight in type interface {}
```

So, I must be missing something.  I thought that if I set "starlight" as an alias, that `plugins.starlight` would work... but only `plugins.Starlight` works.  So now I'm not sure how to use the alias field....

natefinch:
I know this needs tests, but I wanted to put it up as a proof of concept to see if the concept is acceptable.  It's usable as-is, I tried it out with some simple scripts.

natefinch:
There's a POC hugo site to test it with here: https://github.com/natefinch/hugo-starlight

tpl/plugins/init.go
```diff
@@ -0,0 +1,43 @@
...
+
+		ns := &internal.TemplateFuncsNamespace{
+			Name:    name,
+			Context: func(args ...interface{}) interface{} { return ctx },
+		}
+
+		ns.AddMethodMapping(ctx.Starlight,
+			[]string{"starlight"},
+			[][2]string{
+				{`{{plugins.starlight "uppercase.star" .Page.Title }}`, `Blockhead`},
```
bep: Should be: `plugins.Starlight` (which should get the build go green)

tpl/plugins/init.go
```diff
@@ -0,0 +1,43 @@
...
+	f := func(d *deps.Deps) *internal.TemplateFuncsNamespace {
+		ctx := New(d)
+
+		ns := &internal.TemplateFuncsNamespace{
+			Name:    name,
+			Context: func(args ...interface{}) interface{} { return ctx },
+		}
+
+		ns.AddMethodMapping(ctx.Starlight,
+			[]string{"starlight"},
```
natefinch: I'm unclear what the alias here is for, if it's not to make `plugins.starlight` work in a template...

bep: The alias is there to make `starlight` work in a template. As in `where` vs `collections.Where` etc.

natefinch: oh I see... hmm ok.  What's your opinion?  I think I like keeping plugins namespaced, since they have code outside of the template, being more visible and searchable is nice.

tpl/plugins/init.go
bep: Just replace the alias slice with nil.

tpl/plugins/init.go
natefinch: btw, I'm assuming here that other plugin schemes will at some point exist, otherwise we could just call it "plugin" ... but I presume someone's going to want to do their own lua thing or whatever.

natefinch:
Say you want to display a sorted list of author display names  for an article.  I'm not entirely sure that's even possible in pure Go template code, but it's trivial in imperative code.

Even if something is possible in template code, It's hard to reuse template code that just manipulates data, rather than producing text.  Like producing a sorted list of display names from a map of authors.  Maybe different pages want to display those names differently, but the code to get that list could be the same.

Real code mixed in can make your templates much easier to read and understand.

```
{{$displayNames := plugins.Starlight "displayNames.star" "authors" page.Authors}}
```

^^ That's a lot better than 20 lines of gnarly template code.

natefinch:
Yes, exactly.  It's for writing your own template functions.  The nice thing about that is that it takes the responsibility off of Hugo to implement every function in existence, and allows a theme author to write very specific functions for their particular need. 

The code for the above template could be implemented in starlark code like this:

```python
authors = page.Authors()
display_names = [authors[k].DisplayName for k in authors]
display_names = sorted(display_names)
output(display_names)
```

I'm not sure you could even do it in go templates without some very narrowly focused template functions.

bep:
>I'm not sure you could even do it in go templates without some very narrowly focused template functions.

I'm pretty sure your current example is possible as-is, and I don't see how a "sort function" should be considered very narrowly focused.

The thing is, when I think about "plugins in Hugo", this area isn't what's first popping into my my head (this is discussed elsewhere, so I'm not going into those defails). 

This PR has a high degree of overlap with what we have ("a plugin is ... almost like a partial,  but it's in Python and not Go template syntax ..."), and I don't think the arguments I have seen adds enough weight to warrant the added complexity/maintainance/documentation/support (we would now have to teach people Python in the forum) ...

We will eventually (hopefully very soon) add some  kind of "build hook plugins", and I think it would be much easier to add something like this PR **after** that. It would be really unfortunate if we merged this PR and then decide we need to go with Lua or something for the other stuff.

moorereason:
This PoC is really helpful for showing a way to solve the custom template functions problem, so thank you for the PR.  I also like the fact that you are able to return non-textual results from such a function (if I'm reading the PR correctly, you're example would return a slice).

But I also agree with @bep.  We need to agree upon an overall design and how custom template functions may fit into that design.  I feel like we can implement just about anything we dream up, but it needs to be well thought out or we'll regret it.

Again, thanks for this PoC, @natefinch.  👍

natefinch:
Ok, so I just looked at the currently available functions, and you're right.  Because @moorereason already effectively built an entire programming language in hugo's template functions, this is perhaps somewhat less necessary than I thought. 

I do still think custom template functions are valuable, but I also agree that it needs to be part of an overarching strategy, and if you decide to do lua for another part, you don't want to have to support two built-in plugin languages.

moorereason:
Related #5510

PR #5535, opened 2018-12-17
Retrieve GitInfo from contentDir

### LINKED ISSUE #5533: Support GitInfo for content folders in different Git repositories
`GitInfo` provides metadata about content files, but from my observations and reading of the code, it actually searches the working directory for a git repo, *not* the content directory.

As a result, if content and theme/configuration are kept in separate repos (including when content is a submodule), Hugo fails to find any git metadata for the content files because it's querying the wrong repo.

Because `GitInfo` is for information about content, not theme/build configuration etc., Hugo should search for a git repo from the content directory, not the working directory, which is generally where the configuration and themes are found, and may or may not be in the same repo.

This should make `GitInfo` behave correctly for anyone who keeps their content in a separate repo, while continuing to work as before for mono-repos.

deanishe:
Get metadata for `GitInfo` from `contentDir` instead of `workingDir`.
This decouples the `GitInfo` feature from the working directory, making
it dependent on configuration, not the directory `hugo` is run in.

Rationale
---------

This patch aligns the `GitInfo` feature more closely with the way the
rest of Hugo works. The standard `cd /my/site && hugo` build idiom also
works when content isn't in the same repo as site, and it allows the
`GitInfo` feature to be used regardless of which directory `hugo` is run
from (when configured correctly), as is the case for other features.

Currently, Hugo searches the working directory for the git repo to
populate `GitInfo` content metadata. This means that, in order to
retrieve the correct metadata, `hugo` must be run from within the repo
containing the content.

If `hugo` isn't run from within a repo, `--enableGitInfo` causes a
repo-not-found build error, even if the content *is* in a repo. And if
`hugo` is run in a different repo, say if content is a submodule of
site, it will find no/bad data, as it's looking in the wrong repo.

This restriction is unique to `GitInfo` (AFAIK). If you aren't using
that feature, you can run Hugo in any directory you want, as long as you
pass the appropriate flags.

This patch makes it possible to use the standard `cd /my/site && hugo`
build process also when the content is in a different repo.

More generally, it makes it possible to run `hugo` from any directory
when using `--enableGitInfo`, as you can when not using it.

Effects
-------

`GitInfo` currently only works correctly if `workingDir` and
`contentDir` are in the same repo, so it will continue to work
everywhere it already works, particularly in the normal case of having
site and content in a single repo.

This patch will also make `GitInfo` Just Work ^(TM) when content is in a
separate repo to the site. Currently, using `GitInfo` requires `cd`-ing
to the content repo and running `hugo` from there with a bunch of flags.

If the content is unversioned, this patch will cause Hugo to *always*
fail to build the site when run with `--enableGitInfo`.

Currently, because Hugo looks in the working directory, not the content
directory, a build with unversioned content can succeed when Hugo is
called with `--enableGitInfo` as long as Hugo is *run* within a git
repo. Hugo won't find any (correct) file metadata, but missing file
metadata don't cause a build error, whereas a missing git repo does.

Fixes #5533

### DIFF
--- a/docs/content/en/variables/git.md
+++ b/docs/content/en/variables/git.md
@@ -25,7 +25,7 @@ Hugo's Git integrations should be fairly performant but *can* increase your buil
 
 ## `.GitInfo` Prerequisites
 
-1. The Hugo site must be in a Git-enabled directory.
+1. The global `contentDir` must be in a Git-enabled directory. Any language-specific `contentDir` must be in the same repo as the global `contentDir`.
 2. The Git executable must be installed and in your system `PATH`.
 3. The `.GitInfo` feature must be enabled in your Hugo project by passing `--enableGitInfo` flag on the command line or by setting `enableGitInfo` to `true` in your [site's configuration file][configuration].
 
--- a/hugolib/gitinfo.go
+++ b/hugolib/gitinfo.go
@@ -36,9 +36,12 @@ func (g *gitInfo) forPage(p page.Page) *gitmap.GitInfo {
 }
 
 func newGitInfo(cfg config.Provider) (*gitInfo, error) {
-	workingDir := cfg.GetString("workingDir")
+	contentDir := cfg.GetString("contentDir")
+	if !filepath.IsAbs(contentDir) {
+		contentDir = filepath.Join(cfg.GetString("workingDir"), contentDir)
+	}
 
-	gitRepo, err := gitmap.Map(workingDir, "")
+	gitRepo, err := gitmap.Map(contentDir, "")
 	if err != nil {
 		return nil, err
 	}


bep:
There are test failures.

deanishe:
> There are test failures.

Working on them.

> Also note that there may be more than one `contentDir`.

Yes, and I suppose it's improper to assume that they're all in the same repo. Would it be an acceptable solution to move `hugolib.gitInfo` from `hugolib.HugoSites` to `hugolib.Site`?

bep:
>Would it be an acceptable solution to move hugolib.gitInfo from hugolib.HugoSites to hugolib.Site?

I'll assume that the normal use case would have to have a "mono rep" with content folders. So, if I have a big site with 10 languages, having to do a "git log" and parse the same Git log 10 times does not sound optimal.

deanishe:
> So, if I have a big site with 10 languages, having to do a "git log" and parse the same Git log 10 times does not sound optimal.

No, indeed it doesn't.

As I see it, that leaves the following options:

1. Make GitInfo dependent on the global `contentDir` and require any language-specific `contentDir` to be in that same repo (which is the current state of the PR).
2. Require all content to be in the same repo, but prefer a language-specific `contentDir` to the global one (i.e. no need to set the global `contentDir` when you're not using it).
3. Make `hugolib.gitInfo` capable of handling multiple roots. This would presumably require rejiggering GitInfo lookups to use absolute paths.
4. Ensure that any repo is only queried once. I believe `bep/gitmap` is currently responsible for actually finding the repo, and doesn't separate that step from loading the metadata, so that would mean changing `bep/gitmap` or incorporating repo resolution into Hugo.

In terms of my particular use case, 1 works just dandy. In terms of correctness, 3 or 4 strike me as the optimal options, but I have *no idea* whether one-repo-per-language is a thing. (Keeping content separate, so authors/editors can't break the build system, or so content can be managed via Dropbox etc. has very obvious advantages. Keeping English separate from German not so much…)

I'm happy to try to implement whichever solution you prefer. Your project, your choice. I'd just like to be able to use the standard `cd /my/hugo/site && hugo` build method when my content is in a different repo without having to maintain my own fork forever.

deanishe:
> Will I receive any surprises related to this in my current mono repo projects (hugoDocs and friends?) that uses this feature?

No. If `contentDir` is a subdirectory, `gitInfo` still finds the same repo. I've tested it with my own monolingual mono-repo site, and also with `gohugoio/hugoDocs`.

The only breakage I've found so far is if you're running `hugo` with `--enableGitInfo` against content that's not actually in any repo. Currently, as long as `workingDir` is within *some* repo, Hugo looks there for metadata and doesn't complain that the content is actually unversioned. My PR turns that into a "no git repo found" build error.

In any case, I would be very grateful for any feedback from folks with more "exotic" setups.

bep:
I'm closing this in favour of #6848

Issue #5436, opened 2018-11-13
"# more" does not work in org-mode in 0.51


fliar:
same issue of #5381 , but not fixed in 0.51

I've posted a test org file in #5381 , sorry for duplicating but I cannot reopen the old one

bep:
```bash
---
title: "Test"
date: 2018-11-09T23:02:52+08:00
draft: true
---

this is summary

# more

this is content
```

The above is not "ORG mode front matter". Note that I have fixed/patched/tested this ORG mode format without any existing tests nor spec. And I can say that the above is not something that I plan to support, even if it has worked before. If someone wants to implement this with proper tests, they are welcome.

kaushalmodi:
I am traveling, so won't be able to look at this in detail until 3 weeks from now. 

@bep the example you pasted looks alright. It's not clear if that's what you support now or is just a proposal. 

If it's a proposal, I'd suggest having `#+hugo: more` (#+hugo` is case insensitive as per Org mode spec) instead `# more`.

bep:
>If it's a proposal

It's a failing Org Mode test case (whatever Org Mode is). I just leave this issue open for anyone interested.

PR #5545, opened 2018-12-20
Improve org mode support

niklasfasching:
As the title says, this PR improves the Org mode support of hugo. For a basic overview of the changes, check out [this page](https://niklasfasching.github.io/go-org/go-org-vs-goorgeous) - it showcases the html output of `goorgeous` (left, currently included with hugo) and `go-org` (right, this PR) for the same input. The css is adapted to `go-org`, so open up dev tools for a more fair comparison.

For an overview of the open issues of `goorgeous` that are fixed in `go-org` take a look [here](https://niklasfasching.github.io/go-org/#misc.org)

If you want to try out any Org mode snippets, you can do so in your browser [here](https://niklasfasching.github.io/go-org/).

Also take a look at the [discussion](https://discourse.gohugo.io/t/improved-org-mode-support/15859/18) for more information.

I included a second commit in this PR that adds the Org mode front matter to the main content as Org mode itself also uses the front matter for configuration - the impact of this is much smaller (as go-org only uses the `TODO` front matter variable right now - so we could move that to a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ Hugo stands on the shoulder of many great open source libraries, in lexical orde
  | [github.com/bep/debounce](https://github.com/bep/debounce) |    MIT License |
  | [github.com/bep/gitmap](https://github.com/bep/gitmap) |  MIT License |
  | [github.com/bep/go-tocss](https://github.com/bep/go-tocss) | MIT License |
- | [github.com/chaseadamsio/goorgeous](https://github.com/chaseadamsio/goorgeous) | MIT License |
+ | [github.com/niklasfasching/go-org](https://github.com/niklasfasching/go-org) | MIT License |
  | [github.com/cpuguy83/go-md2man](https://github.com/cpuguy83/go-md2man) | MIT License |
  | [github.com/danwakefield/fnmatch](https://github.com/danwakefield/fnmatch) | BSD 2-Clause "Simplified" License |
  | [github.com/disintegration/imaging](https://github.com/disintegration/imaging) |  MIT License |
--- a/docs/content/en/content-management/formats.md
+++ b/docs/content/en/content-management/formats.md
@@ -19,7 +19,7 @@ toc: true
 
 **Markdown is the main content format** and comes in two flavours:  The excellent [Blackfriday project][blackfriday] (name your files `*.md` or set `markup = "markdown"` in front matter) or its fork [Mmark][mmark] (name your files `*.mmark` or set `markup = "mmark"` in front matter), both very fast markdown engines written in Go.
 
-For Emacs users, [goorgeous](https://github.com/chaseadamsio/goorgeous) provides built-in native support for Org-mode  (name your files `*.org` or set `markup = "org"` in front matter)
+For Emacs users, [go-org](https://github.com/niklasfasching/go-org) provides built-in native support for Org-mode  (name your files `*.org` or set `markup = "org"` in front matter)
 
 But in many situations, plain HTML is what you want. Just name your files with `.html` or `.htm` extension inside your content folder. Note that if you want your HTML files to have a layout, they need front matter. It can be empty, but it has to be there:
 
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/bep/debounce v1.2.0
 	github.com/bep/gitmap v1.1.0
 	github.com/bep/go-tocss v0.6.0
-	github.com/chaseadamsio/goorgeous v1.1.0
 	github.com/cpuguy83/go-md2man v1.0.8 // indirect
 	github.com/disintegration/imaging v1.6.0
 	github.com/dustin/go-humanize v1.0.0
@@ -38,6 +37,7 @@ require (
 	github.com/muesli/smartcrop v0.0.0-20180228075044-f6ebaa786a12
 	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646 // indirect
 	github.com/nicksnyder/go-i18n v1.10.0
+	github.com/niklasfasching/go-org v0.0.0-20190112190817-da99094e202f
 	github.com/olekukonko/tablewriter v0.0.0-20180506121414-d4647c9c7a84
 	github.com/pelletier/go-toml v1.4.0 // indirect
 	github.com/pkg/errors v0.8.1
@@ -66,7 +66,5 @@ require (
 	google.golang.org/genproto v0.0.0-20190522204451-c2c4e71fbf69 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 )
-
-exclude github.com/chaseadamsio/goorgeous v2.0.0+incompatible
 
 replace github.com/markbates/inflect => github.com/markbates/inflect v0.0.0-20171215194931-a12c3aec81a6
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -27,8 +27,8 @@ import (
 	"unicode/utf8"
 
 	"github.com/gohugoio/hugo/common/maps"
+	"github.com/niklasfasching/go-org/org"
 
-	"github.com/chaseadamsio/goorgeous"
 	bp "github.com/gohugoio/hugo/bufferpool"
 	"github.com/gohugoio/hugo/config"
 	"github.com/miekg/mmark"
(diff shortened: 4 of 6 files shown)


parser/pageparser/pagelexer_intro.go
```diff
@@ -155,7 +155,13 @@ LOOP:
 		}
 	}
 
+	// The front matter that was just lexed is in a format native to Org mode - it is,
+	// contrary to other document and front matter formats, part of the main section.
+	// l.emit advances the offset (start) of the lexer (and thus excludes the front matter
+	// from the main section) so we have to reset it.
+	frontMatterStart := l.start
```
bep: This doesn't look right. I don't understand the comment, but even if I did: This will produce more content than it consumes, and I'm fairly confident that this can be solved without this workaround.

niklasfasching: I'll try to my goal, maybe you can point me into a direction:
I want hugo to send the complete file contents to go-org - not just what hugo calls the MainSection part after the front matter.

---- 
The reason for this is that the front matter possibly contains information that is important for the rendering as `#+FOO: bar` is the [settings format of the org-mode parser](https://orgmode.org/manual/In_002dbuffer-settings.html).

An example:
```
#+TITLE: Foo
#+TODO: TODO DONE CUSTOM

content [...]
```
 
Hugo currently only sends  
```
content [...]
```
to the renderer. However, the `#+TODO: TODO DONE CUSTOM` is actually configuration for the Org mode renderer - `go-org` would use it to update the list of keywords that are recognized as TODO keywords in headlines - see [the headline with CUSTOM todo status example](https://niklasfasching.github.io/go-org/#headlines.org).

bep: OK; I think I understand. What you get today is a processed version of the content (after front matter). If you also need the front matter, I suggest that you add that to the `RenderingContext`. You can easily get to that with the parse tree you have.

bep: You could store away the start/end positions for front matter here:

https://github.com/gohugoio/hugo/blob/master/hugolib/page_content.go#L92

We currently don't use it.

bep: >So I guess the problem this change is trying to solve is that the front matter of an org file actually serves two purposes: (1) configure hugo (2) configure the Org mode parser.

Why not send the unmarshaled frontmatter?

https://github.com/gohugoio/hugo/blob/master/hugolib/page_content.go#L94

niklasfasching: Thanks! this sounds like it will get more involved than a 2 line change. 

As the benefit of this is pretty small for now (there's not much to be configured in go-org for now) i suggest we split this into a separate PR and just focus on org mode support itself for now.

bep: >As the benefit of this is pretty small for now

Agree. And Hugo's API in this area is also rather clumsy. I have been planning on redoing that for a while ... Bu time.

bep: Also note that If you need some "configuration for the rendererd", we should add that as configuration in Hugo which you could receive in some constructor.

niklasfasching: > Also note that If you need some "configuration for the rendererd", we should add that as configuration in Hugo which you could receive in some constructor.

Maybe I shouldn't have called it configuration for the renderer. It's not configuration in the sense that it should be passed to the constructor. Rather, we currently remove part of the org document because we conflate hugo configuration (front matter) and org mode configuration (buffer settings) by using the same syntax (`#+KEY: VALUE`), location (top of the file) and namespace for both of them. So... hugo remove front matter from the document to be rendered because that's the right thing - the problem with org is that the front matter is not just front matter - it's also part of the actual document.

Coming back to the point before
> Why not send the unmarshaled frontmatter?
I like the idea because it sounds like a good way to provide go-org with the full document without polluting the hugo code base with org specific stuff - having the front matter available sounds like it could be useful for other things as well.

I hope that makes sense...

niklasfasching:
Updated with support for `.TableOfContents`
Please let me know if there's anything else that should be done - I'll try to keep this branch rebased in the meanwhile.

kaushalmodi:
I'm traveling at the moment. I'll test out this branch once I get back home (Jan 3).

iNecas:
I've started exploring hugo as a platform for my blog. As Emacs user, the native support for org mode was definitely +1 when selecting the tool for the job. I've already hit limitations of `goorgeous` (such as missing support for _definition lists_).

Not that it would matter much, but I've checkout this branch and rendering of org-mode work just fine for me: looking forward for hugo having an active maintenance of the org-mode backend again.

kaushalmodi:
> Good point! I didn't know about org-lint, thx :).
> It doesn't seem to result in any errors

Thanks for checking that.

> Also Org mode expects at least some keywords to contain square brackets. see `org-element-dual-keywords`

And I did not know that. Good to know :)

---

### About testing go-org

As I wasn't a Goorgeous user, I did not have any handy posts in Org format. But I know that @zamansky uses it and blogs often. So I built `hugo` locally using this PR branch and ran it on his repo (https://github.com/cestlaz/hugo-blog-generator). I got just this one warning:

> go-org: Could not parse token org.token{kind:"beginBlock", lvl:0, content:"COMMENT", matches:[]string{"#+BEGIN_COMMENT", "", "COMMENT", ""}}: Falling back to treating it as plain text.

Looks like you just need to support `#+begin_comment` .. `#+end_comment`.

Awesome job!

kaushalmodi:
> Hugo treats lines starting with #+ as Org mode front matter (link). It would be more robust to check whether the line is a keyword with a regexp like this but that adds quite a bit of complexity I'm not sure is worth it.

I would have thought that `#+begin_comment` .. `#+end_comment` would be treated similar to how you would treat `#+begin_src ..` .. `#+end_src`, and such. Is that not the case?

niklasfasching:
I'm not sure if I understand you correctly, could you expand on that?

Just to make sure there's no misunderstanding: What you quote was about how `hugo` treats lines starting with `#+`, not about `go-org`.

The warning in question is generated by a [file](https://github.com/cestlaz/hugo-blog-generator/blob/3170b7aa4e0ff82a0572d49bdcc991235e15b863/content/posts/2015-05-19-code-org-college-board.html.org) starting with 
```
#+BEGIN_COMMENT
[...]
#+END_COMMENT
```
Hugo thinks the first line is front matter (as it starts with `#+`) and removes it from the content of the file.
`go-org` receives just 
```
[...]
#+END_COMMENT
```
As there's no `#+BEGIN_COMMENT` line, it logs a warning about an out of place `#+END_COMMENT` line.

kaushalmodi:
> I'm not sure if I understand you correctly, could you expand on that?

> Maybe there's a misunderstanding: What you quote was about how hugo treats lines starting with #+, not about what go-org does.

Yes, I quoted correctly. Sorry for not being clear.

I meant to ask what `hugo` does with `#+begin_src` blocks (because those also begin with `#+` but are not confused as keywords). How do hugo and go-org interact with those blocks so that the code blocks are rendered correctly? Can the same be achieved for `#+begin_comment` blocks?

niklasfasching:
Thx, got it now :).

Hugo treats the `#+BEGIN_COMMENT` line as front matter because it starts with `#+` **and** is at the top of the file (i.e. there's no non-empty lines that don't start with `#+` ahead of it). 
A `#+BEGIN_SRC` line at the top of the file would also be treated as front matter.

It's just about where in the file the line is, nothing specific about the `#+BEGIN_COMMENT`.

kaushalmodi:
> It's just about where in the file the line is, nothing specific about the #+BEGIN_COMMENT.

Understood. Thanks. That means that `#+begin_comment` mid post should work as expected then?

This is a separate issue then for hugo.. it should consider `#+..` as Org keywords only if `#+` is not followed by `#+begin_` or `#+end_` (case-insensitive). That is not 100% accurate, but that will cover a lot of cases.

niklasfasching:
> Understood. Thanks. That means that #+begin_comment mid post should work as expected then?
> This is a separate issue then for hugo
:+1: to both

Thanks for taking the time to review this :)

PS: The last force push fixes the front matter logging & updates `go-org` (no functional changes, just refactoring)

kaushalmodi:
> This is a separate issue then for hugo.. it should consider #+.. as Org keywords only if #+ is not followed by #+begin_ or #+end_ (case-insensitive). That is not 100% accurate, but that will cover a lot of cases.

You are officially the most qualified person knowing both Go and Org. Do you think you can tackle fixing that in `hugo`?

niklasfasching:
> Do you think you can tackle fixing that in hugo in a separate PR?
I have it on my todo list (along with #5436) :).

kaushalmodi:
@bep Do you have further comments on this PR?

kaushalmodi:
@bep Another tiny nudge to review this.

niklasfasching:
@bep is there anything specific blocking this? i can totally understand if this just doesn't have priority, just stumbled across this again by chance :)

bep:
@niklasfasching really sorry about this. I was reminded about this when I got some version issues with that other library.

niklasfasching:
No worries, thanks for all your work on hugo :)

Issue #4717, opened 2018-05-12
Indented `highlight` shortcode brings in the indentation as part of the code block


kaushalmodi:
- When using the highlight shortcode for code blocks in lists, the indentation needed to have the code blocks indented gets added to the code block itself. 
- This does not happen when using code fences. 

So the bug seems to be localized to Hugo and not Blackfriday. 

Here is a test page demonstrating the bug:

- [HTML][1] | [Markdown][2]

In the same test page above, the [first section][3] has the same code blocks, but done using the code fences, where that extra indentation issue is not seen.

Another effect of not removing that indentation is that the indentation before the closing `{{< /highlight >}}` creates an extra blank line in the final HTML.

[1]: https://ox-hugo.scripter.co/test/posts/source-block-indented/#code-blocks-in-list-using-highlight-shortcode
[2]: https://raw.githubusercontent.com/kaushalmodi/ox-hugo/master/test/site/content/posts/source-block-indented.md
[3]: https://ox-hugo.scripter.co/test/posts/source-block-indented/#code-blocks-in-list-using-code-fences

kaushalmodi:
Updated the tests showing the bug with smaller/targeted test files.

kaushalmodi:
@bep can you please tag this as a bug (and hopefully see if an easy fix is possible for this?

kaushalmodi:
Can this be tagged as a Bug please? This issue prevents me from using `highlight` shortcode (and thus the line number prefixing using Chroma) in lists.

bep:
Your shortcode variant shows the hightlighting as a table with line numbers, you can adjust the indentation with CSS, I guess, but there will be indentation ...

```
 {{< highlight emacs-lisp "linenos=table, linenostart=1" >}}
````

If you remove the "linenos=table", that should fix it.

kaushalmodi:
> you can adjust the indentation with CSS

The indentation is due to extra spaces in the pre block. And also the closing tag of the highlight shortcode introduces a blank line.

The indentation also increases with each nested list item. So it gets very tricky to fix with CSS (also not sure if it would be possible because then I need to make the CSS rules apply only on the badly indented blocks created by highlight shortcode and not by the fenced blocks).

> If you remove the "linenos=table", that should fix it.

But I need to use the line numbers and so I am using the highlight shortcode. I use the fenced code blocks when I don't need line numbers.

---

[...]

PR #5553, opened 2018-12-21
Fix highlight sc so that it removes shortcode offset from inner code

kaushalmodi:
Fixes https://github.com/gohugoio/hugo/issues/4717.


kaushalmodi:
@bep When you get a chance please review this PR. 

- It passes all the existing tests.
- It passes the new test added specifically to test to fix. I have verified locally that if I revert the fix, this new test starts failing.

bep:
@kaushalmodi I appreciate the effort, but

1. I need to understand this problem, which is a post-Christmas activity
2. We cannot have logic at this level in the shortcode templates. If so, that would have required a different level of test coverage

In the mean time, I suggest you put this shortcode inside your own project(s).

PR #5572, opened 2018-12-28
tpl: Fix hugo package name and add godocs

moorereason:


tpl/hugo/init.go
```diff
@@ -11,7 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package site
+// Package hugo provides template functions for accessing the Site Hugo object.
```
bep: What is a "Site Hugo" object?

I'm not sure what the correct term is, but that isn't. And this doesn't really "provide template functions accessing ..."

tpl/reflect/init.go
```diff
@@ -11,6 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package reflect provides template functions for run-time object reflection.
```
bep: We need to do better than that; what functionality does the package provide?

tpl/site/init.go
```diff
@@ -11,6 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package site provides template functions for accessing the Site object.
```
bep: Does it?

tpl/hugo/init.go
moorereason: How about `Package hugo provides easy access to Hugo version information.`?

I'm working on generating template namespace docs for the docs repo, and I'm using the package godoc as the namespace description.

tpl/site/init.go
moorereason: Does it not?

tpl/reflect/init.go
moorereason: `Package reflect provides template functions for identifying object types.`?

Offer a better description.

tpl/hugo/init.go
moorereason: BTW, the docs generation is a one-time event right now.  I don't intend to update the docs site continually with what I'm doing.  Just writing a bash script to do the heaving lifting.

PR #5591, opened 2019-01-07
Use subtests with server_test.go

tryzniak:
Use Golang's subtests to provide a convenient way to run specific tests.
Example:
go test -run=TestFixURL/Basic_production

If you find this one useful, I could go through the codebase and find some other places to add subtests.

### DIFF
--- a/commands/server_test.go
+++ b/commands/server_test.go
@@ -97,21 +97,23 @@ func TestFixURL(t *testing.T) {
 		{"No config", "", "", true, 1313, "//localhost:1313/"},
 	}
 
-	for i, test := range tests {
-		b := newCommandsBuilder()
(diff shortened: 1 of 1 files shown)


bep:
Hey, and thanks for your interest in Hugo.

I will merge this, but a general update of the tests to use subtests will have to wait. There are several reasons for this:

1. I'm in the middle of a bigger refactoring of the `hugolib` package, and any major parallel work will make the merge impossible.
2. I find sub tests usual when something fail (which should be rare) and I need to debug. But doing a massive "test update only" will also touch the Git history (why was this test modified?) which I value more. So I welcome test updates of this sort for the right reasons (as part of an update to that part of the code), but not for test tests alone.

Issue #4868, opened 2018-06-22
commands: Allow string slices in --theme


jhabdas:
[Theme Components](https://gohugo.io/themes/theme-components/#readout) were introcuded in `0.42` with https://github.com/gohugoio/hugo/pull/4467. Running `hugo -h` on `0.42.1` provides the following output for the `-t` flag:

```
-t, --theme string               theme to use (located in /themes/THEMENAME/)
```

Should the CLI be updated to support theme components? If not, what is the suggested approach for programmatically building Hugo sites with theme components going forward?

jhabdas:
https://github.com/gohugoio/hugo/issues/4869

PR #5603, opened 2019-01-13
Make hugo server -t work again

### LINKED ISSUE #5061: HUGO_THEME succeeds but "-t" fails when config.toml contains theme = ["../.."]
I'm not entirely sure it is a Hugo bug at all because it looks like a rare corner case which manifests only when config.toml has a buggy `theme` entry, as described below.

But HUGO_THEME and `-t` being inconsistent here seems worthy of investigation, though admittedly a low-priority one.

### Details

I came across a perculiar error (tested with Hugo 0.46) when building the Hugo themes site locally with the alpha-church theme after I reverted the `HUGO_THEME` workaround back to using the `-t` flag (see https://github.com/gohugoio/hugoThemes/commit/0b4e98b96f0ee4aee3fd22bc8a7a5e65869d255f):

```
 ==== PROCESSING  alpha-church  ====== 
Building site for theme alpha-church using its own exampleSite to ../themeSite/static/theme/alpha-church/
hugo --quiet -s exampleSite2 -d ../themeSite/static/theme/alpha-church/ -t alpha-church -b http://localhost:1313/theme/alpha-church/
ERROR 2018/08/11 10:22:56 Unable to locate template for shortcode "osmap" in page "about.md"
ERROR 2018/08/11 10:22:56 Unable to locate template for shortcode "contact_form" in page "contact.md"
Error: Error building site: logged 2 error(s)
FAILED to create exampleSite for alpha-church
```

However, if I add `HUGO_THEME=alpha-church`, the "Unable to locate template for shortcode" errors go away.

I then noticed in https://github.com/funkydan2/alpha-church/blob/master/exampleSite/config.toml#L4
a somewhat strange `theme = ["../.."]` entry (`themesDir` not defined in config.toml):

```ini
baseURL = "http://example.org/"
languageCode = "en-us"
title = "Alpha Church"
theme = ["../.."]

# etc.
```

It looks like an erroneous use of the `theme` variable, and I suppose it should be corrected to something like this:

```ini
theme = "alpha-church"
themesDir = "../.."
```

But HUGO_THEME and `-t` being inconsistent here seems worthy of investigation, though admittedly a low-priority one.

### LINKED ISSUE #5569: hugo server -t still doesn't work
Running `hugo server -t <a theme>` still doesn't work.

There's several related Issues, some open and some closed:
- https://github.com/gohugoio/hugo/issues/4851
- https://github.com/gohugoio/hugo/issues/4868

This issue goes back to 10 months ago when support for theme config was added.

I haven't been able to track down exactly where the issue is myself, but I did notice that here (https://github.com/gohugoio/hugo/blob/3f3187de0f62107da19d9341aebd1d8414bff0ea/hugolib/config.go#L469) is where the theme passed by the CLI gets lost. If i disable that line, it works as expected though the tests fail.

tryzniak:
This commit solves an issue where hugo would ignore the cli -t flag
and only use a theme defined in config.toml.

Also allow -t flag to accept a string slice.

Closes #5569
Closes #5061
Related #4868

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -250,7 +250,7 @@ func (cc *hugoBuilderCommon) handleFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP("cacheDir", "", "", "filesystem path to cache directory. Defaults: $TMPDIR/hugo_cache/")
 	cmd.Flags().BoolP("ignoreCache", "", false, "ignores the cache directory")
 	cmd.Flags().StringP("destination", "d", "", "filesystem path to write files to")
-	cmd.Flags().StringP("theme", "t", "", "theme to use (located in /themes/THEMENAME/)")
+	cmd.Flags().StringSliceP("theme", "t", []string{}, "themes to use (located in /themes/THEMENAME/)")
 	cmd.Flags().StringP("themesDir", "", "", "filesystem path to themes directory")
 	cmd.Flags().StringVarP(&cc.baseURL, "baseURL", "b", "", "hostname (and path) to the root, e.g. http://spf13.com/")
 	cmd.Flags().Bool("enableGitInfo", false, "add Git revision, date and author info to the pages")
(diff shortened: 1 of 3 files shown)


bep:
Thanks a lot @tryzniak I know I was the one who most likely introduced this issue, but I'm glad someone else stepped up to the plate and ... just fixed it.

PR #5616, opened 2019-01-18
Support minify setting in config files

### LINKED ISSUE #5142: Make --minify able to be set in config.toml
It would be nice to not have to specify it on the command line, but in a config file.

tryzniak:
Users can now use config files to enable minification.
For example with config.toml to enable minification simply add line
`minify = true` to it.

Running hugo with the `--minify` cli flag will overwrite a config
option. For example if config.toml has minification enabled, and
a user runs hugo with `--minify=false` then minification will be
disabled.

Closes #5142

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -214,33 +214,25 @@ func initializeFlags(cmd *cobra.Command, cfg config.Provider) {
 		"themesDir",
 		"verbose",
 		"verboseLog",
-	}
-
-	// Will set a value even if it is the default.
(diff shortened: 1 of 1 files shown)


bep:
See https://github.com/gohugoio/hugo/issues/5142#issuecomment-455863780

regisphilibert:
But with Hugo config becoming environment aware and one can now set minifiy true in its production config only, do you still stand by the mentioned comment?

PR #5627, opened 2019-01-20
Stop navigating to draft and future pages on edits with livereload

### LINKED ISSUE #5385: --navigateToChanged 404s on edit on drafts

tryzniak:
If hugo was run with
```bash
hugo server --navigateToChanged
```
and a user edited a draft or future page then hugo would navigate to it. Since
drafts don't have urls the navigation would be buggy, sometimes resulting in
404s.
This commit fixes it. Now, the navigation will only be performed only with
--buildDrafts or --buildFuture enabled, like so
```bash
hugo server --navigateToChanged -D
```

Closes #5385

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -981,11 +981,20 @@ func (c *commandeer) handleEvents(watcher *watcher.Batcher,
 					}
 				}
 
-				if p != nil {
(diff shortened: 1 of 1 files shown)


tryzniak:
Hmm, should we rebuild the site if we detect changes on drafts? I think it would be pointless unless hugo is run with --buildDrafts.

PR #5636, opened 2019-01-25
Fix OpenGraph image fallback to site params

statik:
In other parts of the opengraph internal template and in the twitter_cards internal template, the template falls back to using site-level parameters when page level parameters are not specified.

This change fixes the opengraph internal template to have this same behavior for the image meta tags.

I believe this was a simple bug, and with this change the behavior now matches what the documentation describes.

### DIFF
--- a/tpl/tplimpl/embedded/templates/opengraph.html
+++ b/tpl/tplimpl/embedded/templates/opengraph.html
@@ -2,7 +2,7 @@
 <meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
 <meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}" />
 <meta property="og:url" content="{{ .Permalink }}" />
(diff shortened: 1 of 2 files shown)


tpl/tplimpl/embedded/templates/opengraph.html
```diff
@@ -2,8 +2,15 @@
 <meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
 <meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}" />
 <meta property="og:url" content="{{ .Permalink }}" />
-{{- with .Params.images }}{{ range first 6 . }}
+{{ with $.Params.images }}
```
bep: The fix is correct and good, but I suspect this would be easier to understand (and shorter) if you used the `Param` method, e.g:

```
{{ with $.Param "images" }}
```

The above will first look in page, then in site.

statik: That is great, thank you! I changed it to use this way, much better.

gbmhunter:
Thanks for this fix!

PR #5648, opened 2019-01-29
[Incorrect fix?] tpl/data: Prevent getJSON and getCSV fetch failure from aborting build

### LINKED ISSUE #5643: getJSON fetch failure should not abort build
Hugo >= 0.50 (including master HEAD as of today) exits before finishing building a website when getJSON fails to get a resource.

This problem was earlier filed as Issue #5076 "getJSON etc. should log ERROR but return nil in failure situations" and was fixed in 0.49, but perhaps the "greatly improved error messages shown in browser" feature in 0.50 might have undone the fix in PR #5199.

anthonyfok:
This PR supposedly fixes #5643, but how that this PR is pushed, I become convinced that this is not the right fix because the context of the error is lost:

Before (0.53):

> Error: Error building site: failed to render pages: render of "home" failed: "/home/foka/web/example.ca/layouts/index.html:24:13": execute of template failed: template: index.html:24:13: executing "main" at <getJSON "https://api...>: error calling getJSON: failed to get getJSON resource "https://api.example.com/test?q=abc": Failed to retrieve remote file: Not Found

After (with this PR):

> Building sites … ERROR 2019/01/28 17:05:11 failed to get getJSON resource "https://api.example.com/cards/test?q=abc": Failed to retrieve remote file: Not Found

So, perhaps this PR is not the right fix?  More discussion at the issue #5643.

### DIFF
--- a/tpl/data/data.go
+++ b/tpl/data/data.go
@@ -83,7 +83,8 @@ func (ns *Namespace) GetCSV(sep string, urlParts ...string) (d [][]string, err e
 
 	err = ns.getResource(cache, unmarshal, req)
 	if err != nil {
-		return nil, _errors.Wrapf(err, "failed to read CSV resource %q", url)
+		ns.deps.Log.ERROR.Printf("Failed to get CSV resource %q: %s", url, err)
+		return nil, nil
 	}
 
 	return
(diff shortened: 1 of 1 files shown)


bep:
I suspect this issue was introduced as a side-effect of me trying to improve error messages. I guess this PR reverts that in this particular case, but we can revisit that later. This fix is fine, but you need to adjust the tests to make them pass. 

Note that I will probably do a patch release on Monday, so it would be good to get this in before then.

bep:
Merged.

anthonyfok:
> I suspect this issue was introduced as a side-effect of me trying to improve error messages. I guess this PR reverts that in this particular case, but we can revisit that later. This fix is fine, but you need to adjust the tests to make them pass.
> 
> Note that I will probably do a patch release on Monday, so it would be good to get this in before then.

Thank you so much for adjusting the tests for me in ddc6d4e30f282f23b703a3b42da552886062c8c8.  After reading the changes in that commit, I realize that I wouldn't have known how to fix these tests properly (due to me being still a very amateurish Go programmer and my lack of familiarity with much of Hugo's code), so it was wonderful that you just went ahead and fixed it.  And thank you for the new 0.54.0 release!  It is a very pleasant surprise!

pgorod:
@anthonyfok does this allow us to react to the error from our template? If so, how?

I'd like to output a proper result when the GetJSON doesn't return anything.

https://discourse.gohugo.io/t/handling-getjson-errors-in-shortcodes/18597/3

Thanks for any tips

PR #5649, opened 2019-01-29
Respect ignoreFiles config setting for static files

### LINKED ISSUE #1559: IgnoreFiles for static folder

### LINKED ISSUE #4874: ignoreFiles does not work as documented
I cannot get the example from the documentation of `ignoreFiles` to work in a brand new project.

Using Hugo 0.41 on macOS 10.11.6, I created a new site using `hugo new site testing` and added the following to `config.toml` [as per the documentation of `ignoreFiles`](http://gohugo.io/getting-started/configuration/#ignore-files-when-rendering):

```toml
ignoreFiles = [ "\\.foo$", "\\.boo$" ]
```

The documentation says that this should

> cause Hugo to ignore files ending with `.foo` and `.boo` when rendering

but when I create a file at the path `static/example.foo` and run `hugo`, it is copied to `public/example.foo`, i.e. not ignored at all.

It seems that either there is a bug in Hugo or the documentation must be updated.

stapelberg:
fixes #1559
fixes #4874

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 
 	"github.com/gohugoio/hugo/hugofs"
+	"github.com/gohugoio/hugo/source"
 
 	"github.com/gohugoio/hugo/resources/page"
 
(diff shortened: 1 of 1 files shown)


stapelberg:
The travis failure affects windows only and seems unrelated to my change.

stapelberg:
Friendly ping?

marrrcin:
This would be a nice feature to merge

vvasuki:
indeed- I don't want to have to write a script to manually delete ignored files. :-(

stapelberg:
@spf13 @bep friendly ping? :)

hartzell:
ping?

XhmikosR:
This is still valid, but needs a rebase

stapelberg:
> This is still valid, but needs a rebase

Rebased onto current master.

XhmikosR:
@bep friendly ping :)

bep:
To my knowledge, `ignoreFiles` has always been just `content` files.  I have updated the docs to reflect this. This means that we cannot repurpose that same setting to also include some other random files without breaking sites in the wild. Assuming I have this in my config:

```
ignoreFiles = [ "\\.jpg$" ]
``` 

I wanted to ignore /content/blog.jpg, but certainly not /static/images/cat.jpg or /assets/images/cat.jpg.

bep:
That said, I do understand the need for all of this, and I agree that doing this on this as a file system wrapper makes sense.

But I don't want us to build this on top of some legacy config when it doesn't match particularly well.

I'm doing some work adding Goldmark as the new default Markdown renderer now, and 90% of that work is to split what we have into packages.

So, what I suggest we do is to add a new config element called `files`:

```toml
[files]
[[files.ignore]]
fileSystem = "static" (or "all"?)
expressions = [ "\\.jpg$" ]
[files]
[[files.ignore]]
fileSystem = "content")
expressions = [ "\\.jpg$" ]
```

See https://github.com/gohugoio/hugo/blob/master/hugofs/files/classifier.go#L76 for the "file system classifiers".


And then decode that config into `files.Config` and use that when building the different filesystems.


https://github.com/gohugoio/hugo/blob/master/hugofs/files/classifier.go#L76

stapelberg:
I don’t have time to make these changes right now. Could someone else jump in and move this forward please?

ackerleytng:
Looking forward to having this merged!

XhmikosR:
This is still valid.

marrrcin:
I agree, this PR was opened on `29 Jan 2019`...

XhmikosR:
Still valid.

raspopov:
Any news?

jmooring:
This was resolved with https://github.com/gohugoio/hugo/pull/9059.

```
[[module.mounts]]
source = 'static'
target = 'static'
excludeFiles = '*.txt'
```

See <https://gohugo.io/hugo-modules/configuration/#module-config-mounts>

PR #5655, opened 2019-01-31
Fix #5646, ignore .swp files in config folder

### LINKED ISSUE #5646: Vim swap files cause hugo to fail
Opening configuration in vim causes hugo development server to fail with 
```
<...>config/_default/.menus.toml.swp" is not a valid configuration format
```

Possible solution would be to ignore hidden files.

fberrez:
fixes #5646

### DIFF
--- a/hugolib/config.go
+++ b/hugolib/config.go
@@ -283,11 +283,18 @@ func (l configLoader) loadConfigFromConfigDir(v *viper.Viper) ([]string, error) 
 				return nil
 			}
 
-			for _, filenamePart := range strings.Split(fi.Name(), ".") {
-				// Ignore .swp files
-				if filenamePart == "swp" {
-					return nil
+			// Verifies if the file extension is valid.
+			isValid := false
+			ext := strings.ToLower(filepath.Ext(fi.Name()))
+			for _, validExt := range []string{"toml", "yaml", "yml", "json"} {
+				if strings.Contains(ext, validExt) {
+					isValid = true
+					break
 				}
+			}
+
+			if !isValid {
+				return nil
 			}
 
 			name := helpers.Filename(filepath.Base(path))
--- a/hugolib/config.go
+++ b/hugolib/config.go
@@ -283,8 +283,14 @@ func (l configLoader) loadConfigFromConfigDir(v *viper.Viper) ([]string, error) 
 				return nil
 			}
 
-			name := helpers.Filename(filepath.Base(path))
+			for _, filenamePart := range strings.Split(fi.Name(), ".") {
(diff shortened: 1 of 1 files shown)


hugolib/config.go
```diff
@@ -283,8 +283,14 @@ func (l configLoader) loadConfigFromConfigDir(v *viper.Viper) ([]string, error)
 				return nil
 			}
 
-			name := helpers.Filename(filepath.Base(path))
+			for _, filenamePart := range strings.Split(fi.Name(), ".") {
+				// Ignore .swp files
+				if filenamePart == "swp" {
```
kaushalmodi: I don't use vim, but should that check be case insensitive?

fberrez: Editing a file with Vim generates a `.swp` file. It works as a [buffer](https://vi.stackexchange.com/questions/177/what-is-the-purpose-of-swap-files).
That check does not need to be case sensitive since the generation of these files is automatic.

But your comment showed me that something is incorrect here: the `swp` is the file extension, we only need the test with the last part of the filename. Also, the .swp files are hidden files.

Let me correct it.

bep: I understand the problem, but the fix is sub-optimal.

* We have a list of config file extensions we support (lower case, see another PR that pulls this list into a variable)
* Every file that is not in this list can be safely ignored

fberrez: Here is my config folder:
```sh
config
├── _default
│   ├── config.toml
│   ├── .config.toml.swp # while editing config.tom with Vim
```

Here is the error message: 
```sh
Error: "/Users/fberrez/workspace/src/github.com/fberrez/hugotest/config/_default/.config.toml.swp:1:1": "/Users/fberrez/workspace/src/github.com/fberrez/hugotest/config/_default/.config.toml.swp" is not a valid configuration format
```

According to your answer, the `validConfigFilenames` slice containing all valid config file extensions would ignore the `swp` file. But, it returns an error.

Therefore, we have to change the checker behavior, no?

bep: >Therefore, we have to change the checker behavior, no?

 Yes, you have to change. But there is a fundamental difference between "if ext == swp then continue" an "if ext not in valid list then continue"

tryzniak:
It would be nice if this change was merged in after #5657

bep:
I fixed this in https://github.com/gohugoio/hugo/commit/3244cb3b31f8f8c39d9dfa82bc01fb2d6db59257 -- this PR looks correct, but I had to get it ready to ship in the near future patch release, with tests and all.

Thanks!

PR #5657, opened 2019-01-31
Store supported config formats in a variable

tryzniak:
Reduces duplication and defines a single source of truth for config formats.

### DIFF
--- a/config/configLoader.go
+++ b/config/configLoader.go
@@ -20,6 +20,10 @@ import (
 	"github.com/spf13/viper"
 )
 
+var (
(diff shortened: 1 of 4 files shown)


config/configLoader.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/spf13/viper"
 )
 
+var (
+	SupportedFormats = []string{"toml", "yaml", "yml", "json"}
```
bep: The change itself is good, but I think it would be more correct if this variable said, perhaps,`ValidConfigFileExtensions` -- esp. with the duplicate YAML definition.

tryzniak:
For some reason windows build has failed.

PR #5663, opened 2019-01-31
Hosting with AWS Amplify

swaminator:
Adding AWS Amplify hosting to the docs.


bep:

Abby3017:
In docs you have mentioned, **amplify will automatically detect hugo configuration** and it will build. But its not happening, it showing error that config file is not there( my config file is in config dir). When i am trying to pass this info through arg flag to `hugo` command, then its showing hugo command not found. Why dockerfile have fixed version of 0.51, anything can be done to upgrade (not issue to raise here, just saying). Any suggestion to make it work will be helpful.

davidsneighbour:
Reading the conversation above this note probably fit's in on [gohugoio/hugoDocs](https://github.com/gohugoio). 
See https://github.com/gohugoio/hugoDocs/pull/736.

Abby3017:
Thank you for pointing this out. But this has nothing to do with my problem. I can build my site with config.toml kept at root folder. But it is failing when I was using config folder to save my config.toml for production and staging server.

Issue #1687, opened 2015-12-08
Asciidoc TOC stripped in 0.15


jbarrez:
See discussion thread here: https://discuss.gohugo.io/t/asciidoc-toc-stripped-in-0-15/2231

In 0.15, the TOC (table of con tent) of an asciidoc file is stripped away, while in 0.14 this wasn't the case. This is caused by this commit: https://github.com/spf13/hugo/commit/b23b546a30df792ad2c8f0cf3ec1be30c53ba34d. Would it be possible to make this hardcoded setting configurable?

jbarrez:
Duplicates in a way https://github.com/spf13/hugo/issues/1685. However, that particular issue about docs.

anthonyfok:
I took a first look at the problem, and found some interesting twists.

My first question is: The table of content (TOC) is generated within `<body>`, not `<head>`, and shouldn't be stripped by `--no-header-footer`, which keeps everything inside the `<body>` intact.  Then, what is going on?

It turns out that AsciiDoc does not generate the TOC statically, but rather uses JavaScript to generates the TOC on-the-fly!  With the `--no-header-footer` option, the related JavaScript is not loaded, and 

I am happy to report, however, that Asciidoctor is not affected by this issue: Asciidoctor generates the full TOC as static HTML regardless of the `--no-header-footer` setting.  According to http://asciidoctor.org/docs/asciidoc-asciidoctor-diffs/#table-of-contents-toc:

> * Asciidoctor reifies the ToC in the header of the document instead of relying on JavaScript to create it

So, while we look at how to fix this issue #1687 in Hugo code, could you please install Asciidoctor and give it a try and see using Asciidoctor instead of AsciiDoc solves your problem?  :-)

Thanks!
Anthony

jbarrez:
Hi Anthony, thanks for digging into this.

We are using asciidoctor, however we were using 1.5.2. Switching to 1.5.3 (and Hugo 0.15), I could see the behaviour you described above: the ToC gets created in the HTML. Had to do some css changes to get it properly on the screen, but all now works as before.

Sorry for your time spent on this - I should have checked if the asciidoctor version was the problem here before thinking it was a Hugo problem.

anthonyfok:
[...]

PR #5672, opened 2019-02-03
Add support for native Go implementation of AsciiDoc by using libasciidoc

### LINKED ISSUE #1435: Add support for native Go implementation of AsciiDoc
Starting with v0.14 (released in May 2015), Hugo provides preliminary support for AsciiDoc with the help of an external helper such as `asciidoc` (Python-base) and `asciidoctor` (Ruby-based).

Please see Issue #470 "Support for AsciiDoc" (which led to the current experimental "external helper" implementation) for more information.  This issue 

However, the ultimate goal is for Hugo to support a native Go implementation of AsciiDoc, something that e.g. the https://github.com/VonC/asciidocgo project set out to accomplish.  This issue will track the progress on that front.

odknt:
Close #1435

Render asciidoc by using [libasciidoc](https://github.com/bytesparadise/libasciidoc) if `useLibasciidoc` is true.


moorereason:
We're aware of the `libasciidoc` project, but I don't think it's ready for production yet.

Cc: @xcoulon

bep:
0.1 may be a little early, yes -- but @xcoulon has been very busy lately. I'm very impressed by his recent work. I'm not familiar with AsciiDoc, so I cannot judge when this has become "good enough", but I can say this:

* My thought has always that the current AsciiDoc handling is something temporary until something native happens. So, when we do add this to Hugo, the native implementation will be the default, with a flag to use "the old".

xcoulon:
thanks for this PR @odknt, it was actually on my TODO list for a little later, once libasciidoc would have a better support of the Asciidoc syntax. And the plan is actually to use a feature toggle (as you did) to make sure that it's set at an "experimental" level for a bit (while Asciidoctor would remain the default runtime to process the files).

Thanks for the feedback @bep! I hope I can get more features soon (as time allows), and we should definitely discuss about the info/metadata that libasciidoc should return upon processing. I'm thinking about the doc title, author and ToC for instance, but there may be also other pieces of info that it should return to integrate nicely in Hugo.

odknt:
@bep @moorereason @xcoulon Thanks for checking.

May be @moorereason is right. I found issue related asciidoc in hugo and libasciidoc.
I may need to fix them ahead of this PR.

* `{{ .TableOfContents }}` not working in asciidoc (#1687).
* libasciidoc in panic rendering TOC if no sections.

Is there a possibility to be merge this PR?
In that case I want to add commit that use `parser` and `renderer/html5` instead of `libasciidoc.ConvertToHTML()` for TOC support.

xcoulon:
> * libasciidoc in panic rendering TOC if no sections.

@odknt if you found a bug in libasciidoc, would you mind opening an issue with a sample content so I can reproduce and fix the bug, please? I probably missed an edge-case with regards to the ToC ;)

bep:
>Is there a possibility to be merge this PR?

This will have to wait. Mostly to wait for libasciidoc to be a little more mature (it looks really promising). That, and I'm currently working on some restructuring of the relevant code, which will make this integration simpler when this is ready. But thanks for your contribution.

odknt:
OK, thanks.

xcoulon:
> > * libasciidoc in panic rendering TOC if no sections.
> 
> @odknt if you found a bug in libasciidoc, would you mind opening an issue with a sample content so I can reproduce and fix the bug, please? I probably missed an edge-case with regards to the ToC ;)

The ToC bug in libasciidoc was fixed in https://github.com/bytesparadise/libasciidoc/issues/284

PR #5676, opened 2019-02-04
Support nested keys/fields with missing values with the `where` function

### LINKED ISSUE #5416: Improve where with missing values
See https://discourse.gohugo.io/t/get-regular-pages-matching-params/15084/11

### LINKED ISSUE #5637: Allow nested fields in where clause
Currently the following:

```
{{ $pages := where .Site.RegularPages ".Params.author.name" "John" }}
```
Returns an error. 

Maybe this is similar enough to #5522 that it can be treated in the same timeframe/by the same person.

tryzniak:
Before this commit `where` would produce an error and bail building the
site. Now, `where` simply skips an element of a collection and does not
add it to the final result.

Closes #5637
Closes #5416


bep:
We need to redo the error handling in this area, but that will have to wait. Thanks for this.

PR #5686, opened 2019-02-09
WIP Support fetching remotes files with resources.Get

### LINKED ISSUE #5255: Add remote support in resources.Get
It'll be nice to have a `getCSS` function for external URL where the caching are not in our hands and also if we want to make the CSS an inline version instead of adding a link to the external site. Maybe by doing something like:

```
{{ $externalCSS := .getCSS "https://fonts.googleapis.com/css?family=Space+Mono:400,700" }}
<style>
     {{ $externalCSS.Content | safeCSS }}
</style>
```

tryzniak:
It'll let developers use inline CSS as in the example below:

```
{{ $externalCSS := resources.Get "https://fonts.googleapis.com/css?family=Space+Mono:400,700" }}
<style>
     {{ $externalCSS.Content | safeCSS }} Or .RelPermalink would also work.
</style>
```

Closes #5255

### DIFF
--- a/hugolib/resource_chain_test.go
+++ b/hugolib/resource_chain_test.go
@@ -14,6 +14,9 @@
 package hugolib
 
 import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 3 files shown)


bep:
I need to come back to you on this a little later, as I'm a little swamped with work at the moment (also Hugo related). 

The code looks good, I have 2 general comments:

1. I don't think it is right to extend a very specific "read from a filesystem" to also do remote work.  I suggest adding a `client.GetRemote` method and do the isRemote checking one layer up.

2. I have not tested your version, but from what I read, I don't think the MIME type resolution is correct. This may not be important in the simple use case, but is a must if you want to use the resource in further pipes (minify, concat, toCSS etc.).  This is currently detected by the suffix in the filename. So doing MD5 is good as long as it ends up as ZZZZZZ.css etc. if at all possible.

3. I'm a little curious/uncertain about how this ends up in the file cache and how that can be configured. My memory about that implementation is a little hazy, will have to revisit.

4. We need proper testing. I was a little overwhelmed by the amount of work in creating the initial Hugo Pipes implementation, so I went the simple way of adding some integration tests for these in `hugolib`. I think we need package scoped tests for this, but that requires some thinking/work.

tryzniak:
Awesome considerations! I'll need some time to digest those!

atishay:
@tryzniak Do you have any plans to get this merged soon (into the latest build with all tests passing?)

jmooring:
Superseded by https://github.com/gohugoio/hugo/pull/9185

PR #5690, opened 2019-02-10
content_renderer: support mmark code-fence captions

### LINKED ISSUE #4129: pygmentsCodeFences breaks mmark code captioning
Settings pygmentsCodeFences to true breaks mmark code caption functionality. 

Hugo version: 0.31

config.yaml
```
pygmentsCodeFences: true
```

example.mmark
~~~
---
markup: mmark
slug: example
title: Example
---

```bash
highlight bash source code
```
Figure: some bash source
~~~

Expected result: code section with highlighting and code caption.
Actual result: code section with highlighting, missing the code caption.

The caption will be generated when removing the language (`bash` in the example above), but, in turn, it removes the highlighting.

jschaf:
Mmark supports captions on code-fences like:

    ~~~ python
    import this
    ~~~
    Figure: this is a caption

Add hugo support by copying some of the render code from Mmark. Ideally, we'd reuse the MMark BlockCode function, but that would change how code is rendered in HTML and is a breaking change.

There's two more MMark code-fence features that aren't supported by this PR:

1. Callouts
2. Manually specifying a code prefix through an attribute

Fixes #4129

### DIFF
--- a/helpers/content_renderer.go
+++ b/helpers/content_renderer.go
@@ -96,13 +96,32 @@ type HugoMmarkHTMLRenderer struct {
 }
 
 // BlockCode renders a given text as a block of code.
-// Pygments is used if it is setup to handle code fences.
+// Pygments is used if it is setup to handle code fences with support for captions.
 func (r *HugoMmarkHTMLRenderer) BlockCode(out *bytes.Buffer, text []byte, lang string, caption []byte, subfigure bool, callouts bool) {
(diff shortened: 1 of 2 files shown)


jschaf:
Anything I can do to make this easier to review?

bep:
The MMark version we currently use is deprecated/replaced with something new. Until we sort out that situation, I don't want to add new stuff in that particular department. I'm sure it's working, but people would shout high and loud if we, for some reason, had to remove this in the future. For that reason, it is much easier not adding it in the first place.

PR #5703, opened 2019-02-21
List future dates in CSV format

### LINKED ISSUE #5610: Add flag for showing publish date with hugo list future
I'd like to build tooling to automatically publish future dated posts once they become current posts. `hugo list future` prints a list of all future dated posts, but not their publish dates. Would a PR be accepted that added a flag to include the publish dates?

danielcompton:
Prints future dated posts as CSV, including the future date the post will be available from.

Prints like this:

```
$ hugo list future
episodes/19.md,2019-04-09T07:46:51+13:00
```

Things I'm not sure about:

* Should I somehow be writing the CSV through jww? I wasn't sure how to do that here.
* Do you want me to change the output format for the other list commands? Especially `expired`, that seems like a similar kind of case. Not sure what you'd want to do for `draft`, I suppose it could be useful to see the draft dates listed too (note that Hugo doesn't return future dated drafts when listing either drafts or future).

Fixes #5610

### DIFF
--- a/commands/list.go
+++ b/commands/list.go
@@ -14,7 +14,10 @@
 package commands
 
 import (
+	"encoding/csv"
+	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/gohugoio/hugo/hugolib"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


moorereason:
> Should I somehow be writing the CSV through jww? I wasn't sure how to do that here.

AFAIK, jww doesn't provide `io.Writer` interfaces for this use case, so `os.Stdout` is fine here.

> Do you want me to change the output format for the other list commands? Especially `expired` ... `draft`

Adding to `expire` seems logical.

Adding to `draft` makes less sense.  I'm sure people won't know at first glance what date that is.  Is it the pubdate?  The file Last Modified date?

bep:
>s it the pubdate? The file Last Modified date?

It is the `.ExpiryDate()`.

I agree that adding that to the "list expired" command would make this nice and symmetrical.

danielcompton:
I've updated the PR to print for future and expired listings. Ready for review.

PR #5705, opened 2019-02-21
Fix #5685. Convert numeric values to float64 and compare them

### LINKED ISSUE #5685: Number types don't work properly in JSON data files
I have a simple data file like this:

userdata.json
```json
[
	{
		"name": "John doe",
		"salary": 300
	},
	{
		"name": "Frank foobar",
		"salary": 500
	},
	{
		"name": "Rick foobar",
		"salary": 400
	}
]
```
and in my layout I'm doing

```go
{{where .Site.Data.userdata "salary" "<" 500 }}
```
The output is `[]` but the expected output is `[map[salary:300 name:John doe] map[name:Rick foobar salary:400]]`

If I have the same thing as a `yaml` file

```yaml
---
- name: John doe
  salary: 300
- name: Frank foobar
  salary: 500
- name: Rick foobar
  salary: 400
```

It does give me the expected output of `[map[salary:300 name:John doe] map[name:Rick foobar salary:400]]`

I'm not quite proficient in Hugo or go html/templates, so excuse me if this issue has been posted already, but I couldn't find it with my current knowledge.

`>hugo version` output

Hugo Static Site Generator v0.54.0-B1A82C61 windows/amd64 BuildDate: 2019-02-01T09:42:02Z

tryzniak:
Hi. We just need check here if both values are numbers, convert them to float64, and then compare.

Fix #5685

### DIFF
--- a/tpl/collections/where.go
+++ b/tpl/collections/where.go
@@ -114,6 +114,17 @@ func (ns *Namespace) checkCondition(v, mv reflect.Value, op string) (bool, error
 			slv = v.Interface()
 			slmv = mv.Interface()
 		}
+	} else if isNumber(v.Kind()) && isNumber(mv.Kind()) {
+		fv, err := toFloat(v)
+		if err != nil {
+			return false, err
+		}
+		fvp = &fv
+		fmv, err := toFloat(mv)
+		if err != nil {
+			return false, err
+		}
+		fmvp = &fmv
 	} else {
 		if mv.Kind() != reflect.Array && mv.Kind() != reflect.Slice {
 			return false, nil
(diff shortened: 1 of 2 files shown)


tpl/collections/where.go
```diff
@@ -114,6 +114,11 @@ func (ns *Namespace) checkCondition(v, mv reflect.Value, op string) (bool, error
 			slv = v.Interface()
 			slmv = mv.Interface()
 		}
+	} else if isNumber(v.Kind()) && isNumber(mv.Kind()) {
+		fv, _ := toFloat(v)
+		fvp = &fv
+		fmv, _ := toFloat(mv)
```
moorereason: Please handle all error values

tryzniak: Oh yeah, missed that one.

PR #5713, opened 2019-02-23
Fix some lint errors

tryzniak:
I noticed they're some lint errors, so I fixed them.


bep:
Thanks. First: These are not lint errors, they are suggestions. Second: I'm in the process of a larger refactoring at the moment, which means that for the next week or so, changes that will make that merges hard must be important. So, this PR will have to wait for a couple of weeks.

PR #5720, opened 2019-02-27
doc: improve dates i18n workaround

gdeshors:
The existing one worked only when you had one language. My proposal works with several languages.


bep:

PR #5725, opened 2019-03-01
Feature/image watermark

### LINKED ISSUE #4595: Add image watermarking
All issues related to image watermarking are closed  as native image processing comes to Hugo.
Unfortunatly in image feature set I lack of image watermarking (with test or image)
As far as I can see there is a function in library used by Hugo https://gowalker.org/github.com/disintegration/imaging#Overlay

JeanDaniel-Fischer:
Add a Watermark function to Image Resources.

Add a Watermark function to Image Resource that use an image from the
static directory as watermark. It is used the same way that Fit/Fill/Resize function.
A simple example: 
```
{{ $uploadPath := .Site.GetPage "/uploads" }}
{{ range $uploadPath.Resources.Match "towatermark.jpg" }}
        <img src="{{ (.Watermark "topleft static/watermark.png").Permalink }}" />
{{ end }}
```
It supports the anchor to determine where to put the watermark in the image.

Fixes #4595 

This is my first PR, don't hesitate to guide me if something is missing.

### DIFF
--- a/resources/image.go
+++ b/resources/image.go
@@ -184,6 +184,24 @@ func (i *Image) Fill(spec string) (*Image, error) {
 	})
 }
 
+// Watermark image with a given image from static package.
+func (i *Image) Watermark(spec string ) (*Image, error) {
+	// TODO 1 - watermark custom image config 2 - load static 3 - output watermark 4 - chech gen image
+	return i.doWithImageConfig("watermark", spec, func(src image.Image, conf imageConfig) (image.Image, error) {
+		watermarkSrc, err := imaging.Open(conf.secondaryImagePath)
+		if err != nil {
+			return nil, err
+		}
+
+		offset := image.Pt(0, 0)
+		b := src.Bounds()
+		result := image.NewRGBA(b)
+		draw.Draw(result, b, src, image.ZP, draw.Src)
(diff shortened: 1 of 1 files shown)


bep:
Thanks for this. One of the hardest think in Hugo has been to get file reading consistent (I'm not saying where totally clear) and working on all platforms (also Windows) and be fairily easy to use. So, I don't think we can introduce another way to load resources.

So, instead of this:



I would love if we could get to something like:

```bash
{{ $img := (resources.Get("images/sunset.jpg").Fit "500x300" }}
{{ $logo := (resources.Get("watermarks/logo.jpg").Fit "100x100" }}
{{ $result := $logo.Overlay $img "topleft" }}
```

Note that I have not thought too hard about method names or ordering in the above, but we should. With something like the above we would avoid adding any new file handling and we oculd more or less reuse the existing config for sizes and anchoring.

/cc @regisphilibert

regisphilibert:
I agree with Bjørn Erik. This is a great addition to Hugo's image capabilities but it should be consistent with the current way users manipulate those in Hugo. 
Oh and 👍 for `.Overlay`. It makes it more generic, as this PR brings more than just watermark potential.

JeanDaniel-Fischer:
Thanks for the feedback :) I will rename the method and look into using the image resource cache.

bep:
What is the current status of this PR?

jolestar:
What is the current status of this PR?

PR #5729, opened 2019-03-02
Update stretchr/testify to 1.3.0.

QuLogic:
No need for a random commit anymore now that there's a newly tagged version. This version also adds a Go module file.

I don't think I saw anything major in the diff from the used commit to new version: https://github.com/stretchr/testify/compare/04af85275a5c...v1.3.0


QuLogic:
Force-pushed to remove the old snapshot from `go.sum` as `go mod tidy` seems to want to do that.

bep:
Thanks for this.

Issue #5764, opened 2019-03-18
Support for ES6 JavaScript transpilation using Babel.


hmmmmniek:
Issue related to #5587 

Copied from my [post on the forum](https://discourse.gohugo.io/t/support-for-browser-compatible-javascript-compilation/17592):

> I am developing a documentation website for a JavaScript library.
On this website we offer live code examples which are written in ES6 JavaScript because this allows us to write very clean an easily understandable code.
However, ES6 is not a widely supported JavaScript version, especially IE11 does not, and will not support this.
> 
> Currently, we load Babel-standalone libraries and transpile the code examples client side as a temporary solution for better browser compatibility.
Of course, this is definitely not an ideal solution, these libraries are about 3.5Mb in size, and were never really meant to be used in this way.
> 
> A better solution would be to transpile the ES6 JavaScript code during website generation.
This feature will not only be useful for my very specific use case. But it would also allow anyone to write ES6 code for their hugo website, while keeping the website compatible with browsers that do not support ES6.
> 
> This feature will introduce a new dependencies:
> First, we need a library which allows us to interpret JavaScript code, goja seems like a good choice.
> - https://github.com/dop251/goja (MIT License)
> 
> Besides that, we will need a library which allows us to transpile ES6 code. Babel is a widely used JavaScript library which does this very well
> - https://babeljs.io/ (MIT License)
> 
> We will have to figure out how to properly include a JavaScript dependency into HUGO, but I’m sure I’ll figure something out.
> 
> My plan is to use goja to load the babel library. Then the babel library can transpile a JS code string passed by the HUGO user.
> For example, you’d use this function like this:
> ```
> {{ $jsCode := "[1, 2, 3].map(n => n ** 2);" }}
> {{ $jsCode | transpileJS "ES5" }}
> ```
> Which will produce the following browser-compatible output:
> ```
> [1, 2, 3].map(function (n) {
>    return Math.pow(n, 2);
> });
> ```
> I will make an attempt to implement this new feature somewhere in the next week.
I’d love to get some opinions on this, do you agree with this approach or do you feel like I’m making this unnecessarily complex?

I already got Goja working in a local build.
So now the last step is to somehow use to Babel CLI in the hugo project.

I noticed that we already [use the postcss CLI in HUGO](https://github.com/gohugoio/hugo/blob/master/resources/resource_transformers/postcss/postcss.go).

So currently my plan is to basically follow the example of how postcss is implemented, I will:
[...]

PR #5772, opened 2019-03-21
resources: Add JavaScript transpiling solution

hmmmmniek:
Add a new pipe called TranspileJS which uses the Babel cli. This makes it possible for users to write ES6 JavaScript code and transpile it to ES5 during website generation so that the code still works with older browser versions.

In the original issue i wrote that i planned on using a JavaScript interpreter. However I noticed that PostCSS uses a CLI to do its processing, and I realized that there is a Babel CLI as well. So I simply implemented this functionality in almost the exact same way that PostCSS is implemented.

The contributing guidelines tell me that I have to include test cases, which makes sense, however I am not sure where the correct place is in the project to do this.

See #5764

### DIFF
--- a/docs/content/en/hugo-pipes/transformjs.md
+++ b/docs/content/en/hugo-pipes/transformjs.md
@@ -1,0 +1,69 @@
+---
+title: TransformJS
+description: Hugo Pipes can process JS files with Babel.
+date: 2019-03-21
+publishdate: 2019-03-21
+lastmod: 2019-03-21
+categories: [asset management]
+keywords: []
+menu:
+  docs:
+    parent: "pipes"
+    weight: 75
+weight: 75
+sections_weight: 75
+draft: false
+---
+
+Any JavaScript resource file can be transpiled to another JavaScript version using `resources.TransformJS` which takes for argument the resource object and a slice of options listed below. TransformJS uses the [babel cli](https://babeljs.io/docs/en/babel-cli).
+
+
+{{% note %}}
+Hugo Pipe's TranspileJS requires the `@babel/cli` and `@babel/core` JavaScript packages to be installed in the environment (`npm install -g @babel/cli @babel/core`) along with any Babel plugin(s) or preset(s) used (e.g., `npm install -g @babel/preset-env`).
+
+If you are using the Hugo Snap package, Babel and plugin(s) need to be installed locally within your Hugo site directory, e.g., `npm install @babel/cli @babel/core` without the `-g` flag.
+{{% /note %}}
+### Options
+
+config [string]
+: Path to the Babel configuration file
+
+_If no configuration file is used:_
+
+plugins [string]
+: Comma seperated string of Babel plugins to use
+
+presets [string]
+: Comma seperated string of Babel presets to use
+
+minified [bool]
+: Save as much bytes as possible when printing
+
+noComments [bool]
+: Write comments to generated output (true by default)
+
+compact [string]
+: Do not include superfluous whitespace characters and line terminators (true/false/auto)
+
+verbose [bool]
+: Log everything
+
+### Examples
+Without a `.babelrc` file, you can simply pass the options like so:
+```go-html-template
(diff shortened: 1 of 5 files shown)


docs/content/en/hugo-pipes/transformjs.md
```diff
@@ -0,0 +1,69 @@
...
+: Path to the Babel configuration file
+
+_If no configuration file is used:_
+
+plugins [string]
+: Comma seperated string of Babel plugins to use
+
+presets [string]
+: Comma seperated string of Babel presets to use
+
```
westurner: s/seperated/separated/

snap/snapcraft.yaml
```diff
@@ -67,16 +67,20 @@ parts:
...
-    node-packages: [postcss-cli]
+    node-packages: [postcss-cli, @babel/cli]
     filesets:
       node:
         - bin/node
       postcss:
         - bin/postcss
         - lib/node_modules/postcss-cli/*
+      babel:
+        - bin/babel.js
```
danez: I'm not sure which `bin` folder this is and how it gets populated from node, but npm/yarn would install the babel binary as just `babel`  without `.js`.

docs/content/en/hugo-pipes/transpilejs.md
```diff
@@ -0,0 +1,69 @@
+---
+title: TranspileJS
```
danez: I personally would maybe name it by name similar to postcss like `babel`, because there are also other tools that do similar things. For example `bublè` (https://github.com/bublejs/buble). But babel is the de facto standard so it might also make sense to just go with `TranspileJS` and not worry about alternatives.
Not sure.

docs/content/en/hugo-pipes/transpilejs.md
```diff
@@ -0,0 +1,69 @@
...
+weight: 75
+sections_weight: 75
+draft: false
+---
+
+Any JavaScript resource file can be transpiled to another JavaScript version using `resources.TranspileJS` which takes for argument the resource object and a slice of options listed below. TranspileJS uses the [babel cli](https://babeljs.io/docs/en/babel-cli).
+
+
+{{% note %}}
+Hugo Pipe's TranspileJS requires the `@babel/cli` and `@babel/core` JavaScript packages to be installed in the environment (`npm install -g @babel/cli @babel/core`) along with any Babel plugin(s) or preset(s) used (e.g., `npm install -g @babel/preset-env`).
```
danez: I saw the same line in the `postcss` docs, but not sure if this makes sense here (but same applies for postcss). I would never recommend installing this globally and always recommend to install it locally in the project folder (which works btw.). Otherwise if multiple users work on the same repo, they might have different versions. Same if you deploy your website with some service I doubt they all have it installed and if so who knows which version.

But also not sure what the difference is between snap package and whatever this refers to. I use mac and netlify, so do I use snap?

bep: I agree as to the recommended best practice for documentation. But there are certain situations where you would want to use the global one (e.g. for my many ad-hoc test sites), but that is, I guess, a limitation of NPM. It puzzles me a little that they have not implemented some kind of global cache (like Go Modules does) to avoid me having to have a million copies of postcss etc. installed.

westurner:
@bep

> I agree as to the recommended best practice for documentation. But there
are certain situations where you would want to use the global one (e.g. for
my many ad-hoc test sites), but that is, I guess, a limitation of NPM. It
puzzles me a little that they have not implemented some kind of global
cache (like Go Modules does) to avoid me having to have a million copies of
postcss etc. installed.

A global cache distinct from -g (in /usr/local by default)?
https://docs.npmjs.com/files/folders

The cache in ~/.npm could honor XDG desktop spec; but that wouldn't solve
for anything more than the user-global cache in ~/.npm

bep:
I will finish this in https://github.com/gohugoio/hugo/pull/7223

I'm sorry for the delay, and thanks a lot for the work on this. I was a little bit too hooked up on getting something faster than os/exec to talk to Babel ... but there is nothing preventing us from adding this now, and improve it in the future.

PR #5806, opened 2019-04-01
hugolib: Add parameter option for .Summary

### LINKED ISSUE #5800: Fall back to summary set in front matter
Hugo's auto-generated page .Summary is often of no use to me so I write my own as a page variable.  This means, however, that I need to change common templates (e.g. RSS feed) to use said variable.

Is there any interest in a patch that would alter .Summary so that if a page variable called summary was available it used that value rather than auto-generating one from the content?

mcdee:
Add the ability to have a `summary` page variable that overrides
the auto-generated summary.  Logic for obtaining summary becomes:

  * if summary divider is present in content use the text above it
  * if summary variables is present in page metadata use that
  * auto-generate summary from first _x_ words of the content

Fixes #5800 

Tests have been added to ensure all three cases work as expected.  Note that in the tests the summary has markdown elements to ensure that processing occurs correctly; as such RST is explicitly excluded from the list of tested engines in the tests added in `hugolib/page_test.go`.

There is a discrepancy between the summary when auto-generated and when generated using the explicit delimiter; in the former case the output is plain and in the latter case it is wrapped with `<p>`.  I chose to go for the former option when generating from the summary metadata, as in general the summary would be wrapped in `<div> or similar` anyway, and it's cleaner for RSS output.

Talking of which, I added a custom summary to one of the test pages in `hugolib/site_test.go` to ensure that the output XML is correct with summary metadata that contains markdown; test is in `hugolib/rss_test.go`.

I haven't studied the Hugo codebase to any great depth so please do take a good look at the changes and let me know if there is anything that's incorrect here.  Thanks.

### DIFF
--- a/hugolib/page__per_output.go
+++ b/hugolib/page__per_output.go
@@ -134,10 +134,7 @@ func newPageContentOutput(p *pageState) func(f output.Format) (*pageContentOutpu
 						Cfg:        p.Language(),
 						DocumentID: p.File().UniqueID(), DocumentName: p.File().Path(),
 						Config: cp.p.getRenderingConfig()})
-					// Strip enclosing <p>
-					html = []byte(strings.TrimSpace(string(html)))
-					html = []byte(strings.TrimPrefix(string(html), "<p>"))
-					html = []byte(strings.TrimSuffix(string(html), "</p>"))
+					html = cp.p.s.ContentSpec.TrimShortHTML(html)
 					cp.summary = helpers.BytesToHTML(html)
 				}
 			}
--- a/hugolib/page__per_output.go
+++ b/hugolib/page__per_output.go
@@ -128,6 +128,17 @@ func newPageContentOutput(p *pageState) func(f output.Format) (*pageContentOutpu
 							cp.summary = helpers.BytesToHTML(summary)
 						}
 					}
+				} else if cp.p.m.summary != "" {
+					html := cp.p.s.ContentSpec.RenderBytes(&helpers.RenderingContext{
+						Content: []byte(cp.p.m.summary), RenderTOC: false, PageFmt: cp.p.m.markup,
+						Cfg:        p.Language(),
+						DocumentID: p.File().UniqueID(), DocumentName: p.File().Path(),
+						Config: cp.p.getRenderingConfig()})
+					// Strip enclosing <p>
+					html = []byte(strings.TrimSpace(string(html)))
+					html = []byte(strings.TrimPrefix(string(html), "<p>"))
+					html = []byte(strings.TrimSuffix(string(html), "</p>"))
+					cp.summary = helpers.BytesToHTML(html)
 				}
 			}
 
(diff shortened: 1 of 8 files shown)


bep:
Thanks. There are some tests that fail.

mcdee:
Okay, the issue was that the tests were in Markdown source and although I excluded the RST processor from the tests I missed the Asciidoc one due to not having it installed locally.  Tests should be passing now.

hugolib/page__per_output.go
```diff
@@ -128,6 +128,17 @@ func newPageContentOutput(p *pageState) func(f output.Format) (*pageContentOutpu
...
 						}
 					}
+				} else if cp.p.m.summary != "" {
+					html := cp.p.s.ContentSpec.RenderBytes(&helpers.RenderingContext{
+						Content: []byte(cp.p.m.summary), RenderTOC: false, PageFmt: cp.p.m.markup,
+						Cfg:        p.Language(),
+						DocumentID: p.File().UniqueID(), DocumentName: p.File().Path(),
+						Config: cp.p.getRenderingConfig()})
+					// Strip enclosing <p>
+					html = []byte(strings.TrimSpace(string(html)))
```
bep: These multiple `[]byte` conversions look slightly ineffective. I would probably just convert it to a `string` as a first step (as that is what it's going to end up as) and do the trimming on that ...

That said, we've had some discussions on this before, whether we should trim `p` tags or not.

I'm not totally sure what we do in the "summary split" case, but the code I read here will not work as a general case, so I think that needs to be adjusted/removed. The main problem is the handling of multiple paragraphs. I suggest that we get it in line with the other Markdown summary option.

bep: Second thought, if you pull the logic in 

https://github.com/gohugoio/hugo/blob/950295516da882dcc51d83f70835dde230a0b4d6/tpl/transform/transform.go#L116

Into some common function, and then use that.

mcdee: Would you prefer the common function to remain in `transform.go` or move to `helpers/`?

bep: Move it to helpers/content.go (it cannot live in transform.go anyway because that will create a cyclic reference)

bep: Add it as a method on the ContentSpec struct to make less global helper funcs.

mcdee: Done.  I was tempted to let the function trim all HTML passed in to it but that would change existing output so I stuck to principle of least surprise and only do so when the input is recognised as a single paragraph.

kaushalmodi:
@mcdee Please update the documentation too .. grep for Summary in the existing docs and makes changes related to this feature wherever appropriate. It would also be really useful to specify the new `summary` front-matter variable somewhere.

Thanks.

mcdee:
@kaushalmodi I was planning to do that after this PR was accepted; didn't want to write docs against a moving target.

bep:
@mcdee This looks just right, thanks for this. I merged this quickly to get it into my ongoing manual testing for Hugo 0.55 (a rather massive and cool release). I would really appreciate if you could update the docs (in the /docs folder in this repo) in a separate PR with the summary changes.

PR #5851, opened 2019-04-13
Optimize logo

MartijnCuppens:
The logo had a pretty big file size, so I optimized it _a bit_. File size decreased from `113.27kb` to `2.8kb`.


bep:
Thanks for this. The theme repository is located here:

https://github.com/gohugoio/gohugoioTheme/issues

It would be good if you could move this PR to the correct repo.

budparr:
@MartijnCuppens Curious what you used for optimization? We used svgo some time ago with different results, clearly.

MartijnCuppens:
@budparr, I remade the hexagons because there was a lot of room for improvement there. Afterwards I used svgo to further optimize the paths.

PR #5869, opened 2019-04-16
Fixed a typo

JinHan2002:
The word 'valuable' should be variable or value?


bep:

PR #5912, opened 2019-04-29
Update blackfriday to v1.5.2

dbirks:
Hugo is currently using a version of blackfriday from Aug 4, 2018. I'm proposing it to be updated to v1.5.2, which was cut on Sept 23, 2018.

This is to pull in the recent fixes to fenced code blocks:
https://github.com/russross/blackfriday/releases/tag/v1.5.2

The mage tests come back clean. 

Thanks for considering!


heavywatal:
Something might have been fixed in v1.5.2, but it has a serious regression, e.g., https://github.com/russross/blackfriday/issues/484

PR #5931, opened 2019-05-03
deploy: Support invalidating a CloudFront CDN cache

### LINKED ISSUE #5929: deploy: Add support for cache invalidation in CDN

vangent:
Fixes #5929.

* Adds a `invalidateCDN` boolean flag (defaults to true) that enables invalidation of a CloudFront CDN cache if any files changed during the "deploy".
* The CloudFront `DistributionID` is configured as part of the `target` in the config.
* Only CloudFront is supported for now, but other providers could be added.
* The invalidation is for the entire cache, not per-file. `s3deploy` supports a more complicated version which tries to invalidate individual files, but falls back to invalidating the entire cache if more than 8 files have changed. That added a lot of complexity in terms of keeping track of files changes, mapping paths, etc., and my guess is that a fairly high percentage of the time more than 8 files changed anyway, so I'd like to start with this.

### DIFF
--- a/deploy/cloudfront.go
+++ b/deploy/cloudfront.go
@@ -1,0 +1,51 @@
+// Copyright 2019 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package deploy
+
+import (
+	"context"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/cloudfront"
+)
+
+// InvalidateCloudFront invalidates the CloudFront cache for distributionID.
+// It uses the default AWS credentials from the environment.
+func InvalidateCloudFront(ctx context.Context, distributionID string) error {
+	// SharedConfigEnable enables loading "shared config (~/.aws/config) and
+	// shared credentials (~/.aws/credentials) files".
+	// See https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ for more
+	// details.
+	// This is the same codepath used by Go CDK when creating an s3 URL.
+	// TODO: Update this to a Go CDK helper once available
+	// (https://github.com/google/go-cloud/issues/2003).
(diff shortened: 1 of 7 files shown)


deploy/deployConfig_test.go
```diff
@@ -32,9 +32,12 @@ someOtherValue = "foo"
 [[deployment.targets]]
 Name = "name1"
 URL = "url1"
+CloudFrontDistributionID = "cdn1"
```
bep: I'm not sure we ever will get support for something other than CF, but I don't think you will have more than one at a time -- so we might as well call this `CDNDistributionID` or something. We can use CloudFront in help texts etc., but at least make room for another should it happen.

bep: I'm not sure we ever will get support for something other than CF, but I don't think you will have more than one at a time -- so we might as well call this `CDNDistributionID` or something. We can use CloudFront in help texts etc., but at least make room for another should it happen.

deploy/cloudfront.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// InvalidateCloudFront invalidates the CloudFront cache for distributionID.
+// It uses the default AWS credentials from the environment.
+func InvalidateCloudFront(ctx context.Context, distributionID string) error {
+	// SharedConfigEnable enables loading "shared config (~/.aws/config) and
+	// shared credentials (~/.aws/credentials) files".
+	// See https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ for more
+	// details.
+	sess, err := session.NewSessionWithOptions(session.Options{SharedConfigState: session.SharedConfigEnable})
```
bep: Is this the same authentication I would get for the S3 upload step? I say this because I find it a little surprising that the Google Cloud library doesn't have CDN as a concept. It would be ideal if we could share session/at least use the same code path. But we should at least make sure that if 1) S3 upload is OK 2) CDN invalidate should not fail because of an authentication failure.

deploy/deployConfig_test.go
vangent: If we did add a different provider, we'd need a way to identify which provider. So, I could change this to `CDNDistributionID`, but then if we add another one later we'd still need to add a new field like `CDNProvider`, with values like `cloudfront` and `fooprovider`.

I considered this but decided to leave it more explicit; if we add another provider we would add it as `FooProviderDistributionID`.

Let me know which way you want to go here.

deploy/cloudfront.go
vangent: > Is this the same authentication I would get for the S3 upload step?

Yes: https://github.com/google/go-cloud/blob/master/blob/s3blob/s3blob.go#L106-L110.

> It would be ideal if we could share session/at least use the same code path.

I filed an issue for Go CDK to add a helper for this so we can share the code:
https://github.com/google/go-cloud/issues/2003

> But we should at least make sure that if 1) S3 upload is OK 2)
> CDN invalidate should not fail because of an authentication failure.

Given that the code path is the same, this should be true.

bep: OK.

bep:
OK, looks good, I'll merge if you squash and fix the commit message (ref. my comment in another thread).

vangent:
Done.

PR #5933, opened 2019-05-03
deploy: Support configuration of upload order

### LINKED ISSUE #5932: deploy: add support for configuring order of uploads

vangent:
Fixes #5932.

Adds support for ordering of uploads, via a sequence of regular expressions configured in the `order` config under `deployments`. Uploads are sorted by path by default, but paths that match one of the `order` regexes are uploaded in that order.

I think this is similar to https://github.com/bep/s3deploy/pull/37, but (IMO) simpler.

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -23,7 +23,9 @@ import (
 	"mime"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
+	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


vangent:
Hrm, I just realized that we still do uploads in parallel, so even though the `uploads` slice is sorted correctly according to the ordering, they may not be applied strictly in that order.

I may have to change it to be a slice of slices for upload "groups"....

vangent:
This is ready for review now.

bep:
Looks good. Can you rebase it against master?

vangent:
> Looks good. Can you rebase it against master?

Done.

PR #5938, opened 2019-05-03
deploy: Add tests; fix Windows

### LINKED ISSUE #5936: deploy: Add a better end-to-end test
There are some good unit tests, but it would be nice to have a test that does something like:

* Creates a `afero.NewMemMapFs`.
* Creates a bunch of files in that Fs.
* Creates a `Deployer` using a Go CDK `mem://` URL.
* Does a `deployer.Deploy` and verifies that the appropriate number of changes were applied, and that the target looks correct.
* Does another `Deploy` and verifies no changes.
* Add/update/delete a file in the source Fs, re-Deploy and verify that each of the changes are made.
* Etc.

vangent:
Fixes #5936.

Let me know if you have other test ideas.

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -20,6 +20,7 @@ import (
 	"crypto/md5"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"mime"
 	"os"
 	"path/filepath"
@@ -46,6 +47,7 @@ import (
 // Deployer supports deploying the site to target cloud providers.
 type Deployer struct {
 	localFs afero.Fs
+	bucket  *blob.Bucket
 
 	target        *target          // the target to deploy to
 	matchers      []*matcher       // matchers to apply to uploaded files
@@ -56,6 +58,13 @@ type Deployer struct {
 	force         bool             // true forces upload of all files
 	invalidateCDN bool             // true enables invalidate CDN cache (if possible)
 	maxDeletes    int              // caps the # of files to delete; -1 to disable
+
+	// For tests...
+	summary deploySummary // summary of latest Deploy results
+}
+
+type deploySummary struct {
+	NumLocal, NumRemote, NumUploads, NumDeletes int
 }
 
 // New constructs a new *Deployer.
@@ -92,11 +101,18 @@ func New(cfg config.Provider, localFs afero.Fs) (*Deployer, error) {
 	}, nil
 }
 
+func (d *Deployer) openBucket(ctx context.Context) (*blob.Bucket, error) {
+	if d.bucket != nil {
+		return d.bucket, nil
+	}
+	return blob.OpenBucket(ctx, d.target.URL)
+}
+
 // Deploy deploys the site to a target.
 func (d *Deployer) Deploy(ctx context.Context) error {
 	// TODO: This opens the root path in the bucket/container.
 	// Consider adding support for targeting a subdirectory.
-	bucket, err := blob.OpenBucket(ctx, d.target.URL)
+	bucket, err := d.openBucket(ctx)
 	if err != nil {
 		return err
 	}
@@ -107,6 +123,7 @@ func (d *Deployer) Deploy(ctx context.Context) error {
 		return err
 	}
 	jww.INFO.Printf("Found %d local files.\n", len(local))
+	d.summary.NumLocal = len(local)
 
 	// Load remote files from the target.
 	remote, err := walkRemote(ctx, bucket)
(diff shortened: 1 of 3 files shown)


vangent:
Hmm, looks like the test identified a Windows issue.... will look into it next week.

vangent:
@bep please take a look when you have a chance. There were a number of changes needed to get Windows to work; I'll add some notes inline.

deploy/deploy.go
```diff
@@ -269,58 +288,51 @@ func doSingleUpload(ctx context.Context, bucket *blob.Bucket, upload *fileToUplo
...
+	NativePath string
+	// SlashPath is NativePath converted to use /.
+	SlashPath string
 	// UploadSize is the size of the content to be uploaded. It may not
 	// be the same as the local file size if the content will be
 	// gzipped before upload.
 	UploadSize int64
-	// UploadContentReader reads the content to be uploaded. Again,
-	// it may not be the same as the local file content due to gzipping.
-	UploadContentReader io.Reader
```
vangent: UploadContentReader did not play nice with Windows, because it created an open reader based on a file that was never closed (it was opened in the old `contentToUpload`, and can't be closed there without making the `Reader` not work anymore). Windows is persnickity about deleting files that have been opened.

The new approach is to read the file contents into a `byte.Buffer` regardless, and create a `Reader` as needed. This also cleans up the code a bit when computing MD5 hashes, but it does mean we'll read everything into memory. I can optimize this by trying to do the read lazily if you think it's worth it.

deploy/deploy.go
```diff
@@ -269,58 +288,51 @@ func doSingleUpload(ctx context.Context, bucket *blob.Bucket, upload *fileToUplo
 // localFile represents a local file from the source. Use newLocalFile to
 // construct one.
 type localFile struct {
-	// Path is the relative path to the file.
-	Path string
+	// NativePath is the native path to the file (using file.Separator).
```
vangent: To make sure the code is clear, I replaced `Path` with `NativePath` (the local filesystem path, which may include backslashes for example) and `SlashPath`, the native path with `filepath.ToSlash` on it.

The code is now clear about using the `SlashPath` version when matching regular expressions for matchers or ordering, and also (critically) as the key in the map returned by `walkLocal`, so that we don't get tons of spurious diffs on Windows due to "/" vs "\\" mismatches.

deploy/deploy.go
```diff
@@ -56,6 +56,14 @@ type Deployer struct {
 	force         bool             // true forces upload of all files
 	invalidateCDN bool             // true enables invalidate CDN cache (if possible)
 	maxDeletes    int              // caps the # of files to delete; -1 to disable
+
+	// For tests...
+	summary    deploySummary // summary of latest Deploy results
+	testBucket *blob.Bucket  // injected bucket
```
bep: I'm not a fan of having different code paths for test. I suggest that you add a `openBucket` field (func) to the `Deployer` struct.

deploy/deploy.go
bep: Why not set `UploadContentReader`to a `hugio.ReadSeekCloser`? And 1) reset the reader (seek) every time you use it and 2) Close every `localFile` at the end.

There can be lots of potentially big files here and we should not need to have 20 GB of ram to publish 10 GB of files. There may be double work (that may be optimized away) with the above approach, but we should try to avoid Out of mem errors at all cost.

deploy/deploy_test.go
```diff
@@ -361,3 +389,412 @@ func TestOrdering(t *testing.T) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	tmpbucketdir, err := ioutil.TempDir("", "bucket")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	memfs := afero.NewMemMapFs()
+	membucket := memblob.OpenBucket(nil)
```
bep: Since these tests write to memory and not to, say, Amason S3, you should rework the commit message title. These are not end-to-end tests; that would be "hugo && hugo deploy" with a real site and real buckets.

deploy/deploy.go
vangent: Done.

deploy/deploy_test.go
vangent: Done.

deploy/deploy.go
vangent: I changed `Reader()` to return an `io.ReadCloser` that the caller must close. Closing a returned reader is easier to manage than trying to close all the `localFile` objects.

vangent:
PTAL.

Issue #5904, opened 2019-04-27
Add `list all` command to CLI


rusnasonov:
I working on extension to [VSCode](https://marketplace.visualstudio.com/items?itemName=rusnasonov.vscode-hugo) and I want to show list of all posts. Would a PR be accepted that added a new command `hugo list all`?

bep:
> Would a PR be accepted that added a new command hugo list all?

It would if you at the same time refactored list future/draft to share a new common code path. They look a little like copy and paste and doing that a third time would not be ideal.

bep:
I can add that the current "file path" logic in these commands isn't entirely correct for all situations (folders outside of the work dir, projects with multiple content dirs). Or, it is possibly correct, but not good enough if you want to use it in an "open file" plugin in VSCode.

I suggest you replace the current logic with `strings.TrimPrefix .Page.Filename workDir`.

rusnasonov:
> It would if you at the same time refactored list future/draft to share a new common code path.

Ok, no problems.

> I can add that the current "file path" logic in these commands isn't entirely correct for all situations 
> I suggest you replace the current logic with strings.TrimPrefix .Page.Filename workDir.

Ok, I'll try to fix that. Replace current logic for all functions (draft, future), or only for new (all)?

bep:
>for all functions (

Yes.

rusnasonov:
@bep I create pr, but build fails with error for go 1.11.10

PR #5962, opened 2019-05-16
commands: Create new 'hugo list all' command

rusnasonov:
New:
- command `hugo list all`, return all posts meta in csv format

Refactoring:
- move common parts in commands/list.go to function `buildSites`
- change way to detect path to content

See #5904

### DIFF
--- a/commands/list.go
+++ b/commands/list.go
@@ -16,7 +16,8 @@ package commands
 import (
 	"encoding/csv"
 	"os"
-	"path/filepath"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gohugoio/hugo/hugolib"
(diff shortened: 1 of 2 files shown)


rusnasonov:
I don't now why build fails with error for go 1.11.10
```
exec: gofmt -l watcher/batcher.go
Error: improperly formatted go files
The command "mage -v check" exited with 1.
```
I did't touch the `wathcer/batcher.go`

bep:
>I did't touch the wathcer/batcher.go

You need to look further up the log for the actual filename. In short: You need to run `go fmt` on your  files. I use `LiteIDE` wchich does this auto on save.

rusnasonov:
fix

commands/list.go
```diff
@@ -50,117 +77,144 @@ List requires a subcommand, e.g. ` + "`hugo list drafts`.",
...
+
+				writer.Write([]string{
+					"path",
+					"slug",
+					"title",
+					"date",
+					"expiry_date",
+					"publish_date",
+					"draft",
+					"url",
```
bep: This overall PR looks good, but I think we need to stick to whatever fields we had in the existing list commands. It may be tempting to add more, but you will quickly get into questions about what URL / word count, etc. to use (re. output formats).

rusnasonov: Did I understand correctly, you think that additional fields like "url, word count, tags, categories" is unnecessary?

bep: I’m saying that there may be more than one URL, wordcount etc. for a page.

bep: But for most cases, I guess it will give the URL to the HTML version, which is probably what you want.

But:

* drop tags and categories (that is just relevant for people with tags and categories)
* drop word_count (for reason mentioned above)
* use the same naming standard for fields as in the struct; e.g. permalink, publishDate etc.

rusnasonov: > I’m saying that there may be more than one URL, wordcount etc. for a page.

But how? Can you give example?

> drop tags and categories, drop word_count, use the same naming standard

Done.

Issue #6018, opened 2019-06-07
template function sort cannot access an element in slice to sort by


greenstatic:
When using the `sort` template function it is not possible to access an element of a slice.

For example, lets consider the following content:

```markdown
---
title: Example 1
dates:
- 2019-05-30T14:30:00.000Z
- 2019-06-03T11:30:00.000Z
- 2019-06-03T11:30:00.000Z
```
```markdown
---
title: Example 2
dates:
- 2019-07-05T11:30:00.000Z
- 2019-07-07T13:30:00.000Z
```
```markdown
---
title: Example 3
dates:
- 2020-01-04T10:00:00.000Z
```

It is not possible to sort by let's say the first element of the slice `dates`. Providing the actual value of the element (using `(index .Params.dates 0)`) does not work since `sort` expects a field and not a value to sort by.  

(This is in reference to the forum post I created: [https://discourse.gohugo.io/t/sort-by-first-element-in-front-matter-slice/19090)](https://discourse.gohugo.io/t/sort-by-first-element-in-front-matter-slice/19090)

PR #6019, opened 2019-06-07
tpl: Add support to sort and where to access elements of a slice

greenstatic:
Template functions sort and where accept a key in order to perform their taks. However when the key is a slice it is not possible to access elements using their index. This commit adds support to access elements of a slice by appending a period followed by the element's index.

For example:
`{{ range (sort .Pages ".Params.dates.0") }}`
will sort the pages by using the first element of the slice
.Params.dates

This is a possible solution to: #6018

If the proposed solution is viable I can update the documentation to reflect the changes.

### DIFF
--- a/tpl/collections/where.go
+++ b/tpl/collections/where.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


kaushalmodi:
Can the syntax for index fetching in this filter be changed to `{{ range (sort .Pages ".Params.dates[0]") }}`?  What do you think @bep?

kaushalmodi:
What happens if a slice isn't big enough to fetch the specified index.. example `{{ range (sort .Pages ".Params.dates.100") }}`

bep:
This is the first time I have seen this requirement, so I think we should start with the _why_ and _is it important enough?_ before we start adding more complexity in this department.

greenstatic:
@kaushalmodi 
Currently the full string is split by periods, if we would like to change the syntax to `{{ range (sort .Pages ".Params.dates[0]") }}` it would require a bit more refactoring.

I agree however that the `[0]` syntax would be better to denote calling an element of a slice by index as opposed to simply `.0`.

In case the index is greater than the size of the slice, an error is thrown `error calling sort: index is out of bounds` during the build process.

@bep 
In my use-case I have a dates slice in front matter of a post. I wish to sort the posts by the first element of this dates slice. I agree it's a bit of a unique use-case, although the implementation was rather straightforward so I gave it a go.

bep:
@greenstatic let us hold this for a while to let others chime in with some convincing uses for this. I agree about the implementation looks straight forward, but I have learned the hard way that the simplest of things have side effects I did not see.

I think `.0` matches how we do map lookups.

Issue #5618, opened 2019-01-18
Support Apache MultiViews language negotiation file naming


anthonyfok:
It would be nice if Hugo could support the multilingual file naming convention described here:

> https://www.w3.org/International/questions/qa-apache-lang-neg#naming

such as:

> example.html.en, example.html.fr, example.html.de (language extention after `.html`),

or

> example.en.html, example.fr.html, example.de.html (language extension before `.html`)

such that Hugo can be a drop-in replacement for websites that use such Apache MultiViews approach for language negotiation.  For example, https://www.debian.org/

afranke:
[Content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) is not specific to Apache by the way. It’s “the right way” to serve multilangual pages, and it’s sad that hugo currently recommends using language specific paths or domains instead.

PR #6039, opened 2019-06-15
Add NoSuffix output format option to support MultiViews-style content negotiation

jordanwade90:
If the noSuffix option is set on an output format and uglyURLs are enabled, strip the MediaType.FullSuffix() from links to the Page. The intent is to support Apache MultiViews-style content negotiation (see #5618) on such links.

For example, given a config.toml with
```toml
[outputFormats.HTML]
noSuffix = true
```
example.md would be rendered to example.html but links to it would be simply example, without a suffix. I'm still looking into how to do language negotiation (for example, creating files example.en.html and example.fr.html).

### DIFF
--- a/output/outputFormat.go
+++ b/output/outputFormat.go
@@ -79,6 +79,12 @@ type Format struct {
 	// behaviour is wanted.
 	Permalinkable bool `json:"permalinkable"`
 
+	// Enable NoSuffix to omit the MediaType suffix of links to
+	// Pages rendered with this output format.  It is up to the
+	// server to map URIs to the correct file (for example, with
+	// Apache's MultiViews option).
+	NoSuffix bool `json:"noSuffix"`
+
 	// Setting this to a non-zero value will be used as the first sort criteria.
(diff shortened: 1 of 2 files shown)


bep:
I'm pretty sure you can do this today by creating a mediaType with no delimiter/exstension:

https://github.com/gohugoio/hugoDocs/blob/master/config.toml#L37

jordanwade90:
I tried this:
```toml
[outputs]
home = ["TEST", "RSS"]
page = ["TEST"]

[mediaTypes]
[mediaTypes."text/test"]
delimiter = ""

[outputFormats]
[outputFormats.TEST]
mediaType = "text/test"
baseName = "index"
rel = "canonical"
isHTML = true
permalinkable = true
```
and it renamed the generated files as well. For this the permalink needs to have no suffix but the file still needs to end .html (or whatever). Have I misunderstood what you were suggesting?

bep:
> Have I misunderstood what you were suggesting?

No, I misunderstood you, I re-read it now.

Issue #5589, opened 2019-01-07
Feature Request: Add an `Org.toHTML` function [go-org]


kaushalmodi:
Hello,

This is a feature request from @niklasfasching to add a helper function on Hugo-side so that `"*foo*" | Org.toHTML` would return *htmlized* `<strong>foo</strong>`. 

The idea is to same something analogous to `markdownify`, but to convert a string from Org mode instead of from Markdown. 

I already have an idea to use this feature: At present, I allow Webmentions on my blog to be entered as Markdown, and I `markdownify` to render them to HTML. I'd like to start supporting Org mode Webmentions now. 

@bep What do you think of this? 

---

I have mentioned before that the `markdownify` name seems to do the opposite of what it is named .. it sounds like it "converts **to** Markdown", but it's actually doing the opposite. Suggestion for `Org.toHTML` name was so that it's 100% clear as to what's being converted to what. 

Of course, please pitch in ideas for better names.

niklasfasching:
[Markdownify](https://github.com/gohugoio/hugo/blob/master/tpl/transform/transform.go#L100-L125) already goes through `RenderBytes` (which knows how to render Org mode) so this could be done by making `PageFmt` in markdownify customizable (or extracting the common functionality) - in any way this looks simple enough.

I wondered whether this would be a good idea as I'm not sanitizing input in go-org but it turns out blackfriday does nothing to protect against malicious content either, so that shouldn't be a problem (at least not a new one).

PR #6043, opened 2019-06-17
Org mode improvements (extend markdownify to support all markup formats)

niklasfasching:
- document org front matter format
- introduce htmlify (#5589)

### DIFF
--- a/docs/content/en/functions/markdownify.md
+++ b/docs/content/en/functions/markdownify.md
@@ -1,7 +1,7 @@
 ---
 title: markdownify
 linktitle: markdownify
-description: Runs the provided string through the Markdown processor.
+description: Runs the provided string through the Markdown processor. Takes an optional options map specifying the markup format.
 godocref:
 date: 2017-02-01
 publishdate: 2017-02-01
@@ -19,7 +19,8 @@ deprecated: false
 aliases: []
 ---
 
-
 ```
 {{ .Title | markdownify }}
+{{ .Title | markdownify (dict "format" "org") }}
+{{ .Title | markdownify (dict "format" "asciidoc") }}
 ```
--- a/tpl/transform/init.go
+++ b/tpl/transform/init.go
@@ -78,6 +78,8 @@ func init() {
 			[]string{"markdownify"},
 			[][2]string{
 				{`{{ .Title | markdownify}}`, `<strong>BatMan</strong>`},
+				{`{{ "*BatMan*" | markdownify (dict "format" "org")}}`, `<strong>BatMan</strong>`},
+				{`{{ "**BatMan**" | markdownify (dict "format" "markdown")}}`, `<strong>BatMan</strong>`},
 			},
 		)
 
--- a/tpl/transform/transform.go
+++ b/tpl/transform/transform.go
@@ -19,6 +19,7 @@ import (
 	"html/template"
 
 	"github.com/gohugoio/hugo/cache/namedmemcache"
+	"github.com/pkg/errors"
 
 	"github.com/gohugoio/hugo/deps"
 	"github.com/gohugoio/hugo/helpers"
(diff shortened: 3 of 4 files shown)


bep:
We need to stop a minute and think a little about the name. I may have thought these "ify" function names was clever (I think we stole it from Jekyll), but I'm not so sure anymore. `markdownify` was obviously not fantastic.

I suggest we call the new one `Markup`.

That said, we also need to specify the block (paragraphs) vs inline behaviour of such func. There are miles of discussions about this  elsewhere, so I will not go into details.

kaushalmodi:
I agree that the "ify" names are getting old.

> I suggest we call the new one Markup.

How about `toHTML` in `Markdown`, `Org`, .. namespaces?

Can we have something like:

```
{{ "**BatMan**" | Markdown.toHTML }}
```

```
{{ "*BatMan*" | Org.toHTML }}
```

I feel like above are readable as in plain English too.

bep:
The current namespace is `transform.Markdownify` (I don't want 1 function namespaces) and I suggest we do `transform.Markup`. The name is pretty good, and another reason is that I plan to add  `.Page.Markup` method which defaults to the renderer or the page itself. It would be good if they shared the same name (and we're not adding a bunch of new methods to the alrady big Page interface).

regisphilibert:
I'm sorry I don't have any suggestion here. But for me Markup in the global sense is not HTML only, it's an ensemble of factors determining the look of your page. CSS is part of the "Markup".

kaushalmodi:
@regisphilibert As an aside.. that you mention:

> Originally these the little handwritten comments added next to the text to indicate the publisher how those element should be "styled" on once printed.

https://shadycharacters.co.uk/ is an awesome read if you are interested in such topics related to publishing and special characters 😃

niklasfasching:
>  I suggest we do transform.Markup
> So.. how about a middle ground.. transform.ToHTML

I don't care either way but I'm unsure where to go from here.

kaushalmodi:
Given that no one has a preference, but I do, @bep, does my proposal in https://github.com/gohugoio/hugo/pull/6043#issuecomment-502778769 sound good? 🙂

bep:
>Given that no one has a preference, 
I have a preference, which is the more general term mentioned earlier.

* When I say `.Markup` (which is the page contextual variant of what we discuss here) I say: Render this using the page's rendering engine to the current output format.
* I understand that what we currently have renders to HTML only, I would be rather disappointed if the future did not bring ... more.

bep:
@kaushalmodi @niklasfasching sorry, I have been slightly distracted lately. I think I will delay parts of this discussion to a later time. For this particular PR I will say that we cannot add another function with a different name but with almost identical functionality (you have no idea how much time just explaining the difference to people will add up to, even if we add fantastic docs).

So, also saying that in the future we need something to indicate block vs inline for this, I suggest that we:

1. Keep the name `markdownify` for now. We can create an alias for it once we know what that is.
2. Add an optional options `map` so I can do `{{ "=hello=" | markdownify (dict "format" "org") }}`

The above will map fairly nicely to the `.Markdownify` func, which then uses the page context to guess the above.

bep:
Bump.

niklasfasching:
Got distracted - I'll look into this on the weekend :)

bep:
I will probably replace this PR now ...

I have thought about this, and I think `.Page.ToOutputFormat` (which will fall back to HTML for now) would be an OK name. It's a little bit mouthy, but I think it describes what it does well.

/cc @regisphilibert

regisphilibert:
I'm a bit confused here as of the purpose of this `.ToOutputFormat` method. Its name evokes compatibility with multiple output. I understand it can take either markdown or org as input but I'm not sure I can figure what output, other than HTML, it could ever produce?

If I'm working on a JSON template file and I use `.Page.ToOutputFormat`, what does happen?

bep:
@regisphilibert I changed my mind as I wrote it ... See my update. You are too fast.

regisphilibert:
Yes `.RenderString` is great! And now I see we're talking about Rendering a string with shortcode, markdown etc... This will help greatly building pages with Front Matter Blocks! 🎉

PR #6046, opened 2019-06-19
Remove type="application/javascript" type from Analytics

coliff:
This PR removes ` type="application/javascript"` from Google Analytics snippet. If type was to be used it should be ` type="text/javascript"`, but the type isn't needed to be there anyway.

REF:
- https://google.github.io/styleguide/htmlcssguide.html#type_Attributes
- https://codeguide.co/#html-style-script
- https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet

### DIFF
--- a/tpl/tplimpl/embedded/templates/google_analytics_async.html
+++ b/tpl/tplimpl/embedded/templates/google_analytics_async.html
@@ -1,7 +1,7 @@
 {{- $pc := .Site.Config.Privacy.GoogleAnalytics -}}
 {{- if not $pc.Disable -}}
 {{ with .Site.GoogleAnalytics }}
-<script type="application/javascript">
+<script>
 {{ template "__ga_js_set_doNotTrack" $ }}
 if (!doNotTrack) {
 	window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;


bep:
Hugo uses `type` to identify minifier to use, so we cannot merge this. I'm not going into the discussion of what's the correct MIME type for JS, but the one built into Hugo is this:

https://github.com/gohugoio/hugo/blob/master/media/mediaType.go#L133

coliff:
thanks for the reply @bep - fascinating. the link you gave with the MIME types uses MDN as the reference point (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textjavascript) and it states clearly there:

> Per the HTML specification, JavaScript files should always be served using the MIME type text/javascript. No other values are considered valid, and using any of those may result in scripts that do not load or run.

and there is a further note:

> Note: Even though any given user agent may support any or all of these, you should only use text/javascript. It's the only MIME type guaranteed to work now and into the future.

But then at the top of that article it states that the Internet Assigned Numbers Authority (IANA) is responsible for all official MIME types, and you can find the most up-to-date and complete list at their Media Types page and there it states: `text/javascript - OBSOLETED in favor of application/javascript`. I've seen linters (like webhint) state that `application/javascript` is incorrect - so maybe that is wrong!

So I can see now why you don't want to get into a discussion about what the correct MIME type is! 😄 

In any case, removing the type altogether doesn't appear to make a difference to minification with `hugo --minify` in my testing so is removing it not an option you'd consider?

XhmikosR:
Note that there might be more cases like that (Hugo bundles?).

coliff:
Yes, I think there are more cases of this -at least with the other Analytics snippets. What's your opinion on this subject @XhmikosR ? ...

Incidentally, the thing that flagged this to me was testing Hugo sites with http://webhint.io - but they've just merged a change to stop warning over the use of `application/javascript` - though they do recommend using `text/javascript`.

REF: https://github.com/webhintio/hint/issues/2621

XhmikosR:
I just follow the specs personally.

bep:
I can merge this if there is also a test that confirms that the minification is preserved. I assume it is, but it's good to have it confirmed.

bep:
>I just follow the specs personally.

This think the spec in this case says that the type is optional for JS.

XhmikosR:
Yeah, it's optional, i.e. not an error, but a warning. https://validator.w3.org/nu/?doc=https%3A%2F%2Fgohugo.io%2F

coliff:
> I can merge this if there is also a test that confirms that the minification is preserved. I assume it is, but it's good to have it confirmed.

I can confirm for certain that minification is preserved without `type="application/javascript"` but don't have a test to prove it to go with this PR. On the Hugo site it uses a Google Tag Manager script without the type defined though (https://github.com/gohugoio/gohugoioTheme/blob/master/layouts/partials/gtag.html), and that minifies correctly.

PR #6049, opened 2019-06-20
FIX: Do not trigger warning when a user customizes a theme's data files

gcushen:
A recent change was made to `hugo_sites.go` which caused a warning to be generated and displayed in `hugo` command output when a user purposely overrides the data file of a theme to customize parameters.

Overriding files is a *feature* of Hugo and should not generate Warnings/Errors in this scenario of a user customizing a theme's data files.

This PR fixes that logging bug that was introduced by correcting the log level from WARN to INFO.

See https://discourse.gohugo.io/t/hugo-bug-warning-higher-precedence-data-already-in-the-data-tree/19236


bep:
I have an upcoming fix for this in my working branch, and I will close this in favour of that one. It will come soon. There is a potential WARNING situation in there, but it is hard to detect so I'll turn it off.

PR #6063, opened 2019-06-26
Pagination - do not render href if no next item

markmandel:
When using a html link checker with Hugo, this template consistently causes errors, as it renders `href=""` attributes when next/previous is disabled.

This change makes it so that the `href` attribute is not rendered at all if `HasNext` is false - which is better semantically, and makes link checking far easier.


divinerites:
Thanks **a lot** for this fix, because it will solve properly this htmltest issue : https://github.com/wjdp/htmltest/issues/86

PR #6083, opened 2019-07-09
Fix typo s/Meny/Menu/

kaushalmodi:


bep:
It would have been correct in Norwegian ...

PR #6103, opened 2019-07-21
hugolib: remove unused function

peaceiris:
- remove `supports` function because it is no longer used.

This pull request is ready for review. Could you confirm this?


bep:
@peaceiris I appreciate these PRs, but this particular change conflicts with a rather big PR (Hugo Modules) that I will merge in a day or two. It rewrites the entire area around the function you now remove (which will be removed by my changes). It will probably save both of us some work if you could hold on with these "remove unused" PRs for a couple of days.

peaceiris:
Ok, I got it.

PR #6112, opened 2019-07-26
tools: Deploy to cloud using ansible

souvikhaldar:
Add ansible script and helper scripts to easily deploy a static site created using Hugo to Debian based servers on any cloud service provider. I have personally used this tool and still do, to deploy the changes my website content to the digital ocean droplet. Reference:- http://souvikhaldar.info/

I believe this should be useful for beginners as well as experts to easily put their changes to cloud and run the Hugo server as systemd service (for obvious reasons) all in one command.


bep:
Thanks, but we cannot add this to this repo.

souvikhaldar:
May I ask you why? @bep

souvikhaldar:
> Thanks, but we cannot add this to this repo.

May I ask you why?

bep:
>May I ask you why?

It does not belong here. There are millions of ways to deploy to the cloud, we cannot maintain scripts for all of them in this repo. That would be way too much work for this small team.

PR #6186, opened 2019-08-09
Use the SVG logo in README.md

XhmikosR:


bep:
Thanks. As a side note: I'm planning to do a Hugo release on Monday. It will be plenty cool.

PR #6207, opened 2019-08-14
Add `strings.HasSuffix` documentation

dsiddy:
I tracked down this function *via*
<https://discourse.gohugo.io/t/solved-add-hassuffix-function/13160>. It
came in handy for me; may it do so for others as well.


bep:
Thanks, but we have moved the Hugo site source to its own repository, so we would appreciate if you could open this PR directly in `hugoDocs`:  https://github.com/gohugoio/hugoDocs/pulls

Issue #4546, opened 2018-03-30
Options "--cleanDestinationDir" and "--cleanDestinationDirKeepDotFolders"


htr3n:
Hi all,

As hosting my site using `github.io`, I have just set my `github.io` page as a submodule under public.

`git submodule add git@github.com:htr3n/htr3n.github.io.git public`

I've used  the option `--cleanDestinationDir` before (no submodule under `public`) without problems. Now, I realize that the git's metadata (i.e. `.git`) under `public` are all deleted and therefore the submodule is effectively nullified and disconnected. 

I did some research around and found out relevant [reports / suggestions](https://github.com/gohugoio/hugo/issues/3202). Even the option `--cleanDestinationDirKeepDotFolders` was added [here](https://github.com/gohugoio/hugo/pull/3210) in March 2017. Nevertheless, I checked out `hugo version` and did not find the option  `--cleanDestinationDirKeepDotFolders`.

Has the option `--cleanDestinationDirKeepDotFolders` already been incorporated in Hugo or are there any other mitigations for cleaning the destination directory (`public`) but keeping dot files? 

I enclose here some relevant info.

```
$ brew install hugo --HEAD
...
$ hugo version
Hugo Static Site Generator v0.38-DEV darwin/amd64 BuildDate:
$ hugo --help
hugo is the main command, used to build your Hugo site.
...
      --canonifyURLs               (deprecated) if true, all relative URLs will be canonicalized using baseURL
      --cleanDestinationDir        remove files from destination not found in static directories
      --config string              config file (default is path/config.yaml|json|toml)
...
```
Thanks in advance for your help.

bep:
Thanks for getting in touch, but it feels like this is a question  that would be better suited to the [Hugo Discussion Forum](https://discourse.gohugo.io/).  We prefer to use GitHub issues only for bugs and enhancements. Feel free to update this issue with a link to the re-posted question (so that other people can find it) or add some more details if you feel this is a genuine bug. You may also find the answers you search for in [Hugo Documentation](http://gohugo.io/overview/introduction/).

pojebunny:
Sorry for digging out an old issue, but I'm currently in the same situation with HUGO v0.54.0 installed in /usr/local/bin and a setup for GitHub pages "public" being the build destination and a submodule of the root project directory.

Issuing a " hugo -b "$baseURL" -D --cleanDestinationDir --quiet " command removes everything including the .git file.

pojebunny:
[...]

Issue #3202, opened 2017-03-20
Ignore "." dirs in hugo --cleanDestinationDir 


hippora:
hi,when i execute hugo --cleanDestinationDir ,it clean the public directory,but also delete .git in public what i want sync to github.io. 

so can this directive did not remove .git ? thankyou

eamonnmcevoy:
@bep Hi, I'd like to contribute to the project, could I be assigned this issue? thanks

eamonnmcevoy:
I'm going to extend `mostafah/fsync` to allow a filter function to configured to allow certain files to remain in the destination folder

bep:
What is `mostafah/fsync`?

eamonnmcevoy:
It's one of the dependencies in the vendor folder, that's where the code is to clean the destination folder.
`github.com/spf13/vendor/github.com/spf13/fsync` which is a fork of `mostafah/fsync`

vbmade2000:
@bep  I think we can close this issue.

eamonnmcevoy:
The change I've made for fsync won't resolve this though, it's just allows a filter to be applied. I'll need to update Hugo once the fsync pull request is merged.

bep:
The fsync is merged ...

eamonnmcevoy:
@bep 
Should I change the default behaviour of `cleanDestinationDir` to ignore '.' folders? 
I'm considering add a new option instead `cleanDestinationDirKeepDotFolders`? 

What do you think?

eamonnmcevoy:
I think this can be closed now

PR #6261, opened 2019-08-24
Keep ".git" file in in --cleanDestinationDir

g-a-c:
Ignore files specifically called ".git" in order to preserve submodule information for Git post-1.7.8 (which seems to be the point things changed from having an entire “.git” directory inside the submodule to having a “.git” _file_ pointing to a gitdir outside of the submodule)

This should restore what I believe was the intended behaviour of #3202 and fix the issue for pojebunny in #4546

See #3202 and #4546

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -636,7 +636,7 @@ func (c *commandeer) copyStaticTo(sourceFs *filesystems.SourceFilesystem) (uint6
 		c.logger.INFO.Println("removing all files from destination that don't exist in static dirs")
 
 		syncer.DeleteFilter = func(f os.FileInfo) bool {
-			return f.IsDir() && strings.HasPrefix(f.Name(), ".")
+			return (f.IsDir() && strings.HasPrefix(f.Name(), ".")) || (!f.IsDir() && f.Name() == ".git")
 		}
 	}
 	c.logger.INFO.Println("syncing static files to", publishDir)


g-a-c:
> If this is a feature request, and you feel that it is still relevant and valuable, please tell us why.
> This issue will automatically be closed in the near future if no further activity occurs. Thank you for all your contributions.

I don't currently use Hugo but I believe this PR still adds value for other users attempting to use submodules with recent versions of Git.

Jonathan-Ironman:
Same issue here. The .git file should be preserved imo

g-a-c:
I don't remember any of the context around why I opened this. However...

* gohugoio/hugo#3210 shows that there was a separate command added called `cleanDestinationDirKeepDotFolders` which 
* This command is not documented, and if I'm reading the current `master` code correctly it is [part of the regular `cleanDestinationDir` functionality since ~2017](https://github.com/gohugoio/hugo/blob/b660ea8d545d6ba5479dd28a670044d57e5d196f/commands/hugo.go#L648-L656) - so I don't know why I ever even opened this...

Jonathan-Ironman:
Exactly, it leaves the dirs intact but clears .git, .gitignore, etc.
I'm having the public folder as a git submodule and it for instance also has a README.md file which would be nice to keep.
Custom regex filter settings would be ideal here I suppose, but ignoring dot-files is already a big help.

bep:
I didn't know `.git` could be a file -- we can check for that, but we're not making this more complicated than that.

Jonathan-Ironman:
https://github.com/spf13/fsync/pull/10 
This seems related, might it already be possible?

g-a-c:
> I didn't know `.git` could be a file -- we can check for that, but we're not making this more complicated than that.

Honestly, I don't remember all the details of this. But I think it was related to me wanting to pull in a remote theme without committing it in its entirety to my Hugo repo.

That seemed (at the time at least) like a sensible thing to do because it meant updating the theme was just a change of reference, and I think the way I achieved that was by having a `theme` subdirectory or something similar in my output directory, which was a submodule reference (this somehow turned into a `.git` _file_ inside the `theme` subdir, though I don't fully understand/remember why). So when the output directory was being cleared, I actually wanted that submodule reference to remain so that once the output step was finished I could pull the theme from its remote reference into my output directory and then `rsync --delete` the whole thing up to a webserver.

It was a very, **very** early exploration into rudimentary CI/CD for me at the time. It may no longer make sense, and I don't use Hugo anymore, which is why I just let the PR expire at the second time of asking instead of being an ass and trying to force a change through that I didn't need anymore.

luisfelipe3d:
any update?

DominoPivot:
This would still be pertinent, though the codebase has changed quite a bit since this PR was opened. The code now lives in [/commands/hugobuilder.go](https://github.com/gohugoio/hugo/blob/65871d5cf48d50935cf6e2d18483c32736c2730d/commands/hugobuilder.go#L471C3-L471C3). I could open my own PR if needed, though I don't have a Go environment ready and never contributed here.

```diff
		syncer.DeleteFilter = func(f os.FileInfo) bool {
-			return f.IsDir() && strings.HasPrefix(f.Name(), ".")
+			return strings.HasPrefix(f.Name(), ".")
		}
```

I ran into this issue when I found an old site of mine whose public folder contained a redundant clone of the same repository but with the `gh-pages` branch always checked out. This is precisely what a git worktree is for, I probably just didn't know at the time. I ran this command to replace the redundant clone with a worktree:
```sh
rm -rf public && git worktree add public gh-pages
```

This had the effect of replacing the `public/.git` folder with a `.git` file. I knew `--cleanDestinationDir` left `public/.git` alone before so I was surprised when it got deleted now that it was a file instead of a folder. I believe this should be fixed for consistency, though I'm not sure if there are people who rely on the current behavior. Still, I found some workarounds for now:

-   `git worktree repair public` to recreate the `.git` file
-   `git worktree lock public` to avoid accidental pruning if Hugo deletes the `.git` file in the future
-   copy the `.git` file to the static folder, and make sure not to run git commands within that folder

jmooring:
I've read through this thread several times and it no longer seems relevant given current (six years later) hosting methods.     We continue to retain dot directories when cleaning the `publishDir`, though at this point I can't come up with a good reason to do that either.

ilovelinux:
> This would still be pertinent, though the codebase has changed quite a bit since this PR was opened. The code now lives in [/commands/hugobuilder.go](https://github.com/gohugoio/hugo/blob/65871d5cf48d50935cf6e2d18483c32736c2730d/commands/hugobuilder.go#L471C3-L471C3). I could open my own PR if needed, though I don't have a Go environment ready and never contributed here.
> 
> ```diff
> 		syncer.DeleteFilter = func(f os.FileInfo) bool {
> -			return f.IsDir() && strings.HasPrefix(f.Name(), ".")
> +			return strings.HasPrefix(f.Name(), ".")
> 		}
> ```
> 
> I ran into this issue when I found an old site of mine whose public folder contained a redundant clone of the same repository but with the `gh-pages` branch always checked out. This is precisely what a git worktree is for, I probably just didn't know at the time. I ran this command to replace the redundant clone with a worktree:
> 
> ```shell
> rm -rf public && git worktree add public gh-pages
> ```
> 
> This had the effect of replacing the `public/.git` folder with a single file. I knew `--cleanDestinationDir` left `public/.git` alone before so I was surprised when it got deleted now that it was a file instead of a folder. I believe this should be fixed for consistency, though I'm not sure if there are people who rely on the current behavior. Still, I found some workarounds for now:
> 
>     * `git worktree repair public` to recreate the `.git` file
> 
>     * `git worktree lock public` to avoid accidental pruning if Hugo deletes the `.git` file in the future
> 
>     * copy the `.git` file to the static folder, and make sure not to run git commands within that folder

@jmooring 
I also have the `.domains` file for Codeberg Pages, which get removed for the same reason.

jmooring:
@ilovelinux This PR is closed. I suggest you create a new issue (proposal, not a bug).

PR #6268, opened 2019-08-26
fix languageCode format with https://www.ietf.org/rfc/rfc4646.txt

h0wardch3ng:
fix languageCode format with https://www.ietf.org/rfc/rfc4646.txt


davidsneighbour:
I read recently in a release note, that the language code can be anything now. It's probably not using RFC4646 as rule.

PR #6184, opened 2019-08-09
HTTPS links to Vimeo embeds

coliff:
This PR changes the iframe links to load over HTTPS. This is what Vimeo.com themselves recommends.

PR #6281, opened 2019-08-29
tpl: Migrate last shortcodes (YouTube and Vimeo) to HTTPS embeds

jakejarvis:
Hello! This is my first PR here so I hope I'm doing everything correctly.

I noticed while investigating `youtube-nocookie` compatibility that there were a couple of shortcodes still defaulting to HTTP embeds on HTTP sites. With only YouTube and Vimeo remaining, as of this PR all built-in shortcodes would officially fetch embeds securely by default, as both services' docs dictate.

Vimeo was already mentioned in #6184 a few weeks ago but I believe this covers all the bases regarding `mage generate` and `mage check` (which took me quite a while to crack!)

Please let me know if there's anything I can fix and/or improve. Thanks! 😊

### DIFF
--- a/hugolib/embedded_shortcodes_test.go
+++ b/hugolib/embedded_shortcodes_test.go
@@ -173,17 +173,17 @@ func TestShortcodeYoutube(t *testing.T) {
 	}{
 		{
 			`{{< youtube w7Ft2ymGmfc >}}`,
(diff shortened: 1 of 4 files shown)


bep:
This is great, thanks! I'll merge once the Travis builds goes green.

PR #6287, opened 2019-08-31
deps: Update go-org (fix descriptive lists)

niklasfasching:
</dt> and </dd> closing tags were broken, see https://github.com/niklasfasching/go-org/pull/9

also is there any limit to how often i should make pull requests for version updates? it seems like there will be a steady supply of bugs :D


bep:
>also is there any limit to how often i should make pull requests for version updates? 

As long as it's not 10 a day ... These dependency updates doesn't create any work on my side of the fence.

PR #6292, opened 2019-08-31
Escape RSS description as XML

### LINKED ISSUE #3268: RSS XML can contain invalid characters
I found that RSS feeds that Hugo generates can contain characters that are invalid for XML.

The XML 1.0 spec defines valid characters: https://www.w3.org/TR/2006/REC-xml-20060816/#charsets

One I encountered in the wild in a blog using Hugo is U+000b (\v, vertical tab). (It was this blog, if you're interested: https://blog.hypriot.com/)

Trying to parse such XML raises an error with Go's decoder (which is how I noticed this in the first place):

`XML syntax error on line 10: illegal character code U+000B`

My environment:
  * Hugo Static Site Generator v0.20-DEV linux/amd64 BuildDate: 2017-04-02T17:42:16-07:00
  * Debian Linux (testing) 64bit

Here are two small sample Go programs to help demonstrate the problem:

Create a post with an invalid character:

```
package main

import "fmt"

func main() {
	post := `+++
date = "2017-04-02T16:11:58+05:30"
draft = false
title = "New post"

+++

Hi there
`

	post += "\u000bsudo apt-get update\u000b"

	fmt.Println(post)
}
```

Use like this: `$ ./create-problem-post  > ~/t/bookshelf/content/post/newpost.md`

Then re-generate the site: `$ hugo`

Then try to decode the RSS feed with this program:

```
package main

import (
	"encoding/xml"
	"io/ioutil"
	"log"
	"os"
)

func main() {
	buf, err := ioutil.ReadAll(os.Stdin)
	if err != nil {
		log.Fatalf("Reading from stdin: %s", err)
	}

	type TestStruct struct {
		Blah string
	}

	t := TestStruct{}

	if err := xml.Unmarshal(buf, &t); err != nil {
		log.Fatalf("Unmarshal XML: %s", err)
	}
}
```

Like so:

```
$ cat ~/t/bookshelf/public/index.xml | ./read-problem-post 
2017/04/02 21:27:43 Unmarshal XML: XML syntax error on line 22: illegal character code U+000B
```

Thank you!

pwindle:
This is to avoid including characters invalid for XML.

Fixes #3268

I just found an Hugo built rss feed failing at https://pawelgrzybek.com/feed.xml. I think it is due to this old issue. I tested the latest version of Hugo and it still fails as described in #3268. The fix written there by @horgh still seems to work as well, so I've rebased his branch here for your consideration. Thanks.

### DIFF
--- a/docs/content/en/functions/xmlEscape.md
+++ b/docs/content/en/functions/xmlEscape.md
@@ -1,0 +1,28 @@
+---
+title: xmlEscape
+linktitle:
+description: Returns the given string with the reserved XML characters escaped.
+godocref:
+date: 2018-10-13
+publishdate: 2018-10-13
+lastmod: 2018-10-13
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [strings, xml]
+signature: ["xmlEscape INPUT"]
+workson: []
+hugoversion:
+relatedfuncs:
+deprecated: false
+aliases: []
+---
+
+In the result `&` becomes `&amp;` and so on. It escapes characters with
+special meaning in XML 1.0. If a character is not valid for XML 1.0, it
+gets replaced with the Unicode replacement character (U+FFFD).
+
+```
+{{ xmlEscape "Hugo & Caddy > Wordpress & Apache" }} → "Hugo &amp; Caddy &gt; Wordpress &amp; Apache"
+```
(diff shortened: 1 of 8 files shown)


bep:
I don't mind adding this if we need it, but  Is wrapping it in CDATA not an option for this partiular case?

pwindle:
@bep I would've said yes CDATA would be enough. But looking through the issue comments again it looks like that was tried and didn't work (I haven't tried to reproduce this.):

@moorereason commented on 8 Oct 2018
> I attempted to use <!CDATA[ ... ]]> around the <description> contents, but that didn't fix the issue. The illegal character within the CDATA block still violates the XML spec. (Additionally, we use html/template for the RSS feed, so the <!CDATA gets escaped if we try to use that in the RSS template, anyway.)

bep:
><!CDATA gets escaped if we try to use that in the RSS template, anyway.)

There are ways to prevent that. @moorereason may remember ...?

PR #6300, opened 2019-09-02
livereload: Ensure LibreJS compliance

odgalvin:
This commit means that when Hugo injects the livereload code onto a page when
the user is using 'hugo server', and a user using LibreJS tries to access a
webpage, the script is not incorrectly flagged as being non-free. It uses special
comments to inform the LibreJS browser extension that Hugo uses the Apache 2.0
license.
[See GNU's page about correctly flagging free JS.](http://www.gnu.org/software/librejs/free-your-javascript.html)


bep:
I don't use LIbreJS so I cannot reason about "if we need this or not", but I suspect this is more important for "user-facing JS". Live Reload is for development only (e.g. localhost).

XhmikosR:
I doubt this helps TBH. It's only for local development.

PR #6303, opened 2019-09-03
tpl: Remove eq argument limitation

### LINKED ISSUE #6237: remove eq argument limitation
I recently stumbled on this in the [Go Template doc](https://golang.org/pkg/text/template/#hdr-Functions):

> For simpler multi-way equality tests, eq (only) accepts two or more arguments and compares the second and subsequent to the first, returning in effect
>
> arg1==arg2 || arg1==arg3 || arg1==arg4 ...

We'd be inclined to write the following.
```
    {{ $test := "bonjour" }}
    {{ $bool :=  in (slice "bonjour" "goodbye" "ciao") $test }}
    {{*/ becomes /*}}
    {{ $bool :=  eq $test "bonjour" "goodbye" "ciao" }}
```
Hugo errors `wrong number of args for eq: want 2 got 4` at the moment.
 
It's obviously not critical but could help readability and align Hugo with Go Template.

vazrupe:
Fixes #6237

### DIFF
--- a/tpl/compare/compare.go
+++ b/tpl/compare/compare.go
@@ -90,17 +90,13 @@ func (*Namespace) Default(dflt interface{}, given ...interface{}) (interface{}, 
 	return dflt, nil
 }
 
-// Eq returns the boolean truth of arg1 == arg2.
-func (ns *Namespace) Eq(x, y interface{}) bool {
-	if ns.caseInsensitive {
+// Eq returns the boolean truth of arg1 == arg2 || arg1 == arg3 || arg1 == arg4.
+func (n *Namespace) Eq(first interface{}, others ...interface{}) bool {
+	if n.caseInsensitive {
 		panic("caseInsensitive not implemented for Eq")
 	}
-	if e, ok := x.(compare.Eqer); ok {
-		return e.Eq(y)
-	}
-
-	if e, ok := y.(compare.Eqer); ok {
-		return e.Eq(x)
+	if len(others) == 0 {
+		panic("missing arguments for comparison")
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this, but I don't think a `MultiEq` was the intention in #6237, which was to extend the number of args to `Eq` and `Ne` from 2 to a vararg.

vazrupe:
Hi, @bep

At first, I changed 2 args to a vararg. By the way, the function prototype changed so that `TestCompare` was not available.

https://github.com/gohugoio/hugo/blob/1b5c7e327c7f98cf8e9fff920f3328198f67a598/tpl/compare/compare_test.go#L148-L160

To avoid this problem, you should exclude it from testing or change it to:

```
{tstEq, n.Eq},.
// change to
{tstEq, func(a, b interface{}) bool { return n.Eq(a, b) }},
```

But both don't seem to be nice. Instead, it would be better to create a new function and call an existing function.

And `Ne` didn't implement because there is only an extension for `eq` in the [template document](https://golang.org/pkg/text/template/#hdr-Functions).

Thank you.

vazrupe:
But when I read the code to write a comment, I think the name `MultEq` is not good.. 😅😅😅

I think the name of this function is `Eq`.

bep:
If we make the signature of Eq and Ne to be in line with Go's variant, I think we're home free:

https://github.com/golang/go/blob/master/src/text/template/funcs.go#L442

So, no new function, we just change the existing functions.

vazrupe:
I agree with that! 👍

PR #6323, opened 2019-09-12
replace the deprecated github.com/markbates/inflect

fazalmajid:
with the recommended replacement github.com/gobuffalo/flect. This also fixes the annoyance: https://github.com/gobuffalo/envy/issues/30

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/kyokomi/emoji v1.5.1
 	github.com/magefile/mage v1.4.0
 	github.com/magiconair/properties v1.8.1 // indirect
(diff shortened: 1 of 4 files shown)


fazalmajid:
Hmm. Clearly the new library has slightly different behavior:
* old: Humanize(MyCamel) == My camel
* new: Humanize(MyCamel) == My Camel

I don't know which one would be considered correct.

bep:
>I don't know which one would be considered correct.

Their docs says:  "Humanize First letter of sentence capitalized" which is in line with the old behaviour and also makes sense for my _human_ brain.

Could you create an issue on that repo?

Issue #6298, opened 2019-09-02
Make png to jpg converter (flatten to specified background color)


ericselin:
I think this would be a killer feature. Serving .png images is less than ideal compared to .jpg. But keeping the master images as large .png files is a must if you want transparency. Hence something like this would be preferred:

```
{{ $png := .Resources.GetMatch "image.png" }}
{{ $jpg := $png.toJPG "#FFFFFF" }}
```

or using filters as in #6255:

```
{{ $jpg := $png.Filter (images.Flatten "#FFFFFF") }}
```

I hope you get my point. We have an eCom site and our product images are large .png files with a transparent background. Right now we flatten these in a separate step into .jpg files with two different backgrounds depending on what the background on the underlying element is. But I think this kind of processing could (and should!) be included by default in hugo.

roz3x:
hey ! the code base is (very ) huge , can you point the specific  file,  where this feature  could be added directly . like template parser or similar file .

bep:
@roz3x we need to agree on the API first. I don't mind adding this "somehow", but we need to consider where/how.

roz3x:
is this something you wanna fix right now , or you want to add in future updates !?

bep:
I don't mind having it fixed now, but we need to know how before doing it. `$img.ToJPG` does not look "right" to me, but this is not something I plan to spend time on right now.

jansorg:
Speaking for my needs, something like this would feel "okay" to me:

```
{{ $png := ... }}
{{ $jpg := $png.Convert("jpeg bg=#ffffff") }}
// or: {{  $jpg := $png.Convert("jpeg #ffffff") }}
// or: {{  $jpg := $png.Convert("jpeg", (dict "bg" "#ffffff")) }}
{{ $webp := $png.Convert("webp") }}
```

My thinking is that the required parameters depend on the target image format. Required parameters have to be passed to `Convert`. If missing, `Convert` will raise an error.
For example, creating a JPG from an image with transparent background needs the background color. A solution with a flatten filter might be harder to get because converting a png to jpeg without flattening before will have to assume a default background color.

A generic `Convert` could be used for all sorts of image formats, i.e. we wouldn't need functions for each format, e.g. `$img.ToPng`, `$img.ToJpeg`, `$img.ToWebP`, `$img.ToJpeg2000`, etc.

gnalck:
[...]

PR #6354, opened 2019-09-19
Support target image format  for image transformations

jansorg:
## Changes
The image format is defined as the image extension of the known formats.
All of `img.Resize "600x jpeg"`, `img.Resize "600x jpg"`, `img.Resize "600x png"` are valid
format definitions.
If the target format is defined in the operation definition string, then the converted image will be stored in this format.
Unknown extensions passed in the operation definition have not effect.

## Notes
Tests were already failing (for me) with the `master` branch. I've added a partial fix for one of these, but haven't fully fixed that particular test. Tests are still failing, but not more than with `master`.
[Log of go test ./... on master](https://github.com/gohugoio/hugo/files/3637245/tests-master.txt)

See #6298

### DIFF
--- a/resources/image.go
+++ b/resources/image.go
@@ -43,8 +43,6 @@ import (
 	"github.com/gohugoio/hugo/resources/images"
 
 	// Blind import for image.Decode
-
-	// Blind import for image.Decode
 	_ "golang.org/x/image/webp"
 )
 
@@ -220,15 +218,11 @@ func (i *imageResource) Filter(filters ...interface{}) (resource.Image, error) {
 	}
 
 	conf.Key = internal.HashString(gfilters)
+	conf.TargetFormat = i.Format
 
 	return i.doWithImageConfig(conf, func(src image.Image) (image.Image, error) {
 		return i.Proc.Filter(src, gfilters...)
 	})
-}
-
-func (i *imageResource) isJPEG() bool {
-	name := strings.ToLower(i.getResourcePaths().relTargetDirFile.file)
-	return strings.HasSuffix(name, ".jpg") || strings.HasSuffix(name, ".jpeg")
 }
 
 // Serialize image processing. The imaging library spins up its own set of Go routines,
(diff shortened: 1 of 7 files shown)


bep:
Re. MediaType:

If you add `setMediaType` to this interface:
 
https://github.com/gohugoio/hugo/blob/master/resources/resource.go#L173

It only needs to be implemented by the `genericResource`, I think.

Note that there is also a `metaAssigner` interface.

I think we should leave `Name` alone (for now, anyway), but I would expect that `.RelPermalink` would have the `jpg` extension.

jansorg:
@bep Thank you for the hints.
I've implemented `setMediaType` in `metaAssigner`, as this was easier to do and seemed right.
`.RelPermalink` is updated with the right file extension.

I think that this PR is ready now and hope that it's all in order.

resources/image.go
```diff
@@ -173,7 +171,7 @@ func (i *imageResource) cloneWithUpdates(u *transformationUpdate) (baseResource,
 // filter and returns the transformed image. If one of width or height is 0, the image aspect
 // ratio is preserved.
 func (i *imageResource) Resize(spec string) (resource.Image, error) {
-	conf, err := i.decodeImageConfig("resize", spec)
+	conf, err := i.decodeImageConfig("resize", spec, i.Format)
```
bep: `decodeImagConfig` is a method -- it has access to its `Format`, it should not be needed to send it as arg.

resources/image_test.go
```diff
@@ -536,6 +578,14 @@ func TestImageOperationsGolden(t *testing.T) {
 	c.Assert(err, qt.IsNil)
 	dirinfos2, err := d2.Readdir(-1)
 
+	// sort entries of both dirs for stable tests, don't rely on the fs to handle sorting
+	sort.Slice(dirinfos1, func(i, j int) bool {
+		return strings.Compare(dirinfos1[i].Name(), dirinfos2[i].Name()) < 2
+	})
+	sort.Slice(dirinfos2, func(i, j int) bool {
+		return strings.Compare(dirinfos2[i].Name(), dirinfos2[i].Name()) < 2
```
bep: Do we need these?

I was under the impression that `Readdir` had a defined order (it says so in the docs) and that it was stable across OSes. I have not seen any test failures with this?

Also, the less funcs used does not look right.

resources/image.go
jansorg: Makes sense. I wanted to make it clear which fallback is being used, but it's a bit unclean. 
Fixed.

resources/image_test.go
jansorg: This wasn't needed for this PR. I removed it and force-pushed the changes.

jansorg:
@bep I updated the code for your latest comments, it should be fine now.

PR #6361, opened 2019-09-21
wip: Add webp support for image decoding and encoding.

jansorg:
**note: this isn't as straight-forward as I thought. the webp dependency isn't compiling well on macOS and Windows. I'm not sure yet if these issues can be solved.**

This adds support to read and write webp images to hugo.
More on webp at https://developers.google.com/speed/webp . 

It uses https://github.com/chai2010/webp (BSD-3-clause license)
which isn't adding external dependencies. It contains C files, though. If this is too large or not wanted in the base hugo, would it be possible to integrate this in the extended build configuration?

webp images are decoded using Golang's x/image/webp package to simplify
the implementation, https://godoc.org/golang.org/x/image/webp . If this isn't good enough, then we could use the webp lib above to decode webp images, too.

Encoded webp images are stored lossy and with the defined quality
setting (i.e. identical to JPEG). The default quality set for Hugo
is used as a fallback.

(I have used this locally and it seems to be working well so far.)

See #6298


bep:
I'll close this now. We need to discuss/decide if this is something we want to go -- and also, how.

Issue #6322, opened 2019-09-11
Bad include


cup:
Using this file:

~~~
#+INCLUDE: "app.php" src php
~~~

I get this result:

~~~
WARN 2019/09/09 23:04:11 Bad include
org.Keyword{Key:"INCLUDE", Value:"\"app.php\" src php"}:
open posts\app.php: The system cannot find the path specified.
~~~

Using this file:

~~~
#+INCLUDE: "posts\app.php" src php
~~~

I get this result:

~~~
WARN 2019/09/09 23:05:23 Bad include
org.Keyword{Key:"INCLUDE", Value:"\"posts\\app.php\" src php"}:
open posts\posts\app.php: The system cannot find the path specified.
~~~

cup:
https://github.com/niklasfasching/go-org/issues/10

kaushalmodi:
@cup You're an Org mode user too (I have seen you in the Nim community too)! :)

Copying @niklasfasching on this issue as he's the maintainer of **go-org**.

cup:
@kaushalmodi, yeah. The only problem with Org mode syntax highlighting is pretty
bad with Jekyll, as GitHub refuses to open source their highlighter:

https://github.com/cup/rosso/blob/master/publish/markup.md#org-mode

However I think Hugo has its own highlighter, so if I can resolve this issue
with includes I could switch away from Jekyll finally.

cup:
https://github.com/github/markup/issues/1266

jerith:
The problem here is that the path in `DocumentName` is relative to the content dir rather than the repo root. This means that when go-org expands the relative path, it gets with `/path/to/repo/foo.org` instead of `/path/to/repo/content/foo.org` (and similarly for nested dirs within `content`).

I'm currently working around this by symlinking `posts` to `content/posts`, but I think the solution is for hugo to either pass an absolute path to go-org's `Parse` method or pass `DocumentName` with `content/` prefixed to it so that it's relative to the repo root again.

jerith:
I seem to get the correct behaviour if I compile my own hugo with this very hacky change:
```diff
diff --git a/helpers/content.go b/helpers/content.go
index f6576c04..9dce57fc 100644
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -22,6 +22,7 @@ import (
        "fmt"
        "html/template"
        "os/exec"
+       "path/filepath"
        "runtime"
        "unicode"
        "unicode/utf8"
[...]

PR #6383, opened 2019-10-01
Update go-org & fix includes

niklasfasching:
see #6322 

The org mode renderer supports including other files [1]. To resolve relative
paths we need the path of the current file - either absolute or relative to the
directory hugo is run from. DocumentName is the path relative to the content
directory, not the base directory that hugo is run from - we could get the full
path by looking up the content directory from the config and merging both; or
we just make the absolute path of rendered files available in the
RenderingContext.

[1]: e.g. `#+INCLUDE: ./foo.py src python` includes `foo.py` as a python source
block.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/muesli/smartcrop v0.0.0-20180228075044-f6ebaa786a12
 	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646 // indirect
 	github.com/nicksnyder/go-i18n v1.10.0
-	github.com/niklasfasching/go-org v0.1.4
+	github.com/niklasfasching/go-org v0.1.6
 	github.com/olekukonko/tablewriter v0.0.0-20180506121414-d4647c9c7a84
 	github.com/pelletier/go-toml v1.4.0 // indirect
 	github.com/pkg/errors v0.8.1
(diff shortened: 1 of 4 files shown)


bep:
Hugo has this concept of a virtual file system:

* With composable/union mounts etc.
* With some built-in sandboxing (security motivation), most notably: Only the project (not themes) can mount files outside of the project dir, only the project has symlink support.

The first point may not be a deal-breaker, but I'm not thrilled about themes "poking around on my disk and including stuff".

So what I suggest instead is that you instead of sending the filename, send in an "OpenFile` func. that returns a `io.ReadCloser`.

niklasfasching:
> The first point may not be a deal-breaker, but I'm not thrilled about themes "poking around on my disk and including stuff".

Could you elaborate why you say theme? This is part of `go-org`, i.e. the Org mode / `.org` renderer and should be unrelated to themes. I understand the security concerns though and like the OpenFilefunc idea - could you give me some pointers on where I could get that from? I didn't see anything obvious on a glance...

bep:
>This is part of go-org, i.e. the Org mode / .org renderer and should be unrelated to themes. I

When we, recently, introduced Hugo Modules, we added support for file mounts, including content from themes/theme components (including org files). 


As to "Open File"

See https://godoc.org/github.com/gohugoio/hugo/helpers#PathSpec => https://godoc.org/github.com/gohugoio/hugo/hugolib/filesystems#BaseFs

If you use the `Content` filesystem to open the files you can just `Path` and relative filenames.

As to how to get that filesystem to where you need it, I'm not sure. I have plans on moving these "content renderer" into its own package once I add Goldmark.

niklasfasching:
Thx! I got the BaseFS from page.pageCommon.

Would you be fine something along this changeset? If so I'll add `ReadFile` to go-org and update the PR with v0.1.6.

The current ReadFile allows relative includes by checking in the content FS first but falls back to trying to read the path relative to the project root. I guess that's a bad idea security wise so maybe we should just allow the content fs...

niklasfasching:
@bep ping. Ignore my last comment, I restricted it to just `Content.Fs` for now.

bep:
So, I don't see the security problem in reading files from the project root. We have plenty of examples of us doing that (there is even a `Work` filesystem in there that points to the project root.

But there are obvious benefits to having the files live in the standard Hugo folders:

* You can compose/combine the fs via theme components.
* The folders are watched in server mode
* And when a file is changed, we can look at it and do a partial rebuild

PR #6389, opened 2019-10-04
Update README.md

MjAbuz:
Trying with hugo extended  and hugo fresh theme and complied on ubuntu 18.04


bep:
Thanks, but I don't think we can use these changes.

PR #6391, opened 2019-10-05
deps: Upgrade to latest version of emoji dependency

jamietanna:
To add support for new emojis in Hugo, we need to upgrade our internal
dependency on the emoji package.


bep:
Sorry, I fast a little bit fast merging this. I had to revert this due to related test failures. Could you create another one and run the tests?

PR #6392, opened 2019-10-05
deps: Upgrade to latest version of emoji dependency

jamietanna:
To add support for new emojis in Hugo, we need to upgrade our internal
dependency on the emoji package.

Note that we also need to update our tests, as the underlying emoji that
is rendered has changed.

Follow-up to #6391. (170f18d9352d39213170dd9d5e947eb45854c84b and
2df5d202c6fc6eb18ee0b259c3dd1156094947c5).


jamietanna:
Let's hold off on merging until Travis has completed - I've tried against my fork and so far it's happy, and is just building Windows now.

Issue #3040, opened 2017-02-15
Add "block" option/action to markdownify


alizain:
As @robsonsobral commented in the original issue #1025 that added this "functionality", this is really non-standard, confusing behaviour. Why was it added in? More importantly, can it be removed, or put behind a config option?

For context, refer to this commit be627fa718d23278aeb53da5651a8a4a1be14c27. Hugo manually strips the leading and trailing paragraph tags, thereby rendering the `markdownify` function useless for any string that contains block level markdown.

This can cause major bugs! For example, if your yaml frontmatter looks like this:

```yaml
somekey: |
  This is a **valid** markdown paragraph.

  This is another **valid** markdown paragraph. And it's *still valid yaml!*
someotherkey: 2
```

`markdownify` will output the following, incorrect HTML, where the opening tag for the first paragraph, and closing tag for the last paragraph are missing!

```html
This is a <strong>valid</strong> markdown paragraph.</p>
<p>This is another <strong>valid</strong> markdown paragraph. And it's <em>still valid yaml!</em>
```

bep:
You are right, that was a bad idea -- I'm not sure what I was thinking.

alizain:
Thanks @bep! I'm using a fork of `hugo` for my work that reverts the specific changeset, should I submit a PR?

bep:
Or, thinking about it -- I'm not sure. The main use case for `markdownify` is for titles etc. and not blocks of test, and it is easier to add `p` tags than it is to remove it from a template.

robsonsobral:
Can we have a parameter for that? Something like `inlineOnly`, which will allow only inline level elements, like `b`, `strong`, `em`, `a`...

robsonsobral:
> it is easier to add p tags than it is to remove it from a template

Not always.

alizain:
> and it is easier to add p tags than it is to remove it from a template.

It's actually a lot harder. Especially when you have non-`<p>` elements that begin or end the string.

alizain:
Perhaps `markdownifyBlock` & `markdownifyInline` functions? As mentioned, I see the use-case for inline markdown styles in headers, but calling the function `markdown` and then producing non-standard output is highly confusing.

bep:
Here is what we need:

1. `markdownify` must continue to work as inline
2. We need a way to signal a `block` option/action
3. Even if we have variants like this in `partialCached`, I don't like that very much.
4. `markdownify "block"` isn't very pretty, either
[...]

PR #6398, opened 2019-10-08
Document unexpected mardownify behaviour

looeee:
From an end user standpoint the behaviour is confusing and needs to be documented at least. I'm not sure I fully understand the reasoning so I haven't attempted to explain it, but now it's at least documented. 

See: 

* https://github.com/gohugoio/hugo/issues/3040
* https://discourse.gohugo.io/t/single-paragraph-rendered-withing-shortcode-using-inner-is-missing-p-tag/21155
* https://discourse.gohugo.io/t/prevent-markdownify-to-strip-first-p-tags/7854/5

### DIFF
--- a/docs/content/en/functions/markdownify.md
+++ b/docs/content/en/functions/markdownify.md
@@ -23,3 +23,5 @@ aliases: []
 ```
 {{ .Title | markdownify }}
 ```
+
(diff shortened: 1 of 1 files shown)


looeee:
Bump - any chance of getting this added to the docs?

PR #6402, opened 2019-10-08
Added JAMStack themes

remotesynth:
Linking only to Hugo themes but thought these would be a useful tool for devs looking for starter sites that connect to a CMS.


bep:

PR #6404, opened 2019-10-09
common: Fix elements are doubling when append a not assignable type

### LINKED ISSUE #6188: Doubling of last slice element when adding template.HTML
I came across a weird error that I could boil down to this minimal example:

```
{{ $c := delimit (slice "c") " " }}<!-- should be 'c' -->
{{ $abc_slice := slice "a" | append "b" $c  }}<!-- should be ['a','b','c'] -->
{{ $abc := delimit $abc_slice "+" }}<!-- should be 'a+b+c' -->
{{ errorf "'%s' (%T) - '%s' (%T)" $c $c $abc $abc }}
```

Put that in a shortcode and strangely enough, the result is:

`'c' (template.HTML) - 'a+b+b+c' (template.HTML)`

As you can see, b is doubled.

The problem goes away if you add ` | print` to $c, making it a string. I wonder why it's not a string in the first place.

Thanks in advance for any comments and fixing.

vazrupe:
Fixes #6188

### DIFF
--- a/common/collections/append.go
+++ b/common/collections/append.go
@@ -65,6 +65,7 @@ func Append(to interface{}, from ...interface{}) (interface{}, error) {
 		fv := reflect.ValueOf(f)
 		if !fv.Type().AssignableTo(tot) {
 			// Fall back to a []interface{} slice.
+			tov, _ := indirect(reflect.ValueOf(to))
 			return appendToInterfaceSlice(tov, from...)
 		}
 		tov = reflect.Append(tov, fv)
--- a/common/collections/append_test.go
+++ b/common/collections/append_test.go
@@ -14,6 +14,7 @@
 package collections
 
 import (
+	"html/template"
 	"testing"
 
 	qt "github.com/frankban/quicktest"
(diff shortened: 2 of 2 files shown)


tpl/collections/collections.go
```diff
@@ -84,9 +83,9 @@ func (ns *Namespace) After(index interface{}, seq interface{}) (interface{}, err
 	return seqv.Slice(indexv, seqv.Len()).Interface(), nil
 }
 
-// Delimit takes a given sequence and returns a delimited HTML string.
+// Delimit takes a given sequence and returns a delimited string.
 // If last is passed to the function, it will be used as the final delimiter.
-func (ns *Namespace) Delimit(seq, delimiter interface{}, last ...interface{}) (template.HTML, error) {
+func (ns *Namespace) Delimit(seq, delimiter interface{}, last ...interface{}) (string, error) {
```
bep: Can you revert this change. It's not directly relevant to the issue in question, and it may also be a possibly breaking change.

vazrupe:
@bep 

Okay. I will split the patch.

bep:
Thanks for this!

Issue #5056, opened 2018-08-10
Fix Docker/Dockerhub situation


kav:
I'm aware that in #4155 @ellerbrock created dockerhub and quay accounts and @bep mentioned a different dockerhub account and @skoblenick proposed that one day hugo should be an Official image and #4518 incorporated all @ellerbrock's changes into the dockerfile.

However...

Neither Quay nor Dockerhub currently have a remotely up to date version of hugo at present in any of the above mention locations.

What would it take to get automatic publishing on dockerhub tied to this repo at either of the `gohugo` or `gohugoio` dockerhub accounts?

bep:
Some months ago I would probably have answered "why do you need a Docker version?", but with the recent libsass extensions, I can see the need.

I have one Docker image living in my account that belongs to Gohugo, and we should also get a binary version in here. I remember vaguely the person behind the gohugoio said he was willing to transfer ownership.

kav:
Well it’s less so needing or not needing to run Hugo on Docker. It’s pretty straightforward to clone the repo and use the existing file to do that.

It’s more a case that one might expect to find Hugo on Dockerhub or Quay and naively use one of the existing out of date images.

I’m not sure what you mean by a binary version in here. Can you explain a bit more?

jhabdas:
Would like to see a BusyBox roll like the dev image with auto-build from CI/CD as suggested in #4693 which just runs the _binary version_ and weighs only a few MB though keeping intermediate layers from the Alpine build does have it's advantages and can help get more hands on the codebase.

jhabdas:
Would be very handy to see a BusyBox tag, Alpine for dev and then copy the binary into scratch for custom build pipelines with CI/CD as suggested in #4693 by adding another layer to multistage build. BusyBox will enable theme maintainers an easy way to move themes into containers to run alongside the final binary without having to pull the top layers.

nizsheanez:
FYI: i using klakegg/hugo:0.46-ext-alpine
this repo is very good: automatically releasing new versions, has extended hugo, it's small.
but this repo a bit doesn't follow standard hugo folders structure, for example it has /src and /target volumes. 

In my CI it looks this way: 
```
docker run --rm -v $(pwd)/:/src -v $(pwd)/public:/target -e "HUGO_ENV=production" klakegg/hugo:0.46-ext-alpine --baseURL="https://mywebsite.com/"
```

maiki:
[...]

Issue #4693, opened 2018-05-03
Docker image auto build and deploy


kfeofantov:
Please, add automated docker build and push at docker hub in CI Pipeline. 
I can write pipeline for CircleCI or Travis.

jhabdas:
I was looking on Kitematic yesterday and there're quite a few _unofficial_ dockers for hugo, most of which are [not well-maintained](https://github.com/ecma6/hugo) or poorly constructed (i.e. not using scratch).

The current `Dockerfile` in the project repo makes some assumptions, such as to which ports to expose and volumes to use, which, while nice, would still require some adjustments for use in some [Compose](https://docs.docker.com/compose/) projects.

Given the base image, env vars and apt packages will change between builds having an official image will help prevent the need to use something like Docker-in-Docker, which adds a funky abstraction to running hugo in containerized environments.

**TL;DR:** Please update the CI/CD for deploys to DockerHub as OP suggests.

jhabdas:
For those looking for a quick way to build the `0.44` with Docker I've created a Dockerfile to build tagged versions of Hugo here (path subject to change):

https://git.habd.as/comfusion/after-dark/src/branch/fathom/docker/hugo/Dockerfile

mlake:
+1 for Composeability

Yajo:
I think this is still relevant because Hugo users usually perform some kinda autobuild for their websites, usually done in a CI/CD pipeline, usually based on Docker for easiness and reusability.

This would make the project more usable, and still more maintainable because CI would do the dirty job.

jonahbron:
I would also like to see an officially maintained Docker image be added. :+1:

eugene-bright:
There is [an official docker image on dockerhub][hub]. But it is stale like this issue.

[hub]: https://hub.docker.com/r/gohugoio/hugo/tags

bep:
>There is an official docker image on dockerhub. But it is stale like this issue.

That `gohugoio` account isn't official.

eugene-bright:
So it's a well crafted fake that drains reputation then.

pierreozoux:
Gitlab provides one.
I also think that the project should provide one, it would be cleaner for our CICD pipelines :)

jufemaiz:
LGTM on dockerhub?

ChenXiyu:
It is a good practice to wrap a command line tool into a docker image, I believe. 
Could you please release an official image for us?
Or you have other concerns about this?

[...]

PR #6405, opened 2019-10-09
Dockerfile: Switch to mage builds, various optimizations

jakejarvis:
Hello all! This is tangentially related to #5056. I'm suggesting the following changes to the current `Dockerfile`:

- Use Alpine in both stages for consistency.
- Switch to `mage install` from `go install` as dictated in [the latest docs](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md).
- Easy switch to building Hugo Extended by directly setting `HUGO_BUILD_TAGS` flag at image build time (simpler than the current `99notag` workaround):

```
docker build --build-arg HUGO_BUILD_TAGS=extended .
```

- Update to Go 1.13 and Alpine 3.10.
- The only possibly breaking change: Moved the `hugo` binary in the final stage from `/hugo` to `/usr/bin/hugo` so one can simply run `hugo` without having to search around for its location.

---

I realize these are big changes and I'm honestly not sure how much the team depends on Docker — it doesn't seem like very much from the outside! — but I don't want to break anyone's workflow so if you'd like me to revert any one of the changes above just let me know. I've been [using](https://github.com/jakejarvis/hugo-custom/blob/master/Dockerfile) almost these exact steps for the past few releases and they've worked great, even with SASS support. 😉 

Thanks as always!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,32 +2,44 @@
 # Twitter:      https://twitter.com/gohugoio
 # Website:      https://gohugo.io/
 
-FROM golang:1.11-stretch AS build
+FROM golang:1.13-alpine AS build
 
+# Optionally set HUGO_BUILD_TAGS to "extended" when building like so:
+#   docker build --build-arg HUGO_BUILD_TAGS=extended .
+ARG HUGO_BUILD_TAGS
 
-WORKDIR /go/src/github.com/gohugoio/hugo
-RUN apt-get install \
-    git gcc g++ binutils
-COPY . /go/src/github.com/gohugoio/hugo/
-ENV GO111MODULE=on
-RUN go get -d .
-
-ARG CGO=0
+ARG CGO=1
 ENV CGO_ENABLED=${CGO}
 ENV GOOS=linux
+ENV GO111MODULE=on
 
-# default non-existent build tag so -tags always has an arg
-ARG BUILD_TAGS="99notag"
-RUN go install -ldflags '-w -extldflags "-static"' -tags ${BUILD_TAGS}
+WORKDIR /go/src/github.com/gohugoio/hugo
+
+COPY . /go/src/github.com/gohugoio/hugo/
+
+# gcc/g++ are required to build SASS libraries for extended version
+RUN apk update && \
+    apk add --no-cache gcc g++ musl-dev && \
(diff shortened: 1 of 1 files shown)


jakejarvis:
force-pushing to re-trigger Travis — **_completely_** unrelated to this PR but is there some weird non-deterministic race error going on? Not familiar with the tests but it seems odd for this one job to fail out of otherwise successful ones. https://travis-ci.org/gohugoio/hugo/jobs/595643601

bep:
>Not familiar with the tests but it seems odd for this one job to fail out of otherwise successful ones.

https://github.com/gohugoio/hugo/issues/6409 -- we have some global loggers (yes I know) that we naively reset between builds -- I assume this is only a real problem when testing, but we should fix it.

As to the changes itself. I'm not using this Dockerfile myself, but I know other does. I have only one comment on your changes: I would prefer it if we avoided using Mage to build. It doesn't add much usefulness to the mix, and if we would want to use it, I think we need to download a release version and rely on compiling from source.

 @anthonyfok may have something to say.

Dockerfile
```diff
@@ -2,32 +2,44 @@
...
 COPY . /go/src/github.com/gohugoio/hugo/
-ENV GO111MODULE=on
-RUN go get -d .
 
-ARG CGO=0
-ENV CGO_ENABLED=${CGO}
-ENV GOOS=linux
+# gcc/g++ are required to build SASS libraries for extended version
+RUN apk update && \
+    apk add --no-cache gcc g++ musl-dev && \
```
riyad: FYI: `musl-dev` is automatically installed when installing `g++`

jufemaiz:
Just a QQ: https://hub.docker.com/r/gohugoio/hugo seems to be out of date compared with this. The builds are not automated I assume?

https://hub.docker.com/r/gohugoio/hugo/builds

peaceiris:
@jufemaiz 
> That gohugoio account isn't official.
> https://github.com/gohugoio/hugo/issues/4693#issuecomment-518031415

jufemaiz:
@peaceiris eep. Perhaps that should be remedied by the owner?!

peaceiris:
I hope so. Or the Hugo team gets that gohugoio account from the current owner.

PR #6413, opened 2019-10-11
tpl: After now accepts 0 as index

### LINKED ISSUE #6388: After should accept 0 as an index
### Issue Description
Currently [after](https://gohugo.io/functions/after) blows up if you pass in 0 as the index, which feels like a valid usage that should not crash things. The line that causes this is [here](https://github.com/gohugoio/hugo/blob/6027ee11082d0b9d72de1d4d1980a702be294ad2/tpl/collections/collections.go#L63). 

If this gets maintainer approval I'm happy to make the change myself, given that its only a line or two :).

<pre>
$ hugo version
Hugo Static Site Generator v0.58.1-24277B92 linux/amd64 BuildDate: 2019-09-06T09:19:04Z
</pre>

Yes

BaibhaVatsa:
Modified the if conditional because of which after threw an error if called with 0 as index. The function now returns the whole original slice if 0 is passed as an index. Also added tests to test the new behavior.

Fixes #6388


bep:
Looks great, thanks.

PR #6418, opened 2019-10-11
Add optional "title" attribute to iframe in Vimeo shortcode

zbayoff:
Add an optional "title" attribute to the iframe in the vimeo shortcode. If one is not given, the title attribute will default to "vimeo video". It is imperative for iframes to have a non-empty "title" attribute in order to meet WCAG2.0 accessibility guidelines. https://www.w3.org/TR/WCAG20-TECHS/H64.

### DIFF
--- a/tpl/tplimpl/embedded/templates/shortcodes/vimeo.html
+++ b/tpl/tplimpl/embedded/templates/shortcodes/vimeo.html
@@ -4,10 +4,10 @@
 {{ template "_internal/shortcodes/vimeo_simple.html" . }}
 {{- else -}}
 {{ if .IsNamedParams }}<div {{ if .Get "class" }}class="{{ .Get "class" }}"{{ else }}style="position: relative; padding-bottom: 56.25%; height: 0; overflow: hidden;"{{ end }}>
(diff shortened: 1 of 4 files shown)


tpl/tplimpl/embedded/templates/shortcodes/vimeo.html
```diff
@@ -4,11 +4,11 @@
 {{ template "_internal/shortcodes/vimeo_simple.html" . }}
 {{- else -}}
 {{ if .IsNamedParams }}<div {{ if .Get "class" }}class="{{ .Get "class" }}"{{ else }}style="position: relative; padding-bottom: 56.25%; height: 0; overflow: hidden;"{{ end }}>
-  <iframe src="https://player.vimeo.com/video/{{ .Get "id" }}" {{ if not (.Get "class") }}style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; border:0;" {{ end }}webkitallowfullscreen mozallowfullscreen allowfullscreen></iframe>
+  <iframe src="https://player.vimeo.com/video/{{ .Get "id" }}" {{ if not (.Get "class") }}style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; border:0;" {{ end }} {{ if .Get "title"}} title="{{ .Get "title" }}" {{ else }} title="vimeo video" {{ end }}  webkitallowfullscreen mozallowfullscreen allowfullscreen></iframe>
```
bep: This is a little matter of taste, but it gets shorter an easier to maintain if you write something ala:

```
{{ with .Get "title" }} {{ . }}{{ else }}{{ end }}
```

Other than that it looks good. But you need to run `mage generate` to regenerate the Go code. This will also tell us if there are test failures.

PR #6425, opened 2019-10-13
Revert "Add optional "title" attribute to iframe in Vimeo shortcode"

patti24k:
Reverts gohugoio/hugo#6418


bep:
You need to provide some context. Why?

zbayoff:
@bep Why was this reverted?

Issue #5642, opened 2019-01-27
Machine-readable version output


emilmelnikov:
Currently, the output of `hugo version` is not machine-readable:

```
$ hugo version
Hugo Static Site Generator v0.53/extended darwin/amd64 BuildDate: unknown
```

Machine-readable version output would make it easier to use Hugo from scripts.
See example in #5638 that started this discussion.

Hugo also has the `env` command whose output is used to report bugs:

```
$ hugo env
Hugo Static Site Generator v0.53/extended darwin/amd64 BuildDate: unknown
GOOS="darwin"
GOARCH="amd64"
GOVERSION="go1.11.4"
```

My suggestion is to split the info from the version line into something like the following, and also add more identifiable information (e.g. git commit hash).
However, I am not sure what the _extended_ version flavor means and how to represent it here.

```
VERSION="0.53"
BUILD_DATE=""
GIT_COMMIT="5b45961431f3ed3a0104c03a8e78657a6413fb01"
```

Then, the standard unix tools could be used to extract the raw values:

```
$ hugo env | grep GOVERSION | cut -d= -f2 | tr -d \"
go1.11.4
```

renantarouco:
Is this one still open for grab? I would like to start contributing and I think this one is a good start. The purpose of the project is awesome, and I found it right when I was looking to start writing some posts for practice.

I'm looking the other issues as well, but as a first-timer I think this one is achievable for me.

Thanks!

renantarouco:
Another question is if YAML would suit the needs and which information would be relevant. I see in some places people comparing the CLI tool with the docker one, sou I was kind of take inspiration in it. Is that ok?

patti24k:
yep

renantarouco:
Well, for the part of providing a machine readable output of the `hugo version` command I used the *yaml* package already imported by the project, and used its API to output it to os.Stdout, so it can be integrated with terminal pipes and etc.

```go
versionOut := yaml.MapSlice{
	yaml.MapItem{
		Key:   "Program",
		Value: "Hugo Static Site Generator",
	},
	yaml.MapItem{
		Key:   "OS/Arch",
		Value: runtime.GOOS + "/" + runtime.GOARCH,
	},
	yaml.MapItem{
		Key:   "Version",
		Value: hugo.CurrentVersion.String(),
	},
}
return yaml.NewEncoder(os.Stdout).Encode(versionOut)
```

Now the `hugo version` command gives me the following output:

```
[...]

PR #6428, opened 2019-10-15
Machine readable 'hugo version' and 'hugo env' commands #5642

renantarouco:
Changes to YAML the `hugo version` output and adds
version related info to the `hugo env` command.

See #5642


bep:
Thanks, the code looks fine, but I'm not merging this now. See #6298 5642

Issue #6069, opened 2019-07-01
Add an .editorconfig


coliff:
I think it'd be a good idea to add an `.editorconfig` to the projects root for contributors to easily see the code-formatting style. I imagine the file would be something like this:

```
# https://editorconfig.org

root = true

[*]
charset = utf-8
end_of_line = lf
indent_size = 2
indent_style = space
trim_trailing_whitespace = true

[*.go]
indent_size = 8
indent_style = tab

[*.js]
insert_final_newline = true

[*.md]
trim_trailing_whitespace = false
```

Ref: https://editorconfig.org/. Let me know what you think, I'd be happy to open a PR for this.

davidsneighbour:
Somebody familiar with the codebase should write down the requirements and check if the code is ready for this before "just" adding it. I broke a project recently with an activity like this :)

Other than this :+1:

PR #6461, opened 2019-10-29
Create .editorconfig

coliff:
Fix for: #6069

I think it'd be a good idea to add an `.editorconfig` to the projects root for contributors to easily see the code-formatting style. I imagine the file would be something like this but (of course) feel free to make any changes. There is an `.editorconfig` in the HugoDocs repo with the same config rules (https://github.com/gohugoio/hugoDocs/blob/master/.editorconfig)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# https://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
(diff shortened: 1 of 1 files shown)


coliff:
I think this PR is worthwhile being merged still.

bep:
I will close this, as the whitespace/newline handling in all of the test files etc. is very much dependent on the given situation.

PR #6464, opened 2019-10-31
Add goldmark renderer, with no options (yet)

### LINKED ISSUE #5963: Make Goldmark the new default markdown renderer
https://github.com/yuin/goldmark

evankanderson:
Applies to #5963 

I tested on a local build of https://knative.dev/docs/install/knative-with-aks/

Note that this does not yet include features like https://github.com/yuin/goldmark-highlighting

Also, error handling in the existing `RenderBytes` method seemed not great, so I just used a `panic`. I'm willing to plumb an `error` parameter higher if needed.

Signed-off-by: Evan Anderson <evan.k.anderson@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,6 +54,7 @@ require (
 	github.com/spf13/viper v1.4.0
 	github.com/tdewolff/minify/v2 v2.5.2
 	github.com/yosssi/ace v0.0.5
+	github.com/yuin/goldmark v1.1.0
 	go.opencensus.io v0.22.0 // indirect
 	gocloud.dev v0.15.0
 	golang.org/x/image v0.0.0-20190523035834-f03afa92d3ff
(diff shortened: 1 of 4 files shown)


bep:
Thanks for this. #5963 talks about making Goldmark the new default and that obviously means _more_ than this is. There are plenty of Hugo sites out there which we would hate to break (or I would because it creates lots of noise/issues/questions). So we need:

* It to similar enough in config/features, including ToC and footnotes. For the ToC I was thinking about creating a "content map" from the AST and let the end-user do `range .ContentTree.Headers` and similar.
* We probably want to introduce a new `markup` config element (name not sure) and only do `markup.goldmark` for the settings that are really unique for that renderer. We also want to create a config adapter from Blackfriday to the new format.
* We need a way for people to pick Blackfriday as the default, but we can probably use the front matter cascade thing for that for now (there is a `markup` setting).
* Given the above, we cannot keep on adding code to the `helpers` package. This code was originally lazily added to this package without much thinking and then extended.
* I started working on https://github.com/gohugoio/hugo/pull/6465 some time ago which outlines a new package structure

We can introduce the above a little by little, but we cannot make it the "new default" before most of the above is in place. And any new code should at least outline a new design/package structure.

evankanderson:
Yes, this wasn't intended to be a full fix for #5963; I updated the "Fixes" text to simply say "related to".

My thought was to:

1. Introduce this as a basic workaround. As an example, the page I had trouble with in #5963 renders fine (including TOC) when switching the default renderer to goldmark.

2. Introduce additional options for goldmark. I got started here, but I agree that some restructuring of `ContentSpec` would be handy.

3. Introduce a top-level flag default for the renderer (I couldn't find one today, but I am a hugo novice), which would default to `blackfriday`. At some point in the future, we would change the default of the flag to `goldmark`, and explain to existing users how to switch back to `blackfriday` after upgrade in the case of rendering issues. (I hesitate to even say "bugs" -- they might be deliberate design choices.)

I'd be happy to work (with guidance) on achieving some of the items laid out. It sounds like you have some concrete notions for the above (which I was hoping to find with my solicitation in https://github.com/gohugoio/hugo/issues/5963#issuecomment-537278949).

bep:
OK, I'll do a first a stab at pulling the existing "mess" into a package, and you can fit the initial Goldmark into that.

bep:
The refactoring work was a bigger task than I thought; I will add the Goldmark on top of that, with configuration. But thanks for this.

PR #6467, opened 2019-10-31
Update .travis.yml for arm64 support, etc.

anthonyfok:
- Add arm64 support for Linux thanks to Travis CI’s new feature, see
  https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support
- Force GO111MODULE=on
- Cache directories
- Switch to bionic (Ubuntu 18.04) environment
- Use "master" instead of "tip" to match the latest Travis CI docs
- ~~Use "x" in Go patch level: 1.12.x and 1.13.x~~
- Allow "mage -v check" to fail on non-AMD64 architectures
  because "go test -race ./..." currently fails on arm64 and ppc64le.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,59 @@
 language: go
-sudo: false
-dist: xenial
+
+dist: bionic
+
 env:
   global:
-  - GOPROXY="https://proxy.golang.org"
-  - HUGO_BUILD_TAGS="extended"
+    - CACHE_NAME=${TRAVIS_ARCH}
+    - GO111MODULE=on
+    - GOPROXY=https://proxy.golang.org
+    - HUGO_BUILD_TAGS=extended
+
 git:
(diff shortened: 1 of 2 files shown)


bep:
Please don’t do the x version thing. I want to test on whatever is the release version

anthonyfok:
> Please don’t do the x version thing. I want to test on whatever is the release version

Good call!  Thanks for letting me know.  I have now reverted that change.

anthonyfok:
Please note that `go test -race ./...` and `go test -race -run TestPageBundlerSiteRegular github.com/gohugoio/hugo/hugolib` succeeds on amd64 but fails on both arm64 and ppc64le, hence the `mage -v check || true` workaround for arm64 in .travis.yml.

I filed a report about this discrepancy at golang/go#35308.

Issue #6264, opened 2019-08-24
languageCode format is wrong


h0wardch3ng:
https://en.wikipedia.org/wiki/Language_code displays `en-US` is example for english.

In https://gohugo.io/getting-started/quick-start/#site-configuration

I think `languageCode = "en-us"` is wrong, the correct is `en-US`.

XhmikosR:
You should submit a PR upstream https://github.com/gohugoio/hugoDocs/blob/master/content/en/getting-started/quick-start.md and close this issue.

Las-Wonho:
I want to work on it. Can I submit for PR for this?

ky0nch3ng:
@XhmikosR @Las-Wonho Thx, I will.

Las-Wonho:
@ky0nch3ng Oh, "I will" mean is you submit PR?

PR #6469, opened 2019-11-01
[6264] Change language format code to en-US

mariCampos:
Issue #6264


bep:

bep:
OK, I'll notice that part of this PR belongs in this repo. If you could remove/revert those changes that lives below /docs it would be great.

PR #6483, opened 2019-11-06
Adding a page for external resources for Hugo including Hugo In Action.

atishay:
Looking for a place to link [Hugo In Action](https://www.manning.com/books/hugo-in-action) into the Hugo documentation website. This is a great place to have a discussion.


bep:

PR #6488, opened 2019-11-09
Add permalink configuration for non-zero padded dates

look:
This adds new permalink configuration to allow non-zero padded months (`:monthnopad`) and days (`:daynopad`) to be used in URLs. For example, `2018-01-06` can be rendered as `/2018/1/6`.

See: https://discourse.gohugo.io/t/implementing-additional-date-formats-for-permalinks/17860

I also looked into implementing this using `strftime` escapes, as hinted at by this TODO comment introduced way back in 07978e4a4922bc21c230fee65052232b829bd1ab:

https://github.com/gohugoio/hugo/blob/cafecca440e495ec915cc6290fe09d2a343e9c95/resources/page/permalinks.go#L217-L218

I did find a very nice [pure Go strftime library](https://github.com/lestrrat-go/strftime). However, it wasn't clear to me how to fit stftime escapes into the permalink configuration that exists today, so I went with the straightforward solution of adding new configuration.

### DIFF
--- a/docs/content/en/content-management/urls.md
+++ b/docs/content/en/content-management/urls.md
@@ -48,13 +48,19 @@ The following is a list of values that can be used in a `permalink` definition i
 : the 4-digit year
 
 `:month`
-: the 2-digit month
+: the 2-digit month (zero padded)
+
+`:monthnopad`
+: the 2-digit month with no zero padding
 
 `:monthname`
 : the name of the month
 
 `:day`
-: the 2-digit day
+: the 2-digit day (zero padded)
+
+`:daynopad`
+: the 2-digit day with no zero padding
 
 `:weekday`
 : the 1-digit day of the week (Sunday = 0)
--- a/resources/page/permalinks.go
+++ b/resources/page/permalinks.go
@@ -47,8 +47,10 @@ func NewPermalinkExpander(ps *helpers.PathSpec) (PermalinkExpander, error) {
 	p.knownPermalinkAttributes = map[string]pageToPermaAttribute{
 		"year":        p.pageToPermalinkDate,
 		"month":       p.pageToPermalinkDate,
+		"monthnopad":  p.pageToPermalinkDate,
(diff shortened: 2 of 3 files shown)


bep:
>I also looked into implementing this using strftime escapes,

That comment is very old, and I'm not sure how sensible it was.

What would probably make more sense was to use the date format built into Go, e.g. `2006-01-02`

Adding the `daynopad` etc. feels a little off; I guess it's OK if _this is it_, so to speak. But we could maybe also consider using the Go date constants, e.g `/2006/1/2` etc.

look:
Thanks for the quick feedback.

I'm in favor of providing a way for users to specify their own date format for permalinks, that feels the best to me. And given that the Go community really does not use strftime much, using the standard library's date formatting seems reasonable. 👍 

I'd be happy to work on that. However, I'm not sure how to configure that in the `permalinks` entry.  Do you have ideas?

This looks a little weird, but would match the `attributeRegexp` and then could be passed along for further processing.

```
permalinks:
  posts: /:format=2006/:format=1/:format=2/:slug
```

Or perhaps the config could optionally become a map (not sure how hard that would be to implement) to make the format string unambiguous.

```
permalinks:
  posts:
    format: /2016/1/2/:slug
  pages: /:title
```

Thoughts?

bep:
I was more thinking in the line of making "numbers" the special case, e.g.

```
permalinks:
  posts: /2006/1/2/:slug
```

look:
Ah yes, that makes sense. I will try that and see how it works out.

Do you think it will impact anyone who might be using a literal number in their permalinks? It would only be those that are for Go's special case date, I suppose.

bep:
> Do you think it will impact anyone who might be using a literal number in their permalinks?

I didn't think about those ...

We could do:

```
permalinks:
  posts: /:2006/:1/:2/:slug
```

look:
I like that. Giving it a try now.

look:
Closed in favor of #6489.

PR #6489, opened 2019-11-10
Support Go time format strings in permalinks

look:
In the vein of an [ancient TODO](https://github.com/gohugoio/hugo/commit/07978e4a4922bc21c230fee65052232b829bd1ab#diff-0688a3b65c7f5d01aa216f8d9b57fd00R111-R112) about supporting custom date formatting with strftime, this allows `:`-prefixed Go time format strings in permalink segments. This allows users to customize date-based permalinks any way they need to.

For example, with a date of 2019-11-09, the permalink `/:06/:1/:2` will render as `/19/11/9`.

See:

- [Discourse thread](https://discourse.gohugo.io/t/implementing-additional-date-formats-for-permalinks/17860)
- [Discussion of alternative approach](https://github.com/gohugoio/hugo/pull/6488)

### DIFF
--- a/resources/page/permalinks.go
+++ b/resources/page/permalinks.go
@@ -20,6 +20,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


resources/page/permalinks.go
```diff
@@ -173,8 +192,8 @@ func (l PermalinkExpander) validate(pp string) bool {
 		}
 
 		for _, match := range matches {
-			k := strings.ToLower(match[0][1:])
```
look: Lower casing the key turns out to be a bug: If you use `:TITLE` as the key in the permalink configuration, it will pass this check, but then when trying to expand the permalink, it will fail because `TITLE` isn't in the map and the code would panic.

I decided to remove this to lower because it makes date formats like `Friday` not work because `friday` is not a valid Go time format.

I added a test for the behavior that non-normalized permalink configuration keys aren't valid. It would pass on master as well, due to the bug noted above.

resources/page/permalinks.go
```diff
@@ -38,6 +39,24 @@ type PermalinkExpander struct {
...
+// with a Go time format always has a different output than the format itself.
+var referenceTime = time.Date(2019, time.November, 9, 23, 1, 42, 1, time.UTC)
+
+// Return the callback for the given permalink attribute and a boolean indicating if the attribute is valid or not.
+func (p PermalinkExpander) callback(attr string) (pageToPermaAttribute, bool) {
+	if callback, ok := p.knownPermalinkAttributes[attr]; ok {
+		return callback, true
+	}
+
+	if referenceTime.Format(attr) != attr {
```
bep: Smart!

look:
🙇 Thanks for your help with this! Appreciate it.

PR #6511, opened 2019-11-18
docs: Update documentation about slugs

look:
The documentation indicates the slug comes from the name of the file, but it comes from the title of the content (unless provided in frontmatter).

I was pretty confused by this because I had a `title` that contained a `.` and it was used in the URL even though the file name did not.

This information in the [Permalink Configuration Values](https://gohugo.io/content-management/urls/#permalink-configuration-values) section is correct on this point.


bep:

PR #6527, opened 2019-11-23
Minor cleanups

moorereason:
Use fmt.Fprint were appropriate instead of Fprintf.  Remove some unused
code.


bep:
Thanks. Note that the `temphighlighting` package will be pushed upstream in a few days. I kind of lost patience when developing the needed upstream changes in Chroma + Highlighting. Patience when waiting for days for people to merge and tag releases over timezones isn't my strongest character trait ...

Issue #5695, opened 2019-02-14
AsciiDoctor: set working dir to dir of current content


fapdash:
## Report

I'm currently evaluating hugo for our asciidoc based website.
As described in #4344, currently relative includes in adoc files are evaluated relative to the hugo root dir, which defeats the purpose of a relative include.

Would you be willing to change this so that the working directory for calls to the externel tool (AsciiDoctor in this case) is set to the directory of the content file that gets converted? Or give this as an option?

### Content file
```adoc
---
title: "Hello, world"
date: 2019-02-13T15:06:05+01:00
---

== Hello, world

include::to_include.adoc[]
```

### to_include.adoc
```adoc
== Abc

```

### Error
`ERROR 2019/02/14 11:26:28 posts/my-first-post/post.adoc: asciidoctor: ERROR: <stdin>: line 9: include file not found: /home/fap/repos/hugo_test/quickstart/to_include.adoc`

celtic-coder:
It would be great if this could be setup as an option. Thanks!

patti24k:
I made a mistake could you please correct my mistake  #5836  sorry for not replying right away very busy I'm trying to move can you please help

muenchhausen:
I created a very first solution for this problem. Still for discussion, because I think some more discussions are required. Anyhow: This solution works so far (without backward compatibility).

muenchhausen:
I have now improved the solution in pull request #6561

celtic-coder:
Hi Derk (@muenchhausen),

Is there any further progress on this PR? 

Kind Regards, 
Liam

muenchhausen:
I have the feeling, that the pull request [6561](https://github.com/gohugoio/hugo/pull/6561) will be merged soon :)

benibilme:
Hello, 
I have the same problem. I have hugo web-site which shares documentation with antora web site. I really need this feature as well.

PR #6561, opened 2019-11-29
feat(asciidoc): set working dir to current content

muenchhausen:
This pull request will solve the relative path problem with ``asciidoctor`` tooling. An ``include`` will resolve relatively, so you can refer easily to files in the same folder.

Also ``asciidoctor-diagram`` and PlantUML rendering works now, because the created temporary files will be placed in the correct folder.

This patch would be a breaking change, because you need to correct all your absolute ``include`` pathes to relative pathes, so using relative pathes must be configured explicitly.

This patch covers just the Ruby version of ``asciidoctor``. The old AsciiDoc CLI EOLs in Jan 2020, so we left that implementation untouched.

related #5695


markup/asciidoc/convert.go
```diff
@@ -64,7 +65,8 @@ func (a *asciidocConverter) getAsciidocContent(src []byte, ctx converter.Documen
 	}
 
 	a.cfg.Logger.INFO.Println("Rendering", ctx.DocumentName, "with", path, "...")
-	args := []string{"--no-header-footer", "--safe"}
+	folderName := filepath.Dir(filepath.Join(a.cfg.Cfg.GetString("contentDir"), ctx.DocumentName))
```
bep: There isn't ONE config dir. We could probably add `Filename` (full path) to the `DocumentContext` (but should also then rename the existing attributes to reflect what they contain; `DocumentName` => Path).

muenchhausen: The working dir problem is not just to get asciidoc `include::somefile.adoc[]` working. Also `asciidoctor-diagram` rendered images in the wrong folder. 

My proposal to solve that is exposing all asciidoctor args to `config.toml` and adding `--base-dir` and `outdir=` to asciidoctor args if desired. 

So this pull request includes:
 
1. New configuration parameters for `config.toml` e.g.:
```
asciidoctorArgs = ["--no-header-footer", "-r", "asciidoctor-html5s", "-b", "html5s", "-r", "asciidoctor-diagram"]
asciidoctorCurrentContent = true
```

2. Unit tests that prove that these changes can be configured, so we stay backward compatible (changing the working dir is a breaking change!)

3. I tested that in a sample repo https://github.com/muenchhausen/hugo_asciidoc_ddd

What is open? 
- I am not sure if the new configuration parameters follow the hugo naming convention - please check! 
- Please let me know, where I can document the new configuration options.

bep: >I am not sure if I fully understood @bep `s comment - I will try to dig deeper

I suspect that, too. I will try to explain better.

* Hugo can have many content dirs (while the common thing is to have 1). Hugo Modules means that this can be "virtual dirs" on GitHub.
* Given the above, the only current way to get the real filename to a content file is to look at the `Filename`. We have a `DocumentContext` with some badly named fields, but `DocumentName`I think is filled with `Path`; this is the file's path relative to the content dir.  So, to get to the real filename, I think the simplest is if you add `Filename` to the `DocumentContext` (you could probably also do a `.ContentFs.Stat`, but that is a little bit ugly).

A final question: I don't use Asciidoctor myself, and have not invested too much time in it: If we add some flags we must make sure that they don't break if run at a reasonably recent version of both asciidoc and asciidoctor... So, if the above flags is for asciidocor, but not defined in asciidoc ...

muenchhausen: Thanks for explaining! I think adding `FileName` to `DocumentContext` is the correct proposal for converters that need to reference files. This works with modules very well.

I wrote some tests to ensure that `asciidoc` still works. Anyhow, the new features just covers `asciidoctor` because I have just experience with that.

The naming of the configuration parameters are improved now.

Open: Documentation. Should we write a dedicated chapter on `asciidoc`? What are your thoughts?

muenchhausen: AsciiDoc EOLs in Jan. I suggest to depecated this for Hugo from now on.

I renamed FileName to Filename. 

I added documentation today - please have a look if this is fine!

muenchhausen: @bep I would like to deploy this patch to 3 installations this week, test it and continue on some new improvements. Is there anything open I have to improve?

weissjoh:
Great work. The patch should be accepted as early as possible as it contains a breaking change. If the fixed behavior is the intended one it should be corrected very soon.

timowolf:
I really would appreciate the merge of the request. The different handles of the path drives me crazy.

xtarx:
The related issue #5695 is open since almost a year. I would appreciate this pull request, because I really need asciidoctor to work with hugo properly.

bep:
I don't know enough about this to merge this. @anthonyfok have worked on this before, and may have time to take a look.

I have one general comment:

* Any added config for this should follow the config for the others in the `markup` package. I don't want a new top level config option.
* Also, that new config section could be named `asciidocext` or something, to save the spot for a native solution.

muenchhausen:
I implemented @bep 's suggestions.  @anthonyfok - I am looking forward to get your feedback.

muenchhausen:
@anthonyfok you are very busy - let me know if I can help you to review other pull requests so you have some time for mine :)

azah:
I tried using @muenchhausen's fork with the following config and graphviz:
```toml
  [markup.asciidocext]
    args = ["--no-header-footer", "-r", "asciidoctor-html5s", "-b", "html5s", "-r", "asciidoctor-diagram"]
    workingFolderCurrent = true
```
setting `workingFolderCurrent` does indeed generate the file, but it gets generated in my `content/posts` folder. The generated HTML looks like:
```html
<div class="image-block">
  <img src="example.png" alt="example" width="340" height="347">
</div>
```
but it doesn't find the image.

muenchhausen:
@azah thanks testing and sharing your results!! Your site repo is quite complex - I wasn't able to run it quickly. Anyhow: I took over your graphviz sample to my simple testing repo [hugo_asciidoc_ddd](https://github.com/muenchhausen/hugo_asciidoc_ddd). Here graphviz worked very fine. 

With your help, I found a bug with wrong adoc --base-dir folders! So maybe this will also fix your problem. Can you please pull my changes and re-test your issue?

azah:
@muenchhausen I re-built hugo with your changes and ran it again, but I could not get the diagrams to load.

Issues I encountered:
* running `hugo` generated `chapter1` and `chapter3` folders in the root
* I had to change `destination` to `public` in `config.toml` in order to get it to not make those folders in the root. Is this documented anywhere?

Changing `destination` generates the diagrams in the right folders in `public/`, but when I visit the page I still don't see them. For example, the generated html which points to `http://localhost:1313/chapter2/example.svg` returns a 404.

muenchhausen:
@azah: I was able to reproduce the issue. Can you try to start hugo with the command line argument `-d` or `--destination` (instead of defining that in config.toml) ? 
e.g. run ```hugo -v --destination ./public server```.

azah:
@muenchhausen running that command worked.

muenchhausen:
@bep, @anthonyfok : This patch is running since 4 month on several installations without any problems. I would appreciate to finalize this pull request.

bep:
Hey, and sorry for the delay on this.

The code looks great, my only concern now is that it does much more than the title says.

```toml
[markup.asciidocext]
    args = ["--no-header-footer", "-r", "asciidoctor-html5s", "-b", "html5s", "-r", "asciidoctor-diagram"]
    workingFolderCurrent = true
```

So, the last part (workingFolder ...) is great. 

The args part is, I assume, allowing arbitrary arguments sent to the shell. And this is not something I'm prepared to open up for, from a security standpoint. You can probably argue that it would "be fine", but when someone question it, it's probably worth spending time to think about. For what I now Asciidoctor could have a "--exec" flag allowing running arbitrary programs.

So I suggest you either remove that option for now or possibly making it less flexible:

```toml
[markup.asciidocext]
    noHeaderOrFooter = false
...
    workingFolderCurrent = true
```

snipem:
@muenchhausen So following up to this comment, I suggest that only the extensions to be included should be configurable. In this manner:

```yaml
[markup.asciidocext]
    noHeaderOrFooter = false
    extensions
    - asciidoctor-html5s
    - asciidoctor-diagram
    workingFolderCurrent = true
```

The other settings like `--no-header-footer` or `-b html5s` are static and should not be configurable for said reason.

The list of extensions will not run arbitrary code in any case. Even if someone replaces `asciidoctor-html5s` with `asciidoctor-html5s"; rm -rf /`. I haven't tested this but I assume it's the behavior of https://golang.org/pkg/os/exec/#Cmd.Run used here: https://github.com/gohugoio/hugo/blob/5f6b6ec68936ebbbf590894c02a1a3ecad30735f/markup/internal/external.go#L17. `Cmd.Run` uses low level process spawning which should not be vulnerable to shell injection. And if `-r` is prefixed in front of every free defined extensions, only asciidoctor should complain about an unknown extension in case of error.

To improve security, you could also use a whitelist of accepted extensions. Here is a list of extensions which can be easily stored in Hugos code as a whitelist: https://asciidoctor.org/docs/extensions/

bep:
```toml
    extensions
    - asciidoctor-html5s
    - asciidoctor-diagram
```

Any extensionlist have to be in a whitelist. I'm not prepared to spend the time investigating some other way of doing it.

azah:
A lot of those extensions are for AsciidoctorJ[ava]. The current assumption is that the installed asciidoctor is the ruby one. Should we support both?

snipem:
> A lot of those extensions are for AsciidoctorJ[ava]. The current assumption is that the installed asciidoctor is the ruby one. Should we support both?
> 
> Edit: I have no experience with maven/asciidoctorj so I can't test it.

Thanks for the hint. You're right. I haven't looked at the list throughly. I go with the current assumption: the ruby one and it's extensions. Most users are hoping for the asciidoctor-diagram extension as far as I know. I've never used the Java implementation nor it's extensions.

azah:
@snipem @bep thoughts on how I restructured it here? https://github.com/azah/hugo/commit/ac8f2784e2276dca358e5a3fc441197d7fa3b609

* backends, safe-modes, and extensions are now whitelisted to the popular (ruby) extensions and valid values.
* the default for extensions is to not enable any, because they're all installed separately so a build would break if referenced.
* the default backend is `html5` because asciidoctor does not install `html5s` by default.

it seems that the cfg parsing works correctly, I just get an error now:
```
INFO 2020/04/21 17:19:58 destinationDir ./public

INFO 2020/04/21 17:19:58 Rendering posts/2019/12-21_home_server_with_freenas.adoc with /home/andrew/.rbenv/shims/asciidoctor using asciidoc args [-b html5s -r asciidoctor-html5s -r asciidoctor-diagram --no-header-footer --safe-mode safe --base-dir /home/andrew/website/hugo-site/content/posts/2019 -a outdir=/home/andrew/website/hugo-site/public/posts/2019 -] ...

ERROR 2020/04/21 17:19:59 posts/2019/12-21_home_server_with_freenas.adoc: asciidoctor: WARNING: path is outside of jail; recovering automatically
```

muenchhausen:
> WARNING: path is outside of jail;

we need to support rendering of asciidoctor also from various modules and folders, so we have to run asciidoctor without `--safe-mode safe` so the default unsafe is used.

azah:
@muenchhausen setting `safeMode = unsafe` in my config.toml does fix that error, but the issue is still that the generated html links to to `public/posts/<year>/<post-name>/example.svg` but the images get saved one directory higher: `public/posts/<year>/example.svg`. I haven't delved deep enough into hugo's internals to figure out how (safely) get the rendered object's slug/name to figure out the final generated folder.

pgorod:
Don't forget this, the native Go implementation of Asciidoc (which has been advancing):

https://discourse.gohugo.io/t/libasciidoc-a-golang-asciidoc-renderer/21702/3

The Ruby Asciidoctor is slooooow, especially when used with Hugo (I'm talking about a psychological effect here 🤣)

Regarding this PR here, I just thought I'd leave this comment to encourage that we keep in mind `asciidoc rendering <> Ruby asciidoctor`. 

In that thread I linked above, I asked (back in January) about whether everything in already in place (from Hugo side) so that we can test the Go-based libasciidoctor, but I didn't get a reply. I don't want to rush anyone here, but when it is ready enough for a test-drive, I'd like to try it with my own site.

Thanks!

bep:
@pgorod which is why the config element in question is suffixed `ext` ... to make room for the future.

muenchhausen:
This pull request still supports the [old Python based AsciiDoc](https://github.com/asciidoc/asciidoc). That came to end of life in Jan 2020. Is it ok, if I remove the sources to simplify the implementation? See e.g. in convert.go - `func getAsciidocExecPath()` became obsolete.

muenchhausen:
> @muenchhausen setting `safeMode = unsafe` in my config.toml does fix that error, but the issue is still that the generated html links to to `public/posts/<year>/<post-name>/example.svg` but the images get saved one directory higher: `public/posts/<year>/example.svg`. I haven't delved deep enough into hugo's internals to figure out how get the document's slug/name (if it exists) to figure out the final generated folder.

I am not able to reproduce that issue. Is the svg post related to asciidoctor? Can you point me to your sample?

azah:
@muenchhausen I created a new hugo site repo with 1 post: https://github.com/azah/hugo-asciidoc-test

the changes I made to the default config.toml are:
```toml
destination = "./public"

[markup.asciidocext]
    args = ["--no-header-footer", "-r", "asciidoctor-html5s", "-b", "html5s", "-r", "asciidoctor-diagram"]
    workingFolderCurrent = true
```
rendering this puts `example.svg` in `public/posts/2020`

muenchhausen:
> rendering this puts `example.svg` in `public/posts/2020`

Ah ok - this is the issue we were discussing before. It can be solved by setting explicitly the hugo arg `--destination ./public` - right? I do not understand why hugo handles `destination`as configuration differently than using command line arg. I did not find any possible fix in the code for that. I will need help for that.

azah:
@muenchhausen actually I'm not sure what happened before, but even with `--destination ./public` it still puts the files in `public/posts/2020`.

azah:
@bep in `asciidoc/convert.go` is it possible to check if the `ctx.Document` is a `page.Page` without causing a cyclic import?

Or is there another way to get the final folder name for the generated post?

bep:
You can define the subset of that interface you're interested in locally (private) and do a type cast.

azah:
@bep where does hugo determine the resulting folder name? I was looking at `resources/page/page.go` but I don't see where it gets determined.

azah:
@muenchhausen @bep I got it working so far with this:

```go
type pageSubset interface {
	Permalink() string
}
```

```go
func (a *asciidocConverter) getAsciidoctorArgs(ctx converter.DocumentContext) []string {
	args := make([]string, len(a.cfg.MarkupConfig.AsciidocExt.Args))
	copy(args, a.cfg.MarkupConfig.AsciidocExt.Args)

	workingFolderCurrent := a.cfg.MarkupConfig.AsciidocExt.WorkingFolderCurrent

	if workingFolderCurrent {
		contentDir := filepath.Dir(ctx.Filename)
		destinationDir := a.cfg.Cfg.GetString("destination")

		a.cfg.Logger.INFO.Println("destinationDir", destinationDir)
		if destinationDir == "" {
			a.cfg.Logger.ERROR.Println("markup.asciidocext.workingFolderCurrent requires hugo command option --destination to be set")
		}

		postDir := ""
		page, ok := ctx.Document.(pageSubset)
		if ok {
			postDir = filepath.Base(page.Permalink())
		} else {
			a.cfg.Logger.ERROR.Println("unable to cast interface to pageSubset")
		}

		outDir, err := filepath.Abs(filepath.Join(destinationDir, filepath.Dir(ctx.DocumentName), postDir))
		if err != nil {
			a.cfg.Logger.ERROR.Println("asciidoctor outDir: ", err)
		}

		args = append(args, "--base-dir", contentDir, "-a", "outdir="+outDir)
	}

	return args
}

```

This generates the asciidoctor-diagram images in the correct folders:
```
./public/posts/2020/asciidoc_hugo_test_2/example2.png
./public/posts/2020/04-26_test/example.png
```

however when I use `hugo server` the images don't get loaded.

bep:
@azah you cannot use Permalink to do file path logic. I'm not sure what you're trying to do here, but if the goal is to find the "content dir relative" path, I think `.Path()` will be the correct.

azah:
@bep I want to put the generated image in the same directory as the generated post. `Path()` does not work because the generated directory can be different if the user sets a custom slug.

See the below: I set a custom slug for one of the posts, which are otherwise named `04-26_test.adoc` and `04-27_test.adoc`. `.Path()` wouldn't work here.
```
public
├── posts
│  ├── 2020
│  │  ├── 04-26_test
│  │  │  ├── example1.png
│  │  │  └── index.html
│  │  └── asciidoc_hugo_test_2
│  │     ├── example2.png
│  │     └── index.html
│  ├── index.html
│  ├── index.xml
```

bep:
OK, then I understand, but you need to use `RelPermalink` and not `Permalink`.

azah:
:+1: here's the updated code to use `.RelPermalink()`:
```go
package asciidoc

type pageSubset interface {
	RelPermalink() string
}
```
```go
func (a *asciidocConverter) getAsciidoctorArgs(ctx converter.DocumentContext) []string {
	args := make([]string, len(a.cfg.MarkupConfig.AsciidocExt.Args))
	copy(args, a.cfg.MarkupConfig.AsciidocExt.Args)

	workingFolderCurrent := a.cfg.MarkupConfig.AsciidocExt.WorkingFolderCurrent

	if workingFolderCurrent {
		contentDir := filepath.Dir(ctx.Filename)
		destinationDir := a.cfg.Cfg.GetString("destination")

		a.cfg.Logger.INFO.Println("destinationDir", destinationDir)
		if destinationDir == "" {
			a.cfg.Logger.ERROR.Println("markup.asciidocext.workingFolderCurrent requires hugo command option --destination to be set")
		}

		postDir := ""
		page, ok := ctx.Document.(pageSubset)
		if ok {
			a.cfg.Logger.INFO.Println("path: ", page.RelPermalink())
			postDir = filepath.Base(page.RelPermalink())
		} else {
			a.cfg.Logger.ERROR.Println("unable to cast interface to pageSubset")
		}

		outDir, err := filepath.Abs(filepath.Join(destinationDir, filepath.Dir(ctx.DocumentName), postDir))
		if err != nil {
			a.cfg.Logger.ERROR.Println("asciidoctor outDir: ", err)
		}

		args = append(args, "--base-dir", contentDir, "-a", "outdir="+outDir)
	}

	return args
}
```
@muenchhausen can you test this out?

muenchhausen:
@azah: you patch works fine! Should I add it to this branch?

azah:
@muenchhausen yep that's fine.

muenchhausen:
@azah: I successfully rebased my changes on your pull request #7208! 
Currently I still have this in a [branch,](https://github.com/muenchhausen/hugo/tree/asciidoc_config_7208) because your RelPermalink has problems with asciidoctor-diagram images. I try to investigate on this the next days.

azah:
@muenchhausen what issue are you getting?

snipem:
@muenchhausen Any way we can help?

muenchhausen:
@snipem thanks - of course :) I will have time tonight and dig into the problem and try to describe it. Sorry for the delay...

muenchhausen:
@azah the RelPermalink patch just causes problems with index.adoc and _index.adoc. Here the folder name will be duplicated. A possible solution is the following:

```golang
func (a *asciidocConverter) parseArgs(ctx converter.DocumentContext) []string {
	...
		var outDir string
		var err error

		file := filepath.Base(ctx.Filename)
		if file == "_index.adoc" || file == "index.adoc" {
			outDir, err = filepath.Abs(filepath.Dir(filepath.Join(destinationDir, ctx.DocumentName)))

		} else {
			postDir := ""
			page, ok := ctx.Document.(pageSubset)
			if ok {
				postDir = filepath.Base(page.RelPermalink())
			} else {
				a.cfg.Logger.ERROR.Println("unable to cast interface to pageSubset")
			}

			outDir, err = filepath.Abs(filepath.Join(destinationDir, filepath.Dir(ctx.DocumentName), postDir))
		}

		if err != nil {
			a.cfg.Logger.ERROR.Println("asciidoctor outDir: ", err)
		}

		args = append(args, "--base-dir", contentDir, "-a", "outdir="+outDir)
	...
}
```

See here for the full code: [convert.go](https://github.com/muenchhausen/hugo/blob/asciidoc_config_7208/markup/asciidocext/convert.go).

Do you know how to switch to my new branch [asciidoc_config_7208](https://github.com/muenchhausen/hugo/tree/asciidoc_config_7208) to be used for this pull request?

muenchhausen:
I would like to close this pull request, because I combined #7208 and this one into the new pull request #7281. Please follow #7281 from now on!

PR #6575, opened 2019-12-02
deps: Update chroma

jpienaar:
Updating chroma module to head to include some new language support added in chroma. Tried removing unused parts from go.sum but v0.7.0/go.mod gets re-added but seems to get added back when I run `go test ./...` and I'm not sure why.

Saw hugo does allow for packages to be referenced in between releases, but not sure if there is any other criteria.


bep:
This needs to be a tagged release.

PR #6591, opened 2019-12-09
Add inline template function to complement markdownify

### LINKED ISSUE #5975: Allow markdownify as a block for .Inner
Typical construction  is:

```
{{ .Inner | markdownify }}
```

The current logic is that we remove surrounding `<p></p>` if there is only one paragraph.

For people doing `{{ .Inner }}` => `{{ .Inner | markdownify }}` because of the changes made in 0.55, this behaviour would be subtle breaking. 

I don't think we'd want this for every `.Inner` (that would breaking it for others), but I suggest we add a new `block` keyword (it has been discussed before), so you would then do:

```
{{ .Inner | block | markdownify }}
```

Also see https://github.com/gohugoio/hugo/issues/3040

moorereason:
Add inline template function to handle the TrimShortHTML call, and
remove the TrimShortHTML call from markdownify.  To get the old behavior
do `. | markdownify | inline`.

Fixes #5975

### DIFF
--- a/tpl/strings/init.go
+++ b/tpl/strings/init.go
@@ -56,7 +56,8 @@ func init() {
 			[][2]string{
 				{
 					`{{ findRE "[G|g]o" "Hugo is a static side generator written in Go." "1" }}`,
-					`[go]`},
+					`[go]`,
+				},
 			},
 		)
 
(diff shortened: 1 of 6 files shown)


bep:
If we want to do this, I fear we need to do this the other way around. The breaking change is one thing (I can take that shouting), but as discussed in the issue, I also think the current inline behaviour is what most people want in most situations. `markdownify` is typically used for titles etc.

To take this discussion even further, I have a `.RenderString` in another branch that is supposed to be output format agnostic -- and I have a little problem seeing how you can do `.RenderString  | inline` without taking the output format into consideration.

I think this needs to wait a little.

PR #6637, opened 2019-12-19
tpl: Support GitInfo for content folders

### LINKED ISSUE #5533: Support GitInfo for content folders in different Git repositories

asafg6:
Add an option to get git metadata from the content dir instead of the
working dir

Fixes #5533

### DIFF
--- a/docs/content/en/variables/git.md
+++ b/docs/content/en/variables/git.md
@@ -56,3 +56,9 @@ The `GitInfo` object contains the following fields:
 If the `.GitInfo` feature is enabled, `.Lastmod` (on `Page`) is fetched from Git i.e. `.GitInfo.AuthorDate`. This behaviour can be changed by adding your own [front matter configuration for dates](/getting-started/configuration/#configure-front-matter).
 
 [configuration]: /getting-started/configuration/
+
+## Using contentDir instead of workingDir
+
+If your content directory is under a submodule, you can add `useContentDirGitInfo` to `true` to your [site's configuration file][configuration].
+Hugo will then get its content metadata from the content repository (instead of the repository in the working directory).
+
--- a/hugolib/gitinfo.go
+++ b/hugolib/gitinfo.go
@@ -36,9 +36,16 @@ func (g *gitInfo) forPage(p page.Page) *gitmap.GitInfo {
 }
 
 func newGitInfo(cfg config.Provider) (*gitInfo, error) {
-	workingDir := cfg.GetString("workingDir")
+	useContentGitInfo := cfg.GetBool("useContentDirGitInfo")
+	var gitDir string
+	if useContentGitInfo {
+		gitDir = cfg.GetString("contentDir")
+	} else {
+		gitDir = cfg.GetString("workingDir")
 
-	gitRepo, err := gitmap.Map(workingDir, "")
+	}
+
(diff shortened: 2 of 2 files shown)


bep:
He, and thanks for this.

While I do agree that we need something like this, using `contentDir` is too limiting. That setting still works, but it was replaced by the `mounts` slice recently. The way I see it we need to handle:

* Multiple content directories
* A way to detect if a content dir is a Git repo
* A way to handle the special union mounts we have

asafg6:
hi @bep 
I didn't know about `mounts` :) I see that they allow a lot of flexibility around directory config. however i believe this setting will very much appeal more simple users like myself

asafg6:
any chance of merging this?

docs/content/en/variables/git.md
```diff
@@ -56,3 +56,9 @@ The `GitInfo` object contains the following fields:
 If the `.GitInfo` feature is enabled, `.Lastmod` (on `Page`) is fetched from Git i.e. `.GitInfo.AuthorDate`. This behaviour can be changed by adding your own [front matter configuration for dates](/getting-started/configuration/#configure-front-matter).
 
 [configuration]: /getting-started/configuration/
+
+## Using contentDir instead of workingDir
+
+If your content directory is under a subomdule, you can add `useContentDirGitInfo` to `true` to your [site's configuration file][configuration].
```
coliff: I think 'subomdule' should be be 'submodule'?

docs/content/en/variables/git.md
```diff
@@ -56,3 +56,9 @@ The `GitInfo` object contains the following fields:
...
+Hugo will then get it's content metadata from the content repository (instead of the repository in the working directory).
```
coliff: it's should be its I think (no apostrophe)

bep:
>any chance of merging this?

Not in the current state, no. See my comment above.

docs/content/en/variables/git.md
asafg6: done

docs/content/en/variables/git.md
asafg6: done

asafg6:
any updates?

asafg6:
Anyone considering this?

NorseGaud:
This is a huge need for us!

deanishe:
I submitted a patch for this in 2018 (#5535) and it was ignored for over a year, then rejected in favour of some feature that will fix the problem at some point in the future.

So I wouldn't hold your breath…

asafg6:
@bep is there an alternative solution for that problem?

deanishe:
@asafg6 my patch still works if you don't mind applying it and building Hugo yourself.

NorseGaud:
Bump

PR #6640, opened 2019-12-19
Fix incorrect MIME type from image/jpg to image/jpeg

zaitseff:
According to multiple sources, both official ([IANA] and [RFC2046]) and
otherwise (eg, [Mozilla] and [Wikipedia]), the official MIME type for
JPEG images is `image/jpeg`, not `image/jpg`.  Change Hugo to match.

[IANA]: https://www.iana.org/assignments/media-types/media-types.xhtml#image
[RFC2046]: https://tools.ietf.org/html/rfc2046
[Mozilla]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
[Wikipedia]: https://en.wikipedia.org/wiki/JPEG

### DIFF
--- a/docs/data/docs.json
+++ b/docs/data/docs.json
@@ -1501,10 +1501,10 @@
         ]
       },
       {
-        "type": "image/jpg",
-        "string": "image/jpg",
+        "type": "image/jpeg",
+        "string": "image/jpeg",
         "mainType": "image",
-        "subType": "jpg",
+        "subType": "jpeg",
         "delimiter": ".",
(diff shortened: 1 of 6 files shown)


docs/data/docs.json
```diff
@@ -1501,10 +1501,10 @@
         ]
       },
       {
-        "type": "image/jpg",
```
bep: Just a heads up for later: docs.json is autogenerated from the source.

PR #6654, opened 2019-12-21
Add title parameter to YouTube shortcode to allow overriding the default

azenk:
Allowing the user to specify a meaningful `title` improves the [accessibility](https://dequeuniversity.com/tips/provide-iframe-titles) for YouTube videos.  If no `title` is specified, the current behavior is preserved.


obar:
Looks like a useful addition. Even if the person making the website is the video uploader on youtube, it would often be the case that the title on youtube (where you want the video to show up in search) would be different than the title the author would want for the embedded video on their site.

Not only that, it would make the `youtube` shortcode consistent with the `vimeo` shortcode which [already has a title parameter](https://gohugo.io/content-management/shortcodes/#example-vimeo-output.html).

What is blocking this from being merged?

Issue #6636, opened 2019-12-19
Improve warning for missing layout


cup:
When missing layouts, Hugo presents this output:

~~~
WARN 2019/12/18 19:42:47 found no layout file for "HTML" for "home": You should
create a template file which matches Hugo Layouts Lookup Rules for this
combination.

WARN 2019/12/18 19:42:47 found no layout file for "HTML" for "page": You should
create a template file which matches Hugo Layouts Lookup Rules for this
combination.

WARN 2019/12/18 19:42:47 found no layout file for "HTML" for "section": You
should create a template file which matches Hugo Layouts Lookup Rules for this
combination.

WARN 2019/12/18 19:42:47 found no layout file for "HTML" for "taxonomyTerm": You
should create a template file which matches Hugo Layouts Lookup Rules for this
combination.
~~~

However, this output is not quite helpful. A skilled user might find this page:

https://gohugo.io/templates/lookup-order

but a novice user might not, in which case they will have no idea how to fix the
problem. Further, even if they find the page, its not clear still how to resolve
the issue. For example if you search `taxonomyTerm`, it doesnt take you to a
section that lists proper paths. Youd have to know to navigate to this section:

https://gohugo.io/templates/lookup-order#examples-layout-lookup-for-taxonomy-terms-pages

I think it would be more helpful for Hugo to emit these anchored links. That way
upon error a person can navigate directly to the section of the help page to
solve the problem.

Another option would be for Hugo to simply emit what its searching for,
something like:

~~~
WARN 2019/12/18 19:42:47 found no layout file for "HTML" for "taxonomyTerm" in
these paths: [layouts/categories/category.terms.html.html
layouts/categories/terms.html.html layouts/categories/list.html.html
layouts/categories/category.terms.html layouts/categories/terms.html
layouts/categories/list.html layouts/taxonomy/category.terms.html.html
layouts/taxonomy/terms.html.html layouts/taxonomy/list.html.html
layouts/taxonomy/category.terms.html layouts/taxonomy/terms.html
layouts/taxonomy/list.html layouts/category/category.terms.html.html
layouts/category/terms.html.html layouts/category/list.html.html
layouts/category/category.terms.html layouts/category/terms.html
layouts/category/list.html layouts/_default/category.terms.html.html
layouts/_default/terms.html.html layouts/_default/list.html.html
layouts/_default/category.terms.html layouts/_default/terms.html
layouts/_default/list.html]
~~~

[...]

PR #6661, opened 2019-12-22
Make the missing layout file warning more verbose

zaitseff:
Print the full list of template file paths being searched when an
appropriate layout file is not found, with the advice that one of these
missing files needs to be created.

Closes: #6636


bep:
Please don't shortcut open discussions with a PR.

zaitseff:
Sorry, not my intention to shortcut discussions; rather just to provide a suggested patch that you are free to accept or reject.  Still learning my way around how you prefer to do development!

PR #6680, opened 2019-12-29
Recursively merge if both sides are maps

### LINKED ISSUE #5192: Recursively merge configurations
Configuration files are currently merged by Hugo by looping through the keys of a given map and overwriting the values for those keys in another map. This is not done recursively, which leads to scenarios where themes set a bunch of default configuration options nested in a maps under e.g. `params.foo.bar` with the expectation that a user can override an option at `params.foo.bar.baz` while all of the other defaults remain intact.

Instead, all of the theme settings under `params.foo` are wiped out and replaced with a nearly-empty map containing only `params.foo.bar.baz`.

My theme relies heavily on configurations being applied recursively as in the first paragraph, as it has a plugin system that has settings for the bundled plugins. I currently have to manually namespace all of the theme settings and, in the templates, determine whether a setting is set in the user configuration and use the theme namespaced one otherwise. It would greatly reduce boilerplate in the theme to be able to assume things got merged recursively.

The code for merging the maps should include a check to see if a value matching the key is a `map[string]interface{}` and, if so, recursively merge that map first. Everything else, including slices, should remain copied as-is.

There was a [thread](https://discourse.gohugo.io/t/cannot-access-theme-params/11730/5) on the Hugo forums about this, however I never got a response after I clarified how I expected the merging to work and even provided some code demonstrating it. This issue is a follow-up to that thread.

Shadow53:
Recursively merging ensures that modules get to keep the configuration
items they set, while allowing the user config to override anything.
Recursion only happens if the key in both user and module configurations
are maps.

Fixes #5192

I added a test for the recursive merging function, including making sure that map keys explicitly set to `nil` do not get overwritten. I did not do anything with testing that merged configurations were otherwise "correct"/as expected, as there do not seem to have been tests for that previously. At least, it didn't break the ones that exist.

I also did not update documentation as I am not sure where that would be added outside of release notes. I am more than willing to make an update there if the maintainers deem that necessary, though.

### DIFF
--- a/hugolib/config.go
+++ b/hugolib/config.go
@@ -539,6 +539,35 @@ func (l configLoader) applyThemeConfig(v1 *viper.Viper, theme modules.Module) er
 
 }
 
+// Helper function for mergeStringMapKeepLeft
+// If both sides of the configuration map are maps, merge rather than replace
+// contents.
+func mergeStringMapRecursively(l, r map[string]interface{}) {
+	var isMap bool
+	var lMap, rMap map[string]interface{}
+	for key, val := range r {
+		if _, found := l[key]; !found {
+			l[key] = val
+		} else {
+			rMap, isMap = val.(map[string]interface{})
+			if !isMap {
+				rMap, isMap = val.(maps.Params)
+			}
+
+			if isMap {
+				lMap, isMap = l[key].(map[string]interface{})
+				if !isMap {
+					lMap, isMap = l[key].(maps.Params)
+				}
+
+				if isMap {
+					mergeStringMapRecursively(lMap, rMap)
+				}
+			}
+		}
+	}
+}
+
 func (configLoader) mergeStringMapKeepLeft(rootKey, key string, v1, v2 config.Provider) {
 	if !v2.IsSet(key) {
 		return
@@ -558,6 +587,10 @@ func (configLoader) mergeStringMapKeepLeft(rootKey, key string, v1, v2 config.Pr
 				continue
 			}
 			m1[k] = v
+		} else if m1v, isMap := m1[k].(map[string]interface{}); isMap {
+			if m2v, isMap := v.(map[string]interface{}); isMap {
+				mergeStringMapRecursively(m1v, m2v)
+			}
 		}
 	}
 }
--- a/hugolib/config_test.go
+++ b/hugolib/config_test.go
@@ -524,3 +524,41 @@ resamplefilter = "CatmullRom"
 	c.Assert(cfg.Get("intSlice"), qt.DeepEquals, []interface{}{5, 8, 9})
 
 }
+
+func TestRecursiveMerging(t *testing.T) {
+
+	c := qt.New(t)
+
+	left := map[string]interface{}{
+		"foo": "test",
+		"bar": 12,
+		"baz": map[string]interface{}{
+			"foo": "nested",
+			"bar": nil,
+		},
+	}
+
+	right := map[string]interface{}{
+		"foo": "not applied",
+		"qux": "applied",
+		"baz": map[string]interface{}{
+			"foo": "ignored",
+			"bar": "untouched",
+			"baz": "is set",
+		},
+	}
+
+	result := map[string]interface{}{
+		"foo": "test",
+		"bar": 12,
+		"qux": "applied",
+		"baz": map[string]interface{}{
+			"foo": "nested",
+			"bar": nil,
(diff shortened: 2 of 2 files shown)


bep:
I understand the problem and we need to do something about it, but I don't think we can go "all in" as in this PR (I don't see any tests, though, so I'm not sure ...)

If the theme defines:

```toml
[outputFormats]
  [outputFormats.MyEnrichedFormat]
    baseName = "myindex"
    isPlainText = true
    mediaType = "text/enriched"
    protocol = "bep://"
```

And I have that overridden in my project, it may look like this:

```toml
[outputFormats]
  [outputFormats.MyEnrichedFormat]
    baseName = "foo"
    mediaType = "text/enriched"
    protocol = "http://"
```

If I merge this PR, the above setup will be broken in a very subtle way that I don't think we can fix by adding a note in the release notes.

/cc @regisphilibert

bep:
.. note that the above may be an acceptable breakage (the net gain is good), but we should really think about it ...

Shadow53:
Something I think I mentioned in a previous discussion is that I would welcome having a flag for Hugo that would change how it parses configuration files. E.g. `--recursive-merge` could enable this recursive behavior, or `--recursive-merge=false` disables it, depending on what the default ends up being.

I think this would be better as the default behavior, but maybe we hide it behind a flag for a few releases and let people opt into testing it?

Shadow53:
Alternatively, the documentation could be edited to include a warning at the top of the relevant pages that modules may set configuration items and they will keep those values if you don't change them in your site's `config.toml`. More experienced Hugo users may not see that warning, but it should cover newcomers, and a note in the release notes would help warn those experienced users too.

Shadow53:
I was sure I had committed the new test, but apparently not. The recent forced push should fix that.

regisphilibert:
I personally think recursive merging is great for I often find myself having to copy/paste complete maps in order to overwrite just one key. I for one expect recursive merging in settings (for some reason) so wouldn't complain too much about this implementation.

But then again, Björn Erik is right. It will probably break some current setups 🤷‍♂ especially among theme users. 

I feel a good way to introduce a potential "breaking" improvement is to only enable it for well informed/prepared users.  

I like the idea of a CLI flag that would defaults to  `false` for a few releases and only defaults to `true` when people got some time to get acquainted with the feature.

This way, when first introduced, the noise made by the breaking changes do not shadow the benefit of the feature.

Shadow53:
My only concern with a CLI flag is that users who don't need it won't use it and, if/when it gets enabled by default, they'll run into those same breakages.

Then again, I haven't interacted with the community much recently, and maybe more people would be willing to test the flag than I'm expecting.

bep:
A flag wouldn't solve the real problem (themes). It also means more work on the dev side, which I don't like.

I have thought about this and my initial conclusion is that we cannot do this as a general thing. And that is not just because of backwards compatibility.  People use `params` for both configuration and data, and while there may be possible to null/set default value for config options, that isn't always easy/possible for data (you can set values to nil/zero, but you cannot remove them).

So, if we want to do something about this, I think it needs to be restricted to the **params maps, and one level only*.

Shadow53:
My first instinct is to disagree about the "one level only". My use case in particular requires multiple levels of recursion to work: multiple kinds of "plugins" with namespaced configuration. So a search bar would use `params.sidebar.search`, a shortcode to render math using KaTeX would use `params.shortcodes.katex`, so on. Modules set information on how to integrate with the theme and users set configuration options, both under that namespace.

I'm not sure what you mean by people using `Params` for data, but isn't that what the [`data` folder](https://gohugo.io/templates/data-templates/) is for? I think I understand what you are talking about in principle, but I can't picture an example of what that looks like.

bep:
>I'm not sure what you mean by people using Params for data, but isn't that what the data folder is for? 

I have no way of knowing what users in the wild are using Params for. Perhaps `params.sidebar.flags`? This sounds reasonable to store in `Params`-- and if that is a  map, I would have no way to remove the Swedish flag ...

With the fear of making this too complicated ... the `Params` map is already a very special citizen in Hugo, so we could consider adding a processing instruction to it:

```toml
[params]
__h_nomerge = true
foo = "bar"
```

A better suggestion for the key above would be needed, but as we would remove it as part of the merge, it should be reserved/unique. It would work downwards, so we could put it on the `params.sidebar.flags` if needed.

This is me thinking ... 

/cc @regisphilibert

Issue #6595, opened 2019-12-10
Live reload doesn't respect --baseURL setting


sth:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
Hugo Static Site Generator v0.54.0/extended linux/amd64 BuildDate: 2019-06-04T19:16:36Z
</pre>

### Does this issue reproduce with the latest release?

Yes, the current git master version has the same behaviour.

### livereload.js

Hugo always tries to load `livereload.js` from the hardcoded path `/livereload.js`, even when there is a `--baseURL` parameter that specifies a different root path. So, for example, with `--baseURL=https://tejp.de/drafts/` I would have expected the livereload script to be loaded from `https://tejp.de/drafts/livereload.js` or simply `/drafts/livereload.js`, but not from `/livereload.js` which lies "outside" of the baseURL.

Other page resources respect the specified baseURL, so I would have expected the support scripts for livereload to do the same.

### What's wrong with that?

So why is that a problem? My use case for `--baseURL` including a path was to access the hugo server through a proxy as a subdirectory on my webserver, accessible as `https://tejp.de/drafts/`.
So `https://tejp.de/drafts/` and everything below gets proxied to `http://localhost:1313/`. I add some authentication and now I can view my draft posts conveniently at that URL. It works nicely, only live reload doesn't work because it uses a different path.

I imagine such a proxy to be the main use case for `--baseURL`.

### livereload websocket

If I reconfigure things so livereload.js gets loaded successfully, it tries to open a websocket connection to `/livereload`, again without using the `--baseURL` setting. This websocket should also be located below baseURL and not "leak" into the toplevel directory.

This can easily be accomplished by passing an additional `path=...` URL parameter to livereload.js.

### Fixing it

It is quite possible that I misunderstand how live reload is supposed to work and am going in the wrong direction, but not I have a proof of concept patch that implements these changes. It needs some polishing but should work out without major problems.

bep:
>This websocket should also be located below baseURL and not "leak" into the toplevel directory.

Can you tell me what **real** problem you're trying to solve? Note that this is a development server.

sth:
I have a web server with a hugo installation. Nginx on that server is configured to serve the generated pages. To change something, I ssh into the server, write some text and run hugo to re-generate the page.

When I want to run `hugo serve` on the server while writing new content I need some way to access that with my browser on my local machine (not on the web server).

[...]

PR #6698, opened 2020-01-02
Use --baseURL path for live-reload URL

sth:
Currently the live-reload script always uses the top-level path '/livereload.js', even if the page lives in a subdirectory. This PR changes that so that a subdirectory in `--baseURL` will be respected. So with `--baseURL=http://example.com/subdir` the live-reload will use `/subdir/livereload.js`. (See #6595)


transform/livereloadinject/livereloadinject_test.go
```diff
@@ -31,10 +32,15 @@ func doTestLiveReloadInject(t *testing.T, bodyEndTag string) {
...
+	u, err := url.Parse("http://localhost:1234/subpath")
+	if err != nil {
+		t.Errorf("Parsing test URL failed")
+		return
+	}
+	tr := transform.New(New(*u))
 	tr.Apply(out, in)
 
-	expected := fmt.Sprintf(`<script data-no-instant>document.write('<script src="/livereload.js?port=1313&mindelay=10&v=2"></' + 'script>')</script>%s`, bodyEndTag)
+	expected := fmt.Sprintf(`<script data-no-instant>document.write('<script src="/subpath/livereload.js?mindelay=10&v=2&port=1234&path=subpath/livereload"></' + 'script>')</script>%s`, bodyEndTag)
```
bep, line 43: I assume you have tested this, so I guess it's correct, but the `path=subpath/livereload` looks ... odd?

sth: It looks odd, but [the default][lr-1] is `path=livereload`, so with added "subpath", which is the path used in the test, it becomes `path=subpath/livereload`. Livereload [builds a full absolute URL][lr-2] from this.

 [lr-1]: https://github.com/livereload/livereload-js#options
 [lr-2]: https://github.com/livereload/livereload-js/blob/5cba9245931bbb9bbdc61703794521ce06e4402b/src/connector.js#L11-L12

sth:
A note: The pull request uses a `url.URL` to pass the Livereload settings around, but only `Port()` and `Path` of that URL are actually used. So instead of the URL we could also use two separate parameters for port and path everywhere. Using an URL just felt convenient.

commands/server.go
```diff
@@ -446,8 +448,15 @@ func (c *commandeer) serve(s *serverCmd) error {
...
-			mu.HandleFunc("/livereload", livereload.Handler)
+			u, err := url.Parse(baseURLs[i])
+			if err != nil {
+				return err
+			}
+			if !strings.HasSuffix(u.Path, "/") {
+				u.Path += "/"
+			}
+			mu.HandleFunc(u.Path+"livereload.js", livereload.ServeJS)
+			mu.HandleFunc(u.Path+"livereload", livereload.Handler)
```
Maldris, line 459: considering the repeated checking for trailing slashes here and below, then concatenating a string.
Would using the utility `helpers.SanitizeURL(string) string` be more appropriate? at that point, as it strips trailing slashes, all the string literals may start with a slash for the same effect.

Though it may instead be better to ensure the baseURL flag gets this treatment at the command entrypoint rather than down here

transform/livereloadinject/livereloadinject.go
```diff
@@ -16,25 +16,41 @@ package livereloadinject
...
+		}
+
+		src := baseURL.Path + "livereload.js?mindelay=10&v=2"
+
+		src += "&port=" + baseURL.Port()
+
+		path := baseURL.Path + "livereload"
+		if strings.HasPrefix(path, "/") {
+			path = path[1:]
+		}
```
Maldris, line 44: given `strings.TrimPrefix(str, prefix)`  does exactly this internally, is there a particular motivation for handling this manually like this?

commands/server.go
sth: `SanitizeURL()` is definitely a good idea

transform/livereloadinject/livereloadinject.go
bep: No.

sth:
Since there would be conflicts when merging I'm going to rebase it to the current master.

yihui:
I have also run into this issue and will truly appreciate it if the livereload URL could take `baseURL` into consideration when it contains a subpath. Thank you!

PR #6727, opened 2020-01-08
tpl: Add internal template enabling web monetization

sabinebertram:
Add internal template that enables web monetization by adding a meta tag to the head of each page. It grabs the payment pointer from the params section in the config file.
Documentation has been added to explain how to include web monetization into the website, just like google analytics.

**Background**
[Web monetization](https://webmonetization.org/) is a proposed W3C standard that allows the creation of a payment stream from the user agent to the website. It uses the [Interledger Protocol](https://interledger.org/), a currency agnostic micro-payment protocol, to stream money while a user is visiting the website. 
That user has to have a means of streaming payments. There is no visible difference for users who are streaming and for those who are not streaming. 
Web monetization offers a new possibility for content creators to monetize their work. It is an alternative to the advertisement model.

### DIFF
--- a/docs/content/en/templates/internal.md
+++ b/docs/content/en/templates/internal.md
@@ -197,6 +197,27 @@ To add Twitter card metadata, include the following line between the `<head>` ta
 {{ template "_internal/twitter_cards.html" . }}
 ```
 
+## Web Monetization
+
+An internal template to enable [web monetization](https://webmonetization.org/). It allows the creator of the website to receive micropayments from readers. 
+
+### Configure Web Monetization
+
+Provide your [payment pointer](https://paymentpointers.org/) under the params section in your configuration file:
+
+{{< code-toggle file="config" >}}
+[params]
+  monetization = "$twitter.xrptipbot.com/sabinebertram_"
+{{</ code-toggle >}}
+
+### Use the Web Monetization Template
+
+You can then include the Web Monetization internal template:
+
+```
+{{ template "_internal/webmonetization.html" . }}
+```
+
 ## The Internal Templates
 
 * `_internal/disqus.html`
@@ -206,7 +227,7 @@ To add Twitter card metadata, include the following line between the `<head>` ta
 * `_internal/opengraph.html`
 * `_internal/pagination.html`
 * `_internal/schema.html`
-* `_internal/twitter_cards.html`
+* `_internal/webmonetization.html`
 
 [disqus]: https://disqus.com
 [disqussignup]: https://disqus.com/profile/signup/
--- a/tpl/tplimpl/embedded/templates.autogen.go
+++ b/tpl/tplimpl/embedded/templates.autogen.go
@@ -550,4 +550,5 @@ if (!doNotTrack) {
 <meta name="twitter:creator" content="@{{ . }}"/>
 {{ end -}}
 {{ end -}}`},
+	{`webmonetization.html`, `<meta name="monetization" content='{{ $.Site.Params.monetization }}'>`},
 }
--- a/tpl/tplimpl/embedded/templates/webmonetization.html
+++ b/tpl/tplimpl/embedded/templates/webmonetization.html
@@ -1,0 +1,1 @@
+<meta name="monetization" content='{{ $.Site.Params.monetization }}'>


bep:
Hey, and thanks for this.

Web monetization was new to me, so the first instinct was to say that this may not sound like a general useful template yet ... but I like the money and getting paid for my work ...

The simplest form of this is easy to add to your site/theme as it is today. For the more complex scenarios, you would probably have multiple authors, each with their own wallet.

We need to think a little before adding this, is what I'm saying.

/cc @regisphilibert @onedrawingperday etc.

onedrawingperday:
>[Web monetization](https://webmonetization.org/) is a proposed W3C standard

We cannot add an internal template for a W3C proposal. Can we?

If it is approved then maybe yes, we could add this.

sabinebertram:
> The simplest form of this is easy to add to your site/theme as it is today. For the more complex scenarios, you would probably have multiple authors, each with their own wallet.

I started with the most simple case to get you guys' reaction. I also have an advanced implementations for exclusive content (i.e. hiding stuff on the website if the user visiting it is not streaming). Having multiple payment pointers for different sites when multiple people write blog posts or something is also on my agenda. 

> We cannot add an internal template for a W3C proposal. Can we?

What are the criteria for an internal template?

onedrawingperday:
> What are the criteria for an internal template?

These are not set somewhere, but I suppose that we ought to support features that are part of the W3C spec.

However that is my opinion, the others might think otherwise.

bep:
>We cannot add an internal template for a W3C proposal. Can we?

Of course we can if it is "useful enough".  The web is full of useful stuff that isn't W3C. But I think we need to think a little about this one.

sabinebertram:
Take some time and think it over. Feel free to ask for any additional explanations or resources regarding web monetization and interledger. 

By the way, there is a demo at https://hugowebmonetization.netlify.com/ that also includes the "exclusive content" part. It was a hackathon project of mine.

regisphilibert:
I feel like this belong more in a module/theme than an internal template. Any reason why the Hugo maintainers should be responsible for this implementation and its potential maintaining should the API evolve?

davidsneighbour:
https://www.w3.org/TR/ - there are lots of proposals and standards. Some proposals stay that way for years and then disappear. 

I feel like adding a meta tag should be the function of a "theme component" and not an internal template. Established internal templates add larger functionality than a single header tag. 

Also: as long as it's a W3C proposal only adding it to Hugo might be interpreted as taking a side.

sabinebertram:
> I feel like adding a meta tag should be the function of a "theme component" and not an internal template. 

I have not come across theme components so far but from what I've read after a quick search sounds like this may be an alternative for now. I need to look more into it and have to understand the merge better. I appreciate any good resource links.

> Also: as long as it's a W3C proposal only adding it to Hugo might be interpreted as taking a side.

There is already a wordpress plugin. And it is taking the right side ;-).

davidsneighbour:
You prove my point :) Make it a theme component. Wordpress plugins are not part of Wordpress, but an extension of it. Comparable to theme components. 

You install it like a theme, then add it to your theme configuration in `config.toml` (as an array). 

Have a look at https://github.com/kaushalmodi/hugo-debugprint to see how it's done.

I wonder though if there is a way to "add to `<head>`" without having to temper with templates of a theme itself in the end.

regisphilibert:
> I wonder though if there is a way to "add to <head>" without having to temper with templates of a theme itself in the end.

There is no template "hooks" à la Wordpress. But the component can document that users should add the "partial" call to their head.

sabinebertram:
> But the component can document that users should add the "partial" call to their head.

Wouldn't that mean the the user is either writing their own `baseof.html` or they need to edit the one their main theme is providing? Ideally, I want a solution where the user only has to add the payment pointer in the `config.toml`.

regisphilibert:
>  Ideally, I want a solution where the user only has to add the payment pointer in the config.toml

Understandably, but then theme editors will have to include your module.

sabinebertram:
I am fairly new to all of this and I already like your community - you reply so fast! Thank you!
Are the `baseof.html` files fairly standard across themes? Because then I could write a _default/baseof.html file that looks like all the others but includes web monetization. Because I am assuming that I cannot somehow merge those files or import one into the other since it is the base template.

regisphilibert:
> Are the baseof.html files fairly standard across themes?

No I would say `baseof.html` are very different among themes.

I would gather that most theme uses a `head` partial. So if you write a parital that display your meta, then all theme user would have to do, is overwrite their theme's `head` partial with a copy/paste of the original including the partial call to your own partial.

I'm not sure there is any way around that. But others might chime in.

sabinebertram:
I created a component: https://github.com/sabinebertram/hugo-webmonetization-component
 
I'll close the PR for now and move further discussions to the forum. Thanks for your help so far!

PR #6753, opened 2020-01-14
Allow raw string literals in shortcode params

### LINKED ISSUE #6726: Allow raw string literals in shortcode params
If you try a shortcode like this:

~~~
{{< sunday `monday tuesday` >}}
~~~

You get this:

~~~
unrecognized character in shortcode action: U+0060 '`'. Note: Parameters with
non-alphanumeric args must be quoted
~~~

It seems if Go allow this, Hugo should too:

https://golang.org/ref/spec#raw_string_lit

hcwong:
Fixes #6726 

This PR allows users to use raw string literals in shortcode params.

Examples for both named and positional params include:
```
{{< param `testparam` >}}
{{< param=`testparam` >}}
```

However, it does not support escaped backticks (eg {{< param \\\`testparam\\\` >}}), because it dosen't work on Go (unrecognized escape character error).  A new error is thrown for this as previously, it used to be handled either by the unrecognized unicode character error for the backtick or the unterminated quote error.

### DIFF
--- a/parser/pageparser/pagelexer_shortcode.go
+++ b/parser/pageparser/pagelexer_shortcode.go
@@ -85,6 +85,8 @@ func lexShortcodeRightDelim(l *pageLexer) stateFunc {
 // 2. "param" or "param\"
 // 3. param="123" or param="123\"
 // 4. param="Some \"escaped\" text"
+// 5. `param`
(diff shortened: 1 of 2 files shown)


bep:
This looks great, thanks!

PR #6774, opened 2020-01-18
Skip over temporary files when building pages

### LINKED ISSUE #6773: Temporary files not ignored when (re)building site

v0.62.2-83E50184/extended linux/amd64

Yes

I wasn't sure why I'd often be seeing the live preview of my site break on saves, when the state on the filesystem was fine. Digging into the `no such file or directory` message, I see this has to do with [Emacs lock files](https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html#File-Locks) being followed where I don't believe they should be. The full message is in the form of:

```
Error: add site dependencies: load resources: loading templates: walk: Readdir: decorate: lstat $SITE_ROOT/layouts/partials/$USER@$MACHINE.$PID:$BOOT: no such file or directory
```

Of course the lock file is a symlink pointing to this path, with the prefix `.#` added to the file name in an unsaved, modified buffer.

This is true for both content and templating files, as long as a buffer is open with a modified version. This happens for both `hugo` and `hugo server` commands. It's easiest to reproduce by triggering a rebuild (running `hugo` or saving a change to a different file with `hugo server` running), and having a file open in emacs with a modified state so the lock symlink exists. It seems to also happen as a race condition when editing only a single file with `hugo server` running, where the rebuild reaches the lock symlink before emacs removes it.

I believe `hugo server` ignores events around temporary files so a rebuild isn't triggered, but does not ignore these files in the building of a site. I further believe that, well… that's the wrong behavior.

I have a fix that solves my issue, see imminent pull request.

obar:
The same logic for ignoring a temp-related filesystem event in
`handleEvents` from `commands/hugo.go` is now used to ignore files
when (re)building the site.

Fixes #6773 and #7345

### DIFF
--- a/hugofs/decorators.go
+++ b/hugofs/decorators.go
@@ -191,6 +191,26 @@ func (l *baseFileDecoratorFile) Readdir(c int) (ofi []os.FileInfo, err error) {
 	for _, dirname := range dirnames {
 		filename := dirname
 
+		ext := filepath.Ext(filename)
+		baseName := filepath.Base(filename)
+		istemp := strings.HasSuffix(ext, "~") ||
+			(ext == ".swp") || // vim
+			(ext == ".swx") || // vim
+			(ext == ".tmp") || // generic temp file
+			(ext == ".DS_Store") || // OSX Thumbnail
+			baseName == "4913" || // vim
+			strings.HasPrefix(ext, ".goutputstream") || // gnome
+			strings.HasSuffix(ext, "jb_old___") || // intelliJ
+			strings.HasSuffix(ext, "jb_tmp___") || // intelliJ
+			strings.HasSuffix(ext, "jb_bak___") || // intelliJ
+			strings.HasPrefix(ext, ".sb-") || // byword
+			strings.HasPrefix(baseName, ".#") || // emacs
+			strings.HasPrefix(baseName, "#") // emacs
+
+		if istemp {
+			continue
+		}
+
 		if l.Name() != "" && l.Name() != filepathSeparator {
 			filename = filepath.Join(l.Name(), dirname)
 		}
--- a/hugofs/walk_test.go
+++ b/hugofs/walk_test.go
@@ -47,6 +47,34 @@ func TestWalk(t *testing.T) {
 	c.Assert(names, qt.DeepEquals, []string{"a.txt", "b.txt", "c.txt"})
 }
 
+func TestWalkIgnoreTemp(t *testing.T) {
+	c := qt.New(t)
+
+	fs := NewBaseFileDecorator(afero.NewMemMapFs())
+
+	afero.WriteFile(fs, "b.txt", []byte("content"), 0777)
+	afero.WriteFile(fs, "c.txt", []byte("content"), 0777)
+	afero.WriteFile(fs, "a.txt", []byte("content"), 0777)
+	afero.WriteFile(fs, "4913.txt", []byte("content"), 0777)
+	afero.WriteFile(fs, "a.txt.swp", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "a.txt.swx", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "a.txt.tmp", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "a.DS_Store", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "4913", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, ".goutputstream", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "a.jb_old___", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "a.jb_tmp___", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "a.jb_bak___", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, "a.txt.sb-", []byte("ignoreme"), 0777)
+	afero.WriteFile(fs, ".#a.txt", []byte("ignoreme"), 0777)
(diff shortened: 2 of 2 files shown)


obar:
Anything else this needs before merging @bep? The new test in this pull request fails before the fix is applied and passes after, which is promising :)

obar:
I have used this small patch in 0.62-0.64 with no issues and it fixes the heisenbug that still exists when using hugo with emacs. @bep is there anything I can do to make this pull request ready for 0.65? No new logic is added, just applying the same check in an additional place.

pjeby:
This looks pretty good, though I'd like to suggest adding `.~` to the prefix list; it's used by Typora, the WYSIWYG markdown editor.  (I assume it would need to be added to `handleEvents` as well, though.)

obar:
That sounds like a good idea to add. This pull request and the associated issue are only about skipping temporary files on build, perhaps you should open a new issue about compatibility with another text editor and reference it here.

hugofs/decorators.go
```diff
@@ -191,6 +191,26 @@ func (l *baseFileDecoratorFile) Readdir(c int) (ofi []os.FileInfo, err error) {
 	for _, dirname := range dirnames {
 		filename := dirname
 
+		ext := filepath.Ext(filename)
+		baseName := filepath.Base(filename)
+		istemp := strings.HasSuffix(ext, "~") ||
```
khayyamsaleem, line 196: would it be possible to make this data driven / config driven? I feel like instituting a pattern like `.hugoignore` supporting regular expressions might make this pattern more extensible, rather than inlining support for a bunch of editors of the implementer's choice.

obar:
It's a good idea @khayyamsaleem. The proposed check doesn't do introduce new code, [this is the same check performed elsewhere](https://github.com/gohugoio/hugo/blob/master/commands/hugo.go#L985-L1002) in hugo. They could both go behind a config parameter, with smart defaults so most users wouldn't need to touch it.

I'd be happy to make that change in both parts of the code base if it would be merged into the project. As of now I haven't heard anything about merging this change, and it's not clear if that's because of a problem with the pull request, a problem with the issue it closes, or just the backlog of other pull requests. I think it's still a bug worth addressing though, as several text editors require changes to their configuration in order to avoid a race condition with hugo's live server. @bep is there something here that should be done differently for a merge to happen? Thanks!

jmooring:
This may also fix https://github.com/gohugoio/hugo/issues/7345

sciurius:
I have applied this small patch to git head with no issues and I can confirm it fixes the hugo-with-emacs annoyance.

bep:
>I have applied this small patch to git head with no issues and I can confirm it fixes the hugo-with-emacs annoyance.

Then I assume you don't have a content file named `4913.md`?

I know we're doing the same to filter file chance events, but some false positives is not critical on that side of the fence. Just dropping random files when running `hugo` isn't something we can do.

jmooring:
@bep https://github.com/gohugoio/hugo/blob/master/commands/hugo.go#L1032

obar:
> Then I assume you don't have a content file named 4913.md?

That's a good point! I just checked and this does not skip 4913.md. `filepath.Base` [includes the file extension](https://golang.org/pkg/path/filepath/#example_Base).

Should I add a test condition to make this explicit in the test suite?

garrison:
> Should I add a test condition to make this explicit in the test suite?

Not a Hugo developer myself, but I would be willing to bet that this has a better chance of getting merged with such a test, yes.

obar:
@garrison fair enough :)

@bep there's now a passing test which includes the 4913 case you pointed out. Thanks for taking a look at this!

ChordPro:
I solved it by no longer using Hugo. 
Thanks.

davidsneighbour:
> I solved it by no longer using Hugo.
> Thanks.

I am not sure if that is a passive agressive act or if you just don't know how these things work, but telling the world, that you don't use Hugo on an issue that is marked as stale keeps it open longer:

> This issue will automatically be closed in the near future if no further activity occurs. Thank you for all your contributions.

ChordPro:
I had no other intention than to indicate that I am no longer interested.

pjeby:
@ChordPro The simplest way to indicate a lack of interest in an issue is to unsubscribe from it, so you don't receive further notifications.  Commenting on it just generates notifications to other people...  most of who have no way to unsubscribe you from issues...  and who are likely to reply and create more notifications to everyone involved, yourself included.

To unsubscribe, visit the issue on the website, and look for an unsubscribe button on the right hand side.  (Making sure you're logged in first.)  If you are subscribed to other Hugo issues as well, you may need to visit your Github subscription settings generally.

garrison:
I solved it by rebasing this commit on top of every since-released version of Hugo and building it on my own.  Luckily, Go makes this very easy to do -- I am actually quite impressed with the convenience, even though one downside is that I often forget to set `CGO_ENABLED=1`, so I usually don't get SASS working until my second try after each release.

Still, I would prefer to see this merged one day so I can use Hugo as-is, without any modifications.  It is a shame that `hugo server` does not work out of the box with venerable text editors (vim and emacs), even though such editors are [recommended on Discourse](https://discourse.gohugo.io/t/recommended-reading-reference/14815).

obar:
I've been doing the same as @garrison (well, not every single release) and can confirm this bug still exists / this pull request still fixes it.

To summarize the current state, this proposed bug fix had a potential issue highlighted: what if it breaks existing sites? I added a test case to explicitly cover the discussed edge case and can definitively say this cannot break any sites, no intentional content is saved as `my_post.md.swp` and filenames can't begin with the URL fragment separator `#` because that wouldn't work with a browser.

@bep can you comment on any other reasons why this fix couldn't be merged? Could it at least be merged as a stop-gap to a more elaborate and elegant solution involving redoing the filesystem components of Hugo (discussed elsewhere)?

moorereason:
We're rejecting this PR.  Hugo ships with everything you need to ignore these files via the hugo config.

PR #6775, opened 2020-01-19
Allow multiple arguments in ne/ge/gt/le/lt functions

### LINKED ISSUE #6619: Allow multiple arguments in ne function
This is currently not possible

```
{{ if ne .RelPermalink "/link1/" "/link2/" }}
```

while this is possible

```
{{ if eq .RelPermalink "/link1/" "/link2/" }}
``` 

This may be a good addition because it makes things a lot easier.

Maybe this construct would make sense for `ge`, `gt` …, too?!

le0tan:
This PR is trying to close #6619:

I adopted `&&` as the connective between logical values because of https://github.com/gohugoio/hugo/issues/6619#issuecomment-565811571:

`!(a==b || a==c)` is equilvalent to `a!=b && a!=c` so I understand `ne a b c` should be translated the same way

### DIFF
--- a/tpl/compare/compare.go
+++ b/tpl/compare/compare.go
@@ -135,33 +135,58 @@ func (n *Namespace) Eq(first interface{}, others ...interface{}) bool {
 	return false
 }
 
-// Ne returns the boolean truth of arg1 != arg2.
-func (n *Namespace) Ne(x, y interface{}) bool {
-	return !n.Eq(x, y)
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this; the implementation looks great, but I think we need some added test cases for this...?

le0tan:
@bep I added the requested tests, are they sufficient?

PR #6804, opened 2020-01-26
A solution to easily enable full-text RSS feeds

vhbelvadi:
A proposed `fullTextRSS` Boolean site config parameter that can optionally be used to easily switch between .Summary and .Content to generate either truncated or full-text RSS feeds as required. In my opinion this has to be core Hugo RSS functionality.

### DIFF
--- a/tpl/tplimpl/embedded/templates/_default/rss.xml
+++ b/tpl/tplimpl/embedded/templates/_default/rss.xml
@@ -27,7 +27,7 @@
       <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
       {{ with .Site.Author.email }}<author>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</author>{{end}}
       <guid>{{ .Permalink }}</guid>
-      <description>{{ .Summary | html }}</description>
+      <description>{{ if .Site.fullTextRSS }}{{ .Content | html }}{{ else }}{{ .Summary | html }}{{ end }}</description>
     </item>
     {{ end }}
   </channel>


obar:
Have you gotten any feedback on this yet?

This functionality makes a lot of sense to me, I think it belongs in the base template. You see the same feature built in to feeds in other website authoring tools, eg [wordpress](https://en.support.wordpress.com/settings/reading-settings/), [jekyll](https://github.com/jekyll/jekyll-feed/blob/master/README.md#excerpt-only-flag), [hexo](https://github.com/hexojs/hexo-generator-feed/blob/master/README.md#options)…

vhbelvadi:
@obar I haven’t heard back yet. I guess the 38 PRs in the pipeline are what’s causing the delay? I was under the assumption this was part of the core until I needed it one day and found it wasn’t, so I coded it in myself. Is it any good? Perhaps @betz can chime in.

nekr0z:
One argument against this would be the fact that `{{ .Content | html }}` can produce a lot of things, including nasty and non-conformant ones. It can be argued that possible implications should at least be documented (and at most mitigated) before including such an option.

For example, if Hugo's standart youtube shortcode is used, the resulting xml would be frowned upon by W3C RSS [validator](https://validator.w3.org/feed/) because it would have `iframe` in `<description>`.

Another example would be Hugo's default way of rendering footnotes. Hugo creates relative references (like `<a href="#fn:1">`) that would be invalid in RSS.

There are other such issues, too. Documenting all of them would be quite a big endeavour in itself. With current state of affairs, a user (or a theme author) advanced enough to be able to make the necessary change to the RSS template is supposedly also advanced enough to at least be aware of these issues.

Kaligule:
@nekr0z  

I would argue that this is already a problem, this PR just makes it more visible. [`Summary`](https://gohugo.io/content-management/summaries/) doesn't solve any of the issues you described, it just makes it less likely that they show up in the feed because they might be cut of.

Would this PR be considered if we documented the option and noted that some hugo-features might show up non-conformant in the feed?

PR #6833, opened 2020-01-31
Documentation and Template Author Updates

kc0bfv:
I was confused about how to get author to show up in RSS feeds, and as I figured that out I realized some of the documentation around specifying authors was wrong.  The method used in the RSS template is partially consistent with methods in a couple other places, and the templates for opengraph and twitter_cards do things slightly differently and sometimes slightly impossibly...

[Initial question on Discourse here.](https://discourse.gohugo.io/t/author-configuration-for-rss/23140/2)

These commits should bring that into standardization, and update documentation to explain.

I'm sorry - I don't have "mage".  I have modified some internal templates here, but not combined them into the go file in the parent directory...

### DIFF
--- a/docs/content/en/getting-started/code-toggle.md
+++ b/docs/content/en/getting-started/code-toggle.md
@@ -30,12 +30,13 @@ permalinks:
   posts: /:year/:month/:title/
 params:
   Subtitle: "Hugo is Absurdly Fast!"
-  AuthorName: "Jon Doe"
   GitHubUser: "spf13"
   ListOfFoo:
     - "foo1"
     - "foo2"
   SidebarRecentLimit: 5
+Author:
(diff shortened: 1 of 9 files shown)


kc0bfv:
Well, I did get mage.  Then, updating the template and building again caused errors which I should've seen before.  Mostly surrounding the fact that in some places these things I counted on as being maps are simply strings...  I'm currently working through fixing that by adding detection where important.

Sorry for the half-finished state of my pull request...

kc0bfv:
Based on @bep's comment elsewhere, much of this needs to get moved over to the website's repo.

I think I'll open a better-organized pull request, then close this one.  But I'm still curious about what y'all think about proposed changes.

Issue #6510, opened 2019-11-17
Documentation inaccurate about .Pages on homepage


look:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
Hugo Static Site Generator v0.60.0-DEV darwin/amd64 BuildDate: unknown
</pre>

### Does this issue reproduce with the latest release?

I'm using a pre-release build, but this is also present with the released version, hugo 0.59.

----

In the documentation, [it says that](https://github.com/gohugoio/hugo/blob/master/docs/content/en/templates/homepage.md#pages-on-the-homepage):

> `.Pages` on the Homepage 
> 
> In addition to the standard page variables, the homepage template has access to all site content via `.Pages`.

However, I'm not seeing this. I have this template:

```
{{ range .Pages }}
  {{ . }} <br>
  Kind: {{ .Kind }} <br>
  Type: {{ .Type }} <br>
{{ end }}
```

And the output shows only my top-level section pages:

```
Page(/posts)
Kind: section
Type: posts
Page(/pages)
Kind: section
Type: pages
```

If I use `.Site.Pages` I get all the pages on the site. I ended up adjusting my template like this to get just the posts:

```
{{ range where .Pages "Section" "posts" }}
  {{ range first 10 .Pages }}
    {{ . }} <br>
    Kind: {{ .Kind }} <br>
    Type: {{ .Type }} <br>
  {{ end }}
{{ end }}
```

I'm not sure if this is a bug in Hugo or the documentation.

PR #6838, opened 2020-01-31
Remove inaccurate comment about .Pages on homepage documentation

nerg4l:
This PR removes a line of the documentation which contains an inaccurate statement.

It is related to #6510


bep:

PR #6841, opened 2020-02-01
The hugo function is not documented in the functions section

kc0bfv:
This basically copies the params/variables .Hugo over, which is fine because the interface remained the same.


bep:

kc0bfv:
No problem @bep!  Thanks for all your work on this.

Should the docs directory still be present in the hugo repo then?  It seems like it'll get way out of date...

Issue #6544, opened 2019-11-27
Consider markdown extensions for math typesetting in Hugo


onedrawingperday:
Since the release of Hugo 0.60.0 the new default markdown library is Goldmark and mmark is deprecated and going to be removed at some point.

I am opening this issue to track the matter of math typesetting in Hugo with the new default markdown library.

The present situation is like so: 

https://github.com/litao91/goldmark-mathjax is an extension that offers support for goldmark.

Currently there is no KaTex extension that offers support for goldmark. (ref: https://github.com/yuin/goldmark/issues/32 )

severen:
The lack of support for maths typesetting with Goldmark (and specifically with using KaTeX to render it) is currently a blocker for me.

mobop:
Some one should write a KaTex extension for goldmark

onedrawingperday:
I have just come across a KaTex extension for Goldmark.

https://github.com/graemephi/goldmark-qjs-katex

As mentioned in the first post there is another for MathJax over here:
https://github.com/litao91/goldmark-mathjax

cc: @bep

oblitum:
Would be nice to have a rationale for this here. I came here from build warning and have no idea why of the deprecation, as Mmark coupled with KaTeX has been working great for me and I use it in all of my posts....

onedrawingperday:
@oblitum The release notes for 0.60.0 provide all the rationale you request.

Also the former default Blackfriday is largely unmaintained. mmark decided to switch to another Markdown engine that unfortunately came with its own performance penalties. So in the end it was decided that Goldmark which is Commonmark compliant should be adopted.

jemus42:
Proper KaTeX and/or mathjax support by hugo would be huge. So far I always had to resort to various hacky solutions that never lasted without breaking in one way or another.

oblitum:
Well this sucks and I used to recommend Mmark for math savvy hugo bloggers (it's a great and simple solution that works), I hope they don't come hunting me now, I made a dozen of people adopt it.

oblitum:
[...]

PR #6842, opened 2020-02-01
Add KaTeX server side math rendering support

awwong1:
Addresses #6544

To enable, set in `config.toml`:
```text
[markup]
  [markup.goldmark]
    [markup.goldmark.extensions]
      katex = true
```

Renders inline math `$1+2=3$` as well as display math `$$ \dfrac{1}{2} = 0.5 $$`.


awwong1:
Closing the PR due to no cgo rule.

entron:
@awwong1 This would be great. What can't someone merge this?

awwong1:
@entron I can empathize with the maintainers for not wanting to support a
complicated cgo dependency for this feature.
I know that this approach does break the build in certain environments.

As an aside, because server side rendered math is important to me, I
migrated over to Hexo.

https://hexo.io/
https://github.com/hexojs/hexo-math

Issue #6880, opened 2020-02-11
Simplifying doWithCommandeer


MarkRosemaker:
- **Problem:** purpose of doWithCommandeer unclear, seemingly superfluous
- **Proposal:** coding its functionality directly, replacing with boolean if necessary

I've been examining the code for a couple of days. Even though I didn't understand the purpose of some structs right away, over time they seemed to make a bit more sense.

However, one object seems to me to be superfluous:
`doWithCommandeer func(c *commandeer) error`

It's a field in commandeer and passed around in `newCommandeer` and `initializeConfig`. Do a search for it and you'll find 12 results.
You'll also find that it never does anything except once in `loadConfig`:

```
doWithCommandeer := func(cfg config.Provider) error {
	c.Cfg = cfg
	if c.doWithCommandeer == nil {
		return nil
	}
	err := c.doWithCommandeer(c)
	return err
}
```

That's it: Besides running a previous `doWithCommandeer` (which would only do the same thing), it just assigns `cfg` to `c.Cfg`. Since it's always called like `c.doWithCommandeer(c)`, that would be `c.Cfg = c`.

The only other functionality not mentioned yet is to be passed on to `hugolib.LoadConfig`, which does:

```
if err := d(v); err != nil {
	return v, configFiles, err
}
```

with d being `doWithCommandeer` (an input among other functions) and v being a pointer to a viper (`*viper.Viper`), so it would do `c.Cfg = v` (and err always nil).

Could we simplify this by directly calling `c.Cfg = c` and `c.Cfg = v`? LoadConfig returns the `*viper.Viper` so it can be done.

If necessary, we could replace doWithCommandeer with a boolean telling us which needs to be done.

bep:
I welcome any simplification in this package! And the ugliness is all my fault, but to defend myself a little, a short history:

* The commands package was all globals and no tests.
* So when I had the "get rid of all globals so we can run tests in parallel" I had some really gray hairs when it the commands package. The Cobra API is a little bit clumsy.

That said, now we have decent test coverage, so I suggest that you just "do it" and open a PR.

MarkRosemaker:
Great, will do!

MarkRosemaker:
I think you did a great job btw. Hugo is very sophisticated so there is bound to be some complexity.

Regarding doWithCommandeer, I found out its purpose and decided that a simple renaming will make it much clearer:
#6881

PR #6881, opened 2020-02-11
Renaming doWithCommandeer to cfgInit to make clearer what it does and for consistency

MarkRosemaker:
Regarding my [proposal to simplify doWithCommandeer](#6880), I found that renaming alone will clear things up.

For one, I realized that there are two different object types:
1. `doWithCommandeer func(c *commandeer) error`
2. `doWithCommandeer func(cfg config.Provider) error`

When `initializeConfig(mustHaveConfigFile, running bool, h *hugoBuilderCommon, f flagsToConfigHandler, doWithCommandeer func(c *commandeer) error)` is called, the last argument is either `nil` or was named `cfgInit`.

Therefore, renaming 1. to `cfgInit` clears things up because it's the only thing it ever does.
I renamed 2. `cfgSetAndInit` because it sets the Cfg field of commandeer (to a *viper.Viper pointer) and runs `cfgInit`.

### DIFF
--- a/commands/commandeer.go
+++ b/commands/commandeer.go
@@ -171,7 +171,7 @@ func newCommandeer(mustHaveConfigFile, running bool, h *hugoBuilderCommon, f fla
 		h:                   h,
 		ftch:                f,
 		commandeerHugoState: newCommandeerHugoState(),
-		cfgInit:    cfgInit,
+		cfgInit:             cfgInit,
 		visitedURLs:         types.NewEvictingStringQueue(10),
 		debounce:            rebuildDebouncer,
 		fullRebuildSem:      semaphore.NewWeighted(1),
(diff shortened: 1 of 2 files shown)


bep:
You need to gofmt your source files. I recommend you use an editor where you can configure this to happen automatically on save. I use LiteIDE where this is default.

MarkRosemaker:
That's strange. I do have gofmt on save and I now manually ran it. There are no changes to commit.

bep:
What Go version are you running? We do the "gofmt check" on Travis with Go 1.13 only as they tend to change their format rules slightly from version to version. So if you gofmt with Go 1.12 it could still be "wrong" for Go 1.13.

MarkRosemaker:
> What Go version are you running? We do the "gofmt check" on Travis with Go 1.13 only as they tend to change their format rules slightly from version to version. So if you gofmt with Go 1.12 it could still be "wrong" for Go 1.13.

Fixed it, thanks!

bep:
OK, the change itself looks good, but can you squash the 3 commits and make a proper commit message  (it does not have to be long, important part is the title).

jtierney4:
Sorry.

PR #6889, opened 2020-02-12
Remove unneeded type attribute from gist embed shortcode

coliff:
Gist itself uses the embed code as:
`<script src="https://gist.github.com/spf13/7896402.js"></script>`
I think it makes sense to use that and keep things shorter.


bep:
I have a nagging feeling that we have an open issue discussing this.

XhmikosR:
#6046

Issue #6877, opened 2020-02-10
HTTP code blocks not fully rendering


rlepinski:
HTTP code blocks not fully rendering using code fences or the `highlight` shortcode.

Input:
```http
GET /foo HTTP/1.1
Content-Type: application/json
User-Agent: foo

{
  "hello": "world"
}
```

Output:
```<div class="highlight"><pre style="color:#f8f8f2;background-color:#272822;-moz-tab-size:4;-o-tab-size:4;tab-size:4"><code class="language-http" data-lang="http"><span style="color:#a6e22e">GET</span> /foo <span style="color:#66d9ef">HTTP</span><span style="color:#f92672">/</span><span style="color:#ae81ff">1.1</span>
Content-Type<span style="color:#f92672">:</span> <span style="color:#ae81ff">application/json</span>
User-Agent<span style="color:#f92672">:</span> <span style="color:#ae81ff">foo</span>

</code></pre></div>
```

I am not sure if this is a Chroma issue or not. Using Chroma (0.71) directly it seems to render fine:

```
$ chroma -s monokailight --html --html-lines --html-lines-table --html-inline-styles example

<html>
<body style="color:#272822;background-color:#fafafa">
<div style="color:#272822;background-color:#fafafa">
<table style="border-spacing:0;padding:0;margin:0;border:0;width:auto;overflow:auto;display:block;"><tr><td style="vertical-align:top;padding:0;margin:0;border:0;">
<pre style="color:#272822;background-color:#fafafa"><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">1
</span><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">2
</span><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">3
</span><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">4
</span><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">5
</span><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">6
</span><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">7
</span></pre></td>
<td style="vertical-align:top;padding:0;margin:0;border:0;;width:100%">
<pre style="color:#272822;background-color:#fafafa">GET /foo HTTP/1.1
[...]

Issue #6856, opened 2020-02-05
Less-than not escaped in code blocks with Blackfriday


jgreely:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
Hugo Static Site Generator v0.60.1-96066756 darwin/amd64 BuildDate: 2019-11-29T14:59:30Z
$ hugo version
Hugo Static Site Generator v0.64.0-241DB8F7 darwin/amd64 BuildDate: 2020-02-04T09:10:23Z
</pre>

The less-than character is not being escaped in code blocks if you render with Blackfriday under 0.60-0.64. Worked in 0.59.1.

Here's a sample one-page site:

https://github.com/jgreely/hugo-blackfriday-code-bug

satotake:
@jgreely 

Hugo >= 0.60 is supposed to have a bug of syntax highlight feature with blackfriday.
So, unless you require highlighting feature, this can be avoided with configuration like 

```toml
baseURL = "http://example.org/"
languageCode = "en-us"
title = "My New Hugo Site"
theme = "ananke"

[markup]
  defaultMarkdownHandler = "blackfriday"
  [markup.highlight]
    codeFences = false
```

`goldmark` has its own highlighting feature.
And Hugo >= v0.6.0 uses it when `goldmark` is set as markdown handler.
Therefore it works fine with `goldmark`

PR #6901, opened 2020-02-16
markup/highlight: Fix chroma highlight

satotake:
* Use chroma.Coalesce
    * for https://github.com/gohugoio/hugo/issues/6877
    * This method was used in [Hugo < v0.60.0](https://github.com/gohugoio/hugo/blob/07a203406a1863ba7c48f9090de7c4587fae913d/helpers/pygments.go#L188) and are missing in >= v0.60.0
* Escape code strings if lexer is nil
    * for https://github.com/gohugoio/hugo/issues/6856

Note: this fixation is only for `choma`'s highlight. 
We need more fixation of highlight function with `goldmark` because its highlight does not  work properly in CodeFences yet.

### DIFF
--- a/markup/highlight/highlight.go
+++ b/markup/highlight/highlight.go
@@ -15,9 +15,9 @@ package highlight
 
 import (
 	"fmt"
+	gohtml "html"
 	"io"
 	"strings"
-	gohtml "html"
 
 	"github.com/alecthomas/chroma"
 	"github.com/alecthomas/chroma/formatters/html"
--- a/markup/highlight/highlight.go
+++ b/markup/highlight/highlight.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	gohtml "html"
 
 	"github.com/alecthomas/chroma"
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this, can you run go fmt on your code to make the build go green.

bep:
Also, can you use "Fixes #issueid"  in your commit message, so they get closed.

frodriguezsmartclip:
@satotake Hi there, in new [Hugo's version >v.60.0 they use Goldmark....](https://gohugo.io/getting-started/configuration-markup#highlight)

I configured PrismJS to highlight the code with Hugo, and perfect. It has a lot of cool features!! 👌 

You can [see my comment here](https://github.com/google/docsy-example/issues/37#issuecomment-586157096). Regards!! 😃

satotake:
@bep  sorry...

- > can you run go fmt on your code to make the build go green.
- > can you use "Fixes #issueid" in your commit message, so they get closed.

Done!

@frodriguezsmartclip 
I have not checked either `goldmark` usage on Hugo or the details of `PrismJS` yet.
But anyway, thanks for your sharing information.

PR #6912, opened 2020-02-18
Fix goMinorVersion on non-final Go releases.

### LINKED ISSUE #6910: TestGoMinorVersion fails on Go RCs
Fedora 32 is going to have Go 1.14, but since neither are released yet, it's on the release candidate. This causes a failure in `TestGoMinorVersion` because the minor is incorrectly parsed:

```
--- FAIL: TestGoMinorVersion (0.00s)
    quicktest.go:306: 
        error:
          values are not equal
        got:
          bool(false)
        want:
          bool(true)
        stack:
          /builddir/build/BUILD/hugo-0.64.1/_build/src/github.com/gohugoio/hugo/common/hugo/version_test.go:87
            c.Assert(GoMinorVersion() >= 11, qt.Equals, true)
```

The Go runtime version is "1.14rc1".

<pre>
$ hugo version
0.64.1
</pre>

Yes

QuLogic:
This should work for alpha/beta/rc releases.

Fixes #6910.

### DIFF
--- a/common/hugo/version.go
+++ b/common/hugo/version.go
@@ -15,7 +15,7 @@ package hugo
 
 import (
 	"fmt"
-	"strconv"
+	"io"
 
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


QuLogic:
Oh, I meant to add a small test too.

PR #6948, opened 2020-02-23
Added math.Sqrt (and docs for math.Log)

StarsoftAnalysis:
I wanted to use math.Sqrt for working with areas and aspect ratios of images, so here it is.

I noticed that math.Log was already there, so I added it to the documentation.

### DIFF
--- a/tpl/math/math.go
+++ b/tpl/math/math.go
@@ -72,6 +72,18 @@ func (ns *Namespace) Log(a interface{}) (float64, error) {
 	return math.Log(af), nil
 }
 
+// Sqrt returns the square root of a number.
+// NOTE: will return for NaN for negative values of a
+func (ns *Namespace) Sqrt(a interface{}) (float64, error) {
+	af, err := cast.ToFloat64E(a)
+
+	if err != nil {
+		return 0, errors.New("Sqrt operator can't be used with non integer or float value")
+	}
+
+	return math.Sqrt(af), nil
+}
+
 // Mod returns a % b.
 func (ns *Namespace) Mod(a, b interface{}) (int64, error) {
 	ai, erra := cast.ToInt64E(a)
--- a/tpl/math/math_test.go
+++ b/tpl/math/math_test.go
@@ -153,6 +153,51 @@ func TestLog(t *testing.T) {
 		c.Assert(err, qt.IsNil)
 		c.Assert(result, qt.Equals, test.expect)
 	}
+
+	// Separate test for Log(-1) -- returns NaN
+	result, err := ns.Log(-1)
+	c.Assert(err, qt.IsNil)
+	c.Assert(result, qt.Satisfies, math.IsNaN)
+}
+
+func TestSqrt(t *testing.T) {
+	t.Parallel()
+	c := qt.New(t)
+
+	ns := New()
+
+	for _, test := range []struct {
+		a      interface{}
+		expect interface{}
+	}{
+		{81, float64(9)},
+		{0.25, float64(0.5)},
+		{0, float64(0)},
+		{"abc", false},
+	} {
+
+		result, err := ns.Sqrt(test.a)
+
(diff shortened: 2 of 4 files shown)


bep:
This looks good. Can you press the CLA button?

tpl/math/math.go
```diff
@@ -72,6 +72,18 @@ func (ns *Namespace) Log(a interface{}) (float64, error) {
 	return math.Log(af), nil
 }
 
+// Sqrt returns the square root of a number.
+// NOTE: will return for NaN for negative values of a
+func (ns *Namespace) Sqrt(a interface{}) (float64, error) {
+	af, err := cast.ToFloat64E(a)
+
+	if err != nil {
+		return 0, errors.New("Sqrt operator can't be used with non integer or float value")
```
moorereason: This error scenario needs some work.  The only two reasons `cast.ToFloat64E` will fail are for a malformed string (using `strconv.ParseFloat`) or for unknown/unsupported data types.  I would suggest something like:

```go
af, err := cast.ToFloat64E(a)
if err != nil {
    return 0, fmt.Errorf("Sqrt operator error: %s", err)
}
```

Also, `cast.ToFloat64E` will allow negative values.  It would probably help our users if we caught negative values before calling `math.Sqrt`.  So, I'd add a separate if-statement before the final return.

```go
if af < 0 {
    return 0, errors.New("Sqrt operator must be greater than or equal to zero")
}
```

I just noticed that `Log` has similar issues.  Let's fix `Log` in a separate PR.

tpl/math/math_test.go
```diff
@@ -153,6 +153,51 @@ func TestLog(t *testing.T) {
...
+
+	for _, test := range []struct {
+		a      interface{}
+		expect interface{}
+	}{
+		{81, float64(9)},
+		{0.25, float64(0.5)},
+		{0, float64(0)},
+		{"abc", false},
+	} {
```
moorereason: Add a test for negative values.

moorereason: I meant to edit this comment before submitting it after seeing your test for negative at the end of the func.  Can we incorporate that test into this table-based test loop?

StarsoftAnalysis: I didn't include the negativity test in the existing loop because it required using `qt.Satisfies` instead of `qt.Equals`, and I couldn't see a neater way (not being familiar with QuickTest).  Any suggestions for a better way?

moorereason: If you take my advice on changing the Sqrt error handling, will you get a NaN result?  I suggested returning zero, but I suppose you could return NaN if `af < 0`.  If you do return NaN, you can make a special case in the loop similar to the `test.expect.(bool)` case on line 181 below.

I'm probably being a little pedantic here.  I just prefer having all the tests in the table if we're going to use table-based tests.

StarsoftAnalysis: I don't think `math.Sqrt(-1)` should return zero because that's wrong maths, and wouldn't alert the user to the problem.  

An alternative would be to raise an error, as it currently does if the argument is not numeric.  But Hugo is quite capable of handling NaN values -- for example `printf "%f" (math.Sqrt -1)` displays the string "NaN" which would make the problem visible without crashing anything.

So I still think that returning NaN is the best approach.  

And while I agree that keeping the test suite table-based would be ideal, it's not possible to use `qt.Equals` with NaNs, so a separate test is needed.

bep: I'm going to decide here: We don't need to do anthing special here: Just return the value returned from `math.Sqrt`. Also, I'm not too interested in tests that tests math.Sqrt -- just smoke test it (verify that we call the right function) and then test some of the error cases (on the input side). Which makes this PR look perfectly fine to me as it is now.

StarsoftAnalysis: OK, thanks.

PR #6962, opened 2020-02-26
deploy: Implement include/exclude filters for deploy

### LINKED ISSUE #6922: Hugo Deploy Include/Exclude Filters
We are currently trying to adopt `hugo deploy` with our large site, but need more fine tuned control over which assets we deploy to s3 and where. We currently deploy our html files to one bucket, and all other static assets (images, videos) to another s3 bucket. We would like to continue serving these static assets from their own bucket (and domain), but that is currently not possible with `hugo deploy`. 

AWS CLI currently provides this functionality: 
https://docs.aws.amazon.com/cli/latest/reference/s3/#use-of-exclude-and-include-filters

Would it be possible to include this functionality with `hugo deploy`?

For example, the config might look like: 
```
[[deployment.targets]]
# An arbitrary name for this target.
name = "live-html"
URL = "s3://<Bucket Name>?region=<AWS region>"
exclude = "*"
include = "*.html, *.xml, *.txt"

[[deployment.targets]]
name = "live-statics"
URL = "s3://<Bucket Name>?region=<AWS region>"
include = "*"
exclude = "*.html, *.xml, *.txt"
```

Here is a similar post in hugo discourse: 
https://discourse.gohugo.io/t/hugo-deploy-include-exclude-filters/23279

Thanks!

vangent:
Fixes #6922.

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -467,17 +467,17 @@ func walkLocal(fs afero.Fs, matchers []*matcher, include, exclude glob.Glob) (ma
 		}
 
 		// Check include/exclude matchers.
-		if include != nil && !include.Match(path) {
-			jww.INFO.Printf("  dropping %q due to include\n", path)
+		slashpath := filepath.ToSlash(path)
+		if include != nil && !include.Match(slashpath) {
+			jww.INFO.Printf("  dropping %q due to include\n", slashpath)
 			return nil
 		}
-		if exclude != nil && exclude.Match(path) {
-			jww.INFO.Printf("  dropping %q due to exclude\n", path)
+		if exclude != nil && exclude.Match(slashpath) {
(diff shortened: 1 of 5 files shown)


deploy/deploy.go
```diff
@@ -461,6 +466,16 @@ func walkLocal(fs afero.Fs, matchers []*matcher) (map[string]*localFile, error)
 			path = norm.NFC.String(path)
 		}
 
+		// Check include/exclude matchers.
+		if include != nil && !include.Match(path) {
```
bep: I notice you're doing the ToSlash lower down. The Glob matches on Unix style paths, so I don't think this would work in the non-simple cases on Windows.

vangent: Good catch! Fixed to match against `slashpath`.

Issue #6892, opened 2020-02-13
“hugo --minify” strips spaces and preserve some new lines


inwardmovement:
<pre>
v0.64.1/extended windows/amd64
</pre>
---

When I use hugo --minify in my [branch/repo](https://github.com/inwardmovement/inwardmovement.github.io/tree/test-hugo-minify):
- some new lines are not minified,
- some spaces inside tags (that need to be preserved) are striped.

New lines not minified:
![image](https://user-images.githubusercontent.com/9438102/74409961-40c4dd00-4e38-11ea-818f-3a4cf4ad7c35.png)
Space inside tag striped:
![image](https://user-images.githubusercontent.com/9438102/74410057-779af300-4e38-11ea-86fb-ae620560d9d1.png)
![image](https://user-images.githubusercontent.com/9438102/74409968-43bfcd80-4e38-11ea-93f4-8f3c0bed6f5e.png)

aalfiann:
yes me too, but this just happen on <a> tag only.

tdewolff:
I want to clarify that "newlines are not minified" is incorrect. The minifier removes superfluous whitespace, but sometimes whitespace in HTML has meaning and cannot be removed. In that case either a space or a newline are required, and both have the same effect. The minifier will let the remaining _significant_ whitespace be a newline when any of the original whitespace characters contained a newline. This newline _cannot_ be removed without changing the display of HTML.

You second issue that whitespaces get remove in tags, this is because you use the CSS `white-space` property or you change `display` to `inline`. This changes the semantical meaning of HTML and may influence layouts that depend on whitespace for formatting. It is *not recommended* to let your HTML layout depend on whitespace!

bep:
@tdewolff thanks for the input and your work on the magnifiers, much appreciated. I have not looked into this particular issue, but we just merged way to configure and turn off a given minified, so that should solve most of these issues.

virgofx:
@tdewolff  Just a minify note. It appears that the first nested `<source>` block inside a `<picture>` preserve newlines between the parent picture and first source element. I don't see why they should be preserved.

![image](https://user-images.githubusercontent.com/739719/79080974-11aed880-7cce-11ea-9444-db8eb0293694.png)


![Untitled-2](https://user-images.githubusercontent.com/739719/79081024-7a965080-7cce-11ea-88ef-dac77295eabc.png)

tdewolff:
[...]

Issue #6724, opened 2020-01-07
Change minification flag to be global 


arturdryomov:
Running Hugo with the `--minify` flag does not affect assets. I. e. having a Markdown file which is converted to HTML with referenced CSS asset files will minify the HTML but not CSS. Seems like this can be solved by this:

```html
{{ $style := resources.Match "css/*.css" | resources.Concat "bundle.css" | minify }}
<link rel="stylesheet" href="{{ $style.RelPermalink }}" />
```

This way both HTML and CSS assets are minified. Great! But we have a mirror issue then. Running Hugo without the `--minify` flag will minify CSS and will not minify HTML.

From my point of view it would be better to follow the `--minify` flag globally — both for the target content (HTML in our case) and assets (CSS). This way it would be possible to have a staging environment without the minification and a production environment with it.

bep:
I assume that you mean that this:

```
{{ $style := resources.Match "css/*.css" | resources.Concat "bundle.css" }}
```

Would, when run with `--minify` should behave like this?:

```
{{ $style := resources.Match "css/*.css" | resources.Concat "bundle.css" | minify }}
```

arturdryomov:
Yep!

arturdryomov:
@bep, it is possible to achieve the desired effect with latest configuration changes? Cannot connect dots together 🤔

PR #6990, opened 2020-02-29
Add minify config

### LINKED ISSUE #6621: Keep single/double quotes in `--minify`
`--minify` strips out all superfluous quotes. This breaks Netlify Forms handling, which is checking for `name="field-name"`, and fails on `name=field-name`.

Yes, Netlify should be able to process valid HTML. But, it would be nice to have a setting in Hugo to keep single/double quotes, OR change the defaults to have 'em.

In my opion, minifying HTML should be less aggressive, and be just about removing unnecessary white spaces  and comments.

See also: https://github.com/tdewolff/minify#html

### LINKED ISSUE #6750: Add option to disable inline SVG minification
The `minify` tool used by hugo to minify the output works great on HTML, but there have been several issues with it destroying inlined SVGs:

- https://github.com/tdewolff/minify/issues/275
- https://github.com/tdewolff/minify/issues/284
- https://github.com/tdewolff/minify/issues/269
- https://github.com/gohugoio/hugo/issues/6475

These issues stem from the fact that `minify` changes the svg `<path>` attributes and attempts to optimize the drawing of lines on the svg canvas. This is a very hard problem, and as a result hugo users are affected by SVG optimization bugs from time to time. Furthermore, the testing for SVG minification in the `minify` repo seems to still be rudimentary: https://github.com/tdewolff/minify/blob/master/svg/pathdata_test.go 

In `minify` project readme https://github.com/tdewolff/minify it says:
**"Status: SVG: partially implemented; in development"**

I think in light of this, we should add an option to disable inline SVG minification, or - even better - be able to pass options through to minify if another class of problems emerges.

satotake:
related https://github.com/gohugoio/hugo/issues/6750

This PR enables users to turn on/off by minifier types and to configure `minify` with its options 
This is proposal and any comments and suggestions are welcome

## content
- [mod] ~Update github.com/tdewolff/minify~
- [minifiers] New config.go
- [docshelper] Add func merge DocProviders
- [docs] update docs.json

## related issues

- Fixes #6750 
- partially #6892 (preserve space)
- ~Fixes~ Upstream #6621 

edit:
By downgrading `minify` version, #6621 became an upstream issue

### DIFF
--- a/minifiers/config.go
+++ b/minifiers/config.go
@@ -41,16 +41,18 @@ var DefaultConfig = minifiersConfig{
 		KeepEndTags:             true,
 		KeepDefaultAttrVals:     true,
 		KeepWhitespace:          false,
-		KeepQuotes:              false,
+		// KeepQuotes:              false, >= v2.6.2
 	},
 	Css: css.Minifier{
-		Precision: 0,
-		KeepCSS2:  true,
+		Decimals: -1, // will be deprecated
+		// Precision: 0,  // use Precision with >= v2.7.0
+		KeepCSS2: true,
 	},
 	Js:   js.Minifier{},
 	Json: json.Minifier{},
 	Svg: svg.Minifier{
-		Precision: 0,
+		Decimals: -1, // will be deprecated
+		// Precision: 0,  // use Precision with >= v2.7.0
 	},
 	Xml: xml.Minifier{
 		KeepWhitespace: false,
--- a/minifiers/config.go
+++ b/minifiers/config.go
@@ -27,7 +27,7 @@ import (
 	"github.com/tdewolff/minify/v2/xml"
 )
 
-var DefaultConfig = minifiersConfig{
+var defaultTdewolffConfig = tdewolffConfig{
 	EnableHtml: true,
 	EnableCss:  true,
 	EnableJs:   true,
@@ -59,7 +59,7 @@ var DefaultConfig = minifiersConfig{
 	},
 }
 
-type minifiersConfig struct {
+type tdewolffConfig struct {
 	EnableHtml bool
 	EnableCss  bool
 	EnableJs   bool
@@ -75,8 +75,16 @@ type minifiersConfig struct {
 	Xml  xml.Minifier
 }
 
+type minifiersConfig struct {
+	Tdewolff tdewolffConfig
+}
+
+var defaultConfig = minifiersConfig{
+	Tdewolff: defaultTdewolffConfig,
+}
+
 func decodeConfig(cfg config.Provider) (conf minifiersConfig, err error) {
-	conf = DefaultConfig
+	conf = defaultConfig
 
 	m := cfg.GetStringMap("minifiers")
 	if m == nil {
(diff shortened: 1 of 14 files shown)


bep:
This looks really good. I have 3 comments:

1. We cannot update tdewolff deps before we have verified the reason I had to roll back recently (it broke sites badly in the wild, including the Hugo site)

2. I have learned some lessons from Chroma/Blackfriday/Pygments ... there may be a new super minifier on the scene in a year ... with totally different options. So I suggest we split out the general and put the specifics below `tdewolff` (same as in the markup config)

3. I'm not sure about this one, but if I disable XML and do `{ { $xml : $xml | minify }}` what happens?  I assume I get `myxml.min.xml` but with unchanged content. I think we need to split the minifiers in two (one for the final content and the other for the `minify` func) ...?

satotake:
Thanks for comment, @bep 

1 & 2 are done

Please note that #6621 is not solved even by merging this because `minify` is not updated and does not have `KeepQuotes` option.

I admit that I am also not so confident about 3 but let me explain.

>  I disable XML and do { { $xml : $xml | minify }} what happens?

Minification of `resource` raises error and cause build fail when you unset the minfier.
On the other hand, minification of inline contents in `.html` file never raises error.

That is my intention but it is so vague as specification that I have just added a test case and improved the error message.

Second,
> I think we need to split the minifiers in two (one for the final content and the other for the minify func) ...?

It is possible to separate `content` minifier from `resource` minifier in terms of both configurations and processing behaviour (I mean #6724 ).
However I believe that it will be better to use at least single configuration because

1. It reduces configuration simplicity (Hugo's advantage)
2. Most of users who configure minification demand lower level of minification and higher validity of contents. It will be better to apply more secure settings for them in sites' scope.
3. Rquired settings of minification vary with users' situation but they are independent of whether minification targets are derived from `content (or layout)` or `resource`

minifiers/config.go
```diff
@@ -27,54 +27,64 @@ import (
...
-	EnableHtml: true,
-	EnableCss:  true,
-	EnableJs:   true,
-	EnableJson: true,
-	EnableSvg:  true,
-	EnableXml:  true,
-
-	Html: html.Minifier{
+var defaultTdewolffConfig = tdewolffConfig{
+	EnableHTML: true,
```
bep: I think we can assume that any future minifier will support all of these fomats, so I suggest we keep all the Enable* on top level.

satotake: done!
thanks you for reviewing

bep:
Thanks for this. I think we need to settle 3 before we can wrap this.

```bash
{ { $xml := $xml | minify }} 
```

I _think_ I would expect the above to just silently drop the _minify_ statement if I disable minify, that statement could well live in a theme I don't control.

But you are probably right (?) about having 1 disable for both of these. We can consider adding some flag later if needed.

/cc @kaushalmodi @regisphilibert

regisphilibert:
I can see enabling/disabling is made by Type so it seems very controllable. I'm not familiar with mifnication technology.

I'm simply curious of what happens if a CSS file's minified `.Content` is printeed in an HTML file. I suppose if CSS minification is disabled, but HTML is, then the HTML minification will take care of minifying the CSS?

bep:
> then the HTML minification will take care of minifying the CSS?

Not if you include it in style tags (which I assume you should/would).

satotake:
> I think I would expect the above to just silently drop the minify statement if I disable minify, that statement could well live in a theme I don't control.

You are right @bep . I failed to be careful for `theme`. I will fix

bep:
This looks really good now, thanks a lot. I will pull this and apply some adjustments.

bep:
I will complete this in #7077

PR #6998, opened 2020-03-02
Updated README.md: added missing build command

bilke:
When following the build instructions in the README there is no `hugo` binary generated. The missing step is:

```
go build
```


bep:
Thanks. But bulding Go programs you should use _either_ `go build` or `go install`. The main  difference is that install puts them in a common place. If you want to control _where_ to install, you can set the `GOBIN` environment variable.

That said, I think I prefer using `go install` in the README.

PR #7004, opened 2020-03-02
Support unComparable args of uniq/complement/in

### LINKED ISSUE #6105: Uniq not working on a slice of maps
Apology the vocabulary used here for Golang's types is inaccurate. I will use `slice` to define what is commonly known as non-associative arrays and `map` for associative arrays.

Consider the following Front Matter field:

```
slice_of_maps:
  - firstname: Victor
    lastname: Hugo
  - firstname: Émile
    lastname: Zola
  - firstname: Victor
    lastname: Hugo
```

Using uniq on it:
```
{{ .Params.slice_of_maps | uniq }}
```
Will return the following error: `error calling uniq: elements must be comparable`.

Following the motivation behind #5852 I suspect we would want any kind of "collection" to be processed by `uniq`.

[Repo here](https://github.com/regisphilibert/hugo-slice-methods-maps)

Note that `intersect`, though not producing an error would simply return an empty strings when asked to process two "slices of maps" sharing common entries. If `intersect` is not impacted by the resolution of this issue, I'll create a seperate one.

satotake:
Fixes #6105

### DIFF
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -271,18 +271,13 @@ func (ns *Namespace) In(l interface{}, v interface{}) (bool, error) {
 	lv := reflect.ValueOf(l)
 	vv := reflect.ValueOf(v)
 
-	if !vv.Type().Comparable() {
-		return false, errors.Errorf("value to check must be comparable: %T", v)
-	}
-
-	// Normalize numeric types to float64 etc.
 	vvk := normalize(vv)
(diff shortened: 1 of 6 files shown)


bep:
Can you resolve the conflict and squash your commits into 1.

satotake:
I am not familiar with unsafe operation of git ... is this ok?

Issue #7020, opened 2020-03-04
deploy: Make delete respect Include/Exclude Filters


bep:

bep:
I have thought a little about this, and if I were to split my site into domains, I would split it into different buckets or into different subpaths in the same bucket. Either way, there would only be files matching the filters in that bucket/path. So, when I delete I probably always want to clean it up completely, and in case I change my filters, that means "all files".

So, I'm not sure this issue is needed/wanted.

/cc @zbayoff @vangent

zbayoff:
@bep @vangent I understand your viewpoint. For our use case, we have processes that create files outside of Hugo in our pipeline and upload them to the remote bucket. We need to be able to exclude those files from being deleted. This process works for us currently with AWS CLI, so we were hoping to replicate this behavior with `hugo deploy`.

vangent:
Was this fixed in #7024 ?

bep:
@vangent I think so, I'm closing it -- if it's still is an issue, we'll need to create another one.

PR #7024, opened 2020-03-06
deploy include exclude support for remote

davidejones:
Fix for #7020

First contribution to gohugo and still fairly new to golang so go easy on me 🙇

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -138,7 +138,7 @@ func (d *Deployer) Deploy(ctx context.Context) error {
 	d.summary.NumLocal = len(local)
 
 	// Load remote files from the target.
-	remote, err := walkRemote(ctx, bucket)
+	remote, err := walkRemote(ctx, bucket, include, exclude)
 	if err != nil {
 		return err
 	}
@@ -499,7 +499,7 @@ func walkLocal(fs afero.Fs, matchers []*matcher, include, exclude glob.Glob) (ma
 }
 
 // walkRemote walks the target bucket and returns a flat list.
-func walkRemote(ctx context.Context, bucket *blob.Bucket) (map[string]*blob.ListObject, error) {
+func walkRemote(ctx context.Context, bucket *blob.Bucket, include, exclude glob.Glob) (map[string]*blob.ListObject, error) {
 	retval := map[string]*blob.ListObject{}
 	iter := bucket.List(nil)
 	for {
@@ -509,6 +509,15 @@ func walkRemote(ctx context.Context, bucket *blob.Bucket) (map[string]*blob.List
 		}
 		if err != nil {
 			return nil, err
+		}
+		// Check include/exclude matchers.
+		if include != nil && !include.Match(obj.Key) {
+			jww.INFO.Printf("  remote dropping %q due to include\n", obj.Key)
+			continue
+		}
+		if exclude != nil && exclude.Match(obj.Key) {
+			jww.INFO.Printf("  remote dropping %q due to exclude\n", obj.Key)
+			continue
 		}
 		// If the remote didn't give us an MD5, compute one.
 		// This can happen for some providers (e.g., fileblob, which uses the
--- a/deploy/deploy_test.go
+++ b/deploy/deploy_test.go
@@ -720,6 +720,88 @@ func TestIncludeExclude(t *testing.T) {
 	}
 }
 
+// TestIncludeExcludeRemoteDelete verifies deleted local files that don't match include/exclude patterns
+// are not deleted on the remote.
+func TestIncludeExcludeRemoteDelete(t *testing.T) {
+	ctx := context.Background()
+
+	tests := []struct {
+		Include string
+		Exclude string
+		Want    deploySummary
+	}{
+		{
+			Want: deploySummary{NumLocal: 3, NumRemote: 5, NumUploads: 0, NumDeletes: 2},
+		},
+		{
+			Include: "**aaa",
+			Want:    deploySummary{NumLocal: 2, NumRemote: 3, NumUploads: 0, NumDeletes: 1},
+		},
+		{
+			Include: "subdir/**",
+			Want:    deploySummary{NumLocal: 1, NumRemote: 2, NumUploads: 0, NumDeletes: 1},
+		},
+		{
+			Exclude: "**bbb",
+			Want:    deploySummary{NumLocal: 2, NumRemote: 3, NumUploads: 0, NumDeletes: 1},
+		},
+		{
+			Exclude: "bbb",
+			Want:    deploySummary{NumLocal: 3, NumRemote: 4, NumUploads: 0, NumDeletes: 1},
+		},
+	}
+	for _, test := range tests {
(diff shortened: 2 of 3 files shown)


bep:
This looks good and correct to me. @vangent could you have a quick look? Note that we (or I) may want some additional flag to control/override this behaviour, but let's wait with that.

deploy/deploy.go
```diff
@@ -510,6 +510,14 @@ func walkRemote(ctx context.Context, bucket *blob.Bucket) (map[string]*blob.List
 		if err != nil {
 			return nil, err
 		}
+		// Check include/exclude matchers.
+		slashpath := filepath.ToSlash(obj.Key)
+		if include != nil && !include.Match(slashpath) {
+			jww.INFO.Printf("  remote dropping %q due to include\n", slashpath)
```
vangent: Neither of these appear to do anything other than log that it's doing something...?

deploy/deploy.go
```diff
@@ -510,6 +510,14 @@ func walkRemote(ctx context.Context, bucket *blob.Bucket) (map[string]*blob.List
 		if err != nil {
 			return nil, err
 		}
+		// Check include/exclude matchers.
+		slashpath := filepath.ToSlash(obj.Key)
```
vangent: I don't think that this `ToSlash` is necessary; gocloud blob keys should already be slashified. Note that below, the object is added to `retval` under the key `obj.Key`, and that's compared directly to the local file path; that wouldn't work if my assertion was false.

deploy/deploy.go
```diff
@@ -527,7 +535,9 @@ func walkRemote(ctx context.Context, bucket *blob.Bucket) (map[string]*blob.List
 				r.Close()
 			}
 		}
-		retval[obj.Key] = obj
+		if include == nil || include.Match(slashpath) {
```
vangent: IIUC, you want to add `continue` to the two sections above instead of this.

deploy/deploy_test.go
```diff
@@ -720,6 +720,88 @@ func TestIncludeExclude(t *testing.T) {
...
+		{
+			Include: "**aaa",
+			Want:    deploySummary{NumLocal: 2, NumRemote: 3, NumUploads: 0, NumDeletes: 1},
+		},
+		{
+			Include: "aaa",
+			Want:    deploySummary{NumLocal: 1, NumRemote: 1, NumUploads: 0, NumDeletes: 0},
+		},
+		{
+			Exclude: "**bbb",
```
vangent: I think this testcase only passes because the initial deploy is applying the filters; it's not really testing the filtering being applied to the remote.

You might need to do the first deploy with a target with no include/exclude filters, to make sure remote is set up with *all* the files, do the deletes, and then verify the appropriate deletes are skipped.

davidejones:
Thank you both for taking the time to review its much appreciated. I'll make some updates and get back to you.

deploy/deploy.go
davidejones: I've added continue statements here as well, do you think i should remove the logs?

deploy/deploy.go
davidejones: Thanks yes you are correct without ToSlash it works too

deploy/deploy_test.go
davidejones: You are right, one test wasn't actually uploading anything on the second deploy because nothing changed so it was testing the outcome of the first upload which wasn't useful.

I've changed the tests and made sure the initial deploy is just uploading with no include/exclude.

davidejones:
Hey @vangent i've made [some changes](https://github.com/gohugoio/hugo/pull/7024/commits/ce3397d9b994f821683d7fbeabde125cce7f4aa2). Let me know what you think

deploy/deploy_test.go
```diff
@@ -720,6 +720,93 @@ func TestIncludeExclude(t *testing.T) {
 	}
 }
 
+// if we delete files locally that don't match the include/exclude pattern
```
vangent: Nit: comments are sentence, start with "If".

deploy/deploy_test.go
```diff
@@ -720,6 +720,93 @@ func TestIncludeExclude(t *testing.T) {
...
+
+			// Second sync
+			tgt := &target{
+				Include: test.Include,
+				Exclude: test.Exclude,
+			}
+			if err := tgt.parseIncludeExclude(); err != nil {
+				t.Error(err)
+			}
+			deployer = &Deployer{
```
vangent: Nit: I think it should also work to just update `deployer.target = tgt` here and re-use the same one?

docs/content/en/hosting-and-deployment/hugo-deploy.md
```diff
@@ -85,8 +85,9 @@ cloudFrontDistributionID = <ID>
 # Optionally, you can include or exclude specific files.
 # See https://godoc.org/github.com/gobwas/glob#Glob for the glob pattern syntax.
 # If non-empty, the pattern is matched against the local path.
-# If exclude is non-empty, and a file's path matches it, that file is dropped.
-# If include is non-empty, and a file's path does not match it, that file is dropped.
+# If exclude is non-empty, and a file's path matches it, that file is not synced with the remote
```
vangent: Nit: end with period.

docs/content/en/hosting-and-deployment/hugo-deploy.md
```diff
@@ -85,8 +85,9 @@ cloudFrontDistributionID = <ID>
...
+# If include is non-empty, and a file's path does not match it, that file is not synced with the remote
+# Files not synced with the remote will cause both local and remote copies of the file to remain intact should they exist.
```
vangent: Hmm, I don't think these changes quite capture the semantics; "both local and remote" here implies that sometimes local files are deleted, which is never the case.

How about:

```
# All paths are matched against in their filepath.ToSlash form.
# If exclude is non-empty, and a local or remote file's path matches it, that file is not synced.
# If include is non-empty, and a local or remote file's path does not match it, that file is not synced.
# As a result, local files that don't pass the include/exclude filters are not uploaded to remote,
# and remote files that don't pass the include/exclude filters are not deleted.

**Comment by davidejones** [comment]
Nice feedback @vangent, I think your wording on the docs is much clearer so I've adopted that and cleared up the other points you mentioned in my [last commit](https://github.com/gohugoio/hugo/pull/7024/commits/fcbe6b90a71d7624fb8fb7b1f39b30a65653185c)

File: deploy/deploy_test.go [hunk 9]
```diff
@@ -720,8 +720,8 @@ func TestIncludeExclude(t *testing.T) {
 	}
 }
 
-// if we delete files locally that don't match the include/exclude pattern
-// the sync should not delete them on the remote.
+// TestIncludeExcludeRemoteDelete verifies deleted local files that don't match include/exclude patterns
+// are not deleted on the remote
```
vangent: Tiny nit: end with period.

davidejones:
Ok I've done that last change and I've rebased/squashed into one commit 👍

PR #7038, opened 2020-03-10
Add languageDirection to language configuration

### LINKED ISSUE #6550: Add languageDirection to language configuration
A number of languages are written from [right to left](https://en.wikipedia.org/wiki/Right-to-left) (rtl). It would be good to have a setting for that in Hugo languages.

You can already add any settings you like as a param and that is how I do it now for the sites that use a "rtl" language. But I think this deserves an official setting with a fixed name to make it easier for themes to implement.

The attribute that is set on the html tag is named "[dir](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir)" so I suggest we name the setting "languageDir". Similar to "languageName" that already is present.

A html template could then look like this:

``` html
<html class="nojs" lang="{{ .Site.Language.Lang }}" dir="{{ .Site.Language.LanguageDir | default "ltr" }}">
```

And the language configurations like this:

```
languages:
  sv:
    weight: 1
    languageName: "Svenska"
  en:
    weight: 2
    languageName: "English"
  ar:
    weight: 3
    languageName: "العربية"
    languageDir: "rtl"
```

le0tan:
Closes #6550

### DIFF
--- a/docs/content/en/content-management/multilingual.md
+++ b/docs/content/en/content-management/multilingual.md
@@ -47,6 +47,13 @@ weight = 2
 linkedin = "https://linkedin.com/fr/whoever"
 [languages.fr.params.navigation]
 help  = "Aide"
+
+[languages.ar]
+title = "مدونتي"
+weight = 2
+languagedirection = "rtl"
+[languages.ar.params]
+linkedin = "https://linkedin.com/ar/whoever" 
 {{< /code-toggle >}}
 
 Anything not defined in a `languages` block will fall back to the global value for that key (e.g., `copyright` for the English `en` language). This also works for `params`, as demonstrated with `help` above: You will get the value `Aide` in French and `Help` in all the languages without this parameter set.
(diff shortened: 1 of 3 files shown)


bep:
This looks good; could you add a line in the documentation about this, too. In this repo.

docs/content/en/content-management/multilingual.md
```diff
@@ -47,6 +47,13 @@ weight = 2
 linkedin = "https://linkedin.com/fr/whoever"
 [languages.fr.params.navigation]
 help  = "Aide"
+
+[languages.ar]
+title = "مدونتي"
+weight = 2
+languagedirection = "rtl"
+[languages.ar.params]
+linkedin = "https://linkedin.com/ar/whoever"
```
bep: Drop that param (we use example.com for all example links, and that would not work here, so just drop it).

PR #7052, opened 2020-03-13
Make IsZero take the entire PageMeta into account

evankanderson:
I can't see where `set` is filled out (for example), by `DecodeBuildConfig`, which makes me suspect that `IsZero` is returning true in `pageMeta.applyDefaultValues` when it shouldn't.

All tests still passed with this change, but I'm not entirely sure where the right place to add a test would be.


evankanderson:
... nevermind, I figured it out -- `defaultBuildConfig` is further filled in by `WeakDecode`, not overwritten.

Closing.

PR #7055, opened 2020-03-14
Use .Lastmod for og:updated_time

dtip:
This PR fixes a bug (?) where `.Date` was being used for `og:updated_time` instead of `.Lastmod`. Couldn't find an open issue for this. It was quicker to fix it than to open one.

.Lastmod is the time at which the website was most recently updated,
rather than .Date which is the time at which the website content file
was created.


dtip:
Anyone planning to look at this? It's a minor change and fixes a bug.

dtip:
@bep Thanks!

Issue #7061, opened 2020-03-16
GetTerms: runtime error: invalid memory address or nil pointer dereference 


carlmjohnson:
Bug: If a page doesn't have its taxonomy set, .GetTerms panics.

Workaround (e.g. for topic):

```
{{ $topics := slice }}
{{ if .Param "topics" }}
  {{  $topic = .GetTerms "topics" }}
{{ end }}
```

### What version of Hugo are you using (`hugo version`)?

Hugo Static Site Generator v0.67.1/extended darwin/amd64 BuildDate: unknown

### Does this issue reproduce with the latest release?

Yes.

bep:
I tried to create a failing test for this, but failed ... So there must be something I don't understand.

carlmjohnson:
Hmm, it definitely happens on my site, but I am having trouble reproducing it as a Hugo test. Maybe it's because I'm using non-standard taxonomies?

carlmjohnson:
If you checkout https://github.com/spotlightpa/poor-richard/commit/f89ff87534f6e976d1a71de15ca1fed5b6d5c644#diff-5725f8ca9fd37c1438a052e56a003075 and change 

```
{{ $topic := "" }}
{{ if .Param "topics" }}
  {{  $topic = index (.GetTerms "topics") 0 }}
{{ end }}
```

to 

```
{{ $topic := index (.GetTerms "topics" | default slice) 0 }}
```

You get:

```

ERROR 2020/03/16 16:03:04 Render of "404" failed: execute of template failed: template: 404.html:56:5: executing "404.html" at <partial "ld-json.html" .>: error calling partial: execute of template failed: template: partials/ld-json.html:17:14: executing "partials/ld-json.html" at <partialCached "page-params.html" .Page .RelPermalink>: error calling partialCached: "/Users/cjohnson/src/github/poor-richard/layouts/partials/page-params.html:11:20": execute of template failed: template: partials/page-params.html:11:20: executing "partials/page-params.html" at <.GetTerms>: error calling GetTerms: runtime error: invalid memory address or nil pointer dereference
```

carlmjohnson:
I think I've figured out how to reproduce it. The `404.html` layout needs to be defined and reference `.GetTerm`.

carlmjohnson:
```
github.com/gohugoio/hugo/hugolib.(*Site).renderForTemplate
        /tmp/hugo/hugolib/site.go:1667
[...]

PR #7062, opened 2020-03-17
Fix GetTerms nil pointer deref?

carlmjohnson:
Refs #7061 

I don't understand Hugo internals, so this could be a totally bogus fix. But the test is real.


bep:
Thanks!

PR #7082, opened 2020-03-21
tpl: Extend Jsonify to support options

### LINKED ISSUE #5040: Allow "jsonify" to pretty print JSON

moorereason:
Fixes #5040

### DIFF
--- a/tpl/encoding/encoding.go
+++ b/tpl/encoding/encoding.go
@@ -17,6 +17,7 @@ package encoding
 import (
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"html/template"
 
 	"github.com/spf13/cast"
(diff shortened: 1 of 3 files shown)


bep:
I don't remember what we discussed in #5040 but thinking about it now I think I would prefer to use a dict option, e.g. `$foo  |  (dict "indent" "    ")`

Esp. if we should get more options in the future.

moorereason:
PTAL

zbayoff:
@moorereason  Thank you very much for this addition!! Is there any way you might update the Hugo Docs page as well to include the new options? https://gohugo.io/functions/jsonify/

cshoredaniel:
I would suggest you checkout the Hugo Docs repo and add it yourself.  
That would take some load off the Hugo developers, and it's an open 
source project to enable / encourage collaboration...

PR #7101, opened 2020-03-27
[helpers] Fix TrimShortHTML

### LINKED ISSUE #7081: Markdownify produces invalid HTML with paragraph followed by list
```
$ hugo version
Hugo Static Site Generator v0.67.1/extended linux/amd64 BuildDate: 2020-03-16T00:07:38Z
```

Here's a minimal example of what I'm seeing:
`content/_index.html`
```
---
content: |
  Here's a list:                                                                                                                                                                    
  * A
  * B
  * C
---
```

`layouts/index.html`
```
{{ .Params.content | markdownify }}
```

Hugo output in `public/index.html`
```
Here&rsquo;s a list:</p>
<ul>
<li>A</li>
<li>B</li>
<li>C</li>
</ul>
```

You can see that markdownify produced a close paragraph tag with no open paragraph tag...  So - I looked at the other bugs for this, and recommended solutions - including surrounding the whole `{{ .Params.content | markdownify }}` with open and close paragraph.  But that won't work in this case - because there is no second open paragraph tag in this example.

Now my trick to workaround this is to change `content/_index.html` to this:
```
---
content: |
  Here's a list:                                                                                                                                                                    
  * A
  * B
  * C
  
  &nbsp;
---
```

That produces this in `public/index.html`:
```
<p>Here&rsquo;s a list:</p>
<ul>
<li>A</li>
<li>B</li>
<li>C</li>
</ul>
<p> </p>
```

Which is at least valid HTML.  I think the logic markdownify uses to determine when to strip/not strip first/last paragraph tags is the problem here.

In my actual code I'm activating this issue using markdownify as a shortcode, trying to supply markdown inside of another shortcode's inner inside other shortcodes...  Not with content as part of frontmatter - but the result and workaround are the same.

satotake:
Fixes #7081
where some tags are siblings of p tag

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -44,6 +44,7 @@ var (
 	openingPTag        = []byte("<p>")
 	closingPTag        = []byte("</p>")
 	paragraphIndicator = []byte("<p")
+	closingIndicator   = []byte("</")
 )
 
 // ContentSpec provides functionality to render markdown content.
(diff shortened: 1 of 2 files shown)


bep:
The fix looks good, but you could help me by getting the commit message into shape.

For this PR I would write something ala:

```
helpers: Fix TrimShortHTML

Where some tags are siblings of p tag.


Fixes #7081 
```

Theimportant part being the title and the "Fixes" syntax.

satotake:
Fixed
I must be careful about it. Thanks

kc0bfv:
That's a pretty slick solution.

PR #7117, opened 2020-04-01
tpl: Fix bad rounding in NumFmt

### LINKED ISSUE #7116: lang.NumFmt Numbers greater than or equal to 5 are not rounded up correctly

<pre>
$ hugo version
Hugo Static Site Generator v0.68.3/extended darwin/amd64
</pre>

Yes it can be reproduced with the latest release with the following numbers:

```
{{ lang.NumFmt 2 927.675 }} -> 927.67 (expected 927.68)
{{ lang.NumFmt 2 1927.675 }} -> 1927.67 (expected 1927.68)

{{ lang.NumFmt 2 2927.675 }} -> 2927.68 (as expected)
```

sams96:
strconv.FormatFloat doesn't round properly sometimes, this adds a different method of rounding, fixes #7116


moorereason:
@bep, this PR looks good.  Rounding floats is hard, especially with configurable precision.  The approach taken here by @sams96 is the simplest all-around solution.  It's not bulletproof since it can overflow at `n*exp`, but solving that in a general way is not at all trivial.

The Travis CI builds originally failed to run properly.  I've restarted them.  Once they're green, we should be clear to merge. 👍

Issue #7095, opened 2020-03-26
`replace` and `replaceRE` don't work in shortcode anymore


Jos512:
As [reported on the forum](https://discourse.gohugo.io/t/replace-tableofcontents-id-in-rendered-toc/24222) and confirmed there by another user, string functions like `replace` and `replaceRE` don't work in shortcodes anymore.

This _page_ template code works:

~~~
{{ replace .TableOfContents "TableOfContents" "toc" | safeHTML }}
~~~

Creating the _shortcode_ equivalent doesn't perform the replacement operation:

~~~
{{ replace .Page.TableOfContents "TableOfContents" "toc" | safeHTML }}
~~~

The issue happens when I call the shortcode with `{{<toc>}}` and with `{{%toc%}}`.

### What version of Hugo are you using (`hugo version`)?

<pre>
Hugo Static Site Generator v0.68.3/extended windows/amd64 BuildDate: unknown
GOOS="windows"
GOARCH="amd64"
GOVERSION="go1.14.1"
</pre>

### Does this issue reproduce with the latest release?
Yes.

### Since when does the issue happen?
The problem started with [Hugo 0.55](https://github.com/gohugoio/hugo/releases/tag/v0.55.0). The issue wasn't present with the version before that, [Hugo 0.54](https://github.com/gohugoio/hugo/releases/tag/v0.54.0).

### Is there an example site to replicate?
Yes, I uploaded an example site here: https://github.com/Jos512/hugo-shortcode-issue

That repository has `example-post.md` call the `toc.html` shortcode with `{{<>}}` and `{{%%}}` format. The content of that shortcode is:

~~~
{{ replace .Page.TableOfContents "TableOfContents" "toc" }}
~~~

### How does the problem happen with the example site?
On Hugo 0.54, both `{{<toc>}}` and `{{%toc%}}` output:

~~~html
<nav id="toc"> <ul> <li> <ul> <li><a href="#heading-1">Heading 1</a></li> <li><a href="#heading-2">Heading 2</a></li> </ul></li> </ul> </nav> 
~~~

This is correct; `TableOfContents` is replaced by `toc`.

On Hugo 0.55, both `{{<toc>}}` and `{{%toc%}}` output:

~~~html
<nav id="TableOfContents">
<ul>
<li>
<ul>
<li><a href="#heading-1">Heading 1</a></li>
<li><a href="#heading-2">Heading 2</a></li>
</ul></li>
</ul>
</nav>
~~~

[...]

Issue #6811, opened 2020-01-27
Allow overriding tableOfContents levels in frontmatter


inwardmovement:
Would be great to allow overriding [tableOfContents](https://gohugo.io/getting-started/configuration-markup#table-of-contents) in a page's frontmatter.

inwardmovement:
keep?

Issue #6081, opened 2019-07-09
.TableOfContents generate "hahahugoshortcode" text if {{< icon >}} on ## level headers


divinerites:
Hello,

My `{{ .TableOfContents }}` works flawlessly. But when I decided to **add some icon on my headers**, the markdown parsing produced a funny result with `hahahugoshortcode` text 😸 :

from source markdown :

```md
## {{< icon name="fa-battery-1" >}} My title with icon 

my text 1

### My title without icon

my text 2
```

produce this HTML

```html
<nav id="TableOfContents">
<ul>
<li>
<ul>
<li><a href="#hahahugoshortcode-s0-hbhb-my-title-with-icon">HAHAHUGOSHORTCODE-s0-HBHB My title with icon</a>
<ul>
<li><a href="#my-title-without-icon">My title without icon</a></li>
</ul></li>


```

divinerites:
Ah ... may be this is already known, I finally found a reference on this [Bep's post](https://discourse.gohugo.io/t/hugo-0-40-help-test-it/11655) ...

plinss:
I found the fix for this, use the `{{% %}}` syntax for the shortcodes. e.g. 
```
## {{% icon name="fa-battery-1" %}} My title with icon 
```
That will cause the shortcodes to be rendered before being passed to the markdown processor (which generates the .TableOfContent variable)

subhashb:
@plinss's solution works well for icons. But using true shortcodes with separate <style> sections still generates this text in ToC.

bep:
You want to use the `{{%` delimiter in titles.

rodrigoalcarazdelaosa:
> You want to use the `{{%` delimiter in titles.

Is this confirmed to work with Hugo 0.72.0? I'm testing it and the icon shows in the first line of the body of the text, not in the title.

rodrigoalcarazdelaosa:
> > You want to use the `{{%` delimiter in titles.
> 
> Is this confirmed to work with Hugo 0.72.0? I'm testing it and the icon shows in the first line of the body of the text, not in the title.

This is still happening for me with Hugo 0.73.0.

rodrigoalcarazdelaosa:
> This is still happening for me with Hugo 0.73.0.

Also happening with Hugo 0.74.3

bep:
Note that any fix for this issue would be to issue some kind of warning/error instead of the haha... placeholder. The proper tix is what I say above. If you see some other issues related to this, that is something different.

[...]

PR #7127, opened 2020-04-04
Proposal: Add "ContentData" to customize TableOfContents

satotake:
This is proposal.  Further improvements will be required.
Any comments are welcome.

Add new Page Param ".TableOfContentsCollection" with goldmark.
This enables users to customize table of contents as they like.
You can be used it in shortcodes but have to use with `{{< >}}` (as non-markdown shortcode) because headers in contents cannot be determined until markdown shortcodes (`{{% %}}`) are rendered.

First, I was tackling the issue #7095 . It was harder than it looks.
Rather than fixing the single issue, I added the new param that gives Hugo some flexibility

Here is [usage example](https://github.com/satotake/hugo-test/tree/proposal-toc-collection)

---

Addresses #7095, #6811, #6081, #225

### DIFF
--- a/hugolib/page__per_output.go
+++ b/hugolib/page__per_output.go
@@ -35,7 +35,6 @@ import (
 	"github.com/gohugoio/hugo/tpl"
 
 	"github.com/gohugoio/hugo/helpers"
-	"github.com/gohugoio/hugo/markup/tableofcontents"
 	"github.com/gohugoio/hugo/output"
 	"github.com/gohugoio/hugo/resources/page"
 	"github.com/gohugoio/hugo/resources/resource"
@@ -142,7 +141,7 @@ func newPageContentOutput(p *pageState, po *pageOutput) (*pageContentOutput, err
 						cfg.TableOfContents.Ordered,
 					),
 				)
-				cp.tableOfContentsHeaders = toc.Headers
+				cp.contentData.TableOfContents = toc.Headers
 			} else {
 				tmpContent, tmpTableOfContents := helpers.ExtractTOC(cp.workContent)
 				cp.tableOfContents = helpers.BytesToHTML(tmpTableOfContents)
@@ -272,10 +271,10 @@ type pageContentOutput struct {
 	contentPlaceholders map[string]string
 
 	// Content sections
-	content                template.HTML
-	summary                template.HTML
-	tableOfContents        template.HTML
-	tableOfContentsHeaders tableofcontents.Headers
+	content         template.HTML
+	summary         template.HTML
+	tableOfContents template.HTML
+	contentData     page.ContentData
 
 	truncated bool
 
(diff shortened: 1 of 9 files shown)


bep:
I agree about the general idea, but not the param bit.

I have another issue somewhere describing a "page map"; the golden end goal would be that you could do

```
{{ range .Page.ContentMap.Headings }}

{{ end }}
```

Etc. etc. in a very efficient way. I need some help on the Goldmark side to get to the above, but it would be really cool.

That said, I admit that the above ideal end goal may be some time in the future, so I'm open to adding ToC as the first step ... If we could agree on a name (not params) for the container collection, e.g.:

```
{{ range .Page.ContentData.TableOfContents }}
{{ end }}
```

satotake:
Thank you for commenting @bep 

> it would be really cool.
Absolutely. 
I am afraid that I cannot find your description even now but I hope it will be done in the future.

I agree with your suggestions. Especially, I am appreciating your attention to names which interface with general user. 

I will update.

satotake:
Done.

- `ContentData` was adopted
- `TableOfContentsCollection (as param?)` was removed
- PR title was changed

bambooom:
Hi, will this feature be added in the future? I'd like to customize toc content by shortcode as I need to filter out some headings.

I've tried [this solution](https://discourse.gohugo.io/t/sections-wrapped-in-a-shortcode-do-not-appear-in-the-toc/21304/15?u=bambooom), but it throws out error `can't evaluate field Content in type *hugolib.ShortcodeWithPage`.
If I change to use `.Page.Content` instead of `.Content`, there won't be error, but the toc content is never rendered.

reinhrst:
Thanks for the link @bambooom , it works for me (in a partial, not in a shortcode). Expanded a bit to also catch the tag name and make it configurable (not sure how to access the markup.tableOfContents settings, so using custom params. It (obviously) needs CSS to style now.

```
{{ $startLevel := .Site.Params.tocStartLevel | default 2 }}
{{ $endLevel := .Site.Params.tocEndLevel | default 3 }}
{{ $tagRe := printf "h[%d-%d]" $startLevel $endLevel }}
{{ $tocRe := printf "<%s.*?>(.|\n])+?</%s>" $tagRe $tagRe }}
{{ $headers := findRE $tocRe .Content   }}
<toc>
      <ul class="menu-list">
          <li class="toc-h0"><a href="#page-top">{{.Title}}</a></li>
          {{ range $headers }}
              {{ $tagname := substr . 1 2 }}
              {{ $header := . | plainify | htmlUnescape }}
              <li class="toc-{{$tagname}}">
                  <a href='#{{ $header | anchorize }}'>{{ $header }}</a>
              </li>
        {{ end }}
      </ul>
</toc>
```

Still feels like a hack though; would be great if something like this proposal were to be added.

Asday:
I'd also like this feature to render only parts of a table of contents.  My use case is a "here's what's in this part of the document" at the top of a section, and the table of contents for only that section.  Currently what I'm doing is rendering the whole table of contents where I want it with a shortcode, then copy pasting the rendered HTML into another shortcode and editing it, and then using _that_ shortcode for the final render, which really is kinda sub-optimal.

(Note "section" here is the English use of the word, not the hugo use).

Issue #6985, opened 2020-02-28
hugo crashes with enableGitInfo=true and log.showSignature=true in ~/.gitconfig


jaakristioja:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
Hugo Static Site Generator v0.65.3-211BA42A linux/amd64 BuildDate: 2020-02-23T09:59:37Z
</pre>

### Does this issue reproduce with the latest release?
That was the latest release binary from https://github.com/gohugoio/hugo/releases/download/v0.65.3/hugo_0.65.3_Linux-64bit.tar.gz.

### So what is the problem
If I have `enableGitInfo = true` in my `config.toml` then `hugo server` crashes:

<pre>
$ hugo server
Building sites … panic: runtime error: index out of range [5] with length 1

goroutine 82 [running]:
github.com/bep/gitmap.toGitInfo(0xc000876000, 0x30, 0x1b70f21, 0x1, 0x0)
        /go/pkg/mod/github.com/bep/gitmap@v1.1.1/gitmap.go:128 +0x3f7
github.com/bep/gitmap.Map(0xc000040094, 0x21, 0x0, 0x0, 0x21, 0x1, 0x0)
        /go/pkg/mod/github.com/bep/gitmap@v1.1.1/gitmap.go:90 +0x4cd
github.com/gohugoio/hugo/hugolib.newGitInfo(0x1feb980, 0xc000342d80, 0xd, 0x401, 0x7f247bc9c400)
        /root/project/hugo/hugolib/gitinfo.go:41 +0x6a
github.com/gohugoio/hugo/hugolib.(*HugoSites).loadGitInfo(0xc000774b40, 0x20300000000000, 0x7f247be52fff)
        /root/project/hugo/hugolib/hugo_sites.go:378 +0x8b
github.com/gohugoio/hugo/hugolib.newHugoSites.func4(0x400, 0x7f247bc9c500, 0x20300000000000, 0x7f247be52fff)
        /root/project/hugo/hugolib/hugo_sites.go:344 +0x2e
github.com/gohugoio/hugo/lazy.(*Init).Do.func1()
        /root/project/hugo/lazy/init.go:97 +0x136
github.com/gohugoio/hugo/lazy.(*onceMore).Do(0xc0007b62c8, 0xc0006b8848)
        /root/project/hugo/lazy/once.go:53 +0x146
[...]

PR #7131, opened 2020-04-04
updated to gitmap version v1.1.2

dragtor:
see: https://github.com/gohugoio/hugo/issues/6985


bep:
Thanks, I'll merge once the test passes. For future contributions, please use "Fixes #1234" as that will close the issue automatically on merge.

PR #7143, opened 2020-04-07
Add roller2hugo to migrations.md

jlevon:
As I was unable to find a good-to-go converter from Apache Roller, I wrote a quick one.


bep:

jlevon:
Guess I'll fix the pull request link too then

PR #7145, opened 2020-04-07
Properly close <meta "robots" ... > tags

fekete-robert:
In the generated output HTML files, Hugo didn't close some meta tags, raising warning in strict HTML validity checks.
This PR addresses this issue.


davidsneighbour:
HTML5 does not need to "properly close" any most of the tags, definitely not META tags. This should be fixed on a higher level.

bep:

Issue #4780, opened 2018-05-26
On build: convert JPGs larger than 1M into WEBP


vitaly-zdanevich:
And produced HTML for WEBP must contains `picture` instead of `img` for fallback:

```
<picture>
  <source srcset="img/awesomeWebPImage.webp" type="image/webp">
  <source srcset="img/creakyOldJPEG.jpg" type="image/jpeg"> 
  <img src="img/creakyOldJPEG.jpg" alt="Alt Text!">
</picture>
```

bep:
Can you be more specific?

onedrawingperday:
Also WEBM is not supported by all browsers, so this couldn’t be implemented as things are now.

Maybe this file format will replace JPG or maybe not.

vitaly-zdanevich:
> Also WEBP is not supported by all browsers, so this couldn’t be implemented as things are now.

`picture` tag handle it - Firefox will show JPG.

onedrawingperday:
No support in Safari. https://caniuse.com/#feat=webm

vitaly-zdanevich:
> No support in Safari (desktop and iOS), Edge.

Safari and others support [picture](https://caniuse.com/#search=picture), please see [this article](https://css-tricks.com/using-webp-images/).

flother:
Quite unfairly, JPEG has picked up a bad reputation. With [`mozjpeg`](https://github.com/mozilla/mozjpeg), Mozilla have shown that the assumed inadequacies of JPEG images are caused more by unoptimised encoders than the format itself. You can [read more at Performance Calendar](https://calendar.perfplanet.com/2014/mozjpeg-3-0/).

As a quick example, I took [this 4896×3264 image](https://unsplash.com/photos/5UjBY00ToG0) and then resized and re-encoded it using `libwebp`, `libjpeg`, and Mozilla's drop-in replacement for the latter, `mozjpeg`.

```bash
$ cwebp -resize 800 533 339839-orig.jpg -o 339839.webp
$ convert -resize 800x 339839-orig.jpg TGA:- | cjpeg -outfile 339839-libjpeg.jpg -targa
$ convert -resize 800x 339839-orig.jpg TGA:- | /usr/local/opt/mozjpeg/bin/cjpeg -outfile 339839-mozjpeg.jpg -targa
$ ls -lS 339839*
-rw-rw-rw- 1 matt 5224595 May 29 09:22 339839-orig.jpg
[...]

PR #7155, opened 2020-04-10
resources/images: Adding webp support

### LINKED ISSUE #5924: Add image processing support for WebP image resources
I tried this code

```
{{ $imageResource := .Page.Resources.GetMatch "**/*.webp" }}
{{ $resized := $imageResource.Resize "180x" }}
{{ $resized.RelPermalink }}
```
but I get an error:

```
<$imageResource.Resize>: can't evaluate field Resize in type resource.Resource
```

The .ResourceType is correctly identified as image. If I replace that with a jpg version it works.

blaubaer:
This is based on the `cwebp` binary of the [WebP utilities of Google](https://developers.google.com/speed/webp/download).
If on the system the binary is installed methods like `$image.Resize "600x webp"` are available.

The reason why this is based on the binaries of [WebP utilities of Google](https://developers.google.com/speed/webp/download) because there is currently no golang native WebP implementation available. On the other side using C-bindings to access libraries just make the whole build process of hugo too complicated. To execute binaries is not my favorite but seems under this conditions the best compromise.

Fixes #5924
Supports #4780

### DIFF
--- a/docs/content/en/content-management/image-processing/index.md
+++ b/docs/content/en/content-management/image-processing/index.md
@@ -193,11 +193,15 @@ See https://github.com/disintegration/imaging for more. If you want to trade qua
 
 By default the images is encoded in the source format, but you can set the target format as an option.
 
-Valid values are `jpg`, `png`, `tif`, `bmp`, and `gif`.
+Valid values are `jpg`, `png`, `tif`, `bmp`, `gif`, `webp`.
 
 ```go
 {{ $image.Resize "600x jpg" }}
 ```
+
+{{% note %}}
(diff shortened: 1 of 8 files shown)


bep:
Thanks, but we need to determine if/how in #5924 before implementing this.

I don't think we have the bandwidth to implement this using C bindings.

blaubaer:
@bep I understand your concerns, but I want to point out that this PR does not change syntax or whatever. It "just" adding the webp encoding support with all the commands which are there before. This needs to be done in any way, also if you want to change the whole domain language or whatever. When you want to add other functions to "change" file types or whatever, you can do it afterwards.

PR #7162, opened 2020-04-12
use environment variable `HUGO_ASCIIDOCTOR_ARGS` to override the default arguments

### LINKED ISSUE #5842: AsciiDoc: Add support for global document attributes
I'm helping to migrate the Okta developer blog from Jekyll to Hugo. We're using both Markdown and AsciiDoc and Jekyll renders AsciiDoc using the following plugins:

```
group :jekyll_plugins do
  gem 'jekyll-asciidoc'
  gem 'asciidoctor-rouge'
end
```

We have most things working in Hugo, but we've discovered (with the help of @mojavelinux) that Hugo doesn't provide a way to pass configuration to an external helper command, like `asciidoctor`. Here's the code that executes Asciidoctor: https://github.com/gohugoio/hugo/blob/master/helpers/content.go#L626-L650

This means I have to define global attributes in every `.adoc` file. For example:

```
:page-liquid:
:source-highlighter: rouge
:icons: font
```

It'd be great if Hugo had a way to pass attributes from the Hugo configuration to the `asciidoctor` command. According to Dan, we're losing a lot of power without that ability.

gfreezy:
fixed #5842

### DIFF
--- a/docs/content/en/content-management/formats.md
+++ b/docs/content/en/content-management/formats.md
@@ -47,7 +47,7 @@ Some of the formats in the table above needs external helpers installed on your 
 
 Hugo passes reasonable default arguments to these external helpers by default:
 
-- `asciidoc`: `--no-header-footer --safe -`
+- `asciidoc`: `--no-header-footer --safe -` use environment variable **HUGO_ASCIIDOCTOR_ARGS** to override the default arguments.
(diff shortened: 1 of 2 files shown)


reitzig:
Superseded by https://github.com/gohugoio/hugo/pull/7281 ?

PR #7164, opened 2020-04-12
Read sass imports from virtual fs

mraerino:
When experimenting with running Hugo on fully-virtual filesystems (remote, in-mem) i noticed that it is not possible to use `ToCSS` with that setup, since it does not load imports from the hugo filesystem.

This makes sure that sass imports are going through the Hugo filesystem.

### DIFF
--- a/resources/resource_transformers/tocss/scss/tocss.go
+++ b/resources/resource_transformers/tocss/scss/tocss.go
@@ -18,6 +18,7 @@ package scss
 import (
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


bep:
I suspect you have another issue, we do resolve imports against the virtual fs, but we return the real filename (full path) to LibSass to open. Open a regular issue and describe your issue.

mraerino:
Opened https://github.com/gohugoio/hugo/issues/7165

PR #7184, opened 2020-04-18
Add ability to generate nginx 30X configuration

lermit:
Hi

This PR add the ability to generate an nginx configuration in order to handle aliases as real 30X HTTP response.

Details:
Add the ability to set a new build option `nginxAliases`. If set to true
when building a new `nginx.conf` file will be generated in the project's
root directory. This file will be filled with all configurated aliases
and can be directly include in website nginx configuration.

The main objective of this feature is to prevent the http page loading
(HTTP status of 200) containing only a meta refresh html tag, but
directly serving a 30X HTTP status reponse. This is mainly SEO related.

Note: I'm not a Go developer, maybe not a developer at all, so fell free to consider this PR as a POC and don't hesitate to criticize :).

### DIFF
--- a/config/commonConfig.go
+++ b/config/commonConfig.go
@@ -30,6 +30,7 @@ import (
 var DefaultBuild = Build{
 	UseResourceCacheWhen: "fallback",
 	WriteStats:           false,
+	NginxAliases:         false,
 }
 
 // Build holds some build related condfiguration.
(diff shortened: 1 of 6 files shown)


bep:
Thanks for this. The code looks good, but I suspect that this is already possible using Hugo's custom output formats.

We do it for Netlify's redirects here:

https://github.com/gohugoio/hugoDocs/blob/master/_vendor/github.com/gohugoio/gohugoioTheme/layouts/index.redir

(there is a output format definition in the config)

I don't see why this should not be possible for nginx.

lermit:
Oh ... I've may be lost 2 hours of my life :).

I'll give it a try and close this PR after.

Thank you for you answer.

lermit:
I've tested your solution and this is (like all you work) awesomely simple !

Thank you for your assistance.

Regards

Issue #7178, opened 2020-04-16
Page is no longer the descendant of top-level section in v0.69.0


reuixiy:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
Hugo Static Site Generator v0.69.0/extended linux/amd64 BuildDate: unknown
</pre>

### Does this issue reproduce with the latest release?

Yes.

---

![2020-04-16-15-51-39](https://user-images.githubusercontent.com/28993043/79432354-425c7e00-7ffe-11ea-9664-55ffa163e4fa.png)

```html
{{ $sectionPage := .CurrentSection }}
{{ $.Scratch.Delete "pages" }}

{{ range $.Site.RegularPages }}
    {{ if (.IsDescendant $sectionPage) }}
        {{ $.Scratch.Add "pages" (slice .) }}
    {{ end }}
{{ end }}

{{ $pages := $.Scratch.Get "pages" }}
{{ $pages }}
```

I think this may be related to #7096. 

Downstream issue https://github.com/reuixiy/hugo-theme-meme/issues/116#issuecomment-614369427 (Non-English).

Related code:

1. [tree-sections.html](https://github.com/reuixiy/hugo-theme-meme/blob/af709a0529c9d34022a0834602bd40c8ba2e57a1/layouts/partials/pages/tree-sections.html#L17-L25)
2. [list.html](https://github.com/reuixiy/hugo-theme-meme/blob/86bfc7748271b2e6c3e573c2555b0a85af723703/layouts/_default/list.html#L27-L32)

reuixiy:
Seems fixed in `v0.69.2`.

PR #7204, opened 2020-04-23
Fix IsAncestor and IsDescendant under subsection

tekezo:
Fix an issue that IsAncestor and IsDescendant always return false if the page is in subsection.

Related issue: https://github.com/gohugoio/hugo/issues/7178

## Detail

If the page is in subsection, ref1 and ref2 values are as follows in `IsDescendant`.

```
ref1: /blog/subsection__hb_/page4__hl_
ref2: /blog
```

So, the following original code always return false since `ref2.key+cmBranchSeparator (/blog__hb_)` is not matched.

```go
cmBranchSeparator = "__hb_"
strings.HasPrefix(ref1.key, ref2.key+cmBranchSeparator), nil
```

We have to check `ref2.key + "/"`, too.


tekezo:
Thank you for reviewing and merging!

PR #7208, opened 2020-04-24
allow asciidoc extensions settings to be configurable

azah:
Following the discussion in #6561, I reworked the asciidoc configuration.

---

* backends, safe-modes, and extensions are now whitelisted to the [popular (ruby) extensions](https://asciidoctor.org/docs/extensions/) and valid values.
* the default for extensions is to not enable any, because they're all external dependencies so the build would break if the user didn't install them beforehand.
* the default backend is `html5` because `html5s` is an external gem dependency.
* the default safe-mode is `safe`, explanations of the modes: https://asciidoctor-docs.netlify.app/asciidoctor/1.5/safe-modes/
* the config is namespaced under `asciidocext_config` and the parser looks at `asciidocext` to allow a future native golang `asciidoc`.

---

example config.toml:

```toml
[markup]
  [markup.asciidocext]
    extensions = ["asciidoctor-html5s"]
    safeMode = "safe"
    backend = "html5s"
    noHeaderOrFooter = true

```

### DIFF
--- a/markup/asciidocext/asciidocext_config/config.go
+++ b/markup/asciidocext/asciidocext_config/config.go
@@ -1,0 +1,65 @@
+// Copyright 2020 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package asciidoc_config holds asciidoc related configuration.
+package asciidocext_config
+
+// DefaultConfig holds the default asciidoc configuration.
+var (
+	Default = Config{
+		Backend:          "html5",
+		DocType:          "article",
+		Extensions:       []string{},
+		NoHeaderOrFooter: true,
+		SafeMode:         "safe",
+		SectionNumbers:   false,
+		Verbose:          false,
+	}
+
+	ExtensionsWhitelist = map[string]bool{
+		"asciidoctor-html5s":           true,
+		"asciidoctor-diagram":          true,
+		"asciidoctor-interdoc-reftext": true,
+		"asciidoctor-katex":            true,
+		"asciidoctor-latex":            true,
+		"asciidoctor-question":         true,
+		"asciidoctor-rouge":            true,
+	}
+
+	SafeModeWhitelist = map[string]bool{
+		"unsafe": true,
+		"safe":   true,
+		"server": true,
+		"secure": true,
+	}
+
+	BackendWhitelist = map[string]bool{
+		"html5":     true,
+		"html5s":    true,
+		"xhtml5":    true,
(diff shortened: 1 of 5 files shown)


bep:
Have you communicated this with @@muenchhausen ?

azah:
@bep I have not, other than mentioning it in the thread. If this is merged, @muenchhausen would have to rebase and make some minimal changes to add a `WorkingFolderCurrent` in `markup/asciidoc/asciidocext_config/config.go` and add the workingfolder logic to `convert.go`.

muenchhausen:
@azah I like your changes - they complete merge request #6561. I propose to work together on the existing fork. I would like to run some tests with your changes, before we propose them here.

bep:
Yes, this looks good. Can I suggest that we rename the Go package to `asciidocext` as well? To make room for a hopefully native solution in the future. We could just do that later, but that makes a messy Git history.

markup/asciidoc/convert.go
```diff
@@ -68,16 +67,70 @@ func (a *asciidocConverter) getAsciidocContent(src []byte, ctx converter.Documen
...
+
+	for _, extension := range cfg.Extensions {
+		if contains(asciidocext_config.ExtensionsWhitelist, extension) {
+			args = append(args, "-r", extension)
+			continue
+		}
+		a.cfg.Logger.ERROR.Println("Unsupported asciidoctor extension was passed in.")
+	}
+
+	if cfg.NoHeaderOrFooter == true {
```
bep, line 97: It reads better if you just write

```go
if cfg.NoHeaderOrFooter {
}
```

markup/asciidoc/asciidocext_config/config.go
```diff
@@ -0,0 +1,65 @@
...
+		Backend:          "html5",
+		DocType:          "article",
+		Extensions:       []string{""},
+		NoHeaderOrFooter: true,
+		SafeMode:         "safe",
+		SectionNumbers:   false,
+		Verbose:          false,
+	}
+
+	ExtensionsWhitelist = []string{
```
bep, line 29: I think it's easier to use a map[string]bool for these whitelists. Also, no need to export them (use lower case first letter).

azah: @bep I change it to `map[string]bool`, but I had to export the variables because the `asciidoc` module uses them in the `parseArgs()` method.

The reason `parseArgs` is in `convert.go` is because attempting to pass a `converter.DocumentContext` object would cause a cyclic dependency issue. @muenchhausen 's fork uses that `ctx`.

azah:
@bep I renamed the `asciidoc` module to `asciidocext`.

muenchhausen:
@azah I don't want to get into merge conflicts. Let me know when you are done, then I will continue...

azah:
@bep is there anything else I need to do? #6561 now looks like it depends on this being merged first.

muenchhausen:
Yesterday, I failled trying to rebase #6561 onto this. I will try tonight to get these changes by cherry pick.

azah:
@muenchhausen unfortunately I'm fairly busy at the moment. I tried rebasing as well but it does look a bit complicated. It may be easier to start a new branch based off of this one, and copy/paste your changes in. I can do that once I get more time if needed.

muenchhausen:
@azah no problem - I will try that. I think I will manage that the next days.

muenchhausen:
@azah your changes are now available in our combined pull request #7281 . If this is fine for you, please close this pull request.

PR #7236, opened 2020-05-02
Added tests to ensure that sorting same value elements will not change their order

Mipsters:
allows to chain sort operations  
use case: sort pages by both pinned boolean and date  
for example:  
page A:
```yaml
section: "A"
date: 2020-04-27T00:0:0+00:00
```
page B:
```yaml
section: "B"
date: 2020-05-27T00:0:0+00:00
```
page C:
```yaml
section: "A"
date: 2020-06-27T00:0:0+00:00
```
```
{{ sort (sort site.RegularPages "Date") "Params.section" }}
```
will always be ordered like so:  
page C  
page A  
page B

### DIFF
--- a/tpl/collections/sort_test.go
+++ b/tpl/collections/sort_test.go
@@ -166,6 +166,31 @@ func TestSort(t *testing.T) {
 			"asc",
 			[]map[string]mid{{"foo": mid{Tst: TstX{A: "a", B: "b"}}}, {"foo": mid{Tst: TstX{A: "c", B: "d"}}}, {"foo": mid{Tst: TstX{A: "e", B: "f"}}}},
 		},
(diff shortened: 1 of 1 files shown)


bep:
I think I understand the motivation (verify stable sort order), but I don't see how you test this here?  I don't see how your tests can fail even with a non-stable sort?

Mipsters:
This is more of a sanity test, sort of test driven development  
I am aware that there is not a great chance that it will fail, its more of a trend set, so that I could relay on the assumption that ordering same values will not change their order  
so that if someone change the behavior majorly and the test fails, they can look at this merge and understand the reasoning

bep:
I don't want to add tests that have no failing case.

PR #7257, opened 2020-05-07
error message: add link to documentation

stapelberg:


stapelberg:
This isn’t an issue, it’s a pull request, and it’s still trivial and should still be merged.

davidsneighbour:
Of course it's something that is needed, but a description-less pull request get's overlooked often.

IMHO (not a gohugo developer) it might be better to add some kind of system that links error-message-links to errors with a configuration file that is the single source of these links so that they might be updated later on easily if required.

PR #7268, opened 2020-05-13
deploy: Do not suppress .well-known/ directory

### LINKED ISSUE #6691: Make deploy compliant with IETF RFC 8615 by not skipping hidden directories (.well-known)
Hugo deploy currently skips hidden directories through [this line of code](https://github.com/gohugoio/hugo/blob/6b59b64f02341c67a3606d2bbb3b9934d6276a74/deploy/deploy.go#L448).

[IETF RFC 8615](https://tools.ietf.org/html/rfc8615) uses a ".well-known" directory to implement site-wide metadata such as security information. Sites like keybase.io use these to guarantee identity, but it's also used for vulnerability disclosure purposes such as the [security.txt ](https://securitytxt.org/)initiative:

“When security risks in web services are discovered by independent security researchers who understand the severity of the risk, they often lack the channels to disclose them properly. As a result, security issues may be left unreported. security.txt defines a standard to help organizations define the process for security researchers to disclose security vulnerabilities securely.” 

**Proposal: skip hidden directories, but include ".well-known"**

johnweldon:
Deployments ignore directories with a leading `.`, but should not ignore
certain well known 'hidden' directories like `.well-known/`

Fixes #6691

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -440,6 +440,21 @@ func (lf *localFile) MD5() []byte {
 	return lf.md5
 }
 
+// knownHiddenDirectory checks if the specified name is a well known
+// hidden directory.
+func knownHiddenDirectory(name string) bool {
+	var knownDirectories = []string{
+		".well-known",
+	}
+
+	for _, dir := range knownDirectories {
+		if name == dir {
+			return true
+		}
+	}
+	return false
+}
+
 // walkLocal walks the source directory and returns a flat list of files,
 // using localFile.SlashPath as the map keys.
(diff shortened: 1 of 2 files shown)


deploy/deploy.go
```diff
@@ -440,6 +440,21 @@ func (lf *localFile) MD5() []byte {
 	return lf.md5
 }
 
+// knownHiddenDirectory checks if the specified name is a well known
+// hidden directory.
+func knownHiddenDirectory(name string) bool {
```
johnweldon, line 445: I used this approach to open the door for configurable 'well known' directories in the future.

PR #7281, opened 2020-05-16
rework external asciidoc integration 

muenchhausen:
This pull request is a combination of #7208 and #6561. It was initially opened because of issue #5695. 

This pull request will solve the relative path problem with asciidoctor tooling. An include will resolve relatively, so you can refer easily to files in the same folder.

Also `asciidoctor-diagram` and PlantUML rendering works now, because the created temporary files will be placed in the correct folder.

This patch covers just the Ruby version of asciidoctor. The old AsciiDoc CLI EOLs in Jan 2020, so this variant is removed from code.

The configuration is completely rewritten and now available in `config.toml` under the key `[markup.asciidocext]`:

```toml
[markup.asciidocext]
    extensions = ["asciidoctor-html5s", "asciidoctor-diagram"]
    workingFolderCurrent = true
    trace = true
    [markup.asciidocext.attributes]
        my-base-url = "https://example.com/"
        my-attribute-name = "my value"
```

- backends, safe-modes, and extensions are now whitelisted to the popular (ruby) extensions and valid values.

- the default for extensions is to not enable any, because they're all external dependencies so the build would break if the user didn't install them beforehand.

- the default backend is html5 because html5s is an external gem dependency.

- the default safe-mode is safe, explanations of the modes: https://asciidoctor.org/man/asciidoctor/

- the config is namespaced under asciidocext_config and the parser looks at asciidocext to allow a future native golang asciidoc.

- `uglyUrls=true` option and `--source` flag are supported

- `--destination` flag is required

Follow the updated documentation under `docs/content/en/content-management/formats.md`. 
  
This patch would be a breaking change, because you need to correct all your absolute include pathes to relative pathes, so using relative pathes must be configured explicitly by setting `workingFolderCurrent = true`.


bep:
The code looks very good, but the build fails, and this also needs a rebase against master.

muenchhausen:
@azah @snipem Do you have any idea why the build fails here? We renamed the package from asciidoc to asciidocext - but I thought we found all parts(?)

snipem:
@muenchhausen Is it related to this line: https://travis-ci.org/github/gohugoio/hugo/jobs/687972704#L666 ? there are lot's of errors in front of that failed test. I don't know if that is the expected behavior.

muenchhausen:
@snipem yes correctly - I don't understand the correlation of our changes and the errors in the template `layouts/_default/baseof.html` (???)

snipem:
After looking into it and testing it locally the error is the following. Which is also shown on Travis

```
--- FAIL: TestPageWithMoreTag (0.28s)
398    page_test.go:714: Page content is:
399        "<div class=\"paragraph\"> <p>Summary Same Line</p></div> <div class=\"paragraph\"> </div> <div class=\"paragraph\"> <p>Some more text</p></div>"
400        Expected:
401        "<div class=\"paragraph\"> <p>Summary Same Line</p></div> <div class=\"paragraph\"> <p>Some more text</p></div>" ([])
402--- FAIL: TestPageWithDelimiter (0.27s)
403    page_test.go:576: Page content is:
404        "<div class=\"paragraph\"> <p>Summary Next Line</p></div> <div class=\"paragraph\"> </div> <div class=\"paragraph\"> <p>Some more text</p></div>"
405        Expected:
406        "<div class=\"paragraph\"> <p>Summary Next Line</p></div> <div class=\"paragraph\"> <p>Some more text</p></div>" (["ad"])
```

For some reason the asciidoc conversion is behaving differently than expected with an additional `<div class=\"paragraph\">` added to the generated html. Can you @muenchhausen  reproduce from your older branches and pull requests if this error was introduced by your asciidoctor refactoring before or if it was introduced while fixing the tests?

Older tests from your branch were failing because of the tests being incorrect. So we can't check when the error was introduced. I expect this error is older than the first fail but was not detected before.

snipem:
093a296 looks promising. Looks like you found the missing div tag.

muenchhausen:
@snipem yes, thanks a lot :) I hope we can now finalize this pull request. 
@bep please proceed.

snipem:
I'm currently trying to integrate this into my project [Monako](https://github.com/snipem/monako).  It seems that the `--source` flag is ignored by the asciidoc generation of diagrams if set. For example if I set `--source` to `compose` and `--destination` to `public`, the HTML files are generated to `compose/public` while the diagrams are copied to `public` on root.

```
compose
├── config.toml
├── content
│   ├── content
│   ├── docs
│   │   └── test
│   │       ├── README.md
│   │       ├── profile.png
│   │       ├── subfolder
│   │       │   ├── subfolderprofile.png
│   │       │   └── test_doc_asciidoc_include_me_subfolder.adoc
│   │       ├── test_doc_asciidoc.adoc
│   │       ├── test_doc_asciidoc_include_me.adoc
│   │       └── test_doc_markdown.md
│   ├── monako_menu_directory
│   │   └── index.md
│   └── themes
├── public <----- expected path with all the static HTML files and assets from Hugo
│   ├── 404.html
│   ├── asciidoc_toc_fix.min.58499385d89856294f477ff41bf382768345086326d4789e843c6f841f537de3.js
│   ├── book.min.fd9659ba79459b4d484275976dcfb37feeecc367c71a460cc480a6181acad16c.css
│   ├── categories
│   │   ├── Asciidoc
│   │   │   ├── index.html
│   │   │   ├── index.xml
│   │   │   └── page
│   │   │       └── 1
│   │   │           └── index.html
│   │   ├── Markdown
│   │   │   ├── index.html
│   │   │   ├── index.xml
│   │   │   └── page
│   │   │       └── 1
│   │   │           └── index.html
│   │   ├── Test
│   │   │   ├── index.html
│   │   │   ├── index.xml
│   │   │   └── page
│   │   │       └── 1
│   │   │           └── index.html
│   │   ├── index.html
│   │   └── index.xml
│   ├── docs
│   │   ├── index.html
│   │   ├── index.xml
...
public     <--- not expected, should be compose/public
└── docs
    └── test
        └── test_doc_asciidoc
            └── ditaa-diagram.svg

```

snipem:
I found another issue when the "uglyUrls" option is set. The diagram does not honour this option correctly and instead creates a folder named "markdown_file.html". Consequently, Hugo cannot create the converted  html file since there is already a folder with the same name. In this case Asciidoctor should go up one folder for the working dir.
```
Error: Error building site: failed to render pages: open [.....]/docs/test/test_doc_asciidoc.html: is a directory
```
Path edited by me.

reitzig:
Would also fix #5842, I think.

muenchhausen:
I just want to remember, that main purpose was to fix @fapdash 's issue #5695 from Feb 2019(!). The rebase is done, thanks for all community suggestions and further improvements - is anything missing for a final merge?

markup/asciidocext/convert.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/gohugoio/hugo/markup/internal"
+)
+
+/* ToDo: RelPermalink patch for svg posts not working*/
+type pageSubset interface {
+	RelPermalink() string
+}
+
+// Provider is the package entry point.
+var Provider converter.ProviderProvider = provider{}
```
SchnWalter, line 35: I see that you are using this exported variable to create a new provider, shouldn't this `var Provider` be inside a factory function in order to avoid creating an instance until it's really needed? You shouldn't have an import side-effect.

(Sorry if the question seems wrong, but I'm just learning Go.)

bep: The struct created is really cheap to create, so I see no win adding complexity making this more complex.

markup/asciidocext/asciidocext_config/config.go
```diff
@@ -0,0 +1,83 @@
...
+		NoHeaderOrFooter:     true,
+		SafeMode:             "unsafe",
+		SectionNumbers:       false,
+		Verbose:              true,
+		Trace:                false,
+		FailureLevel:         "fatal",
+		WorkingFolderCurrent: false,
+	}
+
+	ExtensionsWhitelist = map[string]bool{
```
bep, line 34: First, apologies for the tame taken for me to get back to this issue. I had so many "in-progress" notifications that I didn't catch that it was done and ready for review.

bep: The Go repository have been doing some renames of their whitelists/blacklists in the light of the recent events in the US, so we might at least try to avoid adding new ones. I suggest we rename these to AllowedExtension and Disallowed*

bep: I would prefer if we could unexport these whitelists (e.g. `allowedExtensions`); this would mean that you move all/most of the `parseArg` function to the config package, which would keep all the config handling in one place.

Other than that this PR looks ready to go.

azah: @bep what would moving it to the config package look like, in order to avoid cyclic imports?

bep: I'm not sure which cyclic imports you're talking about, but given this extract:

```go
var cfg = a.cfg.MarkupConfig.AsciidocExt
	args := []string{}

	if asciidocext_config.BackendWhitelist[cfg.Backend] && cfg.Backend != asciidocext_config.Default.Backend {
		args = append(args, "-b", cfg.Backend)
	}

	for _, extension := range cfg.Extensions {
		if !asciidocext_config.ExtensionsWhitelist[extension] {
			a.cfg.Logger.ERROR.Println("Unsupported asciidoctor extension was passed in. Extension `" + extension + "` ignored.")
			continue
		}

		args = append(args, "-r", extension)
	}
```

The above is basically decoding of configuration and I don't see any packages outside of `asciidocext_config` in the above ...?

muenchhausen: Asciidocext is an external extension. It would be also nice if we keep this extension self contained and leave parseArgs as private function where it is. What do you think?

bep: The parseArg can live where it is, my main point was to keep the "config code" in the config package, if possible.

But we can look at that later, it doesn't matter for the end user.

If you can rebase this against master (I see there is a conflict) and possibly also squash the commits, I will pull this in.

muenchhausen: The rebase was cumbersome. If the result is fine, please continue.

celtic-coder:
Great work on this by all concerned! Well done! 👍

bwklein:
How can I add `asciidoctor-mathematical` and `asciidoctor-bibtex` extensions to the whitelist?

azah:
@bwklein they would need to be added to the allowed list of extensions. These should be straightforward to make a pull request for.

In the meantime you could add them manually and build Hugo from source:
 https://github.com/gohugoio/hugo/blob/f0266e2ef3487bc57dd05402002fc816e3b40195/markup/asciidocext/asciidocext_config/config.go#L34

bwklein:
@azah Thank you, just made PR #7447 to add them in.

capsulecorplab:
I feel like this opens an opportunity to also add `asciidoctor-pdf` to the whitelist, but I'm not sure how that would look in usage or if it would require a breaking change to the Hugo API. Any thoughts?

snipem:
@capsulecorplab You should test it out. From what I've observed the generators return rendered strings with the HTML content of the formats that Hugo support. `asciidoctor-pdf` would return bytes instead of HTML.

00krishna:
Hey folks, I was wondering what the latest status on getting `asciidoctor-diagram` to work with Hugo was? I can see that this PR was merged, but even when I configure my `config.toml` file as in the original post comment, I still don't get any diagrams generated. Is there any updated documentation or a good example of using Hugo to generate asciidoctor-diagrams from an Asciidoc file?

muenchhausen:
This is an external helper intergation. If you want asciidoctor-diagram, you need to install it by running `gem install asciidoctor-diagram`.

PR #7352, opened 2020-06-03
Expand lookup order

KalleDK:
Can't find a spot where the lookup order for cli and environment are, so added it here. Please fell free to deny this request or change it as you see fit.


bep:

Also, if you consider the `config` directory, the lookup order for config is slightly more elaborated.

PR #7334, opened 2020-05-29
tpl/crypto: add hmac

edouardparis:

PR #7358, opened 2020-06-05
Add genDocsHelper mage target

moorereason:
Expose the functionality of the hidden `hugo gen docshelper` command as
a mage target.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -66,6 +66,7 @@ func flagEnv() map[string]string {
 	}
 }
 
+// Generate autogen packages
 func Generate() error {
 	generatorPackages := []string{
 		"tpl/tplimpl/embedded/generate",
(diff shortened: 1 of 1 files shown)


moorereason:
See https://github.com/gohugoio/hugo/pull/7334#issuecomment-639517032

bep:
I'm a little hesitant about this one, 3 reasons:

* It's not that much easier than doing `go run main.go gen docshelper > data.json`
* It requires us to remember to update the import path if we add more 
* I think it's enough to generate this file before every release

moorereason:
I understand.  Some comments:

* My main motivation for this PR is to make the docshelper command more visible.  For whatever reason, my brain isn't interested in remembering the exact command, so I have to go digging for it each time.
* I'm only interested in adding a mage target.  Executing `go run` in mage is fine with me.
* I prefer to update `data.json` myself (when I remember to) instead of waiting for a release.

bep:
Yes, with "go run" this would be good to go.

moorereason:
Simplified to use `go run`.

PR #7383, opened 2020-06-14
Fix aliases with relativeURLs

helfper:
This PR fixes alias' redirect URLs when using the `relativeURLs` setting. This applies to both aliases explicitly defined in the page front matter and aliases implicitly created for the default language (e.g. `/` -> `/en`).

Taking the language redirect as an example, when setting `relativeURLs` to `true`, currently, the redirect portion of `/index.html` file looks like this:
```
<meta http-equiv="refresh" content="0; url=/en" />
```

While the correct should be (the last `/` is not strictly necessary, but I added it to be consistent with other URLs generated by Hugo):
```
<meta http-equiv="refresh" content="0; url=./en/" />
```

Using the [example from Hugo's documentation](https://gohugo.io/content-management/urls/#example-aliases), it currently generates two "redirect" files like this:
* `/posts/my-original-url/index.html` -> `/posts/my-awesome-post/`
* `/2010/01/01/even-earlier-url.html` -> `/posts/my-awesome-post/`

After this change:
* `/posts/my-original-url/index.html` -> `../../posts/my-awesome-post/`
* `/2010/01/01/even-earlier-url.html` -> `../../../posts/my-awesome-post/`

(These are my first lines of go in a real project, so let me know if I'm doing things in a non-idiomatic way.)

### DIFF
--- a/hugolib/alias.go
+++ b/hugolib/alias.go
@@ -99,8 +99,11 @@ func (s *Site) publishDestAlias(allowRoot bool, path, permalink string, outputFo
 		OutputFormat: outputFormat,
 	}
 
-	return s.publisher.Publish(pd)
+	if s.Info.relativeURLs || s.Info.canonifyURLs {
+		pd.AbsURLPath = s.absURLPath(targetPath)
+	}
 
+	return s.publisher.Publish(pd)
 }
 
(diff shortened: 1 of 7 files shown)


bep:
This looks very good, appreciate the updated tests.

PR #7399, opened 2020-06-17
Fix incorrect directory name in quick-start

SIkebe:
Fixed a minor typo in quick-start guide. The name of the theme directory should be `themes` not `theme`.


bep:

PR #7424, opened 2020-06-24
Allow empty alt (alt="") in img element

yaaax:
It's important to allow an empty alt attribute for img element.
See https://www.w3.org/WAI/tutorials/images/decorative/


apexskier (approved):

apexskier:
I still think this is a valid and useful PR.

Issue #7291, opened 2020-05-19
Markdown Render Hook for Footnotes?


maelle:
I'd like to be able to use the markdownified content of a footnote as its footnote link title, to make the footnote content accessible on hover without using JavaScript.

I.e. at the moment

```markdown
footnote.[^bla]

[bla]: I am the text of the footnote
```

Will give

```html
 <p>footnote.<sup id="fnref:1"><a href="#fn:1" class="footnote-ref" role="doc-noteref">1</a></sup</p>

<li id="fn:1" role="doc-endnote">
<p>I am the text of the footnote <a href="#fnref:1" class="footnote-backref" role="doc-backlink">&#x21a9;&#xfe0e;</a></p>
</li>
```

And I'd like to get

```html
 <p>footnote.<sup  id="fnref:1"><a title="I am the text of the footnote" href="#fn:1" class="footnote-ref" role="doc-noteref">1</a></sup</p>

<li id="fn:1" role="doc-endnote">
<p>I am the text of the footnote <a href="#fnref:1" class="footnote-backref" role="doc-backlink">&#x21a9;&#xfe0e;</a></p>
</li>
```

I asked about this [in the forum](https://discourse.gohugo.io/t/how-to-add-title-tooltip-with-footnote-content-to-footnote-link/25498) but after reading about [Markdown hooks for headers](https://github.com/gohugoio/hugo/issues/6713) I wondered whether I was looking for a Markdown render hook for footnotes, and whether it might be useful for others. If not, I'll try to hack something else -- and I'm sorry if I missed something in the docs.

bep:
I would say that that needs to be solved in another way.

Issue #6807, opened 2020-01-26
footnoteReturnLinkContents not used by Goldmark


inwardmovement:
> v0.62.2/extended windows/amd64

The `footnoteReturnLinkContents` config parameter is not taken into account by Goldmark.

bep:
Is this configuration specificed for Goldmark?

inwardmovement:
I guess no: how can we customize the footnote return link content then?

As a workaround I did it with CSS:
```
.footnote-backref {
    visibility: hidden;

    &::before {
      content: "↑";
      visibility: visible;
    }
  }
```

PR #7133, opened 2020-04-05
Add render template hooks for headings

elihunter173:
This commit also

* Renames previous types to be non-specific. (e.g. hookedRenderer rather
  than linkRenderer)

Resolves #6713

This was implemented much in the same way the current hooks are implemented, pulling in method definitions from [goldmark](https://github.com/yuin/goldmark). It still needs testing and documentation, but on my personal website I was able to successfully replace the `replaceRE` partial hack described in the issue above with a `render-heading.html` template similar to the one described.

Issue #5087, opened 2018-08-16
New ToC and Footnotes data structures


bep:
Go 1.11 is right around the corner. With that comes Go modules, which should solve the versioning issues we currently have with an upgrade to Blackfriday v2. I have started some investigations in another project. 

The goal of this first "sprint" is to

1. Upgrade to BF2 in a way that makes it easy to plug in some hooks that we have described in other issues (URL rewriter etc.), allow for renderer per output format etc (HTML, LatEx ...).
2. Add some proper data structures for table of contents and foot notes

This issue discusses 2) Above.

For ToC we currently have this HTML `.TableOfContents` blob. We cannot remove that over night so I suggest we add a new "container" in Page that contains Markdown-related  metadata.

* Name? `.MD.TableOfContents`?

That object will in the first go contain two data structures:

* `.TableOfContents`
* `.Footnotes`

For `TableOfContents`, I suggest a list of:

```bash
Title
Children
Parent
```

For `Footnotes` I suggest a list of:

```
ID
Link
```

Instead of `Link` maybe we should just reuse the term `RelPermalink` etc.

Note that the `Footnotes` list is meant to be used to render "somewhere else" in a template, so we also need a way to not render it in the main content.

rdwatters:
Is it unrealistic to keep `.TableOfContents` with your default behavior above but instead add arguments for start/stop based on `h` element?

```
{{ .TableOfContents 2 3 }}
```

Or the same as your comment above by adding the `.MD` to my example?

rgov:
This goes well with the new render template hook for headings. There is now a `HeadingContext` class which could be used:

https://github.com/gohugoio/hugo/blob/2fd0a5a6781456e88745b370d12aaf5351a020ff/markup/converter/hooks/hooks.go#L37-L48

gj52:
> This goes well with the new render template hook for headings. 

Can we collect this heading structures in a map/list and inject it in render-toc.html?

PR #7427, opened 2020-06-26
Add support for Goldmark footnote render hooks

samuelhorwitz:
This pull request will allow footnote/endnote customization with Goldmark via the footnote extension render hooks Goldmark provides.

I simplified the Hugo-side render hooks API from what Goldmark exposes and made the way they work for the Hugo end user much more idiomatic in terms of Hugo. There are just two new render templates: footnote-link (the superscript index in flow) and footnotes (all the notes at the bottom, ranged over)

See #7291
See #6807
See #6804

### DIFF
--- a/docs/content/en/getting-started/configuration-markup.md
+++ b/docs/content/en/getting-started/configuration-markup.md
@@ -94,6 +94,7 @@ The features currently supported are:
 * `image`
 * `link`
 * `heading` {{< new-in "0.71.0" >}}
+* `footnote-link` and `footnotes`
 
 You can define [Output-Format-](/templates/output-formats) and [language-](/content-management/multilingual/)specific templates if needed.[^hooktemplate] Your `layouts` folder may look like this:
 
@@ -149,6 +150,39 @@ Text
 PlainText
 : The plain variant of the above.
 
+The `render-footnote-link` template will receive this context:
+
+Page
+: The [Page](/variables/page/) being rendered.
+
+Index
+: The footnote numeric index (indicating which number footnote it is in order)
+
+The `render-footnotes` template will receive this context:
+
+Page
+: The [Page](/variables/page/) being rendered.
+
+Footnotes
+: An array of footnote contexts (in ascending order)
+
+Footnote contexts:
+
+Page
+: The [Page](/variables/page/) being rendered.
+
+Ref
+: The internal name used in the markdown file to connect a footnote to it's content
+
+Index
+: The footnote numeric index
+
+Text
+: The rendered (HTML) text.
+
+PlainText
+: The plain variant of the above.
+
 #### Link with title Markdown example:
 
 ```md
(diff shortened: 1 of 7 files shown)


docs/content/en/getting-started/configuration-markup.md
```diff
@@ -149,6 +150,39 @@ Text
 PlainText
 : The plain variant of the above.
 
+The `render-footnote-link` template will receive this context:
+
+Page
+: The [Page](/variables/page/) being rendered.
+
+Index
+: The footnote numeric index (indicating which number footnote it is in order)
```
apexskier, line 159: Is this a zero-based index? (I'd assume so, but since the text content is 1-based I'd want to be sure)

docs/content/en/getting-started/configuration-markup.md
```diff
@@ -94,6 +94,7 @@ The features currently supported are:
 * `image`
 * `link`
 * `heading` {{< new-in "0.71.0" >}}
+* `footnote-link` and `footnotes`
```
apexskier, line 97: You might want to add a `{{< new-in ` shortcode here, like in the heading

apexskier (commented):
This is great, I have a use case for custom footnote rendering that this will solve!

I have a couple minor suggestions around documentation, but overall the code looks solid. I don't fully understand the push/pop position stuff but the tests make me confident it works.

FYI, when I wrote documentation for the heading render hooks feature (https://github.com/gohugoio/hugo/pull/7133) I did it in https://github.com/gohugoio/hugoDocs (https://github.com/gohugoio/hugoDocs/pull/1100). I don't really understand the relationship between that repo and this one though.

bep:
Sorry for the delay. Thanks for this. These render hooks are a little bit too heavy to enable for all kinds of Markdown elements. I thought long and hard before I added it for images and links, but those was kind of a must. I will close this down for now, we need to revisit this problem later.

swamidass:
I'm hopefully this can be renewed. It would be very helpful for our use case.

swamidass:
@bep I still think this would be really helpful, for a whole range of reasons, particularly for academic/scholar users of hugo.

rootwork:
@bep I also think this would be really helpful. While shortcodes can accomplish a lot of things, I think it makes sense that when functionality is already provided by Goldmark, as footnotes are, render hooks should be provided. The academic use of footnotes is not a small use-case, and having to use relatively heavy shortcodes (`{{< f "1" >}}` is as short as it could be, compared to the considerably easier to type `[^1]`) makes for unnecessary complication.

That said, as to this particular PR I think the default render hook should generate the same HTML as Goldmark, which from what I can see wraps the footnotes list in `<div class="footnotes" role="doc-endnotes">` and prefaces it with an `<hr>`. Providing the same HTML as Goldmark would allow there to be no disruption when the default render hook is added.

cmpute:
I want to vote for this feature! It's hard to customize the footnote area without using JS. I'm also a using this a lot in my writings, and I've also seen a lot of footnotes in the blogs I read about programming and research.

> These render hooks are a little bit too heavy to enable for all kinds of Markdown elements. I thought long and hard before I added it for images and links, but those was kind of a must.

In my opinion, it's kind of confusing to close this PR without given an alternative approach. If this customization ability cannot be achieved without a hook, then how is it not a must?

> I will close this down for now, we need to revisit this problem later.

Is there any further discussion on this? Maybe it's time to revisit it now? @bep Thanks!

cmpute:
If #5087 can be implemented, then it's not a problem anymore, but that issue seems no progress as well..

PR #7433, opened 2020-06-27
Add support for native Org dates in frontmatter

sometimesfood:
When using Org mode with Hugo, it would be nice to have support for [Org timestamps](https://orgmode.org/manual/Timestamps.html) for the `DATE` header.

At the moment (hugo v0.72.0), `DATE` only supports Hugo's native syntax in Org files, therefore this is parsed correctly:
```
#+DATE: 2020-06-10
```

While this is not:
```
#+DATE: <2020-06-10 Wed>
```

### Expected result

Using the following template to format a page's `Date` attribute should produce the correct date:
```
{{ dateFormat "January 2, 2006" .Date }}
```

### Actual result
Output for `#+DATE: 2020-06-10`:
```
June 10, 2020
```

Output for `#+DATE: <2020-06-10 Wed>`:
```
January 1, 0001
```

Since Org mode defaults to the latter syntax, it would be great to have support for that in Hugo. Discussing this issue with @niklasfasching, he explained that this would fit better into Hugo than into go-org because of how Org frontmatter is handled by Hugo.

This pull request adds a simple parseORGDate function that does not touch the date value if it does not match an Org timestamp regexp, so that users who are currently using Hugo's native syntax do not experience any breakage.

### DIFF
--- a/parser/metadecoders/decoder.go
+++ b/parser/metadecoders/decoder.go
@@ -18,6 +18,7 @@ import (
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/gohugoio/hugo/common/herrors"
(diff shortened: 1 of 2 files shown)


bep:
This looks good to me. @niklasfasching ?

sometimesfood:
I just noticed I left a named capture in the regexp that is not used anymore (`?P<date>`). Sorry for that. I'll force-push an update to remove it and simplify the regexp.

niklasfasching:
LGTM

sometimesfood:
Thanks for your fast response!

Since I don't know if updated commits currently trigger a notification on GitHub: I removed the superfluous named capture in baa7a9d78eeb5f6efd67ded0617bbc576a8d1d51 (replacing 96d4e130f5f2837c26c940367c979307e95b8c1b) and left the rest of the commit untouched.

PR #7439, opened 2020-06-29
deps: Update go-org to v1.3.0

niklasfasching:
- Add support for image links (i.e. link with image as description)
- Kinda breaking: Wrap headlines and headline+content in containers as emacs
  Org mode does. Could break people's styles - I don't expect it, but who knows
  what counts as sane when it comes to css.

  Before:

  ```
  <h2 id="headline-1">FOO</h2>
  foo bar
  ```

  After:

  ```
  <div id="outline-container-headline-1" class="outline-2">
  <h2 id="headline-1">FOO</h2>
  <div id="outline-text-headline-1" class="outline-text-2">
  foo bar
  </div>
  </div>
  ```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/muesli/smartcrop v0.3.0
 	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646 // indirect
 	github.com/nicksnyder/go-i18n v1.10.0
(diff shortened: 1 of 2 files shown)


bep:
> Could break people's styles - I don't expect it, but who knows
what counts as sane when it comes to css.

Don't worry, it's me who'll get the shouting :-)

Issue #7321, opened 2020-05-26
Feature request: ESBuild pipeline


carlmjohnson:
[ESBuild](https://github.com/evanw/esbuild) is a very fast JavaScript bundler written in Go. It would be nice if in addition to minifying out of the box, Hugo could also bundle JS files. 

Usage might looks like:

```html
{{ $jsOptions := dict "minify" true "sourceMap" false "module" true }}
{{ $js := resources.Get "js/entrypoint.js" | toJSBundle $jsOptions | fingerprint }}
<script type="module" src='{{ $js.RelPermalink }}'></script>
```

And entrypoint.js could import files in the ES6 style:

```javascript
import { onLoad } from './utils.js';
import myFunc from './my-func.js';

onLoad(myFunc);
```

With all the concatenation and tree shaking happening automatically.

bep:
See https://github.com/evanw/esbuild/issues/152

bep:
How did that project get under my radar? Oh, well, hope the Hugo boss accepts this proposal.

carlmjohnson:
I thought you were the Hugo boss? If it's not you, then is it [him](https://en.wikipedia.org/wiki/Harald_V_of_Norway)?

PR #7446, opened 2020-07-01
Add support for Rollup JS asset bundling

remko:
**Warning: This patch is not finished yet. The documentation still needs to be written, but submitting to get confirmation that there is interest in finishing this.**

This patch adds support for bundling JavaScript assets with [Rollup](https://rollupjs.org/guide/en/). 

If you want to use JavaScript on your website that uses external dependencies (or that is organized internally into separate files/modules), all the external dependencies need to be bundled into the final `.js` file. This is exactly what Rollup does. 

This plugin is used in the same way as the Babel plugin:

    {{ $script := resources.Get "myscript.js" | rollup | minify | fingerprint }}
    <script src="{{ $script.Permalink }}"></script>

Where `myscript.js` can contain external dependencies, that will be bundled in the final result:

    import $ from "jquery";
   
    $.ajax(....);

Can also be used with Babel earlier in the asset chain:

    {{ $script := resources.Get "myscript.js" | babel | rollup | minify | fingerprint }}
    <script src="{{ $script.Permalink }}"></script>

The implementation is almost exactly the same as the Babel resource processor. 

The current version of the patch works, but the documentation needs to be written. I didn't do this yet because I wasn't sure that there was interest in merging this.

### DIFF
--- a/resources/resource_transformers/rollup/rollup.go
+++ b/resources/resource_transformers/rollup/rollup.go
@@ -1,0 +1,163 @@
+// Copyright 2020 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rollup
+
+import (
+	"io"
+	"os"
+	"os/exec"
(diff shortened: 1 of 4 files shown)


bep:
Not having too much opinion about this, but we have this issue:

#7321

remko:
I didn’t know ESBuild. A priori, I’m happy with anything that can bundle JavaScript in Hugo. 

Rollup has a large plugin ecosystem, and is widely used in production, but comes at the order of magnitude of execution time. That said, ESBuild seems to have a large feature set, and I would guess that if some Rollup plugin does something that ESBuild doesn’t, it can be moved to a (slow) Babel stage. I can imagine Hugo’s goals are more aligned with ESBuild than with Rollup. 

If wanted, I can probably adapt this patch to use ESBuild instead.

bep:
Now that we have ESBuild in place, I don't think we can add another (even if Rollup has features ESBuild doesn't). But thanks for this, the implementation looks solid.

PR #7449, opened 2020-07-02
Add support for ESBuild asset bundling

remko:
Adds support for  ESBuild pipeline, as requested in #7321

The current implementation goes through a temporary file, because ESBuild doesn't support in-memory inputs for bundling yet. (see https://github.com/evanw/esbuild/issues/212). The temporary file isn't created with Afero, because
- Hugo uses Afero 1.2.2, and Afero only supports TempFile wildcards in 1.3
- This will hopefully be useless in the future anyway, when ESBuild supports in-memory inputs

Still needs documentation, but gathering feedback first.


bep:
Thanks for this, the code looks good. But since adding some API in Hugo is very hard to remove/change, we need to get the basis more or less right out of the gate, but I'm thinking out loud, please don't your time by incorporating any of the changes/things I talk about.

* We have many "product name" type of functions in `resources`, e.g. (`resources.PostCSS`). Wen we add "native" JS bundling/transform support, I think we need to make it a feature that is named after "what it does", .e.g. `js.Transform` and `js.Build`.
* It's unfortunate that we cannot use Hugo's virtual filesystem in the Build part of this (Transform should be fine), but I guess we have to start somewhere. I have looked at the upstream code, and it should not be hard to add an option to allow a custom filesystem implementation, but the owner of that repo needs to be convinced it's a great idea ...

@regisphilibert what do you think about the naming discussion above?

regisphilibert:
To properly answer I would need to understand the properties and attributes of both methods of the suggested `js` keyword.

I'm trying to write a code example, but fail to grasp what `js.Transform` would do (though I suspect Babel) and what attribute it would take (though I suspect of a slice of js resources)
So about the `js` keyword and the method you suggest:
Would `js.Transform` be used for Babel, and `js.Build` for ESBuild?

Do you have a code example of both `js` methods being used to reference a bundle? All I can think of is:
```
{{ with resources.Match "js/*.js" }}
  {{ $asset := js.Build . }}
  {{ $asset = js.Transform $assets "ES2020" }}
  <script src="{{ $asset.RelPermalink }}"></script>
{{ end }}
```
Where Build bundles and Transform applies the ES wished transformations...

bep:
I think the short answer would be:

* `js.Transform` = `resoures.Babel` on steroids (as in: a million times faster): it converts one JS file from a format/standard/react to another. Typically used to support older browsers using newer syntax, but lots of use cases.
* `js.Build` takes 1 input file and resolves ES module imports (node_modules and relative for now I guess) and does tree shaking (remove unused code) etc. and bundles into 1 JS file. I assume that you can do transformation in `js.Build` as well, but not sure.

regisphilibert:
Awesome, I get it. 

Ok, this might sound a bit far fetched, but I'll just throw this out there.
For the sake of consistency with what we already have for image resources and to avoid the addition of yet another "function/keyword/shorthand" , couldn't we introduce those as methods on the resource:
```
{{ with resources.Get "js/index.js" }}
  {{ with .Build "ES2020" }}
    <script src="{{ .RelPermalink }}"></script>
  {{ end }}
{{ end }}
```
In the future we could bring other functions in alignment...

- `$jsAsset.Transform` > `resources.Babel`
- `$jsAsset.Build `
- `$cssAsset.Transform` > `resources.ToCSS`/`resources.PostCSS`

bep:
I suspect that we've had this discussion before ...

* So the image implementation supersedes the resources transformation package. I'm not sure I would have implemented it this way had I started fresh, not sure.
* The beauty of the "function pipes" is that @remko can just add another function and it does not need to know anything about the surroundings (a good thing, according to [Robert C. Martin](https://en.wikipedia.org/wiki/Single-responsibility_principle)
* The other beauty of pipes is ... they can be piped, so `$css | resources.PostCss | minify` vs. `$css.PostCSS.Minify`.
* I think that the method approach could work for "cross cut" concerns (transformations that would be applicable for most Resources, e.g. `.Minify`, but I'm not sure `.Transform` is "cross cut"?)

I don't see how this avoids "adding another" thing. If we want to make `resources.Babel` work like  `js.Transform` we could easily just call that function.

regisphilibert:
Yes I see.

The only thing I have against `js` is that it evokes a file extension or a media type, even though it could be used with files of a different extension (`jsx` etc...). Just like if we add `css.Transform` it could be used on `scss` files or else. 

I like `templates.Exists` because it is agnostic of file extension or media type, it's just an internal concept people can rely on. So I would suggest something more aligned with this concept like:
`scripts.Transform`, `scripts.Build`? 
And maybe in the future `styles.Transform` for CSS etc...

bep:
>The only thing I have against js is that it evokes a file extension or a media type, even though it could be used with files of a different extension (jsx etc...). 

It certainly belongs to "a group of media types" where `application/javascript` is the most common.  I don't think `Transform` works that fantastic as a function name, as "transformation" is what all of these functions does. So `styles.ToCSS` looks descriptive and good to me. For JavaScript I think we should just call the spade a space (even if there are variants, like JSX -- which is JS DSL) and say `js.Transform` (or maybe `javascript.Transform`); the little vague `Transform` kind of works because there are so many output variations (not just `CSS`).

regisphilibert:
I agree with everything said above :)

remko:
A note on the current implementation itself: the mechanism to go through a temporary file doesn't work well in watch mode it seems. The temporary file is generated in the assets/ dir (to make relative paths work), but this triggers a rebuild, which in turns creates a new temporary file, which triggers ... 

If there's currently no easy way to ignore files ad-hoc in Hugo, I guess we'll have to wait for the API to change so it can get its input from memory.

bep:
>A note on the current implementation itself: the mechanism to go through a temporary file doesn't work well in watch mode it seems.

If you name the temp file to match one of these, that should work:

```
istemp := strings.HasSuffix(ext, "~") ||
			(ext == ".swp") || // vim
			(ext == ".swx") || // vim
			(ext == ".tmp") || // generic temp file
			(ext == ".DS_Store") || // OSX Thumbnail
			baseName == "4913" || // vim
			strings.HasPrefix(ext, ".goutputstream") || // gnome
			strings.HasSuffix(ext, "jb_old___") || // intelliJ
			strings.HasSuffix(ext, "jb_tmp___") || // intelliJ
			strings.HasSuffix(ext, "jb_bak___") || // intelliJ
			strings.HasPrefix(ext, ".sb-") || // byword
			strings.HasPrefix(baseName, ".#") || // emacs
			strings.HasPrefix(baseName, "#") // emacs
		if istemp {
			continue
		}
```

remko:
> If you name the temp file to match one of these, that should work:

Great, that works!

bep:
Given the naming discussion above, what should we name the new package? I will add a note that even if other "script types" are involved (TypeScript, JSX), the end result is (currently) always some form of JavaScript.

[](https://api.gh-polls.com/poll/01ECPXRB4XGPT9D6W3M6VXRYFW/js.Build/vote)
[](https://api.gh-polls.com/poll/01ECPXRB4XGPT9D6W3M6VXRYFW/javascript.Build/vote)
[](https://api.gh-polls.com/poll/01ECPXRB4XGPT9D6W3M6VXRYFW/script.Build/vote)

remko:
Is there a reason you need a `"mybundle.js"` parameter in `javascript.Build`?  If `$myjs` is a single file, the filename can be preserved. If you want to bundle files together (such as what `Concat` does), you can either pipe it into `Concat` first, or you can create your own bundle `.js` that `import`-s all the files you want bundled, so `Build` doesn't really need to support slices?

bep:
>Is there a reason you need a "mybundle.js" parameter in javascript.Build? If $myjs is a single file, the filename can be preserved.

I just quickly wrote the "example" above without thinking. But we need a way to set the targetPath (with a default if not set), e.g. https://gohugo.io/hugo-pipes/scss-sass/#readout

Does the Build support TypeScript as input? If yes, we also need to consider the change in MIME types.

The thing is, if I do:

```
{{ $js := resources.Get "myscript.js" }}
{{ $bundle := $js | js.Build }}
```
I suspect there will be use cases where you would want to use both of the above, so we cannot use the same name as the only option.

This gets more obvious if this would be possible:

```
{{ $ts := resources.Get "myscript.ts" }}
{{ $bundle := $js | js.Build }}
```

What would be the extension in the end result?

bep:
@remko to conclude: If you rename the template package to match `js.Build` (ref the vote above), then I will merge this. There are still some API-questions that needs to be decided, but that I can ponder on my own.

remko:
Should be done.

Some notes:

- I left the client in `resource_transformers` (because nothing changed there). It could be folded into the resource package.
- I did not add a method mapping (e.g. `jsBuild`)
- I copy-pasted `resolveArgs` from `resources` (because `js.Build` needs the same behavior), but didn't look at specializing it for the use or moving it to a shared function.
- The ESBuild issue for in-memory input was just closed ( https://github.com/evanw/esbuild/issues/212 ). I'll wait until a release to remove the workaround with the temporary file

resources/resource_transformers/js/js.go
```diff
@@ -55,17 +55,17 @@ func New(fs *filesystems.SourceFilesystem, rs *resources.Spec) *Client {
...
-type esbuildTransformation struct {
+type jsTransformation struct {
 	options Options
 	rs      *resources.Spec
 	sfs     *filesystems.SourceFilesystem
 }
 
-func (t *esbuildTransformation) Key() internal.ResourceTransformationKey {
-	return internal.NewResourceTransformationKey("esbuild", t.options)
+func (t *jsTransformation) Key() internal.ResourceTransformationKey {
```
bep, line 65: All of this looks good, I have one final "suggestion":

As we will (most likely) get more transformations in this package, I suggest that you name this file `build.go`, the key `jsbuild` and the struct `buildTransformation`. That will leave room for `jstransform` etc.

bep: Also, this need a test. The simplest is to just expand on the tests in /hugolib.

remko: @bep Is this enough as a test? This tests the filesystem interaction.

If you want me to test various options, I'll have to create a test at a lower level (i don't need the filesystem for those)

bep:
Thanks, I have squashed/rebased this branch into https://github.com/gohugoio/hugo/pull/7473 -- I need to do some testing/adjustments on my own before I pull it into master, but great job on this one, thanks, much appreciated!

PR #7466, opened 2020-07-07
Add DisablePathUnicodeEscaping config option

RaeesBhatti:
Allows users to opt out of escaping of non-ASCII path.


RaeesBhatti:
This PR is attempting to solve https://github.com/gohugoio/hugo/issues/7467. I'm not familiar with the codebase, any help would be appreciated.

RaeesBhatti:
Not stale!

PR #7473, opened 2020-07-12
Add js.Build asset bundling

### LINKED ISSUE #7321: Feature request: ESBuild pipeline

bep:
Fixes #7321

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/bep/tmc v0.5.1
 	github.com/disintegration/gift v1.2.1
 	github.com/dustin/go-humanize v1.0.0
-	github.com/evanw/esbuild v0.6.1
+	github.com/evanw/esbuild v0.6.2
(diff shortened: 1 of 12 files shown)


carlmjohnson:
🤯 Incredible.

PR #7484, opened 2020-07-15
Fix broken fragment identifier on the Taxonomy Templates docs page

mariozig:
Small typo in docs caused a broken link to the "Taxonomy List Templates" section.

On https://gohugo.io/templates/taxonomy-templates/ the following text contains a broken link using a singular fragment identifier -- it should be plural: 

`Order the way content associated with a taxonomy term is displayed in a taxonomy list template`


bep:

PR #7494, opened 2020-07-18
Fix Asciidoctor args

### LINKED ISSUE #7493: Inconsistent arguments in Asciidoctor
At the moment, the handling of the arguments passed to Asciidoctor CLI is very inconsistent.

A few examples:

With no `asciidocext` options in the `config.toml`, the args passed to Asciidoctor are:
```
--no-header-footer --trace -
```
---
`config.toml`:
```
[markup.asciidocext]
trace = true
```
Asciidoctor args:
```
--no-header-footer --trace --trace -
```
---
`config.toml`:
```
[markup.asciidocext]
trace = false
```
Asciidoctor args:
```
--no-header-footer --trace -
```
---
`config.toml`:
```
[markup.asciidocext]
verbose = false
```
Asciidoctor args:
```
--no-header-footer -v --trace -
```
---
`config.toml`:
```
[markup.asciidocext]
verbose = true
```
Asciidoctor args:
```
--no-header-footer --trace -
```
---

It seems the implementation had two goals in mind:

1. Allow Hugo to have different defaults from Asciidoctor CLI
2. Minimize the arguments passed to Asciidoctor by skipping options that are already the default in Asciidoctor CLI

The problem is that to achieve this, we need two sets of default values: Hugo's and Asciidoctor's. However, the current implementation only has one set of defaults which for some options it seems to mean Hugo's defaults and for others it means Asciidoctor's defaults.

I think there are three options to solve the issue:

1. The cleanest one would be to just have two default configs: Hugo's and Asciidoctor's, so that both goals above can be easily met
2. Only have Hugo's defaults and don't minimize arguments - avoids the need for a second var with config defaults, but it may be too noisy in the logs while debugging
3. Specify explicitly which options are meant to be Hugo's defaults and which ones are Asciidoctor's defaults. In this case, default values for options of type string have to either match Asciidoctor's defaults, or not be minimized and always be set explicitly; default values for options of type boolean can mean Hugo's defaults and still be minimized, because in Asciidoctor CLI, boolean flags are always false by default.

The option 3 seems to have been the aim of the current implementation, because the defaults are not exactly the same as Asciidoctor's defaults, but then it assumes that's the case when trying to minimize them.

helfper:
Fixes #7493

### DIFF
--- a/markup/asciidocext/asciidocext_config/config.go
+++ b/markup/asciidocext/asciidocext_config/config.go
@@ -14,21 +14,26 @@
 // Package asciidoc_config holds asciidoc related configuration.
 package asciidocext_config
 
-// DefaultConfig holds the default asciidoc configuration.
-// These values are asciidoctor cli defaults (see https://asciidoctor.org/docs/user-manual/)
 var (
+	// Default holds Hugo's default asciidoc configuration.
 	Default = Config{
 		Backend:              "html5",
-		DocType:              "article",
 		Extensions:           []string{},
 		Attributes:           map[string]string{},
 		NoHeaderOrFooter:     true,
 		SafeMode:             "unsafe",
 		SectionNumbers:       false,
-		Verbose:              true,
-		Trace:                false,
+		Verbose:              false,
+		Trace:                true,
 		FailureLevel:         "fatal",
 		WorkingFolderCurrent: false,
+	}
+
+	// CliDefault holds Asciidoctor CLI defaults (see https://asciidoctor.org/docs/user-manual/)
+	CliDefault = Config{
+		Backend:      "html5",
+		SafeMode:     "unsafe",
+		FailureLevel: "fatal",
 	}
 
 	AllowedExtensions = map[string]bool{
@@ -72,7 +77,6 @@ var (
 // Config configures asciidoc.
 type Config struct {
 	Backend              string
-	DocType              string
 	Extensions           []string
 	Attributes           map[string]string
 	NoHeaderOrFooter     bool
--- a/markup/asciidocext/convert.go
+++ b/markup/asciidocext/convert.go
@@ -69,7 +69,6 @@ func (a *asciidocConverter) getAsciidocContent(src []byte, ctx converter.Documen
 	}
 
 	args := a.parseArgs(ctx)
-	args = append(args, "--trace")
 	args = append(args, "-")
 
 	a.cfg.Logger.INFO.Println("Rendering", ctx.DocumentName, "with", path, "using asciidoctor args", args, "...")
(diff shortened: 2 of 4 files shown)


markup/asciidocext/asciidocext_config/config.go
```diff
@@ -64,6 +71,13 @@ var (
 		"manpage":   true,
 	}
 
+	AllowedDocType = map[string]bool{
```
muenchhausen, line 74: I did not see a usecase for these doctypes in Hugo context. If there is a usecase, we should mention them in the documentation.

markup/asciidocext/convert.go
```diff
@@ -81,17 +80,17 @@ func (a *asciidocConverter) parseArgs(ctx converter.DocumentContext) []string {
...
+
+	args = a.appendArg(args, "--doctype", cfg.DocType, asciidocext_config.CliDefault.DocType, asciidocext_config.AllowedDocType)
 
 	for _, extension := range cfg.Extensions {
 		if !asciidocext_config.AllowedExtensions[extension] {
 			a.cfg.Logger.ERROR.Println("Unsupported asciidoctor extension was passed in. Extension `" + extension + "` ignored.")
 			continue
 		}
 
-		args = append(args, "-r", extension)
```
muenchhausen, line 94: The list of arguments can get quite long. Personally, I would prefer to use the short forms because the calls to `asciidoctor` get generated and are not typed in manually by humans.

markup/asciidocext/asciidocext_config/config.go
helfper: I don't have a use case for it either, but it was already in the type `asciidocext_config.Config` without being used at all, so I just added the missing parts. I'll remove it completely in that case.

markup/asciidocext/convert.go
helfper: I'd argue it's the opposite: short flags are more useful when written manually, while in scripts/code, I usually prefer the long form because it makes it more obvious for other people reading the code and code is read more often than it is written. Not sure if there is an overall preference in Hugo for short vs long flag format, or if it is decided on a case by case basis. I'm happy to revert this change if it's the latter.

helfper:
@muenchhausen I've updated the PR based on your comments.

markup/asciidocext/asciidocext_config/config.go
```diff
@@ -25,12 +24,20 @@ var (
 		NoHeaderOrFooter:     true,
 		SafeMode:             "unsafe",
 		SectionNumbers:       false,
-		Verbose:              true,
-		Trace:                false,
+		Verbose:              false,
```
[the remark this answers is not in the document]
muenchhausen: I double checked your new standard values. They are fine for me!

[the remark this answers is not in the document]
helfper: I changed `Verbose` and `Trace` defaults here to keep the current default behaviour, as these values were not reflecting real defaults as you can see in #7493.

I agree with changing `WorkingFolderCurrent` default to true. That's the most useful option.

PR #7507, opened 2020-07-22
Fix data race in templating code

andreasf:
This should fix the concurrent map access crashes reported in #7293.

Changes:
* Use existing "muFunc" lock in Template.AddParseTree and Template.Lookup
* Call Template.Lookup from outside instead of accessing map directly

The additional lock doesn't appear to impact performance noticeably.

### DIFF
--- a/tpl/internal/go_templates/texttemplate/exec.go
+++ b/tpl/internal/go_templates/texttemplate/exec.go
@@ -396,7 +396,7 @@ func (s *state) walkRange(dot reflect.Value, r *parse.RangeNode) {
 
 func (s *state) walkTemplate(dot reflect.Value, t *parse.TemplateNode) {
 	s.at(t)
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this; but this needs to be fixed upstream (https://github.com/golang/go) -- we have a (hopefully) temporary fork with some controlled patches to the execution of templates, but I'm not opening up that fence to bug fixes in the parser etc., that I don't have any plans to maintain here.

bep:
Sorry for the delay on this. If this will not be fixed for Go 1.16 then we need to fix it on our side.

andreasf:
It's not your fault, I didn't have time in the past weeks to look further into the discussion on the Go side.

jmooring:
Superseded by https://github.com/gohugoio/hugo/commit/9650e568418a316e71ad94d7e27caf544a4a2d0d

PR #7540, opened 2020-08-08
Generate custom error pages

### LINKED ISSUE #5066: Generate custom error pages
Referencing [https://discourse.gohugo.io/t/how-to-create-custom-error-pages/755](https://discourse.gohugo.io/t/how-to-create-custom-error-pages/755)

It would be nice if Hugo could generate custom error pages. Not only for 404 errors. It could use a regex like `/\d{3}.html/` for pages in `/layouts/`.

satotake:
Render layouts/(4xx|5xx).html
resolves #5066

### DIFF
--- a/hugolib/page__new.go
+++ b/hugolib/page__new.go
@@ -177,7 +177,7 @@ func newPageFromMeta(
 
 }
 
-// Used by the legacy 404, sitemap and robots.txt rendering
+// Used by the legacy error pages, sitemap and robots.txt rendering
 func newPageStandalone(m *pageMeta, f output.Format) (*pageState, error) {
 	m.configuredOutputFormats = output.Formats{f}
 	m.standalone = true
--- a/hugolib/site.go
+++ b/hugolib/site.go
@@ -1215,7 +1215,7 @@ func (s *Site) render(ctx *siteRenderContext) (err error) {
 			}
 		}
 
-		if err = s.render404(); err != nil {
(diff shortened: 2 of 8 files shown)


bep:
Thanks for this.

I find it a little bit odd that this is restricted to HTML, which is the format where I would least expect a 403 etc. to be useful ...?

satotake:
Thank you

Probably I miss your point but I assume that they are always HTML.
For example, these error pages are used for [CloudFront's error pages](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesErrorPages)
Could you describe more details?

bep:
>Probably I miss your point but I assume that they are always HTML.

I have never seen a 400 or 500 error for a static Hugo site. I could however maybe these error endpoints being used by some JSON service, not sure.

For HTML I guess 403's could happen.

satotake:
> I have never seen a 400 or 500 error for a static Hugo site. 

This is true when we use Hugo and static site hosting service (such as [Netlify](https://www.netlify.com/)).
Instead, these error pages are generated in cases where Hugo is combined with a front server.
For example, Nginx for the purpose of auth or proxy. 
These cases can raise 500 etc and have the chance to serve customized error pages.
The use cases may be minor, I mean, out of Hugo's scope. 
If so, I will close this PR.

PR #7570, opened 2020-08-18
Remove last comma from schema.html template

0urobor0s:
With the introduction of delimit in #683 it is possible to use it
instead of range while also removing the last comma in keywords tag.


0urobor0s:
This still happens in master but is not critical.
The relevant url is [https://schema.org/keywords](https://schema.org/keywords)

PR #7573, opened 2020-08-19
Add Victor Hugo Kontent boilerplate

makma:
Add Victor Hugo Kontent boilerplate into starter kits.


bep:

Issue #7579, opened 2020-08-20
Newline after "Building Sites …" message?


jwarner112:
Hey there,

I tend to do print-debugging until something goes real wrong. So when I'm writing a shortcode or template file, I do a lot of `hugo server ...` and see a lot of
```
hugo server -D
Building sites … WARN 2020/08/19 23:45:42 warn: this is debug message 0
WARN 2020/08/19 23:45:42 warn: this is debug message 1
WARN 2020/08/19 23:45:42 warn: this is debug message 2

                   | EN  
-------------------+-----
  Pages            | 28  
  Paginator pages  |  1  
  Non-page files   |  0  
  Static files     |  1  
  Processed images |  2  
  Aliases          |  4  
  Sitemaps         |  1  
  Cleaned          |  0  

Built in 1698 ms
Watching for changes in /home/jeff/code/website.com/{archetypes,assets,content,data,i18n,static,themes}
Watching for config changes in /home/jeff/code/website.com/config.toml
Environment: "development"
Serving pages from /home/jeff/code/website.com/public
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
```
This means a lot of the time I miss the very first line -- on the same line as `Building Sites ...`, so I was thinking we could add a newline afterwards?

```
hugo server -D
Building sites …
WARN 2020/08/19 23:45:42 warn: this is debug message 0
WARN 2020/08/19 23:45:42 warn: this is debug message 1
WARN 2020/08/19 23:45:42 warn: this is debug message 2
```

It's a small nit-pick and I'm issuing a PR for it in a sec. Maybe there's some reason this hasn't been done, but otherwise it would at least be a tiny bit easier to read. Thoughts?

PR #7580, opened 2020-08-20
Adds newline after 'Building Sites' message

jwarner112:
As mentioned in #7579, the logs keep tripping me up so I made this change, and I'm hoping it helps annoy others less as well.

Going over the contributors' guidelines:
- [x] Sign the CLA
- [ ] Test cases for the code -- this was so small I'm not sure if you'd make this rule apply? I manually verified it, at least
- [x] `go fmt` was run, though it made no changes
- [ ] No docs, same reasoning as Test Cases. Let me know if you think otherwise!
- [x] Squashed to one commit
- [x] `mage check` finished with exit code `0`
- [x] Git commit message guidelines followed as best as I understood them

 I think that's everything but I'm happy to discuss and make changes. 🤘


bep:
2 comments:

* I suspect this change will break the "hide when done" logic; whether that logic is worth keeping is worth discussing.
* Also, have a look at the commit message guidelines (ref. the commit message title/subject)

jwarner112:
> I suspect this change will break the "hide when done" logic; whether that logic is worth keeping is worth discussing.

I'm not sure what you mean about the "hide when done" logic; I've run all the tests with no failure, and haven't found it yet to test manually. Where would I find it?

bep:
>I'm not sure what you mean about the "hide when done" logic;
 
The "Building sites..." message is just showing when the site is... building. It's removed from terminal when done. If you add a newline, that logic is broken.

jwarner112:
Ooh, okay. I guess I've got a bug to report, then -- it's never hidden itself for me, to the point that I didn't know it was even supposed to. Here's what I've always gotten:

```
$ hugo
Building sites … WARN 2020/08/21 00:04:19 I'm a little teapot
WARN 2020/08/21 00:04:19 short and stout

                   | EN  
-------------------+-----
  Pages            | 28  
  Paginator pages  |  1  
  Non-page files   |  0  
  Static files     |  1  
  Processed images |  2  
  Aliases          |  4  
  Sitemaps         |  1  
  Cleaned          |  0  

Total in 96 ms
$
```

I would say that if that behavior were fixed, my problem would go away -- however, I think it would reach and display my `warnf` statements, and that the first of those might get erased when the `Building Sites ...` clears?

bep:
So, I added that clever construct, but I have seen it failing in some situations myself, so what I suggest you do as part of this is:

1. Add a newline (like this PR)
2. Remove the ifTerminal(hideCursor) logic (remove the function)
3. Edit the text from "Building sites …" to ... "Start building sites …"

Issue #4691, opened 2018-05-02
Use alternative tree root instead of index.md (such as README.md)


robcresswell:
It'd be great to be able to configure the root file that hugo assumes when building its front matter, when creating documentation sites, for the following reasons:

- GitHub / GitLab naturally render README.md files  in each directory when browsing source code online.
- Many modern web frameworks, especially in JS ecosystems, have adopted component-based code bases, where each component contains its own documentation, tests, and implementation. This may look something like:

```console
$ tree src/components/table
src/common/table
├── README.md
├── index.js
├── table.vue
└── test.js
```

This seems to me to be extremely well aligned with Hugo's tree structure for organisation, other than the hardcoded index filename requirement.

I think this is useful because it helps developers consume documentation in this manner how ever they'd like; either reading directly from github, or in a more professional generated output from Hugo.

bep:
I see your point, but it looks like an expensive feature. If we look beyond the hardcodings, I don't see a way to do this without either 1) add some kind of config option or 2) break stuff. 1) is doable, of course, but that would kind if limit the "content portability" that the issue is trying to improve in the first place.

Once people get used to these "index.md" files etc., having people share articles (as bundles possibly) in zip files and just add it to the content and it just works ... 

As a contributor to some magazine I would love to not have to do: "Hey, editor, here is my article. Oh, I have used a non-standard naming scheme, so you must toggle some setting and have Hugo version at least ..."

robcresswell:
I see your point, and 1) is the route I would suggest. Also with regards to it not being a standard for Hugo, it is a "sort of standard" for many source control sites, which is why it would be great to be able to take advantage of that.

evankanderson:
Hey, this is of interest to me, and looks like it may be less of a challenge than it was in 2018 (at least based on a fairly quick glance through the code for `_index`).

Is this still of interest/would a PR be accepted?

evankanderson:
Looking at this further, it looks like a matter of plumbing the configuration into https://github.com/gohugoio/hugo/blob/master/hugofs/files/classifier.go#L84 and a few other places.

My initial thought was to make this a [build option](https://gohugo.io/content-management/build-options/), but that means that we may need to open bundle files during the filesystem traversal to read front matter.

@bep , do you have a preference on implementation here (global option, or build option)?

bep:
So, my recent work in this area has been slowly making this more friendly for "non file pages" (aka pages from data).

[...]

PR #7599, opened 2020-08-26
Implement alternate branch root files for #4691

evankanderson:
Fixes https://github.com/gohugoio/hugo/issues/4691

This is a different approach than the first time.

I still need to write up the docs, but there is an e2e test. There are also several tests where the full `FileMetaInfo` isn't present in the filesystem -- what's your preference on how those go in or get cleaned up?


create/content_template_handler.go
```diff
@@ -97,8 +97,7 @@ func executeArcheTypeAsTemplate(s *hugolib.Site, name, kind, targetPath, archety
 
 	if name == "" {
 		name = f.TranslationBaseName()
-
-		if name == "index" || name == "_index" {
+		if f.FileInfo().Meta().Classifier().IsBundle() || name == "index" {
```
evankanderson, line 100: There are 3 test failures if I remove the "index" test here: `TestNewContent/lang-11` through `TestNewContent/lang-13`.

deps/deps.go
```diff
@@ -87,9 +87,6 @@ type Deps struct {
 
 	Metrics metrics.Provider
 
-	// Timeout is configurable in site config.
-	Timeout time.Duration
```
evankanderson, line 91: This was never used, it turned out.

hugofs/files/classifier.go
```diff
@@ -100,22 +100,13 @@ func ClassifyContentFile(filename string, open func() (afero.File, error)) Conte
...
 	}
-
-	if strings.HasPrefix(filename, "_index.") {
-		return ContentClassBranch
-	}
-
-	if strings.HasPrefix(filename, "index.") {
-		return ContentClassLeaf
-	}
-
```
evankanderson, line 118: This determination moves into the only caller, `filter_fs.go`.

hugolib/fileInfo.go
```diff
@@ -89,7 +89,18 @@ const (
...
+		switch fim.Meta().Classifier() {
+		case files.ContentClassBranch:
+			return bundleBranch, true
+		case files.ContentClassLeaf:
+			return bundleLeaf, true
+		default:
+			return bundleNot, false
+		}
+	}
+	// Fallback path, generally used for tests
```
evankanderson, line 103: It turns out this fallback isn't needed, removing.

evankanderson:
(I'll squash commits after I have the docs in)

bep:
Hey,

Thanks for this. The code looks good, but I have 2 general comments. Regarding this I'm mostly convinced that:

1. Any config for this should live on the mount config (as in: overridden value for "_index" for a given mount)
2. This should be normalized when  I call `.Path`. That sentence is obviously simplified, but there will be situations where 2 content files take up the same place and needs some merge, e.g.: (content1)/posts/README.md and (content2)/posts/_index.md. This will be even more relevant/complicated once we start talking about non-markdown content. Doing this will also tell us if this "if _index" logic is too much spread out/duplicated.

evankanderson:
For 1, let me take a look at where modules tie in. To be honest, I didn't find the documentation on modules at all until you gave me that keyword. From a quick look, it looks like a reasonable suggestion.

For 2, see the changes in `filter_fs.go`, where the first (highest priority) prefix in the list is promoted to be a bundle, and other files that may be present are treated as regular content pages. I _think_ that this allows a migration for sites (like my own) that are currently using kludges like `_index.md` files that just contain a `{{%readfile file="README.md"%}}` as content (this breaks things like use of shortcodes inside the included README). You can see this working (I think) in the e2e test here:
https://github.com/gohugoio/hugo/blob/f859d32d175dbff1693d0e43b02721a8a2fba949/hugolib/hugo_sites_build_test.go#L863
```
branchBundlePrefix = ["README", "test"]
...
		"sub2/test.md", `---
title: lower priority, just content
---
# Test content`,
....
	b.AssertFileContent("public/sub2/test/index.html", "Test content")
```

evankanderson:
I'm assuming that the `branchBundlePrefix` option would only apply to content mounts?

evankanderson:
For #1, it looks like the best option is to pass the list of files via FileMeta from the composite filesystem in `rootmapping_fs.go` to the filter list in `filter_fs.go`. This seems like it ha a few problems:

1. It *should* be the case that the FileMeta is the same for all the files in `applyMeta`, but there's nothing that guarantees that is true. This could potentially produce strange results based on mounts in subdirs.
2. I'm not sure about the layering; it seems strange to have the content determination split across the layers like this.

What would you think of having the RootMappingFs include a "branch score" on each file in a dir, and having `applyMeta` look at the top-scoring files to determine if the directory has a branch or leaf index file?

bep:
>It should be the case that the FileMeta is the same for all the files in applyMeta, but there's nothing that guarantees that is true. 

I'm not sure I'm following:

* Every file inside /content will be member of a file mount
* That file mount can have it's `branchName="README"` set or not.

The above should be enough to determine if `Name()` (or whatever FileInfo method) should return `_index.foo` or not.

Issue #7547, opened 2020-08-11
Add post encryption to content


GoGoGoDoge:
I want to add a post encryption feature to hugo. 
Right now there is a work around as proposed by this author in this repo:
https://github.com/Li4n0/hugo_encryptor/blob/master/README.md

It would be nice to integrate this solution into hugo.

GoGoGoDoge:
It is actually related to #7154 
The repo I mentioned above implements in a way to use python with beautiful soup to parse a specific html node and encrypt it on demand.

moorereason:
First, cryptography is surprisingly hard.  Generally speaking, only cryptographers should write crypto solutions.

Second (and perhaps making my first point), the examples linked to above use AES-CBC.  AFAIK, CBC encryption is now considered weak due to padding oracle attacks.

Third, the easiest solution would be to use a verified, turn-key solution such as nacl [secretbox](https://pkg.go.dev/golang.org/x/crypto/nacl/secretbox?tab=doc), but the algorithms used by nacl are not widely supported (by crypto-js or WebCrypto).

Forth, I, for one, don't want to be responsible for maintaining a turn-key solution in Hugo.  I'm not a cryptographer.  The most I'd be willing to support is a general crypto API via template functions that essentially exposes the Go stdlib API, but even that would be need some careful thinking.  Which all feels like a lot of work for a small reward.

I'm interested in the idea of a general crypto API, but I'll have to thinking about it.

@bep, what are your thoughts?

GoGoGoDoge:
Thank you for the detailed response. This is indeed a lot of work for a small reward. I have to admit that encrypted web page is not a common use case in static website.

moorereason:
@Lednerb,
As I said before, I would only be willing to support nacl secretbox.  I'm not interested in supplying a javascript solution around the template functions.  I only mentioned javascript and WebCrypto to point out the difficulty in writing a client-side companion piece to go along with nacl secretbox.

Would a nacl secretbox implementation solve your need?  It would look something like this, I suppose:

```
{{ $Password := "change this password to a secret" }}
{{ $enc := secretbox.Seal $Password "hello world" }}

{{ $enc | base64Encode }} ->
rjfsV4uz4m9XV+znsnm/bNfHDkOKrafW0uIi0P2nL/0GskOXc8N08WlRJ3zq+1chMpvN

{{ secretbox.Open $Password $enc }} ->
hello world
```

[...]

PR #7605, opened 2020-08-28
tpl: Add functions for hex encoding and secretbox encryption

moorereason:
Add a new hexEncode and hexDecode to the encoding namespace.

Add a new secretbox namespace (under crypto) for nacl secretbox
operations, Seal and Open.

Updates #7547

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,6 +57,7 @@ require (
 	github.com/yuin/goldmark-highlighting v0.0.0-20200307114337-60d527fdb691
 	go.opencensus.io v0.22.0 // indirect
 	gocloud.dev v0.15.0
+	golang.org/x/crypto v0.0.0-20190422183909-d864b10871cd
 	golang.org/x/image v0.0.0-20191214001246-9130b4cfad52
 	golang.org/x/net v0.0.0-20191209160850-c0dbc17a3553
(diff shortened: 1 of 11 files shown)


moorereason:
@bep, let me know what you think of this.  I'm not a cryptographer (no surprise there!), but NaCl was made to be easy and full-proof.

bep:
> easy and full-proof.

Was that typo intentional?

I think we need to demonstrate "usefulness" here before we add this. The only type of bugs I'm worried about here is what we've seen in Outlook (I think) where they have for some odd reason sent encrypted messages as plain text. 

"We have test", you say.

Yes, but I have seen stranger things happen where people comment out code paths and tests to debug something, then forgets to enable it ...

moorereason:
> Was that typo intentional?

Sort of.  The whole point of NaCl is to not be a footgun.  It's about as easy as crypto can get, and it's about the max level of complexity I'm willing consider exposing in Hugo.

Usefulness discussions are in the original issue.

Part of my desire to create this PR was to see what the implementation would look like.  What sharp edges are there?  Can I make a sub-package under crypto and import it into the template func namespace?

I share your concern with user mistakes, however, I don't know how to defend against that.  Any ideas?  I assume you mean site creators not Hugo developers.

bep:
>I assume you mean site creators not Hugo developers.

No, I mean you and me.

PR #7615, opened 2020-09-01
fix: Links are not crawlable

MarlonLuan:
Google Lighthouse

Links are not crawlable 

Search engines may use `href` attributes on links to crawl websites. Ensure that the `href` attribute of anchor elements links to an appropriate destination, so more pages of the site can be discovered.

Uncrawlable Link

```
<a class="page-link" aria-label="Previous">
```


jmooring:
Superseded by several previous PRs.

PR #7650, opened 2020-09-09
markup/asciidocext: Fix AsciiDoc TOC with code

### LINKED ISSUE #7649: AsciiDoc TableOfContents doesn't handle `code` properly
When a title in AsciiDoc contains `` `inlined code` ``, it doesn't appear in the table of contents and multiple entries are created with the text surrounding it.

For example, this AsciiDoc:
```
== Some `code` in the title
```

Will generate a TOC with two entries pointing to the same place:

* Some
* in the title

(This is handled correctly in Markdown.)

helfper:
Running this adoc (https://asciidoctor.org/docs/install-toolchain/) through Hugo with Hugo Learn Theme, currently produces:

old_toc

With this PR:

new_toc

Fixes #7649

### DIFF
--- a/markup/tableofcontents/tableofcontents.go
+++ b/markup/tableofcontents/tableofcontents.go
@@ -40,19 +40,19 @@ type Root struct {
 }
 
 // AddAt adds the header into the given location.
-func (toc *Root) AddAt(h Header, y, x int) {
-	for i := len(toc.Headers); i <= y; i++ {
+func (toc *Root) AddAt(h Header, row, level int) {
+	for i := len(toc.Headers); i <= row; i++ {
 		toc.Headers = append(toc.Headers, Header{})
 	}
 
-	if x == 0 {
-		toc.Headers[y] = h
+	if level == 0 {
+		toc.Headers[row] = h
 		return
 	}
 
-	header := &toc.Headers[y]
+	header := &toc.Headers[row]
 
-	for i := 1; i < x; i++ {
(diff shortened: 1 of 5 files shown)


markup/tableofcontents/tableofcontents.go
```diff
@@ -40,19 +40,19 @@ type Root struct {
 }
 
 // AddAt adds the header into the given location.
-func (toc *Root) AddAt(h Header, y, x int) {
-	for i := len(toc.Headers); i <= y; i++ {
+func (toc *Root) AddAt(h Header, row, level int) {
```
bep, line 43: This does not look related.

bep (changes requested):

markup/tableofcontents/tableofcontents.go
helfper: It's only related in the sense that it's called in
https://github.com/gohugoio/hugo/blob/64fcb35c13829bdc50d0f92628cc60674ed5a193/markup/asciidocext/convert.go#L276-L279

And I had to read it to understand how the ToC works in Hugo. Parameters named `y` and `x` didn't help with that.

I'm happy to revert this change in this PR if you prefer, but I think you should consider adding it in a future commit as it helps with the readability and it's consistent with the variable names used in the only other place it's called:
https://github.com/gohugoio/hugo/blob/64fcb35c13829bdc50d0f92628cc60674ed5a193/markup/goldmark/toc.go#L58

Issue #7667, opened 2020-09-12
Support external source maps for esbuild


richtera:
I was able to get sourcemaps to build out, but the fingerprint pipeline doesn't rename sourcemaps along with the files.
I think OpenResourcePublisher doesn't track the file when it goes through the fingerprint transformation.
But otherwise this should be a pretty good solution. Opening this issue so I can attach a PR.

PR #7668, opened 2020-09-12
js.Build: Add support for SourceMap = external

richtera:
Added support to allow SourceMap files to be external to the build.
In addition added more information when the js compilation has an error.
Correctly append sourceMappingURL to output file.

External sourcemaps for js.Build.
image


richtera:
This refers to #7667

richtera:
This change is also in the iss7667 PR, because I needed some of the changes to make that one work and debug it. Would you like me to rebase this or should we just close this one?

richtera:
Can I close this, because it's also contained in https://github.com/gohugoio/hugo/pull/7670?

richtera:
@bep Since the other commit got closed the external sourcemap support never made it into any hugo builds. I just noticed this because I upgraded from my custom build to the latest. I can rebase and recreate the PR. I currently created a workaround that extracts inline sourcemaps from js files and rewrites all integrity sha's in my HTML files. This seems very cumbersome just to remove sourcemaps. The use case is... Build with sourcemaps and upload them to rollbar or raygun and delete them locally before deployment to CDN. Then errors can be resolved in rollbar or raygun stacktraces.

richtera:
@bep Ok so this is fine. The problem is that the argument got renamed from sourceMaps to sourceMap. I also noticed that the documentation PR seems to be missing and there are no docs for either sourceMaps or sourceMap on the babel page. https://gohugo.io/hugo-pipes/babel/. Here is where I put the other comment https://github.com/gohugoio/hugo/pull/8148.

Issue #7656, opened 2020-09-10
Mount common JS config files into /assets/_jsconfig and add these filenames to the JS environment


bep:
This issue talks about using PostCSS with the TailwindCSS plugin in different situations in Hugo, from the main project, from the theme in `/themes`, from modules, from vendored modules ... But this is also a general issue which is mostly relevant to these JS tools living outside of Hugo.

I created this local "test project" to illustrate the problem for myself:

```bash
├── mod
│   ├── babel.config.js
│   ├── package-lock.json
│   ├── package.json
│   ├── postcss.config.js
│   └── tailwind.config.js
└── project
    ├── out.css
    ├── package-lock.json
    ├── package.json
    ├── run.sh
    └── styles.css
```

So, when running `postcss` from `project` with the configuration living in `mod`, I can do:

```bash
postcss  --config ../mod/postcss.config.js -o out.css styles.css;
```

This is in essence what Hugo does today, and it breaks in, at least, the Tailwind setup above. There are some other "gotchas" to the current situation (one being vendored modules), but I'm not going into details about those.

The main problem with the above is that it fails to find the TailwindCSS configuration. So I could do:

```js
module.exports = {
        plugins: [ require('tailwindcss')('/Users/bep/mod/tailwind.config.js') ]
};
```

Which works, but how do I, running in Hugo, know where to find that file?

I have thought long and hard about this, and jumping to my conclusion, I thought we could do something ala this:

```js
console.warn('In postcss.config.js, root.');

let tailwindConfig = process.env.HUGO_FILE_TAILWIND_CONFIG_JS || './tailwind.config.js';

module.exports = {
	plugins: [ require('tailwindcss')(tailwindConfig) ]
};

````

* We allow mounting of single files in the root of a project/module. For the project I think we can mount the *.config.js files by default (if no other root files are mounted)
* We add these filenames to the OS (and Node) environment mapping them to their absolute filename, normalized on the form `tailwind.config.js` => `HUGO_FILE_TAILWIND_CONFIG_JS`. In the example above, we could be even more specific doing:

```js
let tailwindConfig = process.env.HUGO_MY_TAILWIND_CONFIG_JS || './tailwind.config.js';
```

Thoughts? Any better ideas? 

/cc @regisphilibert @onedrawingperday @digitalcraftsman @jmooring @coliff  and gang.

Also see #7635 and some others ....

richtera:
[...]

PR #7670, opened 2020-09-12
js.Build: Add support to js.Build to automatically generate tsconfig files to search assets and js folders

richtera:
When working with assets folders containing js files, it would be nice to search for js files the same way hugo searches for assets.
So this way when using `import "@js/somefile"` it will automatically search for
./assets/js/somefile
./themes/THEMEMOD1/assets/js/somefile
./themes/THEMEMOD2/assets/js/somefile
and so on.
There was a discussion about resolve inside of #7656 so I attached this PR to this ticket.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,10 +47,10 @@ cache:
 
 before_install:
   - df -h
-    # https://travis-ci.community/t/go-cant-find-gcc-with-go1-11-1-on-windows/293/5
+  # https://travis-ci.community/t/go-cant-find-gcc-with-go1-11-1-on-windows/293/5
   - if [ "$TRAVIS_OS_NAME" = "windows" ]; then
-        choco install mingw -y;
-        export PATH=/c/tools/mingw64/bin:"$PATH";
+    choco install mingw -y;
+    export PATH=/c/tools/mingw64/bin:"$PATH";
     fi
   - gem install asciidoctor
   - type asciidoctor
@@ -66,10 +66,11 @@ script:
   - go mod download
   - go mod verify
   - mage -v test
-  - if [ "$TRAVIS_ARCH" = "amd64" ]; then
-        mage -v check;
+  - >
+    if [ "$TRAVIS_ARCH" = "amd64" ]; then
+      travis_wait 20 mage -v check;
     else
-        HUGO_TIMEOUT=30000 mage -v check;
+      HUGO_TIMEOUT=30000 travis_wait 20 mage -v check;
     fi
   - mage -v hugo
   - ./hugo -s docs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,12 +65,12 @@ install:
 script:
   - go mod download
   - go mod verify
-  - mage -v test
+  - travis_wait 20 mage -v test
   - >
     if [ "$TRAVIS_ARCH" = "amd64" ]; then
-      travis_wait 20 mage -v check;
+      mage -v check;
     else
-      HUGO_TIMEOUT=30000 travis_wait 20 mage -v check;
+      HUGO_TIMEOUT=30000 mage -v check;
     fi
   - mage -v hugo
   - ./hugo -s docs/
--- a/hugolib/js_test.go
+++ b/hugolib/js_test.go
@@ -81,7 +81,9 @@ document.body.textContent = greeter(user);`
   "scripts": {},
 
   "dependencies": {
-    "to-camel-case": "1.0.0"
+		"to-camel-case": "1.0.0",
+		"react": "^16",
+		"react-dom": "^16"
   }
 }
 `
--- a/hugolib/js_test.go
+++ b/hugolib/js_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gohugoio/hugo/htesting"
 
+	"github.com/spf13/afero"
 	"github.com/spf13/viper"
 
 	qt "github.com/frankban/quicktest"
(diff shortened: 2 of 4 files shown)


richtera:
The error seems to be a timeout. Checking.

richtera:
NOTE: There is a problem with the resolver inside of ESBuild, so resolving index.js files is a bit misleading. When a file is called index.js then it will be found in the first folder as long as it exists. So in the following example
```javascript
import "@js/folder"
import "@assets/folder"
```
will always search for the folder in either assets or assets/js but then assume there is an index.js file in the folder (i.e. it will not specifically check for the existence of the index.js file within the folder and will not move on to the next folder in the paths list) Related to evanw/esbuild#380

bep:
I suspect this pseudo-code would work:

```go
afero.Walk(assetsFs, func(path, fi) {
  if fi.IsDir() {
    ifs, err := assetsFs.Stat(filepath.Join(path, "index.js");
    if err == nil {
        // add ifs's real filename to the list
   }
 }
})

**Comment by richtera** [comment]
That's cool. It might be able to simplify some of my code :)

**Comment by richtera** [comment]
Rebase to current master and integrate to deal with new "package.json" in the directory list.

**Comment by richtera** [comment]
This will not work with a readonly mod file system. Refactoring.

**Comment by richtera** [comment]
Created a new version which only writes new tsconfig/jsconfig files at the project root but repoints the config to the original root it found the innermost tsconfig/jsconfig file at.

**Comment by richtera** [comment]
Not sure how to increase test timeouts. I tested the failing ones locally and they did pass.

**Comment by bep** [comment]
They pass on *nix and that is good enough for now. Reading your code, I'm not sure I understand what it does. Or, what I read does not match my view of the problem at hand. So we could probably start by defining what we want to do.

**Comment by richtera** [comment]
Ok can write some additional docs.
Essentially it first searches for the closes tsconfig/jsconfig. Once it finds it it will load it's content and rewrite it by making a new version at the root and rewriting all paths and extends and baseUrl to be relative to the original location. By default it will end up doing this to the root tsconfig/jsconfig. It writes the newly generated version into a temp file and points esbuild to it instead of allowing it to search for one.
If you comment out the Remove command you can see the generated tsconfig/jsconfig file at the project root.
Here is a sample config file it would generate.
```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "*": [
        "/Users/andy/Development/platform-themes/node_modules"
      ],
      "@assets": [
        "./assets/js/index.tsx",
        "./assets/js/index.ts",
        "./assets/js/index.jsx",
        "./assets/js/index.mjs",
        "./assets/js/index.cjs",
        "./assets/js/index.js",
        "./assets/js/index.json",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.tsx",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.ts",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.jsx",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.mjs",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.cjs",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.js",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.json",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.tsx",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.ts",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.jsx",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.mjs",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.cjs",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.js",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.json"
      ],
      "@assets/*": [
        "./assets/*",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/*",
        "./themes/module-student-template-combined@feature-data-builds/assets/*"
      ],
      "@js": [
        "./assets/js/index.tsx",
        "./assets/js/index.ts",
        "./assets/js/index.jsx",
        "./assets/js/index.mjs",
        "./assets/js/index.cjs",
        "./assets/js/index.js",
        "./assets/js/index.json",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.tsx",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.ts",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.jsx",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.mjs",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.cjs",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.js",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/index.json",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.tsx",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.ts",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.jsx",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.mjs",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.cjs",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.js",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/index.json"
      ],
      "@js/*": [
        "./assets/js/*",
        "./themes/module-student-template-test-prep@feature-data-builds/assets/js/*",
        "./themes/module-student-template-combined@feature-data-builds/assets/js/*"
      ]
    }
  }
}
```
My theme module names have an "@" in their directory names so that's just a side effect of something else.

bep:
OK, I think this discussion will be easier if we ignore any existing ts/jsconfig. I had not even heard of them until recently, so I think we can safely push that problem ahead.

So,

* We always write a brand new jsconfig to a a temp folder representing the root.
* We walk /assets and searches for index.js/ts files and add mappings to those

richtera:
Hmmm, that wouldn't work in our configuration, but maybe there might be a way to allow "configuration" of additional aliases in another way.

bep:
OK, then I think I need to test this out myself.

richtera:
I was thinking I could write a test scripts for this. Do you have a good place in hugo to get such a testscript sample from. I was thinking the base fs testscript I previously fixed. But if there is a better place that would be great. I could create various js and jsconfig files and compile them in the test script.

richtera:
@bep I am trying to write a testscript for this. I can build normal js files, but when trying to attach multiple theme modules I am not quite sure how to do it. I inserted theme=[mod1,mod2] into the config Viper but it's not searching the folders for some reason.
There might be some kind of trick I am missing:
```golang
func TestJSBuildOverride(t *testing.T) {
	// if !isCI() {
	// 	t.Skip("skip (relative) long running modules test when running locally")
	// }
	
	wd, _ := os.Getwd()
	defer func() {
		os.Chdir(wd)
	}()

	c := qt.New(t)

	workDir, clean, err := htesting.CreateTempDir(hugofs.Os, "hugo-test-js2")
	c.Assert(err, qt.IsNil)
	defer clean()

	v := viper.New()
	v.Set("workingDir", workDir)
	v.Set("themeDir", "themes")
	v.Set("disableKinds", []string{"taxonomy", "term", "page"})
	v.Set("theme", []string{"mod1", "mod2"})
	b := newTestSitesBuilder(t).WithLogger(loggers.NewWarningLogger())

	b.Fs = hugofs.NewDefault(v)
	b.WithWorkingDir(workDir)

	jsDir := filepath.Join(workDir, "assets", "js")
	b.Assert(os.MkdirAll(jsDir, 0777), qt.IsNil)
	b.Assert(os.Chdir(workDir), qt.IsNil)
	jsDir = filepath.Join(workDir, "themes", "mod1", "assets", "js")
	b.Assert(os.MkdirAll(jsDir, 0777), qt.IsNil)
	b.Assert(os.Chdir(workDir), qt.IsNil)
	jsDir = filepath.Join(workDir, "themes", "mod1", "assets", "js")
	b.Assert(os.MkdirAll(jsDir, 0777), qt.IsNil)
	b.Assert(os.Chdir(workDir), qt.IsNil)
	jsDir = filepath.Join(workDir, "themes", "mod2", "assets", "js")
	b.Assert(os.MkdirAll(jsDir, 0777), qt.IsNil)
	b.Assert(os.Chdir(workDir), qt.IsNil)
	jsDir = filepath.Join(workDir, "themes", "mod2", "assets", "js")
	b.Assert(os.MkdirAll(jsDir, 0777), qt.IsNil)
	b.Assert(os.Chdir(workDir), qt.IsNil)

	b.WithContent("p1.md", "")

	b.WithTemplates("index.html", `
{{- $js := resources.Get "js/main-project.js" | js.Build -}}
{{ template "print" (dict "js" $js "name" "root") }}

{{- $js = resources.Get "js/main-mod1.js" | js.Build -}}
{{ template "print" (dict "js" $js "name" "mod1") }}

{{- $js = resources.Get "js/main-mod2.js" | js.Build -}}
{{ template "print" (dict "js" $js "name" "mod2") }}

{{- define "print" -}}
{{ printf "rellink-%s-%s" .name .js.RelPermalink | safeHTML }}
{{ printf "mime-%s-%s" .name .js.MediaType | safeHTML }}
{{ printf "content-%s-%s" .name .js.Content | safeHTML }}
{{- end -}}
`)
	
// Override project included file
	b.WithSourceFile("assets/js/override.js", `
const name = "root-override";
export default name;
`)
	b.WithSourceFile("themes/mod1/assets/js/main-mod1.js", `
import override from "@js/override";
import overrideMod from "@js/override-mod";
window.override = override; // make sure to prevent tree-shake
window.overrideMod  = overrideMod; // make sure to prevent tree-shake
`)
	b.WithSourceFile("themes/mod2/assets/js/main-mod1.js", `
import override from "@js/override";
import overrideMod from "@js/override-mod";
window.override = override; // make sure to prevent tree-shake
window.overrideMod  = overrideMod; // make sure to prevent tree-shake
`)
	b.WithSourceFile("themes/mod2/assets/js/main-mod2.js", `
import override from "@js/override";
import overrideMod from "@js/override-mod";
import config from "@config";
window.override = override; // make sure to prevent tree-shake
window.overrideMod  = overrideMod; // make sure to prevent tree-shake
window.config = config;
`)
	b.WithSourceFile("themes/mod2/assets/js/jsconfig.json", `
{
	"compilerOptions": {
		"baseUrl": ".",
		"paths": {
			"@config": ["./config.json"]
		}
	}
}
`)
	b.WithSourceFile("themes/mod2/assets/js/config.json", `
{
	"data": {
		"sample": "sample"
	}
}
`)
	b.WithSourceFile("themes/mod1/assets/js/override.js", `
const name = "mod1-override";
export default name;
`)
	b.WithSourceFile("themes/mod2/assets/js/override.js", `
const name = "mod2-override";
export default name;
`)
	b.WithSourceFile("themes/mod1/assets/js/override-mod.js", `
const nameMod = "mod1-override";
export default nameMod;
`)
	b.WithSourceFile("themes/mod2/assets/js/override-mod.js", `
const nameMod = "mod2-override";
export default nameMod;
`)
	b.theme = "mod1"
	b.WithThemeConfigFile("yaml", `
assetsDir: assets`)
	b.theme = "mod2"
	b.WithThemeConfigFile("yaml", `
assetsDir: assets`)
	b.WithViper(v)

	b.Build(BuildCfg{})

	content := b.FileContent("public/index.html")
	for _, line := range strings.Split(content, "\n") {
		fmt.Println(line)
	}

	b.AssertFileContent("public/index.html", `
console.log(&#34;included&#34;);
`)
}
```
I this the right approach for this?

richtera:
@bep  Figured out how to get the test scripts to work. Will commit in a few minutes.

richtera:
oops. My commit message contains `@data` and github is trying to send it as a notification. Is that a big deal?

richtera:
@bep tried to push a commit to extend travis wait for mage -v test but by mistake pushed a version for mage -v check. Not sure how to cancel a build. Another question... the windows build is still failing due to some aliases. Something to look at?

richtera:
Rebased to current master.

bep:
* Does any of this require writing files into the module/theme folders?
* Also, I see mention of node_modules in there, which also, I assume, is related to my first question. A solution that requires us to write to the source folders is not something we can use.

richtera:
No I am not writing files in either locations but I read from there. The only modifications happen at the project root (i.e workingDir)

richtera:
Actually I use a temporary file there, which could even be in the /tmp. So really no writes visible to hugo at all.

richtera:
Figured out the windows test script failure. Will commit in a little. We could use the /tmp file for the temporary files, it will cause a bunch of extra ./../../ in front of the resolved files, but I don't think that's a problem with esbuild. Testing that as well.

richtera:
I was also able to change the tsconfig/jsconfig file to be only created inside of /tmp. It just added a little complexity to the relative calculations, but I needed to adjust those for the windows case already.

richtera:
One file is getting caught in the wrong directory. Fixing.

richtera:
Ok tests are passing now. The windows version had problems with \ because the jsconfig file uses URLs and not file paths but I used file paths.Rel to calculate the relative path.

richtera:
@bep How should this integrate with the new hugo mod npm? The imports should work as is, but there maybe additional hints I could read from the generated config files. It may also make sense to simplify this feature and just to use "assets/" and "data/" as prefixes rather than "@js" and "@data", but that's kind of subjective. Do you have a preference? It's really cool to have such a great a testing harness inside of hugo to be able to actually test all this stuff out, so the test script actually validates imports, compilation and sourcemaps. Sorry that it took a few tries to get this to work on Mac, Windows and Linux. Thanks for pointing out that hugo modules are "readonly" during the hugo compile, I hadn't realized that initially, since theme modules is where we do most of our development for our platform. Using this jsconfig/tsconfig setup allows us to implement only necessary js files on child themes to override functionality on inherited theme modules.

bep:
>I was also able to change the tsconfig/jsconfig file to be only created inside of /tmp. It just added a little complexity to the relative calculations, but I needed to adjust those for the windows case already.

I expect most people would want that. I know that VS Code is using these files to help with code navigation, so it would be cool to test that out.

As to this PR, I will eventually/soonish get to this. My Hugo efforts tends to happen in bursts between a fair amount of "real work" these days.

bep:
I will look at this more closely during the weekend; sorry for the delay, but you know, work and life...

richtera:
It's fine. Same here. Let me know if you want tighter integration with your Hugo mod npm. I would also like to talk about building sites with different module setups per page bundle which is linked to my requirement to use /tmp for the jsconfig file that might have an alternate solution that might fit better in your plans. What's a good channel to talk?

bep:
Considering the recent additions, I'll close this, but thanks a lot for your work on this.

PR #7672, opened 2020-09-13
postcss: Mockup simple test to setup NODE_PATH and PATH

richtera:
I was playing around with jsBuild and took some of the path iteration stuff
to allow testing this theory.

This is to play with #7656

### DIFF
--- a/resources/resource_transformers/postcss/postcss.go
+++ b/resources/resource_transformers/postcss/postcss.go
@@ -35,6 +35,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cast"
 
+	"github.com/gohugoio/hugo/config"
 	"github.com/gohugoio/hugo/hugofs"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


resources/resource_transformers/postcss/postcss.go
```diff
@@ -140,69 +143,88 @@ func (t *postcssTransformation) Key() internal.ResourceTransformationKey {
...
-		if err != nil {
-			if t.options.Config != "" {
-				// Only fail if the user specificed config file is not found.
-				return errors.Wrapf(err, "postcss config %q not found:", configFile)
+	nodePath := ""
+	binPath := ""
+	for idx, mod := range t.rs.PathSpec.Paths.AllModules {
+		dir := mod.Dir()
+		if localCommand == "" {
+			commandPath := path.Join(dir, "node_modules", ".bin", binaryName)
```
bep, line 162: So, the idea is good, but this assumes that `npm install` has been run -- which Hugo cannot do as a general thing (modules pulled in via Hugo Modules are read-only)

richtera: Interesting. That would assume that npm folders need to either be published as part of the module, or updated in some other way. It does make sense, but that will make a lot of things more difficult if modules' js files require things which are not necessarily available on the project node_modules and package.json. That's in interesting dilemma :)

bep:
I'm closing this.

richtera:
Ok, thanks

Issue #7618, opened 2020-09-02
Add a feature to index function to allow keys being a slice


richtera:
When calling `(index data "key" "subkey" "subsubkey")` it's hard to dynamically access the data. It would help a lot if index supported either `(index data "key.subkey.subsubkey")` or `(index data (slice "key" "subkey" "subsubkey"))` as well.
Otherwise to retrieve information by specifying the "index from a variable" is not easy.
Our current workaround to the problem is a partial. This reads the location in the data specified by .Params.data which is a slice assigned within our content.
```hugo
{{ $scratch := newScratch }}
{{ if .Params.data }}
  {{ $scratch.Set "data" $.Site.Data }}
  {{ range .Params.data }}
    {{ if $scratch.Values.data }}
      {{ $scratch.Set "data" (index $scratch.Values.data .) }}
    {{ end }}
  {{ end }}
{{ end }}
{{ return $scratch.Values.data }}
```

bep:
Arrest me if I'm wrong, but I think that the last argument of `index` is a vararg, so this should work:

```bash
index $data key subkey subsubkey
```

I understand that that does not solve your question, but we have had many similar questions about the other vararg functions.

I'm not sure how easy this is, but it would be nice if we could do something ala:

```html
{{ index $data (expand $myslice) }}
```

/cc @moorereason

richtera:
That would also be a great solution and more general to call any function with args from a slice. Does a pipeline fill varargs?

bep:
>Does a pipeline fill varargs?

Yes. But note that I'm not totally sure how to solve this without doing any changes to the "Go stdlib runtime".

richtera:
Right, kind of like `{{ index $data $myslice... }}` that would be cool.

bep:
>{{ index $data $myslice... }}

Well, no, that may be a stretch.

moorereason:
I wouldn't recommend `{{ index $data "1.2.3" }}` since that could be a valid index/key.  Assuming that `index` could be extended to support a slice of indices, you could accomplish the same thing with `{{ index $data (split "1.2.3" ".") }}`.

richtera:
Ok, it might be easier to implement a new function `{{ indexIn $data $keySlice }}` for example. But I think both would be a great solution to the weird `{{ range ... }}` hack.

PR #7673, opened 2020-09-13
tpl: Add indexIn function into templates

richtera:
The indexIn function is similar to index, but uses a slice
as the last arg to supply the path within the document.
Ref #7618


bep:
Let's come to some kind of conclusion in #7618 before we start with the implementations.

richtera:
I called it indexIn so there is no name conflict and no change to the index function. But we can revisit it.

PR #4686, opened 2018-04-30
Added 'async' attribute to the Github gist embedded shortcode (remove…

lroguet:
… render-blocking Javascript).

PR #7678, opened 2020-09-13
shortcode: gist JS loads asynchronously

bcollard:
effectively ships the suggested modification from PR https://github.com/gohugoio/hugo/pull/4686


bcollard:
actually doesn't work, sorry.
`A call to document.write() from an asynchronously-loaded external script was ignored.`

Issue #7616, opened 2020-09-02
Problem mounting non-rooted mount into data root


richtera:
compile on master

<pre>
$ hugo version
Hugo Static Site Generator v0.75.0-DEV-ED754F5A/extended darwin/amd64 BuildDate: 2020-08-31T07:15:27-0400
</pre>

### Does this issue reproduce with the latest release?
Yes.

Using a mount with

`target: data/somefolder`

mounts the source at the root of the data folder.

richtera:
Sample output of hugo config mounts
```zsh
% hugo --config ./.config.yml config mounts
{
   "path": "project",
   "dir": "/Users/andy/Development/platform-themes",
   "mounts": [
      {
         "source": "content",
         "target": "content"
      },
      {
         "source": "data",
         "target": "data"
      },
      {
         "source": "repos/builder-course-PlvZSzcF2VNeXanE8IWQ/content",
         "target": "content/courses/PlvZSzcF2VNeXanE8IWQ"
      },
      {
         "source": "repos/builder-course-PlvZSzcF2VNeXanE8IWQ/data",
         "target": "data/PlvZSzcF2VNeXanE8IWQ"
      },
      {
         "source": "layouts",
         "target": "layouts"
      },
      {
         "source": "i18n",
         "target": "i18n"
      },
      {
         "source": "archetypes",
         "target": "archetypes"
      },
      {
         "source": "assets",
         "target": "assets"
      },
      {
         "source": "static",
         "target": "static"
      }
   ]
}
{
   "path": "module-student-template-combined@feature-data-builds",
   "dir": "/Users/andy/Development/platform-themes/themes/module-student-template-combined@feature-data-builds/",
   "mounts": [
      {
         "source": "archetypes",
         "target": "archetypes"
      },
      {
         "source": "assets",
         "target": "assets"
      },
      {
         "source": "data",
         "target": "data"
      },
      {
         "source": "layouts",
         "target": "layouts"
      },
      {
         "source": "static",
         "target": "static"
      }
   ]
}
```
Sample config file
```yaml
baseURL: /
languageCode: en-us
title: SelfStudy Learning Site
disablePathToLower: true
params:
  type: shortcodes
  date_format: Monday, Jan 2, 2006
  copyright: SelfStudy Inc. All rights reserved.
  disable_mathjax: true
  features:
    lock: false
    studybar: false
    theme: light
blackfriday:
  plainIDAnchors: true
  hrefTargetBlank: true
  fractions: false
[...]

PR #7679, opened 2020-09-13
mounts: Allow mounting of data into subfolders

richtera:
This allows mounts for data to be in target folders not at the root
For example
```toml
[module]
[[module.mounts]]
source = "data2"
target = "data/test"
[[module.mounts]]
source = "data"
target = "data"
```
will correctly mount the data2 folder to be available as .Site.Data.test
Ref #7616


richtera:
Removed unnecessary changes which were creating a failure in the TestNewBaseFs test script.

SirUli:
This PR is still relevant.

PR #7689, opened 2020-09-14
Documents empty attributes and attribute unsetting

bric3:
Asciidoctor allows to define empty valued attributes and
allows to _unset attributes_, or even to _soft unset attributes_.
For people unfamiliar to TOML, those example can be interesting.

https://asciidoctor.org/docs/user-manual/#using-attributes-set-assign-and-reference


bep:

PR #7756, opened 2020-10-01
tpl: Updated Google Analytics internal template to use GTM (gtag.js)

### LINKED ISSUE #4327: Update Google Analytics internal template to new style(gtag.js)
Hello I found Google Analytics tracing code has updated from old analytics.js to gtag.js .

Some refers:
- https://support.google.com/analytics/answer/1008080
- https://developers.google.com/analytics/devguides/collection/gtagjs/migration

Look like they are the same for basic usage. I think Hugo's Google Analytics internal template can safely update to new style.

### LINKED ISSUE #4479: Update the internal GA template
https://github.com/gohugoio/hugoDocs/blob/master/themes/gohugoioTheme/layouts/partials/gtag.html#L9

Not sure what to do with the custom dimensions above.  We could add them to site config somehow? We can probably make the `GoogleAnalytics` optionally be a map.

Also see https://discourse.gohugo.io/t/track-outbound-links-with-google-analytics/6289

joshgerdes:
Updated the internal templates for Google Analytics to use the new Google Tag Manager (gtag.js).
Added basic page data to be sent with the new Google Analytics call.
Added additional privacy flag to disable sending enhanced page data to google.

Fixes #4327, Fixes #4479


bep:
This looks good to me, @coliff  could you take a look?

coliff:
Hey, thanks for the PR. The Gtag implementation of Google Analytics isn't strictly replacement for `analytics.js` - rather it's a wrapper for it. If you include the Gtag implementation then it still actually loads analytics.js - in addition to gtag.js meaning extra HTTP requests and more data consumed. I did a lot of research and testing into this for the [HTML5Boilerplate project](https://github.com/h5bp/html5-boilerplate/blob/master/src/doc/html.md#google-universal-analytics-tracking-code).

There are still times when using gtag takes more sense (if you're using gtag manager to manage your tracking - or using additional Google services such as Optimize, Google Ads etc)... so overall it's good to give the user the option for which implementation to use.

For all sites I build and manage I use the following snippet for optimal performance:
```

coliff:
There are some useful options I think we could add to Hugo via params such as:

Disable all advertising features & Disable advertising personalization
https://developers.google.com/analytics/devguides/collection/gtagjs/display-features

But as it is right now, I'd suggest not to merge this PR. Analytics.js is still available and supported by Google and should be available to users. Even if analytics.js is deprecated in the future, I think it'd be better to mark the template as deprecated and not 'auto-upgrade' users to gtag as it's possible there are breaking changes. gtag.js doesn't support some things which analytics.js does.

For the DisableEnchancedData attribute - do you have a link for documentation on that? I can't see it mentioned on Google's docs. https://developers.google.com/analytics/devguides/collection/gtagjs

joshgerdes:
👋 @coliff, thanks for the review and feedback. I understood gtag to be a wrapper and thought it was a full replacement. I didn't realize it was not. Google also got me with some choice marketing 😄 :

> By updating your current Google Analytics tag (analytics.js) to the latest tag (gtag.js), you will benefit from: 
> 
> - Future-proof conversion measurement: Benefit from the latest features and integrations as they become available so that you'll be better equipped to continue measuring conversions accurately if non-Google changes occur.
> - Code-less tag management: Turn measurement features on and off within your Google Analytics account without having to manually change tags on your page.
> - Faster and easier integration: Seamlessly implement and integrate with other Google products, such as Google Ads and Google Optimize.



Also, for the the `DisableEnchancedData` attribute, it was just meant as way for the user to limit the amount of extra site related data sent but I just made it up and use it to limit the extra 'gtag' site information that is being sent.

I did add the [user opt out](https://developers.google.com/analytics/devguides/collection/gtagjs/user-opt-out) feature which seems like a nice addition but the other advertising features seem like a great add as well.

Moving forward do you think it makes sense for me to move the work in this PR into a new `google_analytics_gtag.html` template? This would allow people to have a choice between the 3 options? 

And for those additional attributes (disable all advertising features & Disable advertising personalization) I could incorporate them to this PR (adding flags to privacy) or create a new PR to keep things simple.

Happy to help, just let me know your thoughts.

coliff:
hey again @joshgerdes - awesome, yes, I suggest you make all those changes to a new `google_analytics_gtag.html` template and leave the older `google_analytics` and `google-analytics-async` unchanged for now.
I approve of the idea to add those additional attributes (disable all advertising features & Disable advertising personalization) within the same PR. 👍

joshgerdes:
Shiny new and ready for review #7780

awhitford:
What version will this new `google_analytics_gtag` template be available?

joshgerdes:
@awhitford when #7780 gets merged.

sepehr-alipour:
How we can use both UA and GTM params as old google analytics version and GA4?

PR #7758, opened 2020-10-01
fix: Add ability for plainify to remove <style> and <script>

### LINKED ISSUE #3235: plainify should remove <style> and <script> sections

joshgerdes:
Added StripTag helper function to allow the ability to remove any tag and all
of its contents from a string.  Leveraged that new function within the plainify
function to remove <style> and <script> tags and contents.

Fixes #3235

### DIFF
--- a/helpers/content.go
+++ b/helpers/content.go
@@ -119,6 +119,38 @@ func StripHTML(s string) string {
 	}
 	s = stripHTMLReplacer.Replace(s)
 
+	for _, value := range []string{"script", "style"} {
+		startTag := "<" + value
+		endTag := "</" + value + ">"
+
+		start := strings.Index(s, startTag)
+		if start >= 0 {
+
+			result := make([]byte, start, len(s))
+			copy(result, s)
+
+			for {
+				end := strings.Index(s[start+len(startTag):], endTag)
+				if end < 0 {
+					result = append(result, s[start:]...)
+					break
+				}
+				end += (start + len(startTag)) + len(endTag)
+
+				start = strings.Index(s[end:], startTag)
+				if start < 0 {
+					result = append(result, s[end:]...)
+					break
+				}
+				start += end
+
+				result = append(result, s[end:start]...)
+			}
+
+			s = string(result)
+		}
+	}
+
 	// Walk through the string removing all tags
 	b := bp.GetBuffer()
 	defer bp.PutBuffer(b)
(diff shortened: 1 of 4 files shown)


bep:
Hey, and thanks for this. The code looks clean and good, but I have one general comment:

* This issue is, in my head, relevant to all the callers of `StripHTML`, so this new logic might as well be included in that function.
* Also, the way `Plainify` works with this commit, you will get a fair amount of wasteful memory allocation, but moving this into `StripHTML` should give you some benchmarks to compare to.

joshgerdes:
Heya 👋 , thanks for the feedback.

I agree it probably makes sense that someone using `StripHTML` would want those tag sections removed as well.  So were you thinking of just moving something like this into `StripHTML` function after the `stripHTMLReplacer` call:

```go
for _, value := range []string{"script", "style"} {
	stripTag(s, value)
}
```

Or moving the whole logic in and eliminating the `stripTag` function?

Or trying to integrate the logic of removing content with the tags in the current switch statement of `StripHTML`?

bep:
>Or moving the whole logic in and eliminating the stripTag function?

See my comment about memory allocation. `StripHTML` should have a benchmark. If you run that before/after and compare with `benchstat` (also see my util project `gobench`), I suspect that with the `stripTag` function you will see a major hit in memory allocation. This will cost _something_, of course, but since we're running this on thousands of content pages, this matters.

joshgerdes:
Thanks for the clarification and sorry I'm new to Go so still getting a handle on tooling, how allocation works, etc. You were correct though, here are the benchmark results I got comparing the two options.

With `stripTag` referenced in `StripHTML`:
```
name         old time/op    new time/op    delta
StripHTML-8    3.45µs ± 1%    4.84µs ± 1%   +40.17%  (p=0.008 n=5+5)

name         old alloc/op   new alloc/op   delta
StripHTML-8      896B ± 0%     2304B ± 0%  +157.14%  (p=0.008 n=5+5)

name         old allocs/op  new allocs/op  delta
StripHTML-8      4.00 ± 0%      8.00 ± 0%  +100.00%  (p=0.008 n=5+5)
```
With logic in `StripHTML`:
```
name         old time/op    new time/op    delta
StripHTML-8    3.45µs ± 1%    4.04µs ± 1%   +17.03%  (p=0.008 n=5+5)

name         old alloc/op   new alloc/op   delta
StripHTML-8      896B ± 0%     2080B ± 0%  +132.14%  (p=0.008 n=5+5)

name         old allocs/op  new allocs/op  delta
StripHTML-8      4.00 ± 0%      8.00 ± 0%  +100.00%  (p=0.008 n=5+5)
```

I have updated the PR, let me know if the latest is more what you were thinking.

joshgerdes:
@bep is this what you were thinking?

joshgerdes:
<bump>

@bep thoughts?

D4D3VD4V3:
I'd love to see this pushed upstream so I too can get more accurate estimates of words in a post.

joshgerdes:
@bep is this going to be used?  If so, any changes needed?  Or should I just close it?

bep:
No, keep it open, it will be used, it fixes an important issue. I just need to understand the performance tradeoffs. I will look at this later.

hacdias:
Any updates on this?

joshgerdes:
The stalebot cometh...

I will keep open per your request @bep

Issue #7759, opened 2020-10-01
Math in reStructuredText isn't formatted with MathJax


sumnerevans:
By default, `rst2html` exports math using the "HTML math.css" formatter. This means that any math typeset in reStructuredText is not formatted using MathJax which is not ideal.

`rst2html` has a `--math-output` option and you can pass `MathJax` to make it output as MathJax.

I propose that this option is added to the list of options passed to `rst2html` so that math can be formatted nicely.

PR #5350, opened 2018-10-24
Use short token for code highlighting

stephenfin:
Closes: #5349

As noted in #5349, `rst2html` can generate HTML with short or long form class names for syntax highlighting. Pygments (via the `pygmentize` command) seems to generate short form by default without any way to switch this. This means most/any stylesheets for code provided with various themes won't work when using rST output. This can easily be fixed by providing the `--syntax-highlight=short` flag to `rst2html`.

Issue #5349, opened 2018-10-24
reStructuredText uses long token names for syntax highlighting


stephenfin:
`rst2html` can generate HTML with short or long form class names for syntax highlighting. Pygments (via the `pygmentize` command) seems to generate short form by default without any way to switch this. This means most/any stylesheets for code provided with various themes won't work when using rST output. This can easily be fixed by providing the `--syntax-highlight=short` flag to `rst2html`.

stephenfin:
This is a bug. Without this fix, we do not have syntax highlighting when using rST resources.

cup:
https://github.com/gohugoio/hugo/pull/5350

PR #7763, opened 2020-10-02
Add --math-output=MathJax for rst2html

sumnerevans:
Closes: #5349
Closes: #7759 

Supersedes and Closes: #5350 

There are two changes in this PR:

* As noted in #5349, rst2html can generate HTML with short or long form class names for syntax highlighting. Pygments (via the pygmentize command) seems to generate short form by default without any way to switch this. This means most/any stylesheets for code provided with various themes won't work when using rST output. This can easily be fixed by providing the `--syntax-highlight=short` flag to rst2html.

* As noted in #7759, rst2html exports math using the "HTML math.css" formatter by default. This means that any math typeset in reStructuredText is not formatted using MathJax which is not ideal. This commit adds a `--math-output` parameter configure rst2html to output math ready for MathJax to process. Note that I had to pick a URL for MathJax, but it doesn't end up in the resulting output because the `<script>` tag is in the header which gets cut out by code further down.

I have added tests to verify that each of the above fixes are working. I also had to add a step to install pygments to make the test work.


sumnerevans:
I think this is finally ready for review. I've gone through the `CONTRIBUTING` document and I believe that I've followed all of the guidelines. I also added tests to verify the functionality is correct. Please let me know what else I need to do to get this merged.

PR #7768, opened 2020-10-03
Update robots.txt internal template

davidsneighbour:
It's actually wrong syntax to just have a catch-all user agent without allow or disallow rule. Google proposes to have an `Allow` to allow access to everything. While this is (as of now) working it's false and might stop working at some point. So let's have it properly done :)


bep:
The file you edited is autogenerated. The source lives in a txt file in the templates folder. When you edit one of those you need to run `mage generate`.

davidsneighbour:
updated

PR #7770, opened 2020-10-04
Create install.sh

DH4CK1:
Im create this file for install requirements


davidsneighbour:
Spam. Hugo does not use Python.

PR #7780, opened 2020-10-05
tpl: Added Google Analytics GTM (gtag.js) internal template

### LINKED ISSUE #4327: Update Google Analytics internal template to new style(gtag.js)

### LINKED ISSUE #4479: Update the internal GA template

joshgerdes:
Added an internal template that uses the new Google Tag Manager (gtag.js).
Added basic page data to be sent with the new Google Analytics call.
Added additional privacy flags to disable  google ad signals, good ad personalization signals, and sending enhanced page data to google.

Fixes #4327, Fixes #4479

### DIFF
--- a/config/privacy/privacyConfig.go
+++ b/config/privacy/privacyConfig.go
@@ -53,6 +53,15 @@ type GoogleAnalytics struct {
 
 	// Enabling this will make it so the users' IP addresses are anonymized within Google Analytics.
 	AnonymizeIP bool
+
+	// Enabling this will disable all advertising, reporting, and remarketing features, and override any property settings established in the Google Analytics user interface (gtag only).
+	DisableGtagAdGoogleSignals bool
+
+	// Enabling this will disable Google advertising personalization features (gtag only).
+	DisableGtagAdPersonalizationSignals bool
+
+	// Enabling this will make it so extra page information is not sent to Google Analytics (gtag only).
+	DisableGtagEnhancedData bool
 }
 
 // Instagram holds the privacy configuration settings related to the Instagram shortcode.
--- a/config/privacy/privacyConfig_test.go
+++ b/config/privacy/privacyConfig_test.go
@@ -36,6 +36,9 @@ disable = true
 respectDoNotTrack = true
 anonymizeIP = true
 useSessionStorage = true
+disableGtagAdGoogleSignals = true
+disableGtagAdPersonalizationSignals = true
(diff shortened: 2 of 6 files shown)


coliff:
Great PR. I haven't tested it, but at a glance it looks good. 👍

davidsneighbour:
This is just a personal preference, but the whole "signal" term part in the configuration department could be left out, as it's understood by the parts left of `_signal` what they mean. At least for the hugo config part.

AntoineSoetewey:
Hello,

Is this what we need in order to use the new Google Analytics 4 properties (https://support.google.com/analytics/answer/10089681?hl=en&utm_id=ad)?

Regards,
Antoine

joshgerdes:
@coliff @davidsneighbour do you feel this is going to be used? If so, any changes needed? Or should I just close it?

davidsneighbour:
I was under the impression that this was already merged. There was recently a patch for the analytics layout to work with the new shorter number format for GA4, which is only working with gtag.js, so this PR might be obsolete now. On the other hand, those privacy options are definitely not in yet and they should.

joshgerdes:
Thanks for the reply @davidsneighbour. I might be missing it, but the only gtag reference I see in code is in a comment here:
https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/shortcodes.go#L28-L31

And there is one other related gtag PR with a `keep` label but it doesn't seem as robust as this one: https://github.com/gohugoio/hugo/pull/4208

That's great to hear that we're getting some sort of gtag update in though.

joshgerdes:
Looks like https://github.com/gohugoio/hugo/pull/8287 was merged so should have some sort of GA4 support soon.

djatwood:
With the merge of #8287 is this PR still necessary?

Issue #7221, opened 2020-04-28
ignoreFiles and layouts folder


nu8:
I currently use this to ignore everything except Markdown:

~~~yaml
ignoreFiles:
- \.[^m]
- \.m[^d]
- \.md.
~~~

However I noticed that it ignores **everything**, even files in **layouts**.
This seems to contrast with the documentation:

> The following statement inside `./config.toml` will cause Hugo to ignore
> **content** files ending with `.foo` and `.boo` when rendering:
>
> `ignoreFiles = [ "\\.foo$", "\\.boo$" ]`

<https://gohugo.io/getting-started/configuration#ignore-content-files-when-rendering>

muv1:
Workaround:

~~~yaml
ignoreFiles:
- content/.*\.[^m]
- content/.*\.m[^d]
- content/.*\.md.
~~~

zor-el:
In addition, I suggest to improve the documentation to make it clear

1. what the regular expressions apply to: Only the file name? The full path? The relative path? (And relative to what?)
1. how they are applied (weather they can match anywhere or from the beginning of a string)

Because for ex. the pattern to ignore a content subfolder like [described here](https://discourse.gohugo.io/t/ignore-a-directory/8880/2) doesn't work any more, and it is unclear why without the information listed above.

muv1:
> I second this. In particular, I can't figure out why `ignoreFiles:
> ["\.suffix$"]` ignores `content/post/foo/file.suffix` but not
> `content/post/foo/bar/file.suffix`. I guess there could be an interference
> with page bundle handling, but that's unclear.

If you have a file like:

~~~
content/sunday/monday/a.go
~~~

You need this:

~~~
content/sunday/_index.md
~~~

Else the file will remain.

> **Update:** To make things even more confusing, `ignoreFiles: ["/bar$"]`
> ignores the entire `content/post/foo/bar` directory (as well as any other
> directory named `bar` under content), but `ignoreFiles: ["/bar/"]` doesn't.
> confused

If I have something like `content/sunday/index.md`, then `/sunday/` does ignore
it, not sure what you are trying to do.

>     1. what the regular expressions apply to: Only the file name? The full
>     path? The relative path? (And relative to what?)

The input string is the entire path. So for Windows, that is something like

~~~
C:\sunday\content\monday\index.md
~~~

>     2. how they are applied (weather they can match anywhere or from the
>     beginning of a string)

You can match from beginning or end or both, as long as you respect that the
[...]

PR #7813, opened 2020-10-10
configuration: watchFiles

### LINKED ISSUE #3818: Request - Have a way to specify additional files to watch - e.g. 'watchFiles'
I suppose this would be the opposite of the existing `ignoreFiles` setting. Therefore it should probably be implemented as a regexp just like ignoreFiles was too.

**Reason for this request:**

I am using asciidoc, and have created a site-wide `config.adoc` configuration file. Which that all of my `.adoc` files include. And is in the page template. This is because asciidoc really needs its own configuration file. And the perfect place to put it would be allongside `config.toml` in the root folder. I would very much like for hugo to watch and reload the site whenever this global asciidoc settings file is updated.

Now I've tried putting my file in content, and static. However it seems most of these default watch folders are not really appropriate for this file as usually Hugo expecting something else in those folders. Causing various little issues depending upon which subfolder its in. So for the time being, I have shoved it in `layouts/`, which seems to be working OK for now. In as far far as I don't know anything else is being broken by that. So hopefully I can keep doing this until this requested feature might be implemented in the future.

Please consider. Kind regards.

nu8:
This allows for a `watchFiles` configuration option, which is similar to
`ignoreFiles`, but will ignore files that do not match. Example:

    watchFiles = [ "\\.md$", "\\.jpg$" ]

Behavior is as follows:

1. If only `ignoreFiles` is set, then just use normal `ignoreFiles` behavior.

2. If just `watchFiles` is set, then any files not matching are ignored. Note
   that `watchFiles` will not match a directory.

3. If both `ignoreFiles` and `watchFiles` are set, then only items which pass
   both tests will be included in the output.

Fixes #3818

### DIFF
--- a/source/sourceSpec.go
+++ b/source/sourceSpec.go
@@ -35,6 +35,7 @@ type SourceSpec struct {
 
 	// This is set if the ignoreFiles config is set.
 	ignoreFilesRe []*regexp.Regexp
+	watchFilesRe []*regexp.Regexp
 
 	Languages              map[string]interface{}
 	DefaultContentLanguage string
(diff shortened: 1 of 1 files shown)


nu8:
https://github.com/gohugoio/hugo/issues/7221

nu8:
https://github.com/gohugoio/hugo/issues/3811

PR #7814, opened 2020-10-10
update to golang 1.15 and alpine 3.12

ducksecops:
Golang 1.13 used for build is now EOL, updated to latest 1.15. Updated main image to alpine 3.12 to match alpine used in build.


bep:
Thanks for this. Can you press the CLA button above?

Issue #7551, opened 2020-08-12
Hugo extended BuildDate shows as unknown on Windows 10


jflipp:
Here's a minor bug.

I use hugo v0.74.3 and hugo extended v0.74.3 on Windows 10 64 Bit.

When I execute the version subcommand, I get the following output.

<pre>
C:\Test> C:\Test\hugo_0.74.3_Windows-64bit\hugo.exe version
Hugo Static Site Generator v0.74.3-DA0437B4 windows/amd64 BuildDate: 2020-07-23T16:23:30Z

C:\Test> "C:\Test\hugo_extended_0.74.3_Windows-64bit\hugo.exe version
Hugo Static Site Generator v0.74.3/extended windows/amd64 BuildDate: unknown
</pre>

We see that the regular version of Hugo shows a reasonable BuildDate, but the extended version of Hugo shows **BuildDate: unknown**.

Maybe you want to look into this.

As far as I know, v0.74.3 is currently the latest release of Hugo.

Thank you.

jmooring:
Unable to reproduce with v0.83.1.

```
hugo v0.83.1-5AFE0A57+extended windows/amd64 BuildDate=2021-05-02T14:38:05Z VendorInfo=gohugoio

```

PR #7817, opened 2020-10-12
Update goreleaser.yml

ivanvc:
This PR is related to issue #7551. Although the fix is not in this PR, I updated (as I stated in that bug) the file to be compatible with newer `goreleaser` versions.

The issue is that an old version of `goreleaser` has been used to build the releases, and that's what breaks the windows extended build.

This PR updates the `goreleaser.yaml` file to be compatible with the latest current version (v0.144.1), **but to fully fix the bug the person building the release MUST update its copy of `goreleaser`**.

Given that the process to build the release seems somewhat undocumented, I wasn't sure where to document the `goreleaser` required version. Happy to document it if required.


jmooring:
This was addressed with https://github.com/gohugoio/hugo/pull/8015.

PR #7824, opened 2020-10-13
output: Improve layout path construction

moorereason:
```
name      old time/op    new time/op    delta
Layout-2    77.7µs ± 8%    14.0µs ±14%  -81.98%  (p=0.008 n=5+5)

name      old alloc/op   new alloc/op   delta
Layout-2    41.9kB ± 0%     6.3kB ± 0%  -84.88%  (p=0.016 n=5+4)

name      old allocs/op  new allocs/op  delta
Layout-2       487 ± 0%       112 ± 0%  -77.00%  (p=0.008 n=5+5)
```

### DIFF
--- a/output/layout.go
+++ b/output/layout.go
@@ -14,7 +14,6 @@
 package output
 
 import (
-	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


bep:
>Afterward, the layout handler had already cached all the work.

One could argue that that is the correct behaviour for this benchmark. Caches is why most fast software is fast, so they need to be in the benchmarks to get real numbers. It was certainly intended. I suggest you duplicate the Benchmark and name it something else.

bep:
Looks good. Can you squash + reword the commit message. I'm not a fan of commit messages that takes a stab at the existing implementation for no particular great reason. A method that takes 77 microseconds to run where the result is cached is plenty fast enough. But I agree that your implementation is cleaner.

moorereason:
Squashed and neutralized the commit message.  :+1:

PR #7825, opened 2020-10-13
Implement --prettify parameter (only supports HTML for now).

### LINKED ISSUE #7190: Add HTML tidying as an option
https://gohugo.io/news/0.69.0-relnotes/

Given the hugo templating engine creates whitespace hell in output now that  `resources.PostProcess` exists it would be awesome to have a similar hook to tidy the output HTML.

Here's a great Go lib that does it perfectly https://github.com/yosssi/gohtml

fvbommel:
This is similar to `--minify`, but has essentially the opposite effect: instead of making the output *less* readable it makes it *more* readable.

This uses `github.com/yosssi/gohtml` to perform the actual transformation.

Fixes #7190

---

The default configuration currently looks like this:

```yaml
prettify:
  prettifyOutput: false
  disableHTML: false
  html:
    condense: false
    inlineTags:
      - a
      - code
      - em
      - span
      - strong
    inlineTagsMaxLength: 40
```

which is based on the defaults of `gohtml`.

I'm especially looking for feedback on whether the interface (parameter name and config syntax) are reasonable.

## TODOs:
- [ ] For configuration stability the defaults should probably be hard-coded instead of taken from `gohtml`, right?
  - Currently the default configuration just imports the `gohtml` globals in the initializer.
- [ ] I haven't added the line wrapping configuration yet.
  - `gohtml` supports `LineWrapColumn` and `LineWrapMaxSpillover`.
- [ ] Documentation.

Stuff that may happen later:
- [ ] Support other media types (I'd prefer these to be separate PRs though)
  - At least JSON should be doable using `encoding/json`, I haven't looked into others.
- [ ] Add a `Resources.Prettify` pipeline function so it can be used from there too.
  - Less useful if only HTML is supported?

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ Hugo stands on the shoulder of many great open source libraries, in lexical orde
  | [github.com/spf13/viper](https://github.com/spf13/viper) | MIT License |
  | [github.com/tdewolff/minify](https://github.com/tdewolff/minify) | MIT License |
  | [github.com/tdewolff/parse](https://github.com/tdewolff/parse) | MIT License |
+ | [github.com/yosssi/gohtml](https://github.com/yosssi/gohtml) | MIT License |
  | [github.com/yuin/goldmark](https://github.com/yuin/goldmark) | MIT License |
  | [github.com/yuin/goldmark-highlighting](https://github.com/yuin/goldmark-highlighting) | MIT License |
  | [go.opencensus.io](https://go.opencensus.io) | Apache License 2.0 |
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -313,6 +313,7 @@ func (cc *hugoBuilderCommon) handleFlags(cmd *cobra.Command) {
 	cmd.Flags().StringSlice("disableKinds", []string{}, "disable different kind of pages (home, RSS etc.)")
 
 	cmd.Flags().Bool("minify", false, "minify any supported output format (HTML, XML etc.)")
+	cmd.Flags().Bool("prettify", false, "prettify any supported output format (HTML)")
 
 	// Set bash-completion.
 	// Each flag must first be defined before using the SetAnnotation() call.
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -241,6 +241,7 @@ func initializeFlags(cmd *cobra.Command, cfg config.Provider) {
 	}
 
 	setValueFromFlag(cmd.Flags(), "minify", cfg, "minifyOutput", true)
+	setValueFromFlag(cmd.Flags(), "prettify", cfg, "prettifyOutput", true)
 
 	// Set some "config aliases"
 	setValueFromFlag(cmd.Flags(), "destination", cfg, "publishDir", false)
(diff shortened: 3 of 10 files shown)


nu8:
Can you turn this on without `--prettify`? Ideally Id like to put something like
`prettify: true` in my `config.yaml`.

I have always thought it a pain to use `{{-` and `-}}`, but not too much because
then you get minified output. If I understand correct you could just use `{{`
and `}}` and then turn on prettify for a similar output.

fvbommel:
Oh, sorry @nu8, I forgot to add that to the configuration syntax above. (I'll edit it in)

You can enable like this:

```yaml
prettify:
  prettifyOutput: true
```

This is basically the same as [how `--minify` can be enabled via the configuration file](https://gohugo.io/getting-started/configuration/#configure-minify), but with "minify" replaced with "prettify".

fvbommel:
I think it depends on how you structure your pages, but yes I've noticed that too if there are no tags on either end of the white space.

Like the first message says, this implementation uses `github.com/yosssi/gohtml` (as suggested in #7190), and it appears to mostly leave the "middle" of pure-text nodes alone. Mostly though, it seems to do a pretty good job.

If you know of a better HTML formatting lib written in Go though, feel free to suggest it. (I did look around a bit and this is definitely the most popular one I found at least)

Alternatively, you could raise this issue in https://github.com/yosssi/gohtml and/or submit a PR there.

nu8:
The repo itself does remove whitespace. Example:

~~~go
package main
import "github.com/yosssi/gohtml"

func main() {
   s := `<!doctype html><html lang="en"><head><meta charset="utf-8">
<title>May June</title></head><body><p>July August</p>

<p>July August</p></body></html>`
   s2 := gohtml.Format(s)
   println(s2)
}
~~~

Output:

~~~html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>
      May June
    </title>
  </head>
  <body>
    <p>
      July August
    </p>
    <p>
      July August
    </p>
  </body>
</html>
~~~

So not sure what is happening with how the library is being called in this pull
request.

nu8:
Sorry I am dumb. I was using the Pull Request, but I didnt add this:

~~~yaml
prettify:
   prettifyOutput: true
~~~

fvbommel:
Ah yes, that will also "break" it. :rofl:

bep:
Hey, thanks for this and sorry for the delay.

I need to wait with this. I looked into the gohtml package in relation with something else, and there are some bugs in there that means we cannot use this (my memory is a little hazy, but it's related to whitespace handling with text and span elements). I looked into fixing this upstream, but that looked hard ... I have been tinkering with my own html formatter Go package, which I think will get ready soonish ... will let you know.

jwflory:
Is there a chance of this work making it into a future release of Hugo?

BBaoVanC:
Please don't close this, I would love this feature!

hiAndrewQuinn:
I'd be interested in this too. Commenting to keep it alive.

mjp4:
I too would love to see this get in.

awfulwoman:
Just to say that I was looking for exactly this in Hugo! Would love to see it integrated. ❤️

PR #7828, opened 2020-10-13
Allow aliases to have htm suffix

### LINKED ISSUE #7827: Allow aliases to have htm suffix
When specifying an alias, the only allowable extension is `html`. Allow `htm` as well. 

Example:

```md
+++
title = "Post 1"
date = 2020-10-13T06:50:53-04:00
draft = false
aliases = ["foo.html", "bar.htm"]
+++
```

See <https://discourse.gohugo.io/t/allow-aliases-with-htm-suffix/28780>.

PR to follow...

jmooring:
Closes #7827

### DIFF
--- a/hugolib/alias.go
+++ b/hugolib/alias.go
@@ -166,7 +166,7 @@ func (a aliasHandler) targetPathAlias(src string) (string, error) {
 	alias = strings.TrimPrefix(alias, "/")
 	if strings.HasSuffix(alias, "/") {
 		alias = alias + "index.html"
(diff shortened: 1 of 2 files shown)


moorereason:
Can you add tests for both of these changes?

bep:
I think we need to agree that we should do it before we do it.

jmooring:
I proposed this to address a [need expressed in the forum](https://discourse.gohugo.io/t/allow-aliases-with-htm-suffix/28780). I cannot find any other requests for this capability. Closing.

PR #7840, opened 2020-10-15
output: Remove irregular layout lookups

moorereason:
Avoid adding "stuttering" lookup variations such as `index.html.html`.

### DIFF
--- a/output/layout.go
+++ b/output/layout.go
@@ -208,9 +208,11 @@ func resolvePageTemplate(d LayoutDescriptor, f Format) []string {
 }
 
 func (l *layoutBuilder) resolveVariations() []string {
-	var layouts []string
+	var (
+		layouts    []string
+		variations []string
+	)
 
-	var variations []string
 	name := strings.ToLower(l.f.Name)
 
 	if l.d.Lang != "" {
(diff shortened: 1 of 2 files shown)


moorereason:
Travis tests are failing due to Twitter API issues unrelated to this PR.

bep:
It's a little hard to see from your tests, but what happens if I name my home page layout `layouts/index.html.html`?

moorereason:
It's excluded.  `layouts/index.html.html` would no longer be a valid template lookup path.

moorereason:
I realize this technically a breaking change, but I consider this more of a bug than intentional behavior.  I'm also assuming no one is actually using `index.html.html`, but I have no evidence to back up that assumption.

bep:
>I realize this is technically a breaking change, but I consider this more of a bug than intentional behavior.

It's documented and tested, so there is surely some intention living in there somewhere. So me it looks like this is adding complexity and since it's also a breaking change (the "assuming no one using this" is an assumption I've been bitten by many times), and I think we need a better reason than "it's prettier" to change this.

moorereason:
Sounds good.

PR #7849, opened 2020-10-16
Added a new article on hugo

jaivanti:
A worth reading article on how to deploy websites using goHugo


bep:

PR #7851, opened 2020-10-16
highlight: Avoid making unnecessary allocation

moorereason:
Avoid creating a local copy of the highlight configuration when no
options are passed.

Benchmarks of building the docs site:
```
name        old time/op    new time/op    delta
DocsSite-2     1.94s ± 4%     1.93s ± 4%    ~     (p=0.841 n=5+5)

name        old alloc/op   new alloc/op   delta
DocsSite-2     666MB ± 1%     656MB ± 0%  -1.48%  (p=0.008 n=5+5)

name        old allocs/op  new allocs/op  delta
DocsSite-2     8.85M ± 0%     8.76M ± 0%  -1.04%  (p=0.029 n=4+4)
```

### DIFF
--- a/markup/highlight/highlight.go
+++ b/markup/highlight/highlight.go
@@ -37,12 +37,15 @@ type Highlighter struct {
 }
 
 func (h Highlighter) Highlight(code, lang, optsStr string) (string, error) {
-	cfg := h.cfg
-	if optsStr != "" {
-		if err := applyOptionsFromString(optsStr, &cfg); err != nil {
-			return "", err
-		}
+	if optsStr == "" {
+		return highlight(code, lang, h.cfg)
 	}
+
+	cfg := h.cfg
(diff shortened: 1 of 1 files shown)


moorereason:
PS - Benchmark from #7809

bep:
A tip when running benchmarks is to turn off any "turbo boost". If you're using a tool such as my `gobench` which runs before/after in a sequence, with turbo boost on the first run is often artificial better. Since the delta is negative here, I have no doubt there is a positive effect of the above, I just thought I should mention it. 

My lessons learned with Go benchmarks:

* Turn off turbo boos
* Also try to not have a lot of other CPU stuff going on on your computer

bep:
Oh, and the comment above is only relevant for time/cpu, which is not relevant at all for your bench.

moorereason:
Agreed, but I was only looking for mallocs this time around.

PR #7861, opened 2020-10-19
tpl: Update Hugo time to support optional [LOCATION] parameter

### LINKED ISSUE #7860: Add `dateFormatTimeZone` function to allow formatting a date/time string from a specific timezone
Add a new function that internally utilizes Go's `time.ParseInLocation()` such that a date can be formatted from a dynamic `time`, `date`, and `timezone`. 

Specifically, allow a date string that does not contain any timezone information to parse and output a new `dateFormat` like string where the date/time is loaded from a specific timezone identifier (e.g. `America/Los_Angeles`).

Reference: #1882

virgofx:
Fixes #7860 

This PR adds an optional [LOCATION] parameter to the Hugo `time` function which will internally utilize Go's `ParseInTimezone()` functionality to return a `time.Time` instance that is associated with a specific timezone.

- ✔️ No breaking changes. (Added functionality only)
- ✔️ Documentation updated.

### DIFF
--- a/docs/content/en/functions/time.md
+++ b/docs/content/en/functions/time.md
@@ -10,8 +10,8 @@ categories: [functions]
 menu:
   docs:
     parent: "functions"
-keywords: [dates,time]
-signature: ["time INPUT"]
+keywords: [dates,time,location]
+signature: ["time INPUT [LOCATION]"]
 workson: []
 hugoversion:
 relatedfuncs: []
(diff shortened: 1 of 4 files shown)


virgofx:
@bep @moorereason Hey guys. Mind taking a look? I think I addressed the concerns from the other PR. Thanks.

moorereason:
I pushed a commit to your branch to refactor a few things, but it's essentially the same logic.  I tried to clean up the docs to make it clear that the location is a default.  I also try to use "location" instead of "time zone" since it makes more sense to me, but maybe I'm off base there.  Comments welcome.

@bep, PTAL.

virgofx:
Thanks @moorereason . Minor tweaks. LGTM

virgofx:
Thanks @bep and @moorereason

Issue #7826, opened 2020-10-13
Build tags to not compile-in 'hugo deploy' cloud SDKs


fazalmajid:
`hugo deploy` is convenient but pulls in many dependencies and cloud SDKs that can break on certain platforms (#6324) and generally increase the compilation time and size of the executable.

It would be nice to have a a negative build tag `nodeploy` so that functionality could be omitted by building using `go install --tags nodeploy`

emhagman:
Could I tackle this?

fazalmajid:
@emhagman be my guest, I just didn't want to impose on bep's time.

bep:
>Could I tackle this?

Please do.

emhagman:
Great, thanks :+1:

emhagman:
https://github.com/gohugoio/hugo/pull/7868 took a stab at it. Let me know if you'd like me to remove the command all together or have `hugo deploy` return an error (which is currently what I coded)

Issue #6324, opened 2019-09-12
Azure support breaks Solaris compatibility


fazalmajid:
```
local64 ~/go>go get github.com/gohugoio/hugo
# github.com/Azure/azure-storage-blob-go/azblob
src/github.com/Azure/azure-storage-blob-go/azblob/highlevel.go:144:7: undefined: mmf
src/github.com/Azure/azure-storage-blob-go/azblob/highlevel.go:146:12: undefined: newMMF
src/github.com/Azure/azure-storage-blob-go/azblob/highlevel.go:275:13: undefined: newMMF
```

I have zero interest in Azure, and this is breaking my SmartOS hosted environment. Is there a way to disable Azure support in Hugo at compile time?

bep:
>Is there a way to disable Azure support in Hugo at compile time?

Currently, no. But we should at least make sure to turn it off for unsupported platforms.

You're the first Hugo user on Solaris I've heard of. I thought we had a bug in file watching on Solaris (in `fsnotify`)

/cc @vangent 

https://github.com/Azure/azure-storage-blob-go/blob/8a1deeeabe0a24f918d29630ede0da2a1c8f3b2f/azblob/zc_mmf_unix.go#L12

uniapi:
He is **not** the only Solaris user as there are plenty of us!

The same issue:
```
# github.com/Azure/azure-storage-blob-go/azblob
../../Go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.7.0/azblob/highlevel.go:144:7: undefined: mmf
../../Go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.7.0/azblob/highlevel.go:146:12: undefined: newMMF
../../Go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.7.0/azblob/highlevel.go:275:13: undefined: newMMF
```
Still not fixed?

jmooring:
https://github.com/Azure/azure-storage-blob-go/pull/117 was merged on 23 June 2020. Can anyone confirm that _this_ issue is resolved?

fazalmajid:
I can confirm I can build Hugo@latest on my SmartOS (Illumos) box, using Go 1.20.6

PR #7868, opened 2020-10-21
build: allow optional "nodeploy" tag to exclude deploy command from bin

emhagman:
Fixes https://github.com/gohugoio/hugo/issues/7826

You can run `HUGO_BUILD_TAGS=nodeploy mage install` or `go install --tags nodeploy` to build `hugo` without the `deploy` command which will make the binary smaller and prevent crashes on certain systems https://github.com/gohugoio/hugo/issues/6324

I kept the command in there by creating `nodeploy.go` that simply returns an error if the user tries to call `hugo deploy`.

```
eric@eric-manjaro-ml ~/p/hugo (feature/build-without-deploy)> hugo deploy
Error: build without HUGO_BUILD_TAGS=nodeploy to use this command
```

Please let me know if you'd like me to remove the `deploy` command all together or keep it like this to warn the user.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 
 FROM golang:1.15-alpine AS build
 
-# Optionally set HUGO_BUILD_TAGS to "extended" when building like so:
+# Optionally set HUGO_BUILD_TAGS to "extended" or "nodeploy" when building like so:
 #   docker build --build-arg HUGO_BUILD_TAGS=extended .
 ARG HUGO_BUILD_TAGS
 
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -11,6 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build !nodeploy
+
 package commands
 
 import (
--- a/commands/nodeploy.go
+++ b/commands/nodeploy.go
@@ -1,0 +1,49 @@
+// Copyright 2019 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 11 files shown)


tuananh11091992 (approved):

Dockerfile
```diff
@@ -4,7 +4,7 @@
 
 FROM golang:1.15-alpine AS build
 
-# Optionally set HUGO_BUILD_TAGS to "extended" when building like so:
+# Optionally set HUGO_BUILD_TAGS to "extended" or "nodeploy" when building like so:
```
moorereason, line 7: `HUGO_BUILD_TAGS` is a comma-separated list of tags.  `extended` and `nodeploy` are not mutually exclusion.

emhagman: I can remove it from the Dockerfile if you'd like or change the copy

bep: Just leave it as it is now. We should not have this info spread out into ... hundreds of files, and I'm not even sure the Mage specific env var is the right form of documentation either.

But for now, leave this PR as it is. It is fine for now.

emhagman: Okay, sounds good :+1:

emhagman:
@bep Looks like CI failed for some unrelatedreason:

```
ERROR 2020/10/21 23:25:43 Failed to get JSON resource "https://api.instagram.com/oembed/?url=https://instagram.com/p/BWNjjyYFxVx/&hidecaption=1": Failed to retrieve remote file: Bad Request
```

Anything else I need to do for this?

bep:
>Anything else I need to do for this?

No, that error is unrelated (and fixed in the main branch).

PR #7875, opened 2020-10-22
Allow cascade _target to work with non toml frontmatter

### LINKED ISSUE #7874: Cascade with _target discriminator only works for TOML frontmatter

<pre>
$ hugo version
Hugo Static Site Generator v0.76.5/extended darwin/amd64 BuildDate: unknown

</pre>

**Yes** (and with `master`)

Defining a cascade with the new `_target` discriminator (#7782) using TOML for the frontmatter works as expected, using other formats such as YAML causes the cascade to be ignored completely

This appears to be due to the way that the TOML library unmarshals slices.  I have a fix working locally; will open a PR shortly

gwatts:
The TOML lib unmarshals slices of string maps to []map[string]interface{}
whereas YAML and JSON decode to []interface{}

The existing tests only check for TOML working correctly, and _target
with cascade did not work at all for frontmatter defined in other formats.

Add a function to normalize those slices

Fixes #7874

### DIFF
--- a/common/maps/maps.go
+++ b/common/maps/maps.go
@@ -14,6 +14,7 @@
 package maps
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/gobwas/glob"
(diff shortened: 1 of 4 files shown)


bep:
Thanks for this, much appreciated. I should have remembered this by now.

PR #7890, opened 2020-10-26
Use ContentBaseName() as fallback page Title

### LINKED ISSUE #3805: Use filename as title when it’s not defined in front matter

### LINKED ISSUE #7370: If frontmatter `title` property is empty use filename instead
This allow to use hugo to generate sites from directories of plain markdown files. For example my personal notes directory or a directory of markdown documentation in a similar way to how GitHub and Gitlab allows you to view markdown files.

rigtorp:
Closes #7370 
Closes #3805
Related to #6098

The use case is to quickly generate a site from a directory of markdown files without front matter.

It's also possible to parse `ContentBaseName()` and generate a better title, something like a `unslugify()` function.

### DIFF
--- a/hugolib/page__meta.go
+++ b/hugolib/page__meta.go
@@ -285,6 +285,9 @@ func (p *pageMeta) Sitemap() config.Sitemap {
 }
 
 func (p *pageMeta) Title() string {
+	if p.title == "" && !p.File().IsZero() {
+		return p.File().ContentBaseName()
+	}
 	return p.title
 }
 
--- a/hugolib/page_test.go
+++ b/hugolib/page_test.go
@@ -1300,7 +1300,7 @@ Content.
 `)
 
 	b.Build(BuildCfg{})
-	b.AssertFileContent("public/index.html", "Title||")
+	b.AssertFileContent("public/index.html", "Title|_index|")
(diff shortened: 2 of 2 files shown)


bep:
Thanks for this, and sorry for the delay. I understand the problem, and I suspect many people would want this patch applied. But it would be a breaking change (having no title could be a deliberate choice). One could think of adding some config option here, but I want to hold that thought until we see how "non file" content sources work in this context.

Issue #5710, opened 2019-02-22
Add to oss-fuzz


bep:
https://github.com/google/oss-fuzz

I have done some fuzzing in the past (with https://github.com/dvyukov/go-fuzz). Might as well take it the extra mile, as fuzzing can find some really curious bugs ...

PR #7918, opened 2020-11-02
Add a fuzzer

AdamKorcz:
This PR answers https://github.com/gohugoio/hugo/issues/5710 by adding the first fuzzer to Hugo.

I can confirm that this fuzzer runs on oss-fuzz's infrastructure. I have setup the integration with oss-fuzz which can be found [here](https://github.com/google/oss-fuzz/pull/4589). The build on the oss-fuzz side will fail, and that is because this fuzzer is not merged in yet upstream. Once this fuzzer is merged in here, I will run the integration test again within oss-fuzz and the tests will then pass.

### DIFF
--- a/tpl/transform/fuzz.go
+++ b/tpl/transform/fuzz.go
@@ -1,6 +1,6 @@
 // +build gofuzz
 
-// Copyright 2017 The Hugo Authors. All rights reserved.
+// Copyright 2020 The Hugo Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


bep:
I will merge this to get the "machinery" up and running, but I would appreciate if we could have a discussion (as to where/what) before we add random fuzzers around Hugo. 

But, could you:

* Update the copyright year in the file you added
* Get the commit message in line with the guidelines here ("Added" => "Add").

AdamKorcz:
Sounds great with opening a discussion around the best approach for fuzzing Hugo.

The reason for adding a fuzzer before starting a discussion was to get things started with fuzzing Hugo with a specific example and suggestion.

PR #7928, opened 2020-11-04
tpl: Add functions for hex encoding

moorereason:
Use case that prompted this PR: https://discourse.gohugo.io/t/29130


doggzkie (approved):

bep:
I have seen 1 person ask for this. This seem like a rather special thing and we need to wait for the demand to rise before adding it.

bep:
I will close this. We can revisit this if more people chime in.

PR #7946, opened 2020-11-11
Fix Typo in 0.78 Release Notes

vsopvsop:
Curtsy has a whole different meaning.

curtsy of he  => courtesy of the


bep:

PR #7947, opened 2020-11-11
Fixed wrong var assignment

marceloverdijk:


bep:

PR #7965, opened 2020-11-17
remove unneeded meta tag from blog example

coliff:
```html
    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
```
This was only needed for IE10 and lower (which are no longer supported) and chromeframe which was a Google plugin for IE8 and lower and was discontinued many years ago. So this meta tag can be safely removed.

### DIFF
--- a/examples/blog/layouts/partials/meta.html
+++ b/examples/blog/layouts/partials/meta.html
@@ -1,6 +1,5 @@
 
     <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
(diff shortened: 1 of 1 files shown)


davidsneighbour (approved):
correct. 

on a sidenote: the author tag was never a "real" meta tag and can be safely removed too because Google is not using it to assign articles to author entities. ;)

coliff:
Yep, I think there's quite a lot of improvements and optimizations could be made to this example. If this one can be merged I'll look at making separate PRs later.

PR #7983, opened 2020-11-24
Allow /static (only) to be written to disk in server mode

satotake:
Currently, Hugo has 2 kinds of render destination `memory` or `disk`.
During local development, I prefer `renderToMemory` to `renderToDisk` in order to enjoy Hugo's advantage. I mean fast rendering.

However, staticfiles-intensive site development with `rendreToMemory` consumes machine memory linearly.
For example. 3GB static directory consumes 6GB memory on this setting. 
(yes, this is exceptionally huge example but, unfortunately, it is true of mine)

To resolve this situation, I would like to introduce another render target, `composite`, using afero's composite backend.
This filesystem serves only dynamic contents (`content` and etc) from writable memory.
On the other hand, static files are served directly from os filesystem without syncing to memory.

Contrary to its smallness of diff, it significantly reduces memory consumption.
In addition, as far as I consider, it does not have any major disadvantages.

I attached simple benchmark results. 
In this benchmark, ~60MB file was copied to static direcotry per 10 sec up to ~1.2GB (total ~200 sec).

[memory.log](https://github.com/gohugoio/hugo/files/5592335/memory.log)
[composite.log](https://github.com/gohugoio/hugo/files/5592336/composite.log)

I believe that this we could even replace `renderToMemory` mode with `composite` mode as long as we confirm its compatibility.
But this is prototyping. I am not so confident with Hugo's module system that I added `composite` as another option rather than rewriting `renderToMemory` for now.


bep:
I understand the problem: Limit the memory allocation while stile getting the speed benefits of files written often. But as /content seems to be the new /static, the above may work for you, but not for me (and others). I tend to have very few files in static.

But I agree that we need to fix this. And since we cannot determine "is it dynamic" by its location, maybe we can:

* Write all files to this composite
* Somehow detect the file size before write and have a threshold per file
* Maybe also have a global threshold (if written more than n bytes and file does not exist already, write to disk)

Or something ....?

satotake:
Thank you for your time @bep

> as /content seems to be the new /static
Sorry, I cannot understand this. What is your concern?
It suppose to me that many of Hugo users have a little effect by this.
But I think it is positive effect at least.

> we cannot determine "is it dynamic" by its location

I assume that "files in static directory can be serve as it is. without processing". Is this wrong?

At the implementation level, all files are written to this composite fs in this mode.
So I feel it works correctly.

bep:
>Sorry, I cannot understand this phrase. What is your concern?

You say you have 3GB in your static directory. I assume these are mostly images. If I put these images inside /content I would have the same problem as you, but this PR does not solve it.

satotake:
> I assume these are mostly images

Yes. Most of them are media or data files in terms of file size.

>  If I put these images inside /content I would have the same problem as you, but this PR does not solve it.

I see.
So you mean that target directories are not only `static` but `content` and etc.


It was my intention that this targets only `static` dir because it is easy to determine their files are "static".
At least, we do not have to keep them inside memory.
As a result, logic is simple and it is easy to start trying without redundant disk space. we need just a single flag.
(In addition, it is bottle neck directory in my case.)

I bet you have got a my point. I mean, my original goal is reducing memory usage by static directory.
So if your method is more comprehensive, it will be ok.
At the same time, I think it will be beneficial (at least not harmful) for everyone even if we focus on static directory at first.

Anyway, thanks for your consideration!

satotake:
Second thought.

We may be able to add `content` directory to this composite, too
Dynamic content results inside MemoryFs are always prior to read-only layer. they are always served correctly in this situation.
We do not need to care about dynamic content results. 

In that case, unexpected items are only precursors of dynamic contents, such as `content/post/foo.md`.
Is there any good way to filter them? .. but it might be more simple than something like dependency tree.
Personally, I do not mind that Hugo serves markdown files during local development.

davidovich:
I am a bystander here, but would it not be possible to specify the *big files* in the config file à-la .gitignore that would allow the author to choose and select the files that are loaded from disk? This would permit the flexibility of storing these files wherever the user chooses.

satotake:
@davidovich Thank you for your comment opinion.

> would it not be possible to specify the big files in the config file à-la .gitignore that would allow the author to choose and select the files that are loaded from disk? 

Possible.
If my understanding is correct, this discussion has 2 points.

1.  how to determine target files
2. whether these targets are written to disk or directly served from original directories

Now, I agree with that more hugo users will have benefits when the setting is configurable or we can detect static or not.

In my opinion, this is hugo's config so it will be good to be written in Hugo's config file.
On the other hand, your comment hit the spot for me.
Because it implies glob pattern may be suitable in this case. like `static/**/*` `content/**/!(*.md)` `assets/**/!(*.scss)`
How do you think about this? @bep 

canditates
- file size based threshold
- global threshold
- directory-based (not be flexible enough)
- glob pattern
- media types?

Second, I'm wondering if we need to render these to another disk space.
If it is ok, Why not serve static files from original direcotries? I think it will be cheaper than disk writing operation.

satotake:
Retried.

> Write all files to this composite

At this time, I intended to wite all files into composite fs as @bep said.
Exception is ".md", ".html" and etc. It is based on file extension and follows [currrent `IsContentFile` logic](https://github.com/gohugoio/hugo/blob/d90e37e0c6e812f9913bf256c9c81aa05b7a08aa/hugofs/files/classifier.go#L68).
If we apply not whitelist (which files we should include) but the blacklist (which files we should exclude),
allmost all of the files are managed under comositeFs efficiently without further configurations.
To filter exceptions, `FilenameFilterFs` is added.

For now, I have specified only `contentDir` and `staticDir` since I assume that the other dirs do not have any "static" files.
(it will be useless to use composite Fs)

But even if necessary, it is as easy as `contenteDir` to add extra target dir (maybe assetDir?) with `FilenameFilterFs`

as far as I tested, it works well.

Any suggestions are welcome. @bep , @davidovich
Especially, I am wondring the possiblity of negative effects by this change

satotake:
benchmark of hugoDocs
memory usage was reduced by 10-20% even though hugoDocs' static directory is not so large

[composite.memstats.log](https://github.com/gohugoio/hugo/files/5831604/composite.memstats.log)
[memory.memstats.log](https://github.com/gohugoio/hugo/files/5831606/memory.memstats.log)

satotake:
I have added 2 configurations and related tests
- file size threshold
- filename-based exclusion

Here is example yaml configuration. you can add and configure any directories

```yaml
staticDir:
    - static
    - static_en

staticFileRules:
   static:
   static_en:
   contentDir:
    ignore:
      - "\\.md$"
      - "^content/non-static/*$"
    size: 500 # KB
```

@bep 
How do you think?
Do we have to care about any other points?

bep:
Hey, 

And sorry for the late response on this; I agree that we really should add this, but the added complexity has been pulling me away from it ... 

I have had a look and to me it looks a little too complicated.

Note that the below is probably a little too simplified, but:

In my head, we could add a new flag/config option for the server:

```
renderStaticFilesToDisk = true
```

If that is set to true we:

1. Use the OS filesystem when syncing in `copyStaticTo`
2. Use a composite for the server.

satotake:
@bep 

I am willing to simplify this.
In my case, I will simply apply this for the whole `static` direcoty when it is released, do not need further configuration.
ok, I will revise user-interfaced option/flag/configuration before merge.

By the way,

> Use the OS filesystem when syncing in copyStaticTo

If my understanding is correct, you mean static files are written into `public` dir, right?
I have tried [here](https://github.com/satotake/hugo/commit/0b068eb1a002b4877415f734c1fb6a959c5d340d)
But this does not reduce memory eating.

Probably, this may be because `afero.CopyOnWrite` fs copy all `public` files into memory on full builds happen.

> Any attempt to modify a file found only in the base will copy the file to the overlay layer before modification (including opening a file with a writable handle).

Previous implementation does not exec any ops of static files. thus, memory consumption is relieved.
Or, do I misunderstand something here?

satotake:
@bep 

> Or, do I misunderstand something here?

It was my misunderstanding.  I have to consider files in contentDir.
Here, I implemented as `--renderTo hybrid` as you said.

I would like you to 
- check this implementation
- compare `hybrid` with `composite` 

My recommendation is `composite` because it is faster.
But It is ok for you to select `hybrid` considering complexity.
Finally, I will clean up and simplify configurations and so on using `renderStaticFilesToDisk`

bep:
>If my understanding is correct, you mean static files are written into public dir, right?
>I have tried here
>But this does not reduce memory eating.

The "memory eating" comes from the static files being written to memory. So if we make sure that we don't do that, this should work.

Again, sorry if I'm oversimplifying it, but if the flag mentioned above is enabled:

* Make sure that the FS in the static file syncer is the OS FS.
* Make sure the FS used by the HTTP server is a composite of /public (OS) and the memory destination.

bep:
Do you have a test project online that demonstrates this "memory eating" problem?

satotake:
> this should work.
Yes. After fixation, both modes improve memory usage.
OK, I will simplify my implementation soon. Please wait.

> Do you have a test project online that demonstrates this "memory eating" problem?

No but it is easy to reproduce this.
- create minimal hugo site
- copy some files into static or content directory

Hugo consumes about ~15MB memory in my environment. So you can also confirm this by copying 50-100MB file(s)

satotake:
@bep Simplification done. Could you review again?
For testing, use this
```sh
hugo new site memory-eating-test
cd memory-eating-test
# generate 100MB static files
head -c 100000000 /dev/urandom > static/large-file
cp static/large-file content/large-file

```

bep:
@satotake OK, thanks for your pushing this, much appreciated. The latest patch looks very clean. I'm planning a Hugo release on Monday and will get this merged before then. I will test this a little myself ... The only thing I think about is that we may consider making write static files to disk the default option ... I suspect many have this issue without knowing it.

satotake:
@bep
>  I will test this a little myself

Thank you. I recovered flag's description

> The only thing I think about is that we may consider making write static files to disk the default option

As an author of this, it is welcome.
But it may surprise users those who use only `renderToMemory`, generating public directory after updating.
It seems to be a breaking change.


I know this is not your interest ... and this is my last emphasis but it may be worth taking your time to implement the approach, which I call `composite` here, even after merging this.
If it can cover current static syncer's behaviour, 
- it will not be breaking change
- some static file sync logic can be reduced
- it gains some performance

I believe you can because the basic concept is not complicated as it looked here.

commands/server.go
```diff
@@ -46,15 +46,16 @@ type serverCmd struct {
...
-	serverAppend      bool
-	serverInterface   string
-	serverPort        int
-	liveReloadPort    int
-	serverWatch       bool
-	noHTTPCache       bool
+	disableLiveReload       bool
+	navigateToChanged       bool
+	renderToDisk            bool
+	renderStaticFilesToDisk bool
```
bep, line 52: I sugget renaming renderStaticFilesToDisk => renderStaticToDisk -- a little shorter and easier to type.

commands/static_syncer.go
```diff
@@ -56,6 +57,9 @@ func (s *staticSyncer) syncsStaticEvents(staticEvents []fsnotify.Event) error {
 		syncer.ChmodFilter = chmodFilter
 		syncer.SrcFs = sourceFs.Fs
 		syncer.DestFs = c.Fs.Destination
+		if c.renderStaticFilesToDisk {
+			syncer.DestFs = afero.NewOsFs()
```
bep, line 61: I don't think this is good enough. There's more to `c.Fs.Destination` than just the filesystem; in server mode we also (sometimes) wraps it with a change detector and a duplicate counter. So I suggest you add a `Fs.DestinationStatic` that mostly gets to be identical to the regular one.

hugolib/filesystems/basefs.go
```diff
@@ -55,6 +55,9 @@ type BaseFs struct {
 	// This usually maps to /my-project/public.
 	PublishFs afero.Fs
```
bep, line 57: All the code below needs to be reverted. If I remember correctly, there is no static files copied in the regular site building process.

bep (changes requested):

hugolib/filesystems/basefs.go
bep: Or is this used for the files that gets copied only? I don't see that distinction in the code below.

commands/commandeer.go
```diff
@@ -360,6 +362,13 @@ func (c *commandeer) loadConfig(mustHaveConfigFile, running bool) error {
 		if c.destinationFs != nil {
 			// Need to reuse the destination on server rebuilds.
 			fs.Destination = c.destinationFs
+		} else if createMemFs && c.renderStaticFilesToDisk {
+			// Writes the dynamic output on memory,
+			// while serve others directly from publishDir
+			publishDir := config.GetString("publishDir")
+			writableFs := afero.NewBasePathFs(afero.NewMemMapFs(), publishDir)
+			publicFs := afero.NewOsFs()
+			fs.Destination = afero.NewCopyOnWriteFs(afero.NewReadOnlyFs(publicFs), writableFs)
```
bep, line 371: Also re my comment below, I don't think you can change this global `fs.Destination`.

hugolib/filesystems/basefs.go
satotake: > Or is this used for the files that gets copied only?

Yes. This is used [here](https://github.com/gohugoio/hugo/blob/59d3913e9af6d99481e3a8fb82ce4122ae8c6ead/hugolib/pages_process.go#L162)

If not, static files in `contentDir` are copied to memory. This will not be what we want.

(I fixed the name)

commands/commandeer.go
satotake: This is changeable and, actually,  required.
Without this, we cannot access any static files (,or files in `public` directory) via local server.
We can see it almost identical to "afero.MemMapFs" (like the below clause).
The difference is that the fs can serve files from `public` directory as well as memory

commands/static_syncer.go
satotake: fixed

satotake:
@bep 
Revised or commented. My response may be delayed? If so, sorry

satotake:
I added some test cases and caught one bug, which is related to file deletion event of static sync.
Do you have any considerations?, @bep

bep:
@satotake I'm sorry that this got lost in the backlog, I was reminded of this by a discussion on the forum. I will have a look at this today.

bep:
I have rebased and will complete this in here: #9626

PR #7987, opened 2020-11-25
add template execution tracing

isotopp:
I am struggling with understanding how Hugo works. Part of the problem is that I cannot see
how templates are being considered and which are chosen for execution.

This small patch is my first attempt at Go. I am confident that I am logging template execution
properly. I seems that there is some need for this (https://discourse.gohugo.io/t/see-which-template-is-used-by-hugo/9983 et al).

Example usage:

```
t$ ~/Source/hugo/hugo -DEF -cleanDestinationDir --debug
INFO 2020/11/25 09:54:29 Using config file:
Start building sites …
INFO 2020/11/25 09:54:29 syncing static files to /home/kris/hugo/quickstart/public/
DEBUG 2020/11/25 09:54:29 Render page Categories (template _default/terms.html) to "/categories/index.html"
DEBUG 2020/11/25 09:54:29 Execute template: "_default/terms.html"
DEBUG 2020/11/25 09:54:29 Execute template: "partials/site-favicon.html"
...
```

Using the --debug flag,

- the template chosen is printed with the Render message
- template.Execute() is being logged, showing also partials


bep:
Thanks. This looks good. Can you

* Push the CLA button above
* Adjust the commit message according to guidelines (in your case "add => Add"

isotopp:
- Email certified
- CLA signed
- Commit message edited as per request

bep:
>Commit message edited as per request

I can easily do it for you, but I guess it could be useful for you to know: If you update the title of the PR in GitHub that will not update the title of the commit message itself. To do that you need to do something ala:

```
git commit --amend
# edit commit message
git push -f
```

isotopp:
Thanks.

I did the force-push as per your request.

I refrained from doing this before, as I have been taught that amending and force-pushing usually leads to a lot of breakage.

bep:
> as I have been taught that amending and force-pushing usually leads to a lot of breakage.

In general you are right, but I think one accepted exception to that rule is pull requests.

isotopp:
Since I got your attention: The second half to the question I am trying to answer would be "As a user, I want to be able to trace Hugo as it loads or considers templates/files, and find/does not find them."

As I am learning Go and am familiarizing myself with the codebase, my theory would be that I have to look
into the hugofs part of the codebase? Or where would I have to look?

bep:
It's a 2 step operation:

1. https://github.com/gohugoio/hugo/blob/master/output/layout.go#L66
2. The above gives a set of candidates for a given page, we then pick the first of we find of those.

tpl/tplimpl/template.go
```diff
@@ -224,6 +224,7 @@ func (t *templateExec) Execute(templ tpl.Template, wr io.Writer, data interface{
 		defer t.Metrics.MeasureSince(templ.Name(), time.Now())
 	}
 
+	t.d.Log.Debug().Printf("Execute template: %q", templ.Name())
```
bep, line 227: OK, I didn't notice this the first time I looked at this. I'm not a big fan of permanent debug statements in the first place, but especially here in what I would consider to be somewhat of a hot path. There is a cost of doing that debugging (even if it's turned of), so unless it's really important, I'd rather we delete that line.

bep (changes requested):

tpl/tplimpl/template.go
moorereason: I'd like to keep this as it's valuable debug info.  We should be able to protect this line with an if-statement if we're concerned about the Printf exec time:

```go
if t.d.Cfg.GetBool("debug") {
```
The config lookups are cached.

Even with debug enabled on the Hugo docs site, this doesn't add much to the build time (if you redirection stdout to a file).

isotopp: I amended the PR as requested.

tpl/tplimpl/template.go
```diff
@@ -224,6 +224,9 @@ func (t *templateExec) Execute(templ tpl.Template, wr io.Writer, data interface{
 		defer t.Metrics.MeasureSince(templ.Name(), time.Now())
 	}
 
+	if t.d.Cfg.GetBool("debug") {
+		t.d.Log.Debug().Printf("Execute template: %q", templ.Name())
```
bep, line 228: I'm not merging this debug statement (see my previous comment; in this case it's the cost of the `.GetBool`). I'm fine with the first change

PR #7997, opened 2020-11-28
resources: Extend HasShortcode to check for any shortcodes

### LINKED ISSUE #7996: .HasAnyShortcode : return TRUE if a page uses at least one shortcode
Similar to `.HasShortcode`, but without having the need to specify a shortcode name as parameter.

`.HasAnyShortcode` would Return TRUE if at least one shortcode is used in the page else Return FALSE

_Use case: CSS/JS libraries containing the code needed for a set of shortcodes. This would avoid to have to check for existence of all shortcodes one by one._

Alternatives could be:
- a specific parameter for `.HasShortcode`, like `.HasShortcode Any`
- use `.HasShortcode` with an array to test if one of the shortcodes in the array is used

moorereason:
Make HasShortcode parameter variadic to allow for the testing of any
shortcodes instead of only named shortcodes.

Fixes #7996

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -23,38 +23,28 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/mitchellh/mapstructure"
-
-	"github.com/gohugoio/hugo/identity"
-
-	"github.com/gohugoio/hugo/markup/converter"
-
-	"github.com/gohugoio/hugo/tpl"
-
-	"github.com/gohugoio/hugo/hugofs/files"
-
-	"github.com/bep/gitmap"
-
-	"github.com/gohugoio/hugo/helpers"
-
-	"github.com/gohugoio/hugo/common/herrors"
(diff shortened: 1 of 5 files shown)


hugolib/page.go
```diff
@@ -319,12 +308,20 @@ func (p *pageState) Resources() resource.Resources {
 	return p.resources
 }
 
-func (p *pageState) HasShortcode(name string) bool {
+// HasShortcode returns whether the page has a given named shortcode or any
+// shortcodes at all.  To see if any shortcodes are present, do not pass a name
+// parameter.
```
bep, line 313: This behaviour does not feel natural to me. I would say that:

* HasShortcode nil => true if any shortcodes at all
* HasShortcode a b => true if both a and be exists

Also:

* All the whitespace/formatting changes I assume does not come from `gofmt`?
* I don't see any tests for this new logic.

PR #8010, opened 2020-12-03
Adds ability to sort ByVersion (following semver format)

kevinawoo:
User are are able to specify `version` in the frontmatter, following the [semver format](https://semver.org/) for sorting. Pages that don't have `version` will be sorted to the end.

### DIFF
--- a/docs/content/en/templates/ordering-and-grouping.md
+++ b/docs/content/en/templates/ordering-and-grouping.md
@@ -1,5 +1,5 @@
 ---
-title: Ordere and Grouping Hugo Lists
+title: Order and Grouping Hugo Lists
 linktitle: List Ordering and Grouping
 description: You can group or order your content in both your templating and content front matter.
 date: 2017-02-01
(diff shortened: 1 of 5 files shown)


alexlafreniere:
If there is any way I can contribute to moving this PR along please let me know. We are desperately seeking proper sorting for semantic version numbers on our Hugo-powered documentation site. Given the age of the PR, does it make more sense to re-open a new PR against the latest version of Hugo rather than attempt to solve the merge conflicts?

jmooring:
> Does it make more sense to re-open a new PR against the latest version of Hugo rather than attempt to solve the merge conflicts?

Neither.

Please create an issue (proposal) to describe the problem and why it is important. I wouldn't invest any time in a PR unless the project lead indicates that it would even be considered.

Personally, I would rather see this functionality added to the existing sort function. Something like:

```
{{ range (sort .Pages "Params.version" "semVerAsc" ) }}
{{ range (sort .Pages "Params.version" "semVerDesc" ) }}
```

Benefits:
- I can use any front matter field name (e.g., version, release, sw_ver)
- I can sort anything, not just page collections

alexlafreniere:
> Personally, I would rather see this functionality added to the existing sort function.

An even better idea! Drawing up a proposal now. Thank you for the input and guidance.

Issue #7297, opened 2020-05-20
Add .RenderShortcodes


bep:
I have been hit by this issue twice recently -- so I assume this is a problem worth solving (I know there are similar issue, and this also relates to `.RenderString`.

Currently we have:

* `.Content` which gives rendered HTML with shortcodes rendered.
* `.RawContent` which gives the raw source excluding front matter. No shortcodes rendered.

A common pattern is to include files using `.GetPage` via a shortcode, e.g.:

```md

# Title

{{< include "foo.md" >}}

```

Using the `{{<` means you can include just about anything and it will be preserved as-is. But it will not partake in the global rendering context re. table of contents, foot notes etc.

You can do this in your shortcode:

```html
{{ (site.GetPage "foo.md").RawContent n}}
```

And then

```md

# Title

{{% include "foo.md" %}}

```

The above will work but any shortcodes in `foo.md` will not be rendered.

Which this issue is all about.

I assume that it makes sense that `.RenderString` uses this method internally by default, but what I need is basically:

```html
{{ $markdownWithRenderedShortcodes := .RawContent | .RenderShortcodes }}

/cc @regisphilibert @onedrawingperday @digitalcraftsman @moorereason  and gang

And yes, this is tightly connected to a `Page` so I think a method makes sense.

theory:
Would this feature, if implemented, render the shortcodes for the current output format? That'd be super great!

bep:
Yes.

blattersturm:
Is this going to get implemented at all any time soon? People keep asking how to build my PR and since no equivalent functionality exists in mainline and @bep outright rejected my PR likely as they felt this was a better way to implement this (and therefore it became _the_ only implementation they'd accept, apparently) I don't have any actual alternative option to get shortcodes processed in JSON output.

davidsneighbour:
Checking your PR it was rejected not outright for a feeling but for missing formalities. Open an issue, describe and discuss your issue, then if it is deemed a good addition to hugo open a PR referencing that issue number.

jmooring:
A need for this came up again yesterday:
<https://discourse.gohugo.io/t/export-to-markdown-with-shortcodes/39590>

Rumperuu:
And again: https://discourse.gohugo.io/t/table-of-contents-render-hook/45352

theory:
Tried it out, works great, thanks!

Issue #6703, opened 2020-01-02
Make `.RenderString` render shortcodes


regisphilibert:
As [requested](https://github.com/gohugoio/hugo/issues/6666#issuecomment-568779670) I'm re-creating this issue.

Now that we have `.RenderString`. The problem mentioned in https://github.com/gohugoio/hugo/issues/6094#issuecomment-513116608 has been lifted and we should consider having `.RenderString` process shortcodes.

### Do we really need this?

When building a page with blocks of data (CMS Blocks,  Page Builders and so on), most of those are saved as Front Matter. It then falls on the developer to range on those blocks and apply the proper data processing/partial loading etc...

In the context of a very nice to have page builder, Hugo shortcodes are currently unusable by the editor and any search-and-replace workaround cooked up by developers (been there) will never match the reliability and usability of Hugo's shortcode API.

### Will it break code.

The `.RenderString` method is fairly new and I don't see anyone including shortcodes in there with the expectation that they would not be processed. So I think: No

aaronpowell:
I just came across `.RenderString` as a way to render some markdown that's in the front matter of a page and was surprised it didn't handle shortcodes (then I read the docs and saw it stated).

I see value in it handling that, if not, an alternative like `shortcodify` like was mentioned in the parent issue.

mlake:
also seeing a need for this..

bep:
Filling up my notification screen with "I need this, too" does not help.

regisphilibert:
Hello @mlake, usually a "thumb up" emoji on the issue's description is enough to notify of your interest in this issue being treated. You'll get the same effect without the noise. Thanks a ton!

NorseGaud:
This is a huge need for us. Partials allow for a lot of DRYness in our docs site and not being able to include markdown with other shortcodes in it (like relrefs) is painful.

bric3:
Hi I stumbled on this issue, and I wonder if the _new_ `.RenderString` that could process shortcodes could help me in the below case.

----

I have a mardown table that I'd like to wrap in a div, and I'd have shortcodes to highlight / mark some text in this table.

```
{{< wrapTable >}}

| Memory Type | # of Segments | Used Memory(bytes) | Used Memory(%) | Free Memory(bytes) | Free Memory(%) | Total Memory(bytes) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
[...]

Issue #6094, opened 2019-07-18
Add a template function to process strings with shortcodes in them.


zivbk1:
With a frontmatter variable like:

`biography = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut sed nunc vitae turpis eleifend tincidunt. In hac habitasse platea dictumst. {{< figure src="/media/spf13.jpg" title="Steve Francia" >}} Pellentesque viverra a tellus in mollis. Sed vitae congue odio, vitae tristique magna. Mauris id fermentum libero, sit amet ultricies metus. "`

Provide a template function `shortcodify` that will render the figure shortcode into the output string.

This would allow shortcodes to be used in frontmatter data, within data file content, and expand the capabilities of Hugo for pages that require more than one 'content' block.  See the [Sawmill layout technique](https://forestry.io/blog/sawmill-layout-composer-for-hugo-and-forestry/) for an example of this use case, which is already supported by the Forestry.io CMS.

bep:
>Provide a template function shortcodify that will render the figure shortcode into the output string.

Shortcodify? :-)

The shortcode's API requires a `Page`, so it couldn't be used in data files, and I think the "more than one content" block needs to be solved differently. I have plans for that. We could possibly, maybe adjust the shortcode's API to allow different things, and make this somehow work, but that is a rather big change and will need more than one person needing it.

bwklein:
Yeah, I was trying to follow the naming model of some of the other functions. 😄  Call it whatever you like, I really just want the ability to convert a shortcode in the string into the HTML (or whatever) in the shortcode template.

I didn't realize there was a requirement of a `Page`, I assumed that it was more of a converter that took some shortcode name and the (positional or named) variables passed into the process, and replaced that shortcode call into the resulting output based on whatever was in the Shortcode file.

I am a bit surprised that more people haven't run into this yet.  The Sawmill method makes page building so easy for people.  Setting up design components (blocks) of the page and then letting the editor build a page from those components without requiring a developer to build a special layout for every possible combination of blocks the editor might want.

If you want to see the project that I am working on and where I have run into this issue, I'd be happy to show you.

zivbk1:
I would like to have something in my templates like this to process the string and get what I want out of it. `{{ .Section.biography | shortcodify | markdownify }}`

The shortcodes in the string would be rendered and then the output string of that function would pipe into the markdownify function to convert any markdown in the string to HTML.

That's the hopes and dreams I have at the moment. 😄

[...]

PR #8029, opened 2020-12-06
hugolib: Add .ProcessedContent page variable

blattersturm:
Add a page variable similar to .RawContent which returns the page as if it were a HTML file originally, that is, not passing it through the content converters.

We needed this functionality for rendering Hugo pages in an external SPA frontend with a normal Markdown processor on said frontend, and it suffices for this use case.

While preparing this PR I did stumble across #7297, which proposes a different interface for a similar concern, but which would have additional complexity as the shortcode state seems to be directly tied to the page input currently.

### DIFF
--- a/docs/content/en/variables/page.md
+++ b/docs/content/en/variables/page.md
@@ -120,6 +120,9 @@ See also `.ExpiryDate`, `.Date`, `.PublishDate`, and [`.GitInfo`][gitinfo].
 .PlainWords
 : the Page content stripped of HTML as a `[]string` using Go's [`strings.Fields`](https://golang.org/pkg/strings/#Fields) to split `.Plain` into a slice.
 
+.ProcessedContent
+: the content, before being converted to HTML, but after processing shortcodes.
+
 .Prev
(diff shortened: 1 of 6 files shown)


blattersturm:
I see CI failed, will investigate soon. Worked locally on Windows.

blattersturm:
Hi, did I do anything obviously wrong here? I see this is still not looked at 😕

bep:
Thanks for this,

As this is a non-trivial change, it needs to reference an open [bug or enhancement](https://github.com/gohugoio/hugo/issues) before we can review it to be merge into Hugo.

blattersturm:
I'm unsure about the implications of that - does that mean that since the other issue specified a different implementation for a similar concern, it has become a formal specification?

bep:
> it has become a formal specification?

Just create a new issue describing what you want/need. It's not ideal to discuss an implementation before we even know if/what to add. I'm closing this for now.

jirutka:
> Just create a new issue describing what you want/need.

There are already three issues describing the needed functionality and suggesting a solution, one of them referenced in the PR description: #7297, another ones are #6094 and #6703. Just what I found, there are possibly more. Moreover, you’re the author of the first one!

So what else do you need?

PR #8044, opened 2020-12-12
fix: typo in lib

mayocream:


bep:
I'm closing this. I would appreciate that in the future combine these spelling mistakes into one PR/commit. As it is now, I need to review/edit commit message etc. for very minute changes, which is not very cost effective.

PR #8061, opened 2020-12-17
tpl: Added configuration options for 'TableOfContents' generation

kasisoft:
This commit adds the following options to control the outcome of the 'TableOfContents' rendering:

* writeLevels: Add 'data-level' attributes to '<ul/ol>' elements which allows better control for the styling.
* wrapperElement: Allows to use alternative elements rather than '<nav>'
* wrapperId: Allows to use an alternative ID rather than 'TableOfContents'
* wrapperClass: Allows to attach a CSS class to the wrapperElement

The default behavior remains unchanged, so if none of these attributes is being used nothing will be changed.


kasisoft:
FYI: I signed the CLA agreement multiple times and the CLAssistant showed be the window that this succeeded. However this badge doesn't seem to change so I'm stuck here.

moorereason:
The CLA issue is that your commits and your PR are made with two different accounts.  Signing the CLA with the commit user should fix it.

kasisoft:
@moorereason: Thanks for this hint (didn't know that)

PR #8069, opened 2020-12-17
tpl: Added configuration options to influence 'TableOfContents' generation

kasisoft:
This commit adds the following options to control the outcome of the 'TableOfContents' rendering:

* writeLevels: Add 'data-level' attributes to '<ul/ol>' elements which allows better control for the styling.
* wrapperElement: Allows to use alternative elements rather than '<nav>'
* wrapperId: Allows to use an alternative ID rather than 'TableOfContents'
* wrapperClass: Allows to attach a CSS class to the wrapperElement

The default behavior remains unchanged, so if none of these attributes is being used nothing will be changed.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -3,3 +3,5 @@
 ### Reporting a Vulnerability
 
 Please report (suspected) security vulnerabilities to **[bjorn.erik.pedersen@gmail.com](mailto:bjorn.erik.pedersen@gmail.com)**. You will receive a response from us within 48 hours. If we can confirm the issue, we will release a patch as soon as possible depending on the complexity of the issue but historically within days.
+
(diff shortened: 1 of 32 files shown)


kasisoft:
Can someone help me out and explain to me why my builds are failing?
The travis builds mention that there are some improperly formatted files but I followed the contribution guideline and executed the command "mage -v check" which didn't put out any error message. Apart from that I've gone through my changes again but didn't spot anything suspicious.
It would be helpful to know which file is supposedly improperly formatted.
Thanks in advance.

moorereason:
Looks like you need to use gofmt to format any Go files you edit.

kasisoft:
Thanks again @moorereason for the help ;-) 
I will add this info to the general developer contribution documentation later on.

Issue #8091, opened 2020-12-24
Fix RelURL (and RelLangURL)


jbar:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
Hugo Static Site Generator v0.80.0-DEV linux/amd64 BuildDate: 24-12-2020
</pre>

### Does this issue reproduce with the latest release?

I Just need an issue number for the pull request.

### What is the bug ?

RelURL did nothing if URL begin by http (which may aims a local page and not an absolute URL).
RelURL did unwanted things when input was indeed an absolute URL, but not begining with "http".

jbar:
cf: #8092

jmooring:
Please provide additional details. I do not understand the problem. The examples here are correct:
<https://gohugo.io/functions/relurl/#readout>

Please provide specific examples of what is not working. I am continuing the discussion here instead of the [forum](https://discourse.gohugo.io/) because you have an open PR.

jmooring:
@jbar I am closing this issue and the associated pull request (https://github.com/gohugoio/hugo/pull/8092) due to lack of information. See my previous comment for details.

PR #8092, opened 2020-12-24
#8091: Fix RelURL (and RelLangURL)

jbar:
RelURL did nothing if URL begin by http (which may aims a local page and not an absolute URL).
RelURL did unwanted things when input was indeed an absolute URL, but not begining with "http".


jbar:
Didn't know an issue was automatically create whit pull requests. Good ! (it will save a little time for the next ones)

jmooring:
@jbar I am closing this pull request and the associated issue (#8091) due to lack of information.

PR #8093, opened 2020-12-24
SafeURL: considerer  "tel:" URI sheme as safe

jbar:
Since "mailto:" is considered as safe, then also should be "tel:".
BTW, the existence of this function does not make me very comfortable.

### DIFF
--- a/docs/content/en/functions/safeURL.md
+++ b/docs/content/en/functions/safeURL.md
@@ -20,7 +20,7 @@ aliases: []
 
 `safeURL` declares the provided string as a "safe" URL or URL substring (see [RFC 3986][]). A URL like `javascript:checkThatFormNotEditedBeforeLeavingPage()` from a trusted source should go in the page, but by default dynamic `javascript:` URLs are filtered out since they are a frequently exploited injection vector.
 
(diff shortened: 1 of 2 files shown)


davidsneighbour:
> BTW, the existence of this function does not make me very comfortable.

Maybe you think that everything without this function might be unsafe? That's not the case. The safe in these functions just means, "safe for printing without encoding". And the theme developer is the one adding it. So... apart from that, read this: https://gohugo.io/about/security-model/

+1 for having tel:+1234567890 accepted in safeHTML. If that is not already the case. Maybe you are having an issue with arbitrary signage like dots or dashes?

jbar:
> Maybe you think that everything without this function might be unsafe?

Not at all. What make me uncomfortable is more something related with its name, its place in the overall design, and about its hard coded whitelist.

I am sorry, that is yet just a feeling, and I don't have time to mind more about this. So I won't be able yet to find word to explain and determine if there is something pertinent to improve, or no.

bep:
You need to take this upstream in the Go project. We do maintain a fork here, but I'm not making changes of this type.

Issue #3983, opened 2017-10-20
Make GroupByParamDate work with string params


clementbalestrat:
I'm trying to group some of my articles by date using the ``project_date`` param defined in the front matter, but can't make it work. I end up with a ``runtime error: index out of range
`` error in the console.

According to the documentation, it should look like this:

```{{ range .Data.Pages.GroupByParamDate "project_date" "2006" "desc"}}```

Any thoughts?

Update: I was using version 0.26 and upgraded to 0.30.2 and still experiencing the issue.

MadeByMike:
This is because the function does not convert the a timestamp string into a `time.Time` structure: https://gohugo.io/functions/time/

In my testing when when I `printf` the value of the parameters I got:

date: time.Time{wall:0x0, ext:63642202692, loc:(*time.Location)(nil)}
expirydate: time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}
lastmod: time.Time{wall:0x0, ext:63642202692, loc:(*time.Location)(nil)}
project_date: "2017-09-28T13:38:12Z"
publishdate: time.Time{wall:0x0, ext:63642202692, loc:(*time.Location)(nil)}

I found `GroupByParamDate` worked with all of the parameter that were `time.Time` format above.

timn:
I'm seeing the same problem with 0.32.4. Is there a workaround for this (other than not using GroupByParamDate)?

MadeByMike:
@timn For my project I remapped the field in the CMS to the `date` param which for some reason does get converted to a `time.Time` format by Hugo. Luckily for my purposes I didn't need the actually date field for anything else for I could override it with my param.

Not an awesome solution and no other workaround that I am aware of. 

I'd love to fix this. I was able to locate the problem but I don't know enough Go to get started.

MadeByMike:
This is still a bug. It's also not possible to sort a collection of a page resources and normal pages by a custom date field for the same reasons.

MadeByMike:
This is still a bug :)

moorereason:
First, I'm unable to reproduce the `runtime error: index out of range` error today.

Second, Hugo treats a few(?) reserved page parameter and forces them to a `time.Time` type.  All other date parsing and conversion is left to the front matter parsers.

**JSON** does _not_ support timestamps.  They're all treated as strings.  Don't use JSON.

[...]

PR #8098, opened 2020-12-28
GroupByParamDate not supporting datetimes

zerok:
My goal was to sort my readinglist per year. Here I have three date-params:

- date (just the usual page attribute)
- started (when I started reading the book)
- finished (when I finished the book)

Now I wanted the grouping to happen for the `finished` field in a way similar to how GoodReads does it. So I tried the following code:

```go
{{ range .Pages.GroupByParamDate "finished" "2006" }}
```

Sadly, this resulted in only a single group with the key `0001`. This kind of started smelling like an issue with the date format so I went digging.

Turns out that the `Pages.GroupByParamDate` method converts the parameter to a lower-case representation and tries to convert that into a time.Time if it isn't already of that type. That fails for things like RFC3339-formatted dates because the `T` separator is now also a `t` and therefore no longer matches.

This was never an issue simply because before that method didn't support string parameters and therefore never did any actual type conversion before.

This is basically a follow-up to https://github.com/gohugoio/hugo/issues/3983.

### DIFF
--- a/resources/page/pagegroup.go
+++ b/resources/page/pagegroup.go
@@ -316,7 +316,7 @@ func (p Pages) GroupByParamDate(key string, format string, order ...string) (Pag
 		var r Pages
 
 		for _, p := range pages {
-			param := resource.GetParamToLower(p, key)
+			param := resource.GetParam(p, key)
 			var t time.Time
 
 			if param != nil {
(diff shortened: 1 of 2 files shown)


zerok:
@bep w00t, thanks 🥳

PR #8115, opened 2021-01-04
tpl: Do not return errors in substr for out-of-bounds cases

### LINKED ISSUE #8113: `substr` function gets error if `START` equals to string's length

<pre>
$ hugo version
Hugo Static Site Generator v0.80.0/extended linux/amd64 BuildDate: unknown
</pre>

Not tested.

### Issue
Hi, all.

After upgrading to version v0.80.0, there is an error:
```
start position out of bounds for x-byte string
```

I expected that `{{ substr "foo" 3 }}` gives me empty string, and this always works until v0.80.0.

I believe commit 64789fb5dcf8326f14f13d69a2576ae3aa2bbbaa caused this problem. It was
```golang
if start > len(asRunes) {
	return "", fmt.Errorf("start position out of bounds for %d-byte string", len(aStr))
```
before, and now:
```golang
if start > rlen-1 {
	return "", fmt.Errorf("start position out of bounds for %d-byte string", rlen)
}
```

Can removing "-1" fix this?

Thank you in advance, and happy new year!

moorereason:
Most other substr implementations don't error out in out-of-bounds cases
but simply return an empty string (or a value that's printed as an empty
string). We'll follow their lead and not exit template execution.  Allow
the user decide what to do with the empty result.

Fixes #8113

### DIFF
--- a/tpl/strings/strings.go
+++ b/tpl/strings/strings.go
@@ -16,7 +16,6 @@ package strings
 
 import (
 	"errors"
-	"fmt"
 	"html/template"
 	"strings"
 	"unicode/utf8"
@@ -322,7 +321,7 @@ func (ns *Namespace) Substr(a interface{}, nums ...interface{}) (string, error) 
 	}
 
 	if start > rlen-1 {
-		return "", fmt.Errorf("start position out of bounds for %d-byte string", rlen)
+		return "", nil
 	}
 
 	end := rlen
@@ -337,7 +336,7 @@ func (ns *Namespace) Substr(a interface{}, nums ...interface{}) (string, error) 
 	}
 
 	if start >= end {
-		return "", fmt.Errorf("calculated start position greater than end position: %d > %d", start, end)
+		return "", nil
 	}
 
 	if end < 0 {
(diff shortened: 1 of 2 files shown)


jmooring (commented):
Should the Hugo `substr` function behave identically to PHP's `substr` function in every case? If so, there are still some discrepancies. My comparison testing is a practical application of PHP's `substr` function where the result is automatically cast to string (i.e., FALSE -> ""). Please advise. Thanks.

moorereason:
> Should the Hugo substr function behave identically to PHP's substr function in every case?

Not necessarily.  I haven't compared PHP/Hugo to Javascript's version of `substr`, but the Javascript version always returns a simple string [from what I can tell](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-string.prototype.substr).

Here's my thinking in making this PR:

PHP's `substr` _does not raise an error_ but instead returns a boolean False.  It's up to the caller to handle that properly.

This PR removes the "returning an error" part and simply returns an empty string.  It's up to the caller to decide if an empty string is a special/error condition.

We have no way of return an error without the template engine aborting the template execution.

To mimic PHP, we may be able to return a special Go type that's a boolean but prints as an empty string for the String interface, but that's getting too complicated for our templates IMHO.  And it may not even work.

Thoughts?

bep:
I agree with @moorereason  -- it makes sense for us to be more lenient here. I even think this would make sense if we had better client side error handling, but I don't think we need to take that debate here.

Issue #7698, opened 2020-09-16
Allow for arbitrary Asciidoctor extensions


gzagatti:
First of all thanks for putting this project together.

I am currently using Asciidoctor to write all the posts for my website. However, I use custom made extensions to render my posts according to my specifications.

Unfortunately, Hugo comes with a hard coded [pre-specificed list of allowed Asciidoctor extensions](https://github.com/gohugoio/hugo/blob/efaed306b1efe5de7587c15f2e802cab3eec2f1d/markup/asciidocext/asciidocext_config/config.go#L39). When Hugo is preparing the Asciidoctor converter [it basically checks whether the passed extension is on the list of allowed extension and if not it simply ignores the extension](https://github.com/gohugoio/hugo/blob/efaed306b1efe5de7587c15f2e802cab3eec2f1d/markup/asciidocext/convert.go#L106).

I have created [a fork that simply removes the check](https://github.com/gohugoio/hugo/compare/master...gzagatti:master) and Hugo is able to use any arbitrary Asciidoctor extension which means I can render my website just by calling `hugo`.

This behavior is quite puzzling to me as I do not see any reason to not allow the user to determine the extensions which he/she wants to use. Of course, It should be the final user's responsibility to ensure that her/his desired extension is compatible with Hugo. I might be missing something here.Would it be possible to turn off this check on allowed extensions? Is there any reason otherwise.

Thanks again for developing this awesome software.

xexio:
This will be *very* helpful. @gzagatti maybe you can add a PR to add this functionality.

gzagatti:
Sure, will do. Just need some additional time.

moorereason:
> This behavior is quite puzzling to me as I do not see any reason to not allow the user to determine the extensions which he/she wants to use.

See #796, esp. starting with https://github.com/gohugoio/hugo/issues/796#issuecomment-188409011 and below.  I doubt we'll merge your PR without safeguards in Hugo first.

gzagatti:
Thanks for considering the PR and pointing to the right discussion. 

[...]

PR #8131, opened 2021-01-11
markup: Allow arbitrary Asciidoc extension in unsafe mode

gzagatti:
This PR implements proposal #7698 which allows for arbitrary Asciidoc extensions in unsafe mode expanding the range of possibilities when using this markup language. 

The proposal allows for arbitrary extension when [`SafeMode`](https://docs.asciidoctor.org/asciidoctor/latest/safe-modes/) option is set to `unsafe`.  Any stricter mode would prevent arbitrary extensions and only those listed in the [configuration](https://github.com/gohugoio/hugo/blob/master/markup/asciidocext/asciidocext_config/config.go#L40) would be allowed.

### DIFF
--- a/docs/content/en/content-management/formats.md
+++ b/docs/content/en/content-management/formats.md
@@ -100,6 +100,8 @@ Below are all the AsciiDoc related settings in Hugo with their default values:
 
 {{< code-toggle config="markup.asciidocExt" />}}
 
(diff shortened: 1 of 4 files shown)


bep:
How is this PR making _untruested_ sites safer to build?

sergio-carlavilla:
By default AsciiDoctor has this security level selected. The only thing this PR does is add something **very** important to the integration with AsciiDoctor, the use of extensions.

If this PR is rejected, another way to use custom extensions should be allowed.

bep:
Looks good to me, agree, @moorereason ?

moorereason (approved):
LGTM

Issue #8132, opened 2021-01-11
Add external source maps to js.Build


richtera:
The current symbol resolver seems to work very well and I have not seen any problems using jsconfig files anymore in the last version. So I think things are working very well, but there are two areas I had fixed in my PR which I would love to create a new PR for. These are external map files and better error reporting.
I was trying to extract these from my jsconfig/tsconfig PR, but I think a lot of things have changed around it. I will put together a new PR for it, but I don't quite understand how a file and it's symbol file is piped through the fingerprint pipeline. I know that this happens properly in sass for example, but I couldn't quite figure out where to look for the code. Also the error reporting of JS doesn't seem to output all of the error information returned from esbuild and will just output the error token without any line or file context. It also skips some kind of import stack to be able to find and repair the error.

bep:
>but I don't quite understand how a file and it's symbol file is piped through the fingerprint pipeline. 

What do you mean by this? What's a "symbol file"?

bep:
OK, I assume you mean "source maps". Yes, I reverted the `external` source map option that was added. If that is needed, there is example code in, e.g. the `dartsass` package.

richtera:
Perfect. That's what I needed. Yes. I will recover the "external" source map support from my PR and the output of ESBuild error messages (currently it seems to only show the main error and no context information to find it unless this was fixed after 79.1) Thanks for the info I will make the PR associated with this issue number.

bep:
https://github.com/gohugoio/hugo/blob/master/resources/resource_transformers/tocss/dartsass/transform.go#L143

richtera:
In order to support es5 targets we might need to support sourceMaps for babel as well and use babel inside of the esbuild pipeline.

richtera:
There is a little bit of a rabbit hole here.

1. I was able to get source maps working for babel
2. It was able to read the inline source maps from esbuild
3. But the output source maps work for some files and not others. I think it's related to requiring larger node dependencies within a particular file, then it will lose context of lines and character offsets within that file.
4. I tried to add a rollup pipeline which also works, but with my current javascript project, rollup is complicated due to the polyfill and resolve configuration. If someone has a good rollup configuration it should work as long as it has `input: "-"` as the input config in the rollup.config.js file.

I have a WIP commit that's mostly working, but I needed to test it on a netlify build. With previous versions I just committed a binary hugo into the repo and then used it to build, but with this particular hugo version I get a
```sh
[...]

PR #8148, opened 2021-01-15
pipes: rollup, babel, esbuild

richtera:
WIP
Hopefully there is a way to make a test binary to validate production creation of source maps. In dev it seems to work as expected. Still needs test scripts.

Added rollup for testing.
Added sourceMaps option for babel. Babel can also read sourcemaps from
  the output of an esbuild if they are "inline", it will then
  convert the file and remap the source maps and then re-export them
  but currently not if they are "external".
Added support for sourcemap = "external" for esbuild.

Ref #8132

### DIFF
--- a/resources/resource_transformers/babel/babel.go
+++ b/resources/resource_transformers/babel/babel.go
@@ -16,7 +16,11 @@ package babel
 import (
 	"bytes"
 	"io"
+	"io/ioutil"
+	"os"
+	"path"
 	"path/filepath"
+	"regexp"
 	"strconv"
 
 	"github.com/cli/safeexec"
@@ -43,6 +47,7 @@ type Options struct {
 	Compact    *bool
 	Verbose    bool
 	NoBabelrc  bool
+	SourceMaps string
 }
 
 func DecodeOptions(m map[string]interface{}) (opts Options, err error) {
@@ -56,6 +61,14 @@ func DecodeOptions(m map[string]interface{}) (opts Options, err error) {
 func (opts Options) toArgs() []string {
 	var args []string
 
+	switch opts.SourceMaps {
+	case "true":
+		args = append(args, "--source-maps")
+	case "external":
+		args = append(args, "--source-maps")
+	case "inline":
+		args = append(args, "--source-maps=inline")
+	}
 	if opts.Minified {
 		args = append(args, "--minified")
 	}
@@ -141,6 +154,8 @@ func (t *babelTransformation) Transform(ctx *resources.ResourceTransformationCtx
 		}
 	}
 
+	ctx.ReplaceOutPathExtension(".js")
+
 	var cmdArgs []string
 
 	if configFile != "" {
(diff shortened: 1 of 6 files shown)


richtera:
Figured out the temporary build and it works. Used a docker to build the hugo bin. Will now work on adding test scripts for the above. @bep Should I make rollup a separate PR?

bep:
I'm not merging Rollup, so you need to remove that from this PR.

richtera:
No problem. It was an experiment because I couldn't get the other stuff working initially. I will remove it and create test for the other two. I marked the commit WIP anyways.

richtera:
Ok this has been tested and has test scripts now and I removed rollup.
My use case is to use js.Build with sourcemap "inline" and pipe it through babel with sourcemap "external" to render correct es5 output and source maps.

resources/resource_transformers/babel/babel.go
```diff
@@ -56,6 +61,14 @@ func DecodeOptions(m map[string]interface{}) (opts Options, err error) {
 func (opts Options) toArgs() []string {
 	var args []string
 
+	switch opts.SourceMaps {
+	case "true":
```
bep, line 65: Drop this. Adding these clever defaults is almost always just adding work (documentation etc.). Take either external or inline.

resources/resource_transformers/babel/babel.go
```diff
@@ -153,13 +168,21 @@ func (t *babelTransformation) Transform(ctx *resources.ResourceTransformationCtx
 	}
 	cmdArgs = append(cmdArgs, "--filename="+ctx.SourcePath)
 
+	compileOutput, err := ioutil.TempFile("", "compileOut-*.js")
```
bep, line 171: Can you add a comment to this section as to "why we're doing it" (link to the issue about the stderr issue). I'm going to look at this in 1 year and think "this does not look optimal".

resources/resource_transformers/babel/babel.go
```diff
@@ -177,6 +200,27 @@ func (t *babelTransformation) Transform(ctx *resources.ResourceTransformationCtx
 		return errors.Wrap(err, errBuf.String())
 	}
 
+	content, err := ioutil.ReadAll(compileOutput)
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(compileOutput.Name() + ".map"); err == nil {
```
bep, line 208: Pull the map filename out into a var (you're creating it 3 times).

bep (changes requested):

resources/resource_transformers/babel/babel.go
richtera: Ok, I added a comment. The reason this was there is that babel doesn't understand "external" really.
.sourceMaps must be a boolean, "inline", "both", or undefined
Added "external" and "inline" as the only options. I didn't think "both" was useful.

resources/resource_transformers/babel/babel.go
richtera: Adding comment.

resources/resource_transformers/babel/babel.go
richtera: Good idea.

richtera:
@bep Just noticed something that's kind of "hidden":

The option is called `SourceMaps` in babel because of `--source-maps` being the preferred command line in babel.
The option is called `SourceMap` in `js.Build` because we decided to make it so.
 I can rename the babel one to SourceMap or support if you'd like.

resources/resource_transformers/babel/babel.go
richtera: @bep NOTE the argument used to be called SourceMaps in this commit, because babel's command line argument is --source-maps and not --source-map. It makes sense, to change it, but this generates no warning or error causing source maps all of a sudden not to work. Not sure what the best thing to do would be, I'll change my references to sourceMaps.

Issue #8110, opened 2021-01-02
Hugo compiles non deterministic output when using custom output format


noreff:
### What version of Hugo are you using (`hugo version`)?
<pre>
Hugo Static Site Generator v0.80.0-792EF0F4 darwin/amd64 BuildDate: 2020-12-31T13:36:47Z
</pre>

### Does this issue reproduce with the latest release?
Yes

### Details

I'm using Hugo to create php files. And it was working just fine, but starting from version 0.63.0 Hugo start occasionally producing different output when running `hugo` (without changing any files at all). Version 0.62.2 works without any issues.

See https://github.com/noreff/hugo-non-deterministic-output-bug/raw/master/hugo_bug.gif

I created a minimal reproducible example and upload it to github - https://github.com/noreff/hugo-non-deterministic-output-bug.

This starts happening in this release - https://github.com/gohugoio/hugo/releases/tag/v0.63.0.
More specific - after this commit - https://github.com/gohugoio/hugo/commit/c6d650c8c8b22fdc7ddedc1e42a3ca698e1390d6 (tpl/tplimpl: Rework template management to get rid of concurrency issues) which is quite huge. I tried to find the reason myself, but didn't succeed yet. I will be very grateful for any help.

moorereason:
@noreff,
I removed the link to the 9MB gif in the OP.  It's too big and not that helpful anyway.

@bep,
I can replicate the issue with Hugo master.  Periodically, Hugo renders foo's content for the bar page. I've [submitted a PR](https://github.com/noreff/hugo-non-deterministic-output-bug/pull/1) to the demo repo to trim it down to the bare minimum ([my branch](https://github.com/moorereason/hugo-non-deterministic-output-bug/tree/simplify)).  It looks like the bug is in the baseof functionality.  If I remove the blocks from `layouts/bar/single.php`, I haven't seen it generate the wrong content after 5000 iterations.  With blocks, it will generate the wrong content in under 500 iterations or so.  In my PR/branch, I've included a `run.sh` script to build the site repeatedly until the sha1sum for `public/bar/index.php` is different.

moorereason:
I can confirm that the issue appears to start with https://github.com/gohugoio/hugo/commit/c6d650c8c8b22fdc7ddedc1e42a3ca698e1390d6.

Also, I'm running go1.16beta1 linux/amd64.

moorereason:
[...]

PR #8149, opened 2021-01-17
tpl: Fix race condition in text template baseof

moorereason:
Copy most of the htmltemplate cloning to the textemplate implementation
in the same function.

### DIFF
--- a/tpl/tplimpl/template.go
+++ b/tpl/tplimpl/template.go
@@ -598,10 +598,16 @@ func (t *templateHandler) applyBaseTemplate(overlay, base templateInfo) (tpl.Tem
 			}
 		}
 
-		templ, err = templ.Parse(overlay.template)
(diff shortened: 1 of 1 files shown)


moorereason:
@bep,
Updates #8110.

I created this draft PR to point to the problem area (which you're way more familiar with than I am).  Two things that need more attention:

1. It would be nice to have tests for this PR.  Do we have explicit tests for this race in the HTML engine?  If so, I can't find them.
2. I left the extra lookup workaround section commented out because it didn't break the best repo for #8110.  I'm not sure if that line is necessary, so I left it out until you could take a look at it.

bep:
1. If we had such a test, we would not have a race (as the test would have told us to fix it). But these tests are rather hard to create -- I'll merge this as is.
2. The extra lookup was needed in some older Go version (do not remember which)

PR #8169, opened 2021-01-22
commands: add maxReadFileSize command line argument.

avdva:
This allows to override the default file size limit for readFile.
More context here: https://discourse.gohugo.io/t/error-calling-readfile-file-is-too-big/25460.


moorereason:
I wonder why we have that limit in the first place.  I'm more interested in removing the limit than adding another config option.

bep:
Yea, let's just remove it. We don't do checks on other files.

PR #8172, opened 2021-01-22
tpl/os: remove 1mb limit for readFile.

avdva:
Instead of [adding an option](https://github.com/gohugoio/hugo/pull/8169) that would allow to change the limit, we just remove this limit. 
More context here: https://discourse.gohugo.io/t/error-calling-readfile-file-is-too-big/25460.
Will need to update the docs too.

### DIFF
--- a/tpl/os/os.go
+++ b/tpl/os/os.go
@@ -18,7 +18,6 @@ package os
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


tsal:
THANK YOU!!

balibebas:
w00t

PR #8183, opened 2021-01-28
List headless pages

Jaza:
Expose a `HeadlessPages` collection, so that it's possible to to list headless pages, filter them, count them, paginate them, etc, without any template hackery needed.


Jaza:
After all this time, I finally realised that this is already possible in Hugo, see https://gohugo.io/content-management/build-options/ , can do it for content that has `render = "never"` and `list = "always"` (or `list = "local"`).

PR #8186, opened 2021-01-29
Fallback on .Date if .PublishDate not set

djatwood:
Fallback on `.Date` like in [opengraph template](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/opengraph.html)

### DIFF
--- a/tpl/tplimpl/embedded/templates/schema.html
+++ b/tpl/tplimpl/embedded/templates/schema.html
@@ -2,7 +2,8 @@
 <meta itemprop="description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}">
 
 {{- if .IsPage }}{{ $ISO8601 := "2006-01-02T15:04:05-07:00" }}{{ if not .PublishDate.IsZero }}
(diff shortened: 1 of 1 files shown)


bep:
Hey, thanks for this.

These templates has a history and we may have forgotten to update them as Hugo has evolved, but in my head `publishDate` will (in its default config) always be set if there is a date:

https://gohugo.io/getting-started/configuration/#configure-dates

djatwood:
Maybe I need to take a look at all of the internal templates and update them then. The date attribute is what is checked when adding updated time in the open graph template. I am a bit confused why there is Published Date, Date, and Lastmod. Is Published Date the "correct" parameter here?
```html
{{- if not .Date.IsZero }}<meta property="og:updated_time" {{ .Lastmod.Format $iso8601 | printf "content=%q" | safeHTMLAttr }} />
```

bep:
Yes, those templates was built before we improved the date logic in Hugo.

```html
{{- if not .Date.IsZero }}<meta property="og:updated_time" {{ .Lastmod.Format $iso8601 | printf "content=%q" | safeHTMLAttr }} />
```

So, if you found the above in a template, that obviously isn't right.

I think we can say that, in general:

* If you want the _published date_, use `PublishDate` -- no need to check if it's set.
* If you want the _last modified date_, use `Lastmod` -- no need to check if it's set.
* If you want _a date_, use `Date` -- no need to check if it's set.

PR #8220, opened 2021-02-09
Add 'linktitle' option to figure shortcode

patdavid:
If a link is included and the user wants to include the `title` attribute on the anchor, this will parse it into the `<a>` element.


patdavid:
This contribution fits with the code contribution description.

PR #8222, opened 2021-02-09
Change version string format, and add VendorInfo to help with issue triaging

anthonyfok:
This PR is intended to add an optional "vendorInfo" string to help pinpoint the exact origin of a Hugo binary when `hugo version` is reported in a GitHub issue.  For example:

```
Hugo Static Site Generator v0.80.0/extended linux/amd64 BuildDate: 2021-02-09T18:47:48Z (debian 0.80.0-6)
```

which may help determine whether an issue is generic (Hugo upstream) or vendor-specific.

Example snippet from debian/rules:

```
include /etc/os-release
PACKAGE := github.com/gohugoio/hugo
VENDOR_INFO := $(ID) $(shell dpkg-parsechangelog --show-field Version)

all:
	go build -ldflags '-X "$(PACKAGE)/common/hugo.vendorInfo=$(VENDOR_INFO)"'
```

See https://salsa.debian.org/go-team/packages/hugo/-/blob/debian/sid/debian/rules for a full example.

I have just started using this in the hugo 0.80.0-6 Debian package.

Would other package maintainers be interested to add support for this too?  :-)

### DIFF
--- a/common/hugo/hugo.go
+++ b/common/hugo/hugo.go
@@ -40,6 +40,9 @@ var (
 
 	// buildDate contains the date of the current build.
 	buildDate string
+
+	// vendorInfo contains vendor notes about the current build.
+	vendorInfo string
 )
 
 // Info contains information about the current Hugo environment
--- a/common/hugo/version.go
+++ b/common/hugo/version.go
@@ -127,14 +127,15 @@ func (v Version) NextPatchLevel(level int) Version {
 // BuildVersionString creates a version string. This is what you see when
 // running "hugo version".
 func BuildVersionString() string {
-	program := "Hugo Static Site Generator"
+	// program := "Hugo Static Site Generator"
+	program := "hugo"
 
 	version := "v" + CurrentVersion.String()
 	if commitHash != "" {
 		version += "-" + strings.ToUpper(commitHash)
 	}
 	if IsExtended {
-		version += "/extended"
+		version += "+extended"
 	}
 
 	osArch := runtime.GOOS + "/" + runtime.GOARCH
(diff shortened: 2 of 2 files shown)


bep:
>Would other package maintainers be interested to add support for this too? :-)

It would certainly be useful, but we can take that step-by-step. But do we have some documentation for these build flags somewhere that we could put this info in?

moorereason:
How open are we to changing this version string?  We might need a separate issue to discuss this, but here's my ideal version string:

```
hugo v0.81.0-DEV-4D2B6FC4+extended linux/amd64 BuildDate=2021-02-09T18:47:48Z
hugo v0.80.0+extended linux/amd64 BuildDate=2021-02-09T18:47:48Z VendorInfo=debian:0.80.0-6
```
1. shorten program name
2. make version string semver compliant with `+extended`.  The slash is not allowed in semver.
3. Use key=value pairs for BuildDate and VendorInfo

anthonyfok:
> > Would other package maintainers be interested to add support for this too? :-)
> It would certainly be useful, but we can take that step-by-step. But do we have some documentation for these build flags somewhere that we could put this info in?

@bep Good point!  While a packager can look into magefile.go (and maybe .github/workflows/test.yml) for commitHash and buildDate, there is currently nowhere to look for vendorInfo.  Yes, it would be nice to have a place documentation especially for package maintainers, wouldn't it?  :-)

> `hugo v0.81.0-DEV-4D2B6FC4+extended linux/amd64 BuildDate=2021-02-09T18:47:48Z`
> `hugo v0.80.0+extended linux/amd64 BuildDate=2021-02-09T18:47:48Z VendorInfo=debian:0.80.0-6`

@moorereason I quite like your proposed change to the version string actually!  The *key*=*value* pairs idea does make things look cleaner!

/cc @eclipseo

PR #8234, opened 2021-02-14
Fix: wrong space for humanize func in some case

### LINKED ISSUE #7912: Humanize add spaces between hastags letters

<pre>
$ hugo version
Hugo Static Site Generator v0.77.0/extended darwin/amd64 BuildDate: unknown
</pre>

Yes

I use `humanize` function to make my title uppercase (since title function does not work for my language where only the first letter of the title is uppercase). When text which I am humanizing contains a hashtag and is uppercase, Hugo splits letters after hashtag sign with spaces. Example:
```
{{humanize "this is a #TITLE"}} → "This is a # t i t l e"
```  
I know that humanize is used for changing URLs to human-readable text but even in cases of URLs, I do not know why it should be useful. I think it is a bug.

The bonus would be to add an another `titleCaseStyle` where the only first letter of the text is upper case, this will support title styles for other languages which have the only first letter of the title upper.

susiwen8:
Close #7912 

This issue was caused by [`markbates/inflect`](https://github.com/markbates/inflect) which is deprecated now. use [github.com/gobuffalo/flect](https://github.com/gobuffalo/flect) instead which is recommended solve this issue and won't break `hugo`.

### DIFF
--- a/tpl/inflect/inflect.go
+++ b/tpl/inflect/inflect.go
@@ -16,8 +16,9 @@ package inflect
 
 import (
 	"strconv"
+	"strings"
 
-	_inflect "github.com/markbates/inflect"
+	_inflect "github.com/gobuffalo/flect"
 	"github.com/spf13/cast"
 )
 
@@ -53,7 +54,8 @@ func (ns *Namespace) Humanize(in interface{}) (string, error) {
 		return _inflect.Ordinalize(word), nil
 	}
 
-	return _inflect.Humanize(word), nil
+	str := _inflect.Humanize(word)
+	return _inflect.Humanize(strings.ToLower(str)), nil
 }
 
 // Pluralize returns the plural form of a single word.
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/getkin/kin-openapi v0.32.0
 	github.com/ghodss/yaml v1.0.0
+	github.com/gobuffalo/flect v0.2.2
 	github.com/gobwas/glob v0.2.3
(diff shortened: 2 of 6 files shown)


tpl/inflect/inflect.go
```diff
@@ -53,7 +54,8 @@ func (ns *Namespace) Humanize(in interface{}) (string, error) {
 		return _inflect.Ordinalize(word), nil
 	}
 
-	return _inflect.Humanize(word), nil
+	str := _inflect.Humanize(word)
+	return _inflect.Humanize(strings.ToLower(str)), nil
```
bep, line 58: I understand the ToLower, but why the double Humanize?

susiwen8: In most case `buffalo/reflect` capitalize the first character. At first, I was trying to use `ToUpper` for the first character, and `ToLower` for the rest. 
```go
return str[:1] + strings.ToLower(str[1:]), nil
```
but this would break this test case.
> {ns.Humanize, "óbito", "Óbito"},

So I decide to lowercase the whole string first then use `Humanize ` to capitalize the first character

bep: Sure, but why doing it twice, couldn't you just:


```go
return _inflect.Humanize(strings.ToLower(word)), nil
```

susiwen8: No,  that would break those case

> {ns.Humanize, "MyCamel", "My camel"},

for which result is `Mycamel` rather than `My camel`

`word` is raw, unprocessed input string, so it should get `Humanize ` before `ToLower`. Because of inconsistency between ` markbates/inflect` and `buffalo/reflect`. Second `Humanize` is more like a workaround to keep `hugo` unchanged.

PR #8253, opened 2021-02-16
Fix typo in multilingual.md

tomkonidas:
While going through the docs, I realized that the wrong text was being printed.

I believe it was supposed to be printed without the `s`


coliff:
This is a duplicate of https://github.com/gohugoio/hugoDocs/pull/1228
Changes to the Docs should be made in that repo. I'm unsure if the issue is valid or not.
Have you tested it? If so, please chime in with your feedback on the PR over there.

tomkonidas:
@coliff Thanks will do!

PR #8256, opened 2021-02-17
Update date logic of opengraph and schema internal templates

djatwood:
This is in response to my previous PR #8186.

* Removed date checking of `PublishDate` and `Lastmod` 
* Merged duplicate `IsPage` blocks in opengraph.html
* Formatted template tags so they're easier to follow. There are a lot of nested blocks and keeping track of what's open and what's closed was extremely tedious. I hope the changes I made makes this easier.

Feel free to cherry pick the date changes and exclude the formatting change.

### DIFF
--- a/tpl/tplimpl/embedded/templates/opengraph.html
+++ b/tpl/tplimpl/embedded/templates/opengraph.html
@@ -2,51 +2,51 @@
 <meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
 <meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}" />
 <meta property="og:url" content="{{ .Permalink }}" />
-{{ with $.Params.images }}{{ range first 6 . -}}
-<meta property="og:image" content="{{ . | absURL }}" />
-{{ end }}{{ else -}}
+
+{{- with $.Params.images -}}
+{{- range first 6 . }}<meta property="og:image" content="{{ . | absURL }}" />{{ end -}}
+{{- else -}}
 {{- $images := $.Resources.ByType "image" -}}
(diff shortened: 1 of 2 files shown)


tpl/tplimpl/embedded/templates/opengraph.html
```diff
@@ -2,56 +2,51 @@
...
-{{ end }}
-{{- if not .Lastmod.IsZero }}<meta property="article:modified_time" {{ .Lastmod.Format $iso8601 | printf "content=%q" | safeHTMLAttr }} />{{ end }}
-{{- else }}
-{{- if not .Date.IsZero }}<meta property="og:updated_time" {{ .Lastmod.Format $iso8601 | printf "content=%q" | safeHTMLAttr }} />
-{{- end }}
-{{- end }}{{/* .IsPage */}}
+{{- $iso8601 := "2006-01-02T15:04:05-07:00" -}}
+<meta property="article:section" content="{{ .Section }}" />
+<meta property="article:published_time" content="{{ .PublishDate.Format $iso8601 }}" />
+<meta property="article:modified_time" content="{{ .Lastmod.Format $iso8601 }}" />
```
bep, line 23: Your template code looks clean and good. There is one thing I think we need to fix, though. I don't remember my exact comment in that other thread, but my main take on this is that

* We don't need to fallback on some other date if PublishDate is not set
* But PublishDate may still be zero (e.g. in cases where there are no date set on a given document)

So, given the construct above, I suggest just wrap it in a with, e.g.

```
{{ with .Lastmod }}<meta property="article:modified_time" content="{{ .Format $iso8601 }}" />{{ end }}
```

bep (changes requested):

tpl/tplimpl/embedded/templates/opengraph.html
djatwood: Just so I'm on the same page. You mean I should individually wrap both the PublishDate and Lastmod?

bep: Yes, assume all dates can be zero.

Issue #8261, opened 2021-02-18
Google Analytics internal templates change


djatwood:
There's `google_analytics.html` and `google_analytics_async.html`. Both templates load asynchronously but the "async" one allows preloading.

> While the Google Analytics tag described above ensures the script will be loaded and executed asynchronously on all browsers, it has the disadvantage of not allowing modern browsers to preload the script.

> The alternative async tag below adds support for preloading, which will provide a small performance boost on modern browsers, but can degrade to synchronous loading and execution on IE 9 and older mobile browsers that do not recognize the async script attribute. Only use this tag configuration if your visitors primarily use modern browsers to access your site. 

[View in docs](https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tag)

What do you think about changing the way we define GA in the config?
```toml
[googleAnalytics]
  id = "UA-PROPERTY_ID"
  preload = true
```

Or at the very least change the name from `google_analytics_async.html` to `google_analytics_preload.html`?

djatwood:
There is no option to not use the async attribute when using v4 gtag.js see #8264

coliff:
I don't think we should rename those older Universal Analytics templates, it could break for users/themes which depend on it. I guess aliases/redirects could be setup, but that complicates things.

It's clear from Google that Universal Analytics will no longer receive updates and they are pushing for all users to update to using GA4, so I suggest we leave these older templates as they are.

Issue #7954, opened 2020-11-13
The internal GA template does not work with the new analytics version


onedrawingperday:
The internal GA analytics template has been reported as non functioning with the [new Google Analytics version](https://support.google.com/analytics/answer/10089681).

Also see: https://discourse.gohugo.io/t/google-analytics-ga4-property-doesnt-work/29362

Related Issues: #7879 #7893

debanjanbasu:
This needs to be addressed soon.

flyinprogrammer:
It also should support CSP: https://developers.google.com/tag-manager/web/csp#enabling_the_google_tag_manager_snippet with a nonce variable

djatwood:
Since analytics.js is not depreciated so I think we should support both UA-PROPERTY_ID<sup>analytics.js</sup> and G-MEASUREMENT_ID<sup>gtag.js</sup>. I'm assuming a simple `hasPrefix . "UA-"` and `hasPrefix . "G-"` should be enough to tell the difference.

 I will have to learn about implementing the same [privacy options](https://gohugo.io/about/hugo-and-gdpr/#googleanalytics) with gtag.

---
Anyone know why the functions are namespaced differently? There's `hasPrefix` and `strings.HasSuffix`.

djatwood:
I forgot to mention my PR #8264

onedrawingperday:
Since https://github.com/gohugoio/hugo/pull/8287 was merged I am closing this.

PR #8264, opened 2021-02-19
Add support for google analytics v4

djatwood:
The Google analytics template doesn't work with Google Analytics v4 (gtag.js) #7954. v3 analytics.js is not depreciated so we should support both using `hasPrefix` to tell the difference.

It should be noted that gtag uses the `async` attribute. Which is more like `google_analytics_async.html`. I'm not sure if we should move the changes to that template instead. The naming of the template is also confusing. See #8261.

Privacy settings `anonymizeIP` and `respectDoNotTrack` are supported while `useSessionStorage` is not. I could only find one [document with instructions for session storage](https://medium.com/swlh/how-to-use-google-tag-manager-and-google-analytics-without-cookies-7d041c73cc76) but in my testing it still set cookies.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,7 +50,9 @@ Hugo has become a fully featured static site generator, so any new functionality
 
 If it is of some complexity, the contributor is expected to maintain and support the new feature in the future (answer questions on the forum, fix any bugs etc.).
 
-It is recommended to open up a discussion on the [Hugo Forum](https://discourse.gohugo.io/) to get feedback on your idea before you begin. If you are submitting a complex feature, create a small design proposal on the [Hugo issue tracker](https://github.com/gohugoio/hugo/issues) before you start.
+It is recommended to open up a discussion on the [Hugo Forum](https://discourse.gohugo.io/) to get feedback on your idea before you begin. 
+
+Any non-trivial code change needs to update an open [issue](https://github.com/gohugoio/hugo/issues). A non-trivial code change without an issue reference with one of the labels `bug` or `enhancement` will not be merged.
 
 Note that we do not accept new features that require [CGO](https://github.com/golang/go/wiki/cgo).
 We have one exception to this rule which is LibSASS.


coliff:
This looks like a good PR! I haven't built/tested it, but glancing over the code it works how I would have expected it to. We'll have to update the Docs too.
We'll have to mention that if a user has a GA4 property ID then they should use the `"_internal/google_analytics.html"` template as the `"_internal/google_analytics_async.html"` won't render the new gtag template you added.

bep:
OK, @djatwood could you add a couple of lines to the relevant docs section below /docs in this repo?

djatwood:
I'm happy to update the docs. I could also add gtag support to `google_analytics_async.html` I just wasn't sure. What do you think about that?

bep:
>I could also add gtag support to google_analytics_async.html I just wasn't sure. What do you think about that?

No, let's drop that, that creates too much stuff to maintain.

I suggest that we mention that in the docs is that "for v4, use that other template".

djatwood:
Alright I believe everything should be ready for review now.

natalianunez:
@djatwood, can you help me? I'm trying to use ga v4. So I'm setting the config .toml file as the docs, the `baseof.html`  it's using the `{{ template "_internal/google_analytics.html" . }}` but still not tracking. May I miss something because I don't realize how to change this variable `doNotTrack` , when I render the site:
```javascript

var doNotTrack = false;
if (!doNotTrack) {
	window.dataLayer = window.dataLayer || [];
	function gtag(){dataLayer.push(arguments);}
	gtag('js', new Date());
	gtag('config', 'G-xxxxxxxxxx', { 'anonymize_ip': false });
}
```

It's always false and doesn't execute the script.

Sorry if I sound stupid, but I'm a noobie in all that stuff.

djatwood:
I don't know if I have enough information. `doNotTrack` needs to be false for tracking to run. It's a double negative. Do you have any errors in your console? Are you using any browser other than chrome? Safari and Firefox will block GA

bfeif:
@natalianunez I'm having the same issue as you. Everything looks fine, and the request goes through with `Status: 200` in the browser. However, still nothing is getting sent to google analytics. Is anybody else still having an issue with this? I raised [the issue on discourse.hugo.io](https://discourse.gohugo.io/t/google-analytics-4-not-working-with-cactus-theme/38880/7) as well.

PR #8304, opened 2021-03-06
tpl: Modify 'Querify' to take lone slice argument

### LINKED ISSUE #6735: Allow `querify` to take a lone slice argument.
`querify` helps build URL queries by passing an even number of arguments which are interpreted a set of `key/value` pair à la `dict`.

### Currently 
It is not possible to dynamically generate a set of `key/value` pair to be passed to `querify`.

### Proposal.

Enable the function to take a lone slice of strings argument which would be treated the same way the current string arguments are.

importhuman:
Querify used 'Dictionary' to add key/value pairs to a map to
build URL queries. Changed to dynamically generate ordered
key/value pairs. Cannot take string slice as key (earlier
possible due to Dictionary).

Fixes #6735


importhuman:
Closing pull request as I had configured the incorrect commit email address.

PR #8305, opened 2021-03-06
tpl: Modify 'Querify' to take lone slice argument

### LINKED ISSUE #6735: Allow `querify` to take a lone slice argument.

importhuman:
Querify used 'Dictionary' to add key/value pairs to a map to
build URL queries. Changed to dynamically generate ordered
key/value pairs. Cannot take string slice as key (earlier
possible due to Dictionary).

Fixes #6735

### DIFF
--- a/tpl/collections/collections_test.go
+++ b/tpl/collections/collections_test.go
@@ -564,9 +564,16 @@ func TestQuerify(t *testing.T) {
 	}{
 		{[]interface{}{"a", "b"}, "a=b"},
 		{[]interface{}{"a", "b", "c", "d", "f", " &"}, `a=b&c=d&f=+%26`},
+		{[]interface{}{[]string{"a", "b"}}, "a=b"},
+		{[]interface{}{[]string{"a", "b", "c", "d", "f", " &"}}, `a=b&c=d&f=+%26`},
+		{[]interface{}{[]interface{}{"x", "y"}}, `x=y`},
+		{[]interface{}{[]interface{}{"x", 5}}, `x=5`},
 		// errors
 		{[]interface{}{5, "b"}, false},
 		{[]interface{}{"a", "b", "c"}, false},
+		{[]interface{}{[]string{"a", "b", "c"}}, false},
+		{[]interface{}{[]string{"a", "b"}, "c"}, false},
+		{[]interface{}{[]interface{}{"c", "d", "e"}}, false},
 	} {
 		errMsg := qt.Commentf("[%d] %v", i, test.params)
 
@@ -579,6 +586,32 @@ func TestQuerify(t *testing.T) {
 
 		c.Assert(err, qt.IsNil, errMsg)
 		c.Assert(result, qt.Equals, test.expect, errMsg)
+	}
+}
+
+func BenchmarkQuerify(b *testing.B) {
+	ns := New(&deps.Deps{})
+	params := []interface{}{"a", "b", "c", "d", "f", " &"}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := ns.Querify(params...)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkQuerifySlice(b *testing.B) {
+	ns := New(&deps.Deps{})
+	params := []string{"a", "b", "c", "d", "f", " &"}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := ns.Querify(params)
+		if err != nil {
+			b.Fatal(err)
+		}
 	}
 }
 
--- a/tpl/collections/init.go
+++ b/tpl/collections/init.go
@@ -122,6 +122,10 @@ func init() {
 					`<a href="https://www.google.com?{{ (querify "q" "test" "page" 3) | safeURL }}">Search</a>`,
 					`<a href="https://www.google.com?page=3&amp;q=test">Search</a>`,
 				},
+				{
+					`{{ slice "foo" 1 "bar" 2 | querify | safeHTML }}`,
+					`bar=2&foo=1`,
+				},
 			},
 		)
 
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -424,13 +424,39 @@ func (ns *Namespace) Last(limit interface{}, seq interface{}) (interface{}, erro
 // Querify encodes the given parameters in URL-encoded form ("bar=baz&foo=quux") sorted by key.
 func (ns *Namespace) Querify(params ...interface{}) (string, error) {
 	qs := url.Values{}
-	vals, err := ns.Dictionary(params...)
-	if err != nil {
-		return "", errors.New("querify keys must be strings")
+
+	if len(params) == 1 {
+		switch v := params[0].(type) {
+		case []string:
+			if len(v)%2 != 0 {
+				return "", errors.New("invalid query")
+			}
+
+			for i := 0; i < len(v); i += 2 {
+				qs.Add(v[i], v[i+1])
+			}
+
+			return qs.Encode(), nil
+
+		case []interface{}:
+			params = v
+
+		default:
+			return "", errors.New("query keys must be strings")
+		}
 	}
 
-	for name, value := range vals {
-		qs.Add(name, fmt.Sprintf("%v", value))
+	if len(params)%2 != 0 {
+		return "", errors.New("invalid query")
+	}
+
+	for i := 0; i < len(params); i += 2 {
+		switch v := params[i].(type) {
+		case string:
+			qs.Add(v, fmt.Sprintf("%v", params[i+1]))
+		default:
+			return "", errors.New("query keys must be strings")
+		}
 	}
 
 	return qs.Encode(), nil


moorereason:
I don't think this will do what you intend.  To resolve the issue, `len(params)` will equal `1` since `params` will be something like `[]interface{ []string{"key","value"} }`.  I would suggest adding a test for the scenario you want to pass--we're almost always require tests anyway before we'll merge.

My hunch is that you could do something like this instead:

```
var (
  vals map[string]interface{}
  err error
)

// check for special case of a single slice param
if len(params) == 1 && reflect.TypeOf(params[0]).Kind() == reflect.Slice {
    vals, err = ns.Dictionary(params[0]...)
} else {
    vals, err = ns.Dictionary(params...)
}
if err != nil {
    // ...
```

Reflection is one of the trickier parts of the Go stdlib, so I may be a little off with the above...but it's probably close.  😄

importhuman:
So if I understand correctly, what I've done so far only solves the issue of key/value pairs being unordered due to map, but I haven't implemented the lone slice argument as was the issue.

Should I add test cases to the original test only, considering it's the same function?

moorereason:
> Should I add test cases to the original test only, considering it's the same function?

Yes

moorereason:
Are you sure there's a problem with key ordering in the current implementation?  If there were, wouldn't our existing tests be failing randomly with indeterminate results?  I'm wondering if you're solving a problem that doesn't exist.

importhuman:
@moorereason Yes I was wondering the same thing about the tests, but as mentioned [here](https://github.com/gohugoio/hugo/issues/6735#issuecomment-572704093) it couldn't be a Go map, so I stayed away from maps and implemented a different solution.

tpl/collections/collections_test.go
```diff
@@ -564,9 +564,14 @@ func TestQuerify(t *testing.T) {
...
 		{[]interface{}{"a", "b"}, "a=b"},
 		{[]interface{}{"a", "b", "c", "d", "f", " &"}, `a=b&c=d&f=+%26`},
+		{[]interface{}{[]string{"a", "b"}}, "a=b"},
+		{[]interface{}{[]string{"a", "b", "c", "d", "f", " &"}}, `a=b&c=d&f=+%26`},
 		// errors
 		{[]interface{}{5, "b"}, false},
 		{[]interface{}{"a", "b", "c"}, false},
+		{[]interface{}{[]string{"a", "b", "c"}}, false},
+		{[]interface{}{[]string{"a", "b"}, "c"}, false},
+		{[]interface{}{[]interface{}{"a", "b"}}, false},
```
moorereason, line 574: This test should pass.

importhuman: By pass do you mean they shouldn't be false tests? If so, here's my understanding and reasoning of the function and why I've written them as false tests:
- The original querify function took an even number of string parameters and returned a URL query. The issue was to allow it to take a lone string slice argument. While `[]string{"a", "b", "c"}` is a string slice, it has an odd number of elements, insufficient for a URL query.
- `[]string{"a", "b"}, "c"` has a string slice and a string argument. String slices aren't supported as keys (and don't need to be, [as discussed here](https://discourse.gohugo.io/t/how-does-the-dictionary-function-in-collections-go-work/31599/3?u=importhuman)). If they are meant to be supported, please let me know what the URL query should look like and I will try to implement it.
- `[]interface{}{"a", "b"}` is an interface slice, while we are only concerned with string slices. String assertion of the interfaces might lead to incorrect URL queries in some cases.

If you meant that this test is failing, then I don't understand where it's going wrong. It's passing the Github checks and I've tested on my system as well. How can I replicate the failed tests?

Please let me know which area needs fixing, as I'm not sure I understood what you meant.

moorereason: Sorry. I only meant that I thought the test on line 574 should be a passing test case. However, upon further reflection, I can't think of a scenario where a template would pass an actual slice of interface values, so let's leave it alone for now.

importhuman: Is there anything I need to do for now?

moorereason:
@importhuman,
I've modified your PR locally with some optimizations, but I'll let you try to beat me if you're up for it.  :smile:  Otherwise, I'll just show you my work (but you'll learn more if you try it yourself).  No shame if you don't have time for this.

I pushed a couple benchmarks to your branch.  To run them, I'd use the following to get a baseline of your PR:

```
$ go test -bench='BenchmarkQuerify.*' -count=5 ./tpl/collections | tee bench.querify
WARNING: calling IsSet with unsupported type "struct" (time.Time) will always return false.

WARNING: calling IsSet with unsupported type "invalid" (<nil>) will always return false.

goos: linux
goarch: amd64
pkg: github.com/gohugoio/hugo/tpl/collections
cpu: Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz
BenchmarkQuerify-2                643951              1975 ns/op             464 B/op         18 allocs/op
BenchmarkQuerify-2                689896              1837 ns/op             464 B/op         18 allocs/op
BenchmarkQuerify-2                696940              1833 ns/op             464 B/op         18 allocs/op
BenchmarkQuerify-2                748134              2201 ns/op             464 B/op         18 allocs/op
BenchmarkQuerify-2                738346              1846 ns/op             464 B/op         18 allocs/op
BenchmarkQuerifySlice-2           454816              2263 ns/op             552 B/op         23 allocs/op
BenchmarkQuerifySlice-2           473956              2322 ns/op             552 B/op         23 allocs/op
BenchmarkQuerifySlice-2           615118              2245 ns/op             552 B/op         23 allocs/op
BenchmarkQuerifySlice-2           568017              2278 ns/op             552 B/op         23 allocs/op
BenchmarkQuerifySlice-2           441758              2330 ns/op             552 B/op         23 allocs/op
PASS
ok      github.com/gohugoio/hugo/tpl/collections        13.997s
```
Run it again after making changes, write the output to another file, and then use [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) to compare results.  Here's your version compared to mine:

```
$ benchstat bench.querify worktrees/iss6735-querify/bench.querify
name            old time/op    new time/op    delta
Querify-2         1.94µs ±14%    1.02µs ±17%  -47.55%  (p=0.008 n=5+5)
QuerifySlice-2    2.29µs ± 2%    0.91µs ±17%  -60.40%  (p=0.008 n=5+5)

name            old alloc/op   new alloc/op   delta
Querify-2           464B ± 0%      152B ± 0%  -67.24%  (p=0.008 n=5+5)
QuerifySlice-2      552B ± 0%      176B ± 0%  -68.12%  (p=0.008 n=5+5)

name            old allocs/op  new allocs/op  delta
Querify-2           18.0 ± 0%       9.0 ± 0%  -50.00%  (p=0.008 n=5+5)
QuerifySlice-2      23.0 ± 0%       9.0 ± 0%  -60.87%  (p=0.008 n=5+5)
```
Focus your efforts on reducing memory allocations.  The time results may not be super-accurate since I'm running these on a VM on my laptop that will mess with the CPU clock speed as it sees fit (but they're useful nonetheless).

importhuman:
@moorereason I'd like to try it myself, but I didn't understand this step:
> Run it again after making changes, write the output to another file
What exactly am I writing to another file, and where (as in the branch, etcetera). Also, I haven't read the Go spec yet, should I refer to that or some other resource for help on what will help memory allocations? I'm guessing I'll also have to study up a bit on the O(n) notation stuff?

moorereason:
Man, my language is imprecise. I mean tee the output to another file on the command line. benchstat just needs two files with benchmark results to compare. 

O(n)? No, it's not that complicated really. Hint: adding items to pairedSlice requires an allocation. Is it necessary?

importhuman:
@moorereason This is all I've managed:

```
~/s/hugo ❯❯❯ benchstat bench.querify bench2.querify
name            old time/op    new time/op    delta
Querify-4         3.25µs ± 3%    2.94µs ± 3%   -9.47%  (p=0.008 n=5+5)
QuerifySlice-4    3.96µs ± 2%    3.75µs ± 1%   -5.33%  (p=0.008 n=5+5)

name            old alloc/op   new alloc/op   delta
Querify-4           464B ± 0%      440B ± 0%   -5.17%  (p=0.008 n=5+5)
QuerifySlice-4      552B ± 0%      528B ± 0%   -4.35%  (p=0.008 n=5+5)

name            old allocs/op  new allocs/op  delta
Querify-4           18.0 ± 0%      16.0 ± 0%  -11.11%  (p=0.008 n=5+5)
QuerifySlice-4      23.0 ± 0%      21.0 ± 0%   -8.70%  (p=0.008 n=5+5)
```

What I've done is remove the allocation for adding items to pairedSlice, and define pairedSlice with specific length in both the cases like so:
```go
if len(params) == 1 && fmt.Sprintf("%v", reflect.TypeOf(params[0])) == "[]string" {
		pairedSlice := make([][]interface{}, 0, len(params[0].([]string)))
		if len(params[0].([]string))%2 != 0 {
			return "", errors.New("invalid query")
		}
		for i := 0; i < len(params[0].([]string)); i += 2 {
			pairedSlice = append(pairedSlice, []interface{}{params[0].([]string)[i], params[0].([]string)[i+1]})
		}
		for _, vals := range pairedSlice {
			qs.Add(fmt.Sprintf("%v", vals[0]), fmt.Sprintf("%v", vals[1]))
		}
	}
```
And similarly for the other case. I did try using the Dictionary function in one case (had to disable one of the false tests), but that didn't make too much of a difference except giving a positive delta in one case.

moorereason:
Another hint:  refactor the entire function without using `pairedSlice`.  In both cases (len==1 and len>1), you already have access to the keys as strings.  Skip the `pairedSlice` step and use `qs.Add` directly in your loops.

tpl/collections/collections_test.go
```diff
@@ -582,6 +587,34 @@ func TestQuerify(t *testing.T) {
...
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkQuerifySlice(b *testing.B) {
+	ns := New(&deps.Deps{})
+	params := []string{"a", "b", "c", "d", "f", " &"}
+
+	b.ReportAllocs()
```
bep, line 608: We dont't need/want ReportAllocs in our benchmarks (you can toggle that on/off when running these.

importhuman:
@moorereason Thanks, quite a notable difference now:
```
name            old time/op    new time/op    delta
Querify-4         3.25µs ± 3%    1.92µs ± 1%  -40.88%  (p=0.008 n=5+5)
QuerifySlice-4    3.96µs ± 2%    2.87µs ± 2%  -27.49%  (p=0.008 n=5+5)

name            old alloc/op   new alloc/op   delta
Querify-4           464B ± 0%      152B ± 0%  -67.24%  (p=0.008 n=5+5)
QuerifySlice-4      552B ± 0%      288B ± 0%  -47.83%  (p=0.008 n=5+5)

name            old allocs/op  new allocs/op  delta
Querify-4           18.0 ± 0%       9.0 ± 0%  -50.00%  (p=0.008 n=5+5)
QuerifySlice-4      23.0 ± 0%      17.0 ± 0%  -26.09%  (p=0.008 n=5+5)
```
Thinking about pointers to further improve performance, but not sure if/where I can use them.

moorereason:
> Thinking about pointers to further improve performance, but not sure if/where I can use them.

I wouldn't bother going down that path.  I don't think there's any gains to be made there.

Based upon your last benchstat output, it looks like you've matched my results for the Querify benchmark.  However, QuerySlice needs work.  You can see in my earlier benchstat results that both benchmarks use 9 allocs.

importhuman:
@moorereason got it down to 10 allocs:

```go
if len(params) == 1 && fmt.Sprintf("%v", reflect.TypeOf(params[0])) == "[]string" {
		if len(params[0].([]string))%2 != 0 {
			return "", errors.New("invalid query")
		}
		for i := 0; i < len(params[0].([]string)); i += 2 {
			qs.Add(params[0].([]string)[i], params[0].([]string)[i+1])
		}
	}
```

I did get to 9 allocs by doing a type switch instead of reflect, but it gives an unexpected success in the last test:
```go
{[]interface{}{[]interface{}{"a", "b"}}, false}
```
I modified it to have an odd number of elements (including numbers, like `{[]interface{}{[]interface{}{"a", "b", 3, 4, 5}}, false}`), but it still gave the same result, so this doesn't seem the right approach.
The relevant code:
```go
if len(params) == 1 {
		switch params[0].(type) {
		case []string:
			if len(params[0].([]string))%2 != 0 {
				return "", errors.New("invalid query")
			}
			for i := 0; i < len(params[0].([]string)); i += 2 {
				qs.Add(params[0].([]string)[i], params[0].([]string)[i+1])
			}
		default:
			break
		}
	}
```

moorereason:
I've been thinking.  The main limitation I see with our current approach is that we don't allow something like `{{ slice "a" 1 | querify }}` since that slice call would result in a single `[]interface{}` param instead of a `[]string`.  This gets back to my earlier assertion that the `[]interface{}{[]interface{}{...}}` test should pass.  I think the behavior and type limitations should be the same regardless of how the parameters are passed.

So, here's my proposed implementation that maintains consistency with the existing behavior:

```go
func (ns *Namespace) Querify(params ...interface{}) (string, error) {
        qs := url.Values{}

        if len(params) == 1 {
                switch v := params[0].(type) {
                case []string:
                        // []string is the simplest case, so avoid the extra overhead of
                        // converting to []interface{}.  Just build the query string here
                        // and return early.

                        if len(v)%2 != 0 {
                                return "", errors.New("invalid query")
                        }

                        for i := 0; i < len(v); i += 2 {
                                qs.Add(v[i], v[i+1])
                        }

                        return qs.Encode(), nil

                case []interface{}:
                        params = v

                default:
                        // we should only get here if there were no string params
                        return "", errors.New("query keys must be strings")
                }
        }

        if len(params)%2 != 0 {
                return "", errors.New("invalid query")
        }

        for i := 0; i < len(params); i += 2 {
                switch v := params[i].(type) {
                case string:
                        qs.Add(v, fmt.Sprintf("%v", params[i+1]))
                default:
                        return "", errors.New("query keys must be strings")
                }
        }

        return qs.Encode(), nil
}
```

importhuman:
@moorereason Since this method allows checking for string keys without using the Dictionary function, I agree with you, and have added the code, alongwith a couple of test cases. I wasn't sure if I was supposed to squash this commit as well as there was yours as well. Please let me know if changes are required.

moorereason (changes requested):
@importhuman,
Looking good.  As @bep mentioned earlier, remove the `b.ReportAllocs()` line from the benchmarks I created and then you can squash the commits.

importhuman:
@moorereason Done! Any suggestions for picking up issues in the future?

moorereason (approved):

moorereason:
@importhuman, great work.  I'll leave it to @bep for final review and merge.

> Any suggestions [...]?

- Check the [GoodFirstIssue](https://github.com/gohugoio/hugo/issues?q=is%3Aissue+is%3Aopen+label%3AGoodFirstIssue) label.  I'll try to find some time to locate more candidates for that label.
- Changes like this PR that deal with a single template function are often pretty narrow and easier to reason about, so they are a good place to get started.
- Any of the 85 open [Bug](https://github.com/gohugoio/hugo/issues?q=is%3Aopen+is%3Aissue+label%3ABug) issues could use some attention.
- Any issue assigned to a minor milestone (such as v0.82) can be treated as "planned."  It may take us a year to get around to it, and we may keep pushing it to the next minor release, but we intend to do it at some point.

jmooring:
I merged this locally to test a practical application of the new functionality:
<https://discourse.gohugo.io/t/querify-a-dict-structure/31874/2/>

In this example, the key/value pairs are initially in an indeterminate order anyway, so why can't `querify` just take a map? Don't get me wrong, the new functionality is good and much appreciated, but the user still needs to jump through some hoops to use it.

cc @davidsneighbour

importhuman:
This was for #6735, and it was mentioned that it couldn't be a map, so I didn't use it. I'm not exactly clear on why it can't be a map either, but I did as was described in the issue.

moorereason:
@jmooring,
I agree on allowing a lone map argument, but let's deal with that proposal outside of this PR.

moorereason (changes requested):
@importhuman,
I know I already approved this PR, but can you do two more things?

1. Add a simple usage example to [tpl/collections/init.go](https://github.com/gohugoio/hugo/blob/v0.82.0/tpl/collections/init.go#L116).
1. Clean up the commit message to more cleanly consolidate all of the previous messages.

importhuman:
@moorereason Done. Added a simple example, but for future reference, is there a simple way to test out the function/code? I tried online URL encoders and they gave different results than expected, and I don't know how to use the Querify function itself to get the result for the example URL query.

tpl/collections/init.go
```diff
@@ -122,6 +122,10 @@ func init() {
 					`<a href="https://www.google.com?{{ (querify "q" "test" "page" 3) | safeURL }}">Search</a>`,
 					`<a href="https://www.google.com?page=3&amp;q=test">Search</a>`,
 				},
+				{
+					`{{ (querify ["foo" 1 "bar" 2]) | safeHTML }}`,
```
moorereason, line 126: Square brackets are not allowed.  You have to use the `slice` fuction.  Try  `{{ slice "foo" 1 "bar" 2 | querify | safeHTML }}`.

The examples are tested in the `tpl/tplimpl` package.  I usually just do `go test ./tpl/...`

importhuman: @moorereason Thanks!

Yeah I know how to run the test, I'm asking is there a way to run querify to get the URL encoded query for the example, other than using it for a website and seeing the result?

importhuman: I meant how can I run a function to get the output? Say I want to see what will be the result if I run querify for a particular input, is there a way I can do that without adding it to a Hugo website and seeing what happens?

importhuman: It's still failing, the test passed locally, where am I going wrong?

bep:
There are test failures ...?

importhuman:
@bep I recently added an example to tpl/collections/init.go but I'm not sure about the syntax, which is why the error. Trying to fix.

moorereason:
Check your tests:  `template_funcs_test.go:132: collections[2]: got "bar=2&foo=1" expected "foo=1&bar=2"`.  Change the order of the resultant string in your init test.

importhuman:
@moorereason Thanks. A couple questions:

- I read that doing `go test ./...` runs all the tests in that directory, so why didn't `go test ./tpl` work (as in why did it pass when it shouldn't have) and I had to be more specific like `go test -v ./tpl/tplimpl -run TestTemplateFuncsExamples`?
- Seeing the other examples in the file makes it look like the url queries are sorted alphabetically by key. I didn't find anything about this online, is this behaviour specific to Hugo? And if this is how the order is decided, why couldn't we use maps as was the case originally?

moorereason:
Use `go test ./tpl/...` to recurse into subdirectories.

> Seeing the other examples in the file makes it look like the url queries are sorted alphabetically by key. I didn't find anything about this online, is this behaviour specific to Hugo? And if this is how the order is decided, why couldn't we use maps as was the case originally?

To be clear, we're talking about this use case: `{{ slice "foo" 1 "bar" 2 | querify | safeHTML }}`.

Our `slice` function [returns items in the order they're submitted](https://github.com/gohugoio/hugo/blob/v0.82.0/common/collections/slice_test.go#L109), so it shouldn't affect sort order.

`querify` uses `url.Values` to store it's entries.  The description of [`Values.Encode`](https://golang.org/pkg/net/url/#Values.Encode): *Encode encodes the values into “URL encoded” form ("bar=baz&foo=quux") **sorted by key**.*

moorereason (approved):

importhuman:
@bep Can you please take a look at this? :sweat_smile:

bep:
@importhuman sorry for the delay on this...

PR #8327, opened 2021-03-13
exif: Allow more spacing characters in strings

### LINKED ISSUE #8079: Image descriptions in EXIF data with non ASCII characters is missing

Hugo Static Site Generator v0.79.1/extended darwin/amd64 BuildDate: unknown
(Installed using homebrew)

Yes

### Description 
I'm using Hugo to build a photo gallery website, and using the data in the `imagedescription` tag in the EXIF data to display a description below the image. I have noticed that images with non ASCII characters in the field have this element missing in the EXIF object. 

### Steps to reproduce
Use the `Exif.Tags.ImageDescription` to display the image description in a template with a description like this:

> Città del Vaticano #vatican #vaticancity

Observe that the `Exif.Tags.ImageDescription` variable is empty

Use the same template with an image description like this:

> Citta del Vaticano #vatican #vaticancity

Observe that the `Exif.Tags.ImageDescription` variable contains the description from the image. 

Other characters that I've seen that cause this issue are:

`“` instead of `"` or `‘` instead of `'` or any emoji character

### Additional information

The EXIF tags are being written with version 12 of [exiftool](https://exiftool.org/)

moorereason:
The root cause of issue #8079 was a non-breaking space (U+0160).
`unicode.IsPrint` only allows the ASCII space (U+0020).  Be more lenient
by using `unicode.IsGraphic` instead.

Fixes #8079


bep:
Can you fix the commit message?

moorereason:
Fixed.

resources/images/exif/exif.go
```diff
@@ -227,7 +227,7 @@ func (e *exifWalker) Walk(f _exif.FieldName, tag *tiff.Tag) error {
 func nullString(in []byte) string {
 	var rv bytes.Buffer
 	for _, b := range in {
-		if unicode.IsPrint(rune(b)) {
+		if unicode.IsGraphic(rune(b)) {
```
bep, line 230: OK, taking a second look at this, there is one more error in this, as this will not handle multi-byte runes. See the loop in this example:

https://golang.org/pkg/unicode/utf8/#DecodeRune

bep (changes requested):

resources/images/exif/exif.go
moorereason: Commit pushed.  Fixing this should remove the need for the `utf8.ValidString` check.  PTAL

PR #8328, opened 2021-03-14
Fix `new theme` command description

rootkea:
`hugo new theme foo` creates theme `foo` in `./themes` and not in current directory.


davidsneighbour:
That is not really true. 

```
[21:45:20][patrick@bookra] ~/Projects/dnb-hugo/hugonewsletter.com (+1|%0|!896|0)
❯ hugo new theme bla
Creating theme at /home/patrick/Projects/dnb-hugo/hugonewsletter.com/themes/bla
```

If you are in root, then the theme will be in the themes directory.

PR #8329, opened 2021-03-14
tpl: Introduce embed directive

### LINKED ISSUE #8297: Use the new embed package to embed the buit-in Hugo templates
https://golang.org/doc/go1.16#library-embed

Would replace this script:

https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/generate/generate.go

shimst3r:
Introduce the embed package and directive that has been added to the
language with version 1.16.

With the embed directive, using the pre-generated embedded templates is
no longer necessary and the embedded sub-package can be removed safely.

Fixes #8297


bep:
@shimst3r there are test failures.

shimst3r:
@bep you're right, I should've a) read the contributing guidelines more thoroughly and b) not just ran that single test file. :sweat_smile:

I will take a close look at the error messages and update my code accordingly.

bep:
@shimst3r looking at the test error, the solution may not be obvious, but have some embedded templates that depend on the order they get parsed -- using the order from `filepath.Walk`. If the order we get from the new embedded fs (that would suprise me?) we need to do some sorting.

shimst3r:
Hey @bep, I got a bit stuck over the error message.

> If the order we get from the new embedded fs (that would suprise me?) we need to do some sorting.

Could you elaborate a bit what needs to be sorted? The test result or the result from walking the embedded fs?

bep:
>Could you elaborate a bit what needs to be sorted? The test result or the result from walking the embedded fs?
 
We have a couple of shortcodes that calls some "helper" functions in some other internal templates; the thing about Go templates is that they need to be present at parse time, so if you have a template in another file, the order matters.

We have handled this by prefixing these helpers with "_", which lexicographically sorts before regular characters.

* Go's `ioutil.Walk` sorts the directories before it is passed to the callback.
* I don't know what `fs.Walk` do, but if not, it should be easy to apply `sort.Strings`  to the template filename slice before it's sent to be parsed.

PR #8331, opened 2021-03-14
chore: more explicit support link to discourse

davidsneighbour:
Changed the description for the discourse link to a little more forceful wording. I also suggest to create a `discourse` label that has some description with a link to the discourse to  brush off people to the discourse.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,5 @@
 blank_issues_enabled: false
 contact_links:
-    - name: Support
+    - name: SUPPORT, ISSUES and TROUBLESHOOTING
       url: https://discourse.gohugo.io/
-      about: Please do not use Github for support requests. Visit https://discourse.gohugo.io for support!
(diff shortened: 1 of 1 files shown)


davidsneighbour:
This might make it easier to "unceremoniously" close issues by adding a label and moving the conversation to the discourse.

jmooring:
When I look at <https://github.com/gohugoio/hugo/issues/new/choose> my eyes don't make it past the first two entries. Massaging the text might make a difference, but probably not much of a difference. 

Although you can change the sort order of the issue templates by changing the file name (title remains the same), the entries created by the config.yml file always float to the bottom; that's unfortunate.

Additional thoughts:
- The H in GitHub should be uppercase.
- Uppercase text doesn't get my attention; it's just annoying.
- I wonder if the word Support makes people think of a formal support organization (paid, multiple tiers, phone trees, etc.).
- The word "Issues" is confusing; the user is already on the Issues tab. Maybe something like Problems, Questions, and Troubleshooting?
- I'd leave out the last two sentences.

Issue #3088, opened 2017-02-23
proposal: Author model


bep:
Most the work I've done on Hugo lately, have been with simplicity in mind; mostly for the end users, but also to simplify future maintenance and extensions. We may have confused some with the \_index.md thing, but the code cleaned up nicely (we have now only one `pageRender` method, for starters), and the `multilingual` feature landed nicely.

So, any work to make `authors` a first-class Hugo citizen should not be accompanied by a three pages manual.

Here is my design proposal, in pseudo code. 

**Note that the attributes on `author`  below are picked by random, and not really important to this discussion. We should agree on a set of common attributes, so they can be used in themes, but `author` should be schema-less, i.e. a map, so if someone wants to add _favourite_colour_, so be it.**

Pseudo-coded site config:

```
authors:
   id: spf13
   weight: 1
   name: Steve Francia
   favouriteColour: green
 
languages:
no: 
authors:
   id: bep
   name: Bjørn Erik Pedersen
   favouriteColour: blå
   id: spf13
   favouriteColour: grønn
   slogan: "Hugo er best!"  
 
de: 
authors:
   id: spf13
   favouriteColour: blau
```

So given the above, I see the following rule set:

* `Author` and `Authors` should be on `Page`; `Author` should be the first in a sorted list, sorted by weight etc., see `menus`.
* Both `author` and `authors` (a slice) should work in frontmatter. This will be the `id` of the authors, e.g. `bep` in the example above.
* `author` and `authors` should also be possible to be set in site config. These will be default values for pages with no values.
* The author configuration can be parsed top-down: Each language can override author attributes from the global configuration, but not from siblings (`spf13` will only have a slogan in Norwegian).
* The configuration loading should take inspiration from the `output` and `media` package.
* I guess this should even have its own package ... `people`, or `author`?

* The author config should also be externalised at some point, see #3090.

## Social links
I suggest we just name it `socialLinks`, so `.Author.socialLinks` will give you an ordered (ordered by `Weight`) list of:

```
id: "twitter"
name: "Twitter"
weight: 1
url: "https://twitter.com/bepsays"
```

The SocialLinks can be defined like this:

```
[socialLinks]
[socialLinks.facebook]:
name: "Facebook"
weight: 1
urlTemplate: "https://www.facebook.com/%s"
```

Hugo should support a default definition of the most common, users can add or redefine in site config (see `MediaType` and `OutputFormat`).

The social links can then be added to a given `author`:

```
authors:
   id: bep
[...]

PR #8344, opened 2021-03-18
Remove .Site.Authors from embedded templates

### LINKED ISSUE #4458: References to .Site.Authors need to be cleaned up?
Hello,

I had opened up https://discourse.gohugo.io/t/site-authors-not-implemented/10446 on Discourse for discussion, but haven't received any feedback.

The gist of the problem is that internal templates like `twitter_cards` and `opengraph` reference `.Site.Authors`, but no matter what I do, those variables do not get set.

While diving through the Hugo source code through my Go-unexperienced eyes, I realized that those variables are "declared" but never "assigned" (*not sure what the correct Go-lingo is for that*).

You can read the above linked Discourse thread for details and code snippet references.

jmooring:
Closes #4458

### DIFF
--- a/tpl/tplimpl/embedded/templates.autogen.go
+++ b/tpl/tplimpl/embedded/templates.autogen.go
@@ -226,14 +226,6 @@ if (!doNotTrack) {
 <meta property="article:section" content="{{ .Section }}" />
 {{ with .PublishDate }}<meta property="article:published_time" content="{{ .Format $iso8601 }}" />{{ end }}
 {{ with .Lastmod }}<meta property="article:modified_time" content="{{ .Format $iso8601 }}" />{{ end }}
-
-{{- range .Site.Authors }}
-{{ with .Social.facebook }}<meta property="article:author" content="https://www.facebook.com/{{ . }}" />{{ end }}
-{{ with .Site.Social.facebook }}<meta property="article:publisher" content="https://www.facebook.com/{{ . }}" />{{ end }}
-{{- with .Params.tags }}{{ range first 6 . }}
-<meta property="article:tag" content="{{ . }}" />
-{{- end }}{{ end -}}
-{{- end -}}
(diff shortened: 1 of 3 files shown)


jmooring:
This came up again this morning:
<https://discourse.gohugo.io/t/opengraph-template-help/32613>

Please merge or provide feedback. Thanks.

cryptic-code:
`article:author`, `article:publisher` and `twitter:creator` meta tags totally removed from the social media templates?

They need to be there. Why not only remove their dependency on `.Site.Authors` variable?

jmooring:
@cryptic-code, Thank you for sharing your concerns on this. Here are my thoughts:

1. We didn't remove anything, because none of it was rendered... ever. 
2. It is trivial for site creators to copy and customize the embedded templates as needed.
3. There's an open issue to define an authors data model: https://github.com/gohugoio/hugo/issues/3088. Perhaps when this is settled we can revisit these embedded templates.

If you have questions about how to copy and customize an embedded template, please create a topic on the forum (https://discourse.gohugo.io/). Thanks.

cryptic-code:
> @cryptic-code, Thank you for sharing your concerns on this. Here are my thoughts:
> 
> 1. We didn't remove anything, because none of it was rendered... ever.
> 2. It is trivial for site creators to copy and customize the embedded templates as needed.
> 3. There's an open issue to define an authors data model: #3088. Perhaps when this is settled we can revisit these embedded templates.
> 
> If you have questions about how to copy and customize an embedded template, please create a topic on the forum (https://discourse.gohugo.io/). Thanks.

@jmooring Yes. I'm aware that the tags were never rendered and that they are not critical for any site feature.

I raised the concern since having them is still recommended.

I have already customized my template accordingly.

Anyways, thanks. `.Site.Authors` caused a lot of confusion for beginners (including me).

Issue #7222, opened 2020-04-28
A field in a structure is sometimes protected by Mutex, but sometimes unprotected.


Chain-Fox:
# What version of Hugo are you using (`hugo version`)?

master branch, commit c03ea2b66010d2996d652903cb8fa41e983e787f

**What happened**:
A field in a structure is sometimes protected by Mutex, but sometimes unprotected.
This may lead to data race.
There are 2 such cases:
- [ ] `templateStateMap.templates` is not protected by `t.mu` on L754. But for all the other 3 places, it is protected by `t.mu`.
https://github.com/gohugoio/hugo/blob/c03ea2b66010d2996d652903cb8fa41e983e787f/tpl/tplimpl/template.go#L722-L723
- [ ] `t.text` is not protected by `t.nameSpace.mu` on L35. But for all the other 3 places (https://github.com/gohugoio/hugo/blob/master/tpl/internal/go_templates/htmltemplate/template.go), it is protected by `t.nameSpace.mu`.
https://github.com/gohugoio/hugo/blob/c03ea2b66010d2996d652903cb8fa41e983e787f/tpl/internal/go_templates/htmltemplate/hugo_template.go#L31-L35

**What you expected to happen**:
Data race

paskal:
The only way to properly protect `t.text` (as it's used elsewhere after return) is to make a deep copy of it and return the pointer to that copy instead of the original: that's a cure worse than the disease, in my opinion, as it will complicate code for no clear benefit.

bep:
>This may lead to data race.

Only if it's accessed from multiple go routines at the same time.

1. postTransform is, to my knowledge, always done in 1 thread and before the templates gets used.
2. Any concurrency issue with `t.text` should be addressed in the source (Go stdlib).

Also, note that we have a fair amount of tests that run this code successfully with the `-race` flag enabled.

PR #8355, opened 2021-03-22
Wrap t.main.templates into t.main.mu lock

paskal:
First part of the fix for #7222.

### DIFF
--- a/tpl/tplimpl/template.go
+++ b/tpl/tplimpl/template.go
@@ -767,6 +767,8 @@ func (t *templateHandler) postTransform() error {
 	defineCheckedHTML := false
 	defineCheckedText := false
 
+	t.main.mu.RLock()
+	defer t.main.mu.RUnlock()
 	for _, v := range t.main.templates {
 		if v.typ == templateShortcode {
 			t.addShortcodeVariant(v)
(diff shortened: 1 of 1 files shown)


tpl/tplimpl/template.go
```diff
@@ -767,6 +767,7 @@ func (t *templateHandler) postTransform() error {
 	defineCheckedHTML := false
 	defineCheckedText := false
 
+	t.main.mu.RLock()
```
moorereason, line 770: There's a return in the following for-loop where this lock is not released.  Can you use `defer` here?

paskal: You are right! I've overlooked that one. I don't see that operations below are time-consuming so I agree that defer is preferable over having two unlocks.

paskal: No, actually I think we can't use it because of `t.extractPartials(v.Template)` using the lock inside. I assume tests will highlight that and I'll need to change it.

paskal: @moorereason what do you think now?

paskal:
@moorereason gentle ping for review

moorereason:
The obvious error in your initial commit is fixed, so that's good.  I'll let @bep do the final review since he is way more familiar with the inner workings of the template build system.

paskal:
Gentle ping @bep for review.

paskal:
Another gentle ping for @bep to review.

PR #8365, opened 2021-03-26
tpl: change <a> to <div> for non-link elements in pagination internal template

y1zhou:
When on the first or last page, the previous/next link would be disabled and no href would be generated. This impacts the SEO score in LightHouse. The modification changes `<a>` tags to `<div>` elements.

`mage check` returned no errors. The [Contribution Guide](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md) mentioned having test cases for the new code, but I'm not sure how this modification can be tested.

### DIFF
--- a/tpl/tplimpl/embedded/templates.autogen.go
+++ b/tpl/tplimpl/embedded/templates.autogen.go
@@ -265,7 +265,11 @@ if (!doNotTrack) {
   </li>
   {{ end -}}
   <li class="page-item{{ if not $pag.HasPrev }} disabled{{ end }}">
-    <a {{ if $pag.HasPrev }}href="{{ $pag.Prev.URL }}"{{ end }} class="page-link" aria-label="Previous"><span aria-hidden="true">&laquo;</span></a>
+    {{- if $pag.HasPrev -}}
+    <a href="{{ $pag.Prev.URL }}" class="page-link" aria-label="Previous"><span aria-hidden="true">&laquo;</span></a>
+    {{- else -}}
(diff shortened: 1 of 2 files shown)


jmooring:
This will visually break existing sites where the CSS targets the paginator's anchor tags.

y1zhou:
@jmooring True, and unfortunately fixing this accessibility issue without introducing breaking changes seems impossible. Is the `page-link` class targeted anywhere in the template itself?

davidsneighbour:
The solution to this dilemma is to add a `disabled` and add proper aria attributes. (2021 vs. 2010 solution) ;)

```
<a href="#" role="button" aria-disabled="true" disabled>Section name</a>
```

jmooring:
> fixing this accessibility issue without introducing breaking changes seems impossible. 

@y1zhou When did this become an accessibility issue? I thought it was about SEO. This is the lighthouse report using the existing pagination template:

Lighthouse report

The suggestions from @davidsneighbour are great, and we should re-work this entire template at some point. But for now, the only thing that needs to change is the `href` value (use `#` instead of omitting the attribute). That change will produce this:

Lighthouse report

And it shouldn't break anything.

y1zhou:
@jmooring Ah yes it's about SEO. Should have read my own PL with more care. Thank you @davidsneighbour for the great (modern) suggestions and I'll integrate these into the PL.

jmooring:
Resolved with https://github.com/gohugoio/hugo/pull/8602.

PR #8385, opened 2021-04-03
Adds automatic configuration of Tor and I2P Services to `hugo serve`

eyedeekay:
This adds three flags, 2 booleans(--i2p and --tor) and a string(--keysdir), which allow a Hugo server to automatically configure itself to listen at a Tor Onion Service address and as an I2P Site. It handles every step of the setup process including key generation and connection setup. It can also listen on multiple addresses simultaneously. When the keys required to host the hidden service are generated they are stored in the directory indicated by the keysdir flag.


bep:
Thanks for this,

PR #8387, opened 2021-04-05
Add feed preview

frontaid:
Hi. This PR adds a simple preview to the feed template. It does it by using XSLT to restructure the feed to HTML and add a simple styling. Instead of using the generic preview that I added, please feel free to add your own. But please consider a preview as it helps making a difference. See https://github.com/frontaid/feed-ui for more context.

PS: The website of our content management system https://frontaid.io/ is based on Hugo. We are very happy with it. Thank you for your work!

### DIFF
--- a/tpl/tplimpl/embedded/templates/_default/rss.xml
+++ b/tpl/tplimpl/embedded/templates/_default/rss.xml
@@ -11,6 +11,7 @@
 {{- $pages = $pages | first $limit -}}
 {{- end -}}
 {{- printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
+{{- printf "<?xml-stylesheet type=\"text/xsl\" href=\"https://raw.githubusercontent.com/frontaid/feed-ui/main/preview.xsl\"?>" | safeHTML }}
 <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
   <channel>
     <title>{{ if eq  .Title  .Site.Title }}{{ .Site.Title }}{{ else }}{{ with .Title }}{{.}} on {{ end }}{{ .Site.Title }}{{ end }}</title>


frontaid:
> Developer Certificate of Origin Version 1.1
> 
> Copyright (C) 2004, 2006 The Linux Foundation and its contributors. 660 York Street, Suite 102, San Francisco, CA 94110 USA
> 
> Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.
> 
> Developer's Certificate of Origin 1.1
> 
> By making a contribution to this project, I certify that:
> 
> (a) The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or
> 
> (b) The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or
> 
> (c) The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.
> 
> (d) I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project or the open source license(s) involved.

I do agree with that license.

But I do not agree to sharing this account's information with SAP. Hence, I can't agree to their CLA check, sorry.

troxler:
Just noticed that it does not really work in Chrome. Will have another look.

frontaid:
Just noticed that it does not really work in Chrome. Will have another look.

davidsneighbour:
So, you are not giving your "data" to SAP but expect every user of Hugo to include an XSLT formatting that is hosted on your own Gist?

As long as the XSLT is hosted on an external provider you don't need to re-open this for inclusion. The idea is ok, but people can do that on their own account by adding an RSS template with your link.

PR #8404, opened 2021-04-10
Remove extraneous space from figure shortcode

### LINKED ISSUE #8401: figure shortcode produces unwanted space inside anchor element
When used to generate a link like

```
{{% figure src="foo.png" link="foo.png" %}}
```

the `figure` shortcode produces unwanted whitespace inside the anchor tag, and that whitespace will be shown underlined in many cases, which looks confusing to say the least:

```
<figure><a href="foo.png">
    <img src="foo.png"/> </a>
</figure>
```

This applies to both the `hugo-0.80.0-2.fc34.x86_64` package Fedora is shipping,

```
Hugo Static Site Generator v0.80.0/extended linux/amd64 BuildDate: unknown
```

both when using its builtin `figure` shortcode and when using today's `master` branch (cf the SHA in the link below) copied into `themes/THEME/layouts/shortcodes/figure.html`.

Taking a look at the [implementation of the `figure` shortcode](https://github.com/gohugoio/hugo/blob/3d5dbdcb1a11b059fc2f93ed6fadb9009bf72673/tpl/tplimpl/embedded/templates/shortcodes/figure.html#L4) and playing around a bit, I found that wrapping the newlines and whitespace between the tag beginning the `<a>` element and the `` allowing to keep the existing indentation of the source, while stopping generating the unwanted whitespace:

```
<figure><a href="foo.png"><img src="foo.png"/></a>
</figure>
```

An alternative fix would be to "just" remove the whitespace from the source code, which also works, but that makes the code much more difficult to read.

No patch included, as I have not signed any CLA at this time.

jmooring:
Fixes #8401

### DIFF
--- a/hugolib/page_test.go
+++ b/hugolib/page_test.go
@@ -680,7 +680,7 @@ func TestPageWithShortCodeInSummary(t *testing.T) {
 	assertFunc := func(t *testing.T, ext string, pages page.Pages) {
 		p := pages[0]
 		checkPageTitle(t, p, "Simple")
-		checkPageContent(t, p, normalizeExpected(ext, "<p>Summary Next Line. <figure> <img src=\"/not/real\"/> </figure> . More text here.</p><p>Some more text</p>"))
+		checkPageContent(t, p, normalizeExpected(ext, "<p>Summary Next Line. <figure><img src=\"/not/real\"/> </figure> . More text here.</p><p>Some more text</p>"))
 		checkPageSummary(t, p, "Summary Next Line.  . More text here. Some more text")
 		checkPageType(t, p, "page")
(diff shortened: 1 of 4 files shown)


ndim (changes requested):
A few lines below the `<!-- Closing img tag -->` comment, the `figure` shortcode defines another `<a>` anchor element, this time inside a `<figcaption>` element.

Interestingly, this element uses `{{- end -}}` after closing the opening `<a>` tag, which AFAICT removes the same type of whitespace which is an issue in the first `<a>` element defined by the `figure` shortcode.

So if you copy the solution from line 22 to line 4, that should fix the issue.

PR #8428, opened 2021-04-16
Perform strict unmarshaling of YAML data

### LINKED ISSUE #8427: different behavior on the same config file but with different format

<pre>
$ hugo version
hugo v0.82.0-9D960784+extended linux/amd64 BuildDate=2021-03-21T17:28:04Z VendorInfo=gohugoio
</pre>

Yes

### How to reproduce
1. Create new empty site
2. Create a new yaml config with this content:
    opengraph:
         twitter:

    defaultImage:
        opengraph:
3. launch hugo server - all is ok
4. Create a toml config with the same content but in toml format:
    [opengraph]
         [twitter]

    [defaultImage]
        [opengraph]
5. launching hugo server rise this error:
Error: "/home/xxxxx/hugo_extended_0.82.0/try/config.toml:6:1": unmarshal failed: Near line 6 (last key parsed ''): Key 'opengraph' has already been defined.

jmooring:
Fixes #8427

### DIFF
--- a/docs/content/en/functions/NumFmt.md
+++ b/docs/content/en/functions/NumFmt.md
@@ -2,7 +2,6 @@
 title: lang.NumFmt
 description: "Formats a number with a given precision using the requested `negative`, `decimal`, and `grouping` options. The `options` parameter is a string consisting of `<negative> <decimal> <grouping>`."
 godocref: ""
-workson: []
 date: 2017-02-01
 publishdate: 2017-02-01
 lastmod: 2017-08-21
--- a/docs/content/en/functions/delimit.md
+++ b/docs/content/en/functions/delimit.md
@@ -2,7 +2,6 @@
 title: delimit
 description: Loops through any array, slice, or map and returns a string of all the values separated by a delimiter.
 godocref:
-workson: []
 date: 2017-02-01
 publishdate: 2017-02-01
 lastmod: 2017-02-01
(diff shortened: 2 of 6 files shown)


jmooring:
Feel free to reject this PR if you feel it will be a breaking change. Sometimes the right thing to do isn't the right thing to do.

jmooring:
Yeah, so, this broke 4 pages on the docs site where we had some empty duplicate keys. If our own site is any indication of what would happen in the field, this is a breaking change.

I've cleaned up the docs site with https://github.com/gohugoio/hugoDocs/pull/1420.

docs/content/en/functions/NumFmt.md
```diff
@@ -2,7 +2,6 @@
 title: lang.NumFmt
 description: "Formats a number with a given precision using the requested `negative`, `decimal`, and `grouping` options. The `options` parameter is a string consisting of `<negative> <decimal> <grouping>`."
 godocref: ""
-workson: []
```
bep, line 5: Are you sure this fails with strict unmarshalling? That could be a breaking change. I suspect some may have constructs ala:

```
{{ range .Params.workson }}
{{ end }}
```

jmooring: It failed when there were duplicate keys, which I removed fron the docs site and this repo.

bep: Ah, OK, so it's not the empty slice that triggers the error.

jmooring:
Should we just close this PR and wait for v3 of the YAML package?

devkadirselcuk (approved):

jmooring:
Revisit #8427 once https://github.com/gohugoio/hugo/issues/8822 is complete.

PR #8431, opened 2021-04-17
publisher: Exclude comment and doctype elements from writeStats

### LINKED ISSUE #8396: writeStats adds comments and doctype to used tags
```
❯ hugo version
hugo v0.82.0-9D960784+extended linux/amd64 BuildDate=2021-03-21T17:28:04Z VendorInfo=gohugoio
```

When setting build > writeStats to true Hugo writes a list of used tags into hugo_stats.json. In my list the first two items appear to be erroneous, a comment start (comment end is missing) and the `!doctype` which can't be accessed via CSS anyway. 

I am only aware of the use of this file for purgecss so it might not be a bug if this is useful for any other use case, but I am failing to find usecases for comment-start and doctype. 

If they make sense, then comment-end should be part of the list, so this is a bug in either case ;]

```
{
  "htmlElements": {
    "tags": [
      "!--",
      "!doctype",
      "a",
      ...
```

### LINKED ISSUE #8417: hugo_stats.json: bad tag if html element contains tab character

<pre>
$ hugo version
hugo v0.82.0+extended darwin/amd64 BuildDate=unknown
</pre>

Yes

#### Steps to reproduce
1. One markdown file needs to contain an HTML element (e.g. `form`) declared on several lines, **with tabs for indentation**, as this example:
```
<form
	action="[URL]"
	method="post"
>

## Random markdown content
Foo bar
```
2. Run `hugo server -e production`

#### Expected:

Tag `"form"` should be added within the list of tags within `hugo_stats.json`file.

#### Actual:

Tag `"form\n\taction=\"[URL]\"\n\tmethod=\"post\""` is added within the list of tags in `hugo_stats.json` file.

#### Note:

The issue does not happen if the form is declared in one line :
```
<form action="[URL]" method="POST">
```

Similar issue already raised here: #7958 
Maybe also related to #8396

dirkolbrich:
- Reorder code blocks
- Rename cssClassCollectorWriter to htmlElementCollectorWriter, as it just collects html element information
- Expand benchmark to test for minified and unminified content

Fixes #8396, Fixes #8417

### DIFF
--- a/publisher/htmlElementsCollector.go
+++ b/publisher/htmlElementsCollector.go
@@ -20,21 +20,10 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/gohugoio/hugo/helpers"
 	"golang.org/x/net/html"
+
+	"github.com/gohugoio/hugo/helpers"
 )
-
-func newHTMLElementsCollector() *htmlElementsCollector {
-	return &htmlElementsCollector{
-		elementSet: make(map[string]bool),
-	}
-}
-
-func newHTMLElementsCollectorWriter(collector *htmlElementsCollector) *cssClassCollectorWriter {
-	return &cssClassCollectorWriter{
-		collector: collector,
-	}
-}
 
 // HTMLElements holds lists of tags and attribute values for classes and id.
 type HTMLElements struct {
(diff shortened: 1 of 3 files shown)


dirkolbrich:
I even managed to make it slightly faster. Check branch [htmlcollector-bench](https://github.com/gohugoio/hugo/compare/master...dirkolbrich:htmlcollector-bench) for an updated benchmark test on master to make it comparable.

```bash
➜  publisher git:(htmlcollector-bench) gobench --package ../publisher | tee ~/old.txt
Benchmark branch "htmlcollector-bench"

 go version go1.16.3 darwin/amd64

goos: darwin
goarch: amd64
pkg: github.com/gohugoio/hugo/publisher
cpu: Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
BenchmarkClassCollectorWriter-8                            16903             69535 ns/op          103649 B/op        400 allocs/op
BenchmarkClassCollectorWriterMinified-8                    17094             75559 ns/op          104328 B/op        401 allocs/op
BenchmarkClassCollectorWriterWithMinifyStream-8            13074             90168 ns/op          106299 B/op        412 allocs/op
BenchmarkClassCollectorWriterWithMinifyString-8            13304             90376 ns/op          109320 B/op        420 allocs/op
PASS
ok      github.com/gohugoio/hugo/publisher      8.366s
➜  publisher git:(htmlcollector-bench) gobench --package ../publisher | tee ~/new.txt
Benchmark branch "writestats-exclude-comments-and-doctype"

 go version go1.16.3 darwin/amd64

goos: darwin
goarch: amd64
pkg: github.com/gohugoio/hugo/publisher
cpu: Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
BenchmarkClassCollectorWriter-8                            18169             65161 ns/op           96928 B/op        335 allocs/op
BenchmarkClassCollectorWriterMinified-8                    18256             65314 ns/op           97594 B/op        337 allocs/op
BenchmarkClassCollectorWriterWithMinifyStream-8            13780             87129 ns/op           99536 B/op        348 allocs/op
BenchmarkClassCollectorWriterWithMinifyString-8            14005             85706 ns/op          102549 B/op        356 allocs/op
PASS
ok      github.com/gohugoio/hugo/publisher      8.109s
➜  publisher git:(writestats-exclude-comments-and-doctype) benchstat ~/old.txt ~/new.txt 
name                                    old time/op    new time/op    delta
ClassCollectorWriter-8                    69.5µs ± 0%    65.2µs ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterMinified-8            75.6µs ± 0%    65.3µs ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterWithMinifyStream-8    90.2µs ± 0%    87.1µs ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterWithMinifyString-8    90.4µs ± 0%    85.7µs ± 0%   ~     (p=1.000 n=1+1)

name                                    old alloc/op   new alloc/op   delta
ClassCollectorWriter-8                     104kB ± 0%      97kB ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterMinified-8             104kB ± 0%      98kB ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterWithMinifyStream-8     106kB ± 0%     100kB ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterWithMinifyString-8     109kB ± 0%     103kB ± 0%   ~     (p=1.000 n=1+1)

name                                    old allocs/op  new allocs/op  delta
ClassCollectorWriter-8                       400 ± 0%       335 ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterMinified-8               401 ± 0%       337 ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterWithMinifyStream-8       412 ± 0%       348 ± 0%   ~     (p=1.000 n=1+1)
ClassCollectorWriterWithMinifyString-8       420 ± 0%       356 ± 0%   ~     (p=1.000 n=1+1)
```

publisher/htmlElementsCollector.go
```diff
@@ -59,7 +48,50 @@ func (h *HTMLElements) Sort() {
 	sort.Strings(h.IDs)
 }
 
-type cssClassCollectorWriter struct {
+type htmlElementsCollector struct {
```
bep, line 51: You don't have to revert this, it makes sense, but more of a comment for the future: It would be good to avoid these type renames as it makes the change set really big and very hard to review.

publisher/htmlElementsCollector_test.go
```diff
@@ -129,29 +129,96 @@ func TestClassCollector(t *testing.T) {
 	}
 }
 
-func BenchmarkClassCollectorWriter(b *testing.B) {
```
bep, line 132: The primary motivation behind these benchmarks is to be able to compare before/after (using benchstat); but to do that, we cannot have a branch totally changing the existing benchmarks. I'm fine with adding more benchmarks, but we need to keep the old (at least until the relevant change is merged).

publisher/htmlElementsCollector.go
```diff
@@ -86,109 +126,91 @@ func (w *cssClassCollectorWriter) Write(p []byte) (n int, err error) {
...
-				el := parseHTMLElement(s)
-				el.Tag = tagName
-				if w.isPreFormatted(tagName) {
-					w.inPreTag = tagName
-				}
+			// check if we have processed this element before.
+			w.collector.mu.RLock()
+			// See https://github.com/dominikh/go-tools/issues/723
+			//lint:ignore S1030 This construct avoids memory allocation for the string.
+			seen := w.collector.elementSet[s]
```
bep, line 176: With this revamped implementation, the comments surrounding it does not make sense.

This relates to my benchmark comments, but I'm pretty sure we need to preserve the old behaviour.

bep (changes requested):

publisher/htmlElementsCollector.go
dirkolbrich: ok, understood

publisher/htmlElementsCollector_test.go
dirkolbrich: I readded the old bench test, but left the new additions.

publisher/htmlElementsCollector.go
dirkolbrich: I removed the comment. I don't think we need the old behavior, as further up the buffer is written into a string and reset. So here we just have the string left to compare it to the elementSet list.

Issue #7567, opened 2020-08-15
hugo --minify breaks HTML element extraction


Traumflug:
```
$ ../hugo/hugo version
Hugo Static Site Generator v0.74.3/extended linux/amd64 BuildDate: unknown
```
... where 'unknown' should be 'today'.

Let code speak:
```
$ ../hugo/hugo
[...]
$ wc -l hugo_stats.json 
196 hugo_stats.json
$ ../hugo/hugo --minify
[...]
$ wc -l hugo_stats.json 
176 hugo_stats.json
```
Which means, minification looses 20 HTML elements for unknown reasons. Which breaks my site :-)

#### Further diagnosis:
This shorter list contains this line:
```
{
  "htmlElements": {
    "tags": [
      [...]
      "img",
      "imgdefer.length;i++){if(imgdefer[i].hasattribute('data-src')){imgdefer[i].setattribute('src',imgdefer[i].getattribute('data-src'));imgdefer[i].removeattribute('data-src');}}}\u003c/script",
      "li",
[...]
```
This long line is certainly not a HTML tag, but similar to this chunk in one of the partials:
```
<script type="text/javascript">
  window.onload = function () {
    var imgDefer = document.getElementsByTagName('img');

    for (var i = 0; i < imgDefer.length; i++) {
      if (imgDefer[i].hasAttribute('data-src')) {
        imgDefer[i].setAttribute('src', imgDefer[i].getAttribute('data-src'));
        imgDefer[i].removeAttribute('data-src');
      }
    }
  }
</script>
```

Traumflug:
It looks like HTML parsing trips over the JavaScript `<`. It also looks like HTML extraction happens after minification, which might be not the best idea for consistent results.

Traumflug:
This still happens with latest master, `Hugo 0.80.0-DEV`.

davidsneighbour:
Some more triage ;)

- What is the diff of both runs on the json file? Meaning, not what is inside the minified one, but what actual differences are there?
- Are you aware, that there is a `<` in ` for (var i = 0; i < imgDefer.length; i++) {` - I think that might be the issue here. 

As far as I remember you need to somehow printf the javascript code in connection with a [safeJS](https://gohugo.io/functions/safejs/#readout). 

If you can, move the javascript into a static file outside of the layout file. 

I think this is a limitation of what is going on, not a bug per se. Maybe opening a thread over at discourse.gohugo.io might bring a speedy-er solution.

[...]

PR #8436, opened 2021-04-19
Parsing robustness

Traumflug:
Fallout of a hacking Sunday. Tackles  #7567 and #8417.

Tested against my own project, only, as I'm not aware how the testsuite works.

Not so sure about commit 0d91578, _"don't shortcut on previously seen elements"_, as I'm not sure how much performance impact it has. It solves handling ignored tags, but probably needs either removal of then dead code or a different solution for dealing with ignored tags appearing multiple times.

### DIFF
--- a/publisher/htmlElementsCollector.go
+++ b/publisher/htmlElementsCollector.go
@@ -299,7 +299,10 @@ func parseHTMLElement(elStr string) (el htmlElement, err error) {
 	// and pretend it's a <div>.
 	if exceptionList[tagName] {
 		tagBuffer = tagName
-		elStr = strings.Replace(elStr, tagName, "div", 1)
+
+		fields := strings.Fields(elStr);
+		fields[0] = "<div"
+		elStr = strings.Join(fields, " ")
 	}
 
 	n, err := html.Parse(strings.NewReader(elStr))
--- a/publisher/htmlElementsCollector.go
+++ b/publisher/htmlElementsCollector.go
@@ -244,12 +244,10 @@ func parseStartTag(s string) (string, bool) {
 	s = strings.TrimPrefix(s, "<")
 	s = strings.TrimSuffix(s, ">")
 
-	spaceIndex := strings.Index(s, " ")
-	if spaceIndex != -1 {
-		s = s[:spaceIndex]
-	}
+	fields := strings.Fields(s);
+	tag := fields[0];
 
-	return strings.ToLower(strings.TrimSpace(s)), true
+	return strings.ToLower(tag), true
 }
 
 func (c *htmlElementsCollectorWriter) parseEndTag(s string) bool {
--- a/publisher/htmlElementsCollector.go
+++ b/publisher/htmlElementsCollector.go
@@ -135,8 +135,11 @@ func (w *htmlElementsCollectorWriter) Write(p []byte) (n int, err error) {
 				w.buff.WriteByte(b)
 
 				if !w.inQuote && b == '>' {
-					w.endCollecting()
-					break
+					s := w.buff.String()
+					if w.inPreTag == "" || w.parseEndTag(s) {
+						w.endCollecting()
+						break
+					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


Traumflug:
In case somebody wants to enhance the testsuite, here are additional cases:
```
<script type="text/javascript">
    for (var i = 0; i < imgDefer.length; i++) {
        doNothing();
    }
</script>
```
```
<script
  type="text/javascript"
>
  if (a<b && a>c) {
    haha();
  }
</script>
```
```
<form
	action="[URL]"
	method="post"
>
```
```
<hr	id=a>
```
Yes, newlines and tab vs. space matters :-)

Failure of each is recognized by non-alphabetical characters appearing in the `hugo_stats.json`.

bep:
>Tested against my own project, only, as I'm not aware how the testsuite works.

```
go test ./publisher
```

bep:
I have pulled this into #8442 -- closing this. And thanks.

bep:
Thanks for this.  

There are failing tests. It looks mostly to be elements inside pre/textarea, which I don't see any reason to collect (but please prove me wrong).

Also, it would be great if you could 

1. Add some test cases for the cases this PR is supposed to fix.
2. Run the benchmark inside the ./publisher package and compare it to master (see below) 


```bash
name                     old time/op    new time/op    delta
ClassCollectorWriter-16    21.2µs ± 2%    47.3µs ± 1%  +123.21%  (p=0.029 n=4+4)

name                     old alloc/op   new alloc/op   delta
ClassCollectorWriter-16    35.3kB ± 0%    86.4kB ± 0%  +144.95%  (p=0.029 n=4+4)

name                     old allocs/op  new allocs/op  delta
ClassCollectorWriter-16       155 ± 0%       329 ± 0%  +112.26%  (p=0.029 n=4+4)
```

Note that I don't mind it getting slower if really needed, but I fail to understand the motivation behind the removal of `elementSet`, which I suspect is the reason why it is now so much slower and memory hungry.

bep:
Can you check whether the latest commits re this in the master branch covers your needs?

Traumflug:
Thanks for the hint on how to run these testcases easily. I just pushed a reviewed set of commits.

> There are failing tests. It looks mostly to be elements inside pre/textarea, which I don't see any reason to collect

Well, the simple reason for these failures was me removing special handling of pre/textarea, but not knowing how to remove these tests as well.

That said, removing these was a mistake, I misunderstood the intention of this code earlier. It's not about not collecting non-styleable tags, but about ignoring stuff inside them. Accordingly, I melted down the first commit to just adding another two such tags with preformatted code. Where _<title>_ isn't exactly preformatted, but of no use for styling anyways.

> Can you check whether the latest commits re this in the master branch covers your needs?

Well, some, but not all. This time, knowing how to deal with these tests, I managed to find a testcase for each. Committed testcases separately to allow seeing them failing. Luckily, I also found a fix for each.

Also one or two testcases which fell into the keyboard somehow, without a matching reported issue I'd be aware of :-)

> Note that I don't mind it getting slower if really needed, but I fail to understand the motivation behind the removal of elementSet

This time I moved the commit with this removal, along with a commit with a demonstrating (previously failing, now working) testcase just before it, all to the end to allow (cherry-)picking earlier commits.

The problem to solve appears when such a preformatted tag appears twice and contains a `<` (happens rarely in testcases, almost always in real world usage). First round everything is fine and dandy. On the second round, the processing time saver misses the part where `w.inPreTag` gets set. Accordingly, the following parsing doesn't recognize the preformatted tag as such and treats it like a normal tag.

Perhaps you have a better idea on how to get the best of both, quick processing as well as recognizing repeated preformatted tags.

merchantsedition:
I just rebased the pull request branch to tag _v0.83.1_ and force-pushed it.

bep:
Thanks, but I've already spent a considerably amount of time simplifying this in another. Would welcome any failed test cases after that PR is merged.

PR #8467, opened 2021-04-25
Deploy invalidate specific paths

davidejones:
**Problem:** Hugo deploy will run a cache invalidation when we set the cloudfront distributionid which is great. However this clears everything and we want to be able to target specific paths.

**Solution:** This PR adds an `invalidatePaths` deployment configuration option.  When undefined the current functionality should remain clearing `/*`. However if you specify this option then the invalidation will run with your supplied paths.

**Notes:** I've tested this as successfully working on my aws s3/cloudfront setup both with default and specified paths. I haven't got a gcp setup handy to test on.

### DIFF
--- a/deploy/cloudfront.go
+++ b/deploy/cloudfront.go
@@ -26,7 +26,7 @@ import (
 
 // InvalidateCloudFront invalidates the CloudFront cache for distributionID.
 // It uses the default AWS credentials from the environment.
-func InvalidateCloudFront(ctx context.Context, distributionID string) error {
+func InvalidateCloudFront(ctx context.Context, distributionID string, invalidatePaths []string) error {
 	// SharedConfigEnable enables loading "shared config (~/.aws/config) and
 	// shared credentials (~/.aws/credentials) files".
 	// See https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ for more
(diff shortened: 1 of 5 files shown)


bep:
I think we have an issue about this, and I agree that we need something like this.

But for me the primary use case would be:

1. I edit or add a new blog post (or something)
2. I want to invalidate the cache for the changed files.

I don't see how a configuration option would help me with the above. I remember implementing something _like this_ in [s3deploy](https://github.com/bep/s3deploy). As these cache invalidations comes at a cost (eventually, after the free quota is used), a fall back to clearing everything would be needed.

/cc @vangent

davidejones:
Currently I'm deploying without invalidating the cf cache as I have a low TTL or I have files fingerprinted etc. In my scenario I have a single file I'd like to invalidate during deploy and a solution like this seemed to fit. 

However I think your primary use case is far better and works here as the file contents would change.

So maybe something like passing the uploads/deletes from the deploy diff as paths to invalidate instead?

bep:
>So maybe something like passing the uploads/deletes from the deploy diff as paths to invalidate instead?

Yes, and as I said: You will probably want a way to reduce those paths to e.g. `/mysection/*` if the number of uploads/deletes gets too high (to reduce cost).

bep:
I'm closing this for now. See my comment above.

PR #8474, opened 2021-04-26
Allows user-defined HTTP headers with getJSON and getCSV with interfa…

chamberlainpj:
…ce extension specified in issue 5617:

func (ns *Namespace) GetJSON(args ...interface{}) (interface{}, error) and then checking whether the last  is a map[string]interface{} , then using those for the headers (similar change to GetCSV).

### DIFF
--- a/tpl/data/data.go
+++ b/tpl/data/data.go
@@ -58,8 +58,8 @@ type Namespace struct {
 // The data separator can be a comma, semi-colon, pipe, etc, but only one character.
 // If you provide multiple parts for the URL they will be joined together to the final URL.
 // GetCSV returns nil or a slice slice to use in a short code.
-func (ns *Namespace) GetCSV(sep string, urlParts ...interface{}) (d [][]string, err error) {
-	url := joinURL(urlParts)
+func (ns *Namespace) GetCSV(sep string, args ...interface{}) (d [][]string, err error) {
+	url := joinURL(args)
 	cache := ns.cacheGetCSV
 
 	unmarshal := func(b []byte) (bool, error) {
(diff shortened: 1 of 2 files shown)


tpl/data/data.go
```diff
@@ -97,9 +106,9 @@ func (ns *Namespace) GetCSV(sep string, urlParts ...interface{}) (d [][]string,
 // GetJSON expects one or n-parts of a URL to a resource which can either be a local or a remote one.
 // If you provide multiple parts they will be joined together to the final URL.
 // GetJSON returns nil or parsed JSON to use in a short code.
-func (ns *Namespace) GetJSON(urlParts ...interface{}) (interface{}, error) {
+func (ns *Namespace) GetJSON(args ...interface{}) (interface{}, error) {
 	var v interface{}
-	url := joinURL(urlParts)
+	url := joinURL(args)
```
bep, line 111: All of this looks mostly great. I have a few comments_

If I read correctly, any headers will be joined to the string here. What I would suggest is to create a separate function that splits the URL parts of the slice and the header part, and add a small test for that.

tpl/data/data_test.go
```diff
@@ -119,6 +119,20 @@ func TestGetCSV(t *testing.T) {
 		c.Assert(got, qt.Not(qt.IsNil), msg)
 		c.Assert(got, qt.DeepEquals, test.expect, msg)
 
+		// Test user-defined headers as well
+		gotHeader, _ := ns.GetCSV(test.sep, test.url, map[string]interface{}{"Accept-Charset": "utf-8", "Max-Forwards": "10"})
```
bep, line 123: To improve these header tests I suggest that we just echo these headers in the response and then add an assertion for one of them. 

Maybe we could also add the URL in the response as a custom header and assert on that? That way we can make sure that that is well formed.

bep (changes requested):

bep:
I will finish this in #8609

Issue #8479, opened 2021-04-29
Enhance the countwords function to handle special characters


jmooring:
Current behavior:

Statement|Actual|Expected|Status
:--|:--|:--|:--
|`countwords "Here's to you..."`|3|3|Pass
|`countwords "Here’s to you..."`|8|3|Fail
|`countwords "Here’s to you…"`|11|3|Fail

These are just two examples of special characters that cause problems; there are others.

The .WordCount page method handles these correctly.

bep:
I'm not totally sure how this fail, but there is obviously a faulty CJK language detection in there, which makes some of the examples above count runes instead. We have the same logic in `.WordCount`, so it should be possible to look at that.

ResamVi:
Hello!
I'm new to contribute: [Here's](https://github.com/gohugoio/hugo/pull/8488) a PR using the same logic as proposed:

PR #8488, opened 2021-04-30
tpl: Enhance countwords to handle special chars

ResamVi:
Copies the implementation that already is part of .WordCount
This should handle the given test cases from #8479

### DIFF
--- a/tpl/strings/strings.go
+++ b/tpl/strings/strings.go
@@ -17,6 +17,7 @@ package strings
 import (
 	"errors"
 	"html/template"
+	"regexp"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


tpl/strings/strings.go
```diff
@@ -75,6 +78,11 @@ func (ns *Namespace) CountWords(s interface{}) (int, error) {
 		return 0, _errors.Wrap(err, "Failed to convert content to string")
 	}
 
+	isCJKLanguage := cjkRe.Match([]byte(ss))
```
bep, line 81: Use MatchString (which would save some memory allocation).

bep, line 81: Use MatchString (which would save some memory allocation). Other than that, this looks great.

bep (changes requested):

tpl/strings/strings.go
bep: Also: Could you change the commit message to say "tpl: Fix ..." or something (as this is a bug fix, we use the "fix" word to put them in the right category in the release notes). Also end the commit message with "Fixes #id".

ResamVi: Done!

PR #8526, opened 2021-05-12
Update docs for GitHub Actions

### LINKED ISSUE #8525: confusing documentation: deployment for GH Pages
Yesterday, I migrated from Lektor to Hugo. While the migration was straight forward, I was pretty puzzled with how to deploy to GH pages.

The official documenation ( https://github.com/gohugoio/hugo/blob/master/docs/content/en/hosting-and-deployment/hosting-on-github.md#build-hugo-with-github-action ) only shows a GitHub Action with zero explanation.

The two things which left me confused:
- I had no clue to which repository I should push my new Hugo source; as e.g. for Lektor I had a `my-user.github.io.source` repository for the Lektor data/raw data, and a `my-user.github.io` repository for the rendered HTML.
- at above mentioned mentioned page there is no mention what happens when you add the action to your repository

When I have a closer look at the action, especially at this part...
```
      - name: Deploy
        uses: peaceiris/actions-gh-pages@v3
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: ./public
```
... two more questions pop up....

Do I need to create a secret token? How? Or will this be autocreated?

Also, the last line mentions a `public` directory as deployment target - but what actually happened when I put this action in my repo is, that another branch called `gh-pages` was created.

I suggest to update the documentation and make it clearer what needs to be done and what happens then.

I can't do this, as - even now that I managed to deploy my site - I still have no clue why the action mentions the `public` dir but in reality a `gh-pages` branch was created as the deployment target.

Thanks for Hugo! Other than the deployment process, it is a breeze to use!

GrahamSH-LLK:
Resolves #8525


pplmx (approved):

jmooring:
@GrahamSH-LLK Thanks for this, but you will need to submit the PR to the documentation repository:
<https://github.com/gohugoio/hugoDocs>. 

I transferred the original issue to that repository as well:
<https://github.com/gohugoio/hugoDocs/issues/1466>.

Thanks again.

PR #8550, opened 2021-05-20
fix: --cleanDestinationDir will not be skipped if static isn't present

### LINKED ISSUE #8433: commands: Make cleanDestinationDir work even when staticDir doesn't exist
`hugo v0.81.0+extended linux/amd64 BuildDate=unknown`

Specifically, it seems to be ignored if you have never used this option from CLI. After installing Hugo I just put `cleanDestinationDir: true` in config.yaml of the site I'm trying to make. However, the command `hugo` never deleted any directories or files that were becoming obsolete. For example, I first built the site with the default taxonomies, but later decided that I didn't need them, so I disabled them with

```
disableKinds:
- taxonomy
- term
```

However, after running `hugo` the directories "tags" and "categories" were still in "public". The same was the case with obsolete files/directories in "content". Only after I have commented the option `cleanDestinationDir: true` in the config.yaml and ran `hugo --cleanDestinationDir` from the command line did Hugo build the "public" directory without the obsolete files/directories.

After that, to test the issue, I uncommented the line in the config and ran just `hugo`. But now the option in config is working and the obsolete files are deleted from "public" as they should. It's like something has been initialized by running `hugo --cleanDestinationDir` once.

ResamVi:
Adds a test that executes the string of commands as mentioned by [this comment](https://github.com/gohugoio/hugo/issues/8433#issuecomment-822988001)

Removes all files inside of public when 
- no static directories exist
- --cleanDestinationDir was specified

fixes #8433

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -653,6 +653,28 @@ func (c *commandeer) copyStaticTo(sourceFs *filesystems.SourceFilesystem) (uint6
 		syncer.DeleteFilter = func(f os.FileInfo) bool {
 			return f.IsDir() && strings.HasPrefix(f.Name(), ".")
 		}
+
+		// when `static` directory doesn't exist, using --cleanDestionationDir wouldn't
+		// remove items of an existing public directory that contains files we don't want.
+		// So delete them ourselves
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this, but this isn't what `--cleanDestinationDir` means.

PR #8581, opened 2021-05-26
Add `plainify` to Title - in internal template opengraph.html 

andreseduardop:
Some users can use html tags in the title, for example `<br>`, `<em>` or `<strong>`.

Adding `plainify` to clean up these tags will prevent the title from being rendered inappropriately when sharing a page on Facebook for example.


andreseduardop:
Hello. Can someone with maintainer privileges approve this please?

davidsneighbour:
This makes sense. I just did that to my own title tags everywhere. To get changes like this easier into Hugo you should open an issue where you discuss the reasoning, then add a PR referring to that issue.

andreseduardop:
> This makes sense. I just did that to my own title tags everywhere. To get changes like this easier into Hugo you should open an issue where you discuss the reasoning, then add a PR referring to that issue.

Thanks for the hint. I just opened a problem for this.
https://github.com/gohugoio/hugo/issues/8698

andreseduardop:
Fix the problem described here: [Issue 8698: Open Graph does not support html tags nests inside the og:title property.](https://github.com/gohugoio/hugo/issues/8698)

PR #8588, opened 2021-05-29
Fix invalid timestamp of the "public" folder

### LINKED ISSUE #6161: hugo creating invalid timestamps in folder public
I am using version 0.56.3 on MacOS 10.14.6 in rstudio.
There are problems with the folders `css fonts images js` and `post`  in the folder `public`.
Here are is the listing from the theme folder `static`

```
drwxr-xr-x  4 neuwirth  staff    128 Jul 31 19:00 css
drwxr-xr-x  6 neuwirth  staff    192 Jul 31 19:00 fonts
drwxr-xr-x  5 neuwirth  staff    160 Jul 31 19:04 images
drwxr-xr-x  3 neuwirth  staff     96 Jul 31 19:00 js
```

And here ist the listing in `public` 

```
drwxr-xr-x  4 neuwirth  staff    128 Aug 30  1754 css
drwxr-xr-x  6 neuwirth  staff    192 Aug 30  1754 fonts
drwxr-xr-x  5 neuwirth  staff    160 Aug 30  1754 images
drwxr-xr-x  3 neuwirth  staff     96 Aug 30  1754 js
drwxr-xr-x  5 neuwirth  staff    160 Aug 30  1754 post
```
time is not a valid value in the public directory

anthonyfok:
Special thanks to both Richard Mortimer (@oldelvet) and
Joshua M. Clulow (@jclulow) for their analysis and suggested fix:

 * https://github.com/gohugoio/hugo/issues/6161#issuecomment-574336088
 * https://github.com/gohugoio/hugo/issues/6161#issuecomment-596805273

Fixes #6161

---

I am satisfied with the behaviour I see after applying this patch.  The timestamp of `public` folder itself gets updated to "now" for each `hugo` run; the timestamps of the contents inside the public folder look good.  `hugo server` LiveReload still works as expected, whether rendered to memory or to disk.

### DIFF
--- a/hugofs/fileinfo.go
+++ b/hugofs/fileinfo.go
@@ -280,7 +280,7 @@ func (fi *dirNameOnlyFileInfo) Mode() os.FileMode {
 }
 
 func (fi *dirNameOnlyFileInfo) ModTime() time.Time {
-	return time.Time{}
+	return time.Now()
 }
 
 func (fi *dirNameOnlyFileInfo) IsDir() bool {


moorereason (approved):
The only possible argument I see against this PR is to use `time.Unix(0, 0)` instead of `time.Now()`, but I think that will cause similar confusion with users.  This seems like the simplest solution to me.

LGTM 👍

anthonyfok:
Thanks @moorereason!  I'll "Rebase and merge" this now.  :-)

PR #8602, opened 2021-06-01
Improve pagination performance

### LINKED ISSUE #8599: Improve pagination performance
The [embedded pagination template](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/pagination.html) ranges through every pager while rendering every pager. The number of iterations increases quadratically in relation to the number of content pages.

n = (c/p)^2 where

- n = iterations
- c = content pages
- p = content pages per pager

Using the default number of content pages per pager (10):

Content Pages|Iterations
--:|--:
100|100
1,000|10,000
10,000|1,000,000
25,000|6,250,000
50,000|25,000,000
75,000|56,250,000

There have been a couple of related forum issues over the past month:
- <https://discourse.gohugo.io/t/slow-build/32745>
- <https://discourse.gohugo.io/t/build-performance/33093>

PR to follow...

jmooring:
These calls are equivalent:

```go-html-template
{{ template "_internal/pagination.html" . }}
{{ template "_internal/pagination.html" (dict "page" .) }}
{{ template "_internal/pagination.html" (dict "page" . "format" "default") }}
```

To use an alternate format:

```go-html-template
{{ template "_internal/pagination.html" (dict "page" . "format" "terse") }}
```

Fixes #8599

### DIFF
--- a/tpl/tplimpl/embedded/templates/pagination.html
+++ b/tpl/tplimpl/embedded/templates/pagination.html
@@ -65,7 +65,7 @@
     {{- range $k := seq $start $end }}
       {{- if eq $.Paginator.PageNumber $k }}
       <li class="page-item active">
-        <a href="#" aria-label="Page {{ $k }}" class="page-link" role="button">{{ $k }}</a>
+        <a href="#" aria-current="page" aria-label="Page {{ $k }}" class="page-link" role="button">{{ $k }}</a>
       </li>
       {{- else }}
       <li class="page-item">
@@ -128,7 +128,7 @@
     {{- range $k := seq $start $end }}
       {{- if eq $.Paginator.PageNumber $k }}
       <li class="page-item active">
-        <a href="#" aria-label="Page {{ $k }}" class="page-link" role="button">{{ $k }}</a>
+        <a href="#" aria-current="page" aria-label="Page {{ $k }}" class="page-link" role="button">{{ $k }}</a>
       </li>
       {{- else }}
       <li class="page-item">
--- a/tpl/tplimpl/embedded/templates/pagination.html
+++ b/tpl/tplimpl/embedded/templates/pagination.html
@@ -1,45 +1,154 @@
-{{ $pag := $.Paginator }}
-{{ if gt $pag.TotalPages 1 -}}
-<ul class="pagination">
-  {{ with $pag.First -}}
-  <li class="page-item">
-    <a href="{{ .URL }}" class="page-link" aria-label="First"><span aria-hidden="true">&laquo;&laquo;</span></a>
-  </li>
-  {{ end -}}
-  <li class="page-item{{ if not $pag.HasPrev }} disabled{{ end }}">
-    <a {{ if $pag.HasPrev }}href="{{ $pag.Prev.URL }}"{{ end }} class="page-link" aria-label="Previous"><span aria-hidden="true">&laquo;</span></a>
-  </li>
-  {{- $ellipsed := false -}}
-  {{- $shouldEllipse := false -}}
-  {{- range $pag.Pagers -}}
-  {{- $right := sub .TotalPages .PageNumber -}}
-  {{- $showNumber := or (le .PageNumber 3) (eq $right 0) -}}
-  {{- $showNumber := or $showNumber (le .TotalPages 5) -}}{{/* Issue #7523 */}}
-  {{- $showNumber := or $showNumber (and (gt .PageNumber (sub $pag.PageNumber 2)) (lt .PageNumber (add $pag.PageNumber 2))) -}}
-  {{- if $showNumber -}}
-    {{- $ellipsed = false -}}
-    {{- $shouldEllipse = false -}}
-  {{- else -}}
-    {{- $shouldEllipse = not $ellipsed -}}
-    {{- $ellipsed = true -}}
-  {{- end -}}
-  {{- if $showNumber }}
-  <li class="page-item{{ if eq . $pag }} active{{ end }}">
-    <a class="page-link" href="{{ .URL }}">{{ .PageNumber }}</a>
-  </li>
-  {{- else if $shouldEllipse }}
-  <li class="page-item disabled">
-    <span aria-hidden="true">&nbsp;&hellip;&nbsp;</span>
-  </li>
-  {{- end -}}
+{{- $validFormats := slice "default" "terse" }}
+
+{{- $msg1 := "When passing a map to the internal pagination template, one of the elements must be named 'page', and it must be set to the context of the current page." }}
+{{- $msg2 := "The 'format' specified in the map passed to the internal pagination template is invalid. Valid choices are: %s." }}
+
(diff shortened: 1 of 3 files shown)


jmooring:
Test site...

## Instructions

Clone this branch of the repository and build the site.

```bash
git clone --single-branch -b hugo-github-issue-8602 https://github.com/jmooring/hugo-testing hugo-github-issue-8602
cd hugo-github-issue-8602
npm install
hugo
npx serve
```

Notes:

- We use the [npm serve](https://www.npmjs.com/package/serve) package (`npx serve`) to serve the published site because `hugo server` can be a bit sluggish with larger sites.
- Running `hugo` while running an IDE with an integrated file watcher can produce misleading performance metrics for larger sites. Close the IDE, then run `hugo` from a native terminal.
- Finally, also in the interest of obtaining more accurate performance metrics, remove the public and resources directories before each run.

## Site Structure

There are 10 content types (sections):

Section Title|Content Pages|Section Pages|Pagers|Aliases
:--|--:|--:|--:|--:|
10 Pages|10|1|0|1
20 Pages|20|1|1|1
30 Pages|30|1|2|1
40 Pages|40|1|3|1
50 Pages|50|1|4|1
60 Pages|60|1|5|1
70 Pages|70|1|6|1
80 Pages|80|1|7|1
90 Pages|90|1|8|1
10,000 Pages|10,000|1|999|1

Visit each section to see an example of the pagination format as it varies with the number of pagers.

## Expected Counts

- Pages: 10,461 (content pages + section pages + home page)
- Paginator pages (pagers): 1035
- Aliases: 10

## Format

There are two display formats: default and terse.

The "default" format:

- Always displays a link to the First pager
- Always displays a link to the Previous pager
- Always displays a link to the Next pager
- Always displays a link to the Last pager
- Displays links to a maximum of 5 pagers, 2 on either side of the link to the current pager

The "terse" format:

- Does not display a link to the First pager from the first pager
- Does not display a link to the Previous pager from the first pager
- Does not display a link to the Next pager from the last pager
- Does not display a link to the Last pager from the last pager
- Displays links to a maximum of 3 pagers, 1 on either side of the link to the current pager

To use the "default" format, these are equivalent:

   ```go-html-template
   {{ template "_internal/pagination.html" . }}
   {{ template "_internal/pagination.html" (dict "page" .) }}
   {{ template "_internal/pagination.html" (dict "page" . "format" "default") }}
   ```

To use the "terse" format:

   ```go-html-template
   {{ template "_internal/pagination.html" (dict "page" . "format" "terse") }}
   ```

This approach will allow us to add alternate formats in the future without changing the appearance of existing sites.

I based the formats on a review of pagination formats from Google, Bing, and Amazon. See <https://gist.github.com/jmooring/94e6e0bb0568ae5cdbae3372eb450111>.

## Style

While I have applied minimal styling to this test site, the [Pull Request](https://github.com/gohugoio/hugo/pull/8602) does not contain any CSS. Site creators continue to be responsible for styling.

## Validation

The output is valid HTML5 per <https://validator.w3.org/>.

Scores from [Google Lighthouse](https://developers.google.com/web/tools/lighthouse):

- Performance: 100
- Accessibility: 100
- Best Practices: 100
- SE0: 100

## Performance

Build times are the average of 3 runs.

ID|Pagination Template|Format|Build Time (seconds)
:--|:--|:--|--:
1|None|NA|1.8
2|Old|NA|22.4
3|New|default|2.1
4|New|terse|2.1

tpl/tplimpl/embedded/templates/pagination.html
```diff
@@ -1,45 +1,158 @@
...
+
+{{- $msg1 := "When passing a map to the internal pagination template, one of the elements must be named 'page', and it must be set to the context of the current page." }}
+{{- $msg2 := "The 'format' specified in the map passed to the internal pagination template is invalid. Valid choices are: %s." }}
+
+{{- $page := . }}
+{{- $format := "default" }}
+
+{{- if reflect.IsMap . }}
+  {{- with .page }}
+    {{- if eq (printf "%T" .) "*hugolib.pageState" }}
```
bep, line 11: This doesn't work.

1. It's very fragile to depend on some internal struct names, they may change without notice.
2. `pageState` is only one of the structs that represents a `Page` -- in a shortcode, a `.Page` will be something else. Have no current workaround for you other than just making it fail if it's not a Page.

jmooring: Understood. Thanks. Updated as requested.

bep (commented):
OK, this looks good now, but with the added complexity we need a minimal test for the variants.

Could you add the variants and some very simple assertions to TestEmbeddedTemplates in hugolib package?

Note that I'm mostly worried about someone making a "cometic change" months from now, introducing a nil-pointer or something.

jmooring:
Added minimal test for variants; assertions for first pager only.

bep:
Hey, I have thought about this while sleeping this weekend, and the internals looks very good -- but I'm not sure adding a `dict` config option is worth the complexity for toggling just one option (and the option is basically meaning "choose another template"). So, this: 

```html
{{ template "_internal/pagination.html" . }}
{{ template "_internal/pagination.html" (dict "page" .) }}
{{ template "_internal/pagination.html" (dict "page" . "format" "default") }}
{{ template "_internal/pagination.html" (dict "page" . "format" "terse") }}
```

Would be covered by:

```html
{{ template "_internal/pagination.html" . }}
{{ template "_internal/pagination-terse.html" . }}
```

Which I think is easier to explain to people. What do you think?

jmooring:
I started out with something like this:

```
{{ $options := dict
  "page" .
  "slots" 9
  "showFirst" true
  "showFirstOnFirstPage" false
  "showPrevious" true
  "showPreviousOnFirstPage" false
  "showNext" true
  "showNextOnLastPage" false
  "showLast" true
  "showLastOnLastPage" false
}}
{{ template "_internal/pagination.html" $options }}
```

Or to call it with sensible defaults:

```
{{ template "_internal/pagination.html" . }}
```

But I was concerned that you would say, "This is too complicated." So I simplified, but perhaps I shouldn't have. I like the above better than what I finished with, and I would be happy to revert to this approach, including test cases.

bep:
>But I was concerned that you would say, "This is too complicated."

I would have said that. This needs to be maintained and explained to people. If people want more advanced options, they can create their own template.

This issue is about solving a performance problem, let's keep that focus. I suggest we do it like this:

```html
{{ template "_internal/pagination.html" . }}
{{ template "_internal/pagination-terse.html" . }}
```

varunchopra:
I apologise if this sounds too stupid/ignorant -

1. Does this change reduce memory usage? I believe the core ask with "slow" builds has been that slow becomes impossible due to RAM usage as Hugo doesn't support incremental builds.
2. Can you explain how this impacts templates that are in use right now?

Also, if it helps, your comment here https://github.com/gohugoio/hugo/pull/8602#issuecomment-855415586 certainly makes much more sense than the OP.

PR #8606, opened 2021-06-03
Replace underlying rwcarlsen/goexif with dsoprea/go-exif

### LINKED ISSUE #8586: Replace underlying Exif library
I suggest to replace the current Exif library, which has not been worked on in the last two years and poses some problems with 
https://github.com/dsoprea/go-exif
That one is actively developed and seems to be able to read IPTC data, too. Which would open up a lot more possibilities for image handling.
Cf. https://discourse.gohugo.io/t/chance-of-changing-the-exif-module/33112

danedavid:
Fixes #8586 

- Replace underlying EXIF library ([rwcarlsen/goexif](https://github.com/rwcarlsen/goexif)) with an actively developed one ([dsoprea/go-exif](https://github.com/dsoprea/go-exif))
- Add tests for disable options (date & location) while decoding EXIF.


danedavid:
Couple of questions:

1. In the body of `Decode` method, there are many calls to `goexif` code that return an `error` object as a second value. Being new to Go, I was wondering if we need to add the following after each assignment to `err`? Is there a better, idiomatic way?
```go
if err != nil {
  return
}
```

2. As the values of the `Tags` map are of type `interface{}`, there is no guarantee that a particular tag value has the same type as when we were using the previous library, except the tags (DateTime, LatLong) that are handled specifically. Do we need to add any type specific tests there?

3. The API of `exif` module is unchanged, but would you recommend adding more tests? If so, what parts need to be tested? Have added one new test to check for disable LatLong & Date options at the module level. I believe it's already being tested in the `image` module.

4. Should the old library be removed from dependency list in README?

5. The unexported `exifWalker` struct & its `Walk` method may not be needed as such because the new library doesn't require a `walker` function. However, have maintained the same signatures for brevity. Let me know if those need to be changed.

resources/images/exif/exif.go
```diff
@@ -113,127 +113,104 @@ func (d *Decoder) Decode(r io.Reader) (ex *Exif, err error) {
...
+func processValue(valueRaw interface{}, tagType exifcommon.TagTypePrimitive, tagName string) (val interface{}, err error) {
+	switch tagType {
+	case exifcommon.TypeAscii:
+		s, isString := valueRaw.(string)
+		if isString && strings.Contains(tagName, "DateTime") {
+			if d, err := tryParseDate(s); err == nil {
+				return d, nil
+			}
+		}
+		return s, nil
```
bep, line 177: I'm not sure this could ever happen, but you do have an `isString` check above. If this is not a string at this point, `s` would be empty. So you should probably not return anything here.

bep: The rest of the good looks very good, but I ran this PR on a site I have that uses the `Exif`: 

https://github.com/bep/forsale

Where I now get this error:


```bash
exif" .>: error calling partial: "/Users/bep/dev/sites/forsale/layouts/partials/exif.html:3:6": execute of template failed: template: partials/exif.html:3:6: executing "partials/exif.html" at <.Exif>: error calling Exif: metadata init failed: no exif data
```

If there is no Exif data then `.Exif` should return nil.

bep (changes requested):

bep:
1. No, but I prefer

```go
if err != nil {
  return err
}
```

As in, I almost never used named return values (or whatever they're called).

2. That is a good question. The one I would be curious about, which I don't see a particulary good test for, is that the old library returned `Rat` values for some floats (e.g. f-stop)
3. The test coverage looks fine to me.
4. I'll update the dependency list in README.
5. You're welcome to simplify/remove the Walker struct if its not needed.

danedavid:
@bep Made a few changes. Fixed the build issue when file has no Exif data. Also, added type assertions & type-based test cases for tag values.

resources/images/exif/exif.go
```diff
@@ -113,127 +114,150 @@ func (d *Decoder) Decode(r io.Reader) (ex *Exif, err error) {
...
+func tryParseDate(s string) (time.Time, error) {
+	dateStr := strings.TrimRight(s, "\x00")
+	// TODO(bep): look for timezone offset, GPS time, etc.
+	timeZone := time.Local
+	return time.ParseInLocation(exifTimeLayout, dateStr, timeZone)
+}
+
+func processRational(n int64, d int64) interface{} {
+	rat := big.NewRat(n, d)
+	if n != 1 {
```
danedavid, line 171: @bep This logic existed in the previous code, making the value of rational exif tags either `float64` or `big.Rat` type depending on the numerator. Have copied the logic as such and added corresponding tests; however, let me know if returning just one of the types (`float64` or `big.Rat`) all the time makes more sense.

bep:
Hey,

It still fails for me in the case of an image with no Exif.

What I suggest is that you make:

```go
Decode(r io.Reader) (ex *Exif, err error)
```

Into

```go
Decode(r io.Reader) (*Exif, error)
```

And remove the panic recovery at the top.

And then you need to explicitly handle all errors, including `NewIfdMappingWithStandard`. I'm not sure where to set the line between "no exif data" and a real error that the user should/could do something about.

The above also suggests that we miss out on a test case for a JPG without Exif metadata ...

danedavid:
Hi @bep
I'd run `hugo` on the project you'd mentioned and verified, but apparently I seem to have made a mistake while building. Apologies for that.

I've added a new test case by stripping `sunset.jpg` of its EXIF data (using https://www.verexif.com/en/). I had to add a PNG check & extra dependancy to satisfy the new test case AND this one:
https://github.com/gohugoio/hugo/blob/fcd63de3a54fadcd30972654d8eb86dc4d889784/resources/images/exif/exif_test.go#L65
Hope it doesn't look clumsy.

danedavid:
Hi @bep, were you able to look into this?

bep:
@danedavid looking at it now ...

bep:
This works great now; I'm going to rebase and merge this in #8761 -- thanks!

Issue #8152, opened 2021-01-18
better pandoc support


jirib:
Hi,

it would be nice to have better pandoc integration as I feel commonmark has limitations and pandoc markdown can do much more things. Examples:

- ability to have TOC made by pandoc, just like asciidoctor support (`enableTOC: false` - to pass `--toc` to pandoc, `preserveTOC = false` - to override pandoc TOC html code `<nav id="TOC" role="doc-toc">`with hugo `.TableOfContent` variable; it would be probably nice also to be able to define css selector to find pandoc TOC as one could change pandoc template.
- extra args, eg. to define `--template <file>` or `--data-dir <dir>` ...

jwflory:
If better Pandoc support means that my `toc` appear in AsciiDoc-based HTML, that would be a big win. I see it appear in my local previews with the [Asciidoctor.js extension](https://github.com/asciidoctor/asciidoctor-browser-extension), but it would be nice for the table-of-contents attribute to make it to the generated HTML by Hugo for my site.

jirib:
> If better Pandoc support means that my `toc` appear in AsciiDoc-based HTML, that would be a big win. I see it appear in my local previews with the [Asciidoctor.js extension](https://github.com/asciidoctor/asciidoctor-browser-extension), but it would be nice for the table-of-contents attribute to make it to the generated HTML by Hugo for my site.

Have you tried this?

> By default, Hugo removes the table of contents generated by Asciidoctor and provides it through the built-in variable .TableOfContents to enable further customization and better integration with the various Hugo themes. This option can be set to true to preserve Asciidoctor’s TOC in the generated page.  
-- https://gohugo.io/content-management/formats/#external-helper-asciidoctor

jirib:
BTW, I more wanted to use pandoc markdown with all its extensions, not just TOC...

yaxu:
I had a look at making pandoc more configurable today, the first time I've written any go code but getting somewhere - I managed to get citeproc support working.

I noticed the mathjax support doesn't seem to actually work.. Pandoc does the processing but mathjax itself isn't loaded. I'm not a TeX math user so left that as-is.

yaxu:
As far as I can tell, the YAML header doesn't get passed to pandoc, is that intentional? I can add more parameters to this PR https://github.com/gohugoio/hugo/pull/8610 , but it would be nice to set things in the header too.

jirib:
FYI PR #7529

yaxu:
[...]

PR #8610, opened 2021-06-06
Pandoc - allow citeproc extension to be invoked, with bibliography.

yaxu:
The start of improved pandoc support by being able to pandoc's commandline arguments, ref #8152.


yaxu:
It looks like the venerable https://github.com/gohugoio/hugo/pull/7529 is a more developed + better PR for this.

Issue #8598, opened 2021-05-31
Add image dithering filter to build low-tech websites


rdelaage:
It would be great to be able to select a filter to dither image and reduce final website size in image processing. The [lowtech magazine](https://solar.lowtechmagazine.com/) use a filter to dither images. We could have the same in hugo by using [this library](https://github.com/makeworld-the-better-one/dither) and it could be selecting in [image processing](https://gohugo.io/content-management/image-processing/) filters.

bep:
We could add this **if we could do it by not adding another dependency**. I think the standard library has a simple dithering filter, which I think should be good enough for this.

diantanjung:
I want to solve this issue

diantanjung:
hi @bep 
Is there any documentation for debugging and testing for hugo project? I have changed in filter.go add add some file. But after I run mage install, the change does not work. Is I miss something?

diantanjung:
Ok, Please explain the features of the dithering image. If any input and output images will be very helpful.

bep:
I don't have any stake in this other than that I'll most likely end up maintaining what's added. But I was originally thinking about using `draw.FloydSteinberg` ... but I haven't researched this.

https://golang.org/pkg/image/draw/

diantanjung:
https://github.com/disintegration/gift/issues/25

diantanjung:
hi bep, Could I use `draw.FloydSteinberg` (but I modified it) to implement `gift.Filter`? 
Or Is there another approach?

thanks

PR #8628, opened 2021-06-10
add image dithering filter #8598

diantanjung:
Fix this issue [#8598](https://github.com/gohugoio/hugo/issues/8598)

You can set the dither algorithm and palette colors to use with a 3 or 6 digit hex code starting with #. If you set palettes only one color, it will add Black and White automatically. Use png file for best result. List algorithm :
- FloydSteinberg (default)
- JarvisJudiceNinke
- Stucki
- Atkinson
- Burkes
- Sierra
- TwoRowSierra
- SierraLite

example :
`{{ $img = $img.Filter (images.Dither "Stucki #ffffff #8D1900") }}`
`{{ $img = $img.Filter (images.Dither "FloydSteinberg #ffffff #8D1900 #AD8359 #508A00 #1A6C82") }}`

### DIFF
--- a/resources/images/dither.go
+++ b/resources/images/dither.go
@@ -1,0 +1,179 @@
+// Copyright 2020 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


bep:
The code provide here are entirely written by you?

diantanjung:
No, I modified from https://github.com/brouxco/dithering

diantanjung:
hi @bep 
is there I need to improve the code?

bep:
Thanks, but this was not wha I had in mind when I said that we should try to avoid adding a new dependency for this. It may be OK vs the upstream license, but it's not code that I want to maintain here.

I was originally thinking about using any existing dithering filter in Go's stdlib. Let's discuss this on the original issue.

PR #8636, opened 2021-06-12
Move time notification to after rebuild errors

### LINKED ISSUE #8403: Make sure "Total in ..." is the last thing printed to console
With `hugo serve` and file watching enabled, Hugo indicates a build run has started with the message "`Change detected, rebuilding site.`".

Hugo does not appear to print an equally consistent message to indicate it is no longer building the site.

On successful builds, the last line is usually something like "`Total in # ms`".

However, error messages may be printed after this line.

---

The motivation for this is to write a `problemMatcher` for a [Visual Studio Code Task][].

VSCode Tasks support a long-running, background task. The `problemMatcher` parses the stdout of the executed command, and uses regular expressions to collect information about errors.

VSCode must know when a build run is started, and when it ends. After a build run, it populates the UI with error information.

`endsPattern` is a regular expression that matches whichever line is printed at the end of a build run.

I don't know of one that works in a general case.

My current working attempt is:

```json
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558
    // for the documentation about the tasks.json format
    "version": "2.0.0",
    "tasks": [
        {
            "label": "preview (hugo server)",
            "type": "process",
            "command": "hugo",
            "args": [
                "serve",
                "--bind", "0.0.0.0",
                "--port", "1313",
                "--source", "${workspaceFolder}"
            ],
            "detail": "Generate a live preview of the website",
            "echoCommand": true,
            "isBackground": true,
            "isBuildCommand": true,
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "presentation": {
                "clear": false,
                "focus": false,
                "panel": "shared",
                "showReuseMessage": false,
            },
            "runOptions": {
                "instanceLimit": 1,
                "runOn": "folderOpen",
            },
            "suppressTaskName": true,
            "problemMatcher": {
                "owner": "external",
                "applyTo": "allDocuments",
                "fileLocation": "absolute",
                "background": {
                    "activeOnStart": true,
                    "beginsPattern": "Change detected",
                    "endsPattern": "Build finished",
                },
                "pattern": {
                    "kind": "location",
                    "regexp": "^\\s*?([A-Z]+)\\s+[\\d/]+\\s+[\\d:]+\\s(.*?\"((.*):(\\d+):(\\d+))\".*)$",
                    "severity": 1,
                    "file": 4,
                    "line": 5,
                    "column": 6,
                    "message": 2,
                }
            }
        }
    ]
}
```

[Visual Studio Code Task]: <https://code.visualstudio.com/docs/editor/tasks> "Documentation and Guide on Visual Studio Code Tasks"

jhollowe:
This allows error parsers (VSCode problem matchers) to use the time notification as bounds for detecting errors

(I've never coded in Go before, so if there is a better way to do this, I'm all ears!)

(partially) closes #8403

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -162,14 +162,16 @@ Complete documentation is available at http://gohugo.io/.`,
 				return nil
 			}
 
+			// prevent cobra printing error so it can be handled here (before the timeTrack prints)
+			cmd.SilenceErrors = true
+
 			c, err := initializeConfig(true, cc.buildWatch, &cc.hugoBuilderCommon, cc, cfgInit)
 			if err != nil {
+				cmd.PrintErrln("Error:", err.Error())
 				return err
 			}
 			cc.c = c
 
-			// prevent cobra printing error so it can be handled here (before the timeTrack prints)
-			cmd.SilenceErrors = true
 			err = c.build()
 			if err != nil {
 				cmd.PrintErrln("Error:", err.Error())
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -168,7 +168,13 @@ Complete documentation is available at http://gohugo.io/.`,
 			}
 			cc.c = c
 
-			return c.build()
+			// prevent cobra printing error so it can be handled here (before the timeTrack prints)
+			cmd.SilenceErrors = true
+			err = c.build()
+			if err != nil {
+				cmd.PrintErrln("Error:", err.Error())
+			}
+			return err
 		},
 	})
 
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -735,7 +735,6 @@ func (c *commandeer) handleBuildErr(err error, msg string) {
 }
 
 func (c *commandeer) rebuildSites(events []fsnotify.Event) error {
-	defer c.timeTrack(time.Now(), "Total")
 
 	c.buildErr = nil
 	visited := c.visitedURLs.PeekAllSet()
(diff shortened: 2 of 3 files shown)


jhollowe:
I'd also like all the other `timeTrack()` calls to be after the error handling, but the errors are printed out within cobra's ExecuteC() ([called in commands/hugo.go](https://github.com/gohugoio/hugo/blob/76c95f55a5d18290baa7f23667161d4af9fb9b53/commands/hugo.go#L88)). This would need more understanding of the codebase than I have to print the error out at the correct level without messing with how other errors bubble up.

bep:
FYI: There is an unrelated test failure on GitHub action ... I'm trying to fix that in another branch.

jhollowe:
That most recent commit seems to handle all the errors I could think of around the `timeTrack()` calls (build, server initial build, server partial build, server full build).

I don't know if this is the best code to make it do this or if there are other features that would simplify this, but at least its a start.

bep:
It looks correct, but I quickly tested by introducing an error in `config.toml`:

```bash
➜  hugoDocs git:(master) ✗ hugo
Total in 0 ms
Error: "/Users/bep/dev/go/gohugoio/hugoDocs/config.toml:15:1": unmarshal failed: Near line 15 (last key parsed 'pluralizeListTitles'): expected value but found "falses" instead
```

jhollowe:
Yeah, I was trying to look into the config errors. I had thought it was changed correctly in my testing (with a YAML config), but maybe I am mistaken. I'm having internet issues, so it will be a few days before I can look into this more.

jhollowe:
That commit should fix everything. 

Since most of the errors are now handled in Hugo and not cobra, I was thinking that they could be refactored to use the nicely formatted output of the logger. However, this should be a separate PR and I don't even know if it would be good to change the format of the errors at this point. If there are tools relying on the current format, changing to using the logger would be a breaking change.

PR #8639, opened 2021-06-13
Doorstop file

Time-dragon:
So that the "scripts" folder does not skip to the whatever it is.


Time-dragon:
@jmooring Why did you close this?

davidsneighbour:
Probably because there is no reference to an issue where this implementation is discussed. There is also no concept of a "Doorstop" in Hugo or Go.

PR #8678, opened 2021-06-20
Fix date format in schema and opengraph templates

### LINKED ISSUE #8671: Internal schema templates display &#43; instead of + for dates with positive offsets
These were fixed with #5887 and #7495, then broken again with #8256.

PR to follow...

jmooring:
Fixes #8671


jmooring:
Please merge when you have a moment. This fixes a regression and includes tests to make sure we don't break it again in the future. Thanks.

Issue #8519, opened 2021-05-10
Return nil on EXIF error, but log as WARNING and Include image path


jmooring:
When attempting to read EXIF data from an image with invalid/corrupt EXIF data, the error message does not include the path to the offending image. For example:

> Error: Error building site: failed to render pages: render of "page" failed: "/home/user/code/hugo-testing/layouts/_default/single.html:7:12": execute of template failed: template: _default/single.html:7:12: executing "main" at <.Exif>: error calling Exif: metadata init failed: exif: decode failed (zero length tag value) 

For a site with many images, it may require a lot of effort (bisection) and/or expertise to identify the offending image. We recently ran into this again with <https://discourse.gohugo.io/t/32684>.

It would helpful to include the path to the image (not just the image name) in the EXIF decode error message.

jmooring:
Minimal example:

```bash
git clone --single-branch -b hugo-github-issue-8519 https://github.com/jmooring/hugo-testing hugo-github-issue-8519
cd hugo-github-issue-8519
hugo server
```

jmooring:
Fixed with https://github.com/gohugoio/hugo/pull/8684

PR #8684, opened 2021-06-22
exif: Log warning for metadata decode error

IveGotNorto:
The error is logged as a WARNING and nil is returned.

see #8519

### DIFF
--- a/resources/image.go
+++ b/resources/image.go
@@ -112,8 +112,8 @@ func (i *imageResource) getExif() *exif.Exif {
 
 			x, err := i.getSpec().imaging.DecodeExif(f)
 			if err != nil {
-				i.metaInitErr = err
-				return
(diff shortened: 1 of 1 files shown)


jmooring:
@IveGotNorto Is there anyway to include the full path to the offending image? I have updated the minimal example for https://github.com/gohugoio/hugo/issues/8519 to include an additional assets mount (see config.toml). The resulting warning from this PR is:

```text
WARN 2021/06/29 10:32:46 Unable to decode Exif metadata from image: /images/broken.jpg
WARN 2021/06/29 10:32:46 Unable to decode Exif metadata from image: /images/c.jpg
```

It would be helpful if the warnings were:

```
WARN 2021/06/29 10:32:46 Unable to decode Exif metadata from image: /foo/images/broken.jpg
WARN 2021/06/29 10:32:46 Unable to decode Exif metadata from image: /assets/images/c.jpg
```

I don't know is this is possible. What you have done so far is great. Thank you.

PR #8716, opened 2021-06-30
Fix tab selection of disabled items in internal pagination template

### LINKED ISSUE #8715: Tab-selecting disabled pagination navigation items
The internal template for pagination generates navigation elements that are tab-selectable in disabled state.

BTW: This seems to be a regression since Hugo 0.80

<pre>
$ hugo version
hugo v0.84.0-2C4689F7B linux/amd64 BuildDate=2021-06-18T17:14:27Z VendorInfo=gohugoio
</pre>

raoulb:
Fixes #8715


jmooring (changes requested):
Run tests with `mage -v check`. It will fail.
You also need to update:
<https://github.com/gohugoio/hugo/blob/master/hugolib/embedded_templates_test.go#L155>

jmooring:
Please run `mage -v check`. The test still fails. You need to run tests locally before submitting PRs.

jmooring (approved):

jmooring:
@raoulb

When you ran `mage generate` before testing locally, it updated `tpl/tplimpl/embedded/templates.autogen.go`. You need to commit this change as well. When I update embedded templates...

```
mage generate
go install -tags extended
mage -v check
```

raoulb:
Thanks for the explanation. I did not have a go development setup at hand.

Issue #8913, opened 2021-08-21
Use Chroma fallback when no language is specified


helfper:
Currently, when using code fences without language specified (and `guessSyntax` is disabled), no CSS class or style is added to `pre`/`code`. I think it would be more useful to use the fallback lexer instead -- like it happens when `guessSyntax` is used and the language isn't identified -- as it would use a consistent fore/background color without special CSS handling.

helfper:
This actually may be in https://github.com/yuin/goldmark-highlighting

jmooring:
Related to, or duplicate of, https://github.com/gohugoio/hugo/issues/8824.

sethjensen1:
You can do this using [code block render hooks](https://gohugo.io/render-hooks/code-blocks/).

See https://discourse.gohugo.io/t/add-option-to-use-default-chroma-lexer-for-code-fences-with-no-language/55368/2

jmooring:
Without a language specified in the fenced code block's info string, a fenced code block renders identically to an indented code block. Since you can’t specify a language with an indented block anyway, this is arguably the expected behavior. And changing the default behavior is a breaking change with high visibility.

I'm closing this; use a render hook if you need to set a default language:



<details open>
<summary><code>layouts/_markup/render-codeblock.html</code></summary>

```text
{{- $lang := "text" }}
{{- $options := .Options }}
{{- if transform.CanHighlight .Type }}
  {{- $lang = .Type }}
{{- else }}
  {{- $options = .Attributes }}
{{- end }}

{{- transform.Highlight .Inner $lang $options }}
```
</details>


Incidentally, in David Anson's Markdown linter (which is probably the most popular linter of its kind) rule [MD040][] is enabled by default, requiring a language in the info string of every fenced code block.



[MD040]: https://github.com/DavidAnson/markdownlint/blob/main/doc/md040.md

Issue #8824, opened 2021-07-29
Default Chroma lexer does not support linenos


ynadji:
### What version of Hugo are you using (`hugo version`)?

<pre>
¡ hugo version
hugo v0.86.0+extended linux/amd64 BuildDate=unknown
</pre>

### Does this issue reproduce with the latest release?
v0.86.0 appears to be the latest release, so yes.

### Issue
Here's a small snippet of Zeek (FKA Bro) code that I'd like syntax highlighted with line numbers:

```
{{< highlight bro "linenos=table,linenostart=1" >}}
hook Notice::policy(n: Notice::Info)
    {
    if ( n$note == NewKnownHost )
        add n$actions[Notice::ACTION_TELEGRAM];
    }
```

From the image below, we can see that the syntax highlighting is correct, but there are no line numbers. 

<img width="497" alt="Screen Shot 2021-07-29 at 10 52 04 AM" src="https://user-images.githubusercontent.com/6740/127514953-5be4da41-8a81-4d0e-8d23-fa43c95c51cd.png">

If I change the language to C, it works just fine (albeit with slightly crooked line numbers).

```
{{< highlight c "linenos=table,linenostart=1" >}}
hook Notice::policy(n: Notice::Info)
    {
    if ( n$note == NewKnownHost )
        add n$actions[Notice::ACTION_TELEGRAM];
    }
```

<img width="557" alt="Screen Shot 2021-07-29 at 10 52 26 AM" src="https://user-images.githubusercontent.com/6740/127514975-22417691-60d3-410f-b52d-6b51a143bc20.png">

I suspect this is because Chroma doesn't have a Bro/Zeek lexer, so maybe it's falling back to pygments? But I tried to find another language only in pygments and not in Chroma and couldn't reproduce the error. Would love some direction here so I could try to fix the issue.

bep:
>so maybe it's falling back to pygments?

No, it's falling back to "a default lexer"; I'm a little surprised that that does not support line numbers, though.

pamubay:
Is this related with code fences markdown attributes doesnt work on **chroma-unsupported** language?
Markdown attributes on code fences not parsed if the language is not supported by chroma:

```md
`​`​`bro {.class #id data-attr="val"} 
code
 code
`​`​`
``` 

render:
```html
<pre>
  <code class="language-bro">
    ...
  </code>
</pre>
```

expected: (markdown attributes injected) to `pre`
```html
<pre class="class" id="id" data-attr="val">
[...]

PR #8917, opened 2021-08-22
highlight: Add tabindex when code is not highlighted

helfper:
PR #8568 added `tabindex="0"` to most `<pre>` tags, but it's still missing when using code fences and goldmark-highlighting doesn't highlight that code block. This PR fixes it by making this scenario reuse the same code that was changed in the aforementioned PR.

### DIFF
--- a/markup/goldmark/convert.go
+++ b/markup/goldmark/convert.go
@@ -310,49 +310,52 @@ func newHighlighting(cfg highlight.Config) goldmark.Extender {
 		),
 
 		hl.WithWrapperRenderer(func(w util.BufWriter, ctx hl.CodeBlockContext, entering bool) {
-			l, hasLang := ctx.Language()
 			var language string
-			if hasLang {
+			if l, hasLang := ctx.Language(); hasLang {
 				language = string(l)
 			}
 
-			if entering {
-				if !ctx.Highlighted() {
-					w.WriteString(`<pre>`)
-					highlight.WriteCodeTag(w, language)
-					return
-				}
-
-				w.WriteString(`<div class="highlight`)
-
-				var attributes []ast.Attribute
-				if ctx.Attributes() != nil {
-					attributes = ctx.Attributes().All()
-				}
-
-				if attributes != nil {
-					class, found := ctx.Attributes().GetString("class")
-					if found {
-						w.WriteString(" ")
-						w.Write(util.EscapeHTML(class.([]byte)))
-
-					}
-					_, _ = w.WriteString("\"")
-					renderAttributes(w, true, attributes...)
+			if ctx.Highlighted() {
+				if entering {
+					writeDivStart(w, ctx)
 				} else {
-					_, _ = w.WriteString("\"")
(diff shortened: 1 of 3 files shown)


bep:
We have, I think, at least one other issue discussing using the default Chroma lexer when no lexer is given. Would that not fix this issue?

helfper:
I opened this upstream: https://github.com/yuin/goldmark-highlighting/issues/22

If that's accepted in a way that the code will always be highlighted (in the limit with the fallback lexer) and not as a flag, then yes, `ctx.Highlighted()` will always be true and the outer `else` -- which is where this change has an effect -- will never be executed.

bep:
OK, so if we enable `GuessLanguage` and pass a dummy lexer value that does not exist, that would work, or?

helfper:
The situations in which the code is not highlighted (`ctx.Highlighted() == false`) by goldmark-highlighting are:
* Attribute `nohl` is present, or
* `GuessLanguage` is false and the language is not set or Chroma doesn't have a lexer for it

If `GuessLanguage` is true and the attribute `nohl` is not set, it will always be highlighted. If Chroma can't guess the language, goldmark-highlighting just uses the fallback lexer.

bep:
If I'm reading the above correctly, we should get a better situation by setting the `GuessLanguage=true`.

helfper:
That's correct, but currently that's linked to Hugo's `markup.highlight.guessSyntax`, so it's controlled by the user. You may decide to turn it on by default though.

I think Chroma doesn't do much guessing still, but otherwise I'm usually not a big fan of guessing because of my experience with highlight.js, which does some weird highlightings when it tries to guess languages:
https://learn.netlify.app/en/basics/installation/#create-your-first-chapter-page
https://learn.netlify.app/en/cont/pages/#folders

bep:
>I think Chroma doesn't do much guessing still

It doesn't, I had a look some time ago.

bep:
OK, thinking about it -- I'll merge this, and then follow up this upstream.

helfper:
TBH the main reason I raised this PR was because I was just scrolling through the code trying to figure out where the recently raised issues about highlighting come from, and liked how nice and symmetric this part of the code turned out :smiley:

https://github.com/gohugoio/hugo/blob/7a15edafe240471c072d3548b72ccda0271ffd8f/markup/goldmark/convert.go#L318-L330

bep:
Yes, that's pretty code.

helfper:
Looking further into it, even the `<div>` part could possibly be moved to `highlight.go`, because there's some duplication with code already there. At this point, one may also start wondering if goldmark-highlighting actually adds much value to Hugo, since most of its logic already exists in `highlight.go`. I think the only missing part in `highlight.go` is probably the attributes parsing in code fences, but Hugo already does something similar for other types of blocks, so it may not be too difficult to plug that into `highlight.go`.

The problem with the current overlap is that issues with a similar symptom like #8824 and #8913 need to be fixed in two different places -- the former in Hugo, the latter in goldmark-highlighting.

PR #8939, opened 2021-08-30
site_renderer: add pprof profiler labels

mhansen:
When pprof CPU profiling, I just see a big stacktrace full of
'texttemplate' functions that are hard to relate back to my templates and
site layout.

My intention here is to be able to attribute CPU cycles and work done to
specific causes, and be able to slice the profile by a few dimensions.
I'm hoping this will make it easier to tell where Hugo is spending time.

This took inspiration from https://rakyll.org/profiler-labels/.

Here's an example, building Graphviz's docs,
using the interactive pprof tool, showing the time spent in each tag:

```
$ hugo --profile-cpu /tmp/pprof
$ go tool pprof /tmp/pprof
Type: cpu
Time: Aug 29, 2021 at 6:29pm (AEST)
Duration: 15.61s, Total samples = 24.82s (158.96%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) tags
 format: Total 17.5s
           17.2s (98.34%): HTML
         290.0ms ( 1.66%): print

 kind: Total 17.5s
         16.6s (94.96%): page
       790.0ms ( 4.52%): section
        90.0ms ( 0.52%): home

 section: Total 17.0s
            15.9s (93.40%): docs
          730.0ms ( 4.30%): Gallery
          170.0ms ( 1.00%): Misc
          120.0ms ( 0.71%): download
           80.0ms ( 0.47%): build
           20.0ms ( 0.12%): _pages

 target_filename: Total 17.5s
                  130.0ms ( 0.74%): /docs/outputs/cmap/index.html
                  120.0ms ( 0.69%): /docs/attrs/pencolor/index.html
                  110.0ms ( 0.63%): /docs/attr-types/string/index.html
                  110.0ms ( 0.63%): /docs/attrs/id/index.html
                  [...]

 template: Total 17.4s
             10.2s (58.67%): attrs/single.html
              2.2s (12.85%): output/single.html
              2.0s (11.76%): attr-types/single.html
              1.6s ( 8.93%): docs/single.html
              [...]
```

You can also use pprof's `--tagFocus` argument, like this:

```
(pprof) tagfocus=template=attrs/single.html
(pprof) top
Active filters:
   tagfocus=template=attrs/single.html
Showing nodes accounting for 3.83s, 15.43% of 24.82s total
Dropped 220 nodes (cum <= 0.12s)
Showing top 10 nodes out of 129
      flat  flat%   sum%        cum   cum%
     0.77s  3.10%  3.10%      2.17s  8.74%  runtime.mallocgc
     0.60s  2.42%  5.52%      7.96s 32.07%  reflect.Value.call
     0.56s  2.26%  7.78%      0.70s  2.82%  runtime.heapBitsSetType
     0.46s  1.85%  9.63%      0.46s  1.85%  reflect.name.name
     0.41s  1.65% 11.28%      2.39s  9.63%  reflect.(*rtype).MethodByName
     0.25s  1.01% 12.29%      0.25s  1.01%  reflect.(*rtype).Kind (partial-inline)
     0.24s  0.97% 13.26%      0.24s  0.97%  runtime.nextFreeFast (inline)
     0.21s  0.85% 14.10%      9.92s 39.97%  github.com/gohugoio/hugo/tpl/internal/go_templates/texttemplate.(*state).evalCall
     0.18s  0.73% 14.83%      0.18s  0.73%  runtime.resolveNameOff
     0.15s   0.6% 15.43%     10.16s 40.93%  github.com/gohugoio/hugo/tpl/internal/go_templates/texttemplate.(*state).evalPipeline
```

My hope is that you'll find this as useful as I have for tracing the cause of slowdowns.

Considerations:
- I haven't thought too much about what tags/labels are most useful -- it's possible I'm missing some, or I might have too many.
- I could also break out a subfunction for `renderPage`, as the indentation is getting fairly deep here.
- I haven't documented this, just like the --profile-cpu flag is undocumented
- I'm using `context.Background()` -- perhaps this should be `context.TODO()` until the `main` function passes a Context? It probably doesn't matter too much as this is a command line tool without RPC deadlines.
- Some parts of the hugo pipeline aren't within this `renderPages` function -- perhaps they could get tags too? But for my runs, most of the CPU time is within `renderPages`.
- the 'template' is the top-level template, but the top-level template can include other templates. I haven't captured that here.

### DIFF
--- a/hugolib/site_render.go
+++ b/hugolib/site_render.go
@@ -70,7 +70,7 @@ func (s *Site) renderPages(ctx *siteRenderContext) error {
 
 	for i := 0; i < numWorkers; i++ {
 		wg.Add(1)
-		go pageRenderer(ctx, s, pages, results, wg)
+		go pageRenderer(s, pages, results, wg)
 	}
 
 	cfg := ctx.cfg
@@ -101,7 +101,6 @@ func (s *Site) renderPages(ctx *siteRenderContext) error {
 }
 
 func pageRenderer(
-	ctx *siteRenderContext,
 	s *Site,
 	pages <-chan *pageState,
 	results chan<- error,
--- a/hugolib/site_render.go
+++ b/hugolib/site_render.go
@@ -14,8 +14,10 @@
 package hugolib
 
 import (
+	"context"
 	"fmt"
 	"path"
+	"runtime/pprof"
 	"strings"
 	"sync"
 
@@ -70,7 +72,7 @@ func (s *Site) renderPages(ctx *siteRenderContext) error {
 
 	for i := 0; i < numWorkers; i++ {
 		wg.Add(1)
-		go pageRenderer(s, pages, results, wg)
+		go pageRenderer(context.Background(), s, pages, results, wg)
 	}
 
 	cfg := ctx.cfg
@@ -101,6 +103,7 @@ func (s *Site) renderPages(ctx *siteRenderContext) error {
 }
 
 func pageRenderer(
+	ctx context.Context,
 	s *Site,
 	pages <-chan *pageState,
 	results chan<- error,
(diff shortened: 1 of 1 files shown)


mhansen:
For some more context (probably not necessary to read to understand this PR, but just linking them up), I wrote a blogpost about this investigation: https://www.markhansen.co.nz/attributing-go-cpu-profiles-with-pprof-tags/

moorereason:
Mark,
Cool PR.  I finally took a second to look at it.  Here are my initial thoughts:

1. First, I didn't know about the pprof tags.  If nothing else happens here, you at least taught me something new.
1. If we merge this, I'd keep it as a hidden option for now.  The average user won't know what to do with it.
2. The real value for most users is probably the tags summaries.  This PR tracks higher up in the pipeline than the existing template metrics tracking and gives you a different viewpoint of the site build.  If that is the main goal, we may be able to extend our metrics and avoid pprof altogether.
3. From your blog post, it sounds like this PR helped you see why your site was slow to build (syntax highlighting) but didn't actually help you make it build faster.  Have you found this PR to be of any help in optimizing your site?

mhansen:
> hidden option

Yeah, hidden option sounds good.

> The real value for most users is probably the tags summaries

Agree, given the lack of meaning I can intuit from the stack frames (it's all `texttemplate` and hard to relate that.

> Have you found this PR to be of any help in optimizing your site?

I was experimenting with pprof labels and had a slow Hugo build, but this didn't actually help me fix it yet (maybe if I dig in some more)

FWIW, I'm totally OK if this doesn't get merged. I just thought, it doesn't hurt, and perhaps this will be useful for other people profiling Hugo? It makes the existing cpu profile a bit more relatable.

I'm also considering a different approach here: instrumenting Hugo to output Chrome's Trace Event Format. It's really easy, just write some JSON, and then we could annotate the trace sections with semantic names like pages and template inclusions. Here's an example where I integrated it for Ansible, it didn't take much code: https://github.com/mhansen/ansible-trace

I think that could be really exciting, showing how time is spent in pages executing included templates, for example.

PR #8984, opened 2021-09-20
implement and test a render-document capability

zed-eh:
First time using go. First time contributing to any project on GitHub. Please feel free to inform if I'm not following process.

The proposed code adds the capability to create a markdown hook in layout/_default/_markup/render-document.html that hooks into goldmark render-document. The hook gets invoked twice: just before markdown processing starts; and just after markdown processing finishes. A boolean parameter indicates the reason for invocation.

I found this useful to set up some initial conditions before processing start, and to clean up when processing has completed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /hugo
 docs/public*
 /.idea
+/.vscode
 hugo.exe
 *.test
 *.prof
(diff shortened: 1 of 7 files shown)


bep:
I'm closing this, we need to discuss this in an issue before starting on an implementation.

PR #8991, opened 2021-09-24
tpl/tplimpl: Add .Site.Title to embedded opengraph template 

sean-au:
Added .Site.Title, to go alongside the existing .Site.Params.Title lookup for og:site_name.

.Site.Title has been given preference.

.Site.Title is specified in the hugo docs as the place to record the sites name https://gohugo.io/getting-started/configuration/#title

Its also configured by default when running "hugo new site" 

```TOML
# config.toml
baseURL = "http://example.org/"
languageCode = "en-us"
title = "My New Hugo Site"
```


sean-pwd:
That should work

PR #9016, opened 2021-10-09
Allow multiple plugins in the PostCSS options map

### LINKED ISSUE #9015: Unable to specify multiple plugins in the PostCSS options map
Our [PostCSS options](https://gohugo.io/hugo-pipes/postcss/#options) documentation states:

> **use [string]**
> List of PostCSS plugins to use

Currently you can only specify a single plugin:
- The `use` element of the options map will not accept a slice of plugins; it expects a string
- If you specify a list of plugins as you would using the postcss-cli, it treats the list as a single string

For example, neither of these work:

```text
{{ $css := resources.Get "css/main.css" | postCSS (dict "use" (slice "autoprefixer" "postcss-color-alpha")) }}
{{ $css := resources.Get "css/main.css" | postCSS (dict "use" "autoprefixer postcss-color-alpha") }}
```

You can work around this by (a) creating a postcss configuration file, or (b) creating a postcss object within the project's package.json file. 

But for simple setups, with just a few plugins using default plugin options, specifying the plugins inline is simpler.

Ref: <https://discourse.gohugo.io/t/how-to-list-postcss-plugins-to-use/31293>

jmooring:
Usage:
{{ $options := dict "use" "autoprefixer postcss-color-alpha" }}
{{ $style := resources.Get "main.css" | resources.PostCSS $options }}

Fixes #9015

### DIFF
--- a/docs/content/en/hugo-pipes/postcss.md
+++ b/docs/content/en/hugo-pipes/postcss.md
@@ -15,11 +15,9 @@ sections_weight: 40
 draft: false
 ---
 
-
 Any asset file can be processed using `resources.PostCSS` which takes for argument the resource object and a slice of options listed below. 
 
 The resource will be processed using the project's or theme's own `postcss.config.js` or any file set with the `config` option.
-
 
 ```go-html-template
 {{ $css := resources.Get "css/main.css" }}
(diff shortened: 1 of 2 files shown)


jmooring:
Rebase completed.

PR #9024, opened 2021-10-10
Pinned docker images by SHA

nathannaveen:
Pinned docker images by SHA instead of tags because tags could be moved while SHAs can’t.

https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies


jeanlucmongrain:
Pin by tag break the ability to build multi arch images, no?

PR #9030, opened 2021-10-12
This is very mac-centered, at least mention Linux

balderbach:
Thanks for the great project, trying to amend this in the docs. Am happy to do a larger change if wished.


balderbach:
Hi @bep - I changed the sentence, I think it's better that there is a link with the text "full install instructions", should be easier for people to spot and understand that the Brew method is just one of many.

Issue #4470, opened 2018-03-04
Add XML as a Data File Type


bzerangue:
It would be nice to provide XML as an available data type to load. 

Currently, Hugo accepts, TOML, YAML, JSON, and CSV. Is there any reason why XML is NOT amongst the available formats?

It would be great to add XML as an available data type for the `/data` folder.

rdwatters:
@bzerangue This would be fantastic. Not sure if something like https://github.com/beevik/etree would help said efforts...

fbaube:
+1 on this. Type "XML" could then act as a flag to invoke DOCTYPE parsing and optional validation.

kaushalmodi:
Having a `getXML` similar to `getJSON` would be useful in creating "planets" from feeds. Ref: https://discourse.gohugo.io/t/anyone-here-interested-in-a-hugo-planet/15092.

stephlocke:
This'll be great for surfacing RSS from other sources like Medium

raybellis:
+1 from me, too - I'd like to be able to include references from the RFC index, but it's only published in XML format.

fbaube:
I'm working on a Go library to parse mixed content. If anyone can propose an API (i.e. interfaces) that would be useful for Hugo, I could give it a whack.

stephlocke:
On the RSS specific side, I currently leverage a site that parses RSS to JSON for consumption

```
{{ $rssJ := getJSON "https://api.rss2json.com/v1/api.json?rss_url=https%3A%2F%2Fmedium.com%2Ffeed%2Fnightingale-hq" }}


{{ range $rssJ.items }}
    {{ $post := . }}
    <div class="col-12 col-md-6 col-lg-4 mb-2 mr-2 pb-8 blog" style="background-image: url({{ $post.thumbnail }});">
    <a href="{{ $post.link }}" target="_blank" alt="{{ $post.title }}">
    <h2>{{ $post.title }}</h2>
    </a>
    <h3>{{ $post.author }}</h3>
    </div>
{{ end }}
```

felicianotech:
If this request is too complex as is, what about simplifying it to a `getRSS` function? I know for me the need for `getXML` is simply so that I can have Hugo read in RSS feeds.

lhenze:
Agreed, this would be fantastic.  From the library tech perspective --  we are experimenting with using Hugo to generate Special Collections Finding Aids from Encoded Archival Description (EAD) -- an XML format widely used by libraries.

[...]

PR #9031, opened 2021-10-12
GetXML implementation using antchfx/xmlquery

vanbroup:
An XML data implementation of GetXML using https://github.com/antchfx/xmlquery 

Exposing the Node type which allows users to get sub elements, attributes and use xpath filters.

Fixing #4470

### DIFF
--- a/cache/filecache/filecache_config.go
+++ b/cache/filecache/filecache_config.go
@@ -44,6 +44,7 @@ var defaultCacheConfig = Config{
 const (
 	cacheKeyGetJSON = "getjson"
 	cacheKeyGetCSV  = "getcsv"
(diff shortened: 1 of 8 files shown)


bep:
Thanks for this,

The code looks good, but I'm not merging this now.

I have two comments about this:

* When we add XML as a data format in Hugo it should work as a new data format where the others work (more or less), e.g. `/data` and `transform.Unmarshal`.
* There is another issue that talks about replacing `getJSON` (and friends` with a `resources.GetRemote` (or something), which would be format agnostic (also hopefully).
* I'm not sure how I would expect an XML structure to look like in the templates, but it should be similar to what we have now and not require some special knowledge about some special API.

PR #9040, opened 2021-10-14
Updated Dependabot for GitHub actions

nathannaveen:
Updated Dependabot for GitHub actions. The actions aren't getting updated with the latest release.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


bep:
What actions?

coliff (approved):

bep:
@coliff for security reasons we, for the most parts anyhow, use commit hashes to tell which version we want of a given action. I don't see how that could work with Dependabot?

coliff:
@bep - yes, I see what you mean. If you don't mind manually keeping them up to date as you need to then I guess you can close this then. I don't know if Dependabot would update the commit hashes or not.

bep:
>If you don't mind manually keeping them up to date

I do mind that, but I much rather prefer to know what I'm running ...

PR #9044, opened 2021-10-15
Implement XML data support

### LINKED ISSUE #4470: Add XML as a Data File Type

vanbroup:
Implement XML data support

Sample usage:
```
{{ with resources.Get "https://www.w3schools.com/xml/note.xml" | transform.Unmarshal }}
{{ .note.body }}
{{ end }}
```

A common usage:
```
{{ with resources.Get "https://example.com/rss.xml" | transform.Unmarshal }}
    {{ range .rss.channel.item }}
        <strong>{{ .title | plainify | htmlUnescape }}</strong><br />
        <p>{{ .description | plainify | htmlUnescape }}</p>
        {{ $link := .link | plainify | htmlUnescape }}
        <a href="{{ $link }}">{{ $link }}</a><br />
        <hr>
    {{ end }}
{{ end }}
```

Closes #4470

### DIFF
--- a/docs/content/en/functions/transform.Unmarshal.md
+++ b/docs/content/en/functions/transform.Unmarshal.md
@@ -48,10 +48,6 @@ Example:
 
 ## XML data
 
-{{% warning "Order of XML elements" %}}
-The order of XML elements and data structures is not preserved. An XML element is an order-dependent collection of anonymous values, while a data structure is an order-independent collection of named values.
-{{% /warning %}}
-
 The following example lists the items of an RSS feed:
 
 ```
--- a/docs/content/en/functions/transform.Unmarshal.md
+++ b/docs/content/en/functions/transform.Unmarshal.md
@@ -1,6 +1,6 @@
 ---
 title: "transform.Unmarshal"
-description: "`transform.Unmarshal` (alias `unmarshal`) parses the input and converts it into a map or an array. Supported formats are JSON, TOML, YAML and CSV."
+description: "`transform.Unmarshal` (alias `unmarshal`) parses the input and converts it into a map or an array. Supported formats are JSON, TOML, YAML, XML and CSV."
 date: 2018-12-23
 categories: [functions]
 menu:
--- a/docs/content/en/functions/transform.Unmarshal.md
+++ b/docs/content/en/functions/transform.Unmarshal.md
@@ -48,11 +48,24 @@ Example:
 
 ## XML data
 
+As a convenience, Hugo allows you to access XML data in the same way that you access JSON, TOML, and YAML: you do not need to specify the root node when accessing the data.
+
+To get the contents of `<title>` in the document below, you use `{{ .message.title }}`:
+
+```
+<root>
+    <message>
+        <title>Hugo rocks!</title>
+        <description>Thanks for using Hugo</description>
+    </message>
+</root>
+```
+
 The following example lists the items of an RSS feed:
 
 ```
 {{ with resources.Get "https://example.com/rss.xml" | transform.Unmarshal }}
-    {{ range .rss.channel.item }}
+    {{ range .channel.item }}
         <strong>{{ .title | plainify | htmlUnescape }}</strong><br />
         <p>{{ .description | plainify | htmlUnescape }}</p>
         {{ $link := .link | plainify | htmlUnescape }}
--- a/docs/content/en/functions/transform.Unmarshal.md
+++ b/docs/content/en/functions/transform.Unmarshal.md
@@ -45,3 +45,23 @@ Example:
 ```go-html-template
 {{ $csv := "a;b;c" | transform.Unmarshal (dict "delimiter" ";") }}
 ```
+
+## XML data
+
+{{% warning "Order of XML elements" %}}
+The order of XML elements and data structures is not preserved. An XML element is an order-dependent collection of anonymous values, while a data structure is an order-independent collection of named values.
+{{% /warning %}}
+
+The following example lists the items of an RSS feed:
+
+```
+{{ with resources.Get "https://example.com/rss.xml" | transform.Unmarshal }}
+    {{ range .rss.channel.item }}
+        <strong>{{ .title | plainify | htmlUnescape }}</strong><br />
+        <p>{{ .description | plainify | htmlUnescape }}</p>
+        {{ $link := .link | plainify | htmlUnescape }}
+        <a href="{{ $link }}">{{ $link }}</a><br />
+        <hr>
+    {{ end }}
+{{ end }}
+```
--- a/docs/content/en/templates/data-templates.md
+++ b/docs/content/en/templates/data-templates.md
@@ -6,7 +6,7 @@ date: 2017-02-01
 publishdate: 2017-02-01
 lastmod: 2017-03-12
 categories: [templates]
-keywords: [data,dynamic,csv,json,toml,yaml]
+keywords: [data,dynamic,csv,json,toml,yaml,xml]
 menu:
   docs:
     parent: "templates"
@@ -20,7 +20,7 @@ toc: true
 
 <!-- begin data files -->
 
-Hugo supports loading data from YAML, JSON, and TOML files located in the `data` directory in the root of your Hugo project.
+Hugo supports loading data from YAML, JSON, XML, and TOML files located in the `data` directory in the root of your Hugo project.
 
 {{< youtube FyPgSuwIMWQ >}}
 
@@ -28,7 +28,7 @@ Hugo supports loading data from YAML, JSON, and TOML files located in the `data`
 
 The `data` folder is where you can store additional data for Hugo to use when generating your site. Data files aren't used to generate standalone pages; rather, they're meant to be supplemental to content files. This feature can extend the content in case your front matter fields grow out of control. Or perhaps you want to show a larger dataset in a template (see example below). In both cases, it's a good idea to outsource the data in their own files.
 
-These files must be YAML, JSON, or TOML files (using the `.yml`, `.yaml`, `.json`, or `.toml` extension). The data will be accessible as a `map` in the `.Site.Data` variable.
+These files must be YAML, JSON, XML, or TOML files (using the `.yml`, `.yaml`, `.json`, `.xml`, or `.toml` extension). The data will be accessible as a `map` in the `.Site.Data` variable.
 
 ## Data Files in Themes
 
@@ -95,7 +95,7 @@ Discover a new favorite bass player? Just add another `.toml` file in the same d
 
 ## Example: Accessing Named Values in a Data File
 
-Assume you have the following data structure in your `User0123.[yml|toml|json]` data file located directly in `data/`:
+Assume you have the following data structure in your `User0123.[yml|toml|xml|json]` data file located directly in `data/`:
 
 {{< code-toggle file="User0123" >}}
 Name: User0123
@@ -232,6 +232,7 @@ If you change any local file and the LiveReload is triggered, Hugo will read the
 * [YAML Spec][yaml]
 * [JSON Spec][json]
 * [CSV Spec][csv]
+* [XML Spec][xml]
 
 [config]: /getting-started/configuration/
 [csv]: https://tools.ietf.org/html/rfc4180
@@ -247,3 +248,4 @@ If you change any local file and the LiveReload is triggered, Hugo will read the
 [variadic]: https://en.wikipedia.org/wiki/Variadic_function
 [vars]: /variables/
 [yaml]: https://yaml.org/spec/
+[xml]: https://www.w3.org/XML/
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/bep/golibsass v1.0.0
 	github.com/bep/gowebp v0.1.0
 	github.com/bep/tmc v0.5.1
+	github.com/clbanning/mxj/v2 v2.5.5
 	github.com/cli/safeexec v1.0.0
 	github.com/disintegration/gift v1.2.1
 	github.com/dustin/go-humanize v1.0.0
--- a/go.sum
+++ b/go.sum
@@ -144,6 +144,8 @@ github.com/cheekybits/is v0.0.0-20150225183255-68e9c0620927/go.mod h1:h/aW8ynjgk
 github.com/chzyer/logex v1.1.10/go.mod h1:+Ywpsq7O8HXn0nuIou7OrIPyXbp3wmkHB+jjWRnGsAI=
 github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5PlCu98SY8svDHJxuZscDgtXS6KTTbou5AhLI=
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
+github.com/clbanning/mxj/v2 v2.5.5 h1:oT81vUeEiQQ/DcHbzSytRngP6Ky9O+L+0Bw0zSJag9E=
+github.com/clbanning/mxj/v2 v2.5.5/go.mod h1:hNiWqW14h+kc+MdF9C6/YoRfjEJoR3ou6tn/Qo+ve2s=
 github.com/cli/safeexec v1.0.0 h1:0VngyaIyqACHdcMNWfo6+KdUYnqEr2Sg+bSP1pdF+dI=
 github.com/cli/safeexec v1.0.0/go.mod h1:Z/D4tTN8Vs5gXYHDCbaM1S/anmEDnJb1iW0+EJ5zx3Q=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
--- a/hugolib/resource_chain_test.go
+++ b/hugolib/resource_chain_test.go
@@ -872,16 +872,19 @@ Publish 2: {{ $cssPublish2.Permalink }}
 {{ $toml := "slogan = \"Hugo Rocks!\"" | resources.FromString "slogan.toml" | transform.Unmarshal }}
 {{ $csv1 := "\"Hugo Rocks\",\"Hugo is Fast!\"" | resources.FromString "slogans.csv" | transform.Unmarshal }}
 {{ $csv2 := "a;b;c" | transform.Unmarshal (dict "delimiter" ";") }}
+{{ $xml := "<?xml version=\"1.0\" encoding=\"UTF-8\"?><note><to>You</to><from>Me</from><heading>Reminder</heading><body>Do not forget XML</body></note>" | transform.Unmarshal }}
 
 Slogan: {{ $toml.slogan }}
 CSV1: {{ $csv1 }} {{ len (index $csv1 0)  }}
 CSV2: {{ $csv2 }}		
+XML: {{ $xml.note.body }}
 `)
 		}, func(b *sitesBuilder) {
 			b.AssertFileContent("public/index.html",
 				`Slogan: Hugo Rocks!`,
 				`[[Hugo Rocks Hugo is Fast!]] 2`,
 				`CSV2: [[a b c]]`,
+				`XML: Do not forget XML`,
 			)
 		}},
 		{"resources.Get", func() bool { return true }, func(b *sitesBuilder) {
--- a/hugolib/resource_chain_test.go
+++ b/hugolib/resource_chain_test.go
@@ -591,9 +591,9 @@ func TestResourceChains(t *testing.T) {
 
 		case "/mydata/xml1.xml":
 			w.Write([]byte(`
-				<hello>
-					<world>Hugo Rocks!</<world>
-				</hello>`))
+					<hello>
+						<world>Hugo Rocks!</<world>
+					</hello>`))
 			return
 
 		case "/mydata/svg1.svg":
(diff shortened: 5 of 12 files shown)


jmooring:
Yes, this is much easier to use, but it does not address the comments on the previous PR:

> When we add XML as a data format in Hugo it should work as a new data format where the others work (more or less), e.g. /data and transform.Unmarshal.

Meaning, I should be able to drop foo.xml into the data directory, then do something like:

```text
{{ range $k, $v := site.Data.foo.note }}
  {{ $k }} = {{ $v }}<br>
{{ end }}
```

Or:

```text
{{ $xml := "<?xml version="1.0" encoding="UTF-8"?><note><to>Tove</to><from>Jani</from><heading>Reminder</heading><body>Don't forget me this weekend!</body></note>" }}
{{ $data := transform.Unmarshal $xml }}
```

vanbroup:
I assumed that worked as it's using the same logic as the other implementations, but let me test and see if that can be addressed.

vanbroup:
Fixed. Sorry, I should have run a full test.

jmooring:
This does what I expect it to do, with this known limitation/caveat:

> Mapping between XML elements and data structures is inherently flawed: an XML element is an order-dependent collection of anonymous values, while a data structure is an order-independent collection of named values.

See <https://pkg.go.dev/encoding/xml#pkg-note-BUG>

jmooring:
@vanbroup This branch has conflicts that must be resolved: https://github.com/gohugoio/hugo/pull/9044/conflicts

vanbroup:
> @vanbroup This branch has conflicts that must be resolved: https://github.com/gohugoio/hugo/pull/9044/conflicts

Rebased on master and resolved conflict.

vanbroup:
@jmooring @bep anything I can do to speedup the review/merge?

bep:
First, as I said in a previous PR about something similar: We're not adding another "getSomeFormat" function -- we plan to do a `getRemoteResource` where we would support the Hugo supported data formats.

Second, any XML support would require a solid and well maintained upstream library. I haven't looked at this one in detail, but we're not depending on a minor fork of the original library.

vanbroup:
Thanks for your feedback @bep!

I have switched the xml lib to github.com/clbanning/mxj, the one you listed here https://github.com/gohugoio/hugo/issues/4470#issuecomment-449563597, usage has not changed but mxj exposes the interface as `mxj.Map` of type `map[string]interface{}` by default. If you prefer to remap the interface to a clean `map[string]interface{}`.

I would love to remove the `getXML` but that would require some guidance on how you would like this to be implemented.

Do you want to expose `getRemote`: https://github.com/gohugoio/hugo/blob/282f1aa3db9f6420fdd360e46db1ffadd5b083a1/tpl/data/resources.go#L39 

so that we could do something like this?
```
{{ with transform.Unmarshal (getRemote "https://www.w3schools.com/xml/note.xml") }}
{{ .note.body }}
{{ end }}
```
or should `getRemote` run `transform.Unmarshal` automatically?

can I implement `getRemote` as part of this PR or shall I create a separate PR and rebase this one after that has landed?

vanbroup:
@bep pushed another update removing `getXML` in favor of a generic `getResource`, let me know if you like this direction so I can cleanup further.

Some examples:
```
{{ range $k, $v := site.Data.filename.note }}
  {{ $k }} = {{ $v }}<br>
{{ end }}
```

```
{{ $xml := `<?xml version="1.0" encoding="UTF-8"?><note><to>Tove</to><from>Jani</from><heading>Reminder</heading><body>Don't forget me this weekend!</body></note>` }}
{{ $data := transform.Unmarshal $xml }}
{{ $data.note.body }}
```

```
{{ with transform.Unmarshal ( getResource "https://www.w3schools.com/xml/note.xml" ) }}
{{ .note.body }}
{{ end }}
```

```
{{ with transform.Unmarshal ( getResource "https://example.com/rss.xml" ) }}
{{ range .rss.channel.item }}
    <strong>{{ .title | plainify | htmlUnescape }}</strong><br />
    <p>{{ .description | plainify | htmlUnescape }}</p>
    {{ $link := .link | plainify | htmlUnescape }}
    <a href="{{ $link }}">{{ $link }}</a><br />
    <hr>
{{ end }}
{{ end }}
```

vanbroup:
@bep updated this pull request to only handle the XML data and let #9185 handle the remote resources.

Let me know if you would like to see any further changes.

```
{{ with resources.Get "https://example.com/rss.xml" | transform.Unmarshal }}
    {{ range .rss.channel.item }}
        <strong>{{ .title | plainify | htmlUnescape }}</strong><br />
        <p>{{ .description | plainify | htmlUnescape }}</p>
        {{ $link := .link | plainify | htmlUnescape }}
        <a href="{{ $link }}">{{ $link }}</a><br />
        <hr>
    {{ end }}
{{ end }}
```

vanbroup:
@bep This PR should be in a good state now, let me know if you think that this can be merged or if you would like to see any changes.

docs/content/en/functions/transform.Unmarshal.md
```diff
@@ -45,3 +45,23 @@ Example:
 ```go-html-template
 {{ $csv := "a;b;c" | transform.Unmarshal (dict "delimiter" ";") }}
 ```
+
+## XML data
+
+{{% warning "Order of XML elements" %}}
+The order of XML elements and data structures is not preserved. An XML element is an order-dependent collection of anonymous values, while a data structure is an order-independent collection of named values.
+{{% /warning %}}
```
bep, line 53: I'm not totally sure I follow this logic.

Given the example below, do I not get an ordered slice of items?

```xml
<channel>
			<title>Example feed</title>
			<item>
				<title>Item 1</title>
			</item>
	     <item>
				<title>Item 2</title>
			</item>
		</channel>
```

vanbroup: I added this note following the comment of @jmooring https://github.com/gohugoio/hugo/pull/9044#issuecomment-947645742 and my reply below that.

> @vanbroup
> 
> > That limitation doesn't seem to be relevant to the usage in Hugo, does it?
> 
> For ~the majority of~ some use cases I don't think it is relevant. But technically, a XML node has a name, value, and position. When you convert to a map you lose the position.
> 
> data/foo.xml
> 
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> <item>
>   <b>This is the first node</b>
>   <a>This is the second node</a>
>   <c>This is the third node</c>
> </item>
> ```
> 
> template
> 
> ```
> <pre>{{ jsonify (dict "indent" "  ") site.Data.foo }}</pre>
> ```
> 
> output
> 
> ```
> {
>   "item": {
>     "a": "This is the second node",
>     "b": "This is the first node",
>     "c": "This is the third node"
>   }
> }
> ```

bep: Sure, I understand this re what I consider to be _map values_, but what about this example:

```xml
<channel>
<title>Example feed</title>
<item>
<title>Item 1</title>
</item>
<item>
<title>Item 2</title>
</item>
</channel>
```

vanbroup: Correct, I tested to be sure, but the order of a slice is preserved:

```json 
{
  "channel": {
    "item": [
      {
        "title": "Item 1"
      },
      {
        "title": "Item 2"
      }
    ],
    "title": "Example feed"
  }
}
```
But the order of the values in an item not:
```xml
<channel>
    <title>Example feed</title>
    <item>
        <title>Item 1</title>
        <description>Description 1</description>
    </item>
    <item>
        <title>Item 2</title>
        <description>Description 2</description>
    </item>
</channel>
```
Results in a swapped order of the title and description:
```json
{
  "channel": {
    "item": [
      {
        "description": "Description 1",
        "title": "Item 1"
      },
      {
        "description": "Description 2",
        "title": "Item 2"
      }
    ],
    "title": "Example feed"
  }
}
```

I suggest we remove the warning as the order of a map is not relevant to how the values are used in Hugo.

vanbroup, line 53:
```suggestion
```

bep: We should remove the warning (as it confuses more than it helps), but the order is not irellevant:

* When you iterate a map in a Go template, the keys are sorted (I assume ascending order)
* When you marshal a Go map to JSON, the keys are sorted

What I then would expect for XML is, most importantly, that it is consistent.

We have a `Remarshal` func that has a roundtrip test that it would be good if you could add XML to to verify that these formats can be converted to/from.

vanbroup: The XML specification requires a root element, parsing without a root element works but not for lists (such a resources in this example).

```xml
<root>
  <resources>
    <params>
      <byline>picasso</byline>
    </params>
    <src>**image-4.png</src>
    <title>The Fourth Image!</title>
  </resources>
  <resources>
    <name>my-cool-image-:counter</name>
    <params>
      <byline>bep</byline>
    </params>
    <src>**.png</src>
    <title>TOML: The Image #:counter</title>
  </resources>
  <title>Test Metadata</title>
</root>
```

`TestRemarshal` fails because it will encode with the unexpected root element:
```
--- FAIL: TestRemarshal (0.01s)
    remarshal_test.go:128: [XML => yaml] Expected 
        resources:
        - params:
            byline: picasso
          src: '**image-4.png'
          title: The Fourth Image!
        - name: my-cool-image-:counter
          params:
            byline: bep
          src: '**.png'
          title: 'TOML: The Image #:counter'
        title: Test Metadata
        
        got
        root:
          resources:
          - params:
              byline: picasso
            src: '**image-4.png'
            title: The Fourth Image!
          - name: my-cool-image-:counter
            params:
              byline: bep
            src: '**.png'
            title: 'TOML: The Image #:counter'
          title: Test Metadata
        
        diff:
        [root:]
    remarshal_test.go:128: [XML => json] Expected 
        {
           "resources": [
              {
                 "params": {
                    "byline": "picasso"
                 },
                 "src": "**image-4.png",
                 "title": "The Fourth Image!"
              },
              {
                 "name": "my-cool-image-:counter",
                 "params": {
                    "byline": "bep"
                 },
                 "src": "**.png",
                 "title": "TOML: The Image #:counter"
              }
           ],
           "title": "Test Metadata"
        }
        
        got
        {
           "root": {
              "resources": [
                 {
                    "params": {
                       "byline": "picasso"
                    },
                    "src": "**image-4.png",
                    "title": "The Fourth Image!"
                 },
                 {
                    "name": "my-cool-image-:counter",
                    "params": {
                       "byline": "bep"
                    },
                    "src": "**.png",
                    "title": "TOML: The Image #:counter"
                 }
              ],
              "title": "Test Metadata"
           }
        }
        
        diff:
        ["root":]
    remarshal_test.go:128: [XML => toml] Expected 
        title = 'Test Metadata'
        		
        [[resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [resources.params]
            byline = 'picasso'
        
        [[resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [resources.params]
            byline = 'bep'
        
        got
        [root]
          title = 'Test Metadata'
        [[root.resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [root.resources.params]
            byline = 'picasso'
        
        [[root.resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [root.resources.params]
            byline = 'bep'
        
        
        
        
        diff:
        [[root] [[resources]] [resources.params]]
    remarshal_test.go:128: [XML => TOML] Expected 
        title = 'Test Metadata'
        		
        [[resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [resources.params]
            byline = 'picasso'
        
        [[resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [resources.params]
            byline = 'bep'
        
        got
        [root]
          title = 'Test Metadata'
        [[root.resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [root.resources.params]
            byline = 'picasso'
        
        [[root.resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [root.resources.params]
            byline = 'bep'
        
        
        
        
        diff:
        [[[resources]] [root] [resources.params]]
    remarshal_test.go:128: [XML => Toml] Expected 
        title = 'Test Metadata'
        		
        [[resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [resources.params]
            byline = 'picasso'
        
        [[resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [resources.params]
            byline = 'bep'
        
        got
        [root]
          title = 'Test Metadata'
        [[root.resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [root.resources.params]
            byline = 'picasso'
        
        [[root.resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [root.resources.params]
            byline = 'bep'
        
        
        
        
        diff:
        [[resources.params] [root] [[resources]]]
    remarshal_test.go:128: [XML =>  TOML ] Expected 
        title = 'Test Metadata'
        		
        [[resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [resources.params]
            byline = 'picasso'
        
        [[resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [resources.params]
            byline = 'bep'
        
        got
        [root]
          title = 'Test Metadata'
        [[root.resources]]
          src = '**image-4.png'
          title = 'The Fourth Image!'
          [root.resources.params]
            byline = 'picasso'
        
        [[root.resources]]
          name = 'my-cool-image-:counter'
          src = '**.png'
          title = 'TOML: The Image #:counter'
          [root.resources.params]
            byline = 'bep'
        
        
        
        
        diff:
        [[[resources]] [resources.params] [root]]
```

I can 'fix' this by automatically removing any root element in the `metadecorders` which might give some unexpected results or expect these exceptions in the test.

What do you think @bep?

bep: Just drop the remarshal thing for now. One question:

* What parts of Hugo do XML work with with this PR?
* I assume `reosurces.Get` and /data; for /data; what is the root key for `mydata.xml` with root `<root>`? `mydata.root.*`?

vanbroup: The root key is defined by the xml, so for RSS the key is `rss` as the document starts with `<rss>` and ends with `</rss>`. 

In this example the root key would be `note` and you would use `note.body`:
```
<?xml version="1.0" encoding="UTF-8"?>
<note>
    <to>Tove</to>
    <from>Jani</from>
    <heading>Reminder</heading>
    <body>Don't forget me this weekend!</body>
</note>
```

The current code works for `resources.Get` and the `/data` directory. It doesn't work for configs as Hugo is not expecting a root element. 

Dropping the root element would work as follows:
```
	case XML:
		var xmlRoot xml.Map
		xmlRoot, err = xml.NewMapXml(data)

		var xmlValue map[string]interface{}
		if err == nil {
			xmlRootName, err := xmlRoot.Root()
			if err != nil {
				return toFileError(f, errors.Wrap(err, "failed to unmarshal XML"))
			}
			xmlValue = xmlRoot[xmlRootName].(map[string]interface{})
		}
```

After which you would skip the root element so just `body` in the example above or this for an RSS feed (`{{ range .channel.item }}` instead of `{{ range .rss.channel.item }}`:

```
{{ with resources.Get "https://example.com/rss.xml" | transform.Unmarshal }}
    {{ range .channel.item }}
        <strong>{{ .title | plainify | htmlUnescape }}</strong><br />
        <p>{{ .description | plainify | htmlUnescape }}</p>
        {{ $link := .link | plainify | htmlUnescape }}
        <a href="{{ $link }}">{{ $link }}</a><br />
        <hr>
    {{ end }}
{{ end }}
```

bep: I'll bring @jmooring into this discussion. I think we need to get this right, as these kind of things are extremely hard to take back/change.

```
data
└── notes
    ├── note1.xml
    └── note2.xml
```

So, given the above, we currently get `site.Data.notes.note1.note.title` (or something). Which seems a little stuttery to me ... Also, it would be nice to get the different types (JSON, XML, YAML, TOML) as similar as possible in the structure.

vanbroup: It would indeed be `site.Data.notes.note1.note.title`

I also have a version ready that would strip the root element which would make it `site.Data.notes.note1.title`. But the root element would still be in the source file which could be confusing.

For frontmatter and configs we could wrap the information in a `<hugo>` root element and only trip the root element if it's called `hugo`, which makes the `TestRemarshal` happy.

bep: >For frontmatter and configs

I need to sleep on this, but for frontmatter and config (if we want to got down that route) I would imagine that we would repeat the _topic_ so it would be `<frontmatter></frontmatter>` and `<config></config>` (the latter in a file named `config.xml`).

jmooring: XML for frontmatter and config? Please, no, no, no. 

I don't like the idea of swallowing the root node, because it might provide context. This is contrived, but worth thinking about.

```text
data
└── messages/
    ├── message-1.xml
    ├── message-2.xml
    └── message-3.xml
````

data/messages/message-1.xml

```xml
<info>
  <body>The engine is warm.</body>
</info>
```

data/messages/message-1.xml

```xml
<info>
  <body>The engine is warm.</body>
</info>
```

jmooring: Yeah, well, those aren't messages. If they were messages, they'd be wrapped in a message (root) node. I think it's reasonable to tell a site author, "Hey, if you drop an XML file in the data/books/ directory, it's a book."

So swallow the root node. Sorry for the noise.

vanbroup: The reason to swallow the root node would be to support XML for frontmatter and config as these have a defined format and do not expect the root node.

If we don't care about XML for frontmatter and config, which I can't imagine anyone would want to use, we should keep the root node.

The good thing is, if we would want to support XML for frontmatter or config later, we can just swallow these specific root nodes (or in that context somehow).

I would suggest to leave it as it's now, just for `resources.Get` and `/data` without frontmatter and config support, and without modifying the XML (inclusive the root node).

jmooring: > The reason to swallow the root node would be to support XML for frontmatter and config

And to make XML consumption like JSON/YAML/TOML consumption... bep's point about stuttery syntax is dead on.

A site author says, "So let me get this straight. If I want to access the title, I can do `site.Data.books.mybook.title` if it's a JSON/YAML/TOML file, but I have to do `site.Data.books.mybook.book.title` if it's an XML file. Why did they decide to do it that way?" I don't want to have answer that question.

vanbroup: Is that not exactly the question to expect when you modify the XML?

XML comes with the root element to my knowledge you always need to specify the root element when querying the contents.

It's simple to remove, but I would be concerned that this would raise questions instead of removing them.

jmooring: > I would be concerned that this would raise questions

Can you provide an example?

vanbroup: Simply the reversed, I have XML document file.xml and I want to access the title but this doesn't work:

```
{{ site.Data.directory.file.note.title }}
```

bep: OK, take this example:

```
data
└── notes
    ├── note1.xml
    ├── note2.xml
    └── note3.json
```

The above matches pretty much what we have in the front matter world in Hugo (some prefer YAML, others TOML).

If we preserve the XML root, we would need to have something like this:

```
{{ $title := "" }}
{{ $note :=  site.Data.notes.note1 }}
{{ if dataFileIsXML }}
{{ $title = $note.note.title }}
{{ else }}
{{ $title = $note.title }}
{{ end }}
```

jmooring: With a realistic directory structure, that would be more like `{{ site.Data.notes.note_1.note.title }}`. Personally, I'd prefer that XML, JSON, TOML, and YAML all behave the same way. It's a feature:

"As a convenience, Hugo allows you to access XML data in the same way that you access JSON, TOML, and YAML: you do not need to specify the root element."

vanbroup: I just checked some XML implementations and they don't seem to be consistent with the root element (e.g., Python vs PHP)

So maybe it's not a problem to remove the root node.

I will push the update to remove the root node and add this to the documentation.

Do we agree that we do not implement and test XML for config and frontmatter?

bep: >Do we agree that we do not implement and test XML for config and frontmatter?

Yes.

And also yes to removing the root node for the others. With that I guess the Remarshal tests should pass?

vanbroup: No, only if I also add frontmatter support

vanbroup: By removing the root node, minimizing xml files will fail as the output is not equal to the input (missing the root node).

vanbroup:
Root nodes are removed in the data structure, but not when minifying data.

`TestRemarshal` is implemented and works as expected. To get this working I had to add XML support in `parser.InterfaceToConfig`, located in `frontmatter.go`, but without adding an XML delimter this is not adding support for XML frontmatter or config files.

beebe4:
Paul,

Thanks for writing this and getting it pushed through.  I am trying to consume a Medium RSS feed and using this feature.  

While trying to get the primary content from Medium, it passes an XML node called `content:encoded`.  When I attempt to reference that node with `{{ .content:encoded | plainify | htmlUnescape | truncate 200 }}` I am getting a `expected :=` error.

Any ideas on how to get around this?

vanbroup:
Have you tried dumping the results for debugging `{{.}}` to figure out the path?

Do you have a full example?

PR #9082, opened 2021-10-25
htime: set zone of datetime from from `go-toml`

### LINKED ISSUE #8895: time.Format - timezone details missing with time_full and time_long with TOML dates
This issue title is... terrible.

content/tests/pass.md

```toml
+++
title = "Pass"
date = "2021-08-16T06:00:00-07:00"  # String
draft = false
+++
```

content/tests/fail.md

```toml
+++
title = "Fail"
date = 2021-08-16T06:00:00-07:00    # Date
draft = false
+++
```

||pass.md|fail.md
:--|:--|:--
`{{ .Date \| time.Format ":time_full" }}`|6:00:00 am Pacific Daylight Time|6:00:00 am
`{{ .Date \| time.Format ":time_long" }}`|6:00:00 am PDT|6:00:00 am

satotake:
`go-toml/v2`'s unmarshaler does not specify zone name even if value has
offset explicitly.
To make time-formatting behaviour consistent, convert them into string
in hugo.

Close #8895

### DIFF
--- a/common/htime/time.go
+++ b/common/htime/time.go
@@ -136,6 +136,11 @@ func ToTimeInDefaultLocationE(i interface{}, location *time.Location) (tim time.
 		return vv.AsTime(location), nil
 	case toml.LocalDateTime:
 		return vv.AsTime(location), nil
+	// issue #8895
+	// datetimes parsed by `go-toml` have empty zone name
+	// convert back them into string and use `cast`
+	case time.Time:
+		i = vv.Format(time.RFC3339)
 	}
 	return cast.ToTimeInDefaultLocationE(i, location)
 }


satotake:
> look at my edit of the committ message.
@bep may be too late but where?

bep:
@satotake that comment is old, but the reason this issue surfaces is that I'm looking at this code re another issue, and I see some code

```go
// issue #8895
// datetimes parsed by `go-toml` have empty zone name
// convert back them into string and use `cast`
case time.Time:
    i = vv.Format(time.RFC3339)
}
````

Which I thought I could try to remove and see if a test failed, it didn't. But then I notice that there wasn't any test added for this in the first place.

Do you have a test case?

satotake:
> Do you have a test case?
I will add it soon.
@jmooring gave us a good example in #8895. I will refer it.

bep:
@jmooring 's example isn't very good as it fails with the current code. I assume it works at certain times of the year (where the -7 offset is un-ambigous).

I suspect that the upstream code is fixed now and that the code block in question could be removed, but it would be good to have a test that failed before, so to speak. Looking at the comments I do see some mixup between time zone names and offsets. TOML uses "RFC 3339 formatted date-time with offset" and you will only get a timezone name in there if that offset is unique.

bep:
OK, I suspect that go-toml is using `time.Local` here ... I will check what my computer is set to.

satotake:
@bep I encountered the similar problem.

> I suspect that the upstream code is fixed

But I doubt it  and his comment has a point.

It seems that offset-to-zone lookup table (+9:00 -> `JST`) can differ from machines.
What we fixed here is not offset-to-name conversion but parsing date object in toml format.

As you mentioned, zone name is not deterministic but it is another (upstream) issue.
When you use `:time_long` format, hugo use [github.com/go-playground/locales](https://github.com/go-playground/locales). 
And then it calls `time.Time.Zone()` of golang builtin.

PR #9111, opened 2021-11-01
Fix description of lang.FormatNumberCustom

jmooring:
It currently refers to itself as a simple alternative, when it should
refer to lang.FormatNumber.


bep:
@jmooring thanks, on a side note, I'm planning on doing a Hugo release tomorrow.

PR #9114, opened 2021-11-02
Fix default value of timeout in site configuration

jmooring:
- The existing default value is a string: 30s
- cfg.Language.GetInt() in deps/deps.go cannot cast this to a string,
  discards the error, and returns 0
- Then, also in deps/deps.go, timeoutms <= 0 is true so the default
  timeout value is set to 3000ms (3s)

### DIFF
--- a/docs/content/en/getting-started/configuration.md
+++ b/docs/content/en/getting-started/configuration.md
@@ -381,9 +381,9 @@ The directory where Hugo reads the themes from.
 
 ### timeout 
 
-**Default value:** 10000
+**Default value:** 30000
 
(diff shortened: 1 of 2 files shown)


jmooring:
Reference <https://discourse.gohugo.io/t/default-timeout-value/35403>

jmooring:
Other than the documentation update, I don't think change makes any functional difference. I don't think the timeout value stored in Deps is actually used anywhere due to:

<https://github.com/gohugoio/hugo/blob/master/hugolib/site.go#L527-L542>

jmooring:
Closing, red herring, sorry for the noise. Although the value is incorrect, Deps.Timeout isn't used anywhere. We could:

1) Remove the code
2) Fix the code
3) Leave it as is and hope that no one uses the value in the future

<https://github.com/gohugoio/hugo/blob/master/deps/deps.go#L89>
<https://github.com/gohugoio/hugo/blob/master/deps/deps.go#L258-L261>
<https://github.com/gohugoio/hugo/blob/master/deps/deps.go#L288>

PR #9138, opened 2021-11-06
Add `clock` cli flag

### LINKED ISSUE #8787: Add --date CLI flag
Some builds are run daily with expectations regarding dates. 

It would be interesting to be able to test drive the build on a with date so that Hugo time functions (`now` etc...) uses the time passed through the flag instead of the machine time.

This would allow users to make sure the expected behaviour on or after a certain date/time works as expected.

Parameter could be the same as what the [`time`](https://gohugo.io/functions/time/#readout) function is currently expecting.

```
hugo --date 2023-07-04
```

```
hugo --date 2016-05-28T10:30:00.00+10:00
```

I suspect this might be tricky, but if not, I'd regret not suggesting it.

satotake:
issue #8787
If I understand @regisphilibert 's comment properly, with the flag, users want to check

- future/expired content's manners
- `now` in template
- `hugo list` output

If so, we can implement this flag by just replacing a few of `time.Now()` without changing internal time of go.

This PRs add `buildTime` flag in order to fake hugo's build datetime like so

```
hugo --buildTime 2021-11-06
hugo --buildTime 2021-11-06 server
hugo --buildTime 2021-11-06 list future
hugo --buildTime 2021-11-06 list expired
```

Close #8787


flimzy:
This is a feature that would make my life a lot easier. What are the chances of getting this reviewed and rebased any time soon?

davidsneighbour:
> This is a feature that would make my life a lot easier. What are the chances of getting this reviewed and rebased any time soon?

Thumbs up and useful contributions tend to do that ;)

The "build time" moniker is slightly misleading but I like the concept. If it has enough testing to make sure it works as expected and when the "conflicts that must be resolved" are resolved (probably just a fetch of the current HEAD and re-check) our little conversation here will put it in the TODO pile again and someone who knows Golang will surely have a look.

flimzy:
@satotake Are you still interested in this issue/PR?  If not, I'd be happy to take it over and do a rebase in a new PR.

satotake:
Honestly, this PR has been a dim memory for me but rebasing and resolving were not so tough.
Thanks for your offer, @flimzy .

If you do not mind, could you review this PR or have any idea, @bep?
Especially, my concern is flag name as @davidsneighbour said.

commands/list_test.go
```diff
@@ -69,3 +69,61 @@ func TestListAll(t *testing.T) {
...
+
+`)
+	defer clean()
+
+	c.Assert(err, qt.IsNil)
+
+	hugoCmd := newCommandsBuilder().addAll().build()
+	cmd := hugoCmd.getCommand()
+
+	defer func() {
```
flimzy, line 94: This would probably be better done with `t.Cleanup`

tpl/time/time.go
```diff
@@ -27,17 +27,19 @@ import (
 )
 
 // New returns a new instance of the time-namespaced template functions.
-func New(translator locales.Translator, location *time.Location) *Namespace {
+func New(translator locales.Translator, location *time.Location, buildTime string) *Namespace {
```
flimzy, line 30: Would it make more sense to pass a `time.Time` value here, rather than a string to be re-parsed, so that the build times are consistent across a single run?

As a potential bonus, by doing so the `location` parameter may become redundant.

satotake: > Would it make more sense to pass a time.Time value here, rather than a string to be re-parsed, so that the build times are consistent across a single run?

Yes, you are right. Fixed.

> As a potential bonus, by doing so the location parameter may become redundant.

I do not think so.
This `location` is determined by site lang settings.
It is possible that you would like to use this flag with a different location from the lang settings'

bep: I have looked at this, and agree that we do this.

In short, we need something that allows us to say:

```
`expiryDate.Before(htime.Now())` 
```

Or something.

We should discuss this before doing this, but since I'm really tempting to take the simplest approach here, which is:

1. Add a `htime` package (if not already there)
2. In that package, add a public `var NowOverride time.Time´ and a `Now` func ala:

```go
func Now() time.Now {
   if !NowOverride.IsZero {
      return NowOverride
   }
   return time.Now()
}
```

And use that func where applicable. 

This means that tests using this cannot use `t.Parallel`, but it is extremely easy to reason about.

satotake: > Add a htime package (if not already there)
> In that package, add a public var NowOverride time.Time and a Now func ala:
> This means that tests using this cannot use t.Parallel, but it is extremely easy to reason about.

Though I am concerning about test execution time and so on a little bit, probably it will be fine.

satotake: > Though I am concerning about test execution time and so on a little bit, probably it will be fine.

I assumed there are some test cases to be fixed but I have found it is only `TestShouldBuild` in `page_test.go`.
pretty small.

bep: The test execution time should be negligible, and this i obviously the best choice from a pragmatic standpoint: Very simple to implement and reason about. Let's do it.

satotake: Done!
Could you review this?

bep: Note that all the existing `t.Parallel` tests will still run in parallel, it's just tests that make changes to the global that needs to take extra care.

satotake: You mean any other `t.Parallel` (which uses `NowOverride` internally) should be removed?

bep: No! That would kill the test build time. Tests that _changes the global time_ can not use `t.Parallel`, all existing tests are fine.

satotake: ah...oops, I got what you meant finally.
Fixed.

commands/list_test.go
```diff
@@ -66,3 +66,57 @@ func TestListAll(t *testing.T) {
 		"false", "https://example.org/p1/",
 	})
 }
+
+func TestBuildTimeFlag(t *testing.T) {
```
bep, line 70: I would prefer if you moved this to a "site test" in `hugo_test.go` and use the newly added `newTestHugoCmdBuilder` builder.

hugolib/hugo_sites_build_test.go
```diff
@@ -1263,6 +1263,7 @@ func newMultiSiteTestBuilder(t testing.TB, configFormat, config string, configDa
 		configData = map[string]any{
 			"DefaultContentLanguage":         "fr",
 			"DefaultContentLanguageInSubdir": true,
+			"BuildTime":                      "2021-11-06",
```
bep, line 1266: I'm almost certain this is not what we want.

hugolib/page_test.go
```diff
@@ -1536,7 +1537,8 @@ Content.
 }
 
 func TestShouldBuild(t *testing.T) {
-	t.Parallel()
+	htime.NowOverride = time.Date(2021, 11, 17, 20, 34, 58, 651387237, time.UTC)
```
bep, line 1540: Please don't change existing tests; all of Hugo when this flag is not set should work with the system clock, so it's vital that we keep existing tests as is to verify this. If you have new test cases, create a new test func.

resources/resource/dates.go
```diff
@@ -44,19 +44,19 @@ func (d *Dates) UpdateDateAndLastmodIfAfter(in Dated) {
 }
 
 // IsFuture returns whether the argument represents the future.
-func IsFuture(d Dated) bool {
+func IsFuture(buildTime time.Time, d Dated) bool {
```
bep, line 47: One goal of having a global override was to minimise the changes needed. This means that we only want to change the import from `time.Now()` to `htime.Now()` in the relevant places.

bep (changes requested):

bep:
OK, I have thought some more about this, and I think we need to do an adjustment.

The thing is, when we're changing the build time, I also suspect people would also expect the clock to change in other situations, e.g.:

```
{{ ge .PublishDate now }}
```
Should always print true.

With that in mind, I think we need a ... clock.

What I suggest is that we

* rename the `buildTime` flag to `clock` (?)
* rework the implementation in this PR to also consider the system clock. I have toyed around with this package tonight, which I thinks should be simple enough to reason about: https://github.com/bep/clock
* Replace all usage of `time.Now` with `htime.Now`

Thoughts?

satotake:
> rename the buildTime flag to clock (?)

Looks good. I thought just `time` or `date` might be ambiguous for this.

> rework the implementation in this PR to also consider the system clock. I have toyed around with this package tonight, which I thinks should be simple enough to reason about: https://github.com/bep/clock

I have presumed fixed time for this. yes but it is ok to make it running. I will check it your package.

> Replace all usage of time.Now with htime.Now

Not sure but I suspect that they are not all but most. For example, values for logging and so on? Anyway, I will check it, too.

bep:
Thanks, I have rebased this into a new PR and will finish this in #9864

satotake:
ok

I replaced time.Now with htime.Now as much as possible (including logging) for now.

except for

- values in tests
- [seed value for random](https://github.com/gohugoio/hugo/blob/f2946da9e806c2bafbdd26707fe339db79bd980b/tpl/collections/collections.go#L38)
- [cloudfront deploy](https://github.com/gohugoio/hugo/blob/f2946da9e806c2bafbdd26707fe339db79bd980b/deploy/cloudfront.go#L45)

PR #9154, opened 2021-11-10
accept a class parameter (slice or string) and append to <ul>

sean-pwd:
This allows css manipulation of pagination block via utility classes. 

By default the pagination block is justify:left and it can only be justified to the center via css (not utility classes, which this template is written with)

### DIFF
--- a/tpl/tplimpl/embedded/templates/pagination.html
+++ b/tpl/tplimpl/embedded/templates/pagination.html
@@ -5,6 +5,7 @@
 
 {{- $page := . }}
 {{- $format := "default" }}
+{{- $class := slice -}}
 
 {{- if reflect.IsMap . }}
   {{- with .page }}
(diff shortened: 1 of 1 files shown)


jmooring:
@sean-pwd Thanks for this. I'll be working on this embedded template in the coming days to address an unrelated issue, and I will give this some thought. One of my concerns is, if we do this, are we going to have similar requests for the other embedded templates and built-in shortcodes? That's not a necessarily a bad thing, but it would set a precedent.

sean-pwd:
I'm not sure about the precedent, but as a template that utilizes bootstrap
utility classes it's hard to work with - I wrapped it in a div and was
unable to center the buttons with utility classes on the div.

davidsneighbour:
@sean-pwd you center your pagination using BS5 with flex like this, works with the internal pagination template:

```
<div class="d-flex justify-content-center"></div>
```

sean-pwd:
Thanks for that, that was a big oversight on my part.

I'll leave it up to you guys if its appropriate to include the class
parameter


Sean Emerson
Web Developer, Powerful Web Design
02 8004 3293 <212-931-0000>
| ***@***.***
www.powerfulwebdesign.com.au
<http://www.facebook.com/> <http://www.linkedin.com/> <http://twitter.com/>
<http://instagram.com/>

Issue #9148, opened 2021-11-09
Add debug.List


bep:
See https://discourse.gohugo.io/t/advice-on-how-to-inspect-a-page-context/35504/4

I suggest we add a `debug.List` method that uses reflection to list all fields and method names on structs/pointers, or all keys if it's a map -- non-recursive; returning a string slice would be good, I think (it prints nicely).

mohsalsaleem:
Hello @bep, lets assume the following is the struct 
```
type Address {
    street string
    city     string
}
type User struct {
    name string
    address Address  
}
```
Then, is `name, address` the expected output? Or, should it get the output recursively, in this case the fields of address as well?

bep:
@mohsalsaleem to be more precise:

```go
type Address {
    Street string
    City     string
}
type User struct {
    Name string
    Address Address
    foo     string
}

func (u *User) M1() string
func (u *User) M3() string
func (u *User) M2(v string) string
func (u *User) m4(v string) string
```

For `User`, the expected output (I think) is:

```
["Address", "Name", "M1", "M2", "M3"]
```

As in, only the exported fields and methods.

The above would some less than useful entries for the end user, but so be it.

For `maps` list just the keys sorted (again, non-recursive).

mohsalsaleem:
Thanks for the explanation @bep.
I've implemented the changes in the commit https://github.com/mohsalsaleem/hugo/commit/2994a16d7c116f89f8a79f4dd03e11089aa378a1
Can you please help me with writing tests and testing this functionality? 
I see that, in tpl/debug/init.go, an example needs to be added, but I'm a bit lost as to how I can structure the example.

johannesengl:
@bep @mohsalsaleem I see there hasn't been any activity on this issue since November 2021. Is this issue still valid? Is there help wanted on this? I am looking for a good first issue to pickup.

johannesengl:
@bep Added a PR for this issue: https://github.com/gohugoio/hugo/pull/10806
Would appreciate a review :)

shubmjagtap:
Hello @bep can i work on this issue ?

cubxxw:
Is this issue end-of-life?

EPoY74:
Hello!

Is this issues still relevant?

Lalithsaireddy:
The issue has been resolved in sort.go by correctly handling the sorting of mapValues.





 // Copyright 2018 The Go Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
[...]

PR #9163, opened 2021-11-12
Implement debug.List

mohsalsaleem:
Refers #9148
This PR implements `debug.List` method that uses reflect package to to list exported fields and methods names on structs/pointers and all keys in a map.


bep:
Thanks, can you push the "sign CLA button" above?

mohsalsaleem:
I’ve completed the CLA Signing, but for some reason the bot thinks I have not.

bep:
>I’ve completed the CLA Signing, but for some reason the bot thinks I have not.

You have probably a email mismatch between the mail address you used for the commit (as in: your Git user) and the email address registered on GitHub. You can fix this by adding email aliases in your GitHub profile.

mohsalsaleem:
Thanks for pointing it out, looks like CLA check passes now!

mohsalsaleem:
In case I need to add tests, please let me know. I looked around and could not find anything specific to add tests to apart from the init.go file, which I'm unclear on how to structure the example.

bep:
>In case I need to add tests, please let me know.

We need some basic tests. Just create a `debug_test.go` file and put a basic `TestList` test with a simple test structs and a map. The init test you can ignore.

mohsalsaleem:
@bep I just added some basic tests, please have a look! Thanks.

mohsalsaleem:
I just noticed that the functions also returns fields that are not exported, let me fix that.

mohsalsaleem:
Fixed the issue, please review. 
Thanks.

bep:
Thanks for this. I need to think a little before pulling this into the main branch. I have tested this a little, and while it does work great on technical level, I fear it may expose too much (deprecated and internal methods (not meant to be used in the templates, e.g. `MarshalJSON`) which I think may be more confusing than useful. I need to think about a way to filter out those.

dittrichlucas (approved):

Issue #7699, opened 2020-09-16
Add template function for making general HTTP requests


atishay:
A lot of Jamstack services like Formspree require the `Authorization` header for authentication. Without being able to supply headers, we cannot access the API provided by these services from within Hugo. getJSON and getCSV are very limited in scope and do not support any means for authentication or passing data. We cannot handle non 200 series status codes in Hugo templates as well.

NOTE: With `transform.Unmarshal` hugo already supports parsing textual responses.

moorereason:
I actually started working on such a function a few weeks ago, but it's not finished.  Below is what I was wanting to implement.  Comments welcome.

#### Function signature

```
getHTTP URL [OPTIONS...]
```

#### Optional Parameters Dictionary

* "method" = string (GET, POST, etc)
* "headers" = dict (map[string]string)
* "body" = string
* "no-cache" = bool (bypass Hugo's caching mechanism; not sure this is useful) ❓ 

#### Returns

A HTTP Response object (a simplified view of [`http.Response`](https://golang.org/pkg/net/http/#Response)) containing:

* StatusCode (int)
* Headers (map[string]string)
* Body (string or keep it as `[]byte`?) ❓

#### Example Usage

```
{{ $resp := getHTTP $url (dict "method" "POST" "body" $body "headers" (dict "Authorization" "blah")) }}
{{ if eq resp.StatusCode 200 }}
  {{ $json := transform.Unmarshal resp.Body }}
{{ end }}
```

Cc: @bep

bep:
We have discussed this, and I'm convinced that the method signature should be something like:

```
{{ $resource := $url | resources.GetRemote $opts }}
```

But I would I would love it if we could find a **great** Go library with a struct config that we could fairly easily adapt).

moorereason:
Updated my "proposal" comment above to move the URL parameter to the end of the function signature.

atishay:
@moorereason  Hugo defines [`maxAge`](https://gohugo.io/getting-started/configuration/#configure-file-caches) for controlling its caching. Instead of `no-cache`, `maxAge`which can override the default provided in the config on a per request basis may make more sense.

bep:
To repeat myself, we're not adding some low-level HTTP API to Hugo, so this:

```html
{{ if eq resp.StatusCode 200 }}
  {{ $json := transform.Unmarshal resp.Body }}
{{ end }}
```

Is not something I want to teach the average Hugo user on the forum.

atishay:
[...]

PR #9185, opened 2021-11-18
Add remote support to resources.Get

### LINKED ISSUE #5255: Add remote support in resources.Get

vanbroup:
Implement resources.FromRemote, inspired by #5686

Closes #5255
Supports #9044


hugolib/resource_chain_test.go
```diff
@@ -424,12 +437,16 @@ CSS integrity Data last:  {{ $cssFingerprinted2.RelPermalink }} {{ $cssFingerpri
...
-			`
+			fmt.Sprintf(`
 SUNSET: images/sunset.jpg|/images/sunset.a9bf1d944e19c0f382e0d8f51de690f7d0bc8fa97390c4242a86c3e5c0737e71.jpg|900|90587
 FIT: images/sunset.jpg|/images/sunset_hu59e56ffff1bc1d8d122b1403d34e039f_90587_200x200_fit_q75_box.jpg|200
 CSS integrity Data first: sha256-od9YaHw8nMOL8mUy97Sy8sKwMV3N4hI3aVmZXATxH&#43;8= /styles.min.a1df58687c3c9cc38bf26532f7b4b2f2c2b0315dcde212376959995c04f11fef.css
 CSS integrity Data last:  /styles2.min.1cfc52986836405d37f9998a63fd6dd8608e8c410e5e3db1daaa30f78bc273ba.css sha256-HPxSmGg2QF03&#43;ZmKY/1t2GCOjEEOXj2x2qow94vCc7o=
-`)
+
+SUNSET REMOTE: %[1]s.jpg|/%[1]s.a9bf1d944e19c0f382e0d8f51de690f7d0bc8fa97390c4242a86c3e5c0737e71.jpg|900|90587
+FIT REMOTE: %[1]s.jpg|/%[1]s_hu59e56ffff1bc1d8d122b1403d34e039f_0_200x200_fit_q75_box.jpg|200
```
vanbroup, line 447: @bep The size in the permalink turns 0, any idea why this could happen? The size is not 0.

I tested fit and resizing and the following works:
```
    {{- $image := resources.FromRemote "https://d33wubrfki0l68.cloudfront.net/291588ca4ec28632c45707669f113eced92ea64d/f1a39/images/homepage-screenshot-hugo-themes.jpg" -}}
    {{ $fit := $image.Fit "200x200"}}

    {{$image.Permalink}}
    {{$fit.Permalink}}

    <img src="{{ $image.Permalink }}" alt="{{ $image.Permalink }}" />
    <img src="{{ $fit.Permalink }}" alt="{{ $fit.Permalink }}" />
```

hugolib/resource_chain_test.go
```diff
@@ -537,6 +554,65 @@ func TestResourceChains(t *testing.T) {
...
+				<html>
+					<a href=#>Cool</a>
+				</html>`))
+			return
+		}
+
+		http.Error(w, "Not found", http.StatusNotFound)
+		return
+	}))
+	//defer ts.Close()
```
vanbroup, line 614: The test server should be closed, but I think the sitebuilder runs in a routine and the server will be stopped before finished.

Any suggestions? Can we start a test server as a shared resource?

hugolib/resource_chain_test.go
vanbroup: Fixed usign c8b07b4

hugolib/resource_chain_test.go
vanbroup: Fixed using c8b07b4

vanbroup:
One example use case is the RSS/XML feed parsing as discussed in #9044 another use case could be the generation of Open Graph images (in this case using https://bruzu.com):

```go
   {{- $image := resources.FromRemote ( print "https://img.bruzu.com/?" (querify 
        "bi" "https://example.com/img/og-background.png" "h" "627" "w" "1200"
        "a.oy" "top" "a.ox" "left" "a.x" "100" "a.y" "100" "a.w" "1000" "a.h" "68" "a.c" "#ffffff" "a.ta" "left" "a.fs" "40" "a.lh" "1" "a.fw" "900" "a.ff" "Roboto" "a.maxHeight" "120"
        "b.tp" "textbox" "b.x" "100" "b.y" "200" "b.w" "1000" "b.h" "244" "b.oy" "top" "b.ox" "left" "b.c" "#ffffff" "b.ta" "left" "b.fs" "20" "b.lh" "1.2" "b.fw" "400" "b.ff" "Roboto" "b.ox" "left" "b.oy" "top" "b.maxHeight" "250"
        "a.t" .Title
        "b.t" .Summary
    )) -}}
    <meta property="og:image" content="{{ $image.Permalink }}" />
```

hugolib/resource_chain_test.go
vanbroup: @bep `hugio.NewReadSeekerNoOpCloser` is not storing any files, causing the unavailability of `hugofs.FileMetaInfo` in https://github.com/gohugoio/hugo/blob/master/resources/resource.go#L644, resulting in the 0 source-file size indicator in the publication filename.

I'm not sure why the original file size is included in the filename and if we really care about its or the `hugofs.FileMetaInfo` absence. The `resources.FromString` function has the same limitation (but does not support the image type).

bep: The size is 0 because it's not a file (we read that from the FileInfo object we get when doing `os.Stat`). It's probably fixable (we could probably read the size when we generate the hash), but not important.

vanbroup: Thanks for your feedback, I will mark this as resolved.

resources/resource_factories/create/create.go
```diff
@@ -126,3 +137,137 @@ func (c *Client) FromString(targetPath, content string) (resource.Resource, erro
...
+func (c *Client) FromRemote(args ...interface{}) (resource.Resource, error) {
+	uri, headers := toURLAndHeaders(args)
+	rURL, err := url.Parse(uri)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to parse URL for resource %s", uri)
+	}
+
+	resourceID := helpers.MD5String(uri)
+
+	return c.rs.ResourceCache.GetOrCreate(path.Join(resources.CACHE_OTHER, resourceID), func() (resource.Resource, error) {
```
bep, line 152: There is a long story, but I'm currently on a branch that will improve this "dependency tracking vs cache" situation (also, but not now (we should mention it in the documentation), this only caches to memory, which may not be optimal for lots of remote resources).

But, using `resources.CACHE_OTHER` will effectively trigger a re-download of all remote resources on every rebuild, which is probably not what you would want. I suggest you use `resources.ResourceCacheKey`).

bep (changes requested):

resources/resource_factories/create/create.go
```diff
@@ -126,3 +137,137 @@ func (c *Client) FromString(targetPath, content string) (resource.Resource, erro
...
 	})
 }
+
+// FromRemote expects one or n-parts of a URL to a resource
+// If you provide multiple parts they will be joined together to the final URL.
+func (c *Client) FromRemote(args ...interface{}) (resource.Resource, error) {
+	uri, headers := toURLAndHeaders(args)
+	rURL, err := url.Parse(uri)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to parse URL for resource %s", uri)
```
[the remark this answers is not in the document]
vanbroup: Error message have been updated to start with a lowercase in ba23cbb

resources/resource_spec.go
```diff
@@ -112,6 +114,8 @@ type Spec struct {
 	Logger      loggers.Logger
 	ErrorSender herrors.ErrorSender
 
+	HTTPClient *http.Client
```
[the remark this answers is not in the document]
vanbroup: I moved the http client to the create factory client and made it private with a default timeout of 10 seconds.

tpl/resources/resources.go
```diff
@@ -199,6 +199,12 @@ func (ns *Namespace) FromString(targetPathIn, contentIn interface{}) (resource.R
 	return ns.createClient.FromString(targetPath, content)
 }
 
+// FromRemote creates a Resource from a URL.
+// If you provide multiple parts they will be joined together to the final URL.
+func (ns *Namespace) FromRemote(args ...interface{}) (resource.Resource, error) {
```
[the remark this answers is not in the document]
vanbroup: ad6e83f updates `resources.Get` to check for remote resources, this is also covered by testing

resources/resource_factories/create/create.go
```diff
@@ -126,3 +137,137 @@ func (c *Client) FromString(targetPath, content string) (resource.Resource, erro
...
+
+		return c.rs.New(
+			resources.ResourceSourceDescriptor{
+				Fs:          c.rs.FileCaches.AssetsCache().Fs,
+				LazyPublish: true,
+				OpenReadSeekCloser: func() (hugio.ReadSeekCloser, error) {
+					return hugio.NewReadSeekerNoOpCloser(bytes.NewReader(body)), nil
+				},
+				RelTargetFilename: filepath.Clean(resourceID),
+				SourceFilename:    filename,
```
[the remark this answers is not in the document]
vanbroup: Remove `SourceFilename` in 871bbd4

resources/resource_factories/create/create.go
```diff
@@ -126,3 +137,137 @@ func (c *Client) FromString(targetPath, content string) (resource.Resource, erro
...
+		}
+
+		var contentType string
+		if arr, _ := mime.ExtensionsByType(res.Header.Get("Content-Type")); len(arr) == 1 {
+			contentType = arr[0]
+		}
+
+		// If content type was not determined by header, look for a file extention
+		if contentType == "" {
+			if ext := filepath.Ext(filename); ext != "" {
```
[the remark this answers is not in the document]
vanbroup: Changed `filepath.Ext` to `path.Ext` in 871bbd4

resources/resource_factories/create/create.go
vanbroup: 871bbd4 changes this to `resources.ResourceCacheKey`, does this mean a cache of the remote resource is stored to disk and retained between builds?

As it only returns a string, I doubt it, but I haven't researched the cache system.

tpl/resources/resources.go
```diff
@@ -199,6 +204,12 @@ func (ns *Namespace) FromString(targetPathIn, contentIn interface{}) (resource.R
...
```
bep, line 209: Remove `ns.FromRemove` and call the client directly (we do not want two template methods doing just about the same thing).

vanbroup: Removed the `resources.FromRemote` call, extended test coverage and updated documentation.

docs/content/en/hugo-pipes/introduction.md
```diff
@@ -20,12 +20,46 @@ aliases: [/assets/]
...
 
 ```go-html-template
 {{ $style := resources.Get "sass/main.scss" }}
+{{ $remoteStyle := resources.Get "https://www.example.com/styles.scss" }}
+```
+
+When using an URL the functions accepts [variadic arguments][variadic]:
+
+```
+{{ $resource := resources.Get "url" "arg1" "arg2" "arg n" }}
```
bep, line 35: I will look at the complete PR (in general, it looks very good), so you don't need to do something about this right away -- but I write it now while remembering it:

The original `getJSON` etc. API with variadic arguments was a mistake. If people want to concatenate and URL from different parts, they can do so before passing it to `resources.Get`.

So, I suggest that we stick with the 1 argument, with an optional second argument for HTTP headers. Thinking about this, maybe we should make this a `options` so we could add stuff to it in the future (e.g. HTTP config, timeout, whatever), so we do `options.headers` for now?

vanbroup: Agree, this would be a better approach, I wanted to stick with what is used elsewhere but as this will deprecate the other functions it's a good moment to break that logic.

I have seen some issues asking to support other HTTP methods (e.g., POST), we might be able to include support for that (including the corresponding payload) as well.

resources/resource_factories/create/create.go
bep: No, but the `resources.ResourceCacheKey` prevents it from being re-downloaded on every rebuild in server mode. There is a long story here; I'm doing some major cleanups in this area in another branch, and we will improve this situation before we release Hugo 0.90, so just let it be for now.

docs/content/en/hugo-pipes/introduction.md
bep: OK, I have looked at this one more time -- and if you do the one change as we discussed above, I will merge it (we/I will improve the cache situation in an upcoming commit).

So, we make it into 1 URL argument with an optional second `options` argument. 

```
{{ $resource := resources.Get "https://example.org/api" (dict "headers" (.dict "Authorization" "Bearer abcd" ))  }}
```

resources/resource_factories/create/create.go
```diff
@@ -126,3 +144,136 @@ func (c *Client) FromString(targetPath, content string) (resource.Resource, erro
...
+// FromRemote expects one or n-parts of a URL to a resource
+// If you provide multiple parts they will be joined together to the final URL.
+func (c *Client) FromRemote(args ...interface{}) (resource.Resource, error) {
+	uri, headers := toURLAndHeaders(args)
+	rURL, err := url.Parse(uri)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse URL for resource %s", uri)
+	}
+
+	resourceID := helpers.MD5String(uri)
```
bep, line 157: We need to somehow include the header info in this cache key. It's probably simpler to use `helpers.HashString(args...)`.

vanbroup:
Addressed the comments and implemented options, you can now specify the request headers in the options, optionally combined with a request method and body (see also related issue #7699).
```
{{ $postResponse := resources.Get "https://example.org/api"  (dict 
    "method" "post"
    "body" `{"complete": true}` 
    "headers" (dict 
        "Content-Type" "application/json"
    )
)}}
````

File: resources/resource_factories/create/create.go [hunk 12]
```diff
@@ -126,3 +145,161 @@ func (c *Client) FromString(targetPath, content string) (resource.Resource, erro
...
+// If you provide multiple parts they will be joined together to the final URL.
+func (c *Client) FromRemote(uri string, options map[string]interface{}) (resource.Resource, error) {
+	rURL, err := url.Parse(uri)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse URL for resource %s", uri)
+	}
+
+	resourceID := helpers.HashString(uri, options)
+
+	return c.rs.ResourceCache.GetOrCreate(resources.ResourceCacheKey(resourceID), func() (resource.Resource, error) {
```
bep, line 159: I take that back, just use the `resourceID` as is for the key for now (I will fix this cache situation in a bit).

resources/resource_factories/create/create.go
```diff
@@ -126,3 +145,161 @@ func (c *Client) FromString(targetPath, content string) (resource.Resource, erro
...
+	if method, ok := options["method"].(string); ok {
+		method = strings.ToUpper(method)
+		switch method {
+		case "GET", "DELETE", "HEAD", "OPTIONS":
+			return method, nil, nil
+		case "POST", "PUT", "PATCH":
+			var body []byte
+			if _, ok := options["body"]; ok {
+				switch b := options["body"].(type) {
+				case string:
```
bep, line 292: We may get lots of different string types in here (not likely in this case, but...), so I suggest:

* Make `body` a string and use `cast.ToString` to do the type casting.
* Use strings.NewReader to make it into a reader.

bep (changes requested):

resources/resource_factories/create/create.go
vanbroup: If we make the body a string you can't use a binary resource, for example an image or sound resource, like this: 

```
    {{ $image := resources.Get "https://d33wubrfki0l68.cloudfront.net/291588ca4ec28632c45707669f113eced92ea64d/f1a39/images/homepage-screenshot-hugo-themes.jpg" }}

    {{ resources.Get "https://example.com/imagefilter" (dict 
        "method" "post"
        "body" $image.Content 
        "headers" (dict 
            "Content-Type" "image/jpeg"
        )
    )}}
```

resources/resource_factories/create/create.go
vanbroup: You mean going back to:
```
resourceID := helpers.MD5String(uri)
```

I do think it's good/a must to have optional headers, request method, etc. included in the cache key.
Any reason you think it's better to go back to just the URL?

resources/resource_factories/create/create.go
bep: Ah, ok, never mind.

bep:
I will merge this once the tests go green, it looks very good, I really like how the API turned out.
```
```

PR #9191, opened 2021-11-21
Remove empty href element from pagination template

### LINKED ISSUE #9149: Pagination: consider dropping href="#" for disabled buttons since it is an invalid fragment id
Great update in #8602 @jmooring!

Could we drop the invalid anchors (`href="#"`) for disabled buttons?

```html
<a href="#" aria-disabled="true" ...>
```

It's tripping link checking. For example, see https://github.com/gohugoio/hugo/blob/805c24c32c9b1f118c5f3693bfe1176d4efe4d9c/tpl/tplimpl/embedded/templates/pagination.html#L43

Prior to #8602, the previous button omitted the `href` attribute when there was no previous page.

/cc @austinlparker @nate-double-u

jmooring:
Closes #9149


jmooring:
Please merge.

PR #6370, opened 2019-09-26
Enable page exclusion from sitemap.xml

felicianotech:
This PR adds a new key to .Sitemap called `Exclude`. `Exclude` is a
boolean that is set to false by default meaning all available pages are
included. This is the behavior currently. When `Exclude` is set to true,
it will not appear in any `sitemap.xml` files that Hugo may generate.

`Exclude` can be set to true in the Hugo config turning `sitemap.xml`
into an opt-in rather than an opt-out.

Fixes #653

This is a rebased version of #5259 which was closed.

PR #9193, opened 2021-11-21
sitemap: Hide pages from sitemap if `.Sitemap.Hidden` is true

jwflory:
This commit changes the default `sitemap.xml` to exclude pages from the
sitemap if a page author has set `.Sitemap.Hidden` to `true` in the
front-matter of a page.

The motivation for this change is for cases where a site maintainer
wants to use content from a special category to embed onto another page.
In this case, the content in the special category is already present on
the site, but the pages in the special category are created anyways.
These pages are included in the sitemap and will be indexed by search
engines, but this may not be the desired behavior. So, a site maintainer
is now able to exclude pages from the sitemap without making custom
modifications to the `sitemap.xml` and losing out on any changes made to
the default sitemap in Hugo in the future.

This change was inspired from a support thread from May 2015:

https://discourse.gohugo.io/t/how-to-omit-page-from-sitemap/1126?u=jwf

Documentation is also updated so designers can also take note of this
new change when reading about Hugo's sitemap template.

### DIFF
--- a/docs/content/en/templates/sitemap-template.md
+++ b/docs/content/en/templates/sitemap-template.md
@@ -27,6 +27,9 @@ A sitemap is a `Page` and therefore has all the [page variables][pagevars] avail
 `.Sitemap.ChangeFreq`
 : The page change frequency
 
+`.Sitemap.Hidden`
+: Whether the page should be hidden from the generated sitemap
+
 `.Sitemap.Priority`
 : The priority of the page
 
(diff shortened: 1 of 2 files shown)


jmooring:
A couple of comments...

First, adding a field to the documentation does not add it to the data structure. Running your patch:

> WARN 2021/11/21 08:36:26 Unknown Sitemap field: hidden

Second, there's an open PR that already addresses this: https://github.com/gohugoio/hugo/pull/6370

jwflory:
Hi @jmooring, thanks for the quick review! Given that @felicianotech already sent a more polished version of this change, I will close this PR in favor of #6370.

PR #9194, opened 2021-11-21
Avoid sticky errors in subsequent *Init.Do calls

### LINKED ISSUE #7043: Hugo not recover from error after failed to parse invalid .Data
Since you are working on #7042 , i believe this bug also in the same topic.

<pre>
Hugo Static Site Generator v0.68.0-DEV-B1106F87/extended linux/amd64 BuildDate: 2020-03-11T08:17:58+0700
</pre>

Yes.

### When hugo failed to parse data in `/data` directory, Hugo not recover even after we fixed the invalid data

it happened to all three formats (`toml`,`yaml`,`json`)

Let's says we have this `/data/testdata.json`. 
```json
{
  "key": "value",
  "array": [
    "value1",
    "value2"
  ]
}
```
### Reproducing The Bug
1. Make it becomes invalid JSON by removing a comma or anything. 
2. Hugo throws ERROR. 
```
ERROR 2020/03/11 08:42:34 failed to load data: failed to load data: "data-bug-reload/data/test_json.json:10:1": unmarshal failed: invalid character '"' after object key:value pair
```
3. Fixed the JSON, make it valid JSON.
4. HUGO still throwing ERROR.

but after editing a `config` file, Hugo trigger full rebuild, then the ERROR gone.

#### a repo to repoduce the bug 
[hugo-data-error-reload](https://github.com/pamubay/hugo-data-error-reload)

ptgott:
When *Init.Do executes the *Init's f function, it stores the error
in the *Init's err field and returns ini.err. If we call an *Init's
Do method again, and there is no error in the subsequent call,
ini.err remains unchanged, and *Init.Do returns it again.

This becomes an issue for the "hugo server" command, which calls
h.init.data.Do within *HugoSites.Data every time the site is
rebuilt. If "hugo server" is started with a valid data file, and
that data file is edited to make it invalid (e.g., with poorly
formatted JSON), then any attempt to fix the data file only causes
Hugo to return the same error.

The fix is to ensure that *Init.Do resets ini.err to nil with each
call. The err field of lazy.Init is not used outside of *Init.Do,
so I removed it in order to simplify the code slightly and ensure
that errors would not be part of an *Init's state. Instead, this
change declares an error variable at the top of *Init.Do and
assigns it in place of ini.err.

After making this change, the go race detector began reporting
a race in TestImageOps between one goroutine created in
*Site.renderPages:

```go
go pageRenderer(ctx, s, pages, results, wg)
```

...and another goroutine created in *Init.withTimeout:

```go
	go func(cv chan verr) {
		v, err := f(ctx)
		c <- verr{v: v, err: err}
	}(c)
```

The race affected *pageContentOutput.content. To
prevent this race, I added getter and setter methods for
pageContentOutput.content that make use of a sync.RWMutex.
Otherwise, this field is assigned/read directly by multiple
goroutines.

I made other changes in order to refactor the code and prevent
unintentional memory sharing:

- Removed lazy.AddWithTimeout and its tests. This function is not
  called anywhere, and removing it makes it slightly easier to
  make sense of the lazy package.

- Edited the *Init.withTimeout to remove a redundant select
  statement.

- Edited *onceMore.Do to return an error, making it a bit
  more straightforward to share errors between its function
  argument and its caller without risking unintentional memory
  sharing.

Fixes #7043

### DIFF
--- a/lazy/init.go
+++ b/lazy/init.go
@@ -35,7 +35,6 @@ type Init struct {
 
 	init onceMore
 	out  interface{}
-	err  error
 	f    func() (interface{}, error)
 }
 
@@ -45,13 +44,6 @@ func (ini *Init) Add(initFn func() (interface{}, error)) *Init {
 		ini = New()
 	}
 	return ini.add(false, initFn)
-}
-
-// AddWithTimeout is same as Add, but with a timeout that aborts initialization.
-func (ini *Init) AddWithTimeout(timeout time.Duration, f func(ctx context.Context) (interface{}, error)) *Init {
-	return ini.Add(func() (interface{}, error) {
-		return ini.withTimeout(timeout, f)
-	})
 }
 
 // Branch creates a new dependency branch based on an existing and adds
@@ -76,15 +68,17 @@ func (ini *Init) Do() (interface{}, error) {
 		panic("init is nil")
 	}
 
-	ini.init.Do(func() {
+	var err error
+
+	err = ini.init.Do(func() error {
 		prev := ini.prev
+		var e error
 		if prev != nil {
 			// A branch. Initialize the ancestors.
 			if prev.shouldInitialize() {
-				_, err := prev.Do()
-				if err != nil {
-					ini.err = err
-					return
+				_, e = prev.Do()
+				if e != nil {
+					return e
 				}
 			} else if prev.inProgress() {
 				// Concurrent initialization. The following init func
@@ -94,23 +88,26 @@ func (ini *Init) Do() (interface{}, error) {
 		}
 
 		if ini.f != nil {
-			ini.out, ini.err = ini.f()
+			ini.out, e = ini.f()
+			if e != nil {
+				return e
+			}
 		}
 
 		for _, child := range ini.children {
 			if child.shouldInitialize() {
-				_, err := child.Do()
-				if err != nil {
-					ini.err = err
-					return
+				_, e = child.Do()
+				if e != nil {
+					return e
 				}
 			}
 		}
+		return nil
 	})
 
 	ini.wait()
 
-	return ini.out, ini.err
+	return ini.out, err
 }
 
 // TODO(bep) investigate if we can use sync.Cond for this.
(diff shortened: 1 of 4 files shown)


ptgott:
Seems like my changes trip the go race detector (https://github.com/gohugoio/hugo/runs/4280641814?check_suite_focus=true#step:16:203), probably because of something having to do with goroutine safety in `*Init.Do`. I'll take another look later this week.

lazy/init.go
```diff
@@ -76,14 +75,15 @@ func (ini *Init) Do() (interface{}, error) {
 		panic("init is nil")
 	}
 
+	var err error
+
 	ini.init.Do(func() {
```
bep, line 80: Wouldn't doing a `ini.err = nil` here have the same functional effect?

ptgott: I wanted to remove the `ini.err` field since it wasn't used outside this function in order to simplify the `Init` struct a bit and reduce the risk of accidental memory sharing.

bep:
Can you post the full stack trace of the data race on `content`? I'm not sure I follow this.

bep:
I think I have a simpler fix for this, will create a PR in a minute.

ptgott:
@bep The stack trace of the data race is in [this gist](https://gist.github.com/ptgott/3536c8e9983461593b8b37f525174dc1), beginning at line [106](https://gist.github.com/ptgott/3536c8e9983461593b8b37f525174dc1#file-gistfile1-txt-L106). The gist records the output of removing the locking logic from my PR and running `go test ./... -race`. 

Any ideas as to why this race took place after removing `Init.err` and using a local variable instead? This seemed strange to me because, as far as I know, no other part of the codebase referred to the `err` field outside of `*Init.Do`. 

On another note, after PR #9232, I think there needs to be a way to ensure that `Reset` is called after `Do`, otherwise this bug could come back in a way that's even more difficult to troubleshoot, since assigning `*Init.err` to `nil` is a side-effect of `Reset`. Maybe at least indicate this as a comment above `Do`?

PR #9195, opened 2021-11-22
A new config option to fill translation gaps

goose3228:
Introduces a new config option `mergeLangContent` to fill translation gaps with another language. It is somewhat similar to existing langs.Merge function, but literally copies content on build level, thus fixing the issue of a visitor leaving his native locale by clicking on outputs of langs.Merge.

Here is an example:
```toml
# config.toml
[languages]
[languages.en]
  ...
[languages.es]
  ...
  mergeLangContent = "en"
```

```
# Content tree (note missing translation for post.md):
content/
├── _index.es.md
├── _index.md
├── post2.es.md
├── post2.md
└── post.md

# Output in public:
public/
├── index.html
├── post/index.html
├── post2/index.html
├── es
│   ├── index.html
│   ├── post/index.html
│   ├── post2/index.html
```
Here `/es/post2.index.html` would have the same content as english version, except it would keep spanish header, footer and so on.

Similar behavior is set for i18n translations - if translation file is missing, merged language will take precedence over defaultContentLanguage.

To exclude a single page from this, there is a front matter option `ignoreOnLangMerge`.

#### A note on performance

The current implementation reads content file one more time whenever there is a need to merge it. For me it seems okay for the following reasons:
- I already use bash script to do the same job in production by just copying files in CI
- Current implementation is 2x+ faster than reading copies
- Can't be fixed without rewriting existing methods

### DIFF
--- a/docs/content/en/content-management/multilingual.md
+++ b/docs/content/en/content-management/multilingual.md
@@ -494,10 +494,10 @@ defaultContentLanguage = "en"
 [languages]
 [languages.en]
 title = "My blog"
-mergeLangContentTo = "es"
 
 [languages.es]
 title = "Mi blog"
+mergeLangContent = "en"
 {{< /code-toggle >}}
 
 In this example, if you have the following content:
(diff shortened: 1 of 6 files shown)


goose3228:
Currently seeking help on how to write tests for it, or waiting for a spare time to figure out, but the feature itself is ready for review.

jmooring:
@goose3228 

This functionality is available in v0.96.0 via content mounts. See:
<https://github.com/gohugoio/hugo/releases/tag/v0.96.0>

Issue #9228, opened 2021-12-02
Make `resources.Get` use a file cache for remote resources.


bep:
As I mentioned briefly, I'm in the process of improving some core things in Hugo (including some cache work), but I think we should get the recent remote support in `resources.Get` + XML support out in its own Hugo release.

To do that, I think we should make the remote `resources.Get` cache to disk and not to memory (as it is now).

See:

https://github.com/gohugoio/hugo/blob/master/tpl/data/resources.go#L47

I suggest that we set up a new named file cache, name it `getresource`, and update the docs about it here:

https://gohugo.io/getting-started/configuration/#configure-file-caches

Perhaps make a note about its usage in `resources.Get` but only for _remote resources_.

@vanbroup something you would want to take on and implement?

PR #9230, opened 2021-12-02
Make resources.Get use a file cache for remote resources

vanbroup:
Closes: #9228

### DIFF
--- a/resources/resource_factories/create/create.go
+++ b/resources/resource_factories/create/create.go
@@ -161,7 +161,7 @@ func (c *Client) FromRemote(uri string, options map[string]interface{}) (resourc
 
 	resourceID := helpers.HashString(uri, options)
 
-	_, httpResponse, err := c.cacheGetResource.GetOrCreateBytes(resourceID, func() ([]byte, error) {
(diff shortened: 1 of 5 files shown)


resources/resource_factories/create/create.go
```diff
@@ -156,10 +161,7 @@ func (c *Client) FromRemote(uri string, options map[string]interface{}) (resourc
 
 	resourceID := helpers.HashString(uri, options)
 
-	// This caches to memory and will, in server mode, not be evicted unless the resourceID changes
-	// or the server restarts.
-	// There is ongoing work to improve this.
-	return c.rs.ResourceCache.GetOrCreate(resourceID, func() (resource.Resource, error) {
+	_, httpResponse, err := c.cacheGetResource.GetOrCreateBytes(resourceID, func() ([]byte, error) {
```
bep, line 164: All of this looks very goo. I one comment; I suspect it would be more (memory) effective to to use `GetOrCreate` (which returns a `io.ReadCloser` -- but then you need to remember to close it when done (or in a defer).

Issue #9238, opened 2021-12-03
Text filter that draws text with the given options


vanbroup:
It would be great if Hugo could dynamically generate Open Graph Images.

```
{{ $img := resources.Get "/images/open-graph-bg.png"}}
{{ $img = $img.Filter (images.Text "This text is placed over the image" (dict
    "color" "#ffffff"
    "size" 50
    "x" 100
    "y" 200
))}}
```

PR #9239, opened 2021-12-03
Text filter that draws text with the given options

vanbroup:
Closes: #9238


vanbroup:
@bep I'm creating a Text filter that overlays the given text on an image according to the given options. I would like to hear your thoughts on the current approach, and specifically on how you would like to handle fonts.

I was thinking that fonts could be retrieved using the `resources.Get` function, from local storage or from a remote location such a Google Fonts. While the method below loads the font and passes it on as an option (as `resources.resourceAdapter`), this doesn't allow me to use it.

```
{{ $font := resources.Get "https://github.com/google/fonts/raw/main/apache/roboto/static/Roboto-Regular.ttf" }}
{{ $img := resources.Get "/images/open-graph-bg.png"}}
{{ $img = $img.Filter (images.Text "This text is placed over the image" (dict
    "color" "#ffffff"
    "size" 50
    "x" 100
    "y" 200
    "font" $font
))}}
```

bep:
> While the method below loads the font and passes it on as an option (as resources.resourceAdapter), this doesn't allow me to use it.

Without discussing if it's a great idea or not, I'm not sure what you mean by "doesn't allow me to use it"? You should be able to cast it to `resource.Resource`.

That said, I think this issue would be mentally easier to manage if we took it in two steps and just used the default font in the first iteration.

bep:
If we drop the font discussion for a minute; I tested this, and it works pretty great.

```
{{ $img = $img.Filter (images.Text "This text is placed over the image" (dict
    "color" "#ffffff"
    "size" 50
    "x" 100
    "y" 200
))}}
```

I have initial comments after testing it:

1. I would expect x=0,y=0 to place the text in the upper corner. Seems like the y is the bottom of the text and not the top.
2. I would expect `images.Text "This text is placed over the image"` to work without options with some sensible defaults.

vanbroup:
Agreed, addressed your comments

resources/images/filters.go
```diff
@@ -33,6 +35,47 @@ func (*Filters) Overlay(src ImageSource, x, y interface{}) gift.Filter {
...
+				tf.size = cast.ToFloat64(v)
+			case "x":
+				tf.x = cast.ToInt(v)
+			case "y":
+				tf.y = cast.ToInt(v)
+			}
+
+		}
+
+		if w, ok := opt["font"].(fontSource); ok {
```
bep, line 68: OK, I admit, it would be useful to set a custom font, but then this needs some more work. As it is now, I'm pretty sure that `HashString` will return a new value on every build/restart, and that doesn't work.

I'm doing some related work in this area, so I suggest that we for this case make it simple:

* Fetch the font if set
* Replace the `font` in options with `font.Key` (there is an interface)
* Create the options hash

bep (changes requested):

resources/images/text.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// Correct y position based on font and size
+	f.y = f.y + fontHeight
+
+	// Start position
+	y := f.y
+	d.Dot = fixed.P(f.x, f.y)
+
+	// Draw text and break line at max width
+	parts := strings.Split(f.text, " ")
```
bep, line 96: strings.Fields.

resources/images/filters.go
vanbroup: I have removed the font for now, but custom fonts make the Text filter much more useful; this would be a requirement for many users.

Not sure if I understand you suggestion above.

docs/content/en/functions/images/index.md
```diff
@@ -51,14 +51,12 @@ images.Text TEXT DICT)
 The following example will add the text `Hugo rocks!` to the image with the specified font, color, size and position.
 
 ```
-{{ $font := resources.Get "/fonts/font.ttf" }}
```
bep: I may have given conflicting feedback re font -- and we can take that in its own separate feature -- but I now think this would be really useful. But see my comment re. hash key.

vanbroup: Do you want to include font support for this PR or work that feature out in a separate PR?

If we implement fonts, we need to make sure that the pipeline and font is cached between build/restart. Using resources.Get we utilize the existing logic, for local and remote fonts, including the resource cache. I'm not 100% sure about the pipeline cache, during testing, I had to make changes to the pipeline (even after a restart) to see the changes, but I can imagine that the resource pointer is unpredictable.

I also thought about passing the resource location instead and call the resources get internally but that would cause an import loop.

vanbroup: On the hash key, I suppose you refer to [image.go#L237](https://github.com/gohugoio/hugo/blob/master/resources/image.go#L237)?

What do you mean with:
> Replace the font in options with font.Key (there is an interface)

If we want to use resources.Get to support local and remote fonts, I need to pass the resource at some point.

bep: There is a more complex story to this, but the short story is:

* If you do what you did and pass the `Resource` (as in: what you get in `resources.Get` etc.) as an option, the caching is in place.
* But, and this is important, you need to make sure that what you create in `HashString` is stable -- which is why I made the comment about:

1. First pull the font from the map.
2. Then replace `font` in the map with `fontResource.Key`.
3. Then use that map to create the hash.

You can do the above in its own PR if you want, or just apply that change to this, if that is most effective.

resources/images/filters.go
```diff
@@ -33,6 +35,46 @@ func (*Filters) Overlay(src ImageSource, x, y interface{}) gift.Filter {
...
+				tf.y = cast.ToInt(v)
+			case "linespacing":
+				tf.y = cast.ToInt(v)
+			}
+
+		}
+	}
+
+	return filter{
+		Options: newFilterOpts(text, options),
```
bep, line 73: @vanbroup just to be clear re the cache key vs font; the `options`  map cannot contain a pointer (unless it implements a Hashable interface, but let's not go down that route her), as that will change on every build/restart. So you need to replace the `font` key with its `.Key` value.

vanbroup: > So you need to replace the font key with its .Key value.
Do you mean the `resource.Key`? How would that allow me to access the font later on?

bep: If you do:

```
{{ $font := resource.Get "https://example.org/font.ttf" }}
{{ $img = $img | images.Filter (images.Text "foo" (text font: $font ))}}
```

Then 

1. `$font` gets cached after first invocation.  
2. The images.Text filter will be invoked once for each property combination (which is why we want a stable font Key).


But note that if the font loading (from the Reader) in itself is an expensive operation, that is not covered by the above.

vanbroup: In the last commit I updated the font value for `newFilterOpts` to `$font.Key()`, is that what you mean? 

> `{{ $img = $img | images.Filter (images.Text "foo" (text font: $font ))}}`

did you mean:

`{{ $img = $img | images.Filter (images.Text "foo" (dict "font" $font ))}}`

p.s. Feel free to push directly to this branch, as maintainer you should have the permission to do so, I don't want to keep bothering you while I'm learning the logic of the code.

resources/images/filters.go
```diff
@@ -33,6 +35,53 @@ func (*Filters) Overlay(src ImageSource, x, y interface{}) gift.Filter {
...
+		size:        20,
+		x:           10,
+		y:           10,
+		linespacing: 2,
+	}
+
+	var opt map[string]interface{}
+	if len(options) > 0 {
+		var err error
+		opt, err = maps.ToStringMapE(options[0])
```
bep, line 52: I suggest you just use `maps.MustToParamsAndPrepare`; this makes all keys into lowercase, which makes the compare below simpler.

resources/images/filters.go
```diff
@@ -33,6 +35,53 @@ func (*Filters) Overlay(src ImageSource, x, y interface{}) gift.Filter {
...
+			case "color":
+				tf.color = cast.ToString(v)
+			case "size":
+				tf.size = cast.ToFloat64(v)
+			case "x":
+				tf.x = cast.ToInt(v)
+			case "y":
+				tf.y = cast.ToInt(v)
+			case "linespacing":
+				tf.y = cast.ToInt(v)
```
bep, line 68: tf.linespacing.

resources/images/filters.go
bep: I suggest that we complete this PR without the `font` attribute -- I can create suggested PR for that after and see how that works. I'm planning a Hugo release in the next few days, and this would be a great feature to get better OpenGraph images on the Hugo docs site.

vanbroup: Removed font attribute as suggested

PR #9278, opened 2021-12-12
Improve handling of remote image/jpeg resources

### LINKED ISSUE #9275: Remote image gets: *resources.genericResource is not an image

`hugo v0.90.1-48907889+extended darwin/amd64 BuildDate=2021-12-10T10:56:41Z VendorInfo=gohugoio`

Yes

I'm trying to fetch an image remotely (https://images.unsplash.com/photo-1611323593958-0fec16fa2909) and grab its `.Width` and `.Height` or use `.Resize`. But the following ensue: `error calling Width: *resources.genericResource is not an image` or `error calling Resize: *resources.genericResource is not an image`

Note that `MainType` and `SubType` are `image` and `jpeg` respectively.

This might be because the URL do not point to a file per say, but and endpoint serving a file. But then, the MainType/SubType are misleading and prevent any testing before trying to apply image transformation.

```
  {{ with resources.Get "https://images.unsplash.com/photo-1611323593958-0fec16fa2909" }}
  {{ with .Err }}
    {{ warnf "%s" . }}
  {{ else }}
    {{ with .Resize "300x"}}
    {{ end }}
  {{ end }}
```

jmooring:
Closes #9275

### DIFF
--- a/resources/images/config.go
+++ b/resources/images/config.go
@@ -34,6 +34,9 @@ var (
 	imageFormats = map[string]Format{
 		".jpg":  JPEG,
 		".jpeg": JPEG,
+		".jpe":  JPEG,
+		".jif":  JPEG,
+		".jfif": JPEG,
 		".png":  PNG,
 		".tif":  TIFF,
 		".tiff": TIFF,
--- a/media/mediaType.go
+++ b/media/mediaType.go
@@ -177,7 +177,7 @@ var (
 
 	// Common image types
 	PNGType  = newMediaType("image", "png", []string{"png"})
-	JPEGType = newMediaType("image", "jpeg", []string{"jpg", "jpeg"})
+	JPEGType = newMediaType("image", "jpeg", []string{"jpg", "jpeg", "jpe", "jif", "jfif"})
 	GIFType  = newMediaType("image", "gif", []string{"gif"})
 	TIFFType = newMediaType("image", "tiff", []string{"tif", "tiff"})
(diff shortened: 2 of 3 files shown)


bep:
I think it's better to add `jpe` to the MediaType configuration -- also, we should fix the "is not an Image" error message to display some information about what extension and MediaType Hugo guessed.

jmooring:
Yeah, I went back and forth on that. Instead of choosing the first extension from a somewhat arbitrary list that can vary from system to system, I thought it would be better to choose the "best" one, ensuring consistent builds across multiple systems. But I will make the change.

Also, please see DM on forum.

resources/images/config.go
```diff
@@ -32,8 +32,10 @@ import (
 
 var (
 	imageFormats = map[string]Format{
-		".jpg":  JPEG,
+		".jfif": JPEG,
+		".jpe":  JPEG,
 		".jpeg": JPEG,
+		".jpg":  JPEG,
 		".png":  PNG,
```
bep, line 39: OK, I think I misunderstood the problem. I thought this was a "media type" problem. I see that there are some rooms for improvements in the above, but I suggest that you instead just do the mapping to `jpg` early (inside `FromRemote`).

jmooring: Well, we should also be able to process local files with `.jpe` and `.jfif` extensions, so I'll enable that. But yeah, with image/jpeg, it would nice to say "use the .jpg extension if an extension wasn't provided." So I'll do that too.

bep: OK

Issue #7310, opened 2020-05-24
Check Site.LastChange and Home.LasMod


bep:
See https://discourse.gohugo.io/t/a-way-to-get-last-commit-date-for-the-whole-repo/25665/8

jmooring:
To clarify intent, with `enableGitInfo = true`, the `.Site.LastChange` value will only change if the _content_ changes, correct?

For example, if I modify `.gitignore` and commit the change, `.Site.LastChange` will remain unchanged. Correct? This is what I would expect.

bep:
>value will only change if the content changes, correct?

Yes, the GitInfo is used to set dates to individual files (which then is used to determine `.Site.LastChange`). Only the content files (e.g. `.md`) is considered here.

eallion:
I think all files fetch the last commit is better.
I used a shell script to display the last commit hash in the `footer`.
`GitInfo` does not working well, because `footer` is a global tpl. 

@bep

jmooring:
> .Site.LastChange chould show the most recent non-zero .LastMod

It does.

> .Site.Home.Lastmod will be the same as Site.LastChange if you have no content file [or date] for the home page

It is.

> else it will get that date.

It does.

This works as described.

PR #9283, opened 2021-12-13
rss template: Set correct lastBuildDate

vanbroup:
The embedded rss template does not use the current date for the `lastBuildDate`:

Example: https://gohugo.io/index.xml:
```
<rss xmlns:atom="http://www.w3.org/2005/Atom" version="2.0">
<channel>
<title>Hugo – The world’s fastest framework for building websites</title>
<link>https://gohugo.io/</link>
<description>Recent Hugo news from gohugo.io</description>
<generator>Hugo -- gohugo.io</generator>
<language>en-us</language>
<lastBuildDate>Thu, 02 Mar 2017 12:00:00 -0500</lastBuildDate>
...
```


tpl/tplimpl/embedded/templates/_default/rss.xml
```diff
@@ -21,7 +21,7 @@
     <managingEditor>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</managingEditor>{{end}}{{ with .Site.Author.email }}
     <webMaster>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</webMaster>{{end}}{{ with .Site.Copyright }}
     <copyright>{{.}}</copyright>{{end}}{{ if not .Date.IsZero }}
-    <lastBuildDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</lastBuildDate>{{ end }}
+    <lastBuildDate>{{ now.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</lastBuildDate>{{ end }}
```
bep, line 24: So, `lastBuildDate` should signal change, so we cannot use `now`. But I agree that the current value is not correct -- the best we can do is to use `.Lastmod`.

vanbroup: Now is a bit aggressive, it should indicate the last change of any of the entries, might be possible, let me play with it.

bep: As I said, `.Lastmod` is the value to use.

vanbroup: It looks like pages are sorted to `desc date` by default, updated to:

```
    {{- with (index $pages 0) -}}
    <lastBuildDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</lastBuildDate>{{ end }}
    {{- end -}}
```

If that is not guaranteed we could the following lines after the pages have been selected:
```
{{- $pages = sort $pages "Date" "desc" }}
```

vanbroup: > As I said, `.Lastmod` is the value to use.

Let me check that as well.

jmooring: https://github.com/gohugoio/hugo/issues/7310#issuecomment-981137769

vanbroup: > > As I said, `.Lastmod` is the value to use.
> 
> Let me check that as well.

The `Lastmod` value is from the main page, not relevant to the last new entry to the feed.

Sometimes it could make sense to use the most recent Lastmod value of the all pages (e.g., when a path changed) but for that I would need to create a separate sorting of the included pages and limit to the first entry.

vanbroup: > [#7310 (comment)](https://github.com/gohugoio/hugo/issues/7310#issuecomment-981137769)

Thanks @jmooring 

`.Site.LastChange` works well and is definitely the winner in performance, maybe not 100% accurate (as the pages in this feed might not have not changed) but I don't think that matters much.

vanbroup:
Sorry for all these commits, tested and confirmed ok now.

tpl/tplimpl/embedded/templates/_default/rss.xml
```diff
@@ -21,7 +21,7 @@
     <managingEditor>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</managingEditor>{{end}}{{ with .Site.Author.email }}
     <webMaster>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</webMaster>{{end}}{{ with .Site.Copyright }}
     <copyright>{{.}}</copyright>{{end}}{{ if not .Date.IsZero }}
-    <lastBuildDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</lastBuildDate>{{ end }}
```
bep, line 24: Hugo can have RSS feed on many levels: top level, per section, per category (taxonomy); I thinks this date should best reflect the last modification date of the feed itself, which is `.Lastmod`.

bep (changes requested):

tpl/tplimpl/embedded/templates/_default/rss.xml
vanbroup: Simply `.Lastmod`  doesn't work, my previous commit would get that behavior calculated:
 https://github.com/gohugoio/hugo/commit/22463e919fa9ef76f1ae40d793b4262b75435bc2

Should I revert to that?

bep: If `Lastmod` doesnt't do what you want, we can certainly discuss that, but you need to bring it up in another issue.

PR #9297, opened 2021-12-16
rss template: Set lastBuildDate to latest included page modification

vanbroup:
Set `lastBuildDate` to `Lastmod` of the pages included in the feed.

- Using `Lastmod` gives the last modified date of the top level of the tree but not of the pages included in the feed.
- Using `Site.LastChanged` as suggested by @jmooring gives the date of the last change to the site, it's more effective than `Date` or `Lastmod` as these would indicate no changes when there might be.

(sorry, I unintentionally closed #9283)

### DIFF
--- a/tpl/tplimpl/embedded/templates/_default/rss.xml
+++ b/tpl/tplimpl/embedded/templates/_default/rss.xml
@@ -10,7 +10,6 @@
 {{- if ge $limit 1 -}}
 {{- $pages = $pages | first $limit -}}
 {{- end -}}
-{{- $lastModPage := (sort $pages "Lastmod" "desc") | first 1 }}
 {{- printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
 <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
   <channel>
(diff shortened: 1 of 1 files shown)


tpl/tplimpl/embedded/templates/_default/rss.xml
```diff
@@ -10,6 +10,7 @@
 {{- if ge $limit 1 -}}
 {{- $pages = $pages | first $limit -}}
 {{- end -}}
+{{- $lastModPage := (sort $pages "Lastmod" "desc") | first 1 }}
```
bep, line 13: There is a `.ByLastmod` method on the `Pages` type.

bep, line 13: There is a `.ByLastmod` method on the `Pages` type.

bep: But, I would prefer if we could calculate this date by looping instead of sorting.

vanbroup: I updated the condition to `with index $pages.ByLastmod.Reverse 0` and removed this `$lastModPage` definition which seems the cleanest solution.

```
{{- with index $pages.ByLastmod.Reverse 0 -}}{{- if not .Lastmod.IsZero -}}
    <lastBuildDate>{{ .Lastmod.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</lastBuildDate>
{{- end -}}{{- end -}}
```

PR #9298, opened 2021-12-16
Allow for return partials with falsy arguments

### LINKED ISSUE #7528: No proper return from partials with `0` as arg
Passing `0` integer as an argument to  partial template func with return statement (as a function substitute) prevents this "function" from returning the correct value. It looks like the `0` as partial argument is treated as `nil` context.

Template:
```
add42: 50: {{ partial "add42.tpl" 8 }}
add42: 42: {{ partial "add42.tpl" 0 }}
```

Output:
```
add42: 50: 50
add42: 42: 
```

File `partials/add42.tpl`
```
{{ $v := add . 42 }}
{{ return $v }}
```

This issue is reproducable with the latest release.

ptgott:
Partials with returns values are parsed, then inserted into a
partial return wrapper via wrapInPartialReturnWrapper in order
to assign the return value via *contextWrapper.Set. The
predefined wrapper template for partials inserts a partial's nodes
into a "with" template action in order to set dot to a
*contextWrapper within the partial. However, because "with" is
skipped if its argument is falsy, partials with falsy arguments
were not being evaluated.

This replaces the "with" action in the partial wrapper with a
"range" action that isn't skipped if .Arg is falsy.

Fixes #7528

### DIFF
--- a/tpl/tplimpl/template_ast_transformers.go
+++ b/tpl/tplimpl/template_ast_transformers.go
@@ -112,7 +112,7 @@ func getParseTree(templ tpl.Template) *parse.Tree {
 }
 
 const (
-	partialReturnWrapperTempl = `{{ $_hugo_dot := $ }}{{ $ := .Arg }}{{ with .Arg }}{{ $_hugo_dot.Set ("PLACEHOLDER") }}{{ end }}`
+	partialReturnWrapperTempl = `{{ $_hugo_dot := $ }}{{ $ := .Arg }}{{ range (slice .Arg) }}{{ $_hugo_dot.Set ("PLACEHOLDER") }}{{ end }}`
 )
 
 var partialReturnWrapper *parse.ListNode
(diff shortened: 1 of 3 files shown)


ptgott:
`mage -v test` failed on my local branch with tests that also fail in `origin/master`, so fingers crossed for the CI pipeline.

bep:
Thanks for this, I thought long and hard about this problem ... never thought about using the `range` keyword. I will do some tests on my own tomorrow, but it it works, I'll merge it. Thanks!

tpl/tplimpl/template_ast_transformers.go
```diff
@@ -112,7 +112,7 @@ func getParseTree(templ tpl.Template) *parse.Tree {
 }
 
 const (
-	partialReturnWrapperTempl = `{{ $_hugo_dot := $ }}{{ $ := .Arg }}{{ with .Arg }}{{ $_hugo_dot.Set ("PLACEHOLDER") }}{{ end }}`
+	partialReturnWrapperTempl = `{{ $_hugo_dot := $ }}{{ $ := .Arg }}{{ range (slice .Arg) }}{{ $_hugo_dot.Set ("PLACEHOLDER") }}{{ end }}`
```
bep, line 115: I suspect it would be easier to understand if you make `contextWrapper.Arg` a `[]interface{}` -- and just add a comment there explaining why it's a slice with a single value.

ptgott: I gave this a go, though my attempt at a comment depended on explaining what we were doing with the `partialReturnWrapperTempl`. In the end, I found it simpler to add a comment above `partialReturnWrapperTempl` while leaving `contextWrapper.Arg` as it was. How does that sound?

PR #9322, opened 2021-12-26
Remove unused helpers URLizeAndPrep() and URLPrep()

dgnemo:
Small change while I am getting familiar with the Hugo codebase.
These 2 methods seems unused.


bep:
Thanks, but I have removed these in #9334 -- and merging this would only complicate that merge.

PR #9344, opened 2022-01-03
tpl: Handle `audio` front matter as an array in opengraph

takuti:
Fix a bug that contradicts what the [Front Matter documentation](https://gohugo.io/templates/internal) says:

> audio and videos are URL arrays like images for the audio and video
> metadata tags, respectively.

`audio` was not treated as an array, while `videos` were fine.

### DIFF
--- a/hugolib/embedded_templates_test.go
+++ b/hugolib/embedded_templates_test.go
@@ -92,6 +92,39 @@ title: My Site
 `)
 }
 
+func TestInternalTemplatesAudioVideos(t *testing.T) {
+	config := `
+baseURL = "https://example.org"
+`
+	b := newTestSitesBuilder(t).WithConfigFile("toml", config)
+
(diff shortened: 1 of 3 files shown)


takuti:
Hi @jmooring any update/plan for this? It is more like a bug and the current Hugo version keeps generating an invalid meta tag like:

```
<meta property="og:audio" content="[/audio/recording.mp3]" />
```

takuti:
@jmooring Checking in again - Could you please share what's the plan and when this PR is reviewed/merged?

jmooring:
@takuti I am a contributor, not the project lead.

takuti:
@bep Can I hear who can handle this bug fix PR? It sounds like @jmooring is not an appropriate person while the EmbeddedTemplate label is added by the project member.

Any follow up would be greatly appreciated.

PR #9382, opened 2022-01-14
deps: Update github.com/tdewolff/minify/v2 v2.9.22 => v2.9.29

### LINKED ISSUE #9132: JS destructuring assignment minification bug
Hi, I noticed that Hugo has a problem minifying my javascript. This is my pipeline:

``` go template
{{ $jspath := "js/main.ts" }}
{{- $jscratch := newScratch -}}

{{ if eq hugo.Environment "production" -}}
  {{ $jscratch.Set "mainjs" (resources.Get $jspath | js.Build | minify | fingerprint) -}}
{{- else -}}
  {{ $jscratch.Set "mainjs" (resources.Get $jspath | js.Build | fingerprint) -}}
{{- end -}}

{{ $mainjs := $jscratch.Get "mainjs" -}}

<script src="{{ $mainjs.RelPermalink }}" integrity="{{ $mainjs.Data.Integrity }}" defer></script>
```

It worked fine in development using `hugo server` but when I built for production using `hugo --minify` I would get an error about an unexpected colon, which means the problem is with Hugo's built-in minification function

I traced it to one of Splide's (v3.2.5) source files which I'm importing like so:

``` typescript
import Splide from '@splidejs/splide';
```

The source file that Hugo's minifier is having an issue with (after being bundled by esbuild) is `@splidejs/splide/src/js/utils/math/math/math.ts`, which contains:

``` typescript
export const { min, max, floor, ceil, abs } = Math;
```

Apparently, after running through esbuild and Hugo's minifier it results in the following summarized code, which is invalid:

``` javascript
var S, O, az, H, g // These are declared with a ton of other variables
// . . . Many characters later . . .
{min:S,max:O,floor:az,ceil:H,abs:g}=Math; // And this where the error occurs.
```

I was able to get around this by letting esbuild perform the minification instead (which I didn't know you could do at the time of discovering this error):

``` go template
  {{ $jscratch.Set "mainjs" (resources.Get $jspath | js.Build (dict "minify" true) | fingerprint) -}}
```

Which is probably the right way to minify my JS in this situation, but I decided to report this as a bug anyway because Hugo's minify tool is taking valid JS from esbuild and returning invalid JS.

```console
$ hugo version
hugo v0.88.1-5BC54738+extended windows/amd64 BuildDate=2021-09-04T09:39:19Z VendorInfo=gohugoio
```

Yes.

### LINKED ISSUE #9244: Cannot minify JavaScript file contains a bare Unicode escape identifier
The following JavaScript source code seems odd, but it is completely legal. (See [ECMA-262](https://262.ecma-international.org/12.0/#prod-IdentifierStart))

```javascript
const _\u00bare_unicode_escape_identifier = 42;
console.log(_\u00bare_unicode_escape_identifier);
```

And, such a source code is generated by Scala.js.

When Hugo tries to minify a source containing a bare Unicode escape identifier, it shows the following error:

```
Start building sites …
hugo v0.89.4+extended darwin/amd64 BuildDate=unknown
Error: Error building site: MINIFY: failed to transform "index.js" (application/javascript): unexpected \u00BAre_unicode_identifier in var statement on line 3 and column 8
    3:   var _\u00BAre_unicode_identifier = 42;
              ^
```

Noting that the latest  ESBuild supports minifying source codes with a bare Unicode escape identifier, I think this problem is solved if ESBuild is updated.

<pre>
$ hugo version
hugo v0.89.4+extended darwin/amd64 BuildDate=unknown
</pre>

Yes

jmooring:
Fixes #9244
Fixes #9132
Fixes https://discourse.gohugo.io/t/36523

### DIFF
--- a/hugolib/resource_chain_test.go
+++ b/hugolib/resource_chain_test.go
@@ -699,8 +699,8 @@ Min HTML Remote: {{ ( resources.GetRemote "%[1]s/mydata/html1.html" | resources.
 		}, func(b *sitesBuilder) {
 			b.AssertFileContent("public/index.html", `Min CSS: h1{font-style:bold}`)
 			b.AssertFileContent("public/index.html", `Min CSS Remote: h1{font-style:bold}`)
-			b.AssertFileContent("public/index.html", `Min JS: var x;x=5,document.getElementById(&#34;demo&#34;).innerHTML=x*10`)
-			b.AssertFileContent("public/index.html", `Min JS Remote: var x;x=5,document.getElementById(&#34;demo&#34;).innerHTML=x*10`)
+			b.AssertFileContent("public/index.html", `Min JS: var x=5;document.getElementById(&#34;demo&#34;).innerHTML=x*10`)
+			b.AssertFileContent("public/index.html", `Min JS Remote: var x=5;document.getElementById(&#34;demo&#34;).innerHTML=x*10`)
 			b.AssertFileContent("public/index.html", `Min JSON: {"employees":[{"firstName":"John","lastName":"Doe"},{"firstName":"Anna","lastName":"Smith"},{"firstName":"Peter","lastName":"Jones"}]}`)
 			b.AssertFileContent("public/index.html", `Min JSON Remote: {"employees":[{"firstName":"John","lastName":"Doe"},{"firstName":"Anna","lastName":"Smith"},{"firstName":"Peter","lastName":"Jones"}]}`)
 			b.AssertFileContent("public/index.html", `Min XML: <hello><world>Hugo Rocks!</<world></hello>`)
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,7 @@ require (
 	github.com/spf13/jwalterweatherman v1.1.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.10.1
-	github.com/tdewolff/minify/v2 v2.9.22
+	github.com/tdewolff/minify/v2 v2.9.29
 	github.com/yuin/goldmark v1.4.4
 	github.com/yuin/goldmark-highlighting v0.0.0-20200307114337-60d527fdb691
 	gocloud.dev v0.20.0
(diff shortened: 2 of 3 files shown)


jmooring:
Changes to JS minification are substantial, so this is not without risk. However, the longer we wait...

bep:
There have been 5-6 patch releases we didn't know about? I need to check up on that `dependabot` -- it seem to only bother me with updates I don't want/care about.

hugolib/resource_chain_test.go
```diff
@@ -699,7 +699,7 @@ Min HTML Remote: {{ ( resources.GetRemote "%[1]s/mydata/html1.html" | resources.
 		}, func(b *sitesBuilder) {
 			b.AssertFileContent("public/index.html", `Min CSS: h1{font-style:bold}`)
 			b.AssertFileContent("public/index.html", `Min CSS Remote: h1{font-style:bold}`)
-			b.AssertFileContent("public/index.html", `Min JS: var x;x=5,document.getElementById(&#34;demo&#34;).innerHTML=x*10`)
+			b.AssertFileContent("public/index.html", `Min JS: var x=5;document.getElementById(&#34;demo&#34;).innerHTML=x*10`)
 			b.AssertFileContent("public/index.html", `Min JS Remote: var x;x=5,document.getElementById(&#34;demo&#34;).innerHTML=x*10`)
```
bep, line 703: Not that there was ever an ambition that these tests should somehow "unit test" the upstream library, I cannot help questioning the behaviour above: Why is is behaving differently when applied to the same JS file when pulled in remotely?

anthonyfok:
@jmooring I'm sorry for inadvertently causing a merge conflict when I merged #9294 which bumped viper to 1.10.1.  I hope I resolved that merge conflict correctly, so this PR may be "squashed and merged" after review.

hugolib/resource_chain_test.go
jmooring: > Why is it behaving differently when applied to the same JS file when pulled in remotely?

Because it's not the same JS file.

The "Min JS" test uses this data:
<https://github.com/gohugoio/hugo/blob/master/hugolib/resource_chain_test.go#L942-L944>

The "Min JS Remote" test uses this data:
<https://github.com/gohugoio/hugo/blob/master/hugolib/resource_chain_test.go#L568>

anthonyfok:
> There have been 5-6 patch releases we didn't know about? I need to check up on that `dependabot` -- it seem to only bother me with updates I don't want/care about.

I suspect Dependabot was holding back some PRs as a way to avoid merge conflict.  As soon I as merged the bump to viper 1.10.1:

* #9294

The missing Dependabot "github.com/tdewolff/minify/v2 v2.9.22 => v2.9.28" PR was finally posted:

* #9388

Not that the new minify depends on a newer version of viper, but probably merely because the two PRs modify adjacent lines in go.mod that Dependabot wanted to enforce the "correct" merge order to avoid nasty merge conflicts?

hugolib/resource_chain_test.go
jmooring: I've asked a question about the differences:
<https://github.com/tdewolff/minify/issues/455>

jmooring: Differences unexpected, resolved with new release 2.9.29.

jmooring:
Please review.

anthonyfok (approved):
(I'm not bep, but...) Looks good to me!  :-)

jmooring:
@bep A gentle nudge...

PR #9396, opened 2022-01-16
test: use `T.TempDir` to create temporary test directory

Juneezee:
A small testing enhancement.

We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete.

Reference: https://pkg.go.dev/testing#T.TempDir

### DIFF
--- a/cache/filecache/filecache_test.go
+++ b/cache/filecache/filecache_test.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


bep:
Thanks for this. I would prefer, however, that these "nice to have" tech upgrades comes as part of functional improvements. We have several ongoing development branches/pull request, and these refactorings have a tendency to just create merge conflicts for me going forward -- which creates much more work than the value of the change itself.

So, I'm closing this for now.

PR #9342, opened 2022-01-02
Fix missing page data for alternative formats

ptgott:
When a template calls the .Translations function and a
Hugo environment is using multiple output formats,
a template that calls methods like .Summary and .Len on
each translation will unexpectedly show empty return
values for these methods.

This is because each pageOutput's ContentProvider is
assigned to a page.NopPage in newPageOutput. When
*HugoSites.render assigns pageContentOutputs to
pageOutputs in *pageState.shiftToOutputFormat, it
reuses pageContentOutputs from other pageOutputs,
leaving some pageContentOutputs as NopPages. While this
approach conserves resources, sometimes it means that
a template will unexpectedly call a method on a
pageContentOutput that is actually a NopPage.

In the case of ContentProvider methods called on
translations for alternative output formats, the methods
were called on NopPages.

This change introduces LazyContentProvider, which
performs late initialization when one of its methods is
called. This way, we can reuse content in "normal" cases
but ensure that ContentProvider methods work as expected
when a pageOutput is not assigned a pageContentOutput
during the initial pre-render phase.

Fixes #8919

PR #9398, opened 2022-01-16
Ensure page translation CPs are initialized

### LINKED ISSUE #9383: Investigate .RenderString regression
See https://discourse.gohugo.io/t/multilingual-related-page-renderstring-regression-in-0-92/36555/16?u=bep

I'm pretty sure this came from https://github.com/gohugoio/hugo/pull/9342 -- but I'm also pretty sure that that PR surfaced an already existing bug re output formats and `.Translations`, so I think we need to understand and fix the real issue rather than reverting the above.

@ptgott

ptgott:
PR #9342 introduced a regression in which calling .Translations in a
template and calling RenderString on a translated Page caused a nil
pointer dereference. The issue was that some Pages returned from
.Translations had a nil cp field at the time the calling template was
being executed.

While PR #9342 had attempted to ensure that all ContentProviders were
initialized for translations at build time, it only performed the
initialization for receivers of ContentProvider methods such as
.Summary. However, the ContentProvider's *pageState.pageOutput.cp would
remain uninitialized, causing the nil pointer dereference.

This change edits the .Translations and .AllTranslations methods to
ensure that all of a page's translations have an initialized
content provider in time for a template to be executed.

Since LazyContentProvider is no longer needed with this approach, this
change also reverts the following commits:

- cdcd15b6c2abbb76fd95fbbf90365c56e82f46aa
- 25d645f47ae0a32470d303c9478c9e0b2fff0f0e

Fixes #9383

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -635,7 +635,19 @@ func (p *pageState) RenderString(args ...interface{}) (template.HTML, error) {
 		}
 	}
 
-	c, err := p.pageOutput.cp.renderContentWithConverter(conv, []byte(s), false)
+	var cp *pageContentOutput
+
+	// If the current content provider is not yet initialized, do so now.
+	if lcp, ok := p.pageOutput.ContentProvider.(*page.LazyContentProvider); ok {
+		c := lcp.Init()
+		if pco, ok := c.(*pageContentOutput); ok {
+			cp = pco
+		}
+	} else {
+		cp = p.pageOutput.cp
+	}
+
+	c, err := cp.renderContentWithConverter(conv, []byte(s), false)
 	if err != nil {
 		return "", p.wrapError(err)
 	}
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -16,14 +16,11 @@ package hugolib
 import (
 	"bytes"
 	"fmt"
-	"html/template"
 	"os"
 	"path"
 	"path/filepath"
 	"sort"
 	"strings"
-
-	"github.com/mitchellh/mapstructure"
 
 	"github.com/gohugoio/hugo/identity"
 
@@ -47,7 +44,6 @@ import (
 
 	"github.com/gohugoio/hugo/media"
 	"github.com/gohugoio/hugo/source"
-	"github.com/spf13/cast"
 
 	"github.com/gohugoio/hugo/common/collections"
 	"github.com/gohugoio/hugo/common/text"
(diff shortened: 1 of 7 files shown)


hugolib/page.go
```diff
@@ -369,12 +369,50 @@ func (p *pageState) TranslationKey() string {
...
+
+	// Hugo attempts to reuse content providers while preparing each page for
+	// rendering. This approach means that sometimes, content providers for
+	// translations are not initialized. If needed, initialize them here.
+	for _, tr := range p.translations {
+		tp, ok := tr.(*pageState)
+		if !ok {
+			continue
+		}
+		if tp.pageOutput.cp == nil {
```
bep, line 406: Thanks for this.

But I don't think we can solve this like this. Most notably is the data race which, if it's not already failing the build, it will if you move the `.Translations` into the single template.

I suspect that the cleanest, working solution we will find without doing too much work is to expand on the patch we added a few days ago: Make sure that the lazy content provider is triggered on all access to `cp` (including `RenderString`).

ptgott: Thanks for the feedback. Would you mind explaining the data race a bit more? Which goroutines, processes, I/O, etc. would be racing?

I'll spend some time early this week implementing the approach you've suggested.

bep: 1. We render pages in parallel.
2. Given the below, `AllTranslations` may be called from multiple go routines.
3. Any of the code lines reading or writing mutable page state below will be a potential data race (e.g. `if tp.pageOutput.cp == nil` and `tp.initContentProvider(cp)`). Go's race detector will fail the tests when run with `go test -race` (which we do on the CI server), but I guess you need to have test cases that trigger it.

```go
func (p *pageState) AllTranslations() page.Pages {
	p.s.h.init.translations.Do()

	// Hugo attempts to reuse content providers while preparing each page for
	// rendering. This approach means that sometimes, content providers for
	// translations are not initialized. If needed, initialize them here.
	for _, tr := range p.allTranslations {
		tp, ok := tr.(*pageState)
		if !ok {
			continue
		}
		if tp.pageOutput.cp == nil {
			cp, err := newPageContentOutput(tp, tp.pageOutput)
			if err != nil {
				// This will be caught by texttemplate.safeCall
				panic(fmt.Errorf("error rendering page translation: %w", err))
			}
			tp.initContentProvider(cp)
		}
	}

	return p.allTranslations
}
```

ptgott:
@bep I've suggested initializing a `LazyContentProvider` inside `RenderString`. If this approach looks okay, should I go ahead and see if other `*pageState` methods need a similar treatment?

hugolib/page.go
```diff
@@ -635,7 +635,19 @@ func (p *pageState) RenderString(args ...interface{}) (template.HTML, error) {
 		}
 	}
 
-	c, err := p.pageOutput.cp.renderContentWithConverter(conv, []byte(s), false)
+	var cp *pageContentOutput
+
+	// If the current content provider is not yet initialized, do so now.
+	if lcp, ok := p.pageOutput.ContentProvider.(*page.LazyContentProvider); ok {
```
bep, line 641: So, this is conceptually what we need, but I would love if we could do this in one place. I think that if we use the same "interface approach" for RenderString (and potentially others) and make LazyContentProvider implement that interface, that should work.

bep:
I have some time today and will look closer at this PR -- again, much appreciated.

bep:
@ptgott I have added a commit to your PR which is a little bit more elaborate, but it's just moving the `RenderString` and `Render` code down to where it belongs, I guess. Note that there may still be cases where `.Translations[0].SomeMethod` may still return empty (I have always thought about the translation listings as mostly informative/metadata), but at least now it's not clearly specificed by an interface.

PR #9410, opened 2022-01-20
helpers: Support complex HTML in plainify

### LINKED ISSUE #9199: Symbol '>' breaks plainify function
Also affects auto-generated Page.Summary, Page.Plain and so on.

<pre>
$ hugo version
hugo v0.89.2+extended linux/amd64 BuildDate=unknown
</pre>

Yes, with git/master too

### Test command (in template):
```
{{ print "<div data-action='click->my-controller#doThing'>qwe</div>" | plainify }}
```

### Expected output:
```
qwe
```

### Current output:
```
my-controller#doThing'qwe
```

-------------------------

P.S. arrow `->` is commonly used in html attributes by stimulus.js framework. The problem is sensible when page starts with interactive shortcode, and summary gets garbage from data attributes.

roointan:
Support complex HTML in plainify by replacing the simple HTML sripper with the sanitizer functionality of the bluemonday library, with a strict policy (to strip all HTML elements and their attributes)

Fixes #9199

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/kyokomi/emoji/v2 v2.2.8
 	github.com/magefile/mage v1.11.0
 	github.com/mattn/go-isatty v0.0.14
+	github.com/microcosm-cc/bluemonday v1.0.17 // indirect
 	github.com/mitchellh/hashstructure v1.1.0
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/muesli/smartcrop v0.3.0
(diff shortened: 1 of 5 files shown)


bep:
Bluemonday is a HTML sanitizer. That is about removing unwanted/unsafe HTML -- this is about making it into plain text, which I would hope should be a simpler task not requiring that many new dependencies. I'm closing this for now, but thanks.

bep:
Reopen, in case i don't find anything simpler.

PR #9414, opened 2022-01-22
tpl: Improve internal YouTube Shortcode

### LINKED ISSUE #3694: Improvements to internal YouTube shortcode
The built-in Hugo shortcode for embedding YouTube videos obscures captions when the user is not interacting with the video. This is a simple CSS issue related to **`overflow: hidden;`** being included inline in the go template.  I removed this CSS property in Chrome Dev Tools and everything works as expected.

## overflow: hidden;

overflow-hidden

## Removed overflow:hidden;

overflow-not-hidden

## [From GitHub](https://github.com/gohugoio/hugo/blob/da72805a4304a57362e8e79a01cc145767b027c5/tpl/tplimpl/template_embedded.go):

```
<div {{ if .Get "class" }}class="{{ .Get "class" }}"{{ else }}style="position: relative; padding-bottom: 56.25%; padding-top: 30px; height: 0; **overflow: hidden;**"{{ end }}>
  <iframe src="//www.youtube.com/embed/{{ .Get "id" }}?{{ with .Get "autoplay" }}{{ if eq . "true" }}autoplay=1{{ end }}{{ end }}" 
  {{ if not (.Get "class") }}style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;" {{ end }}allowfullscreen frameborder="0"></iframe>
</div>

```

roointan:
Added support for displaying Embedded YouTube player in 4:3 ratio, in addition to the default widescreen ratio. Also added support for all the latest parameters supported by embedded YouTube player including start/end time, looping, language, playlist, etc. Also added support for for diaplying a linked thumbnail to the YouTube video, instead of the embedded plyaer using the "linkthumbnail" parameter.

Fixes #3694

### DIFF
--- a/docs/content/en/content-management/shortcodes.md
+++ b/docs/content/en/content-management/shortcodes.md
@@ -337,7 +337,7 @@ Using the preceding `tweet` example, the following simulates the displayed exper
 
 ### `vimeo`
 
-Adding a video from [Vimeo][] is equivalent to the [YouTube Input shortcode][].
+Adding a video from [Vimeo][] is similar to the [YouTube Input shortcode][].
 
 ```
 https://vimeo.com/channels/staffpicks/146022717
@@ -375,7 +375,7 @@ Using the preceding `vimeo` example, the following simulates the displayed exper
 
 ### `youtube`
 
-The `youtube` shortcode embeds a responsive video player for [YouTube videos][]. Only the ID of the video is required, e.g.:
+The `youtube` shortcode can be used to embed a responsive video player, or add a linked thumbnail for [YouTube videos][]. Only the ID of the video is required, e.g.:
 
 ```
 https://www.youtube.com/watch?v=w7Ft2ymGmfc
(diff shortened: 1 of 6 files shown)


fgm:
Thanks a lot for linkthumbnail

gwatts:
Thanks for your work on this - I thought i'd try it and see whether it could also be used to play videos stored in Google Drive (in our case, for videos we share internally in the org).

It works great if I change the iframe source from youtube to `https://drive.google.com/file/d/{{ $id }}/preview` - Would be nice to have that as an option.  Not all of the parameters seem to be accepted by the drive embedded player, but it's still very useful

roointan:
> Thanks for your work on this - I thought i'd try it and see whether it could also be used to play videos stored in Google Drive (in our case, for videos we share internally in the org).
> 
> It works great if I change the iframe source from youtube to `https://drive.google.com/file/d/{{ $id }}/preview` - Would be nice to have that as an option. Not all of the parameters seem to be accepted by the drive embedded player, but it's still very useful

I'm very glad to find out people are already finding the changes helpful. I'm kinda new to contributing to Hugo. What you've asked for is an easy change. But I prefer to hear back from maintainers before working on more changes. Maybe you can add that to the original issue, or start a forum post? I'm not sure. (I'm afraid about maintainers not agree about adding this)
After I know supporting google drive also belongs to this shortcode, I'll be more than happy to add it.

davidsneighbour:
It's an internal shortcode named `youtube`. Not `video`. To use specifically modified versions of an internal shortcode copy the content to a file at `layouts/shortcodes/shortcodename.html` and add your changes there. It will then override the internal shortcode.

gwatts:
@davidsneighbour  You're right, it's not intended as a generic video shortcode - my point is that video streamed from google drive basically *is* youtube - It's essentially the same system; i can supply the same "controls=false" parameters, and the controls go away just as they do with youtube.  No surprise given they're both owned by Google.

 Duplicating a shortcode just to change a URL is a waste of time and gives me yet something else to track for updates and improvements.

jhenstridge:
For what it is worth, you could probably make the title and aspect ratio settings automatic using YouTube's oembed endpoint, similar to how the Twitter shortcode does with `getJSON`.

For example:

```
$ curl 'https://www.youtube.com/oembed?url=https://www.youtube.com/watch?v=dQw4w9WgXcQ' | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   793    0   793    0     0   1439      0 --:--:-- --:--:-- --:--:--  1439
{
  "title": "Rick Astley - Never Gonna Give You Up (Official Music Video)",
  "author_name": "Rick Astley",
  "author_url": "https://www.youtube.com/c/RickastleyCoUkOfficial",
  "type": "video",
  "height": 113,
  "width": 200,
  "version": "1.0",
  "provider_name": "YouTube",
  "provider_url": "https://www.youtube.com/",
  "thumbnail_height": 360,
  "thumbnail_width": 480,
  "thumbnail_url": "https://i.ytimg.com/vi/dQw4w9WgXcQ/hqdefault.jpg",
  "html": "<iframe width=\"200\" height=\"113\" src=\"https://www.youtube.com/embed/dQw4w9WgXcQ?feature=oembed\" frameborder=\"0\" allow=\"accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture\" allowfullscreen title=\"Rick Astley - Never Gonna Give You Up (Official Music Video)\"></iframe>"
}
```

Even if you ignore the HTML embed it suggests, we've got the title and thumbnail URL, and checking `width/height` tells us the video aspect ratio.

Some other thoughts:
* The oembed data suggests using i.ytimg.com for thumbnails instead of img.youtube.com. My browser sends my login cookie to the latter but not the former, so ytimg is probably better from a privacy perspective.
* The new linkthumbnail mode seems like it would be useful from a GDPR compliance standpoint. What do you think about having it enabled through the privacy configuration like the "simple" variants of other shortcodes?

Issue #7709, opened 2020-09-19
hugo server shoud print a URL with protocol, not just //localhost:1313/


schoettl:
IMO this is a **usability issue for developers**:

`hugo server` prints
```
Web Server is available at //localhost:1313/ (bind address 127.0.0.1)
```

The developer always wants to open the address in the browser but they
1. cannot click the link
2. cannot use terminal shortcuts to open the link (e.g. Alt-U in urxvt)

I understand that the address is more generic this way – it suggests "use your protocol of choice".

But it would be better if `hugo server` would print
```
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
```

- This would be OK from security perspective if the server binds to 127.0.0.1
- The URL is click-able and can be opened via terminal shortcuts
- HTTPS doesn't work anyway by default (for me)

Would you accept an PR fixing this?

davidsneighbour:
Check your config.toml. For me the http is printed in all hugo setups and is clickable in the terminal. The development server is not a production server, so https is not required.

schoettl:
Oh, thanks. `baseURL` was set to `/` (which I took from a tutorial template).

Setting it to `https://domain.tld/` works.

PR #9416, opened 2022-01-23
Append scheme to the printed server address url

adonese:
In essence, it adds http: to the printed url, when hugo is serving files
It saves the user a keystroke, since most modern terminals support links in the shell.
One can easily just ctrl+click to open the browser for that particular url.

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -519,7 +519,7 @@ func (c *commandeer) serve(s *serverCmd) error {
 			mu.HandleFunc(u.Path+"/livereload.js", livereload.ServeJS)
 			mu.HandleFunc(u.Path+"/livereload", livereload.Handler)
 		}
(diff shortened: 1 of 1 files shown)


jmooring:
With a properly configured baseURL (e.g., `baseURL = 'https://example.org/'`) the "printed server address url" is correct.

With your PR, it displays as "http:http://localhost:1313/"

See also https://github.com/gohugoio/hugo/issues/7709.

In the future, please use the forum (https://discourse.gohugo.io) for questions and troubleshooting. We prefer to use GitHub for verified bugs and vetted enhancements. Thanks.

adonese:
Totally understand! In my case, most of the time I tend to use the default
port and base url (which comes without the scheme part).


Thanks a lot.

Issue #9417, opened 2022-01-23
Panic when reading in invalid config.toml of module


deining:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.92.1-85E2E862+extended windows/amd64 BuildDate=2022-01-27T11:44:41Z VendorInfo=gohugoio
</pre>

### Does this issue reproduce with the latest release?
Yes.

### How to reproduce

At your command prompt, issue the following commands:

```bash
hugo new site github-issue-9417
cd github-issue-9417/
echo -n "[params" >> config.toml
hugo serve
```

Now, hugo panics:

```
$ hugo serve
hugo: collected modules in 1335 ms
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/pelletier/go-toml/v2.expect(...)
        /go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.0-beta.3.0.20210727221244-fa0796069526/parser.go:1033
github.com/pelletier/go-toml/v2.(*parser).parseStdTable(0xc00068a230, {0xc000532000, 0x2, 0x202})
        /go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.0-beta.3.0.20210727221244-fa0796069526/parser.go:198 +0x34a
github.com/pelletier/go-toml/v2.(*parser).parseTable(0x2030001, {0xc000532000, 0x2030001, 0x30001})
        /go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.0-beta.3.0.20210727221244-fa0796069526/parser.go:146 +0x39
github.com/pelletier/go-toml/v2.(*parser).parseExpression(0x2de37b0, {0xc000532000, 0x235f05a4511, 0xb5d7e0})
        /go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.0-beta.3.0.20210727221244-fa0796069526/parser.go:120 +0x9f
github.com/pelletier/go-toml/v2.(*parser).NextExpression(0xc00068a230)
        /go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.0-beta.3.0.20210727221244-fa0796069526/parser.go:61 +0x119
github.com/pelletier/go-toml/v2.(*decoder).nextExpr(...)
        /go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.0-beta.3.0.20210727221244-fa0796069526/unmarshaler.go:143
[...]

PR #9435, opened 2022-01-27
9417/Fix panic when reading invalid config file

patche-v:
Fixed [#9417 ](https://github.com/gohugoio/hugo/issues/9417)


bep:
I don't see how this patch can fix that issue. The bug is upstream and needs to be fixed there.

PR #9441, opened 2022-01-28
Improve readiness waiting in server tests

### LINKED ISSUE #9421: Run "hugo server" deterministically
There is currently [one integration test](https://github.com/gohugoio/hugo/blob/b4f27ef8e75a0a6f6ce693270f7ff3114ef4f59a/commands/server_test.go#L31) that makes assertions against a running Hugo server started via the `hugo server` (the other tests in `server_test.go` do not run `cmd.ExecuteC()`). This test waits for the server to be ready [via `time.Sleep`](https://github.com/gohugoio/hugo/blob/b4f27ef8e75a0a6f6ce693270f7ff3114ef4f59a/commands/server_test.go#L61-L64). 

As mentioned in [this comment](https://github.com/gohugoio/hugo/blob/b4f27ef8e75a0a6f6ce693270f7ff3114ef4f59a/commands/server_test.go#L61-L63), let's find a way to make awaiting server readiness more deterministic. There are a number of [outstanding server-related issues](https://github.com/gohugoio/hugo/issues?q=is%3Aissue+is%3Aopen+server+in%3Atitle+), so this will allow us to write more integration tests for the `server` command without unnecessary sleeping and with more reproducible results between compute environments (e.g., CI pipeline containers).

ptgott:
commands.TestServer waits for the Hugo server to be ready by calling:

```go
time.Sleep(2 * time.Second)

```

While this approach prevents us from needing to modify the server
command for tests, it also means that any new hugo server test would
add another two seconds to the test suite. The server is also not
guaranteed to be ready after two seconds.

commands.TestServer is currently the only test that runs the hugo
server command and waits for the server to be ready, but we will
eventually need more of these tests in order to address bugs when
running the hugo server command.

This change adds a "ready" channel to commands.serverCmd. The server
command sends to the channel when the server is ready for connections.
Tests can then block on receiving from the channel instead of sleeping
for a fixed interval.

This speeds up TestServer. Here are the results of running the test
with a clean test cache:

Sleeping two seconds:

```
$ go test github.com/gohugoio/hugo/commands -run TestServer
ok      github.com/gohugoio/hugo/commands       2.618s
```

Using a ready channel:

```
$ go test github.com/gohugoio/hugo/commands -run TestServer
ok      github.com/gohugoio/hugo/commands       0.556s
```

Fixes #9421

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -48,6 +48,9 @@ type serverCmd struct {
 	// Can be used to stop the server. Useful in tests
 	stop <-chan bool
 
+	// Sent to when the server is ready for connections. Used for testing
+	ready chan struct{}
+
 	disableLiveReload bool
 	navigateToChanged bool
 	renderToDisk      bool
@@ -68,8 +71,17 @@ func (b *commandsBuilder) newServerCmd() *serverCmd {
 	return b.newServerCmdSignaled(nil)
 }
 
+// newServerCmdSignaled takes a channel used to stop the server and returns
+// data required to run the hugo server command. Also returns a channel that
+// the command sends to when the server is ready.
 func (b *commandsBuilder) newServerCmdSignaled(stop <-chan bool) *serverCmd {
-	cc := &serverCmd{stop: stop}
+	cc := &serverCmd{
+		stop: stop,
+		// Buffered channel so the server can keep running without blocking
+		// on the receiver. We only expect a single receiver, the caller of
+		// newServerCmdSignaled.
+		ready: make(chan struct{}, 1),
+	}
 
 	cc.baseBuilderCmd = b.newBuilderCmd(&cobra.Command{
 		Use:     "server",
@@ -530,6 +542,8 @@ func (c *commandeer) serve(s *serverCmd) error {
 	}
 
 	jww.FEEDBACK.Println("Press Ctrl+C to stop")
+
+	s.ready <- struct{}{}
 
 	if s.stop != nil {
 		select {
--- a/commands/server.go
+++ b/commands/server.go
@@ -46,7 +46,7 @@ import (
 
 type serverCmd struct {
 	// Can be used to stop the server. Useful in tests
-	stop <-chan bool
+	stop <-chan struct{}
 
 	// Sent to when the server is ready for connections. Used for testing
 	ready chan struct{}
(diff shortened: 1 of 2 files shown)


commands/server.go
```diff
@@ -48,6 +48,9 @@ type serverCmd struct {
 	// Can be used to stop the server. Useful in tests
 	stop <-chan bool
 
+	// Sent to when the server is ready for connections. Used for testing
+	ready chan struct{}
```
bep, line 52: I have no strong opinion about what to use, but the type should be the same for both of `stop` and `ready`.

commands/server_test.go
```diff
@@ -58,10 +58,14 @@ func TestServer(t *testing.T) {
...
-	// There is no way to know exactly when the server is ready for connections.
-	// We could improve by something like https://golang.org/pkg/net/http/httptest/#Server
-	// But for now, let us sleep and pray!
-	time.Sleep(2 * time.Second)
+	// Wait for the server to become ready or fail on timeout
+	k := time.NewTimer(5 * time.Second)
+	select {
+	case <-k.C:
+		t.Error("hugo server took too long to respond")
+	case <-scmd.ready:
```
bep, line 66: OK, I have read through the code;  I don't see how this detects _when the server is ready for connections_?

ptgott: When the `hugo server` command is executed, it blocks at [this `if` statement](https://github.com/gohugoio/hugo/blob/0303c29b63c8a4b8a1b515251c1b618196a1174a/commands/server.go#L548-L555) until it can receive from either the `stop` channel or `sigs`. In this PR, the server [sends to `ready`](https://github.com/gohugoio/hugo/blob/0303c29b63c8a4b8a1b515251c1b618196a1174a/commands/server.go#L546) just before the `serve` function blocks. I put the `s.ready <- struct{}{}` statement just before the command receives from the `sigs`/`stop` channels to ensure that all setup work completed by `serve` had finished.

bep: * The servers (one or more) are started in their own go routine.
* The printing of "Press Ctrl+C to stop" may come before the server(s) are started.
* I don't know of any way to detect "servers are ready to receive connections", and if it was, I don't see how this PR hooks into the `http` package.

ptgott: It was a misjudgment on my part to conflate the `hugo server` command finishing its setup work with the HTTP servers being ready for requests. 

One solution is to add a lightweight `/ready` endpoint to each HTTP server, probably in `*fileServer.createEndpoint`. This would simply return `200` to any GET request. We could then use a test helper function to repeatedly—and synchronously—poll the readiness endpoint of each HTTP server on a tight loop (say, every few milliseconds). The helper would return `nil` on success (and the test that calls it would continue) or cause the test to fail after a timeout. 

While this would still require sleeping between retries, it would be more deterministic than using a single `time.Sleep` call, and would probably save time. I would just need to figure out how to cleanly pass the base URL of each server to the test helper.

What do you think?

bep: > it also means that any new hugo server test would add another two seconds to the test suite. 

I suspect the starting point for this issue is a little bit flawed: The above statement isn't true; we can 1) Run server tests in parallel (each server gets its own port) and 2) Run many test case for each server start. All of Hugo's tests have been running ... thousands of times, and I don't remember one time the server tests have failed (at least not in recent times).

This should be supported in the Go stdlib, adding to the complexity here isn't worth it.

PR #9447, opened 2022-01-31
Remove the unnecessary type attribute for Google Analytics

squalrus:
image

Example: https://validator.w3.org/nu/?doc=https%3A%2F%2Fsqualr.us%2F


bep:
I think we have had this discussion before. The type attribute may be unnecessary for some, but not if you somehow need it to determine the .. type. Which I think our minfier does.

squalrus:
> I think we have had this discussion before. The type attribute may be unnecessary for some, but not if you somehow need it to determine the .. type. Which I think our minfier does.

Makes sense, thanks @bep! Closing.

PR #9464, opened 2022-02-04
Exclude event attributes when rendering markdown

### LINKED ISSUE #9463: Exclude event attributes when rendering markdown
### Current Behavior

config.toml

```toml
[markup.goldmark.renderer]
unsafe = false

[markup.goldmark.parser.attribute]
block = true
title = true
```

markdown

````md
## Heading {onclick="alert('heading')"}

> Blockquote
{ondblclick="alert('blockquote')"}

```bash {onmouseover="alert('code fence')"}
codefences
```
````

renders

```html
<h2 onclick="alert('heading')" id="heading">Heading</h2>

<blockquote><p>Blockquote</p></blockquote>

<div class="highlight" onmouseover="alert('code fence')">...
```

### Desired Behavior

Remove attribute events from headings and code fences. The block attributes are already being sanitized.

```html
<h2 id="heading">Heading</h2>

<blockquote><p>Blockquote</p></blockquote>

<div class="highlight">...
```

Reference:
- <https://discourse.gohugo.io/t/35150/6>
- <https://discourse.gohugo.io/t/35069/4>

Testing:

```bash
git clone --single-branch -b hugo-github-issue-9463 https://github.com/jmooring/hugo-testing hugo-github-issue-9463
cd hugo-github-issue-9463
hugo server
```

jmooring:
Closes #9463


jmooring:
Note: I intentionally ignore `markup.goldmark.renderer.unsafe` when stripping event attributes. I could go either way on this, but I figured it would be better to be aggressive, and relax later if necessary.

bep:
This looks good. I would, however, appreciate if you could move the test to the Goldmark -- create a integration_test.go file, see some of my recent commits.

You can convert the data in your existing tests by `fmt.Println(b.DumpTxtar())`.

jmooring:
I've moved and changed the test to use NewIntegrationTestBuilder. This format is easier to write and read. Thanks.

markup/goldmark/integration_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+	"github.com/gohugoio/hugo/hugolib"
+)
+
+func TestAttributeExclusion(t *testing.T) {
+	c := qt.New(t)
```
bep, line 24: * Add `t.Parallel()` to make it run potentially faster ...
* You can drop this line and just pass t into T (yes, I know you probably followed my example)

markup/goldmark/integration_test.go
```diff
@@ -0,0 +1,77 @@
...
+	b := hugolib.NewIntegrationTestBuilder(
+		hugolib.IntegrationTestConfig{
+			T:           c,
+			TxtarString: files,
+			NeedsOsFS:   false,
+		},
+	).Build()
+
+	b.AssertFileContent("public/p1/index.html", `
+<h2 class="a" id="heading">Heading</h2>
```
bep, line 60: The lines gets split by `\n` here and tested one by one, so I would suggest that you make the most narrow assertion that tests the actual issue -- it's a pain to adjust 100 similar tests whenever we do need to do a change in HTML rendering.

markup/goldmark/integration_test.go
```diff
@@ -0,0 +1,77 @@
...
+---
+## Heading {class="a" onclick="alert('heading')" linenos="inline"}
+
+> Blockquote
+{class="b" ondblclick="alert('blockquote')" LINENOS="inline"}
+
+~~~bash {id="c" onmouseover="alert('code fence')" linenos=true linenostart=2 hl_lines=[3] hl_style=dracula}
+javascript
+security
+issue
```
bep, line 45: People understand what this issue is about without putting it on a big bold sign :-)

bep (changes requested):

markup/goldmark/integration_test.go
jmooring: Done.

markup/goldmark/integration_test.go
jmooring: Done.

markup/goldmark/integration_test.go
jmooring: Done.

PR #9507, opened 2022-02-16
Sort pages by title length

nobe4:
Opening this PR because I couldn't find a proper way to achieve this with built-in sorters.

This add the required code, test and documentation.


bep:
OK, I did not read this ...

This sounds like a very specialised use case; I suggest you open up a regular issue (proposal) to discuss it.

PR #9582, opened 2022-02-28
docs/chore: use https for some links

imba-tjd:
Trivial


bep:

Also, the documentation below /commands is autogenerated from source.

Issue #9609, opened 2022-03-04
readDir broken in 0.93.2 when directories are symlinks/Linux AR>M64


gj52:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version 0.93.2
works with 0.93.1 !
</pre>

### Does this issue reproduce with the latest release?
YES

---

Dev on Windows is fine.
Prod on Linux breaks, can not find file or directory

path is assets/etc/apache2   with some copied conf files from /etc
should find apache2.conf 

index.md
```
+++
date        = "2018-11-12T11:50:21+01:00"
description = "Konfiguration Apache2"
title       = "Apache2"
draft       = false
chapter     = 1002
+++

## Apache
{{< linux/shortlist path="apache2" type="conf" >}}

### conf-enabled
{{< linux/shortlist path="apache2/conf-enabled" type="conf">}}

### mods-enabled
{{< linux/shortlist path="apache2/mods-enabled" type="conf">}}

### sites-enabled
{{< linux/shortlist path="apache2/sites-enabled" type="conf">}}

## Let's Encrypt Apache 
{{< linux/shortlist path="letsencrypt/" type="conf">}}

```

shortcode is

```
{{ $path     := .Get "path" }}
{{ $filepath := printf "/assets/etc/%s" $path }}
{{ $files    := readDir $filepath }}
{{ $ext := default "-" (.Get "type" )}}
{{ if (fileExists $filepath)}}
<div class="pa2 mh4 ba">
{{- range $files }}
	{{ if (and (ne (substr .Name 0 1) "_") (not .IsDir)) }}{{ $f := resources.Get (printf "etc/%s/%s" $path .Name)}}
		{{ $e := findRE "\\w+$" .Name }}
		{{ if (or ( eq $ext "-") (in $e $ext) )}}
			<a class="w-auto pr2" href={{$f.Permalink }} type=text/plain rel=nofollow download>{{ .Name }}</a>
		{{end}}
	{{end}}
{{- end }}
</div>
{{end}}

```

bep:
@gj52 these files, do they live below the `theme` directory?

gj52:
The shortcode is in the theme directory.  Theme extract is [here](https://github.com/gj52/HugoSample).
The listed files are in <project dir>/<asset dir>/etc/

If you need to look in my complete activ repo, send me a PM.

bep:
@gj52 what I need is something that ... fails, does the "theme extract" fail for you?

[...]

PR #9613, opened 2022-03-05
tpl/os: Add filesystem option to ReadDir

### LINKED ISSUE #9604: Make readDir/readFile more precise and flexible
See #9599 

There are historical reasons to this, I'm sure, but the current behaviour is hard to understand, document and is not as flexible as it easily could be.

So, I suggest adding a `filesystem` option to both of these functions:

* work -- default, will search in _the project work dir only_
* content (composite, project first)
* archetypes  (composite, project first)
* static  (composite, project first)
* layouts  (composite, project first)
* data  (composite, project first)
* assets  (composite, project first)
* i18n  (composite, project first)

The new default (that `work` is project work dir only) is a slightly breaking change, but I suspect it's going to solve more issues than it creates.

lukeryannetnz:
This is my first contribution to this project, the issue was labelled as `[GoodFirstIssue]` so I jumped in. As such, any feedback or guidance :100: appreciated.

Make ReadDir function variadic. Optional second param specifies FileSystem. The default if not specified remains work dir, to retain backwards compatibility.

I largely followed the breadcrumbs @bep left in the issue comments (thanks!).  Things I'm not sure about:
- how best to support the static FS given when in multihost we have one static filesystem per language
- if I need to better test my input checks in `os`
- how best to update docs (I can spend time looking at this if I get positive feedback on this change)

Fixes #9604

### DIFF
--- a/docs/content/en/functions/readdir.md
+++ b/docs/content/en/functions/readdir.md
@@ -8,15 +8,34 @@ menu:
   docs:
     parent: "functions"
 keywords: [files]
-signature: ["os.ReadDir PATH", "readDir PATH"]
+signature: ["os.ReadDir PATH", "readDir PATH", "readDir PATH DIRECTORY"]
 workson: []
 hugoversion:
 relatedfuncs: ['os.FileExists','os.ReadFile','os.Stat']
 deprecated: false
 aliases: []
 ---
-The `os.ReadDir` function resolves the path relative to the root of your project directory. A leading path separator (`/`) is optional.
+The `os.ReadDir` function resolves the path relative to the root of your project directory unless the optional directory parameter is provided. A leading path separator (`/`) is optional.
 
+The following DIRECTORY values are supported
+* assets
+* archetypes
+* data
+* content
+* layouts
+* i18n
+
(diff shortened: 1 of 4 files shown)


bep:
Thanks for this -- the code looks great at first glance. I will need to understand and fix #9609 before I can merge this.

But if I could ask for one change that would make this easier to review:

Instead of creating one test function per variant, I suggest you combine them into one and separate the expectations by an identifier, e.g.:

```
START  assets:|asset.jpg|:END:
START static:|static.jpg|:END:
...
```

And if you create an inline template to print this, you get a pretty compact test.

lukeryannetnz:
Thanks for the feedback, I've combined the tests so they are easier to read as suggested.
Also added a todo to make the gap in the functionality explicit.

I'm a little confused about the failing tests, looks like they timed out. If you can approve the workflow, I'll look into them if it happens again.

lukeryannetnz:
Updated test to include a check for composite archetype files.

lukeryannetnz:
Had a go at adding docs for the new functionality.

lukeryannetnz:
@bep any interest in taking another look at this?

Issue #6226, opened 2019-08-15
snap denials /etc/gitconfig


julian-klode:
Running the hugo snap fails with

```
Building sites … ERROR 2019/08/15 19:04:04 Failed to read Git log: fatal: unknown error occured while reading the configuration files
Total in 125 ms
Error: Error building site: logged 1 error(s)

```

in dmesg I see:

```
[140780.553009] audit: type=1400 audit(1565888471.869:1669): apparmor="DENIED" operation="open" profile="snap.hugo.hugo" name="/etc/gitconfig" pid=22857 comm="git" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0                                                                    
[140780.553095] audit: type=1400 audit(1565888471.869:1670): apparmor="DENIED" operation="open" profile="snap.hugo.hugo" name="/etc/gitconfig" pid=22857 comm="git" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0     
```

julian-klode:
You could probably simply set GIT_CONFIG_NOSYSTEM=true in the app launcher in snapcraft.yaml

gerrywastaken:
Indeed running the server like so causes the issue to disappear:
`GIT_CONFIG_NOSYSTEM=true hugo serve -D`

Thanks @julian-klode !

jmooring:
Reproducible with snap install of v0.80.0.

```
sudo git config --system foo.bar "baz"
sudo chmod 644 /etc/gitconfig
git clone --single-branch -b hugo-github-issue-6226 https://github.com/jmooring/hugo-testing hugo-github-issue-6226
cd hugo-github-issue-6226
hugo server
```

anthonyfok:
This was fixed in
* #9619 thanks to @ifurther 

Once https://forum.snapcraft.io/t/system-files-and-personal-files-request-for-hugo/30598 is approved, the Snap store will allow newer versions of Hugo that can read gitconfig to enter the Snap Store, and then this issue can be closed.

Sorry for my 12-week delay in actually filing that store request.

PR #9619, opened 2022-03-07
Snap: Add {system,user}-gitconfig plugs to read gitconfig

ifurther:
Reference [this code ](https://forum.snapcraft.io/t/system-files-and-personal-files-request-for-onefetch/28963) to add plug, and maybe could resolve issue #6226.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -30,6 +30,8 @@ plugs:
     
 apps:
   hugo:
+    environment:
+      HOME: $SNAP_REAL_HOME
     command: bin/hugo
     completer: hugo-completion
     plugs:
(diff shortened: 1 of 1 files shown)


anthonyfok:
@ifurther Thank you for your great work!  This is totally new to me, but it does look like it, together with a corresponding request to the Snapcraft Team, would resolve issue #6226.

I was studying https://github.com/o2sh/onefetch/issues/588, and looks like there is another commit that maps `$HOME` to `$SNAP_REAL_HOME` for a more complete fix, see https://github.com/o2sh/onefetch/commit/923e486a52a8c7441118a3e1ebeb82380fbf2c86

Could you please look into that and build upon this PR?  Many thanks!

PR #9631, opened 2022-03-08
markup/highlight: Make lineNumbersInTable = false default

### LINKED ISSUE #9596: Change default value markup.highlight.lineNumbersInTable to false
With v0.92.2 and earlier, with `lineNos=true`, the only way to produce a selectable block of code was to set `lineNumbersInTable=true`, so it was a sensible default.

With v0.93.0 and later, `lineNos=true` plus `lineNumbersInTable=false` produces a selectable block of code. It is no longer necessary to set `lineNumbersInTable=true`. This new and desirable behavior was introduced with https://github.com/gohugoio/hugo/commit/ea54a99ca5c4ed9e66574553f6c057fb83d0aca3.

Is the current default causing any problems?
- In the absence of any external styling, no. However, the additional markup is not required.
- With external styling, sometimes.

I ran into this with a Bootstrap 5 implementation, where `_reboot.scss` was setting:

```css
pre,
code,
kbd,
samp {
  font-family: $font-family-code;
  @include font-size(1em); // Correct the odd `em` font sizing in all browsers.
  direction: ltr #{"/* rtl:ignore */"};
  unicode-bidi: bidi-override;
}
```

Producing this:

screen capture

The bidirectional rules have been [removed](https://github.com/twbs/bootstrap/commit/989de20baed8a578fe825649c5b819238887f83c) in HEAD, but anyone using a production release will run into this problem.

Example:

```bash
git clone --single-branch -b hugo-github-issue-9596 https://github.com/jmooring/hugo-testing hugo-github-issue-9596
cd hugo-github-issue-9596
hugo server
```

Follow up....

Even when using Bootstrap 5 at HEAD, setting `lineNumbersInTable=true` still produces undesirable results (space after final line of code):

another screen capture

jmooring:
Closes #9596

### DIFF
--- a/docs/content/en/functions/highlight.md
+++ b/docs/content/en/functions/highlight.md
@@ -35,8 +35,8 @@ lineNos
 Display a number at the beginning of each line.
 
 lineNumbersInTable
-: Boolean. Default is `true`.\
-Render the highlighted code in an HTML table with two cells. The left table cell contains the line numbers. The right table cell contains the code, allowing a user to select and copy the code without line numbers. Irrelevant if `lineNos` is false.
+: Boolean. Default is `false`.\
+Render the highlighted code in an HTML table with two cells. The left table cell contains the line numbers. The right table cell contains the code. Irrelevant if `lineNos` is false.
 
 anchorLineNos
 : Boolean. Default is `false`.\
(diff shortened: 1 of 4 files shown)


kaushalmodi:
Please hold this merge until https://github.com/gohugoio/hugo/issues/9596#issuecomment-1062487823 is resolved.

jmooring:
Closing. See comments on https://github.com/gohugoio/hugo/issues/9596. Firefox...

PR #9664, opened 2022-03-13
Enable hugo server SIGINT after loading bad config

### LINKED ISSUE #8340: `hugo server` doesn't stop on Ctrl+C if theme name is misspelled after starting the server
Hello!

`hugo server` says "Press Ctrl+C to stop" but the server doesn't stop if theme name is misspelled after starting the server. The only way to exit the `hugo server` process then is to close the terminal forcefully.

<pre>
$ hugo version
hugo v0.81.0-59D15C97+extended linux/amd64 BuildDate=2021-02-19T17:07:12Z VendorInfo=gohugoio
</pre>

### Steps to reproduce:
1. Start `hugo server` with correct `theme: <theme_name>` in config
2. Now change `<theme_name>` to some non-existent theme name (e.g. misspelling)
3. Now try to stop server with `Ctrl+C`

ptgott:
If a user starts hugo server and writes to the config file in a way
that renders it invalid, hugo server stops responding to SIGINTs,
meaning that a user needs to send a SIGKILL to the hugo process to stop
it.

*commandeer.serve needs to obtain a *hugolib.HugoSites in order to
close the Hugo environment. After handling signals/the stop channel,
*commandeer.serve calls c.hugo() to get the *hugolib.HugoSites. This
function receives from the c.created channel, which is closed when the
HugoSites is ready. However, if an error took place while loading the
config, c.created would never be closed, so Hugo would block on this
channel even after handling SIGINT.

The solution is to close c.created if we failed to load a config when
rebuilding the site, then check at the end of *commandeer.serve whether
the *hugolib.HugoSites is nil. If it is, we stop waiting for a HugoSites
to close, and exit the hugo process.

One issue that resulted from this fix was how to reset the c.created
channel during site rebuilds. In the current code, this is done by
assigning c.commandeerHugoState to a newCommandeerHugoState in
*commandeer.fullRebuild. However, this creates a race condition, since
other goroutines can be reading the value of c.created just as
*commandeer.fullRebuild is attempting to reassign it.

This change fixes the race condition by adding the lazy.Notifier type,
which preserves the use of c.created while allowing for it to be reset
in a goroutine-safe way. c.created is now a lazy.Notifier. I added this
type to the lazy package because it didn't seem worth it to add a new
package, but we can do this if it's cleaner. (I was thinking about using
sync.Cond for this, but the approach I used is closer to the original
use of the c.created channel.)

Also adds an integration test for interrupting the server after fixing
an invalid configuration. An earlier version of this change caused hugo
server to stop exiting with a SIGINT after fixing a bad config, so I
wanted to ensure that we can prevent this.

Fixes #8340

### DIFF
--- a/commands/commandeer.go
+++ b/commands/commandeer.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	hconfig "github.com/gohugoio/hugo/config"
+	"github.com/gohugoio/hugo/lazy"
 
 	"golang.org/x/sync/semaphore"
 
@@ -58,7 +59,7 @@ type commandeerHugoState struct {
 	*deps.DepsCfg
 	hugoSites *hugolib.HugoSites
 	fsCreate  sync.Once
-	created   chan struct{}
+	created   *lazy.Notifier
 }
 
 type commandeer struct {
@@ -121,12 +122,12 @@ type serverPortListener struct {
 
 func newCommandeerHugoState() *commandeerHugoState {
 	return &commandeerHugoState{
-		created: make(chan struct{}),
+		created: lazy.NewNotifier(),
 	}
 }
 
 func (c *commandeerHugoState) hugo() *hugolib.HugoSites {
-	<-c.created
+	c.created.Wait()
 	return c.hugoSites
 }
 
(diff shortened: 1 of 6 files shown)


bep:
Thanks for this; I'm sure what you have implemented works, but looking at your description I think we need to take a step back and solve this by reducing some of the complexity (which I'm certainly to blame for, but I got some of it from the starting point which was lots of globals) not add more complexity.

I'm not sure what that would be, but I'll think about it. I'll leave this PR open for if I can not come up with something simpler.

ptgott:
👍 I'll keep this rebased in the meantime just in case.

PR #9682, opened 2022-03-17
Fixing Google Analytics Tracking

hugovidafe:


davidsneighbour:
What does this "fix"? It tracks when donottrack is set and if it is not set it does not track. It literally breaks the functionality.

hugovidafe:
I "true" want to don't be tracked, the script will be executed.
I "false" want to don't be tracked, script won't be executed.

hugovidafe:
We're being anonymized when donottrack is true, or do I have the wrong point of view?

davidsneighbour:
I think you are wrong. The code you changed is "is donottrack disabled? then do track". The code that section creates is the initial adding of the tracking script, not about the browser feature do not track. Its basically a security setting so site owners can override the themes setup. It's the question if the site config disables google analytics. 

Have a read through this section:

https://gohugo.io/about/security-model/#security-policy

PR #9691, opened 2022-03-18
Fix trailing comma in keyword meta tag for schema.html

alexanderhansen:
This aimed to fix the trailing comma in the keywords meta tag for schema.html - there seemed to be more opportunities for improvement and so not only tags but all taxonomies are now included.

### DIFF
--- a/hugolib/embedded_templates_test.go
+++ b/hugolib/embedded_templates_test.go
@@ -25,7 +25,8 @@ baseURL = "https://example.org"
 
 [params]
 images=["siteimg1.jpg", "siteimg2.jpg"]
-
+[taxonomies]
+series="series"
 `
 	b := newTestSitesBuilder(t).WithConfigFile("toml", config)
 
@@ -33,6 +34,7 @@ images=["siteimg1.jpg", "siteimg2.jpg"]
 title: My Bundle
 date: 2021-02-26T18:02:00-01:00
 lastmod: 2021-05-22T19:25:00-01:00
+tags: ["test1", "test2"]
 ---
 `)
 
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this. Quickly reading the change; 1) I don't think your template is logically the same as before (if we look beyond the comma), but 2) I'm not sure the old logic makes sense, either.

Here is the gist of what we have today:

```
<meta itemprop="keywords" content="{{ if .IsPage}}{{ range $index, $tag := .Params.tags }}{{ $tag }},{{ end }}{{ else }}{{ range $plural, $terms := .Site.Taxonomies }}{{ range $term, $val := $terms }}{{ printf "%s," $term }}{{ end }}{{ end }}{{ end }}" />
```

So,

* For regular pages, we list the "tags" (which is rather random)
* For all the other node type of pages (e.g. home page, sections), we list all terms

Hugo has admittingly evolved since that template was written, and I think we need to do better, but I'm not sure ...

@jmooring ?

alexanderhansen:
Thanks for considering the PR. After taking another look now, I think the `.IsPage` condition in the meta tag doesn't make sense at all (now and before), because, except the first two lines, everything in the template is only rendered for regular pages since it's wrapped in one `{{- if .IsPage -}}` block. I tried to reflect my understanding in the test case but I'm just getting started with Hugo...

alexanderhansen:
I pushed an update, that includes all taxonomy for the page, for this the test defines a custom taxonomy on the test page and tests for the inclusion. So all taxonomy for the given page ends up in the meta tag (without the trailing comma ;-) )

For all others pages I'm not sure - so for the homepage site-wide taxonomy should be included, for category pages the aggregated taxonomy of all pages for that category? Or maybe only the one keyword for the given category?

PR #9695, opened 2022-03-20
Fixes #9127: Support '-u=patch' in hugo mod get

### LINKED ISSUE #9127: Support '-u=patch' in hugo mod get
The go module system supports updating all modules to the latest patch version with the following command:

`go get -u=patch`

The equivalent hugo command (`hugo mod get -u=patch`) doesn't work and outputs the following message on my hugo project:

`go get: no package in current directory`

A quick look at the code shows that there is special handling for `-u` with no additional arguments here:  https://github.com/gohugoio/hugo/blob/82c33c7105a4d0c0e097f0d074c02995cb2b4d20/modules/client.go#L307

It seems (and I'm no expert) that this code could be updated fairly easily to also support `-u=patch`.

Thanks.

LukeDeWaal:
I believe I have implemented the required fix #9127. 

I changed the aforementioned if-statement to check whether the supplied argument contains `-u`, as opposed to whether the supplied argument is identical to `-u`.  A flag called `patch`, similar to the `update` flag, was added. The `patch` flag will be true if the supplied argument is exactly  `-u=patch`. The appropriate command is then passed down.

I am unsure of how to add tests for this fix, so any pointers on how to get started writing tests would be highly appreciated @bep .

### DIFF
--- a/modules/client.go
+++ b/modules/client.go
@@ -305,8 +305,9 @@ func (c *Client) Vendor() error {
 
 // Get runs "go get" with the supplied arguments.
 func (c *Client) Get(args ...string) error {
-	if len(args) == 0 || (len(args) == 1 && args[0] == "-u") {
+	if len(args) == 0 || (len(args) == 1 && strings.Contains(args[0], "-u")) {
 		update := len(args) != 0
+		patch := update && (args[0] == "-u=patch") //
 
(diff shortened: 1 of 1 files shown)


bep:
This looks good to me, adding test for this is currently a little too hard to be worth it, I'll merge this once it goes green.

PR #9710, opened 2022-03-23
tpl/crypto: Add optional encoding arg to hmac function

### LINKED ISSUE #9709: Add optional encoding parameter to hmac template function
The primary driver for this enhancement is to provide binary output from `crypto.HMAC`. See:
<https://discourse.gohugo.io/t/have-hmac-outputs-raw-binary-data-rather-than-hexits-lowercase/37808/>

cc @regis

jmooring:
Closes #9709

### DIFF
--- a/docs/content/en/functions/hmac.md
+++ b/docs/content/en/functions/hmac.md
@@ -1,7 +1,7 @@
 ---
 title: hmac
 linktitle: hmac
-description: Compute the cryptographic checksum of a message.
+description: Returns a cryptographic hash that uses a key to sign a message.
 date: 2020-05-29
 publishdate: 2020-05-29
 lastmod: 2020-05-29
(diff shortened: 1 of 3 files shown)


docs/content/en/functions/hmac.md
```diff
@@ -10,24 +10,25 @@ menu:
...
 ---
 
-`hmac` returns a cryptographic hash that uses a key to sign a message.
+Set the `HASH_TYPE` argument to `md5`, `sha1`, `sha256`, or `sha512`.
 
-```
-{{ hmac "sha256" "Secret key" "Hello world, gophers!"}},
-<!-- returns the string "b6d11b6c53830b9d87036272ca9fe9d19306b8f9d8aa07b15da27d89e6e34f40"
-```
+Set the optional `ENCODING` argument to either `hex` (default) or `binary`.
```
carlmjohnson, line 23: You could save a step and have encoding BASE64 as one of the options here.

PR #9715, opened 2022-03-24
Extend PageMatcher with Environment keyword

### LINKED ISSUE #9612: Add environment (e.g. production) as a new filter to _cascade.target
Hey there,

After looking this up and [confirming](https://discourse.gohugo.io/t/skipping-page-building-depending-on-environment-config/37502) this was not possible (thanks again @jmooring!) I would like to propose a feature, that works similar to the `draft` keyword, but allows specifying the environment(s) for which pages/posts shall be built for. Specifically, I suggest adding an `environment` keyword, which would accept a list of values:

```yaml
environment:
  - production
  - staging
  - development
```

This would allow for per-page/-post management of individual pages/posts, in which individual teams work on parts of a large Hugo deployment, in which none of them (but one admin-team) have access to the actual environment configuration files. Through a keyword inside the pages/posts, teams could individually decide to which environments pages/posts would get deployed to, without having the admin-team involved for adjusting the `cascade._target` configuration.

I would appreciate thoughts on this! Thank you.

CathrinePaulsen:
Fixes #9612 

This fix allows `environment` to be used as a keyword for `cascade._target`. 
It can be used to apply cascade options to the target in only certain build environments, for example to specify in which environment(s) certain pages should not be rendered.

The documentation has been updated with the new keyword.

### DIFF
--- a/hugolib/page__meta.go
+++ b/hugolib/page__meta.go
@@ -167,6 +167,10 @@ func (p *pageMeta) Description() string {
 	return p.description
 }
 
+func (p *pageMeta) Environment() string {
+	return p.s.Environment()
+}
+
 func (p *pageMeta) Lang() string {
 	return p.s.Lang()
 }
--- a/hugolib/page__meta.go
+++ b/hugolib/page__meta.go
@@ -167,10 +167,6 @@ func (p *pageMeta) Description() string {
 	return p.description
 }
 
-func (p *pageMeta) Environment() string {
-	return p.s.Environment()
-}
-
 func (p *pageMeta) Lang() string {
 	return p.s.Lang()
 }
--- a/docs/content/en/content-management/front-matter.md
+++ b/docs/content/en/content-management/front-matter.md
@@ -189,6 +189,9 @@ kind
 lang
 : A Glob pattern matching the Page's language, e.g. "{en,sv}".
 
+environment
+: A Glob pattern matching the build environment, e.g. "{production,development}"
+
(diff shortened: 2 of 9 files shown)


hugolib/page__meta.go
```diff
@@ -167,6 +167,10 @@ func (p *pageMeta) Description() string {
 	return p.description
 }
 
+func (p *pageMeta) Environment() string {
```
bep, line 170: 0K, this PR looks mostly good, but we don't want to add the `Environment()` everywhere -- the Page interface is big as it is.

So I suggest that you in the page matcher do something ala:

```go
if m.Environment != "" {
		g, err := glob.GetGlob(m.Environment)
		if err == nil && !g.Match(p.Site().Hugo().Environment()) {
			return false
		}
	}

```

bep (changes requested):

hugolib/page__meta.go
bep: Yea, I see your point. Ultimately we may want to split it (e.g. having both a MatchesSite and a Matches Page method), but even then you will have the same test challenge. 

There is a `testSite` implementation in there somewhere, so I suggest you just expand on that. It makes for a little more work, but certainly a better end game.

CathrinePaulsen: I think the most recent change should address the requested change. Let me know what you think!

bep:
Yea, now it looks good, thanks!

PR #9724, opened 2022-03-26
Create JSON schema for Hugo

EmilySeville7cfg:
This JSON schema is taken from [this](https://github.com/SchemaStore/schemastore/pull/2138) PR (I am moving it from `schemastore` to this repo).

image
image


bep:
Please create a proposal for this, we're not adding this without a consensus and a plan on how it was built and how to maintain it.

Issue #9735, opened 2022-03-31
Hugo deploy slow when s3 storing large multipart files


davidejones:
**Note this appears to be a somewhat known issue according to the source code comments. Would love to figure out a direction for improving this** 

### What version of Hugo are you using (`hugo version`)?

I believe this issue exists with any hugo version containing the deploy functionality.
For reference I am using `hugo v0.89.4-AB01BA6E+extended`

### Does this issue reproduce with the latest release?

Yes

### What is Hugo Deploy doing?

Hugo deploy does some MD5 compare of remote files and local files to determine differences and whether to upload.

If you have multipart uploads in s3 this results in them not having an MD5 that can be used for comparison. Hugos current solution is for reliability and is to download the file and generate a MD5 of the contents and then compare like usual.
[Source Code Example of this Here](https://github.com/gohugoio/hugo/blob/db1562e134fc4182cfd063058e9ecd0316a0d906/deploy/deploy.go#L576)

### What is the problem?

If you have an s3 bucket with many large multipart files then the time to deploy can be very slow. This is because its downloading each of these to create MD5s for comparison.

Include/exclude filters may work to a point but isn't ideal

### Solutions

Some suggestions from **vangent** in an similar issue here  https://github.com/google/go-cloud/issues/1898

1. One option is to fallback to Size comparison when no MD5 instead of generating. 
   _This might not be as reliable but 99% of the time should work. perhaps this could be an optional flag?_

2. Consider saving an MD5 hash (or any other hash) explicitly in Metadata. 
   _Perhaps a `Hugo-Hash` attribute is stored on objects and replaces the current obj.MD5 and is used for compare?_

jmooring:
@davidejones Was this issue resolved by #9742? The commit message states "improves" but it is not clear to me what remains to be fixed.

davidejones:
@jmooring Yes this issue is resolved by 9742!

PR #9742, opened 2022-04-03
hugo deploy also sets an md5 attribute & checks it

davidejones:
Improves https://github.com/gohugoio/hugo/issues/9735

### Problem

During Hugo deploy when a remote MD5 is invalid (e.g due to multipart etag) hugo reads the entire remote file and calculates the md5 again which can be slow.

### This PR

- Updates file upload so that it will also store an md5 hash in the cloud providers attributes. e.g in aws this looks like `x-amz-meta-md5chksum: 26fe392386a8123bf8956a16e08cb841`
- Updates the walkremote so that if an MD5 is nonexistent/invalid then it attempts to read the attribute md5 before falling back to reading the entire remote file

### What impact does this have?

Hugo deploy will be faster in the diffing stage between remote and local. 
**_IF_** files come through with `len(obj.MD5) == 0` but store an attribute md5.

### Caveats

Only files uploaded with hugo will have this attribute. 

**Existing users**
Existing hugo deploy users running this in a new release would only get the attributes set on files that are uploaded going forward. If they want all files to have the md5 attribute then running with `--force` should work.

**Other files in the bucket**
If you have files uploaded by another tool in the same bucket that hugo deploy is using and these don't have valid a MD5. Then you can see a perf boost if you manually upload these to have the md5 as an attribute.

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -21,6 +21,7 @@ import (
 	"compress/gzip"
 	"context"
 	"crypto/md5"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -72,6 +73,8 @@ type Deployer struct {
 type deploySummary struct {
 	NumLocal, NumRemote, NumUploads, NumDeletes int
 }
+
+var metaMD5Hash = "md5chksum" // the meta key to store md5hash in
 
 // New constructs a new *Deployer.
 func New(cfg config.Provider, localFs afero.Fs) (*Deployer, error) {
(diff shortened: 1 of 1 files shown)


bep:
Makes sense to me -- do you see any issues with this, @vangent ?

deploy/deploy.go
```diff
@@ -73,6 +74,8 @@ type deploySummary struct {
 	NumLocal, NumRemote, NumUploads, NumDeletes int
 }
 
+var metaMD5Hash = "md5chksum" // the meta key to store md5hash in
```
vangent, line 77: Can make this a const, no?

Also, I don't think you need to ToLower it down below, since it's a const.

deploy/deploy.go
```diff
@@ -566,21 +570,33 @@ func walkRemote(ctx context.Context, bucket *blob.Bucket, include, exclude glob.
...
-			r, err := bucket.NewReader(ctx, obj.Key, nil)
+			var attrMD5 []byte
+			attrs, err := bucket.Attributes(ctx, obj.Key)
 			if err == nil {
-				h := md5.New()
-				if _, err := io.Copy(h, r); err == nil {
-					obj.MD5 = h.Sum(nil)
+				md5String, exists := attrs.Metadata[strings.ToLower(metaMD5Hash)]
+				if exists {
+					attrMD5, err = hex.DecodeString(md5String)
```
vangent, line 586: The `err` here is never used, so it would be more clear to use `_` instead.

vangent (commented):
General approach LGTM.

davidejones:
Thanks for the feedback @vangent, I've followed your suggestions and cleaned up the code a little.

PR #9760, opened 2022-04-08
Changed wording a message

OzLocke:
The message "Start building sites ..." (line 291) reads like a command to the user. Not usually an issue, but when it's taking a long time it looks like it's waiting for the user to do something.

Changed to "Starting to build sites ..." so it's clearer it's just a notification and requires no action.


bep:
Thanks, but we've had it like it is today since the start, I like it, and people may be parsing the output expecting that string,

PR #9823, opened 2022-04-26
hugo: Add basic gemini file type

kallisti5:
* Gemini is a lightweight text-only web protocol which falls
  somewhere between gopher and html.
* This adds native filetype support which the community has
  been hacking into config files for quite some time.
* Hugo is unable to serve gemini (since it isn't http), however the outputed documents
  can be passed to one of the many small + simple gemini server choices out there.
* Based on the examples here:
  https://sylvaindurand.org/gemini-and-hugo/

### DIFF
--- a/media/mediaType.go
+++ b/media/mediaType.go
@@ -233,6 +233,7 @@ var (
 	TypeScriptType = newMediaType("application", "typescript", []string{"ts"})
 	TSXType        = newMediaType("text", "tsx", []string{"tsx"})
 	JSXType        = newMediaType("text", "jsx", []string{"jsx"})
+	GEMINIType     = newMediaType("text", "gemini", []string{"gmi", "gmni", "gemini"})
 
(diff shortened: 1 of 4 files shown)


kallisti5:
Here's an example gemini site in the lagrange browser for context:

image_2022-04-25_190410319

Here's information on the protocol as well:
https://en.wikipedia.org/wiki/Gemini_(protocol)

kallisti5:
Oh, here's a project around building (converting) markdown hugo sites to output gemini sites:
https://github.com/tdemin/gmnhg

bep:
>been hacking into config files for quite some time.

Adding a custom mime type/output format is configuration, not hacking.

Please open a proposal issue about this where we can discuss if we want to add this.

kallisti5:
Thanks @bep   That pretty much sums up everything I need to know about contributing to hugo as a project.

PR #9841, opened 2022-04-28
Use AWS SDK Version 2

frankywahl:
AWS released a V2 of their SDK with a few upgrades: https://aws.amazon.com/blogs/developer/aws-sdk-for-go-version-2-general-availability/

This makes sure Hugo's deploy mechanism uses the V2 of AWS's SDK, and updates dependencies. 

This also fixes (and supersedes) https://github.com/gohugoio/hugo/pull/9837


go.mod
```diff
@@ -82,15 +84,16 @@ require (
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
-	github.com/aws/aws-sdk-go-v2 v1.9.0 // indirect
-	github.com/aws/aws-sdk-go-v2/config v1.7.0 // indirect
+	github.com/aws/aws-sdk-go v1.43.5 // indirect
```
frankywahl, line 87: Unfortunately, this is still needed by https://pkg.go.dev/gocloud.dev@v0.24.0/blob/s3blob

bep: We're not adding both v1 and v2 of the (pretty big) SDK.

frankywahl: V1 and V2 are already present (because of the transient dependencies). We already see both in the go.mod file on the master branch ([v1](https://github.com/gohugoio/hugo/blob/46da0b7aaa231cc5e56ad79eb71c4fc3f7a38a29/go.mod#L8) & [v2](https://github.com/gohugoio/hugo/blob/46da0b7aaa231cc5e56ad79eb71c4fc3f7a38a29/go.mod#L94)). 

This just changes the direct dependency to be V2, and the indirect dependency to be V1. The v2 is meant to be smaller because it only import mdoles  from AWS services we use instead of the whole SDK as well as performance improvements [Announcement Release](https://aws.amazon.com/blogs/developer/aws-sdk-for-go-version-2-general-availability/). 

I would therefore encourage you to reconsider, and maybe ask upstream services to also separate their inclusion of V1 and V2 to have smaller modules so that transient dependencies don't include both V1 and V2 (those would be PRs / Issues in other repositories). 

If after reconsideration, you still think this is not needed, then please feel free to close the PR - and thanks for the time taking to review it.

bep: Oh, well.

bep:
Sorry for the delay. I will complete/test this in #11955. Thanks again.

Issue #9818, opened 2022-04-24
Handle Markdown for .Title and .Description in opengraph properties


bowman2001:
The internal opengraph template `tpl/tplimpl/embedded/templates/opengraph.html` is not expecting markup in the page .Title and .Description. Markup characters for emphasis, which may be useful in .Title and .Description are passed on to the meta properties. 

Opengraph expects plain text, so I propose to plainify these parameters. 

The following lines

```go-html-template
<meta property="og:title" content="{{ .Title }}" />
<meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}" />
```
could be changed into

```go-html-template
<meta property="og:title" content="{{ .Title | $.RenderString | plainify }}" />
<meta property="og:description" content="{{ with .Description | $.RenderString | plainify }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary | plainify }}{{ else }}{{ with .Site.Params.description | $.RenderString | plainify }}{{ . }}{{ end }}{{ end }}{{ end }}" />
```
The same could be done for the itemprops in the Twitter-card template.

bowman2001:
This is the first time, I would like to contribute a small enhancement, so I’m not sure how to proceed. I could change and test the code myself and create a pull request.

bowman2001:
Thanks, I read your contribution guidelines and will work on this in the next days.

Issue #8698, opened 2021-06-26
Add plainify to Title - in internal template opengraph.html


andreseduardop:
### Current circumstance
Inside the `title` parameter, defined in the Front Matter, some users can use html tags, for example `<br>`, `<em>` or `<strong>`.
Example of use:  [Tag in title](https://shower.center/tips/baby-shower/invitaciones-baby-shower/#:~:text=INVITACIONES%20PARA%20BABY%20SHOWER,-%C2%B7GU%C3%8DA%20COMPLETA%C2%B7).
 
### Problem
Open Graph does not support html tags nests inside the `og:title` property:

![og-error-in-title](https://user-images.githubusercontent.com/59033733/123527386-b5eb2800-d6a4-11eb-9e79-9cfcb70a1ccf.PNG)

### Proposal
Adding `plainify` to clean up these tags will prevent the title from being rendered inappropriately when sharing a page on Facebook for example. 

The code of the solution I suggest is in this PR: https://github.com/gohugoio/hugo/pull/8581/commits/68c69c8f0efb2d7458c4156eb26f1ce514c6671f

andreseduardop:
The solution is proposed here: [Pull 8581](https://github.com/gohugoio/hugo/pull/8581)

Issue #8562, opened 2021-05-23
Plainify summary text in internal meta templates


shelbyspees:
If not a bug, it's at least an interesting quirk. I recently started using [semantic line breaks](https://sembr.org) in all my markdown files, including my blog. I tend to use Hugo's `<!--more-->` feature to [generate a post summary](https://gohugo.io/content-management/summaries/#manual-summary-splitting), and this summary gets used in a number [internal templates](https://github.com/gohugoio/hugo/tree/master/tpl/tplimpl/embedded/templates) for Schema.org, Twitter cards, and OpenGraph metadata. These internal templates are preserving the hard line breaks, which then show up in link previews in chat apps and on social media.

I'll use the example of [a recent blog post I wrote](https://shelbyspees.com/2021/05/19/polywork/) (archived [here](https://web.archive.org/web/20210523012346/https://shelbyspees.com/2021/05/19/polywork/)). Here's [that post's summary text](https://github.com/shelbyspees/website/blame/2949eec63ba4f2f46b1607234d44c4431f7fa49e/content/posts/2021/polywork.md#L14-L19) with semantic line breaks:

```markdown
Back in April I was lucky to get an early invite to
[Polywork](https://www.polywork.com/)
and I'm really liking the platform so far.
If you've ever wished you could point people to a single site that shows
all the cool stuff you've done and some more cool stuff you're working on,
Polywork is for you.
```

The post itself renders fine, but when I shared the post in Discord, the summary text had kept all the hard line breaks:

![screenshot of blog unfurl in discord with seemingly random line breaks in the page summary](https://user-images.githubusercontent.com/5217772/119244932-0ca39600-bb2a-11eb-89a2-81a63812dea0.png)

Definitely not what I was going for. I poked around in devtools on that page and noticed that `<meta name="description" ...>` behaves normally, but `<meta name="twitter:description" ...>` and `<meta itemprop="description" ...>` maintain the hard line breaks.

![screenshot of blog post HTML in Chrome devtools showing one description that looks like normal text, and more descriptions with seemingly random line breaks](https://user-images.githubusercontent.com/5217772/119245189-15956700-bb2c-11eb-8d4b-640ef5a68044.png)

It turns out that in my site's theme, [`<meta name="description" ...>` uses the `plainify` function on the summary](https://github.com/shelbyspees/hugo-theme-hello-friend-ng/blob/413547dacf9fc51c7858c914b8f53a82e7a98b7e/layouts/partials/head.html#L5):

[...]

PR #9856, opened 2022-05-04
tpl: Plainify .Title, .Description and .Summary for meta tags

bowman2001:
Opengraph, Schema and Twitter expect strings without markup for their meta tags. The two new internal helper templates plain_title.html and plain_description.html generate them. The direct use of .Title, .Description and .Summary in the internal templates opengraph.html, schema.html and twitter_cards.html is replaced by those helper templates.

Closes [#9818](https://github.com/gohugoio/hugo/issues/9818)
Closes [#8698](https://github.com/gohugoio/hugo/issues/8698) 
Closes [#8562](https://github.com/gohugoio/hugo/issues/8562) at least partially. .Summary is plainified, but there may still remain issues with the resulting formatting.


tpl/tplimpl/embedded/templates/helpers/plain_description_or_summary.html
```diff
@@ -0,0 +1,11 @@
...
+{{- . -}}
+{{- else -}}
+{{- if .IsPage -}}
+{{- .Summary | plainify -}}
+{{- else -}}
+{{- with .Site.Params.description | $.RenderString | plainify -}}
+{{- . -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
chalin, line 11: A few comments:

- You don't need the leading `-` to strip whitespace
-  Use Indentation to make it easier to read.
- Avoid using `.RenderString` in this PR as, IMHO, that would be a breaking change, and it's not really addressing the main issue (#9818)
- Consider naming this file simply `page_description.html` (without create a new `helpers` folder)

So, something like this:

```suggestion
{{ with .Description | plainify -}}
  {{ . -}}
{{ else -}}
  {{ if .IsPage -}}
    {{ .Summary | plainify -}}
  {{ else -}}
    {{ with .Site.Params.description | plainify -}}
      {{ . -}}
    {{ end -}}
  {{ end -}}
{{ end -}}

```

tpl/tplimpl/embedded/templates/helpers/plain_title.html
```diff
@@ -0,0 +1 @@
+{{- .Title | $.RenderString | plainify -}}
```
chalin, line 1: As suggested earlier, I'd avoid introducing `.RenderString` in this PR.

```suggestion
{{ .Title  | plainify -}}

```

(Also, consider not putting this under `helpers`.)

chalin (commented):
Great! I've long wanted an internal template for the page description, used uniformly across other templates (like Twitter, etc). Being able to override these (esp. for the title) has been a need for some projects I work on.  Thanks for the PR, I hope that this gets some traction soon.

tpl/tplimpl/embedded/templates/helpers/plain_title.html
chalin: Hmm, was thinking. Maybe this should just be named `title.html`.

chalin:
One more thought: consider splitting this into two PRs: one for the description update and one for the title update. I think that the title update might be more contentious, and I'd hate for that to delay the acceptance of the description part of this PR. My 2 cents. Cheers!

tpl/tplimpl/embedded/templates/helpers/plain_description_or_summary.html
bowman2001: You’re right about the leading whitespace.

I followed the other templates in the vicinity and they had no indentation, so I’m gonna leave it as it is. But I like the idea and usually do indent template statements.

The `.RenderString` is necessary to translate Markdown into HTML. The following `plainify` only strips HTML and leaves Markdown formatting characters untouched. When you’re mentioned the "breaking change", I guess you had older Hugo versions in mind, which did not have `.RenderString`. So I could change that into an `markdownify`, which has been around much longer. But then again, I read somewhere, that ’.RenderString’ is more suitable than ’markdownify’. I don’t know, which aspect is more important.

I created the ’helpers’ folder, because the template is not meant to be called directly, but only within already existing templates. 

Maybe someone else could also take a look at this. I’m willing to resubmit the PR, if it helps.

tpl/tplimpl/embedded/templates/helpers/plain_title.html
bowman2001: See above and [the docs](https://gohugo.io/functions/plainify).

tpl/tplimpl/embedded/templates/helpers/plain_description_or_summary.html
bowman2001: The `-` at the beginning of the templates are removed, `chomp` added to the `.Summary` pipeline.

chalin:
FYI, see:

- https://github.com/gohugoio/hugo/pull/12901

PR #9858, opened 2022-05-05
Listitem render hook

alexanderhansen:
I needed a way to render lists and list items for different output formats.

I'm writing a food blog and wrote the recipes directly in json-ld format first and used the JSON as input for my template. This has some drawbacks. I use the image processing quite extensively and that's something that couldn't be used in this scenario. Also google recently changed the way json-ld is accepted with sub-recipes so I decided to instead write my recipes in markdown (with the help of some shortcakes) and render them to html and json-ld, so that I can react to changes with a simple edit of my templates. 

But this required the list and especially listitem hooks provided in this PR.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,8 @@
+{
+    "image": "registry.hansenet.local/tools/hugo-devcontainer:0.97.0",
+    "extensions": [
+        "golang.go",
+        "budparr.language-hugo-vscode",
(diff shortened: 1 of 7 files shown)


jordelver:
Not stale, still seems to meet the contribution guidelines

alexanderhansen:
I re-implemented this on HEAD now. I'm still using these changes and build my site from my fork. Let me know if this would be interesting and I create a new PR...

PR #9886, opened 2022-05-12
Fix opengraph URL property

paskal:
Before the change validator http://linter.structured-data.org shows the following warning:

```
property og:url: Object "https://example.com/"@en not compatible with range (ogc:url)
```

Using link instead of meta fixes that. Previously was opened as #8522 but was not merged, @bep can you please look into it?


paskal:
[here](http://linter.structured-data.org/?url=https:%2F%2Fgohugo.io%2Fnews%2F2021%2F0.91.2-relnotes%2F) is a linter scan of  https://gohugo.io/news/2021/0.91.2-relnotes/, I don't see `<link>` mentioned in the Open Graph specification and I don't see linter anymore complaining about it. Thanks for taking your time to review it!

paskal:
Clarification: the warning above is still present but it's unrelevant to storing value in <meta>.

Issue #6101, opened 2019-07-20
Add support for parsing citations and Jupyter notebooks via Pandoc and/or Goldmark extension


gcushen:
# Problems

1. Hugo's Goldmark and Blackfriday don't currently support citations using the popular Cite-Proc standard
   1.  What can we do to empower academics and technical content writers to write citations/bibliographies?
1. Currently Jupyter notebooks cannot be parsed by Hugo into Markdown
   1.   Empower data scientists, analysts, and academics by enabling Pandoc to parse Jupyter notebooks (`.ipynb`) into Markdown
1. Different parsers render things like ToC differently - how can we handle this and let the theme know which parser is being used on a page?
1. Currently, the `markup` option is only available on a per page basis - how can we give users the option to
   1. change the default parser to another parser such as Pandoc in order to parse technical content (if say a user prefer to trade off speed for functionality)
   1.  configure each parser's options in `config.toml`  ✅ **Done**

# Potential solution

Implementation stories:

- [ ] **Parse citations** - in scientific and technical writing, it's common to cite references exported by a reference manager such as Zotero. Here's the [Jekyll equivalent as an example](https://github.com/inukshuk/jekyll-scholar). References exported using the common Bibtex or CSL JSON standards and cited in an article could be parsed and a "References" section generated using:
  - by extending the existing Hugo-Pandoc integration by adding support for [Pandoc's Cite-Proc addon](https://pandoc.org/demo/example19/Extension-citations.html)
  - a Hugo parser or Goldmark extension (perhaps based on libraries such as [Goki Pi](https://github.com/goki/pi/tree/master/langs/bibtex) )
- [ ] **Parse Jupyter notebooks (`.ipynb`) into Markdown** - notebooks are a popular way (over 2.5M on Github) to combine markdown, code, and code outputs such as plots and tables. Conversion of `.ipynb` to a Hugo page bundle of markdown and images can be performed:
  - by extending the existing Hugo-Pandoc integration (Pandoc supports notebook conversion)
  - via Jupyter's nbconvert CLI tool ([Hugo PR](https://github.com/gohugoio/hugo/pull/4281))
  - or parsed directly within Hugo
- [x] Enable **choosing the default markup parser for the entire site** via a `markup` option in `config.toml`
   - Now implemented in #5963 
- [ ] **Allow themes to perform different actions depending on Markup**
[...]

PR #9953, opened 2022-05-30
markup: add --citeproc to pandoc converter

shoeffner:
Adds the citeproc filter to the pandoc converter.

There are several PRs for it this feature already. However, I think
simply adding `--citeproc` is the cleanest way to enable this feature,
with the option to flesh it out later, e.g., in #7529.

Some PRs and issues attempt adding more config options to Hugo which
indirectly configure pandoc, but I think simply configuring Pandoc via
Pandoc itself is simpler, as it is already possible with two YAML
blocks -- one for Hugo, and one for Pandoc:

    ---
    title: This is the Hugo YAML block
    ---
    ---
    bibliography: assets/pandoc-yaml-block-bibliography.bib
    ...
    Document content with @citation!

There are other useful options, e.g., #4800 attempts to use `nocite`,
which works out of the box with this PR:

    ---
    title: This is the Hugo YAML block
    ---
    ---
    bibliography: assets/pandoc-yaml-block-bibliography.bib
    nocite: |
      @*
    ...
    Document content with no citations but a full bibliography:

    ## Bibliography

Other useful options are `csl: ...` and `link-citations: true`, which
set the path to a custom CSL file and create HTML links between the
references and the bibliography.

The following issues and PRs are related:

- Add support for parsing citations and Jupyter notebooks via Pandoc and/or Goldmark extension #6101
  Bundles multiple requests, this PR tackles citation parsing.

- WIP: Bibliography with Pandoc #4800
  Passes the frontmatter to Pandoc and still uses
  `--filter pandoc-citeproc` instead of `--citeproc`.
- Allow configuring Pandoc #7529
  That PR is much more extensive and might eventually supersede this PR,
  but I think --bibliography and --citeproc should be independent
  options (--bibliography should be optional and citeproc can always be
  specified).
- Pandoc - allow citeproc extension to be invoked, with bibliography. #8610
  Similar to #7529, #8610 adds a new config option to Hugo.
  I think passing --citeproc and letting the users decide on the
  metadata they want to pass to pandoc is better, albeit uglier.

Note:
This PR also adds a tiny little bit of unrelated documentation to the external helpers by pointing out MathJax superficially. This certainly needs improvements, but is out of the scope for this PR.

### DIFF
--- a/markup/pandoc/convert.go
+++ b/markup/pandoc/convert.go
@@ -15,10 +15,12 @@
 package pandoc
 
 import (
+	"bytes"
+	"sync"
+
 	"github.com/gohugoio/hugo/common/hexec"
 	"github.com/gohugoio/hugo/htesting"
 	"github.com/gohugoio/hugo/identity"
-
 	"github.com/gohugoio/hugo/markup/converter"
 	"github.com/gohugoio/hugo/markup/internal"
 )
@@ -64,6 +66,9 @@ func (c *pandocConverter) getPandocContent(src []byte, ctx converter.DocumentCon
 		return src, nil
 	}
 	args := []string{"--mathjax"}
+	if supportsCitations(c.cfg) {
+		args = append(args[:], "--citeproc")
+	}
 	return internal.ExternallyRenderContent(c.cfg, ctx, src, binaryName, args)
 }
 
@@ -74,6 +79,45 @@ func getPandocBinaryName() string {
 		return pandocBinary
 	}
 	return ""
+}
+
+var pandocSupportsCiteprocOnce sync.Once
+var pandocSupportsCiteproc bool
+
+// getPandocSupportsCiteproc runs a dump-args to determine if pandoc knows the --citeproc argument
+func getPandocSupportsCiteproc(cfg converter.ProviderConfig) (bool, error) {
+	var err error
+
+	pandocSupportsCiteprocOnce.Do(func() {
+		argsv := []any{"--dump-args", "--citeproc"}
+
+		var out bytes.Buffer
+		argsv = append(argsv, hexec.WithStdout(&out))
+
+		cmd, err := cfg.Exec.New(pandocBinary, argsv...)
+		if err != nil {
+			pandocSupportsCiteproc = false
+			return
+		}
+
+		err = cmd.Run()
+		if err != nil {
+			pandocSupportsCiteproc = false
+			return
+		}
+		pandocSupportsCiteproc = true
+	})
+
+	return pandocSupportsCiteproc, err
+}
+
+// supportsCitations returns true if citeproc is available
+func supportsCitations(cfg converter.ProviderConfig) bool {
+	if Supports() {
+		supportsCiteproc, err := getPandocSupportsCiteproc(cfg)
+		return supportsCiteproc && err == nil
+	}
+	return false
 }
 
 // Supports returns whether Pandoc is installed on this computer.
--- a/markup/pandoc/convert_test.go
+++ b/markup/pandoc/convert_test.go
@@ -25,7 +25,7 @@ import (
 	qt "github.com/frankban/quicktest"
 )
 
-func TestConvert(t *testing.T) {
+func setupTestConverter(t *testing.T) (*qt.C, converter.Converter, converter.ProviderConfig) {
 	if !Supports() {
 		t.Skip("pandoc not installed")
 	}
@@ -34,11 +34,109 @@ func TestConvert(t *testing.T) {
 	var err error
 	sc.Exec.Allow, err = security.NewWhitelist("pandoc")
 	c.Assert(err, qt.IsNil)
-	p, err := Provider.New(converter.ProviderConfig{Exec: hexec.New(sc, "", loggers.NewDefault()), Logger: loggers.NewDefault()})
+	cfg := converter.ProviderConfig{Exec: hexec.New(sc, "", loggers.NewDefault()), Logger: loggers.NewDefault()}
+	p, err := Provider.New(cfg)
 	c.Assert(err, qt.IsNil)
 	conv, err := p.New(converter.DocumentContext{})
 	c.Assert(err, qt.IsNil)
-	b, err := conv.Convert(converter.RenderContext{Src: []byte("testContent")})
+	return c, conv, cfg
+}
+
+func TestConvert(t *testing.T) {
+	c, conv, _ := setupTestConverter(t)
+	output, err := conv.Convert(converter.RenderContext{Src: []byte("testContent")})
 	c.Assert(err, qt.IsNil)
-	c.Assert(string(b.Bytes()), qt.Equals, "<p>testContent</p>\n")
+	c.Assert(string(output.Bytes()), qt.Equals, "<p>testContent</p>\n")
+}
+
+func runCiteprocTest(t *testing.T, content string, expectContained []string, expectNotContained []string) {
+	c, conv, cfg := setupTestConverter(t)
+	if !supportsCitations(cfg) {
+		t.Skip("pandoc does not support citations")
+	}
+	output, err := conv.Convert(converter.RenderContext{Src: []byte(content)})
+	c.Assert(err, qt.IsNil)
+	for _, expected := range expectContained {
+		c.Assert(string(output.Bytes()), qt.Contains, expected)
+	}
+	for _, notExpected := range expectNotContained {
+		c.Assert(string(output.Bytes()), qt.Not(qt.Contains), notExpected)
+	}
+}
+
+func TestGetPandocSupportsCiteprocCallTwice(t *testing.T) {
+	c, _, cfg := setupTestConverter(t)
+
+	supports1, err1 := getPandocSupportsCiteproc(cfg)
+	supports2, err2 := getPandocSupportsCiteproc(cfg)
+	c.Assert(supports1, qt.Equals, supports2)
+	c.Assert(err1, qt.IsNil)
+	c.Assert(err2, qt.IsNil)
+}
+
+func TestCiteprocWithHugoMeta(t *testing.T) {
+	content := `
+---
+title: Test
+published: 2022-05-30
+---
+testContent
+`
+	expected := []string{"testContent"}
+	unexpected := []string{"Doe", "Mustermann", "2022", "Treatise"}
+	runCiteprocTest(t, content, expected, unexpected)
+}
+
+func TestCiteprocWithPandocMeta(t *testing.T) {
+	content := `
+---
+---
+---
+...
+testContent
+`
+	expected := []string{"testContent"}
+	unexpected := []string{"Doe", "Mustermann", "2022", "Treatise"}
+	runCiteprocTest(t, content, expected, unexpected)
+}
+
+func TestCiteprocWithBibliography(t *testing.T) {
+	content := `
+---
+---
+---
+bibliography: testdata/bibliography.bib
+...
+testContent
+`
+	expected := []string{"testContent"}
+	unexpected := []string{"Doe", "Mustermann", "2022", "Treatise"}
+	runCiteprocTest(t, content, expected, unexpected)
+}
+
+func TestCiteprocWithExplicitCitation(t *testing.T) {
+	content := `
+---
+---
+---
+bibliography: testdata/bibliography.bib
+...
+@Doe2022
+`
+	expected := []string{"Doe", "Mustermann", "2022", "Treatise"}
+	runCiteprocTest(t, content, expected, []string{})
+}
+
+func TestCiteprocWithNocite(t *testing.T) {
+	content := `
+---
+---
+---
+bibliography: testdata/bibliography.bib
+nocite: |
+  @*
+...
+`
+	expected := []string{"Doe", "Mustermann", "2022", "Treatise"}
+	runCiteprocTest(t, content, expected, []string{})
 }
--- a/docs/content/en/content-management/bibliography.md
+++ b/docs/content/en/content-management/bibliography.md
@@ -1,0 +1,50 @@
+---
+title: Bibliographies in Markdown
+linkTitle: Bibliography
+description: Include citations and a bibliography in Markdown using LaTeX markup.
+categories: [content management]
+keywords: [latex,pandoc,citation,reference,bibliography]
+menu:
+  docs:
+    parent: content-management
+    weight: 320
+weight: 320
+toc: true
+---
+
+{{< new-in 0.144.0 />}}
+
+## Citations and Bibliographies
+
+[Pandoc](https://pandoc.org) is a universal document converter and can be used to convert markdown files.
+
+With **Pandoc >= 2.11**, you can use [citations](https://pandoc.org/MANUAL.html#extension-citations).
+One way is to employ [BibTeX files](https://en.wikibooks.org/wiki/LaTeX/Bibliography_Management#BibTeX) to cite:
+
+```
+---
+title: Citation document
+---
+---
+bibliography: assets/bibliography.bib
+...
+This is a citation: @Doe2022
+```
+
+Note that Hugo will **not** pass its metadata YAML block to Pandoc; however, it will pass the **second** meta data block, denoted with `---` and `...` to Pandoc.
+Thus, all Pandoc-specific settings should go there.
+
+You can also add all elements from a bibliography file (without citing them explicitly) using:
+
+```
+---
+title: My Publications
+---
+---
+bibliography: assets/bibliography.bib
+nocite: |
+  @*
+...
+```
+
+It is also possible to provide a custom [CSL style](https://citationstyles.org/authors/) by passing `csl: path-to-style.csl` as a Pandoc option.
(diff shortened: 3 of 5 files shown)


bep:
@shoeffner thanks for this. This PR is certainly easier to reason about (because of its size) than some of the others I've seen.

That said, the test failure on Linux shows that this needs some rethinking, as that flag is not available in the version we test on, fetched via:

```
sudo apt-get install -y pandoc
```

I assume this flag was added recently? I'm not sure how to handle this.

shoeffner:
The Linux build seems to use pandoc 2.5, but --citeproc is only supported > 2.11 (https://pandoc.org/releases.html#pandoc-2.11-2020-10-11 – on macOS, 2.18 is installed, so I didn't notice) – before `--filter pandoc-citeproc` was the way to go.

If I remember correctly, pandoc-citeproc had to be installed manually, but I will check it out in a container and see what I can do!

shoeffner:
choco and brew seem to install 2.18, so Windows and macOS should not need any specific handling.

On Linux we have everything from 2.5 (Ubuntu 20.04, e.g. GH actions) to 2.17 (Arch Linux); the most common version seems to be 2.9 (Debian stable, Ubuntu 22.04, and other distros), here `pandoc-citeproc` must be installed additionally.

This makes the pull request much more complicated than what I was aiming for, so feel free to close it again. My current approach is now:

```
if pandoc >= 2.11
    add --citeproc
else if pandoc-citeproc available
    add --filter pandoc-citeproc
otherwise
    keep the old behavior
```

If the citations are not supported (due to the version or pandoc-citeproc dependency), SupportsCitations is false and the corresponding tests are skipped. I also added a remark to the docs.
However, I was unable to test all configurations yet (i.e., Linux with newer pandoc, older pandoc, with and without pandoc-citeproc; nor did I test windows), thus I update the PR to "WIP".

An alternative could be to check for the version and ignore pandoc-citeproc. This would make it much simpler and whoever wants to use citeproc might also want to use other pandoc features anyways. But it puts some burden on the users who have to install a newer pandoc version.

What do you think?

markup/pandoc/convert.go
```diff
@@ -77,6 +80,64 @@ func getPandocBinaryName() string {
...
+
+func getCiteprocBinaryName() string {
+	if hexec.InPath(citeprocBinary) {
+		return citeprocBinary
+	}
+	return ""
+}
+
+func getCiteprocArgs(cfg converter.ProviderConfig) []string {
+	pandocVersion, err := getPandocVersion(cfg)
```
bep, line 93: I think it's safe to assume that the pandoc version can be cached (e.g. via a `sync.Once`). Shelling out to `os/exec` isn't cheap.

markup/pandoc/convert_test.go
```diff
@@ -25,18 +25,114 @@ import (
 	qt "github.com/frankban/quicktest"
 )
 
-func TestConvert(t *testing.T) {
+func SetupTestConverter(t *testing.T) (*qt.C, converter.Converter, converter.ProviderConfig) {
```
bep, line 28: This can be unexported?

markup/pandoc/convert.go
```diff
@@ -77,6 +80,64 @@ func getPandocBinaryName() string {
...
+		cfg.Logger.Errorf("%s --version: %v", pandocBinary, err)
+	}
+
+	outbytes := bytes.Replace(out.Bytes(), []byte("\r"), []byte(""), -1)
+	output := strings.Split(string(outbytes), "\n")[0]
+	version := strings.Split(output, " ")[1]
+	return version, nil
+}
+
+func SupportsCitations(cfg converter.ProviderConfig) bool {
```
bep, line 130: Can be unexported?

markup/pandoc/convert_test.go
shoeffner: I didn't know that concept, as I haven't done much programming in Go, thank you very much!

markup/pandoc/convert.go
shoeffner: Done!

shoeffner:
I rebased the commits and squashed them.
Unless there is anything else or the CI fails again, the PR is for the moment done, so I will also remove the WIP label.

shoeffner:
I updated the PR again, I made a mistake (`err != nil` instead of `err == nil`)which caused the tests to skip and never activated pandoc. It should now work as expected.

shoeffner:
Rebased on master.

markup/pandoc/convert.go
```diff
@@ -77,6 +85,51 @@ func getPandocBinaryName() string {
 	return ""
 }
 
+var versionOnce sync.Once
+
+// getPandocVersion parses the pandoc version output
+func getPandocVersion(cfg converter.ProviderConfig) (string, error) {
+	var version string
```
bep, line 92: You need to pull the `version` out of the function; this will fail he second time `getPandocVersion` gets called.

markup/pandoc/convert.go
```diff
@@ -77,6 +85,51 @@ func getPandocBinaryName() string {
...
+	var err error
+
+	versionOnce.Do(func() {
+		argsv := collections.StringSliceToInterfaceSlice([]string{"--version"})
```
bep, line 96: The above works, but it doesn't make any sense when you can just do `argsv := []string{"--version"}`

markup/pandoc/convert.go
shoeffner: Done, thanks.

markup/pandoc/convert.go
shoeffner: I tried, but it does not work that way:

```go
argsv := []string{"--version"}

var out bytes.Buffer
argsv = append(argsv, hexec.WithStdout(&out))
```
does not compile:
```
markup/pandoc/convert.go:98:25: cannot use hexec.WithStdout(&out) (value of type func(c *hexec.commandeer)) as type string in argument to append
```
It works with StringSliceToInterfaceSlice, but don't ask me why – I adapted the whole function from markup/internal/external.go:
https://github.com/gohugoio/hugo/blob/8e2fd55923a4da38eb2ddda51dc1b96943be0c56/markup/internal/external.go#L23

shoeffner: I assume that the `InterfaceSlice` is compatible to contain not only strings but also `func(*hexec.comandeer)`, so a conversion probably does make sense here. But again, I am new to Go, so I might be wrong.

markup/pandoc/convert.go
shoeffner: I also added a unit test which calls it twice to detect this issue.

markup/pandoc/convert.go
bep: Yea, right, it's a slice of `any`, you need to do

```go
argsv := []any{"--version"}
```

shoeffner:
Sorry for the late reaction. I fixed it to use `argsv := []any{"--version"}` instead of `collections.StringSliceToInterfaceSlice...`.

shoeffner:
Rebased on master.

Setr43:
Вітаємо в буфері обміну Gboard. Скопійований текст зберігатиметься тут.

shoeffner:
Okay, comparing strings 2.5 and 2.11 was a stupid idea... Now it has a proper type and a function to compare two instances.

https://github.com/gohugoio/hugo/commit/3c5e31fd17898637f96e99900613ba9658a96fcf#diff-f662bf93836a7230b77cdb532095a04622eb9eda3054124e8ba9399786870efaR88-R94 

Additionally, I included some test cases for those comparisons. I hope it will work now ;-)

shoeffner:
Unsupported pandoc should now cause the tests to SKIP instead of causing a panic.

shoeffner:
Rebased on the current tip of the master branch.

jacobmerson:
@shoeffner thanks for putting this PR together. It is quite useful. Do you have a sense if this, or something like it will end up getting merged? Being able to put citations in text is quite important.

shoeffner:
Thanks for getting back to this, I rebased the changes on the master.

I don't know whether this will be merged and if the pipelines still work, nor about the progress on alternative approaches, as I haven't followed the current developments much. Maybe @bep knows more about it.

shoeffner:
Is this still relevant to Hugo? I have to admit that I don't even use it right now, so I haven't followed up anymore.
But if it's still useful, I am happy to rebase/update/work on the PR – otherwise, feel free to close it :)

shoeffner:
The docs moved around quite a bit over the last 1.5 years, so I moved the section about citations to its own markdown file during the rebase. Please let me know if I should change that (or, for example, the weight) – or simply change it.

I also found out that the test setup (or provider config) seems to have changed, I am currently working on updating that.

shoeffner:
I found the issue in the test setup (the ProviderConfig was not passed correctly, and I seem to have written that bug the last time I rebased in August '23, whoops).

There's also been some output format changes in Pandoc, so I will relax the tests a bit and make sure they only check for "contains" on the references, not the exact markup (as that will be different between Pandoc versions).

shoeffner:
Done, the tests no longer rely on a specific output format and only check if the author last names, the year, and a word from the title are included in the documents – or not included if no citations was there.

shoeffner:
I received an email [that the tests have failed](https://github.com/gohugoio/hugo/actions/runs/13347029218/job/37279359322), but I think that's what I already fixed in the latest commit.

alametti:
Just tested this rebase and it works fine; I built with `$ CGO_ENABLED=1 go build -tags extended .` after extracting.

### My current workflow with regular extended Hugo:

- Unprocessed markdown content goes in `${DRAFT_DIR}`
- CI pipeline preprocesses with `pandoc` and some ugly kludges:  
```sh
touch .dummy;
ls -1 ${DRAFT_DIR}/*.md \
| xargs basename \
| xargs -i pandoc-${PANDOC_VERSION}/bin/pandoc \
-C -t gfm --resource-path=${DRAFT_DIR} -H .dummy \
${DRAFT_DIR}/{} -o ${POST_DIR}/{}
```
- Build site with `hugo`

### Workflow with this PR:

- Markdown content lives in `content/`
- Add `markup: pandoc` to front matter
- Whitelist `pandoc` executable in `hugo.yaml`
- Build site with `hugo`

Much easier and no preprocessing needed!

jmooring:
@shoeffner I'm not in a position to judge the value of this addition, but I do have a question. Instead of all of the version checking code, can you just test the exit code?

```sh
pandoc --citeproc --dump-args
echo $? # 0 (success)

pandoc --foo --dump-args
echo $? # 6 (failure)
```

shoeffner:
> can you just test the exit code

Sounds plausible, I'll give it a try!


> Add `markup: pandoc` to front matter

Hm, should I add that to the documentation about citations I added? I think
the test examples work without that.

Thanks for trying it out!

alametti:
> Add `markup: pandoc` to front matter

To clarify, this is because I did not set `pandoc` as the global markup engine, only for select content.

shoeffner:
Thanks, that makes sense!

jmooring:
> To clarify, this is because I did not set pandoc as the global markup engine, only for select content.

Why not  change the file extension to `.pdc` or `.pandoc`?

alametti:
> Why not change the file extension to `.pdc` or `.pandoc`?

Because I didn't know I could do that. Thanks for the tip - I'm not a hugo expert by any means!

patch-work:
> * Build site with `hugo`
> 
> ### Workflow with this PR:
> * Markdown content lives in `content/`
> * Add `markup: pandoc` to front matter
> * Whitelist `pandoc` executable in `hugo.yaml`
> * Build site with `hugo`
> 
> Much easier and no preprocessing needed!

Is it available on hugo 0.145.0?

shoeffner:
I changed the citeproc detection to use `--dump-args`, as suggested by @jmooring . It seems to work for me, so feel free to test it. I also rebased on the latest master version again.

shoeffner:
Hm, it seems the tests were unable to pull some data from X.com - I don't think that's related to my PR, though. Or am I mistaken?
The tests I wrote seem to have worked (markup/pandoc/convert_test.go).

bep:
>Hm, it seems the tests were unable to pull some data from X.com - I don't think that's related to my PR, though. Or am I mistaken?

The X.com API has been unstable lately.

jmooring:
You' re logging _and_ returning errors, resulting in repetitive console output:

```text
Start building sites … 
hugo v0.146.0-DEV-038007c117257f13acde3d2d230003bfd8161b9b+extended+withdeploy linux/amd64 BuildDate=2025-03-18T17:16:55Z VendorInfo=jmooring

ERROR Could not call pandoc: access denied: "pandoc" is not whitelisted in policy "security.exec.allow"; the current security configuration is:

[security]
  enableInlineShortcodes = false

  [security.exec]
    allow = ['^((dart-)?sass|git|go|npx|postcss|tailwindcss|gnome-text-editor|nano|asciidoctor|babel|xpandoc|rst2html)$']
    osEnv = ['(?i)^((HTTPS?|NO)_PROXY|PATH(EXT)?|APPDATA|TE?MP|TERM|GO\w+|(XDG_CONFIG_)?HOME|USERPROFILE|SSH_AUTH_SOCK|DISPLAY|LANG|SYSTEMDRIVE)$']

  [security.funcs]
    getenv = ['^HUGO_', '^CI$']

  [security.http]
    methods = ['(?i)GET|POST']
    urls = ['.*']
Total in 14 ms
Error: error building site: render: failed to render pages: render of "/home/jmooring/code/hugo-testing/content/_index.pdc" failed: "/home/jmooring/code/hugo-testing/layouts/_default/home.html:11:5": execute of template failed: template: _default/home.html:11:5: executing "main" at <.Content>: error calling Content: "/home/jmooring/code/hugo-testing/content/_index.pdc:1:1": access denied: "pandoc" is not whitelisted in policy "security.exec.allow"; the current security configuration is:

[security]
  enableInlineShortcodes = false

  [security.exec]
    allow = ['^((dart-)?sass|git|go|npx|postcss|tailwindcss|gnome-text-editor|nano|asciidoctor|babel|xpandoc|rst2html)$']
    osEnv = ['(?i)^((HTTPS?|NO)_PROXY|PATH(EXT)?|APPDATA|TE?MP|TERM|GO\w+|(XDG_CONFIG_)?HOME|USERPROFILE|SSH_AUTH_SOCK|DISPLAY|LANG|SYSTEMDRIVE)$']

  [security.funcs]
    getenv = ['^HUGO_', '^CI$']

  [security.http]
    methods = ['(?i)GET|POST']
    urls = ['.*']
```

shoeffner:
I removed the two log lines, is it better this way? Otherwise, can you share the example you are trying so that I can reproduce the behavior?

shoeffner:
@jmooring, I only ever run the tests with `mage -v check` and they do not produce any meaningful output as they suppress it all.

I don't know what example project you use to test the PR, even though it was clear to me that you removed pandoc from the allow list.
So please excuse that I asked whether it was better; what I wanted to ask is whether this was closer to your preferred behavior, as I was trying to incorporate your feedback. I did not mean to offend and I am sorry if I did.

Of course, I understand that I am not the ideal contributor to work on this because I haven't worked with Hugo since shortly after I originally drafted this PR in 2022, as such, I am not even using this feature myself anymore. I am trying my best to get a feel for the project and improve as I get suggestions, be it by you, @bep - who reviewed the first draft and helped me over a few Go hurdles -, or others in this thread.

If you feel I am not confident enough on contributing this, I will of course step back from the PR, you can close it, and/or someone else can pick it up.

foxyseta:
I am personally working with Hugo and `pandoc --citeproc` right now, if you need somebody to pick this up.

shoeffner:
Sure, I gave you access to my fork so you can work on this PR directly – but feel free to create your own if you like :)

Qodo-Free-For-OSS:
Hi, getPandocSupportsCiteproc shadows the outer `err` with `cmd, err := ...`, so the function always returns a nil error even when pandoc execution fails. This silently disables citeproc on failure and prevents callers/tests from detecting the real cause (e.g. exec policy denial or unsupported flag).

**Severity:** action required | **Category:** correctness

**How to fix:** Fix err shadowing/propagation

> ### Issue description
> `getPandocSupportsCiteproc` declares an outer `err` but then shadows it inside the `sync.Once` closure (`cmd, err := ...`). As a result, the function always returns a nil error even when probing pandoc fails.
> 
> ### Issue Context
> The return signature includes `error`, and `supportsCitations` relies on `err == nil` to decide whether to add `--citeproc`. With the current bug, failures are indistinguishable from success from an error-reporting perspective.
> 
> ### Fix Focus Areas
> - markup/pandoc/convert.go[88-111]
> - markup/pandoc/convert_test.go[67-75]
> 
> ### What to change
> - Avoid shadowing by using `cmd, err = ...` (declare `var cmd hexec.Runner` or similar outside the closure) or rename the inner error variable and assign it to the outer one.
> - Ensure errors from `cfg.Exec.New(...)` and `cmd.Run()` are returned to the caller.
> - Update the test to assert the corrected behavior (e.g., if the probe fails, `err` should be non-nil; if it succeeds, `err` should be nil).

---

shoeffner:
@Qodo-Free-For-OSS feel free to pick up the code in the PR and create your own. I am no longer working on this and will close it now.

Dear maintainers, feel free to take the code if there is any interest :)

PR #9954, opened 2022-05-30
common: add hugo.GoVersion

### LINKED ISSUE #9849: Add hugo.GoVersion
The information is easily obtainable.

khayyamsaleem:
closes #9849. This enables `hugo.GoVersion` in templates to access the
version of Go embedded in the running Hugo binary.


common/hugo/hugo.go
```diff
@@ -61,6 +61,12 @@ func (i Info) Version() VersionString {
 	return CurrentVersion.Version()
 }
 
+// GoVersion returns the version of Go that the hugo binary is running with.
```
bep, line 64: * I think it would be better (or at least more consistent) if you just add GoVersion as a field and add it to the constructor.
* I think it's more precise to say "that the hugo binary is built with"...
* Also, the commit message is not totally in line with the commit message guidelines (does not start with a capital letter after the prefix).

bep (changes requested):

common/hugo/hugo.go
khayyamsaleem: feedback accommodated, thanks for the speedy review!

Issue #9971, opened 2022-06-03
resources: Failed to resolve media type for remote resource


vanbroup:
The mediaTypes defined in the default Hugo configuration are not registered at build time resulting in errors when using `resources.GetRemote` on certain file types.

```
error calling resources.GetRemote: failed to resolve media type for remote resource
```

PR #9972, opened 2022-06-03
resources: Register MediaTypes before build

vanbroup:
This fixes an issue with the `resources.GetRemote` call that fails for some media types, even when these are included in the default Hugo configuring. This is because the media types only get registered when running Hugo as a server and after the build is completed.

Fixes: #9971

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -127,8 +127,10 @@ func initializeConfig(mustHaveConfigFile, failOnInitErr, running bool,
 		return nil, err
 	}
 
-	for _, s := range c.hugo().Sites {
-		s.RegisterMediaTypes()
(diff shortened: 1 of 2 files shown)


commands/hugo.go
```diff
@@ -127,6 +127,10 @@ func initializeConfig(mustHaveConfigFile, failOnInitErr, running bool,
 		return nil, err
 	}
 
+	for _, s := range c.hugo().Sites {
```
bep, line 130: I think we have some commands that needs to complete even without a valid config, meaning `c.hugo()` will just block forever. It would ease my mind if you rewrote the above to:

```go
if h := c.hugoTry(); h != nil {
range h.Sites ...
}
...

PR #9976, opened 2022-06-04
tpl/path: Add path.BaseName function

### LINKED ISSUE #9973: add path.BaseName function
I'm trying to write an image render hook that looks for the presence of smaller image files to fill out the srcset field for responsive images.  I've settled on the following naming convention:

image1.jpg
image1_400.jpg
image1_800.jpg
...etc...

where the numbers are the various widths of the image.  Not all sizes will be available for all images, so the hook needs to test if it exists and only emit it if it exists.

I've settled on having a site-wide parameter that has a list of the sizes to look for.  That way I can have a tool check that parameter and generate the appropriate images.

To generate the path to look for, I am currently doing this:

      {{- $path := replaceRE ".[jJ][pP][gG]$" (printf "_%sw%s" $size (path.Ext $src.Name)) $src.Name -}}

which only works for jpg files.  I'd like to do something like this:

      {{- $path := (printf "%s/%s_%sw%s" (path.Dir $src.Name) (path.BaseName $src.Name) $size (path.Ext $src.Name)) -}}

Which would work for all file extensions.

The File object has BaseFileName which provides this operation, and I don't really care what the operation is named, but I would like to have the operation.

Some test cases (I expect they are similar to the .File.BaseFileName test cases):

```
path.BaseName "image1.jpg"  => image1
path.BaseName "/a/long/path/foo.txt" => foo
path.BaseName "filename.en.jpg" => filename.en
```

jmooring:
Closes #9973

### DIFF
--- a/docs/content/en/functions/path.BaseName.md
+++ b/docs/content/en/functions/path.BaseName.md
@@ -1,0 +1,24 @@
+---
+title: path.BaseName
+description: BaseName returns the last element of a path, removing the extension if present.
+date: 2022-06-04
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [path, base]
+signature: ["path.BaseName PATH"]
+relatedfuncs: [path.Base, path.Clean, path.Dir, path.Ext, path.Join, path.Split]
+deprecated: false
+---
+
+If `PATH` is empty, `.` is returned.
+
+**Note:** On Windows, `PATH` is converted to slash (`/`) separators.
+
+```go-html-template
+{{ path.BaseName "a/news.html" }} → "news"
+{{ path.BaseName "news.html" }} → "news"
+{{ path.BaseName "a/b/c" }} → "c"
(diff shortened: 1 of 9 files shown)


docs/content/en/functions/path.BaseName.md
```diff
@@ -0,0 +1,24 @@
...
+deprecated: false
+---
+
+If `PATH` is empty, `.` is returned.
+
+**Note:** On Windows, `PATH` is converted to slash (`/`) separators.
+
+```go-html-template
+{{ path.BaseName "a/news.html" }} → "news"
+{{ path.Base "news.html" }} → "news"
```
snicolai-blog, line 21: Doesn't Path.Base return news.html instead of news?  That was the purpose of BaseName vs Base.

jmooring: Typo in example. Good catch. Thanks.

PR #9977, opened 2022-06-05
livereload: Use `X-Forwarded-Host` for Codespace livereload

### LINKED ISSUE #9936: LiveReload doesn't work in GitHub Codespace

<pre>
$ hugo version
hugo v0.99.1-d524067382e60ce2a2248c3133a1b3af206b6ef1+extended linux/amd64 BuildDate=2022-05-18T11:18:14Z VendorInfo=gohugoio
</pre>

Yes

## General Problem description
LiveReload doesn't seem to work in GitHub Codespaces. This seems to be due to the port being appended to the url even when `--appendPort=false` is passed to the cli

image

satotake:
Codespace has 2 types of usage

1. in browser
2. vscode on local computer

As long as you select 2 (on local), Hugo handles livereload expectedly.
But if you use it in browser, Hugo does not reload on file change, as #9936 said.

This issue happens because `CheckOrigin` always fails.
Remote server could rewrite request host name.
Fix this by respecting `X-Forwarded-Host` header during origin checking

After merging this, you can preview changes lively with codespaece in browser.

```sh
hugo server --liveReloadPort 443
```

Close #9936

### DIFF
--- a/livereload/livereload.go
+++ b/livereload/livereload.go
@@ -62,7 +62,13 @@ var upgrader = &websocket.Upgrader{
 			return false
 		}
 
-		if u.Host == r.Host {
+		rHost := r.Host
+		// For Github codespace in browser #9936
+		if forwardedHost := r.Header.Get("X-Forwarded-Host"); forwardedHost != "" {
+			rHost = forwardedHost
+		}
+
+		if u.Host == rHost {
 			return true
 		}
 


markgoho:
thank you @satotake -- totally amazed by the quickness with which you figured out what was happening and how to fix it

PR #9982, opened 2022-06-06
hugolib: Add test case for #9082

satotake:
- related 
    - #8895 
    - #9082 
   
I checked the test fails if we revert [this change](https://github.com/gohugoio/hugo/pull/9082/files)

### DIFF
--- a/hugolib/dates_test.go
+++ b/hugolib/dates_test.go
@@ -214,3 +214,32 @@ func TestTimeOnError(t *testing.T) {
 	b.Assert(b.BuildE(BuildCfg{}), qt.Not(qt.IsNil))
 
 }
+
+// Issue 8895
+func TestDateTomlTimeZone(t *testing.T) {
+	b := newTestSitesBuilder(t)
+
+	dateWithQuote := `+++
+title = "String"
+date = "2021-08-16T06:00:00+00:00"
++++
+`
+	dateWithoutQuote := `+++
+title = "Date"
+date = 2021-08-16T06:00:00+00:00
++++
+`
+
+	b.WithContent(
+		"string.md", dateWithQuote,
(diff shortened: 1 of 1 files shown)


bep:
Hmm... OK, I'll have a look.

bep:
OK, this I don't understand ... This passes 

```go
func TestTOMLDates(t *testing.T) {
	t.Parallel()

	files := `
-- config.toml --
-- content/p1.md --
+++
title = "TOML Date with UTC offset"
date = 2021-08-16T06:00:00+00:00
+++

## Foo
Date: {{ .Date | time.Format ":time_long" }}

`

	b := NewIntegrationTestBuilder(
		IntegrationTestConfig{
			T:           t,
			TxtarString: files,
		},
	).Build()

	b.AssertFileContent("public/p1/index.html", `Date: 6:00:00 am UTC`)

}

```

Even if I comment out the `i = vv.Format(time.RFC3339)` fix. Your test (which looks very similar), fails...

bep:
I'll close this, we'll keep that snippet of code for now, but I think it's safe to say that we don't understand the current behaviour fully.

satotake:
I researched this issue.

On unix machine, `/etc/localtime` setting is used as local timezone.
You can change this setting using `TZ` env var.

Only if the offset value is matched this setting, `cast.ToTimeInDefaultLocationE` returns zone name.
(Otherwise, it returns empty string.)

See [this](https://go.dev/play/p/4UjzCwzfsnL).

This explains why we were confused [this issue](https://github.com/gohugoio/hugo/issues/8895).

Actually, this is independent of toml (or `go-toml`).
Machine-level timezone setting matters.

This will be confusing when remote building environments are used (like #9996).

Issue #9978, opened 2022-06-05
Load SVG files with esbuild's api.LoaderText with in js.Build


zmsy:
On my site I've got a custom color theme selector, and I've got a Preact element where I'd like to inline an SVG so I can apply styles to it, and make it responsive to CSS custom properties. This is currently achieved using a custom `Figure` component:

```tsx
// Figure.tsx

import { FunctionComponent } from "preact";

export const Figure: FunctionComponent<{ svg: string }> = ({ svg }) => {
  // This "color-responsive-svg" class adds a number of selectors for the paths, etc within the svg to
  // make them response to various css custom properties, and then the actual SVG file gets inlined
  // here as a child element.
  return (
    <figure class="color-responsive-svg" dangerouslySetInnerHTML={{ __html: svg }} />
  );
};
```

Below, using the component like so doesn't work. You'll get something akin to `JSBUILD: failed to transform "ts/index.tsx" (text/tsx): "my-icon.svg:1:0": Unexpected "<"`

```
// NavModal.tsx

import { Figure } from "./Figure";
import myIcon from "./my-icon.svg";

export const NavModal: FunctionComponent = () => {
  return (
    <Figure svg={myIcon} />
  );
};
```

If you switch the file extension to `.txt`, you can import the file and render it as expected. This looks like it's because `".svg"` is currently not supported in the extensions array: https://github.com/gohugoio/hugo/blob/master/resources/resource_transformers/js/options.go#L134

It looks like loading SVG as text is the [documented approach for using this in ESBuild](https://esbuild.github.io/api/#loader), and that seems like a nice default to me.

stackkrocket:
Has this been fixed?

bep:
From what I understand @zmsy the solution to this is to add `".svg":  api.LoaderText` here:

https://github.com/gohugoio/hugo/blob/master/resources/resource_transformers/js/options.go#L134

Note that any JavaScript ESBuild plugins you find on the interweb is not relevant for Hugo.

stackkrocket:
@bep  I think I may have missed some few pointer in the contributing guidelines. I wish to correct them and make subsequent adequate changes that fixes this issue.

stackkrocket:
@bep Have you had time to go through my edited commit? It seems I have passed all the checks to make a successful commit/PR.

PR #9986, opened 2022-06-07
Added .svg file support in issue #9978

stackkrocket:


bep:
Thanks, could you read this and amend your commit message accordingly:

https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines

bep:
Hey, and thanks, but it's the commit message I'm concerned about (not the PR title):

image

You need to amend the commit and change the title/body.

Note that I could fix this myself, which would be the fast route, but the idea is that it will go more smoothly the next time if we get it right now.

stackkrocket:
I agree. It makes sense to get it right now for future contribution(s) as I have already started using the product myself. 

This seems like a use case for the --amend flag.

 I'll get on it right away. I apologize if I am missing any obvious points, this is one of my early contribution as a first timer.

leonid-shevtsov:
So, will there be an updated PR? I am trying to port a project from Webpack to Hugo's ESBuild and this is a blocker.

Issue #385, opened 2014-08-04
permalinks :slug behaves unexpectedly


natefinch:
In page metadata, the if the slug is set, it is the last part of the page's URL, otherwise the name of the file is used as the last part of the URL.

However, if you set a permalinks format for a section, :slug behaves differently.  It is either the slug, or the title.  There's no way to replicate the slug-or-filename behavior of the page metadata with permalinks.  That's really too bad, because I like the slug-or-filename behavior, and almost never want my title as the slug.

IMO, :slug behaves incorrectly.  I think it should mimic the behavior of the page metadata, where it is slug-or-filename.  However, I would imagine that one can't just change the behavior of :slug, otherwise we're all-of-a-sudden change some peoples' urls on them.

For now, I suggest a :slugorfile token that works the way the metadata slug works.

I'm not sure how Hugo handles deprecation/changing of features, but this is how I'd do it, if you eventually want to make :slug work like :slugorfile, and give everyone as much chance as possible to not have their stuff broken.

in v0.12, add :slugorfile, and add a token called :slugortitle, which works identically to the way :slug works now.  If someone is detected as using :slug, log a warning that :slug will change behavior in the next release, and they should change to using :slugortitle before upgrading, to maintain behavior.  
in v0.13, make :slug work like :slugorfile, and add a deprecation warning to :slugorfile that it should be changed to :slug, because it'll be removed in the next release.  
in v0.14 remove :slugorfile

I'm happy to submit a branch with :slugorfile and, if desired, the rest of the deprecation stuff... but I'd at least like to see a way to mimic the slug-or-filename behavior of the page metadata.

bep:
**Note/Update:** This issue is marked as stale, and I may have said something earlier about "opening a thread on the discussion forum". Please don't.

If this is a **bug** and you can still reproduce this error on the latest <code>release</code> or the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

PR #10003, opened 2022-06-13
resources/page: Add :slugorfilename attribute

### LINKED ISSUE #4739: Permalink variable for using slug, or filename otherwise
I'm new to Hugo so I may be looking in the wrong place, but it seems that when I don't specify custom permalinks for `posts`, it will look for a `slug` in the front-matter, and use the filename if none is found. (Which is great!)

However, I can't seem to specify such a fallback order inside permalinks, because `:slug` will fallback to `:title` instead of `:filename`.

---

There is a [report of this from a while back](https://github.com/gohugoio/hugo/issues/385), which has been closed as "stale". Apologies if this a faux-pas, but I would like to re-open the conversation as a _feature request_. 

I'm migrating my old blog from a personal solution to Hugo, and I have all my posts as `domain.com/post-slug`, and it would be great to use the filename as slug without adding it in the front-matter.

dawidpotocki:
Fixes #4739, #385

It adds a new `:slugelsefile` attribute that behaves the same way as `:slug`
with the only difference being that it fallbacks to filename instead of title.

Both issues are closed because they got marked as stale… but I don't see any
point in opening an another issue.

### DIFF
--- a/docs/content/en/content-management/urls.md
+++ b/docs/content/en/content-management/urls.md
@@ -91,6 +91,9 @@ The following is a list of values that can be used in a `permalink` definition i
 `:slug`
 : the content's slug (or title if no slug is provided in the front matter)
 
+`:slugorfilename`
+: the content's slug (or filename if no slug is provided in the front matter)
+
 `:filename`
 : the content's filename (without extension)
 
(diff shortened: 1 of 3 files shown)


bep:
Yes, this is reasonable, but:

1. I suggest we name it `:slugorfilename`
2. We need a simple test case for it (there should be a test that can be adapted)

dawidpotocki:
1. Done.
2. I now added tests to check if `:slugorfilename` extends to `slug`
when set and also added tests for checking if `:slugorfilename` and
`:slug` properly fallback to filename and title respectively.

bep:
This looks good, thanks.

PR #10071, opened 2022-07-05
Update opengraph.html

stullesha:
Add a 'description' field to avoid "Document does not have a meta description" warning in pagespeed insights and web.dev, especially for themes using the internal template.


coliff (commented):
This isn't a valid Opengraph meta tag so this PR should not be merged.

If you see a '"Document does not have a meta description"' warning then you need to add a description meta tag to your site.

Issue #8594, opened 2021-05-29
Hugo 0.83.1 macOS has a panic for an unknown reason.


jasikpark:
This site built the other day, but today it will not build? I'm not quite sure. The repo in question is https://github.com/jasikpark/jasik-xyz

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.83.1+extended darwin/amd64 BuildDate=unknown
</pre>

### Does this issue reproduce with the latest release?
yes, this is the latest release built from Homebrew.

Hugo Output:
```
hugo --log serve
Start building sites …

                   | EN
-------------------+------
  Pages            | 131
  Paginator pages  |   0
  Non-page files   |  39
  Static files     |  41
  Processed images | 170
  Aliases          |   0
  Sitemaps         |   1
  Cleaned          |   0

Built in 1828 ms
Watching for changes in /Users/calebjasik/Desktop/Github/jasik-xyz/{archetypes,assets,content,layouts,package.json,static}
Watching for config changes in /Users/calebjasik/Desktop/Github/jasik-xyz/config.toml, /Users/calebjasik/Desktop/Github/jasik-xyz/config/development
fatal error: pipe failed

goroutine 1 [running]:
runtime.throw(0x5b2ab88, 0xb)
	runtime/panic.go:1117 +0x72 fp=0xc000b0f890 sp=0xc000b0f860 pc=0x403aad2
runtime.sigNoteSetup(0x6b7d180)
	runtime/os_darwin.go:98 +0xc5 fp=0xc000b0f8b8 sp=0xc000b0f890 pc=0x4036fc5
os/signal.signal_enable(0xc1fb7af500000002)
	runtime/sigqueue.go:228 +0xa5 fp=0xc000b0f8d8 sp=0xc000b0f8b8 pc=0x4070e65
os/signal.enableSignal(...)
	os/signal/signal_unix.go:49
os/signal.Notify.func1(0x2)
	os/signal/signal.go:145 +0x88 fp=0xc000b0f8f8 sp=0xc000b0f8d8 pc=0x549ff68
os/signal.Notify(0xc00398d1a0, 0xc000b0fb38, 0x2, 0x2)
	os/signal/signal.go:165 +0x185 fp=0xc000b0f970 sp=0xc000b0f8f8 pc=0x549f985
github.com/gohugoio/hugo/commands.(*commandeer).serve(0xc0004f2e00, 0xc0000a0900, 0x40, 0x40)
[...]

Issue #6109, opened 2019-07-24
Latest hugo fails in watch mode on "too many open files" on OS X


mattstratton:
Since updating to 0.55.6, when we run `hugo server -w` on our site, we get the following error on OS X:

`ERROR 2019/07/24 09:31:04 Error: listen tcp 127.0.0.1:1313: socket: too many open files`

We have had this happen in the past, and solved it by setting the following:

```
sudo launchctl limit maxfiles 10240 unlimited
sudo sysctl -w kern.maxfiles=65000
```

However, with 0.55.6, this doesn't seem to help. If I run with version `v0.54.0-B1A82C61` it works fine.

For reference, this is the output of the build command (it builds fine, it just can't work in watch mode):
```
                   |  EN
+------------------+------+
  Pages            | 6436
  Paginator pages  |    2
  Non-page files   |   70
  Static files     | 8034
  Processed images |    0
  Aliases          |  462
  Sitemaps         |    1
  Cleaned          |    0
```

To reproduce this, clone the repo at https://github.com/devopsdays/devopsdays-web and run 
`hugo server -w --baseUrl="http://localhost:1313"`

Mac OS Version 10.14.5

bep:
This issue surfaced after Go in 1.12 started to use MacOS' C api to adjust these limits. I did some investigation into this very confusing area that I have now forgotten, but you may look at this thread:

https://github.com/golang/go/issues/30401#issuecomment-480512642

I suspect you can edit /Library/LaunchDaemons/limit.maxfiles.plist and reload that.

The real fix for this issue is, of course, to add a poll based watcher alternative to `fsnotify`. There are implementations ready to use (Docker has one), and there is an open issue about it, but I have been hoping that others would do the work ...

/cc @nathany

mattstratton:
OK, so with some poking around, we came across the following workaround. Running these commands allows it to work now:

```
hugo check ulimit
sudo sysctl -w kern.maxfiles=65536
sudo sysctl -w kern.maxfilesperproc=65536
ulimit -n 65536 65536
```
These changes don't persist past an OS reboot, but at least watch mode works now. I don't know if this issue should stay open, based upon the fact that @bep mentioned there is an existing issue to fit the underlying problem. Feel free to close it if you think that is best.

deanishe:
> The real fix for this issue is, of course, to add a poll based watcher alternative to `fsnotify`

[...]

PR #10098, opened 2022-07-17
add FSEvents-based watcher on macOS

satotake:
On macOS, `fsnotify` uses `kqueue` internally.
This causes some long-lasting issues (#8594, #6109) though we have the workaround for them.

This PR tries to resolve these issues by using `FSEvents` instead of `kqueue`.

After applying it, we can use `--enableFSEvents` flag like so

```
hugo server --enableFSEvents
```

There are some considerations.

- `fsnotify/fsevents` might not be stable yet
- we have to enable CGO if we would apply this as @nathany and @bep pointed.
- ~Related tests look passed but we have to wait for a while just before adding targets to watch.~
    - ~Otherwise, delayed past events are sent.~
    - update: fixed

ref: https://discourse.gohugo.io/t/fsevents-for-watching-on-macos/39053/4

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -208,9 +208,9 @@ type hugoBuilderCommon struct {
 	baseURL     string
 	environment string
 
-	buildWatch bool
-	poll       string
-	clock      string
+	buildWatch     bool
+	poll           string
+	clock          string
 
 	gc bool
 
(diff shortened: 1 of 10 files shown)


bep:
Thanks for this.

I will add some comment, but please don't spend time on adjusting your PR until we decide what to do. I have not tested this myself, but I'm assuming that FSEvents is _significantly better_, and I'm always a fan of that.

Somme comments:

* When we add this, it should be _on_ for all. `--enableFSEvents` is a flag I don't want to explain to people. I suspect that we should get this one battle tested pretty fast, considering there are lots of active MacOS Hugo users out there. If it happens to be a bug for someone, I suggest they fall back to polling.
* This needs to be in the extended build tag (to get `CGO_ENABLED=1`), which would require a small refactor of your 

I will look closer at this later.

nathany:
@arp242 You may want to be aware of this work going on in Hugo.

PR #10111, opened 2022-07-20
externalise and embed livereload.js string

satotake:
This is a small refactoring. `livereload/livereload.go` has a wide
column. Sometimes language server does not work because of it.
Create a new js file and embed it.


bep:
Looks good, reminder for later: Upper case first letter of commit message title (or the first letter after any prefix).

PR #10119, opened 2022-07-23
markup/goldmark/codeblock: get attribute in no language identifier in CodeBlock

### LINKED ISSUE #10118: Cannot get attributes in no language identifier in CodeBlock

<pre>
$ hugo version
hugo v0.101.0+extended darwin/amd64 BuildDate=unknown
</pre>

Yes

### The issue
Sometimes, I don't specify a language identifier in a Code Block as follows:

````markdown
```{foo=true}
bar
```
````

In the above case, the Render Hooks for Code Block do not receive the attributes.
Is this behavior expected?

I guess this cause is that `attrStartIdx` equals 0  if no language identifier in`getAttributes()`.
Ref. https://github.com/gohugoio/hugo/blob/master/markup/goldmark/codeblocks/render.go#L184-L191

If this behavior is a bug, I guess that be fixed by the below diff. 

```diff
-		if attrStartIdx > 0 {
+		if attrStartIdx != -1 {
```
Ref. https://github.com/gohugoio/hugo/commit/b6b15e57602186dcbef0a61c422bdaa9eb295dca

Thanks!

chick-p:
Fixes #10118

### DIFF
--- a/markup/goldmark/codeblocks/integration_test.go
+++ b/markup/goldmark/codeblocks/integration_test.go
@@ -285,7 +285,7 @@ Hello, World!
 -- layouts/_default/single.html --
 {{ .Content }}
 -- layouts/_default/_markup/render-codeblock.html --
-Attributes: {{ .Attributes }}
+Attributes: {{ .Attributes }}|Type: {{ .Type }}|
 `
 
 	b := hugolib.NewIntegrationTestBuilder(
@@ -295,7 +295,7 @@ Attributes: {{ .Attributes }}
 		},
 	).Build()
 
-	b.AssertFileContent("public/p1/index.html", "<h2 id=\"issue-10118\">Issue 10118</h2>\nAttributes: map[foo:bar]")
+	b.AssertFileContent("public/p1/index.html", "<h2 id=\"issue-10118\">Issue 10118</h2>\nAttributes: map[foo:bar]|Type: |")
 }
 
 // Issue 9571
--- a/markup/goldmark/codeblocks/integration_test.go
+++ b/markup/goldmark/codeblocks/integration_test.go
@@ -265,6 +265,39 @@ Position: {{ .Position | safeHTML }}
 	b.AssertFileContent("public/p1/index.html", filepath.FromSlash("Position: \"/content/p1.md:7:1\""))
 }
 
+// Issue 10118
+func TestAttributes(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- config.toml --
+-- content/p1.md --
+---
+title: "p1"
+---
+
+## Issue 10118
+
+§§§ {foo="bar"}
+Hello, World!
+§§§
+
+-- layouts/_default/single.html --
+{{ .Content }}
+-- layouts/_default/_markup/render-codeblock.html --
+Attributes: {{ .Attributes }}
+`
+
+	b := hugolib.NewIntegrationTestBuilder(
+		hugolib.IntegrationTestConfig{
+			T:           t,
+			TxtarString: files,
+		},
+	).Build()
+
+	b.AssertFileContent("public/p1/index.html", "<h2 id=\"issue-10118\">Issue 10118</h2>\nAttributes: map[foo:bar]")
+}
+
 // Issue 9571
 func TestAttributesChroma(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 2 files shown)


markup/goldmark/codeblocks/integration_test.go
```diff
@@ -265,6 +265,39 @@ Position: {{ .Position | safeHTML }}
...
+## Issue 10118
+
+§§§ {foo="bar"}
+Hello, World!
+§§§
+
+-- layouts/_default/single.html --
+{{ .Content }}
+-- layouts/_default/_markup/render-codeblock.html --
+Attributes: {{ .Attributes }}
```
bep, line 288: Could you

* Add a check for the `Type` and make sure it's empty.
* Get the commit message title in line with the guidelines ("get" => "Get")

chick-p: @bep 
I'd fixed the commit message, sorry.

I have a question.
I tried to add to check `Type`, but the `Type` returns Attributes.


Because the parser returns a string of after the code fence if a language identifier is empty.
https://github.com/yuin/goldmark/blob/master/ast/block.go#L273-L276

Would it be better to resolve the problem on this pull request or another Issue?

bep: I'm not sure I follow, it looks like a string to me:

https://github.com/gohugoio/hugo/blob/9c24b86e4b634d7db761a3e57c1b64fead880aff/markup/goldmark/codeblocks/render.go#L158

chick-p: Sorry. The order of my comment was not good.

After adding `.Type`, a result of `TestAttributes()` as follows:
```
--- FAIL: TestAttributes (0.04s)
    /Users/chick-p/Repos/hugo/markup/goldmark/codeblocks/integration_test.go:298: 
        error:
          no substring match found
        comment:
          <h2 id="issue-10118">Issue 10118</h2>
          Attributes: map[foo:bar]|Type: |
        container:
          "<h2 id=\"issue-10118\">Issue 10118</h2>\nAttributes: map[foo:bar]|Type: {foo=&#34;bar&#34;}|"
        want:
          "Attributes: map[foo:bar]|Type: |"
              /Users/chick-p/Repos/hugo/hugolib/integrationtest_builder.go:144
            s.Assert(content, qt.Contains, match, qt.Commentf(m))
              /Users/chick-p/Repos/hugo/markup/goldmark/codeblocks/integration_test.go:298
            b.AssertFileContent("public/p1/index.html", "<h2 id=\"issue-10118\">Issue 10118</h2>\nAttributes: map[foo:bar]|Type: |")
FAIL
FAIL	github.com/gohugoio/hugo/markup/goldmark/codeblocks	0.709s
FAIL
```
`.Type` became `Type: {foo=&#34;bar&#34;}`, so it is not empty.

I Will have the same result even if `attrStartIdx > 0` at L191 in `markup/goldmark/codeblocks/render.go`.

I suppose that the cause is https://github.com/gohugoio/hugo/pull/10119#discussion_r928178328

bep: >.Type became Type: {foo=&#34;bar&#34;}, so it is not empty.

It should be. 

This may be an existing issue, but it feels natural to fix as part of this PR. I suspect the simplest fix is to check for `{` where `Type` is evaluated. I think we can safely make that a reserved character not to be used in `Type`.

chick-p: Thanks for considering the direction.
I have checked for `{`.

markup/goldmark/codeblocks/render.go
```diff
@@ -156,7 +156,13 @@ func (c *codeBlockContext) Page() any {
 }
 
 func (c *codeBlockContext) Type() string {
-	return c.lang
+	i := 0
+	for ; i < len(c.lang); i++ {
```
bep, line 160: Thank, this looks correct, but:

* This needs to be evaluated when codeBlockContext is created. Everyone who later looks at `c.lang` will assume that it contains the `lang` only.
* The loop can be replaced with `bytes.Cut`.

bep (changes requested):

markup/goldmark/codeblocks/render.go
chick-p: Got it. 
I updated the PR again.

- `codeBlockContext` is created at L105. 
I guess that a lang passed to a renderer in L74 looks good be the lang without attributes, too.
- I used `strings.Cut` because both `lang` and  `{` are strings.
If this case is suitable to use `bytes.Cut` as you said, would you like to teach me the reason?

chick-p:
rebase master

PR #10154, opened 2022-08-05
Upgrade dependencies to avoid checksum error for 0.94.x release

alvinlin123:
Hi there

I am one of the [Cortex project](https://github.com/cortexproject/cortex) maintainers, and I am trying to upgrade Hugo from v0.72.0 to v0.94.2, but I hit the problem during build because v0.94.2 depends on [go-toml v2.0.0-beta.6](https://github.com/pelletier/go-toml/releases/tag/v2.0.0-beta.6), which has issue with Go module checksum. 

I can't upgrade to the latest Hugo yet because upgrading Cortex to Go 1.18 will take some time, and my current goal is to upgrade Hugo so that I can use the Google analytics V4 support. So I am attempting to upgrade to Hugo 0.94.X, which is latest version before Go 1.18 upgrade.

So it would be nice if we can do a 0.94.3 release to fix the go-toml issue. I am upgrading go-toml to v2.0.0-beta.7 and not the official v2.0.0 to minimize changes; from go-toml's release note v2.0.0-beta.7's purpose is to only fix the checksum error.

Similar story for localescompressed package.

I am not sure what the release process is, but let me know if I need to do anything differently.

Signed-off-by: Alvin Lin <alvinlin@amazon.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/gohugoio/go-i18n/v2 v2.1.3-0.20210430103248-4c28c89f8013
 	github.com/gohugoio/locales v0.14.0
-	github.com/gohugoio/localescompressed v0.15.0
+	github.com/gohugoio/localescompressed v0.16.0
 	github.com/gohugoio/testmodBuilder/mods v0.0.0-20190520184928-c56af20f2e95
 	github.com/google/go-cmp v0.5.7
 	github.com/gorilla/websocket v1.5.0
--- a/go.mod
+++ b/go.mod
@@ -45,7 +45,7 @@ require (
 	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646 // indirect
 	github.com/niklasfasching/go-org v1.6.2
 	github.com/olekukonko/tablewriter v0.0.5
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this report, I will merge when the test goes green.

alvinlin123:
@bep Sorry, I did not update the go.sum when I published the PR. Sorry about that. Can you run the workflows again see if tests are passing? Thanks!

bep:
OK, I didn't notice this is for an older release. Hmm. Well, it couldn't hurt fixing this, but that branch will not live forever.

alvinlin123:
Understood :-) I am hoping I can upgrade to Go 1.18 in near future so that I can use latest version of Hugo. So as long as the branch lives long enough then it’s all good. I am already glad that you responded to this PR. 

The test failures doesn’t seem to be related to change in this PR, but I will take a deeper look after weekend.

bep:
That test failure isn't something I can merge into an old release branch. You could update go-i18n to a even newer version, but I suggest you do that in your own fork.

alvinlin123:
Yea I think it's easier if I just do this in my fork for now. I plan upgrade to Go 1.18 as well as latest version of Hugo anyways :) Thanks for taking a look @bep !

PR #10155, opened 2022-08-07
livereload: Inject script without head or body tag

### LINKED ISSUE #10105: Print a warning when LiveReload fails to inject its code
There are more than a few posts on the Hugo Discourse over the years through recent times of users having LiveReload not work in `hugo server`. The cause seems to be their templates missing either the `<head>` or the `<body>` tags. As @davidsneighbour has pointed out, HTML5 doesn't require these tags (or the `<html>` tag).

I feel it would make LiveReload more foolproof if the function would inject those html tags in order to inject the LiveReload script if they are missing, instead of failing silently without the user knowing why it isn't working. Or at least a WARNING in the console saying that those tags were not found and so LiveReload was not enabled. Maybe both. The first covers the fact that it'll just work for any user regardless of theme or experience and result in a good user experience (not having anything silently not working as expected) and the second teaches users that they are missing typical tags

I experienced this very thing today and was about to make a post on Discourse about it but was able to parse through the posts and find the issue. The problem is if you "Inspect Page" with DevTools, those missing html tags are injected by DevTools and so it looks like they are there, but you have to go to "View Source" to see that in fact they don't exist in the html.

satotake:
Currently, Hugo does not inject `livereload` script if html does not contain `<head>` or `<body>`.
This sometimes happens if you create new sites without `theme` and it is hard to catch the cause soon.

This PR:

* Inject livereload script even if html does not include `<head>`, `<body>`, or `<html>`
    - Modern browsers execute scripts even if they are outside `<html>`
    - Some js frameworks (confirmed with vite) inject HRM script without `<html>` tag
* Append warning script to html if `<head>` or `<body>` is not in html
* Fix bug that livereload cannot be appended to the tags with attrs

Close #10105

### DIFF
--- a/transform/livereloadinject/livereloadinject.go
+++ b/transform/livereloadinject/livereloadinject.go
@@ -24,16 +24,24 @@ import (
 	"github.com/gohugoio/hugo/transform"
 )
 
+const warnMessage = `"head" or "body" tag is required in html to append livereload script. ` +
+	"As a fallback, Hugo injects it somewhere but it might not work properly."
+
+var warnScript = fmt.Sprintf(`<script data-no-instant defer>console.warn('%s');</script>`, warnMessage)
+
 type tag struct {
 	markup       []byte
 	appendScript bool
+	warnRequired bool
 }
 
 var tags = []tag{
-	{markup: []byte("<head>"), appendScript: true},
-	{markup: []byte("<HEAD>"), appendScript: true},
+	{markup: []byte("<head"), appendScript: true},
+	{markup: []byte("<HEAD"), appendScript: true},
 	{markup: []byte("</body>")},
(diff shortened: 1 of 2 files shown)


hueralin:
Is this problem solved, why am I still having this problem? Hugo version is hugo v0.111.3-5d4eb5154e1fed125ca8e9b5a0315c4180dab192 windows/amd64 BuildDate=2023-03-12T11:40:50Z VendorInfo=gohugoio

hueralin:
warn: "head" or "body" tag is required in html to append livereload script. As a fallback, Hugo injects it somewhere but it might not work properly.

Is it possible that I need to add head, body tags to every page, even the page files in the partials directory?

Issue #9996, opened 2022-06-12
Date format / timezone issue after remote build


willduncanphoto:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
0.100.2
</pre>

### Does this issue reproduce with the latest release?
Yes.

When attempt to use the timezone in a `.Date.Format` for a post/page, it works locally, but when deployed, it displays the timezone as `-0700` no matter what.  And when I use just `{{ .Date }}` it displays `-0700 -0700`.

`{{ .Date.Format "January 2, 2006 3:04 PM MST" }}` and `{{ .Date }}` return the following on both Cloudflare and Netlify.

<img width="544" alt="Screenshot_20220612T005802@2x" src="https://user-images.githubusercontent.com/2657/173223289-a73accf8-a7f0-4e2c-b5f5-0d1885e651d9.png">

Locally it spits out `June 11, 2022 at 2:06 AM PDT // 2022-06-11 02:06:56 -0700 PDT` as it should.

Are there logs I can track to figure out the cause? As far as I can tell, it's a bug of some kind, but that both netlify and pages suffer from? is it a timezone/config issue on my end and I'm crazy?

bep:
So,

This code:

```
{{ .Date.Format "January 2, 2006 3:04 PM MST" }}
```

Uses the local timezone on whatever server it's running on for any zone settings (offset, name). This is unfortunate, but not something Hugo can do anything about, which is why we added a format func:

```
{{ .Date | time.Format "January 2, 2006 3:04 PM MST" }}
```

The above will, if you have set a `timeZone` in your Hugo config (can be set per language), give more consistent results.

I suggest we take further trouble shooting at:

https://discourse.gohugo.io/

willduncanphoto:
Thanks. I applied as you suggested, same results. That is a much better solution though, so thank you!

The timezone / fomatting are correct on the posts after the build on Cloudflare. I assumed that was set by the timezone set in the post front matter so I didn't need timezone in the config. Since I'm just trying to display the PST / PDT with the post time. I'm still not the most savvy Hugo user.

Local: 

<img width="326" alt="Screenshot_20220612T031435@2x" src="https://user-images.githubusercontent.com/2657/173228481-8491b57d-d4bb-4a81-8226-6ccb2927b9d9.png">

On Cloudflare/Netlify builds:

[...]

PR #10157, opened 2022-08-07
time: Fix toml zoneinfo inconsistency

satotake:
TOML's time representation has known issues about zoneinfo name such as `PDT` in hugo

- https://github.com/gohugoio/hugo/pull/9082
- https://github.com/gohugoio/hugo/pull/9982
- https://github.com/gohugoio/hugo/issues/8895
- https://github.com/gohugoio/hugo/issues/9996

These are caused because

* TOML's time format has only offset
    - zoneinfo cannot be retrieved with offset value
* `TZ` environment variable can change the behaviour
* `cast.ToTimeInDefaultLocationE` may or may not use `location` arg

Given that hugo often is used on both local and remote machine, and as #9996 said hugo supports multilingual site generation,

* `TZ` envvar should not have any effect on this
   - Remote environments can be hard to debug
   - **UPDATE: If the site config does not have timezone, use `time.Local` (exception)**
       - see [the comment](https://github.com/gohugoio/hugo/pull/10157#issuecomment-1224537293)
* If only `timeZone` is set explicitly in config file and offsets of
  time are matched with it, hugo should show zoneinfo
   - Otherwise, zoneinfo should not be displayed
* hugo should handle TOML's time object and time string equally

To meet these,

* assure to set `location` derived from conf to time of TOML if possible
* add tests

Close https://github.com/gohugoio/hugo/issues/9996


bep:
@satotake thanks a lot for this. The code looks good to me, even if this "time landscape" is confusing me pretty hard.

@jmooring would you have time to take this PR for a spin and see if this fix makes sense to you?

jmooring:
Sorry, but I don't understand the commit message "time: Fix toml zoneinfo inconsistency". This PR seems to be more about making this configuration work by ignoring the OS tzdata:

```
[languages.en]
languageCode = 'en-US'
languageName = 'English'
timeZone = 'America/Los_Angeles'
weight = 1

[languages.es]
languageCode = 'es-ES'
languageName = 'Espanol'
timeZone = 'Europe/Madrid'
weight = 2
```

Most sites have not set `timeZone` in site configuration, so this is a breaking change. The time zone abbreviations that are visible in < v0.101.0  will not appear after this change unless site authors add `timeZone` to their site configuration. I'm fine with that, but we need to let people know.

bep:
>Most sites have not set timeZone in site configuration, so this is a breaking change.

We could possibly apply this fix only when the locale is established from config. Not sure.

satotake:
> Most sites have not set timeZone in site configuration, so this is a breaking change

> We could possibly apply this fix only when the locale is established from config. Not sure.

It was my intention to show no zone info if config does not have `timeZone` (considering remote/local site building).
But I changed that. If the site config has no `timeZone`, try to use `time.Local` timezone.

PR #10158, opened 2022-08-08
Clarify that blocks in partials are unsupported

fj:
The documentation doesn't currently say one way or the other whether block definitions inside templates are permissible. This clarifies that they are not.


fj:
> I think that this makes it more confusing in some ways. You can use `blocks` in partials, just not override them from the layout templates. Blocks are still useful in partials to define templates and execute them in place. I think if you clarify this better it could be helpful, but it isn't clearly defined what the limitations are and why.

That's good feedback, thanks @mcexit. If there's agreement from a committer on this front, I can update the PR accordingly.

jmooring:
Please submit issues and PRs related to documentation here:
<https://github.com/gohugoio/hugoDocs>

PR #10178, opened 2022-08-15
Add cleanDestinationDir & disableFastRender options.

nfriedli:


bep:
Thanks,

These 2 are defined as flags and documented in the CI section. I'm pretty sure we don't duplicate other flags into the config section, even if most flags can also be set as config.

https://gohugo.io/commands/hugo/

Also, we try to keep docs change not directly related to code changes in the hugoDocs repo:

https://github.com/gohugoio/hugoDocs

PR #10237, opened 2022-08-31
Add `--force` to `hugo new`

### LINKED ISSUE #9243: force flag for hugo new (not hugo new site)
We already have `--force` for `hugo new site` but it seems this flag is not available for `hugo new` to create content.

I have looked at all the issues there and I can not see any relevant issue or PR:
https://github.com/gohugoio/hugo/search?p=15&q=force&type=issues

I have the following usecase:

A specific content type as template. `ENV_VAR=value hugo new my-new-file-name -k my-template-type` to populate the frontmatter with custom keys and values and create the new file.

But if I want to update the content in a specific environment I want to run `ENV_VAR=other-value hugo new my-new-file-name -k my-template-type` to recreate it for compilation.

And if my template-file changes, then I want to just run it with the old `ENV_VAR=value hugo new my-new-file-name -k my-template-type` but get the new template in the resulting file.

Currently this does not work as Hugo says "file-name already exists".
Even the `--force` option for sites does not overwrite files but only folders. See https://github.com/gohugoio/hugo/blob/master/commands/new_site.go#L89 (seems to only add new files but not overwrite existing files).

Maybe such an option to overwrite files should be added for the site command too.

My current workaround is to do a `rm file-name` and then run `ENV_VAR=other-value hugo new my-new-file-name -k my-template-type` again.

satotake:
Closes #9243

### DIFF
--- a/commands/new.go
+++ b/commands/new.go
@@ -32,6 +32,7 @@ var _ cmder = (*newCmd)(nil)
 type newCmd struct {
 	contentEditor string
 	contentType   string
+	force         bool
 
 	*baseBuilderCmd
 }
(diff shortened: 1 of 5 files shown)


commands/new.go
```diff
@@ -54,6 +55,7 @@ Ensure you run this within the root directory of your site.`,
 
 	cmd.Flags().StringVarP(&cc.contentType, "kind", "k", "", "content type to create")
 	cmd.Flags().StringVar(&cc.contentEditor, "editor", "", "edit new content with this editor, if provided")
+	cmd.Flags().BoolVarP(&cc.force, "force", "f", false, "allow to override file")
```
bep, line 58: I think it should say "overwrite file if it already exists".

PR #10250, opened 2022-09-06
tpl: improve youtube shortcode through use of oEmbed data

jhenstridge:
This PR implements some improvements to the youtube shortcode. This covers some of the suggestions from #3694, but rather than implementing everything I looked at what I could accomplish using YouTube's [oEmbed](https://oembed.com/) API endpoint and no additional shortcode arguments.

Unlike the main YouTube API, the oEmbed endpoint doesn't require any API keys or secrets. Using this API, I was able to add the following:

1. set the default value for the iframe title to the video's title on YouTube.
2. determine the video's aspect ratio, and style the iframe to match.
3. get a URL for the video thumbnail image, which I used to implement a "simple" mode modeled after the same mode in the Vimeo shortcode.

(3) is potentially useful for websites concerned about GDPR compliance. The thumbnail URL provided by YouTube uses the `i.ytimg.com`, which Google doesn't seem to be setting any cookies on at present.

The provided thumbnail is the "hq" version. For many videos there is a corresponding "maxres" thumbnail with a higher resolution, but the oEmbed API doesn't tell us whether it'll be usable: for older SD videos, the maxres thumbnail URL will return a placeholder image rather than the actual thumbnail.

### DIFF
--- a/config/privacy/privacyConfig.go
+++ b/config/privacy/privacyConfig.go
@@ -98,6 +98,11 @@ type YouTube struct {
 	// YouTube won’t store information about visitors on your website
 	// unless the user plays the embedded video.
 	PrivacyEnhanced bool
+
+	// If simple mode is enabled, only a thumbnail is fetched from i.ytimg.com and
+	// shown with a play button overlaid. If a user clicks the button, he/she will
+	// be taken to the video page on youtube.com in a new browser tab.
+	Simple bool
 }
 
 // DecodeConfig creates a privacy Config from a given Hugo configuration.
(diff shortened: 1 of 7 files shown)


jhenstridge:
The `getJSON` call is executed while building the site, rather than when someone visits the website. The generated HTML will link to an external image on `i.ytimg.com`. It behaves pretty much the same as the vimeo shortcode's simple mode.

jhenstridge:
@mcexit: at present, the Instagram, Twitter, and Vimeo shortcodes all use `getJSON` to retrieve information about the embedded content. So doing the same for YouTube does not significantly change that.

As for privacy/build performance, Hugo caches the output of getJSON calls. By default these go to the /tmp directory. You can move the cache into your site data in the configuration file if you want more control, or to make cached responses available to a firewalled build machine:

https://gohugo.io/getting-started/configuration/#configure-file-caches

PR #10264, opened 2022-09-13
feat: builtin iso8601 supports

septs:


bep:
iso8601 isn't the same as RFC3339. They have added support for it in an unreleased version of Go: 

https://github.com/golang/go/commit/d2a835fbf3f22ca3b3e5fb51c8c2756c8a7b2550

jhenstridge:
That commit is from 2009. That format code appears to have been removed in 2010 in https://github.com/golang/go/commit/6c124cb879cb7b6613d0d8bcec316e271ded3ee5, both of which are from before Go 1.0.

RFC 3339 describes itself as "a profile of the ISO 8601": that is, all RFC 3339 formatted dates are valid ISO 8601 but not all ISO 8601 dates are valid RFC 3339.

If you are formatting date strings, then it should be safe to use the RFC3339 format string.

bep:
@jhenstridge OK, but there is a recent relevant discussion on the Go's issue tracker.

My point is:

* We use currently have this construct `{{- $iso8601 := "2006-01-02T15:04:05-07:00" -}}`
* RFC3339     = "2006-01-02T15:04:05Z07:00"
* The two strings above are not the same, which means we cannot merge your proposed change.

I would rather wait for Go to come up with some `iso8601` definitions before we create further confusion.

jhenstridge:
The "Z" is part of the time zone format string, rather than a literal Z character. From [the docs](https://pkg.go.dev/time):

> Replacing the sign in the format with a Z triggers the ISO 8601 behavior of printing Z instead of an offset for the UTC zone.

It seems to produce output I'd recognise as ISO 8601: https://go.dev/play/p/Hr7RdQ9VBsg

septs:
@bep rename `:iso8601` to `:rfc3339`?

PR #10305, opened 2022-09-21
hugofs: Fix glob case-sensitivity bug

satotake:
On Linux, `hugofs.Glob` does not hit any directories which include uppercase letters. (This does not happen on macOS.)

Since `resources.GetMatch/Match` uses `Glob`,

```
{{ resources.GetMatch "FOO/bar.css" }}
```

this does not match `assets/FOO/bar.css` .

On the other hand, you can get it with

```
{{ resources.Get "FOO/bar.css" }}
```

### DIFF
--- a/hugofs/glob.go
+++ b/hugofs/glob.go
@@ -26,14 +26,14 @@ import (
 // Glob walks the fs and passes all matches to the handle func.
 // The handle func can return true to signal a stop.
 func Glob(fs afero.Fs, pattern string, handle func(fi FileMetaInfo) (bool, error)) error {
-	pattern = glob.NormalizePath(pattern)
+	pattern = glob.NormalizePathCaseSensitive(pattern)
 	if pattern == "" {
 		return nil
 	}
 
-	g, err := glob.GetGlob(pattern)
+	g, err := glob.GetFilenamesGlob(pattern)
 	if err != nil {
-		return nil
+		return err
 	}
 
 	hasSuperAsterisk := strings.Contains(pattern, "**")
--- a/hugofs/glob/glob.go
+++ b/hugofs/glob/glob.go
@@ -30,15 +30,13 @@ const filepathSeparator = string(os.PathSeparator)
 var (
 	isWindows        = runtime.GOOS == "windows"
 	defaultGlobCache = &globCache{
-		isCaseSensitive: false,
-		isWindows:       isWindows,
-		cache:           make(map[string]globErr),
+		isWindows: isWindows,
+		cache:     make(map[string]globErr),
 	}
 
 	filenamesGlobCache = &globCache{
-		isCaseSensitive: false, // As long as the search strings are all lower case, this does not allocate.
-		isWindows:       isWindows,
-		cache:           make(map[string]globErr),
+		isWindows: isWindows,
+		cache:     make(map[string]globErr),
 	}
 )
 
@@ -49,8 +47,7 @@ type globErr struct {
 
 type globCache struct {
 	// Config
-	isCaseSensitive bool
-	isWindows       bool
+	isWindows bool
 
 	// Cache
 	sync.RWMutex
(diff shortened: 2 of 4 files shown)


hugofs/glob.go
```diff
@@ -55,7 +55,7 @@ func Glob(fs afero.Fs, pattern string, handle func(fi FileMetaInfo) (bool, error
 			return nil
 		}
 
-		if g.Match(p) {
+		if g.Match(strings.ToLower(p)) {
```
bep, line 58: I have a few comments:

* There is a filenamesGlobCache in the glob package that should be used for this (it's used in other similar situations, not sure why not here).
* I do notice, however, that I the `isCaseSensitive` is always set to `false`, and I don't remember why... Changing that value without doing additional checks sounds dangerous.

**Using upper case in filename is something I would highly discourage, but I suspect that comes from the Windows world.**

What I suggest is this, and please arrest me if I'm thinking wrong:

* In globCache.GetGlob we check if the pattern contains any upper case chars.
* If so, we return a `globDecorator` with 2 globs (one all lower case)
* The match Match would then do one or two checks and return the first positive match.
* With that we can remove the isCaseSensitive flag which isn't in use anyway.


Also, if the /commands package tests keep failing on GitHub, do a if htesting.IsCi skip so we can get the tests here green.

bep: Re. failing test, see https://github.com/gohugoio/hugo/commit/08f0984f918bb97d35c1765d3b94b1a9a1a3b9cd

satotake: > Using upper case in filename is something I would highly discourage, but I suspect that comes from the Windows world.

I want to say I am in the same boat ... but, you know, we are sometimes in the undesirable situation against our will.

Thank you for your suggestion. I would follow it mostly.

hugofs/glob.go
```diff
@@ -31,9 +31,24 @@ func Glob(fs afero.Fs, pattern string, handle func(fi FileMetaInfo) (bool, error
 		return nil
 	}
 
-	g, err := glob.GetGlob(pattern)
-	if err != nil {
-		return nil
+	hasUpperCase := pattern != strings.ToLower(pattern)
```
bep, line 34: So, this looks logically correct for this particular case, but one of my motivations behind my previous comment about this was to consolidate the "filename matching" to make sure that it behaved the same in the different places.

Which means that `glob.GetFilenamesGlob` should return an interface that does the double checking (see the struct that is put in filenamesGlobCache).

satotake: Added new interface.
Is this ok?

hugofs/glob.go
```diff
@@ -26,14 +26,14 @@ import (
...
 // The handle func can return true to signal a stop.
 func Glob(fs afero.Fs, pattern string, handle func(fi FileMetaInfo) (bool, error)) error {
-	pattern = glob.NormalizePath(pattern)
+	pattern = glob.NormalizePathCaseSensitive(pattern)
 	if pattern == "" {
 		return nil
 	}
 
-	g, err := glob.GetGlob(pattern)
+	g, err := glob.GetFilenamesGlob(strings.ToLower(pattern))
```
bep, line 34: The lower here doesn't look right.

hugofs/glob/glob.go
```diff
@@ -117,14 +107,50 @@ func (g globDecorator) Match(s string) bool {
 	return g.g.Match(s)
 }
 
+type globDecoratorDouble struct {
+	primary   glob.Glob
+	secondary glob.Glob
+}
```
bep, line 113: I think renaming the above to `lowerCase` and `originalCase` would make it easier to understand.

bep (changes requested):

bep (approved):

bep:
OK, looks good now, but GitHub says I cannot merge due to conflicts (not sure what those are); can you rebase/squash this PR?

bep:
Merged.

satotake:
Sorry for delayed response
Thanks for cleaning up the history

PR #10344, opened 2022-10-02
concatenate newline to UsageString

anonyindian:
concatenate newline to UsageString instead of initiating writer twice.


jmooring:
Superseded by https://github.com/gohugoio/hugo/pull/10602

PR #10374, opened 2022-10-12
preallocate memory

florianl:
**What this PR does / why we need it:**

Preallocate memory instead of enforcing an incremental growth. This will result in less work for the garbage collector.

### DIFF
--- a/hugolib/shortcode.go
+++ b/hugolib/shortcode.go
@@ -248,12 +248,12 @@ func (sc shortcode) String() string {
 	switch v := sc.params.(type) {
 	case map[string]any:
 		// sort the keys so test assertions won't fail
-		var keys []string
+		keys := make([]string, 0, len(v))
 		for k := range v {
 			keys = append(keys, k)
 		}
 		sort.Strings(keys)
-		tmp := make(map[string]any)
+		tmp := make(map[string]any, len(keys))
 
 		for _, k := range keys {
 			tmp[k] = v[k]


florianl:
sorry for the delayed response @bep :pray: 

As the change is trivial, I did not add a benchmark. But I wanted to share the benefits I see here with this proposed change.
For benchmarking I used the following snippet:

```
func BenchmarkShortcodeString(b *testing.B) {
	var sc shortcode
	// Populate sc
	params := make(map[string]any)
	for i:= 0; i<16; i++ {
		params[fmt.Sprintf("%d", i)] = struct{}{} 
	}
	sc.params = params

	b.ResetTimer()
	for i:=0; i< b.N; i++ {
		tmp := sc.String()
		_ = tmp
	}
}
```

With the original code these are the results I get on my machine:
```
goos: linux
goarch: amd64
pkg: github.com/gohugoio/hugo/hugolib
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkShortcodeString-16    	   51668	     22555 ns/op	    4144 B/op	      46 allocs/op
PASS
```

With the proposed change I get these results for the benchmark run:
```
goos: linux
goarch: amd64
pkg: github.com/gohugoio/hugo/hugolib
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkShortcodeString-16    	   56737	     20946 ns/op	    2962 B/op	      40 allocs/op
PASS
```

So even for a map with only 16 elements we can see a reduction of allocs/op and ns/op. I can't tell if 16 elements in this map are common so this number is just a random pick. With more elements in the map the drop in allocs/op and ns/op will be even greater.

PR #10401, opened 2022-10-27
feat: time.Format takes an optional 3rd parameter for language

gjvnq:
This feature allows the `time.Format` template feature to take an optional 3rd parameter for the time formatting language.

```
{{ time.Format ":date_medium" $date }} <!-- will use the page language -->
{{ time.Format ":date_medium" $date "pt" }} <!-- will use Portuguese (pt) -->
```

I added this feature because of citations. I needed a way to write the dates in Portuguese even for articles in English.

### DIFF
--- a/docs/content/en/functions/dateformat.md
+++ b/docs/content/en/functions/dateformat.md
@@ -9,17 +9,21 @@ menu:
   docs:
     parent: "functions"
 keywords: [dates,time,strings]
-signature: ["time.Format LAYOUT INPUT"]
+signature: ["time.Format LAYOUT INPUT [LANGUAGE]"]
 workson: []
 hugoversion:
 relatedfuncs: [Format,now,Unix,time]
 deprecated: false
 ---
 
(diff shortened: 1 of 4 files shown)


bep:
Thanks for this. It looks mostly correct (there is a data race in there somewhere ...), but we need to think a little harder about the API before we commit to this change. Adding the language string to time.Format feels a little arbitratry.

Currently we have these language (Site) aware funcs:

* langs.Translate
* time.Format

(I think I got that right)

I suspect that instead of adding the language string to every relevant func (and then reimplement a cache for every variant) seems sub optimal.

I'm thinking out loud here, so please don't implement this, but I suspect it would make more sense to have a way to create a a language so you could do:

* site.Language.Translate (which would be the same as the template func langs.Translate/i18n)
* {{ $lang := langs.New "en" }}{{ $lang.Translate "foo bar" }}

Etc.

gjvnq:
I liked your idea of giving access to a language object.

PR #10416, opened 2022-11-04
Add support for Google Adsense

WooilJeong:
hello. Added content related to Google Adsense. Please review. Thank you.

### DIFF
--- a/tpl/tplimpl/embedded/templates/google_adsense.html
+++ b/tpl/tplimpl/embedded/templates/google_adsense.html
@@ -1,0 +1,4 @@
+{{- with .Site.Params.googleAdsense -}}
+<script async src="https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js?client={{ .Site.Params.googleAdsense }}"
+     crossorigin="anonymous"></script>
+{{- end -}}


tpl/tplimpl/embedded/templates/google_adsense.html
```diff
@@ -0,0 +1,4 @@
+{{- with .Site.Params.googleAdsense -}}
+<script async src="https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js?client={{ .Site.Params.googleAdsense }}"
+     crossorigin="anonymous"></script>
```
septs, line 3:
```suggestion
{{- $qs := querify "client" . }}
<script async defer src="https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js?{{ $qs }}" crossorigin="anonymous"></script>
```

1. `{{ with pipeline }}`
   If the value of the pipeline is empty, no output is generated;
   otherwise, dot is set to the value of the pipeline and T1 is executed.
   see https://pkg.go.dev/text/template#hdr-Actions
2. `querify` safety build query string
3. allow `defer` loading script, non blocking page

WooilJeong: @septs Thank you for your kind review of my PR. I think you've edited it in a good way.

PR #10422, opened 2022-11-07
Add deprecation notice

adamroyjones:


adamroyjones:
Ugh, opened in the wrong repository because of some misconfiguration on my end. Closing—please ignore.

PR #10433, opened 2022-11-11
Editing the README overview for a change made earlier

vgnh:
I should have edited the README content overview earlier, when I made the change.
Apologies, my bad!


bep:
Thanks for this. I would, however, ask you to rework your GitHub workflow when creating these pull request. I could squash this (notice that this PR contains 2 unrelated commits) on my end and edit the final commit message to get it right, but we might as well do it right to save us some future work:

When you start editing to fix something in your forked repository, create a new branch first. Then create a PR from that branch. This is also the default behaviour you get when you hit the "edit button" directly here on GitHub.

vgnh:
I'll do that, thank you for the pointers!

PR #10455, opened 2022-11-19
common/htime: add `:date_time`, `:date_only`, `:time_only`

Umio-Yasuno:
This PR adds `:date_time ("2006-01-02 15:04:05")`, `:date_only ("2006-01-02")`, `:time_only ("15:04:05")` to the custom layouts.   

Those are popular layouts and will be added to golang's time pkg as well.

Reference: https://github.com/golang/go/commit/0981d9fff1fa5601a2b7833473955252425bd923

### DIFF
--- a/common/htime/time.go
+++ b/common/htime/time.go
@@ -107,6 +107,10 @@ func (f TimeFormatter) Format(t time.Time, layout string) string {
 			return f.ltr.FmtDateMedium(t)
 		case "date_short":
 			return f.ltr.FmtDateShort(t)
+		case "date_time":
+			return t.Format("2006-01-02 15:04:05")
+		case "date_only":
+			return t.Format("2006-01-02")
 		case "time_full":
 			return f.ltr.FmtTimeFull(t)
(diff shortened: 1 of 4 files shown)


bep:
I'm not sure about this. Adding it as constants to the Go stdlib makes sense, adding them to Hugo ... Not sure. The thing is, the set of constants we have now adds date "layouts" with locale support not possible to do with Go's layouts. To pick one of your examples above, `:date_only` => `2006-01-02`. That layout would not make much sense in Norway. So, if we're adding more of these I think they need to be localized somehow.

Umio-Yasuno:
Thank you for your reply.

> To pick one of your examples above, :date_only => 2006-01-02. That layout would not make much sense in Norway.

However, in some countries it is a common layout.  
I think adding popular layouts is an advantage for Hugo themes that offer custom layout settings and its developer and the users of its theme.

> So, if we're adding more of these I think they need to be localized somehow.

That was the easiest part to add.  
Do we need some kind of constant?

Issue #8601, opened 2021-06-01
Hugo crashes when adding invalid data file to data directory.


eazolan:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.83.1-5AFE0A57 windows/amd64 BuildDate=2021-05-02T14:38:05Z VendorInfo=gohugoio
</pre>

### Does this issue reproduce with the latest release?
Yes.

Hugo server crashes when:
1. You use JSON files in your data directory
2. AND you add a new 0 byte JSON file, or change one of the JSON files to 0 bytes

When you add a new 0 byte file, you don't even have to reference it in your code. I think Hugo is evaluating it for validity.

I've gone over this on the Hugo discourse page with another: https://discourse.gohugo.io/t/hugo-crashes-when-theres-a-data-file-with-no-data-in-it/33189/4

FrappeFortyTwo:
Hey, I've made a PR for this issue [here](https://github.com/gohugoio/hugo/pull/8603). Looking for feedback.

P.S It's my first contribution here.

ReCodee:
is this issue still relevant?
If yes, then I'd like to work on this issue

PR #10526, opened 2022-12-12
Add empty JSON data file as empty map

acclassic:
This fixes issue #8601.

Test Case:
1. Create a new Hugo Project or use an existing one
2. Add an empty JSON file to the data folder

Result:
The site will not crash even when the data is accessed in the layout files like this: 
```
{{ range .Site.Data.jazz.bass }}
   {{ . }}
{{ end }}
```

This is my first pull request. So any suggestion for improvement is highly welcome.

### DIFF
--- a/parser/metadecoders/decoder.go
+++ b/parser/metadecoders/decoder.go
@@ -112,7 +112,7 @@ func (d Decoder) UnmarshalStringTo(data string, typ any) (any, error) {
 // Unmarshal will unmarshall data in format f into an interface{}.
 // This is what's needed for Hugo's /data handling.
 func (d Decoder) Unmarshal(data []byte, f Format) (any, error) {
-	if data == nil {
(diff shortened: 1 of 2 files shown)


bep:
Thanks. The fix looks good, but I think we need a test that demonstrates this issue to avoid someone breaking it in the future.

acclassic:
You're welcome. 
Thanks for the feedback. I added an empty JSON to the decoder_test file. If the Unmarshal function is now modified and does not check for the length of the data the test will fail.

acclassic:
Sorry I forgot to rebase my commit.

bep:
Thanks for this, closing this in favour of #10537

For future contributions I would recommend that you start by creating a new branch for the PR and not adding commits to your master branch. This makes for a much simpler work flow.

PR #10550, opened 2022-12-19
tpl/tplimpl: Exclude missing translations from sitemap

### LINKED ISSUE #9073: Multilang sitemap includes empty rel=alternate link if disableKinds=taxonomy is used by one language

<pre>
hugo v0.88.1+extended darwin/amd64
</pre>

Yes

### Issue description
Our website is using Hugo with multi language setup with different `baseURL` for each language. One language (let's say English) has a blog with standard taxonomies (categories, tags) set up, but another language (let's say French) does not show the blog (because we don't have it translated) at all and under `[languages.fr]` has `disableKinds = ["taxonomy"]`. 
This works as expected in most areas, except the sitemap for English contains entries like this:
```
<url>
        <loc>https://example.com/tags/</loc>
        <lastmod>2017-04-03T10:00:07+00:00</lastmod>
        <xhtml:link rel="alternate" hreflang="fr" href=""/>
        <xhtml:link rel="alternate" hreflang="en" href="https://example.com/tags/"/>
    </url>
```
with empty link for French for both categories and tags taxonomy pages.  Sitemap for French is fine - does not contain any reference to taxonomies at all.
All other links in the sitemaps are fine, taxonomy term pages do not have the empty link either.

jmooring:
Fixes #9073

### DIFF
--- a/tpl/tplimpl/embedded/templates/_default/sitemap.xml
+++ b/tpl/tplimpl/embedded/templates/_default/sitemap.xml
@@ -1,24 +1,34 @@
-{{ printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
+{{- printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
 <urlset xmlns="https://www.sitemaps.org/schemas/sitemap/0.9"
   xmlns:xhtml="https://www.w3.org/1999/xhtml">
-  {{ range .Data.Pages }}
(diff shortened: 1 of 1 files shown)


bep:
I'm not totally sure I understand why/how this happens, but fixing it in the sitemap template feels a little arbitrary. What do these pages without a link represent?

jmooring:
> fixing it in the sitemap template feels a little arbitrary
Yeah, I agree. 

> What do these pages without a link represent?
No clue. I didn't want to go down that rabbit hole last night, but I'll have another look as time allows.

PR #10567, opened 2022-12-22
feat: add parents property on page

septs:
Add parents property on page, simplify breadcrumb navigation implementation complexity

community tricks see: https://discourse.gohugo.io/search?q=breadcrumb

---

in HTML:

```html
{{- $page := . -}
<nav class="breadcrumb">
<ul>
{{- range .Parents.Reverse }}
<li><a href="{{ .Permalink }}">{{ .LinkTitle }}</a></li>
{{- end }}
<li class="active">{{ .LinkTitle }}</li>
</ul>
</nav>
```

---

in JSON-LD:

see https://developers.google.com/search/docs/appearance/structured-data/breadcrumb

```html
{{- $breadcrumbItems := slice -}}
{{- range $index, $page := append .Parents.Reverse . -}}
  {{-
  $item := dict
    "@type" "ListItem"
    "position" (add $index 1)
    "name" $page.Title
    "item" $page.Permalink
  -}}
  {{- $breadcrumbItems = append $breadcrumbItems $item -}}
{{- end -}}
<script type="application/ld+json">
{{-
dict
  "@context" "https://schema.org"
  "@type" "BreadcrumbList"
  "itemListElement" $breadcrumbItems
-}}
</script>
```

### DIFF
--- a/hugolib/page__tree.go
+++ b/hugolib/page__tree.go
@@ -178,13 +178,14 @@ func (pt pageTree) Parent() page.Page {
 	return b.p
 }
 
-func (pt pageTree) Ancestors() (parents page.Pages) {
+func (pt pageTree) Ancestors() page.Pages {
+	var ancestors page.Pages
 	parent := pt.Parent()
 	for parent != nil {
-		parents = append(parents, parent)
+		ancestors = append(ancestors, parent)
 		parent = parent.Parent()
 	}
-	return
+	return ancestors
 }
 
 func (pt pageTree) Sections() page.Pages {
--- a/hugolib/page__tree.go
+++ b/hugolib/page__tree.go
@@ -178,6 +178,15 @@ func (pt pageTree) Parent() page.Page {
 	return b.p
 }
 
+func (pt pageTree) Ancestors() (parents page.Pages) {
+	parent := pt.Parent()
+	for parent != nil {
+		parents = append(parents, parent)
+		parent = parent.Parent()
+	}
+	return
+}
+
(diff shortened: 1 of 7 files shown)


bep:
I agree that the idea is good, but `Parents` isn't a good name (a `Page` has never more than 1 `Parent`).

I suggest we use `.Ancestors`.

Also, could you:

* Squash the commits and add one commit message that follows the commit message guidelines in CONTRIBUTION.md
* Add a short description in the Page variables section about this new method.

hugolib/page__tree.go
```diff
@@ -178,6 +178,15 @@ func (pt pageTree) Parent() page.Page {
 	return b.p
 }
 
+func (pt pageTree) Ancestors() (parents page.Pages) {
+	parent := pt.Parent()
+	for parent != nil {
+		parents = append(parents, parent)
```
bep, line 184: Nitpicking, but 

* `parents` should be named something else
* we only use named return values when it adds value (e.g. when we need to capture a panic and assign to err), so I would prefer `Ancestors() page.Pages`

bep:
@jmooring for me this is an obvious improvement, agree?

jmooring:
> this is an obvious improvement, agree?

Yes. Took it for a spin. :+1:

hugolib/page__tree.go
bep: OK, I can do it. I need to merge this soonish, as I'm releasing today.

septs: Sorry, I got COVID-19 and I'm fine now

PR #10577, opened 2022-12-23
Remove forestry from frontends - it has a March 2023 EOL

karlkeefer:
This was announced over email.

That team is moving to a new project called [tina CMS](https://tina.io/guides/tinacms/non-react-based-ssg/guide/) that seems _possible_ to use with Hugo, but it appears less well supported than what we had on forestry (e.g. no automatic schema generation).

### DIFF
--- a/docs/content/en/tools/frontends.md
+++ b/docs/content/en/tools/frontends.md
@@ -26,5 +26,4 @@ toc: false
 ## Commercial Services
 
 - [DATOCMS](https://www.datocms.com) DatoCMS is a fully customizable administrative area for your static websites. Use your favorite website generator, let your clients publish new content independently, and the host the site anywhere you like.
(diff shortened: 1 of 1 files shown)


bep:

karlkeefer:
Opened a PR on the docs repo.

Depending on the behavior you want for other public PR submissions, maybe the readme in `docs` on the main project can be made more minimal to point folks to the other repo? If I understand your _intent_ I can draft some copy and open another PR here. I see [changes](https://github.com/gohugoio/hugo/commit/59af05cabc74e381afb62e624afa5542a9ae8d68#diff-2ac4535c92b2376113d9ff49fad62c4a84b4f6e92708fb479791118627e15e5b) also happen to `docs` in _this_ repo, so it's not clear to me how these are sync'd.

Thanks for sharing Hugo with the world :)

PR #10578, opened 2022-12-24
tpl: Add Classnames function

septs:
port classnames from https://npm.im/classnames

the feature can simplify class names builder, no need to join-strings in the template

after:

```
class="table
{{- if $striped }}table-striped{{ end }}
{{- if $bordered }}table-bordered{{ end }}
{{- if $borderless }}table-borderless{{ end }}
{{- if $hoverable }}table-hover{{ end }}
{{- if $small }}table-sm{{ end }}
"
```

before:

```
{{-
$classes := classnames
  "table"
  (dict
    "table-striped" $striped
    "table-bordered" $bordered
    "table-borderless" $borderless
    "table-hover" $hoverable
    "table-sm" $small
  )
-}}
class="{{ $classes }}"
```

of course you can

```
{{- $classes := classnames "table" .Attibutes -}}
class="{{ $classes }}"
```

### DIFF
--- a/tpl/strings/classnames.go
+++ b/tpl/strings/classnames.go
@@ -1,0 +1,66 @@
+package strings
+
+import (
+	"reflect"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// ClassNames the method ported npm package. see https://npm.im/classnames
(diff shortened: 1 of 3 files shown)


bep:

Any new features needs to start with an issue.

PR #10580, opened 2022-12-25
Remove code apparently obsoleted due to defaultness of pretty

swaldman:
I apologize for the very trivial pull request!

While beginning to learn / find my way around the codebase, I found a couple of methods of PathSpec that seem to be obsolete, replaced by more newer functionality. Maybe they should be removed?


swaldman:
Looking around at code involving pretty vs ugly URLs, it looks like a bunch of code is no longer used since pretty-style URLs are now default.

Issue #4071, opened 2017-11-10
Need option to use full content in rss.xml


rayjohnson:
The default rss.xml template uses .Summary for the content of each item.  If you want to use the full content instead you have to override the entire template.  It would be nice to have an option one can set int he config.toml to control if .Summary or .Content is used.

paskal:
Seems like adding whole content to summary [does the trick](https://discourse.gohugo.io/t/flag-indicating-manual-vs-auto-summary-generation/1155/10), however it's a shame I have to make post summary-less to have it added to RSS properly (with full content).

danielcompton:
https://github.com/gohugoio/hugo/issues/4242 is an issue proposing a solution to this.

jakewies:
Be gone Stale bot! Be gone!

bigblind:
+1 for implementing this.

Kaligule:
I would like to advocate an option to set the RSS-feeds to full-text.

### Hugo users want this

[A quick websearch for "hugo full text rss"](https://duckduckgo.com/?q=hugo+full+text+rss) finds up a lot of blogposts of hugo users who manually adjusted their RSS templates. These are all from the first page of results:

- https://www.gideonwolfe.com/posts/sysadmin/hugo/hugorss/#showing-all-your-content
- https://fuzhouch.com/post/2020-06-11-hugo-tips-generate-full-text-rss-atom-feeds/
- https://blog.eternalrecurrence.space/posts/displaying-the-full-content-in-hugo-rss-feed/
- https://discourse.gohugo.io/t/full-text-rss-feed/8368
- https://www.duskborn.com/posts/fixed-up-rss/

I think that this shows, that full-text rss is a feature that many hugo users want.

### Other static-site-generators have it.

@obar [pointed out](https://github.com/gohugoio/hugo/pull/6804#issuecomment-589710853) that this feature is already present in many other static site generators like:

- [wordpress](https://en.support.wordpress.com/settings/reading-settings/)
- [jekyll](https://github.com/jekyll/jekyll-feed/blob/master/README.md#excerpt-only-flag)
- [hexo](https://github.com/hexojs/hexo-generator-feed/blob/master/README.md#options)…

### The hacky solution is insufficient

@nekr0z  [pointed out](https://github.com/gohugoio/hugo/pull/6804#issuecomment-633144770) that just swapping out `Summary` and `Content` could result in "non-conformant" RSS feeds.

tohn:
[...]

PR #10589, opened 2022-12-29
rss: Add config variable to output full content

kusma:
The RSS 2.0 specification states:
"An item's description element holds character data that contains the item's full content or a summary of its contents, a decision entirely at the discretion of the publisher."

So far, we do the latter only. But it can sometimes be useful to have the full content included instead, so let's add a config variable to also support the former.

### DIFF
--- a/tpl/tplimpl/embedded/templates/_default/rss.xml
+++ b/tpl/tplimpl/embedded/templates/_default/rss.xml
@@ -32,7 +32,13 @@
       <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
       {{ with .Site.Author.email }}<author>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</author>{{end}}
       <guid>{{ .Permalink }}</guid>
-      <description>{{ .Summary | html }}</description>
+      <description>
+      {{- if .Site.Config.Services.RSS.FullContent -}}
+      {{ .Content | html }}
+      {{- else -}}
+      {{ .Summary | html }}
+      {{- end -}}
+      </description>
     </item>
     {{ end }}
   </channel>
--- a/config/services/servicesConfig.go
+++ b/config/services/servicesConfig.go
@@ -72,6 +72,8 @@ type Twitter struct {
 type RSS struct {
 	// Limit the number of pages.
 	Limit int
+	// Output the full content for each article instead of just the summary
+	FullContent bool
 }
 
 // DecodeConfig creates a services Config from a given Hugo configuration.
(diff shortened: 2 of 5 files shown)


kusma:
Hey, this is my first code-contribution to Hugo, so my apologies if I've made a mess here.

There's two things I'm not super-happy with in this MR:

1. The documentation. Yeah, there's some documentation there, but it kinda feels glued on. I've based things on the `rssLimit`-config variable, but that doesn't feel great, because that one has really been renamed, and the docs seems to lag behind. I can't see a very good precedence here, but maybe I'm missing something. I'm not a great documentation writer, so input would be greatly appreciated. 
2. There's an annoying newline added before the closing of the `</description>`-tag. Some quick testing seems to indicate that this somehow comes from the `html` template function, but I'm not 100% sure. It shouldn't matter much, but it feels a bit meh.

My primary reason for adding this is that I need to emit the whole article for use on https://planet.freedesktop.org, and since I changed to using Hugo, my articles are seemingly the only ones on the site that render only the summary. I have done the change locally, but it seems there's [more than just me](https://discourse.gohugo.io/t/full-text-rss-feed/8368) who want this, and some users seem to have some issues getting it to work.

So it seems like this might be useful upstream.

Thoughts?

kusma:
> [#6804 (comment)](https://github.com/gohugoio/hugo/pull/6804#issuecomment-633144770)

Well, yeah... But then again, like [pointed out in the response](https://github.com/gohugoio/hugo/pull/6804#issuecomment-932007612), this is a general problem with RSS feeds in Hugo, and isn't really caused by full content vs summary. It's just a tad less likely with the summary.

I don't disagree that it's not a problem, but I think the problem is orthogonal to such a feature.

jmooring:
Closing this; the original proposal (https://github.com/gohugoio/hugo/issues/4071) has not been accepted.

tpl/tplimpl/embedded/templates/_default/rss.xml
```diff
@@ -32,7 +32,13 @@
       <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
       {{ with .Site.Author.email }}<author>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</author>{{end}}
       <guid>{{ .Permalink }}</guid>
-      <description>{{ .Summary | html }}</description>
+      <description>
+      {{- if .Site.Config.Services.RSS.FullContent -}}
+      {{ .Content | html }}
+      {{- else -}}
+      {{ .Summary | html }}
```
kskarthik, line 39: `{{ .Content | transform.XMLEscape | safeHTML }}` & `{{ .Content | transform.XMLEscape | safeHTML }}`  is a more safe approach

kskarthik:
This patch is quiet good with a small change mentioned above ^ 

Hope the proposal gets accepted.

PR #10593, opened 2023-01-01
dartsass: Import CSS without extension at compile time

### LINKED ISSUE #10592: Unable to import mounted CSS into SCSS.
Hi, Hugo seems doesn't able to load mounted CSS file into SCSS, but normal CSS files are OK.

<pre>
$ hugo version
hugo v0.109.0-47b12b83e636224e5e601813ff3e6790c191e371+extended linux/amd64 BuildDate=2022-12-23T10:38:11Z VendorInfo=gohugoio
</pre>

Yep.

### Log

```
Template changed WRITE         "/home/razonyang/Projects/razonyang/hugo-lab/layouts/index.html"
WARN 2022/12/31 14:13:30 tingle.js/tingle.css
WARN 2022/12/31 14:13:30 tingle.js/tingle.js
ERROR 2022/12/31 14:13:30 Rebuild failed: TOCSS-DART: failed to transform "main.scss" (text/x-scss): "/home/razonyang/Projects/razonyang/hugo-lab/assets/main.scss:2:8": Can't find stylesheet to import.
```

### Reproduce steps

```toml
// config.toml
[[module.imports]]
path = "github.com/robinparisi/tingle"

[[module.imports.mounts]]
source = "src"
target = "assets/tingle.js"
```

```go
// layouts/index.html
<html>
    <head>
        {{- range resources.Match "tingle.js/*" }}
          {{- warnf "%v" .Name }}
        {{- end }}
        {{ $css := resources.Get "main.scss" }}
        {{ $css = $css | toCSS (dict "includePaths" (slice "node_modules") "transpiler" "dartsass") }}
        <link href="{{ $css.RelPermalink }}" rel="stylesheet">
    </head>
    <body>
        <h1 class="text-red">RED HEADING</h1>
    </body>
</html>
```

```scss
// assets/main.scss
@import "foo/bar";
@import "tingle.js/tingle"; # Comment this line works as expected.

// assets/foo/bar.css
.text-red {
    color: red;
}
```

> Full example located at https://github.com/razonyang/hugo-lab/tree/scss-import-css. `cd exampleSite && hugo server`.

Both `libsass` and `dartsass` are not working.

jmooring:
Applicable to Dart Sass only:

- Sass imports with the .css extension indicate a plain CSS @import.
- Sass imports without the .css extension are imported at compile time.

Fixes #10592

### DIFF
--- a/resources/resource_transformers/tocss/dartsass/integration_test.go
+++ b/resources/resource_transformers/tocss/dartsass/integration_test.go
@@ -68,22 +68,24 @@ $moolor: #fff;
 moo {
 	color: $moolor;
 }
--- assets/scss/another.css --
-
+-- node_modules/foo/another.css --
+another {
+	color: #000;
+}
 -- assets/scss/main.scss --
 @import "moo";
 @import "regular.css";
 @import "moo";
-@import "another.css";
+@import "another";
 
 /* foo */
 -- assets/scss/regular.css --
 
 -- config.toml --
 -- layouts/index.html --
-{{ $r := resources.Get "scss/main.scss" |  toCSS (dict "transpiler" "dartsass")  }}
+{{ $cssOpts := (dict "includePaths" (slice "node_modules/foo") "transpiler" "dartsass") }}
+{{ $r := resources.Get "scss/main.scss" | toCSS $cssOpts }}
 T1: {{ $r.Content | safeHTML }}
-
 	`
 
 	b := hugolib.NewIntegrationTestBuilder(
@@ -97,15 +99,18 @@ T1: {{ $r.Content | safeHTML }}
 	// Dart Sass does not follow regular CSS import, but they
 	// get pulled to the top.
 	b.AssertFileContent("public/index.html", `T1: @import "regular.css";
-		@import "another.css";
 		moo {
 		  color: #fff;
 		}
-		
+
 		moo {
 		  color: #fff;
 		}
-		
+
+		another {
+			color: #000;
+		}
+
 		/* foo */`)
 }
 
@@ -291,7 +296,7 @@ body {
 	body {
 		background: url(vars.$image) no-repeat center/cover;
 		font-family: vars.$font;
-	  }	  
+	  }
 }
 
 p {
@@ -341,7 +346,7 @@ image = "images/hero.jpg"
 body {
 	body {
 		background: url(vars.$image) no-repeat center/cover;
-	  }	  
+	  }
 }
 
 p {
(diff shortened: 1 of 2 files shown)


resources/resource_transformers/tocss/dartsass/integration_test.go
```diff
@@ -68,22 +68,24 @@ $moolor: #fff;
...
-@import "another.css";
+@import "another";
 
 /* foo */
 -- assets/scss/regular.css --
 
 -- config.toml --
 -- layouts/index.html --
-{{ $r := resources.Get "scss/main.scss" |  toCSS (dict "transpiler" "dartsass")  }}
+{{ $cssOpts := (dict "includePaths" (slice "node_modules/foo") "transpiler" "dartsass") }}
```
bep, line 86: The original issue, as I understand it, was about importing CSS files from `/assets`; with that I don't understand why you move the CSS to `node_modules`?

jmooring: > The original issue, as I understand it, was about importing CSS files from /assets

The OP incorrectly commented on what was working vs. what was not. I discovered their error when I began testing this, but did not ask the OP to verify their comments until yesterday. They confirmed that the original comments were reversed, but the issue title is correct.

We have no problem importing a CSS file is the assets directory at compile time. But when you import a CSS file from a mounted directory (either through `module.mounts` or the `includePaths` slice), the result is a plain import.

I would be happy to clarify the test if necessary, but I was trying to minimize the `diff` for you.

bep: >I would be happy to clarify the test if necessary, but I was trying to minimize the diff for you.

To minimise the diff would be dropping the includePaths change, which is what confuses me -- even after rereading the original issue. I think I understand the issue, and if I understand it correctly that should be possible to reproduce by just importing a CSS file without using the extension when the file is imported into /assets in a mount (which lives outside of what includePaths is all about).

jmooring:
bep, I apologize for the delay on this. I have been having intermittent failures when testing. It turns out the tests were intermittently skipped due to:

```go
if !dartsass.Supports() {
	t.Skip()
}
```

That is not your problem.

I had been running the tests (a) within VS Code, and (b) with dart-sass-embedded installed via Brew. There are no problems when dart-sass-embedded is installed in /user/local/bin. I still haven't worked out the correct settings in VS Code to address this, so I'll stick with the install in /user/local/bin.

bep:
Bump ... Is this still an issue/relevant?

jmooring:
Yes, still relevant, got distracted. I can look at it again later today (much later today).

PR #10635, opened 2023-01-21
feat: reduce alt output format template code in theme

septs:
after [^1]:
```
{{ range .AlternativeOutputFormats -}}
<link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink | safeURL }}">
{{ end -}}
```

before (if no special requirements [^2]):

```
{{ range .AlternativeOutputFormats -}}
{{ . }}
{{ end -}}
```

or

```
{{ .AlternativeOutputFormats }}
```

[^1]: https://gohugo.io/templates/output-formats/#list-output-formats
[^2]: for example: [OpenSearch](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#autodiscovery_of_search_plugins) need "title" field

### DIFF
--- a/resources/page/page_outputformat.go
+++ b/resources/page/page_outputformat.go
@@ -18,6 +18,7 @@ package page
 import (
 	"fmt"
 	"strings"
+	"text/template"
 
 	"github.com/gohugoio/hugo/media"
 	"github.com/gohugoio/hugo/output"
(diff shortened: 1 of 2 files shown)


bep:
The concept of output formats was added so we could output something other than ... HTML. This PR ties the output format beck to ... HTML, which I don't thin is worth a couple of extra characters in a template.

septs:
why can't it be similar to `(page variable).TableOfContents`?

PR #10638, opened 2023-01-22
refactor: improve math

septs:
allow multi numbers of add (`+`) / sub (`-`) / mul (`*`) / div (`/`) / min value / max value

refactor before:

```
add 1 2 | add 3 | add 4 // equivalent to: 1 + 2 + 3 + 4
sub 1 2 | sub 3 | sub 4 // equivalent to: 1 - 2 - 3 - 4
mul 1 2 | mul 3 | mul 4 // equivalent to: 1 * 2 * 3 * 4
div 1 2 | div 3 | div 4 // equivalent to: 1 / 2 / 3 / 4
math.Min 1 2 | math.Min 3 | math.Min 4
math.Max 1 2 | math.Max 3 | math.Max 4
```

refactor after:

```
add 1 2 3 4
sub 1 2 3 4
mul 1 2 3 4
div 1 2 3 4
math.Min 1 2 3 4
math.Max 1 2 3 4
```

### DIFF
--- a/docs/content/en/functions/math.md
+++ b/docs/content/en/functions/math.md
@@ -3,7 +3,7 @@ title: Math
 description: Hugo provides mathematical operators in templates.
 date: 2017-02-01
 publishdate: 2017-02-01
-lastmod: 2020-02-23
+lastmod: 2023-03-11
 keywords: [math, operators]
 categories: [functions]
 menu:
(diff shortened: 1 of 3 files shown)


septs:
apply e.g: fit image dimension

after:

```
math.Min (div $maxWidth / .Width) (div $maxHeight / .Height) | math.Min 1
```

before:

```
math.Min (div $maxWidth / .Width) (div $maxHeight / .Height) 1
```

bep:
OK, this I like. What do you think, @jmooring ?

jmooring:
I have tested it. Looks good to me, though we might add some test cases, one with 3 arguments to each of the 6 functions.

bep (changes requested):
OK, I like this change.

Could you 

* add some basic test cases to cover this
* adjust the Godoc and also the Markdown docs below /docs

bep (approved):

PR #10647, opened 2023-01-24
opengraph template: CJK series meta tag malfunction

tomy0000000:
The current `opengraph.html` template generate the `og:see_also` meta tags, but it does not work when the `series` front matter includes CJK characters. The root cause is that these characters are being url-encoded after `urlize` filter, while they're not in the `.Site.Taxonomies.series` map.

e.g.
Say I have a tag named `料理`, after `urlize` filter, it becomes `%E6%96%99%E7%90%86`. Meanwhile, the `.Site.Taxonomies.series` looks a bit like:

```
map[料理:[WeightedPage(0,"炒飯") WeightedPage(0,"炒麵")]]
```

which is why it return no series page.

The ideal solution should be removing the `urlize`, but the filter also sanitized other characters (e.g. emojis), so I fix the bug by using the `urls.Parse` function and then take the `Path` attribute, which is roughly URL decode the percentage encoded string back.

### DIFF
--- a/tpl/tplimpl/embedded/templates/opengraph.html
+++ b/tpl/tplimpl/embedded/templates/opengraph.html
@@ -35,7 +35,7 @@
 {{- $siteSeries := .Site.Taxonomies.series }}
 {{- if $siteSeries }}
 {{ with .Params.series }}{{- range $name := . }}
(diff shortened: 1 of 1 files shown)


tomy0000000:
Is there anything I should do to fix the CI checks?

PR #10653, opened 2023-01-25
chore: improve table-of-contents

septs:
if headings is empty, then no generates table-of-contents

### DIFF
--- a/markup/tableofcontents/tableofcontents.go
+++ b/markup/tableofcontents/tableofcontents.go
@@ -88,8 +88,11 @@ func (b *tocBuilder) Build() {
 }
 
 func (b *tocBuilder) writeNav(h Headings) {
+	if len(h) == 0 {
+		return
+	}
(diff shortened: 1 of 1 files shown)


markup/tableofcontents/tableofcontents.go
```diff
@@ -88,8 +88,11 @@ func (b *tocBuilder) Build() {
 }
 
 func (b *tocBuilder) writeNav(h Headings) {
+	if len(h) == 0 {
+		return
+	}
```
septs, line 93: https://discourse.gohugo.io/t/29737

https://github.com/gohugoio/hugo/blob/93ed6e447a24f8d08f136bd35fe3117bd6b29396/hugolib/page__per_output.go#L133-L147

bep:
This is a pretty breaking change that we cannot merge. We have another issue that discusses a proper solution to this. Currently the workaround if people really like to check for this is to do a `{{ if eq (strings.Count "li" .TableOfContents) 0 }}` or something.

PR #10660, opened 2023-01-27
strings: fix Truncate behavior for formatted html

### LINKED ISSUE #10399: truncate doesn't handle formatted HTML very well
E.g.:

```
<a
  href="https://gohugo.io/content-management/front-matter/"  
    target="_blank"  
  >Front matter</a
>
```

khayyamsaleem:
Before this fix, strings.Truncate would erroneously re-include attributes from the opening tag in the closing tag when closing formatted html, due to a bug in how tagnames were extracted from the regex capture group for html tags used in `truncate.go`.

This change ensures that only the tagname is retained and all attributes are discarded when storing the tags for closing them later.

Fixes #10399

### DIFF
--- a/tpl/strings/truncate.go
+++ b/tpl/strings/truncate.go
@@ -18,6 +18,7 @@ import (
 	"html"
 	"html/template"
 	"regexp"
+	"strings"
 	"unicode"
 	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


bep:
Merged in c0d15a2897be0bc9f6df2514f5a477b6df1fa0e5

Issue #5792, opened 2019-03-25
Improve the cond func


bep:
There are two items that I want to address:

1. Make it use the same "truth" function as if/with/not/and -- but we should think a little if this would break stuff
2. Lazily invocation of passed funcs/methods

Basically I want to be able to do something ala:

```bash
n := cond .File .File.Filename .Name
```

Maybe not the best example, but imagine a nilpointer scenario etc.

jamesgeorge007:
@bep I would love to give it a try :clap: I'm new here :beginner: 
It would be great if you could point me to the right files to work with :+1:

bep:
@jamesgeorge007 this is a particulary hard problem. I don't know how to do it myself.

regisphilibert:
One line is always better than 4, especially in Go Template.
```
{{ $n := .Name }}
{{ with .File }}
  {{ $n = .Filename }}
{{ end }}
```

jamesgeorge007:
@bep @regisphilibert if that's the case, can you point me to something that is good to start with :clap:

moorereason:
Given:
```
n := cond .File .File.Filename .Name
```
@bep, as I under the existing implementation, it would blow up if `.File` is nil because the text/template engine would resolve `.File.Filename` before sending it to `cond`. Correct?

bep:
@moorereason It would 2 days ago ... I changed it from a pointer to a struct (which is how was in 0.54) for a very similar reason.

But you are right, there are (at least) 2 reasons why I want to fix this:

* To avoid nilpointers (a comon use case for `cond`
* To avoid costly initialization for values I don't use

The problem, I have no idea how to do this without making modifications to Go's stdlib, so I might bring it up with them (would not be too hopeful). Because now:

```
{{ if and .Foo .Bar }}{{ end }}
```

(same with `or`)

Is very different from:

```go
if .Foo && .Bar {
}
```

bep:
https://github.com/gohugoio/hugo/issues/5792

bep:
I think any help from the Go stdlib in this area is long down the road. But for `cond` I think we could solve it by handling it as (another) special case in the AST transformer:

```
{{ $n := cond .File .File.Filename .Name }}
```

Becomes

```
{{ $_tmp := "" }}
{{ if .File }}
{{ $_tmp = .File.Filename }}
{{ else }}
{{ $_tmp = .Name }}
{{ end }}
{{ $n := $_tmp }}
```

Or similar.

[...]

PR #10666, opened 2023-01-28
tpl: adds `truth` and `bool` template functions

### LINKED ISSUE #9160: Add bool to template functions
#5792 proposes improving `cond` to accept non-bool first arguments. That should also be done, but it would also be good to have a freestanding function to convert values to bools.

As a workaround, now one can do `{{ $x := $y |  not | not }}`, but that's a bit ugly.

khayyamsaleem:
The behavior of `truth` and `bool` is described in the corresponding test cases and examples. The decision-making around the behavior is a based on combination of the existing behavior of strconv.ParseBool in go and the MDN definition of "truthy" as JavaScript has the most interop with the Hugo ecosystem.

Fixes #9160 and (indirectly) #5792

### DIFF
--- a/docs/content/en/functions/truth.md
+++ b/docs/content/en/functions/truth.md
@@ -1,0 +1,53 @@
+---
+title: truth
+linktitle: truth
+description: Creates a `bool` from the truthyness of the argument passed into the function
+date: 2023-01-28
+publishdate: 2023-01-28
+lastmod: 2023-01-28
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+keywords: [strings,boolean,bool,truthy,falsey]
+signature: ["truth INPUT"]
+workson: []
+hugoversion:
+relatedfuncs: [bool]
+deprecated: false
+aliases: []
+---
+
+Useful for turning different types into booleans based on their [truthy-ness](https://developer.mozilla.org/en-US/docs/Glossary/Truthy).
+
+It follows the same rules as [`bool`](/functions/bool), but with increased flexibility.
+
+```
+{{ truth "true" }} → true
+{{ truth "false" }} → false
+
+{{ truth "TRUE" }} → true
+{{ truth "FALSE" }} → false
+
+{{ truth "t" }} → true
+{{ truth "f" }} → false
+
+{{ truth "T" }} → true
+{{ truth "F" }} → false
+
+{{ truth "1" }} → true
+{{ truth "0" }} → false
+
+{{ truth 1 }} → true
+{{ truth 0 }} → false
+
+{{ truth true }} → true
+{{ truth false }} → false
+
+{{ truth nil }} → false
+
+{{ truth "cheese" }} → true
+{{ truth 1.67 }} → true
+```
+
+This function will not throw an error. For more strict behavior, see [`bool`](/functions/bool).
--- a/docs/content/en/functions/truth.md
+++ b/docs/content/en/functions/truth.md
@@ -20,23 +20,21 @@ aliases: []
 
 Useful for turning different types into booleans based on their [truthy-ness](https://developer.mozilla.org/en-US/docs/Glossary/Truthy).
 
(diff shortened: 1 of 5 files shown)


khayyamsaleem:
@bep open to discussion on the behavior / naming for sure; thought this would open a more productive discussion on #9160 as it was low-hanging fruit that has been open for a long time.

khayyamsaleem:
For `truth`, could also just use the rules in `hreflect.IsTruthful`, but thought that the MDN standards may be more familiar to end-users. I guess it's unlikely to come up with something that EVERYONE likes here, so best to provide some capability and gauge the response from discourse or GH issues after the release.

bep:
@khayyamsaleem thanks for this. My 50 cents about this, and please don't adjust your code before we agree on something is.

As to `truth`, these test cases needs to pass and behave the same for all "truthy values" on the left side:

```handlebars
{{ if "foo" }}OK{{ else }}Failed{{ end }}
{{ if ("foo" | truth) }}OK{{ else }}Failed{{ end }}
{{ if ("foo" | not | not ) }}OK{{ else }}Failed{{ end }}
```

Which means that if would be a thin wrapper around `IsTruthFul`.

As to `bool`, I'm tempted to just use `strconv.ParseBool`. Combining it with some JavaScript behaviour is surely going to get us into the Yaml trap (YAML < 3 had "yes/no" as bools).

//cc @jmooring

jmooring:
A colleague of mine told me about a 700+ line YAML file they had to debug a few years ago. It had some country codes. Norway was falsy.

khayyamsaleem:
@bep I'm fine with the suggestions; better to have one definition of truthiness for the codebase.

bep:
@khayyamsaleem OK, if you adjust this PR to that much simpler scenario, I will merge.

To be clear:

* bool == https://pkg.go.dev/strconv#ParseBool
* truth = if reflect.IsTruthful(v) true else false

bep:
OK, I changed my mind again. Looking at the other cast.* functions we have, and also on other places we do similar, I think it's a pretty clear precedence to just use `cast.ToBoolE`, so I suggest we just do that for the `bool` func.

stereobooster:
HI @bep. Does this PR needs more work? As far as I understood `ToBool` function is ok. Do we need to adjust `ToTruth` function (`if reflect.IsTruthful(v) true else false`)? Can I help here?

khayyamsaleem:
@bep pardon the delay here. Updated PR per your last review.

jmooring:
This is not what we agreed to:
<https://github.com/gohugoio/hugo/pull/10666#issuecomment-1443251316>

`cast.ToBool` behaves as I expect, though I think we should return false on error. 

`cast.ToTruth` should use `hreflect.IsTruthful`. The string "false" is truthy.

khayyamsaleem:
Ahhh pardon, you're right. It did seem awkward for Truth and bool to disagree on "false" though.

docs/content/en/functions/truth.md
```diff
@@ -0,0 +1,53 @@
...
+workson: []
+hugoversion:
+relatedfuncs: [bool]
+deprecated: false
+aliases: []
+---
+
+Useful for turning different types into booleans based on their [truthy-ness](https://developer.mozilla.org/en-US/docs/Glossary/Truthy).
+
+It follows the same rules as [`bool`](/functions/bool), but with increased flexibility.
```
jmooring, line 23:
```suggestion
```

jmooring:
@bep I have not tested this yet, but I think the `truth` function belongs in the compare namespace as `IsTruthy`. This function does not cast a value into another data type; it performs a comparison.

```
{{ compare.IsTruthy $foo }}
{{ isTruthy $foo }} (alias)
```

and

```
{{ cast.ToBool $foo}}
{{ bool $foo }} (alias)
```

khayyamsaleem:
Are we waiting for @bep 's feedback here, or would we prefer to move the function to the `compare` namespace?

PR #10674, opened 2023-01-31
all: Updated the deprecated ioutil dependency

zachary-walters:
The ioutil package has been deprecated as of Go v1.16: https://go.dev/doc/go1.16#ioutil This commit replaces ioutil functions with their respective io/os functions.

### DIFF
--- a/cache/filecache/filecache.go
+++ b/cache/filecache/filecache.go
@@ -17,7 +17,6 @@ import (
 	"bytes"
 	"errors"
 	"io"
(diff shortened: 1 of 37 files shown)


bep:
Thanks for this, but we're currently not merging any massive renames/refactoring PRs, as we have a few bigger branches in line -- and merging this particular PR would create a lot of work merging those for no very good reason. We will make sure to migrate to the new package structure in good time.

PR #10693, opened 2023-02-04
markup: fixes regression for linenos codeblock hl option

### LINKED ISSUE #10682: lineNos in code block info string throws error with v0.93.0 and later if value is not boolean
With v0.92.2 and earlier, either of these work (note the mixed case of the key in the first example):

````text
```text {lineNos=inline}
a
```

```text {linenos=inline}
a
```
````

With v0.93.0 and later, this throws an error:

````text
```text {lineNos=inline}
a
```
````

> 1 error(s) decoding: * cannot parse 'LineNos' as bool: strconv.ParseBool: parsing "inline": invalid syntax

With boolean values, naming the key `lineNos` is not a problem. The error occurs when the value is `inline` or `table` (shorthand notation for setting both `LineNos` and `lineNumbersInTable`).

I tested all of the highlighting options in the info string, using both lower and mixed case key names. The only one that fails is `LineNos`.

Git bisect says 08fdca9d9365eaf1e496a12e2af5e18617bd0e66 is the first bad commit, though I'm not sure how helpful that is given `73 changed files with 1,882 additions and 1,981 deletions.`

I ran into this because I blanked on the key name, so I copied and pasted from the [docs](https://gohugo.io/functions/highlight/#options).

khayyamsaleem:
This fixes a regression introduced in v0.93.0 where previously, a
mixed-case key for lineNos would be successfully parsed. This change
moves the configuration key lowercasing step into the configuration
normalization stage, which is called whether the highlighting config
is being parsed from a `string` or a `map`.

Fixes #10682

### DIFF
--- a/markup/highlight/config.go
+++ b/markup/highlight/config.go
@@ -198,7 +198,7 @@ func parseHightlightOptions(in string) (map[string]any, error) {
 
 	for _, v := range strings.Split(in, ",") {
 		keyVal := strings.Split(v, "=")
-		key := strings.ToLower(strings.Trim(keyVal[0], " "))
+		key := strings.Trim(keyVal[0], " ")
 		if len(keyVal) != 2 {
 			return opts, fmt.Errorf("invalid Highlight option: %s", key)
 		}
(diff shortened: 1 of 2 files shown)


khayyamsaleem:
Is it appropriate for this to include an integration test in `markup/highlight/integration_test.go`?

PR #10706, opened 2023-02-08
docs(Diagrams): Update mermaid to esm

sidharthv96:
I was trying to use hugo in Mermaid's new blog, saw it was using the deprecated method to integrate mermaid to websites. 
I've fixed it to use the new ESM method.


bep:

PR #10718, opened 2023-02-12
Fixed hugo:vars casting

### LINKED ISSUE #10632: Transpiled hex-string color from config file cannot be used as color in css functions such as `lighten()` and `darken()`

<pre>
$ hugo version
hugo v0.110.0-e32a493b7826d02763c3b79623952e625402b168+extended
</pre>

Yes

### Context

Followed documentation from [v0.109.0](https://github.com/gohugoio/hugo/releases/tag/v0.109.0) relase notes as well as https://discourse.gohugo.io/t/initialize-sass-variables-from-hugo-templates/42053/1 for LibSass transpiler

**Cannot use css functions such as `lighten()` and `darken()`, if the value comes transpiled from config file, because it's not a color**

Setup is as follows…

config.yaml
```yaml
params:
  style:
    color_palette_white: '#fff'
```

style.sass
```css
@import "bulma/bulma"
@import "mystyle/core"
```

mystyle/core.scss
```css
@import "hugo:vars";

@import 'partials/something';
// and many more .scss imports
```

mystyle/partials/_something.scss
```css
.something {
  border-bottom: 1px solid lighten($color-palette-white, 5%);
}
```

error log
```
Start building sites … 
hugo v0.110.0-e32a493b7826d02763c3b79623952e625402b168+extended
Error: Error building site: TOCSS: failed to transform "style.sass" (text/x-sass): "/path/themes/my-theme/assets/mystyle/partials/_something.scss:101:34": argument `$color` of `lighten($color, $amount)` must be a color
Built in 266 ms
```

acclassic:
Hey there

This one took me a bit to figure out. I thought the problem was the Sass Compiler or the godartsass package but it turned out to be a problem on how we created the varsstylesheet map that is then passed to the compiler to import. All variables were quoted and therefore sass would interpret them as string. Now we pass some of these variables without quotes so that the sass compiler can do the correct type casting. 
In the cssValues.go file I added all the identifier for the different css values. Now we can use them in functions and even pass functions like calc(), rgb(), etc.
I added a new test to check the type casting. All tests are passed other than the TestTransformErrors() function. I checked this with a clean clone of hugo and the error is still present there.

Fixes #10632


bep:
Thanks for this. I don't doubt that you describe a real problem, but I'm a little fearful about pulling parts of the Sass spec into Hugo (how do we know it's complete and how do we maintain it?)

I was rather hoping that Sass' `unquote` function would work as advertised.

* I'm pretty sure I have tested the simple cases and that `unquote("24px")` gets converted correctly.
* Is this PR covering the more complex cases, e.g. `rgb(255, 255, 255)`?

acclassic:
You're welcome. 

I gave it another look and you are not completely wrong either. It seems like that all the variables passed to the Sass compiler are passed as string and interpreted like that even pixel values or numbers. So when using Sass specific functions like darken(), etc. it will throw an error because the variable is of type string. When the Sass file then is converted to Css the values will be casted with the correct type. That is why you can use them in Css functions like calc(). 
I created a showcase for this here https://github.com/acclassic/hugo-issue-10632-test. If you run hugo with the -v flag the debug function will showcase how Sass interprets the variables.

Your fear about implementing this directly in hugo is justified. Just as a clarification the specs come directly from Css and not Sass. Basically I took all the color specs from here https://developer.mozilla.org/en-US/docs/Web/CSS/color_value other than the experimental one. And all the identifiers for dimensions from here https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Values_and_Units#dimensions. This should only change when new features are added directly to Css. 

Whit this PR you are also able to pass rgb(), hsl() and hwb() and use them in Sass functions. You could even pass a calc(x + y) directly from the config file if you wish. I added this so it would cover everything.

PR #10741, opened 2023-02-22
resources/sass: Remove debug statements

jmooring:
Closes #10470


bep:
Good catch, thanks ....

Issue #10737, opened 2023-02-21
Hide the --theme and -themesDir flags (and deprecate them)


willfaught:
`hugo -h` says in part:

```
  -t, --theme strings              themes to use (located in /themes/THEMENAME/)
      --themesDir string           filesystem path to themes directory
```

However, `--theme ../..` works, and has the same result as `--themesDir ../..`, at least on Windows.

The help message for `--theme` should reflect this.

Arguably, `--themesDir` should be deprecated and hidden, since it's redundant.

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.110.0+extended darwin/amd64 BuildDate=unknown
</pre>

### Does this issue reproduce with the latest release?

Yes

ginaesps:
Hello! Saw this good first issue and took the liberty of creating this PR. Would really appreciate feedback on it @bep @willfaught 
#10761

jmooring:
Given this project structure:

```text
./
├── content/
│   ├── posts/
│   └── _index.md
├── my-themes/
│   ├── theme-a/
│   ├── theme-b/
│   └── theme-c/
└── config.toml
```

To try out a theme I need to do:

```
hugo server --themesDir my-themes --theme theme-b
```

So, not sure we want to do this.

willfaught:
I think we need at least --themesDir to run exampleSites, etc. I don't personally use --theme, but I see the value of it. I might have used it when first trying out a couple themes, before building my own.

@bep What's the rationale for removing both flags?

bep:
>@bep What's the rationale for removing both flags?

That was maybe a bit ... rash. But my _thinking_ was that ...

* We have too. many flags. Server flags should be reserved for _tuning on runtime things_ (e.g. enable logging). If you really want to change theme while running (I suspect that is a rare thing), you can do `HUGO_THEME=foo hugo server` or something.
* But also, as we have introduced new and (sometimes) better ways to do things in Hugo, I really want to consolidate the old way with the new to avoid confusing. And now we have `theme` (which can be a slice of themes) and `module.imports`, which is the same thing but with many more options.

So, to the original title of this issue: We do need `themesDir` (the option, I'm not sure about the flag); that you happen to not strictly need it in all cases, that may be true, but in most setups, we do.

willfaught:
So, instead we would have this?

```sh
$ cd sometheme
$ cd exampleSite
$ HUGO_THEMES_DIR=../.. hugo server
```

[...]

PR #10761, opened 2023-02-28
Hide and deprecate --theme and --themesDir Flags

ginaesps:
commit 9030beaeac446d7955ad65ea84280c711d1b88f2
Author: ginaesps <respinosa7@ucol.mx>
Date:   Mon Feb 27 21:19:05 2023 -0600

    commands: Hide and deprecate --theme and --themesDir

    Hide and deprecate --theme and --themesDir flags, taking into consideration that (at least at Windows), the former gives a similar output to the latter one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
-
+.idea
 *.test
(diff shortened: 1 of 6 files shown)


bep:
This needs to wait a little, see the issue.

jmooring:
Closing per https://github.com/gohugoio/hugo/issues/10737#issuecomment-1536645733.

PR #10762, opened 2023-02-28
Modify inner variable shortcode-template explanation

ginaesps:
Added "without the closing TAG by using the self-closing syntax" so that it is more comprehensible. Based on videos and blogs regarding the topic to make sure it makes sense.


bep:

ginaesps:
Great! Thanks for the feedback.
Already created the PR on hugodocs
https://github.com/gohugoio/hugoDocs/pull/1985
Would really appreciate your comments on this @bep

PR #10808, opened 2023-03-08
PoC: hugo.mod as module file

dionysius:
- Works with hugo.mod and go.mod as in `TestHugoModFileV5` and `TestHugoModWithGoModFileV5`
- `TestHugoModFileV6` fails because of the `/v6` "package version suffix" and go does some extra checks, that it probably can find such a module exactly as written in the go.mod file. If such suffix is omitted, the download works, as shown in `TestHugoModFileV6WithoutPackage`
```
go get --modfile=hugo.mod github.com/bep/hugo-mod-with-hugodotmod
go: added github.com/bep/hugo-mod-with-hugodotmod v5.1.1-0.20230308145733-c3e39ab6de5b+incompatible
```
- There will still be some more work needed to continue this path of reading hugo.mod from transitive packages
- `c.GoModulesFilename` should probably be used without the full path. I used this one to check for the "current active module" and whether I need to append `-modfile` to go commands or hacking a temporary go.mod file.

### DIFF
--- a/commands/hugo.go
+++ b/commands/hugo.go
@@ -31,6 +31,7 @@ import (
 	"time"
 
 	"github.com/gohugoio/hugo/hugofs/files"
+	"github.com/gohugoio/hugo/modules"
 	"github.com/gohugoio/hugo/tpl"
 
 	"github.com/gohugoio/hugo/common/herrors"
@@ -942,7 +943,7 @@ func (c *commandeer) handleEvents(watcher *watcher.Batcher,
 		isConfig := configSet[ev.Name]
 		configChangeType := configChangeConfig
 		if isConfig {
-			if strings.Contains(ev.Name, "go.mod") {
+			if modules.IsGoMod(ev.Name) {
 				configChangeType = configChangeGoMod
 			}
 			if strings.Contains(ev.Name, ".work") {
--- a/modules/client.go
+++ b/modules/client.go
@@ -64,9 +64,11 @@ const (
 const vendord = "_vendor"
 
 const (
-	goModFilename = "go.mod"
-	goSumFilename = "go.sum"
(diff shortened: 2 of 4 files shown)


bep:
> If such suffix is omitted, the download works, as shown in TestHugoModFileV6WithoutPackage

Thanks for this, I will look at in more detail, but I suspect the above means that it's not actually reading the `hugo.mod` from the transitive dependency. This would be more clear if we added another dependency below that (I will try to do that sometime later today).

dionysius:
Yes, I'm quite sure it does not read hugo.mod, I never expected it to do so. This is exactly the way and the progress of the discovery for this approach so far I've written in the issue linked.

Hugo would now need to parse that dependencies hugo.mod by himself, decide to accept it (e.g. module name match), listing its dependencies with this module pkg cache as working dir and the go getting all those with the hugo site working dir (optimally also making a separate block in hugo.mod).

If multiple dependencies link to the same transitive dependency but in different versions, hugo graph would need to list both, but in hugo.mod would only one be listed. There is also more work for hugo to do bere, as you don't get this part for free from go amymore.

And it would introduce that you can't use this "package version suffix" (how ever its officially called) if you go the hugo.mod route.

bep:
>Hugo would now need to parse that dependencies hugo.mod by himself, 

Sorry for the delay. The above looks simple, but that's just the start of reimplementing Go Modules on the Hugo side, which is not something I'm prepared to take on. You have mentioned the caveats yourself (version resolution, major versioning would not work).

I have been thinkering about some other workarounds for this (e.g. create some temporary Go workspace file or something), but I'm not able to wrap my head around it.

**So, what if we skip the parts that's hard, and say that we allow `hugo.mod` files in the main project only (at least for now)**

PR #10813, opened 2023-03-10
Unclosed shortcode with inner content: improved error message

deining:
Following up on this [thread](https://discourse.gohugo.io/t/hugo-0-111-2-failure-with-unclosed-shortcodes/43368) in hugo discourse, this PR aims at improving the error message printed out when hugo encounters an unclosed shortcode with inner content. The message is now in line with the [wording](https://gohugo.io/templates/shortcode-templates/#inner) in the official hugo docs.


bep:
Thanks for this; I agree about this change, but there is a test assertion that also needs to be updated (see test failure).

PR #10815, opened 2023-03-10
Run gofmt -s on source files

deining:
This PR corrects format issues reported by the [go report card](https://goreportcard.com/report/github.com/gohugoio/hugo).


bep:
Thanks, these makes sense. I'll merge when it goes green. Note that you will probably find lots of other similar stuff to fix in Hugo (by running some linters); I would, however, appreciate to keep that type of "mass changes" to a minimum for the nearest future -- as I have some bigger change sets planned, and I want to minimise potential merge conflicts.

PR #10823, opened 2023-03-12
feat: parse rss/atom feeds from xml

nikaro:
This uses github.com/mmcdole/gofeed library to parse XML data and treats it as as RSS/Atom feed if the lib parser does not issue an error, otherwise fallback as "raw XML".

I'm not a developer so the code quality is probably poor, but what do you think of this? Would it be an acceptable feature?

The issue with raw XML for RSS/Atom feeds is that the implementation could vary a lot between websites and thus making it very difficult to write a template that would handle the different cases.


nikaro:
Here is an example of how it use then in my template:
```html
<details>
  <summary>
    Articles de blogs que je lis
  </summary>
  {{- range .Site.Params.Webring }}
  {{- warnf "fetching %#v" . }}
  {{- $feed := resources.GetRemote . | transform.Unmarshal }}
  {{- $article := index $feed.items 0 }}
  <h4>
    <a href="{{ $article.link }}" target="_blank" rel="noopener">{{ $article.title }}</a>
  </h4>
  <blockquote>
    {{ ($article.description | default $article.content) | htmlUnescape | plainify | truncate 500 }}
    <br>
    <cite>
      – <a href="{{ $feed.link }}">{{ $feed.title }}</a>
      ·
      {{ $article.published | dateFormat "2 Jan 2006" }}
    </cite>
  </blockquote>
  {{- end }}
</details>
```

nikaro:
Anyone?

parser/metadecoders/decoder.go
```diff
@@ -136,23 +137,30 @@ func (d Decoder) UnmarshalTo(data []byte, f Format, v any) error {
...
-		var xmlValue map[string]any
-		if err == nil {
-			xmlRootName, err := xmlRoot.Root()
-			if err != nil {
-				return toFileError(f, data, fmt.Errorf("failed to unmarshal XML: %w", err))
+		fp := feed.NewParser()
+		if feedData, feedErr := fp.Parse(bytes.NewReader(data)); feedErr == nil {
+			// convert feed parser data to a map
+			d, _ := json.Marshal(feedData)
+			json.Unmarshal(d, v)
```
chloe-the-catgirl, line 144:
```suggestion
			// We marshal and unmarshal the content, so that the feed data matches the target struct in v.
			var marshalled []byte
			marshalled, err = json.Marshal(feedData)
			if err != nil {
				return toFileError(f, data, fmt.Errorf("temporary JSON marshalling of feed into JSON: %w", err)
			}
			if err := json.Unmarshal(marshalled, v); err != nil {
				return toFileError(f, data, fmt.Errorf("unmarshalling temporary feed JSON: %w", err)
			}
```

I got a bit confused by the marshal/unmarshal pattern first, therefore I also added a comment for future readers.

chloe-the-catgirl (changes requested):
First of all, thanks for this PR, we could make use of it as well! 👍

And although I'm not a maintainer, I'd suggested a small fix for better error handling :)

PR #10847, opened 2023-03-17
resources/page: Allow section&taxonomy pages to have a permalink configuration

### LINKED ISSUE #8523: Add Permalink option for section pages
I have something like the following option in the config.toml:

```
[permalinks]
posts = "/new-posts/:filename/"
```

If I visit `http://localhost:1313/new-posts/post-name/` I get the intended page. But if I visit `http://localhost:1313/new-posts/` I don't get the list page, as I would expect. Instead, I get nothing at all.  I still have to go to `http://localhost:1313/posts/` to get the list page.

The version I'm using:

```
hugo v0.83.1+extended linux/amd64 BuildDate=unknown
```

I'm not sure if this behavior is intended. But it would be a good idea to have the ability to change the URL of the list pages using the permalinks option as well. Right now, I have to add the `url` parameter in the front matter of the `_index.md` to make the list page have the different url.

Mai-Lapyst:
Allows using permalink configuration for sections (branch bundles) and also for taxonomy pages. Extends the current permalink configuration to be able to specified per page kind while also staying backward compatible: all permalink patterns not dedicated to a certain kind, get automatically added for both normal pages and term pages.

Fixes #8523

Example config:
```toml
[permalinks]
  docs = "/docs/1.0/:sections[1:]/:filename"  # Will get added to both page & term permalinks
[permalinks.section]
  docs = "/docs/1.0/:sections[1:]"
[permalinks.page]
[permalinks.term]
[permalinks.taxonomy]
```


jmooring:
@Mai-Lapyst I did some quick testing and the first thing I found was...

If an _index.md does not exist for a top level section, the permalink setting is not applied. It is somewhat common to not create an _index.md for top-level content sections, and _very_ common to not create an _index.md file for taxonomy or term pages.

Mai-Lapyst:
@jmooring Thanks for the feedback.

Turns out that the if around the expand is completly redundant now. Removed it and it works like intended.

Also added some debug logging to verify that the permalink for any given file/section is set and to what value exactly.

jmooring:
> WARN 2023/03/18 17:12:36 Permalinks configuration overwrites pattern in kind 'page' for key 'books' with '/my-books2/:slug' (which was '/my-books/:slug')

Nice!

I spent a couple of hours testing this, and functionally it looks great. It behaves as expected:
- When `url` is set in front matter (trumps everything)
- Monolingual
- Multilingual
- Multilingual with permalinks per language
- etc.

FYI, when I run `v0.111.3` (no patch) with the new permalink settings in site config, it panics, but that's much better than creating new keys. I don't see that as a problem.

Mai-Lapyst:
Thanks!

> FYI, when I run `v0.111.3` (no patch) with the new permalink settings in site config, it panics, but that's much better than creating new keys. I don't see that as a problem.

I noticed that too; digged a bit into it and you can get the same crash even in the patched version, by simply using an empty string:
```
[permalinks]
    posts = ""
```

It seems that an unchecked string slice access seems to be the problem: https://github.com/gohugoio/hugo/blob/fb35e7f5e1e5f6f1823abf65c3220db2e1e0457b/resources/page/permalinks.go#L259-L260
Will add a patch for it as well.

jmooring:
@bep Functionally this is great, and scratches a long-term itch. Would appreciate a review if/when you have the chance.

bep:
OK, I guess this is what lots of people would want, and I like the config.

I will merge this if we could

1. Get this rebased against master and get this integrated into my new `allconfig.Config` struct.
2. Get a basic integration test up and running.

Mai-Lapyst:
@bep Thanks for looking over it. Have rebased it onto master and reworked it to work with the new allconfig. Also added a test for it; let me know if more tests are needed.

hugofs/files/classifier.go
```diff
@@ -93,6 +93,7 @@ const (
 	ContentClassBranch  ContentClass = "branch"
 	ContentClassFile    ContentClass = "zfile" // Sort below
 	ContentClassContent ContentClass = "zcontent"
+	ContentClassZero    ContentClass = "zero"  // Special value for zeroFile
```
bep, line 96: I don't think we need/want this. A zero file would have an empty ContentClass string, which is the natural zero value, no need to make one up.

hugolib/page__paths.go
```diff
@@ -143,22 +146,26 @@ func createTargetPathDescriptor(s *Site, p page.Page, pm *pageMeta) (page.Target
...
+		if strings.HasSuffix(opath, "//") {
+			// When rewriting the _index of the section the permalink config is applied to,
+			// we get douple slashes at the end sometimes; clear them up here
+			opath = strings.TrimSuffix(opath, "/")
 		}
 
+		desc.ExpandedPermalink = opath
+
+		if !p.File().IsZero() {
+			s.Log.Debugf("Set expanded permalink path for %s %s to %#v", p.Kind(), p.File().Path(), opath)
```
bep, line 165: Can you remove these debug statement, it makes the debug log very busy,.

config/allconfig/alldecoders.go
```diff
@@ -199,7 +199,48 @@ var allDecoderSetups = map[string]decodeWeight{
...
+
+					// To sucessfully be backward compatible, "default" patterns need to be set for both page and term
+					p.c.Permalinks["page"][k] = v;
+					p.c.Permalinks["term"][k] = v;
+				
+				case maps.Params:
+					// [permalinks.key]
+					//   xyz = ???
+
+					if (k == "page") || (k == "section") || (k == "taxonomy") || (k == "term") {
```
bep, line 224: I think you can use the constants in `pagekinds.Page` etc.

config/allconfig/alldecoders.go
```diff
@@ -199,7 +199,48 @@ var allDecoderSetups = map[string]decodeWeight{
 	"permalinks": {
 		key: "permalinks",
 		decode: func(d decodeWeight, p decodeConfig) error {
-			p.c.Permalinks = maps.CleanConfigStringMapString(p.p.GetStringMapString(d.key))
+			p.c.Permalinks = make(map[string]map[string]string)
+
+			p.c.Permalinks["page"] = make(map[string]string)
```
bep, line 204: pagekinds.Page etc.

resources/page/permalinks_test.go
```diff
@@ -69,6 +69,7 @@ func TestPermalinkExpansion(t *testing.T) {
 	page.date = d
```
bep, line 69: This all looks good, but I think we need a test that shows the new setup working. I suggest you create a new `permalink_integration_test.go` and follow the the cue of similar tests (using the NewIntegrationTest...) that tests the new permalink setup in a site build.

bep (changes requested):

bep:
Never mind the comments above. I suspect we need to test this out on some real projects and then do some adjustments/add some tests, so I might as well merge this early.

PR #10888, opened 2023-04-10
Add out-of-the-box LaTeX in Markdown support to Hugo

janhuenermann:
Hi, 

[from this thread](https://discourse.gohugo.io/t/katex-in-hugo/43274) in the forum a solution emerged to bring proper support for LaTeX in Markdown to Hugo. 

**Disclaimer:** credits for this feature go to @jbirnick, who added support for this originally [in his fork](https://github.com/jbirnick/katugo/commit/2ea6aad9b084e8d48eb5082e16e8bc46ce339744). I just opened this PR because the thread was about to go stale and this seems to be a feature that many people in the community would appreciate (including me).

The advantage of this approach over existing solutions is:
- no need for special syntaxes that are unique to Hugo. **Support standard LaTeX in Markdown out of the box. For example, any Markdown file that already renders on Github, will render now in Hugo.**
- allow rendering of inline LaTeX without shortcodes, which requires a lot of boilerplate

For more details, including previous solutions, I'd recommend to take a look [at the original forum thread](https://discourse.gohugo.io/t/katex-in-hugo/43274).

### How to enable

Put the following in your `config.toml`:

```toml
[markup.goldmark.extensions]
Katex = true
```

Then include the Katex client side stylesheets. Please refer to the [official installation guides](https://katex.org/docs/browser.html) from Katex.

### Testing

Tested manually locally and verified that `Katex = true` renders Katex in the HTML code. Disabling it leads to previous behaviour.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -84,6 +84,7 @@ require (
 	github.com/Azure/go-autorest/autorest/date v0.3.0 // indirect
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
+	github.com/FurqanSoftware/goldmark-katex v0.0.0-20221223060946-0bf781ec55b4 // indirect
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
 	github.com/aws/aws-sdk-go-v2 v1.9.0 // indirect
 	github.com/aws/aws-sdk-go-v2/config v1.7.0 // indirect
@@ -94,6 +95,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sso v1.4.0 // indirect
 	github.com/aws/aws-sdk-go-v2/service/sts v1.7.0 // indirect
 	github.com/aws/smithy-go v1.8.0 // indirect
+	github.com/bluele/gcache v0.0.2 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
 	github.com/dlclark/regexp2 v1.7.0 // indirect
 	github.com/go-openapi/jsonpointer v0.19.5 // indirect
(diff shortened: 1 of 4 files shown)


janhuenermann:
If anyone has concerns regarding [quickjs](https://github.com/lithdew/quickjs) using _cgo_,  I also came up with a solution that doesn't depend on any external packages/C code. It's here, would love to hear some feedback:  https://github.com/janhuenermann/hugo-katex/commit/ad382464769740c73235cbc3b20c4a5143408aa2

bep:
Thanks for this. I would appreciate opening up an issue discussing this before creating a PR.

I'm closing this because of that, but additionally:

* This comes with some dependencies I'm not sure about, but the blocker is `QuickJS`, which requires CGO. I know there are 2 other examples of that in Hugo (libwebp, libsass), but the current policy is that we consider this for policies that we _must have_. LateX doesn't fall into that category.
* Also it seem to be fairly slow (see benchmark below). I see a cache is in place, but I suspect that this feature does not cache very well (I'm guessing most expressions in a site to be different).


```
BenchmarkRender-10    	      61	  18840946 ns/op	    6504 B/op	      30 allocs/op
```

janhuenermann:
Hi @bep,

Thanks for taking the time to look at the PR and provide feedback! I'm sorry for not opening an issue for discussion first. I didn't know that this is part of the contribution best practices in this repo. I'll open an issue after this response.

I completely agree with you. These are reasonable concerns.  However, considering the high demand for this feature among the scientific blogging community, I believe it's worth exploring alternative implementations that don't require CGO. In fact, I've found a solution that enables LaTeX support in pure Go without any external dependencies (see https://github.com/janhuenermann/hugo-katex/commit/ad382464769740c73235cbc3b20c4a5143408aa2). I think it's worth considering this, as it would address the dependency concern.

As for the performance, you're right that it could be an issue for large sites. However, I want to emphasize that this feature is entirely optional and disabled by default. Therefore, users who don't enable it won't experience any performance hit. Also, I think the performance hit in this implementation mainly comes from executing the JavaScript code for SSR of the expressions, which is not necessary for the alternative implementation I mentioned. 

Thanks again! I'll open an issue now.

PR #10897, opened 2023-04-14
tpl/urls: Add JoinPath template function

### LINKED ISSUE #9694: Add support for Go's net/url JoinPath
I ran into an issue trying to use `path.Join` to join a path to a Permalink (I'm new to Hugo and Go): `path.Join` uses `filepath.Join` under the hood, which calls `filepath.Clean`. Among other things, `Clean` will remove any consecutive slashes, clobbering `https://` in `.Permalink`, making it for example `https:/example.com/path/to/a/thing`. Go's `path.Join` also does the same thing with its own implementation of `Clean`.

I dug around and found there's a [very recently added](https://github.com/golang/go/issues/47005) (March 9th) `JoinPath` function in `net/url`. I think it may have missed Go 1.18 by 5 days: it would be a convenient addition to Hugo on the next release of Go.

jmooring:
See https://pkg.go.dev/net/url#JoinPath

Closes #9694

### DIFF
--- a/docs/content/en/functions/urls.JoinPath.md
+++ b/docs/content/en/functions/urls.JoinPath.md
@@ -1,0 +1,24 @@
+---
+title: urls.JoinPath
+description: Joins the provided elements into a URL string and cleans the result of any ./ or ../ elements.
+categories: [functions]
+menu:
+  docs:
+    parent: functions
+keywords: [urls,path,join]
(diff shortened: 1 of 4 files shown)


carlmjohnson:
Won't this panic if it gets no arguments? Would it be better to have an error here?

PR #10909, opened 2023-04-22
langs/i18n: Fallback to defaultContentLanguage instead of English

### LINKED ISSUE #9216: Hardcoded default lang?
Is it possible to NOT use English as default fallback language, or to not use it at all? We have multiple languages on our site, but we don't have English. For that reason we don't have en.toml, just sr.toml and hu.toml. The default lang is "sr", but the fallback for the missing translations are not working. I'm not a Go developer, but I can see that English is kinda must-have lang at this line:
https://github.com/gohugoio/hugo/blob/75a823a36a75781c0c5d89fe9f328e3b9322d95f/langs/i18n/translationProvider.go#L51

The tests are passing, because they are using "en" as `defaultContentLanguage`, but that gives a "fake" success. Changing the `defaultContentLanguage` to some other lang (and also replacing "en" to that lang everywhere in the tests) results to failing tests...

jmooring:
Co-authored-by: 641bill <wo23636@126.com>

Fixes #9216

### DIFF
--- a/langs/i18n/integration_test.go
+++ b/langs/i18n/integration_test.go
@@ -103,3 +103,41 @@ i18n: {{ i18n "a" . }}|
 	i18n: Reading time: 3|
 	`)
 }
+
+// Issue 9216
+func TestI18nDefaultContentLanguage(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- config.toml --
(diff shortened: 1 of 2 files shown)


edhenderson:
This is great, thank you. Forgive my newness to the project, how long till this is available as a release? Please and cheers.

Issue #10434, opened 2022-11-13
Add shortcode for render Mastodon toots


andriisoldatenko:
WDYAT?

olikami:
Should be easy enough to implement for anyone, you just need server, user and tootID:

    <iframe src="https://<server>/@<user>/<tootID>/embed" class="mastodon-embed" style="max-width: 100%; border: 0" width="400" allowfullscreen="allowfullscreen"></iframe>

weddige:
https://github.com/gohugoio/hugo/pull/10920#issuecomment-1685302079
> The proposal (https://github.com/gohugoio/hugo/issues/10434) has not been accepted.

What can be done, to move this issue forward? As Hugo comes with shortcodes for Twitter, Instagram and various other social media platforms, it would be only appropriate to also have one for Mastodon.

I'll happily help implementing the feature. If this is not desired at the moment. Until there is a solution in Hugo I'll continue to use my implementation from #11370 and can push an updated version if necessary.

jmooring:
Although it's not my decision to make, I am not in favor of adding any more shortcodes. Instead, I would prefer to deprecate most of the built-in shortcodes, and provide examples in documentation. 

With the introduction of Hugo Modules (implemented long after the built-in shortcodes), it is trivial to include shortcode libraries in a project.

andriisoldatenko:
> Although it's not my decision to make, I am not in favor of adding any more shortcodes. Instead, I would prefer to deprecate most of the built-in shortcodes, and provide examples in documentation.
> 
> With the introduction of Hugo Modules (implemented long after the built-in shortcodes), it is trivial to include shortcode libraries in a project.

OH, it's a good idea, can you please point me to where in the docs should I add an example? and maybe some links for Hugo Modules?

jmooring:
It it trivial for anyone to create their own shortcode based on Mastadon's embed code:

<details>
<summary>layouts/shortcodes/mastadon</summary>

```text
{{- with .Get 0 }}
  <iframe
    src="{{ . }}/embed"
    class="mastodon-embed" style="max-width: 100%; border: 0"
    width="400"
    allowfullscreen="allowfullscreen">
  </iframe>
  <script src="https://fosstodon.org/embed.js" async="async"></script>
{{- else }}
  {{- errorf "The %q shortcode requires a positional parameter: the url of the post. See %s" .Name .Position }}
{{- end }}
```
</details>
<br>

Then call it with:

```text
[...]

PR #10920, opened 2023-04-28
Add mastodon shortcode

andriisoldatenko:


andriisoldatenko:
@bep could you please take a look?

tpl/tplimpl/embedded/templates/shortcodes/mastodon.html
```diff
@@ -0,0 +1,15 @@
+{{- $pc := .Page.Site.Config.Privacy.Mastodon -}}
+{{- if not $pc.Disable -}}
+  {{- $msg1 := "The %q shortcode requires three named parameters: server, user and id. See %s" -}}
+  {{- $server := .Get "server" | default (.Get 0) -}}
+  {{- $user := .Get "user" | default (.Get 1) -}}
+  {{- $tootID := .Get "tootID" | default (.Get 2) -}}
+  {{- if and $server $user $tootID -}}
+<div style="position: relative; overflow: hidden; height: auto;">
+  <iframe src="https://{{ $server }}/@{{ $user }}/{{ $tootID }}/embed" class="mastodon-embed" style="border: 0px; max-width: 550px; width: 100%;" allowfullscreen="allowfullscreen"></iframe>
+  <script src="https://{{ $server }}/embed.js" async="async"></script>
```
weddige, line 10: Dynamic inclusion of JavaScript from a foreign (and currently completely unknown) domain is a security risk. Especially if it is included in the default functionality of Hugo. Users using this shortcode will assume that it embeds the toot in a (reasonably) secure way.

config/services/servicesConfig.go
```diff
@@ -60,6 +61,14 @@ type Instagram struct {
 	AccessToken string
 }
 
+// Mastodon holds the functional configuration settings related to the Mastodon shortcodes
+type Mastodon struct {
+	// The Simple variant of Mastodon is decorated with a basic set of inline styles.
```
weddige, line 66: Is the inline CSS really necessary? And if so, wouldn't it be easier to override the template instead of having a config option?

config/privacy/privacyConfig.go
```diff
@@ -64,6 +65,15 @@ type Instagram struct {
 	Simple bool
 }
 
+// Mastodon holds the privacy configuration settings related to the Mastodon shortcode.
+type Mastodon struct {
+	Service `mapstructure:",squash"`
+
+	// TODO: revisit it
+	// If simple mode is enabled, a static and no-JS version of the Toot will be built.
```
weddige, line 73: This is obviously not implemented yet. What is the idea here? To load `https://{{ $server}}/api/v1/statuses/{{ $tootID }}/` and statically render something based on it?

Should this be a fully styled layout or has it to be overridden by the user?

weddige:
What is the status of this PR? I overlooked this PR when creating #11370 and will happily contribute here, if it helps to get this merged quicker.

jmooring:
See <https://github.com/gohugoio/hugo#contributing>

The proposal (https://github.com/gohugoio/hugo/issues/10434) has not been accepted.

andriisoldatenko:
@weddige I'm here and ready to fix my PR; the problem for now is that we need to wait for the acceptance of the proposal.

https://github.com/gohugoio/hugo/issues/10434

config/privacy/privacyConfig.go
andriisoldatenko: @weddige thanks for review, I copied it from another structs, so we can potentially remove it.

config/services/servicesConfig.go
andriisoldatenko: Again here, I just follow the same pattern as other services like Twitter or Instagram.

jmooring:
Closing per <https://github.com/gohugoio/hugo/issues/10434#issuecomment-2033349559>.

PR #10927, opened 2023-05-04
tpl/tplimpl: Add img loading attribute to figure shortcode 

InLaw:
to make it possible to e.g. to lazy load an figure/image

### DIFF
--- a/docs/content/en/content-management/shortcodes.md
+++ b/docs/content/en/content-management/shortcodes.md
@@ -117,6 +117,9 @@ height
 width
 : `width` attribute of the image.
 
+loading
+: `loading` attribute of the image.
+
 attr
 : Image attribution text. Markdown within the value of `attr` will be rendered.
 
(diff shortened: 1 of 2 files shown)


bep:
@jmooring this looks good to me, but I don't have strong opinions about this either way.

jmooring (changes requested):
There are already 12 parameters for the built-in figure shortcode: `alt`, `attr`, `attrlink`, `caption`, `class`, `height`, `link`, `rel`, `src`, `target`, `title`, and `width`.

This PR adds 7 new parameters, rendered as `img` element attributes:

- `ismap`: very rare
- `loading`: seems useful
- `longdesc`: deprecated
- `referrerpolicy`: rare
- `size`: invalid
- `srcset`: typically derived via image processing in renderhook, partial, or custom shortcode
- `usemap`: very rare

Remove everything but the loading attribute, and update the documentation (docs/content/en/content-management/shortcodes.md).

Issue #10950, opened 2023-05-17
Use uber-go/automaxprocs to default GOMAXPROCS aware of Linux CPU Quotas


BenTheElder:
See:
- https://github.com/golang/go/issues/33803
- https://github.com/kubernetes/website/pull/41142

Basically: When using hugo in a container with limited CPU quota, hugo can encounter CPU throttling issues because GOMAXPROCS is defaulted to the underlying hardware core count instead of the allocated CPU limits in the container.

There's something of a defacto standard library for defaulting this in a CPU limit aware way in https://github.com/uber-go/automaxprocs, at least until https://github.com/golang/go/issues/33803 is solved.

Users would still be able to override by setting `GOMAXPROCS` env, and this is a fairly small MIT licensed dependency.

bep:
A note from the sideline, Hugo bases its parallelism on

1. HUGO_NUMWORKERMULTIPLIER env var if set
2. `runtime.NumCPU()`

So, setting `HUGO_NUMWORKERMULTIPLIER=1` (or something) could be a workaround.

BenTheElder:
Ack, thanks!

We're getting good results setting GOMAXPROCS=1 in netlify for Kubernetes (it appears free netlify has ~1 core allocated?)

> runtime.NumCPU()

I think it might? be better to base it on `runtime.GOMAXPROCS(0)` (<1 does not set it and only returns the current value).

Even without that though, AIUI the setting GOMAXPROCS will let go internally do better at scheduling within the CPU quota. 

I filed a sample PR in https://github.com/gohugoio/hugo/pull/10951 for the latter, we've seen automaxprocs fix performance issues with containerized Go more generally, e.g. in Kubernetes's CI unit tests.

bep:
I needed to revert this; a new patch that makes the tests pass is welcomed, but I have not time to investigate this at the moment.

BenTheElder:
I'll look into that 👀

Thank you for all your work on hugo!

bep:
Great; Ideally I would want _no output_ from that library.

BenTheElder:
OK, sorry about that, the root automaxprocs just [calls `maxprocs.Set(macprocs.Logger(log.Printf))`](https://github.com/uber-go/automaxprocs/blob/f6364fccf14450d89bc0bf13c2aa1c05d9f7aefd/automaxprocs.go#L32), if we call `maxprocs.Set()` instead there will be no logging this time (per the godocs, checking the implementation, and local testing).

[...]

PR #10952, opened 2023-05-17
add automaxprocs to default GOMAXPROCS to be CPU limit aware

BenTheElder:
fixes https://github.com/gohugoio/hugo/issues/10950

This implementation does NOT log unlike importing the root automaxprocs package in #10951 

`go test ./...` passes locally on this PR, I can reproduce the failures on #10951 related to undesired logging output.


main.go
```diff
@@ -17,10 +17,14 @@ import (
...
 	"os"
 
+	"go.uber.org/automaxprocs/maxprocs"
+
 	"github.com/gohugoio/hugo/commands"
 )
 
 func main() {
+	// default GOMAXPROCS to be CPU limit aware, still respecting GOMAXPROCS env
+	maxprocs.Set()
```
bep, line 27: Could you 

1. Move this into commands.Execute so it actually gets executed during the integration tess.
2. Make the comment into a proper sentence starting with a capital D and ending with a period.
3. Adjust the commit message title in line with the guidelines in this repo (start out with a capital A)

bep (changes requested):

main.go
bep: Also, it would be good if you could rebase it against the recently pushed master.

BenTheElder: All three should be done.

BenTheElder:
So far I can't repro the CI test failure locally with go1.20.4 on latest macOS. `go test -race ./... -tags extended` passes.

PR #10967, opened 2023-05-19
tpl/urls: Return empty string when JoinPath has zero args

jmooring:


tpl/urls/urls.go
```diff
@@ -190,13 +190,15 @@ func (ns *Namespace) AbsLangURL(s any) (template.HTML, error) {
 // of any ./ or ../ elements.
 func (ns *Namespace) JoinPath(elements ...any) (string, error) {
 
+	if len(elements) == 0 {
+		return "", errors.New("one or more arguments required")
```
bep, line 194: I think we should get this behaviour in line with `path.Join` -- just checked, and it returned an empty string in this case.

PR #10991, opened 2023-05-22
Add page content in RSS by default

kirillbobyrev:
Fixes: https://github.com/gohugoio/hugo/issues/4071


jmooring:
Duplicate of https://github.com/gohugoio/hugo/pull/10589, which includes a config option as specified in the https://github.com/gohugoio/hugo/issues/4071.

PR #11032, opened 2023-05-27
tpl/tplimpl: Use .Language.LanguageCode in built-in templates

jmooring:
If this is merged, I will update the documentation in the docs repo after release.

### DIFF
--- a/tpl/tplimpl/embedded/templates/_default/rss.xml
+++ b/tpl/tplimpl/embedded/templates/_default/rss.xml
@@ -16,8 +16,8 @@
     <title>{{ if eq  .Title  .Site.Title }}{{ .Site.Title }}{{ else }}{{ with .Title }}{{.}} on {{ end }}{{ .Site.Title }}{{ end }}</title>
     <link>{{ .Permalink }}</link>
     <description>Recent content {{ if ne  .Title  .Site.Title }}{{ with .Title }}in {{.}} {{ end }}{{ end }}on {{ .Site.Title }}</description>
-    <generator>Hugo -- gohugo.io</generator>{{ with .Site.LanguageCode }}
-    <language>{{.}}</language>{{end}}{{ with .Site.Author.email }}
+    <generator>Hugo -- gohugo.io</generator>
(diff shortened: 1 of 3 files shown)


tpl/tplimpl/embedded/templates/_default/rss.xml
```diff
@@ -16,8 +16,8 @@
     <title>{{ if eq  .Title  .Site.Title }}{{ .Site.Title }}{{ else }}{{ with .Title }}{{.}} on {{ end }}{{ .Site.Title }}{{ end }}</title>
     <link>{{ .Permalink }}</link>
     <description>Recent content {{ if ne  .Title  .Site.Title }}{{ with .Title }}in {{.}} {{ end }}{{ end }}on {{ .Site.Title }}</description>
-    <generator>Hugo -- gohugo.io</generator>{{ with .Site.LanguageCode }}
-    <language>{{.}}</language>{{end}}{{ with .Site.Author.email }}
+    <generator>Hugo -- gohugo.io</generator>
+    <language>{{ or site.Language.LanguageCode site.Language.Lang }}</language>{{ with .Site.Author.email }}
```
bep, line 20: `site.Language.LanguageCode` will fall back to `Lang` if `LanguageCode` is not set. See

https://github.com/gohugoio/hugo/blob/ffdbce5787a45e8537b7a77515ca06e4deee2504/langs/language.go#L111

bep (changes requested):

tpl/tplimpl/embedded/templates/_default/rss.xml
bep: Note that the above was also how `site.LanguageCode` behaved before.

jmooring: Changes made as requested. 

I like that `site.Language.LanguageCode` falls back to `site.Language.Lang`.

Note: I verified that is new behavior. In v0.111.3, site.LanguageCode returned an empty string if languageCode was not set in site configuration. Thanks for the change.

PR #11081, opened 2023-06-10
helpers: Improve schema detection when creating relative URLs

### LINKED ISSUE #11080: relURL and relLangURL ignore schemas other than http/https
Example:

```
{{ relURL "ftp://exampl.org/file.txt" }} --> /ftp:/example.com/file.txt
```

Reference: <https://discourse.gohugo.io/t/44756>

jmooring:
Fixes #11080


bep:
With the very basic benchmark in 5db215d4d6e4a0e353ea9aa21441d9ca4d7caf12 

```
BenchmarkRelURL-10    	 6678015	       174.5 ns/op	     144 B/op	       1 allocs/op
BenchmarkRelURL-10    	 7588068	       161.8 ns/op	     144 B/op	       1 allocs/op
BenchmarkRelURL-10    	 7505914	       158.2 ns/op	     144 B/op	       1 allocs/op
BenchmarkRelURL-10    	 7570666	       157.2 ns/op	     144 B/op	       1 allocs/op
PASS
ok  	github.com/gohugoio/hugo/helpers	5.807s


name       old time/op    new time/op    delta
RelURL-10    14.6ns ± 2%   162.9ns ± 7%  +1012.50%  (p=0.029 n=4+4)

name       old alloc/op   new alloc/op   delta
RelURL-10     0.00B        144.00B ± 0%      +Inf%  (p=0.029 n=4+4)

name       old allocs/op  new allocs/op  delta
RelURL-10      0.00           1.00 ± 0%      +Inf%  (p=0.029 n=4+4)
```

I agree that we need to fix this, but I suggest that you do some basic prefix checking before falling back to the "slow path" using url.Parse.

jmooring:
OK, but this change was modeled after:
<https://github.com/gohugoio/hugo/blob/master/helpers/url.go#L106>

Would you like me to change that too?

bep:
Never mind, I'll take a look at both of them.

PR #11082, opened 2023-06-10
cache: Set default cache path based on $USER

### LINKED ISSUE #7391: user /tmp/hugo_$USER instead of /tmp/hugo_cache
Using 0.72 compiled from master from 2 days ago on Ubuntu 16.04 with go 1.14. Hugo by default uses /tmp/hugo_cache to store temp files. This is a problem on a multiuser system like ours. For example, one user runs `hugo server` on their project folder. Another user attempt to run `hugo` and fails as it can't write to `/tmp/hugo_cache` being owned by the previous user.

I know it can be overcome by `--cacheDir` but it would be better to have hugo by default use `/tmp/hugo_$USER` instead of the hardcoded `/tmp/hugo` (better never to hardcode things in general anyway). Or make it a random unique dir if it's easier for cross-OS compatibility, making sure gc is doing its job.

Sircular:
Change the default cache directory to `$TMPDIR/hugo_cache_$USER`, so that multi-user systems do not have caches that interfere with each other.  The other cache-choosing logic (e.g. Netlify exceptions, configuration options) are not affected.

Fixes #7391

### DIFF
--- a/commands/commandeer.go
+++ b/commands/commandeer.go
@@ -521,7 +521,7 @@ Complete documentation is available at https://gohugo.io/.`
 func applyLocalFlagsBuildConfig(cmd *cobra.Command, r *rootCommand) {
 	cmd.Flags().StringSliceP("theme", "t", []string{}, "themes to use (located in /themes/THEMENAME/)")
 	cmd.Flags().StringVarP(&r.baseURL, "baseURL", "b", "", "hostname (and path) to the root, e.g. https://spf13.com/")
-	cmd.Flags().StringP("cacheDir", "", "", "filesystem path to cache directory. Defaults: $TMPDIR/hugo_cache/")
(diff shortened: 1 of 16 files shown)


bep:
Thanks for this.

I can merge this, but

* You need to update the flag documentation in the code (https://github.com/gohugoio/hugo/blob/73779707a2d5fefd4915f8ee2021f265d841dea2/commands/commandeer.go#L524) and not in the auto generated MD files.
* Also, I think it's `USERNAME` on Windows? So maybe check if that's the case, and add a check for that if `USER` is not found.

jmooring:
On Windows 11, the cache directory is within the user's profile, and it's been that way since Windows 7 (and maybe earlier). For example:

```
C:\Users\joe\AppData\Local\Temp\hugo_cache
```

So if $USER doesn't exist, don't bother appending the user name.

bep:
OK, I'll merge -- the MD files will be overwritten anyhow.

Issue #10857, opened 2023-03-21
Advanced Math Functions for Hugo Templates


marcpabst:
I suggest the addition of advanced math functions to the Hugo framework. These functions would be accessible in templates, enabling users to perform complex mathematical operations more easily. The proposed functions from https://pkg.go.dev/math to be added are:

    math.Acos
    math.Asin
    math.Atan
    math.Cos
    math.Sin
    math.Tan
    math.Pi

Hugo currently supports basic arithmetic operations in templates, but lacks support for more advanced mathematical operations. By adding these functions, we can enhance Hugo's capabilities and make it more versatile for users working with complex math in their projects. 

A possible use case could involve a Hugo template displays an interactive map using map tiles from an API like OpenStreetMap or Mapbox. By pulling map tiles at deploy time, we could avoid calls to external APIs by the clients. To identify the map tiles that are needed for a certain geographic area, trigonometric calculations are needed. These calculations are fairly simple but involve the above-mentioned functions. 

This proposal introduces new template functions and does not modify existing functionality. As such, it is fully backwards compatible and will not impact existing projects or require any changes to existing templates.

Discussed on Discourse here: https://discourse.gohugo.io/t/more-advanced-math-functions/43539/1

raoulb:
Closed by #12660

jmooring:
Implemented in https://github.com/gohugoio/hugo/commit/9d2b5f98d0efa06d5faf1a7596a53051bc1e088c.

Issue #10091, opened 2022-07-15
Ability to parse time string with custom layout


BenSouchet:
## The issue

Currently there is no way to parse date string(s) with custom layout, like the French one `DD/MM/YYYY`, the Spanish one `DD-MM-YY`, the Finnish one `DD.MM.YYYY`.

## Proposal

A hugo time method to parse date string with a provided layout like this:
```go
{{ time.ParseWithFormat FORMAT INPUT [TIMEZONE] }}

{{ time.ParseWithFormat "02/01/2006" "15/03/2021" }} → "2021-03-15T00:00:00Z"
```

## How to

Internally this would use the Go time function `ParseInLocation` like this:
```go
time.ParseInLocation("02/01/2006", "15/03/2021", loc)
```

Go Playground example: https://go.dev/play/p/gsSIgLgAYLL

bep:
>Currently there is no way to parse date string(s) with custom layout,

That's not entirely true. `time.Format` uses the location as defined per language in your site config.

We could consider adding a `timeZone` to `time.Format`, though. Would that work for you?

BenSouchet:
Currently I cannot use `time.Format` with date input string like `15/03/2021`.

If I want to convert a French date string to ISO 8601 format with `time.Format`, like this:
```{{ time.Format "2006-01-02" "15/03/2021" }}```
I get:
```error calling Format: unable to parse date: 15/03/2021```

Even if in my config file I have set:
```
languageCode: "fr-FR"
defaultContentLanguage: "fr"
timeZone: "Europe/Paris"
```

This is why if we had the ability to pass the **custom layout** used by the date string we have, we could parse and convert it to ISO 8601 format.

**Comment by BenSouchet** [comment]
**Idea**: Maybe we could have an optional parameter to `time.Format`:
```
time.Format OUTPUT_LAYOUT INPUT [INPUT_LAYOUT]
```
then we could do:
```go
{{ time.Format "2006-01-02" "15/03/2021" "02/01/2006" }} → "2021-03-15"
```

PR #11085, opened 2023-06-12
Proof of concept: Custom functions

jphastings:
I got the software equivalent of an earworm and decided to turn it into a proof of concept.

👀 I'm seeking feedback of any kind, which could include "this isn't something Hugo is interested in"; I'll be happy to turn this sorry state of affairs into quality, tested code, or to close this line of thinking down.

## What is it?

Code similar to this would allow Hugo users to write and use custom Typescript (or Javascript) functions for use inside their templates.

```go-template
{{ $today := (now.Format "Mon January 2" ) }}

{{ printf $today }} → Mon June 12            # Standard Hugo
{{ fn "enc.Rot13" $today ) }} → Zba Whar 12  # New!
```

See the `external.md` at the top of the PR for more detail.

## Can I test it out?

Yes!

1. Build this branch
2. Clone [my demo site](https://github.com/jphastings/hugo-functions-demo)
3. Edit `layouts/index.html` and `functions/*.ts`
4. Run `../path/to/built/hugo` in the site root
5. Check out `public/index.html` to see the output

## Why is this useful?

A lot of Hugo's Github Issues seem to be about [changing](https://github.com/gohugoio/hugo/issues/10091) or [adding](https://github.com/gohugoio/hugo/issues/10857) new functions to Hugo. Though contributing to Hugo's codebase is an admirable, it's a long-term solution to these requests and likely isn't available option to many.

Writing and sharing function files, storing them in the site files (eg. in `functions/*.ts`), is a cross-platform, simple, accessible way to be able to write and share helper functions.

Some important benefits:

- The JS runtime used is totally sandboxed (no access to the filesystem, web or similar), so malicious custom code can't do harm.
  - I'd like to be able to use a limited `fetch`, but there are lots of new challenges there.
- Unlike using something like Go plugin files, this TS/JS is platform agnostic (so it can be committed to source code along side the site files)
- Other JS modules _should_ be `import`-able*, so custom code writers can build on the shoulders of the JS community. (*Caveat: I haven't tried this, it'd likely be a little complex right now)
- JS execution happens entirely in Go (no v8 here!) and is time-limited. It'll be slower than Go, but not "slow". 

## Why this PR isn't at all ready

I've thrown this together to understand if it's interesting to the community. There are _lots_ of sharp edges. Some of the ones I've found/considered but not addressed:

- The error reporting if you've written bad TS/JS is plain bad or absent.
  - *Good dev tooling here will be crucial, or making these function files will be very hard*
- Go types are auto-cast to Javascript ones, which mostly works (with strings, numbers, structs, even times—though timezone data is lost) but there are undoubtedly arguments people could pass to a JS function that'd be super confusing to interact with.
- The function call is a little clunky (`{{ fn "filename.FunctionName" "args" }}` instead of `{{ filename.FunctionName "args" }}`), as Hugo seems to require namespaced functions to be _methods_ in Go and, because these are dynamically generated from JS, I can't do that (without substantial codechange)
  - This also means that the code I wrote to extract exported constants of the name `<FunctionName>Examples`, expecting it to be in the `[][2]string` format used for other function examples, is useless right now.
- While [Goja](https://github.com/dop251/goja) (the Javascript-VM-in-Go library used here) seems stable as an open source project, [go-typescript](https://github.com/clarkmcc/go-typescript) seems less well maintained.
- I'm re-compiling the JS every time it's called. I can definitely optimise this.

Thanks for reading this far! Let me know what you think.

### DIFF
--- a/docs/content/en/functions/external.md
+++ b/docs/content/en/functions/external.md
@@ -60,20 +60,12 @@ error calling fn: the function file named heLLo has not been loaded
 
 The exported functions can accept any number of arguments, of any type, but must only return a single string. any exceptions `throw`n will be handled gracefully.
 
-Arguments are converted using [goja's `ToValue` method](https://pkg.go.dev/github.com/dop251/goja#Runtime.ToValue)). Note in particular that [dates do not work as expected](https://pkg.go.dev/github.com/dop251/goja#hdr-Handling_of_time_Time) in order to provide access to timezone information. If timezone is unimportant you can use `.UnixNano()`:
+Arguments are automatically converted to Javascript native formats using [goja's `ToValue` method](https://pkg.go.dev/github.com/dop251/goja#Runtime.ToValue)).
 
-```typescript
-export function WorkWithDates(myDateObj: object): string {
-  const myDate = new Date(myDateObj.UnixNano()/1e6);
-
-  return myDate.toString();
-}
-```
-
-Which could be called with:
+Dates/times are also converted to native `Date` objects (ie. _not_ like [default goja](https://pkg.go.dev/github.com/dop251/goja#hdr-Handling_of_time_Time)). This does mean that Timezone information is lost; the `Date` object in your function will be in the timezone of the machine Hugo is running on (not the timezone of the passed argument). You can work around this by sending any needed timezone information as a separate argument, eg:
 
 ```go-template
-{{ fb "yours.WorkWithDates" .Date }} → Sun Jun 11 2023 12:00:00 GMT+0100 (British Summer Time)
+{{ fn "example.Timezone" .Date (.Date.Format "-0700") }}
 ```
 
 ## Imports
--- a/docs/content/en/functions/external.md
+++ b/docs/content/en/functions/external.md
@@ -1,0 +1,69 @@
+---
+title: fn
+description: Calls a custom, external Typescript/Jacascript function.
+categories: [functions]
+menu:
+  docs:
+    parent: functions
+keywords: [function, external, code]
+signature: ["external.Function FILE.FUNCTION ARGUMENTS…", "fn FILE.FUNCTION ARGUMENTS…"]
+relatedfuncs: []
+---
+
+This function allows the use of custom, portable Typescript/Javascript code stored within the hugo site files.
+
+For example if the file `functions/hello.ts` exists in a site with the exported function `Name`:
+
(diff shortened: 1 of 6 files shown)


bep:
You may want to have a look at https://github.com/bep/gojap

Creating a new runtime for each function is bound to be very slow, memory hungry.

I like the idea of this (given the above, I have been thinking about it ...), but there are lots of caveats, and I suspect a proof of concept would have a better chance of success if it tested out the hard parts first, mainly:

* In templates, It's very rare that I have simple problems (addition, date formatting ...) that I think "oh, I wish I could do this in JavaScript".
* So for this to be useful, I suspect many people would expect to be able to pass the Hugo `Page`, `Site` etc. down to the JS runtime. That brings up several issues, but the main part is to get it working.

jphastings:
Thank you for the rapid reply @bep! For me this was indeed hard thing #1 ("is it _at all_ possible"), but you're absolutely right, passing `.Site` or similar large objects over is definitely a hard thing worth testing! I can now build some of the trivial things I'd like to have (for example: custom calendar date formatting), but nothing more complex.

The two things I'll work on next:
1. Making use of [gojap](https://github.com/bep/gojap) (it looks great!) to reduce the number of runtimes (perhaps one per file, or one total — I'll try to measure resource usage)
2. Getting a demo of what'd be needed for passing something more substantial over. **Do you have an example use-case in mind?** (I was considering doing something like implementing a `MakeRSS` function — would this cover the challenges you see?)

That challenge I faced of all function calls needing to be Go methods; do you think this is worth addressing in a proof of concept? I'm not particularly familiar with the Hugo codebase, so I'd be very willing to leave it to a later round of thought (if we get there).

bep:
>That challenge I faced of all function calls needing to be Go methods; do you think this is worth addressing in a proof of concept?

You don't want to tackle all problems -- but what I would have wanted to see is something like (pseudo code, I just made up these, so they're not particulary clever):

```js
function getLastMod(pages) {
   // return the newest .Date from the page list given
}

function filter(pages, predicate) {
  // return the pages matching predicate.
}
```

jphastings:
My apologies; I had significant personal challenges later in 2023, and didn't come back to this. I've lost enough context & have enough else on that I'll close this PR to spare the management overhead — but I'll be sure to take on your advice if I do decide to pick this back up in the future!

PR #11158, opened 2023-06-25
Update README.md

jmooring:
It will be easier to review this preview:
<https://gist.github.com/jmooring/e51bf34d9c1b0bfa6a8baa6237baa517>

The broken test badge image is limited to this preview; the image URL is fine.


bep:
This is much, much better than what we had.

PR #11181, opened 2023-06-27
fixes append regression to allow appending nil

### LINKED ISSUE #11180: `append` fails on `nil`
Given the following in a template file:

```
{{ $res := slice }}
{{ $res | append nil }}
```

... the build will fail with an error like:

```
... execute of template failed: template: "partials/openapi/render-api.html" at <append nil>: error calling append: reflect: call of reflect.Value.Type on zero Value
```

This previously worked with Hugo v1.113.0. I suspect it was introduced by https://github.com/gohugoio/hugo/pull/11104.

<pre>
v0.114.1
</pre>

yes

khayyamsaleem:
Closes #11180

### DIFF
--- a/common/collections/append.go
+++ b/common/collections/append.go
@@ -66,6 +66,10 @@ func Append(to any, from ...any) (any, error) {
 
 		if len(from) == 1 {
 			fromv := reflect.ValueOf(from[0])
+			if !fromv.IsValid() {
+				// from[0] is nil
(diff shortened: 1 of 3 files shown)


bep:
Thanks for this, but for the future, please notice my edit of your commit message re. the contribution guidelines.

richvdh:
wow, that was quick. Thanks!

PR #11213, opened 2023-07-06
Add `rss.disableKinds` configuration

Ty3uK:
This commit will allow users to disable generation of `index.xml` files for some kinds of content.

For example, with this config:

```toml
[rss]
    disableKinds = ["taxonomy", "term"]
```

Hugo will not produce `public/<TAXONOMY>/index.xml` and `public/<TAXONOMY>/<TERM>/index.xml` files.

### DIFF
--- a/config/allconfig/allconfig.go
+++ b/config/allconfig/allconfig.go
@@ -182,6 +182,9 @@ type Config struct {
 
 	// UglyURLs configuration. Either a boolean or a sections map.
 	UglyURLs any `mapstructure:"-"`
+
+    // RSS configuration
+    Rss config.Rss `mapstructure:"-"`
 }
 
 type configCompiler interface {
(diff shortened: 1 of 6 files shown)


bep:
It would be good if you created an issue discussing this before implementing it.

We already have a `outputs` configuration that you can use for this. The default includes `rss` for every kind, but you can easily override that.

```toml
[outputs]
  home = ['html', 'rss']
  page = ['html']
 term = ['html']
```

Ty3uK:
> It would be good if you created an issue discussing this before implementing it.
> 
> We already have a `outputs` configuration that you can use for this. The default includes `rss` for every kind, but you can easily override that.
> 
> ```toml
> [outputs]
>   home = ['html', 'rss']
>   page = ['html']
>  term = ['html']
> ```

My bad, sorry. But it was interesting anyway 🙂

PR #11220, opened 2023-07-07
fix: dotpath for Hugo modules mounted as data

### LINKED ISSUE #10681: Mounting data directory to subdirectory does not allow access by subkey

<pre>
$ hugo version
hugo v0.110.0-e32a493b7826d02763c3b79623952e625402b168 linux/amd64 BuildDate=2023-01-17T12:16:09Z VendorInfo=gohugoio
</pre>

Yes

### Description

I have an external directory containing JSON data files, which I mount in my site with the following configuration:

```
[[module.imports]]
source = "../external-data"
target = "data/my_section"
```

For this example, suppose that the `external-data` directory contains a `my_data.json` file with the contents `{ "my_key": "my value" }`.

Because the target is the `my_section` subdirectory, I expect to be able to access the data in my templates using `.Site.Data.my_section...`. With the example data given previously, this would be `.Site.Data.my_section.my_data.my_key`. However, this does not work; I have to use `.Site.Data.my_data.my_key`. If I want to have the data under `my_section`, I have to create an `external-data/my_section` subdirectory instead of having it as part of the mount.

dvdksn:
Fixes #10681

Before this change, data files from Hugo modules were always mounted at the
root of the `data` directory. The File and FileMetaInfo structs for modules
are different from 'native' data directories.

This changes how the keyParts for data files are generated so that data
from modules or native directories are treated the same.

> **Note**
>
> This might not be the best way of solving this issue. Perhaps it's preferred
> to change how file metadata gets created, so that the properties for native
> and module data directories are the same. I just couldn't figure out how to
> do that.

Signed-off-by: David Karlsson <35727626+dvdksn@users.noreply.github.com>

### DIFF
--- a/hugolib/hugo_modules_test.go
+++ b/hugolib/hugo_modules_test.go
@@ -1178,3 +1178,32 @@ target = "content/resources-b"
 	b.AssertFileContent("public/resources-a/subdir/about/index.html", "Single")
 	b.AssertFileContent("public/resources-b/subdir/about/index.html", "Single")
 }
+
+func TestMountData(t *testing.T) {
+	files := `
+-- hugo.toml --
+baseURL = 'https://example.org/'
+disableKinds = ["taxonomy", "term", "RSS", "sitemap", "robotsTXT", "page", "section"]
+
+[[module.mounts]]
+source = "data"
+target = "data"
+
+[[module.mounts]]
+source = "extra-data"
+target = "data/extra"
+-- extra-data/test.yaml --
+message: Hugo Rocks
+-- layouts/index.html --
+{{ site.Data.extra.test.message }}
+`
+
+	b := NewIntegrationTestBuilder(
+		IntegrationTestConfig{
+			T:           t,
+			TxtarString: files,
+		},
+	).Build()
+
+	b.AssertFileContent("public/index.html", "Hugo Rocks")
+}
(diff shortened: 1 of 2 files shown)


hugolib/hugo_modules_test.go
```diff
@@ -1178,3 +1178,28 @@ target = "content/resources-b"
...
+`
+	b := newTestSitesBuilder(t).WithConfigFile("toml", config)
+
+	b.WithSourceFile("extra-data/test.yaml", "message: Hugo Rocks")
+
+	b.WithTemplatesAdded("index.html", "{{ site.Data.extra.test.message }}")
+
+	b.Build(BuildCfg{})
+
+	b.AssertFileContent("public/index.html", "Hugo Rocks")
```
bep, line 1204: OK, the bug fix itself looks good, but I would appreciate if you could rewrite the test to use `NewIntegrationTestBuilder()`. The assertion API is similar. I'm slowly rewriting existing tests to use that setup, so I would hate to merge new stuff with the old testSitesBuilder that I'm trying to get rid of ...

bep (changes requested):

hugolib/hugo_modules_test.go
dvdksn: OK; done

PR #11225, opened 2023-07-08
deploy: Create AWS session for CloudFront invalidation via Go CDK

mattbnz:
This allows the AWS credentials to be picked up from the configured target URL (like blob does) rather than the current behaviour of only relying on the defaults.

Relying on the defaults here means having to specify credentials twice (once in the URL for the blob, once in the environment for this code path) when non-default AWS credentials are in used (e.g. via a profile).

### DIFF
--- a/deploy/cloudfront.go
+++ b/deploy/cloudfront.go
@@ -18,6 +18,7 @@ package deploy
 
 import (
 	"context"
+	"net/url"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


bep:
Does this make sense to you @vangent ?

vangent:
Yes, looks reasonable to me (but hard to be sure with no tests).

bep:
>Yes, looks reasonable to me (but hard to be sure with no tests).

Right ... I think I need to hold off merging this until we can get a basic set of integration tests going. I have done some work on the similar [s3deploy](https://github.com/bep/s3deploy/tree/master/testscripts), but I'm not sure I'm open to setting up Hugo to test against my private AWS account ... I need to think about this.

vangent:
@mattbnz Hi Matt!

> I think I need to hold off merging this until we can get a basic set of integration tests going

The PR isn't making the situation worse, so if you can verify locally that it works I think it's fine to merge. Obviously integration tests would be better long term.

mattbnz:
Hi Robert, LTNS - nice to run into you ;)

Happy to look at some tests, but I'm not super familiar with what infrastructure is available, are there any mock/fake S3 implementations that go-cloud or similar projects are using for this type of testing already?

PR #11233, opened 2023-07-09
create action for vulnerability check

apollol3:
created action to check for vulnerable dependencies on each push and pr


bep:
This needs a proposal (a regular issue) where we agree that this is a great idea before we create a PR implementing it.

PR #11270, opened 2023-07-19
common/htime: Fix localization of abbreviated month names

### LINKED ISSUE #11267: May (month) displayed as full instead of only 3 letters when `dateFormat` is `Jan` (Italian)

hugo v0.115.3+extended darwin/arm64 BuildDate=unknown

Yes

---

When `defaultContentLanguage` is set to `"it"`, using (for example) `{{ dateFormat "02 Jan 06" .Date }}` returns May as `maggio` instead of `mag.`. This only happens with May, other months are displayed correctly.

`{{ .Date | time.Format ":date_medium" }}` is working correctly, returning `mag.` instead.

Out of curiosity, I've also tested `es` and `fr` and saw some inconsistencies there too. I'm not a Spanish/French native speaker, so I can't say if this is normal behavior, but both have some months shortened as 3 letters and some others as 4 letters (May in Spanish seems to have the same problem).

jmooring:
Fixes #11267

### DIFF
--- a/common/htime/integration_test.go
+++ b/common/htime/integration_test.go
@@ -1,0 +1,83 @@
+// Copyright 2022 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


Zabriskije:
Thanks @jmooring @bep, that was fast!

PR #11281, opened 2023-07-23
tpl/collections: Add like operator to where function

### LINKED ISSUE #11279: Support `where` operators for partial matches
Currently, the `where` function is a concise way to filter a collection by comparing *exact matches* between the collections’ data and known values. However, there’s no way to filter it to contain non-exact matches, such as substring, regex, or other types of matches.

One use-case: on my website, I have a data template containing Webmentions (a type of decentralized commenting) for my entire site. Each Webmention has a target URL. I insert Webmentions whose target URL matches the current page, so that each page can end with its own comments section.

I used to use a `where` filter to select just the mentions whose targets match the canonical URL of the current page; however, this didn’t account for fragments. I wanted the fragments at the end of a target URL to be ignored during comparison. `where` doesn’t support operators regular expressions, substring matching, or URL parsing. I had to resort to iteration and appending. Here’s a diff containing that change: <https://github.com/Seirdy/seirdy.one/commit/6fc55df73639149e715f0dacf7bceb6a61f94d46> ([plaintext](https://git.sr.ht/~seirdy/seirdy.one/commit/6fc55df73639149e715f0dacf7bceb6a61f94d46.patch)).

[Discussed on discourse.gohugo.io](https://discourse.gohugo.io/t/add-like-comparison-operator-to-the-where-function/42013)

jmooring:
Enables constructs such as:

```text
{{ range where site.RegularPages "Params.foo" "like" "^ab" }}
  <h2><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></h2>
{{ end }}
```

Will submit documentation change to the documentation repository is this is merged.

Closes #11279

### DIFF
--- a/tpl/collections/where.go
+++ b/tpl/collections/where.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/gohugoio/hugo/common/hreflect"
+	"github.com/gohugoio/hugo/common/hstrings"
 	"github.com/gohugoio/hugo/common/maps"
 )
 
@@ -272,6 +273,17 @@ func (ns *Namespace) checkCondition(v, mv reflect.Value, op string) (bool, error
 			return false, nil
 		}
 		return false, errors.New("invalid intersect values")
+	case "like":
+		if svp != nil && smvp != nil {
+			re, err := hstrings.GetOrCompileRegexp(*smvp)
+			if err != nil {
+				return false, err
+			}
+			if re.MatchString(*svp) {
+				return true, nil
+			}
+			return false, nil
+		}
 	default:
 		return false, errors.New("no such operator")
 	}
--- a/tpl/collections/integration_test.go
+++ b/tpl/collections/integration_test.go
@@ -196,3 +196,38 @@ title: "p3"
 Home: p1|p3|
 `)
 }
+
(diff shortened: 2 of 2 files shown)


tpl/collections/where.go
```diff
@@ -272,6 +273,14 @@ func (ns *Namespace) checkCondition(v, mv reflect.Value, op string) (bool, error
 			return false, nil
 		}
 		return false, errors.New("invalid intersect values")
+	case "like":
+		if svp != nil && smvp != nil {
+			re := regexp.MustCompile(*smvp)
```
bep, line 278: I guess this both makes sense/would be useful, but:

1. Compile is slowish, so do it once and reuse the `*regexp.Regexp` somehow.
2. Also use `regexp.Compile` and return any error.

bep (changes requested):

tpl/collections/where.go
bep: Yea, a common regexp cache would probably be a good idea. Give me 5 minutes and I'll be back.

bep: I have pushed a commit that adds `hstrings.GetOrCompileRegexp`.

PR #11289, opened 2023-07-27
snap: Allow access to SSH keys and $HOME/.config/hugo

### LINKED ISSUE #11288: snap: Allow access to SSH keys and $HOME/.config/hugo
Access to SSH keys is required when users have configured Git to use SSH. For example:

```
[url "git@github.com:"]
  insteadOf = https://github.com/
```

See <https://discourse.gohugo.io/t/error-failed-to-load-modules/45477/>

Access to `$HOME/.cache/hugo` is required in support of https://github.com/gohugoio/hugo/issues/11286.

Both require approval from the Snapcraft team.

jmooring:
Closes #11288

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -59,6 +59,12 @@ plugs:
     interface: personal-files
     read:
       - $HOME/.config/gcloud
+  dot-cache-hugo:
+    interface: personal-files
+    write:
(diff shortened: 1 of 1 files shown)


jmooring:
The change is correct, but the commit message should be:

```text
snap: Allow access to SSH keys and $HOME/.cache/hugo
```

PR #11335, opened 2023-08-05
Update typo after.md

nooitaf:


bep:
The source for the Hugo documentation lives in its own repository, so we would appreciate if you could open this PR directly in `hugoDocs`:  https://github.com/gohugoio/hugoDocs/pulls

We have a copy in this repository as a Git sub tree, and do allow documentation changes that's related to code contributions, but we try to avoid general documentation changes, as that will lead to merge conflicts later. Also, adding the changes to the `hugoDocs` will getting them faster out on the web.

PR #11356, opened 2023-08-13
markup/goldmark: add support for CJK extension

### LINKED ISSUE #10472: Add goldmark's CJK extension support
When blackfriday was the default markdown renderer, we could enable the `joinLines` extension (introduced in #3574) to use CJK-style line breaks.

This feature was removed when the renderer was changed to goldmark, but goldmark recently added a [CJK extension](https://github.com/yuin/goldmark#cjk-extension) that enables CJK-style line breaks. It would be very helpful for CJK users if this extension can be used from hugo.

henry0312:
This commit fixes #10472.

To make this PR function more accurately, merging https://github.com/yuin/goldmark/pull/409 is needed.


henry0312:
goldmark [v1.5.6](https://github.com/yuin/goldmark/releases/tag/v1.5.6) has been released.
The features implemented in the pull request work fine, and it is ready for merging.

bep:
Thanks, this looks good.

I do, however, notice that there are 2 additional options on this extension:

https://github.com/yuin/goldmark#cjk-extension

I assume someone will ask for these at some point, so I think we might as well make the new config option a map, e.g.:

```toml
[markup]
[markup.goldmark]
[markup.goldmark.extensions.CJK]
enable=false
eastAsianLineBreaks=false
escapedSpace=false
--

henry0312:
I updated and added two options.
At first, I anticipated that those who want to enable the `CJK` option would also want to enable the `eastAsianLineBreaks` and `escapedSpace` features simultaneously. 
However, I now believe that making those two options certainly provides more flexibility.

bep:
Thanks, I will merge once the tests go green.

A side note for future reference; the docs.yaml file is autogenerated (I do it before a new release) so any manual edits will be overwritten (which is why it doesn't matter much ...).

PR #11367, opened 2023-08-18
commands/new: Embed site and theme skeletons

### LINKED ISSUE #11358: Extend default theme templates
The default theme proposed by the command `hugo new theme` is kind of light to say the least. There is some content in `baseof.html` referencing a few other (empty) partials: I believe it would make sense to complete these files just enough so that it creates a crude but still functional skeleton.

I'm thinking in particular of:

1. `head.html` (which is already created but empty) could contain something like (and probably other meta such as `.Description`, `.Keywords`, ...):
```html
<head>
  <title>{{block "title" .}}{{with .Title}}{{.}} - {{end}}{{.Site.Title}}{{end}}</title>
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
</head>
```
2. `baseof.html` could show how to make use of `.Site.LanguageCode` in the `<html>` tag (and maybe `GetLanguageDirection`?)
3. `header.html` (which is already created but empty) could contain a basic list for the menus (`.Site.Menus`)
4. `single.html` and `list.html` could simply have a `.Content`
5. `list.html` could make use of the pages thing

There is no need for style or anything fancy, but simply show how the information provided by the engine is supposed to be used. This would help theme creators make best use of the engine features as well as providing a barely functional out of the box experience.

jmooring:
The skeletons are used when creating new sites and themes with the CLI.

Closes #11358


jmooring:
#### Test

```text
hugo new site mysite
cd mysite/
hugo new theme mytheme
echo "theme = 'mytheme'" >> hugo.toml
hugo server
```

#### Theme skeleton

The theme skeleton demonstrates:

- Page bundles
- Asset pipelines (CSS, JS)
- Taxonomies
- Menus
- Partials
- Cached partials
- Date localization

We can certainly take stuff out, but I figured this was a decent starting point for a discussion.

#### Site skeleton

Although the creation method has changed (we now use embed.FS, even to create empty directories), the result is essentially unchanged from earlier versions. With the exception of archetypes, all directories are empty. My thinking was:

1. If we have layouts in the site skeleton, they will override whatever theme the user creates or installs, which is confusing.
2. We could add content, just as we have in the theme skeleton, but some themes expect certain sections to be present. For example, if we create content in a "posts" section, and `github.com/user/some-theme` expects a "blog" section, the user may wonder why it isn't working.

hugolib/skeletons/theme/layouts/partials/head/js.html
```diff
@@ -0,0 +1,3 @@
+{{ with resources.Get "js/main.js" | js.Build | minify | resources.Fingerprint }}
```
bep, line 1: We can certainly take several steps to get this right, but I think this is important enough to adjust. Fingerprinting during development is (usually) a bad idea.

I'm not sure what would be the general good construct here, but I suspect this construct should be OK:

```
{{ $js :=  resources.Get "js/main.js" | js.Build  }}
{{ if ne hugo.Environment "development" }}
{{ $js = $js | minify | fingerprint }}
{{ end }}
```

The same applies to similar constructs above.

On a more nitpicking point: I much prefer using the built-in minifier in ESBuild.

hugolib/skeletons/site/archetypes/default.md
```diff
@@ -0,0 +1,5 @@
++++
+title = '{{ replace .File.ContentBaseName "-" " " | strings.Title }}'
+date = {{ .Date }}
+draft = true
++++
```
bep, line 5: This is not very important, but is there a good reason why this skeleton package lives below hugolib? I would think that `create/skeletons` would be better.

bep (changes requested):

jmooring:
create/skeletons/theme/hugo.toml

```
[module]
  [module.hugoVersion]
    extended = true
    min = "0.116.0"
```

I waffled on this a bit. The skeleton theme does not require the extended version. But if developers use the skeleton to create a theme, and then include LibSass transpilation or WebP encoding, the theme won't work with the standard edition. So, I tried to get ahead of that, but that means you can't use the skeleton with the standard version... which I'm not so sure is such a great idea.

jmooring:
Made some minor changes:
- Removed extended version requirement from theme's config file
- Moved theme config to config directory (example of menu entries in a menus config file--a common question)
- Use `title` instead of `strings.Title` in theme and site archetypes
- Add section page to theme's content/posts directory

bep:
This looks great, but I cannot get the `hugo new site` to work as expected:

```bash
~/dump ❯❯❯ tree mysite
mysite
├── archetypes
│   └── default.md
├── assets
├── content
├── data
├── hugo.toml
├── i18n
├── layouts
├── static
└── themes

8 directories, 2 files
```

`hugo new theme` produces the correct amount of files, so I'm pretty sure I have the correct version built:

```bash
16 directories, 26 files
```

Tested with commit 3f45ff9b4a480dc35d450da8bdc627b2ae0711a4

Does the above work for you?

PR #11381, opened 2023-08-23
create/skeletons: Move theme's site config to top level

### LINKED ISSUE #11380: hugo new theme: Move the config to a top level hugo.toml
Motivation:

1. For discovery
2. Few themes/sites needs the config directory setup, and I don't think we should start out with that as the default.

jmooring:
Closes #11380


create/skeletons/theme/hugo.toml
```diff
@@ -1,3 +1,7 @@
+baseURL = 'https://example.org/'
+languageCode = 'en-us'
+title = 'My New Hugo Site'
+
 [[main]]
```
bep, line 5: `menus.`

bep (changes requested):

create/skeletons/theme/hugo.toml
jmooring: The menu entries are present.

jmooring: Disregard. Thanks for the catch.

PR #11472, opened 2023-09-21
commands: Fix webserver available at url

hugosjoberg:
Running `hugo serve` yielded the following message: `Web Server is available at //localhost:1313/` The url was missing http.

Prepended the baseURL with http: to correct this issue.

Before:
```
Web Server is available at //localhost:1313/ (bind address 127.0.0.1)
```

After
```
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
```

This is a small issue, but adding `http:` makes the URL clickable from the terminal which is nice

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -954,7 +954,7 @@ func (c *serverCommand) serve() error {
 			mu.HandleFunc(u.Path+"/livereload.js", livereload.ServeJS)
 			mu.HandleFunc(u.Path+"/livereload", livereload.Handler)
 		}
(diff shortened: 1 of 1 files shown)


bep:
>Running hugo serve yielded the following message: Web Server is available at //localhost:1313/ The url was missing http.

Thanks, but the above means that your `baseURL` configuration is wrong and missing the protocol (e.g. `https`). Also note that Hugo's server has an option to start up with HTTPS/SSL, so just hard coding `http` will not work.

PR #10521, opened 2022-12-10
Adds start-aware youtube shortcode

sgharms:
Closes #10520

Issue #10520, opened 2022-12-10
Add support for start offset to the internal youtube shortcode


sgharms:
Sometimes  it's useful to scrub ahead in an embedded video from YouTube. The URLs support a `start=` parameter. Expose that interface? I have a local version of the shortcode that works for me, but this might be useful for others?

SandaruKasa:
Bump

SandaruKasa:
> If this is a **feature request**, and you feel that it is still relevant and valuable, please tell us why.

The pull requests are ready #10575 & #10576 (and have been for over a year), so there is not much work left to be done.

I, personally, am still interested in seeing this merged, even despite [the insistently silent treatment](https://github.com/gohugoio/hugo/pull/11559#issuecomment-1810185406) I have received from the maintainer so far.

PR #11559, opened 2023-10-12
shortcodes: Add `start` parameter for `youtube`

SandaruKasa:


SandaruKasa:
Hey

SandaruKasa:
Thanks for reviewing the PR!

Since communicating with contributors does not seem to be something that Hugo prioritizes in any capacity whatsoever, I will, sadly, have to make some guesses and try to piece together a picture of the current state of affairs and the reasoning behind some of the decisions.

Still, any corrections to my theory would be welcome!

I assume my PR got rejected, because, as stated [in the guidelines](https://github.com/gohugoio/hugo/blob/ac7cffa7e2932fc3c6bd425f86b981dfdef94968/CONTRIBUTING.md#code-contribution),

> Any non-trivial code change needs to update an open [issue](https://github.com/gohugoio/hugo/issues). A non-trivial code change without an issue reference with one of the labels `bug` or `enhancement` will not be merged.

and, I have to assume, the changes proposed in this PR were deemed to be non-trivial (probably should also have been labeled as `enhancement`) but had no referenced issue.

Looking at the issues, I seem to have found one that was already opened 11 months ago: #10520.

Moreover, it had received a PR on the exact same day: #10521, but just 3 days later it got rejected in favor of two even smaller PRs: #10575 & #10576. Which have been inactive ever since. (Aside from getting a label 3 months after the creation).

The reason behind this inactivity is, unfortunately, not something I can make an accurate guess on, but my best bet is that they are of a low priority, and hence have been ignored for nearly a year now.

Unfortunately, I am not familiar with the usual review process for such things in Hugo, so I have no clue on when any progress on those issues can be expected, whether it can be expected at all, what needs to happen for it to take place, and whether there is anything I can personally do to aid it in any way. So if anyone could clarify this to me, it would be very appreciated!

Thank you in advance!

SandaruKasa:
Thanks for taking your time with the response, but I would still appreciate if I could hear back about the review process (to maybe get a hint of why such a trivial change takes months to ~~approve~~ review)?

Issue #11562, opened 2023-10-13
LiveReload script is injected inside <header> without warning when <head> is omitted.


DominoPivot:
`livereloadinject.go` naively scans an HTML output as a byte[] and attempts to append a script element:
1.  after the first `>` that follows the first `<head` or `<HEAD`
2.  before `</body>` or `</BODY>`
3.  after the first `>` that follows `<html` or `<HTML` (logs a warning in the browser console)
4.  at the end of the output (logs a warning in the browser console)

<details><summary>Related code</summary>

https://github.com/gohugoio/hugo/blob/1b5f78b6b7335b02b6207a637498c4c8817999d1/transform/livereloadinject/livereloadinject.go#L39-L46 https://github.com/gohugoio/hugo/blob/1b5f78b6b7335b02b6207a637498c4c8817999d1/transform/livereloadinject/livereloadinject.go#L75-L78

</details>

But despite the many fallbacks, **if the output omits the opening `<head>` tag but includes `<head` anywhere else, the script will be injected at an unpredictable location without a warning.** This could be inside a `<header>` element as well as inside a `<style>`, `<script>`, or `<template>` element that happens to contain `<head`.

I discovered this because I typically omit head and body tags, but use a `<header>` for a site's banner. LiveReload gets injected in the header, which has yet to cause issues, but it is still unexpected behavior. I then intentionally made it break a page that contained an inline script with the regex `/#+\s*(?<heading>.+)/g;`.

I believe the LiveReload script should be injected at the start of the head all the time for better performance, but we need a better algorithm to find the implicit start of the head when tags are omitted:
1.  attempt to insert after the `<head>` tag
2.  otherwise attempt to insert after the `<html>` tag
3.  otherwise attempt to insert before the first tag that isn't the `<!doctype>` or a `<!--comment-->`
4.  otherwise return an error

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.119.0-b84644c008e0dc2c4b67bd69cccf87a41a03937e windows/amd64 BuildDate=2023-09-24T15:20:17Z VendorInfo=gohugoio
</pre>

### Does this issue reproduce with the latest release?
Yes

DominoPivot:
For future reference, even this is a valid HTML document:

```html
<!DocType hTml>
<scrIpt></scriPT>
<TitLe>x</titlE>
```

which produces the same DOM as

```html
<!doctype html>
<html>
  <head>
    <script></script>
[...]

PR #11565, opened 2023-10-15
livereloadinject: Replace naive injection logic with regex

DominoPivot:
This PR addresses some of the concerns I raised in https://github.com/gohugoio/hugo/issues/11562. This is just a first draft and my first contribution here, do not hesitate to share thoughts.

Specifically, this PR should fix issues that arise when a document omits various combinations of html, head and body tags, which is allowed by the standard. It should also solve the bug which would cause accidental injection in a `<header>` element in place of the head.

It still won't solve problems when `<head>` is written inside comments or elements that allow arbitrary text like so:

    <title>The <head> element</title>

The current behavior of `livereloadinject` is to attempt injecting a LiveReload script after the `<head>` tag, or before the `</body>`, or after the `<html>`, or at the end of the document, whichever we can find first.

This PR changes the behavior to inject after `<head>` or `<html>` or `<!doctype>` (now truly case insensitive), or before the first `<`. This keeps the behavior more predictable instead of sometimes appending at the start, sometimes at the end.

~~I went with a single-regex approach because it's more intuitive to me and reduces code complexity quite a bit, but I understand this might come with a performance cost.~~ It indeed has a terrible cost because the regex is greedy, I will split it and force push.

I considered passing only the first 1024 bytes of the document to the regex to stop looking for the head/html/doctype beyond that point... but then I remembered my college teacher who had novels written in `<!--comments-->` at the start of every file he made. Would love a second opinion on that.

### DIFF
--- a/transform/livereloadinject/livereloadinject.go
+++ b/transform/livereloadinject/livereloadinject.go
@@ -14,10 +14,10 @@
 package livereloadinject
 
 import (
-	"bytes"
 	"fmt"
 	"html"
 	"net/url"
+	"regexp"
 	"strings"
 
 	"github.com/gohugoio/hugo/common/loggers"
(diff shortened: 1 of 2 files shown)


DominoPivot:
I knew that greedy regex was likely very inefficient and this code would require changes, but I just realized I introduced a breaking change. Injection worked fine with this file prior to my change:

```html
<!doctype html>
<head><title>x</title></head>
...
<!-- bla bla bla <head> -->
```

All because I made the regex greedy so it would attempt to include the head instead of stopping at html or doctype. I'll fix this and write a test case so it doesn't happen again.

DominoPivot:
I force pushed a new version that loops through a list of regular expressions to look for insertion points, closer to the original code. I added more test cases, including some of which are not covered by this solution (for example, if the string <head> was inside a comment or title).

I'll iterate on alternatives in the upcoming days.

bep:
>I'll iterate on alternatives in the upcoming days.

@DominoPivot your implementation looks good to me, and it should be plenty fast enough for the common cases.

bep:
... but there are some test errors.

DominoPivot:
@bep I knew some tests would fail because they weren't solved by this approach.

I've implemented an alternative solution https://github.com/gohugoio/hugo/pull/11578 which covers all cases and more, I'll close this PR but leave the branch on my fork so we can reopen if someone can point a good reason not to go with the other one.

PR #11567, opened 2023-10-16
hugolib: Deprecate .Site.DisqusShortname

jmooring:
Use .Site.Config.Services.Disqus.Shortname instead.


bep:
There's a merge conflict here....

jmooring:
Rebased.

bep:
Thanks for this, I will merge this. As a note for later; we may want to consider starting to use `site.Config...` instead of `.Site.Config` where the user does not have to think about the "dot".

PR #11573, opened 2023-10-18
commands/mod: update all active modules

razonyang:
Currently, `hugo mod get` won't update indirect import modules.

```toml
// site config
[[module.imports]]
path = "example.org/vendor/foo"
```

```toml
// example.org/vendor/foo module config
[[module.imports]]
path = "example.org/vendor/bar"
```

`hugo mod get` won't update the `example.org/vendor/bar` module.

Related issue/discussion: https://discourse.gohugo.io/t/unable-to-update-modules-those-imported-indirectly/46707.

I'm not sure if this PR is good, or should I create another command instead of overriding the original for this?

### DIFF
--- a/modules/client.go
+++ b/modules/client.go
@@ -318,14 +318,33 @@ func (c *Client) Get(args ...string) error {
 		patch := update && (args[0] == "-u=patch") //
 
 		// We need to be explicit about the modules to get.
-		for _, m := range c.moduleConfig.Imports {
-			if !isProbablyModule(m.Path) {
-				// Skip themes/components stored below /themes etc.
-				// There may be false positives in the above, but those
-				// should be rare, and they will fail below with an
-				// "cannot find module providing ..." message.
-				continue
+		var modules []string
+		// Update all active modules if the -u flag presents.
+		if update {
+			mc, coll := c.collect(true)
+			if coll.err != nil {
+				return coll.err
 			}
+			for _, m := range mc.AllModules {
+				if m.Owner() == nil {
+					continue
+				}
+				modules = append(modules, m.Path())
+			}
+		} else {
+			for _, m := range c.moduleConfig.Imports {
+				if !isProbablyModule(m.Path) {
+					// Skip themes/components stored below /themes etc.
+					// There may be false positives in the above, but those
+					// should be rare, and they will fail below with an
+					// "cannot find module providing ..." message.
+					continue
+				}
+				modules = append(modules, m.Path)
+			}
+		}
+
+		for _, m := range modules {
 			var args []string
 
 			if update && !patch {
@@ -333,7 +352,7 @@ func (c *Client) Get(args ...string) error {
 			} else if update && patch {
 				args = append(args, "-u=patch")
 			}
-			args = append(args, m.Path)
+			args = append(args, m)
 
 			if err := c.get(args...); err != nil {
 				return err
--- a/testscripts/commands/mod_get.txt
+++ b/testscripts/commands/mod_get.txt
@@ -1,0 +1,15 @@
+hugo mod get
+stderr 'withhugotoml.*v1.1.0'
+
+-- hugo.toml --
+title = "Hugo Modules Test"
+[module]
+[[module.imports]]
+path="github.com/gohugoio/hugo-mod-integrationtests/withconfigtoml"
+disable = true
+[[module.imports]]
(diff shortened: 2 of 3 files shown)


bep:
>Currently, hugo mod get won't update indirect import modules.

This is intentional (and as documented https://gohugo.io/commands/hugo_mod_get/), and is also how `go get` behaves.

Could you try to do a `hugo mod get -u`?

razonyang:
> Could you try to do a `hugo mod get -u`?

I can reproduce this on the test repo: https://github.com/razonyang/hugo-mod-tests.

```sh
$ git status
On branch main
nothing to commit, working tree clean

$ hugo mod graph
github.com/razonyang/hugo-mod-tests github.com/razonyang/hugo-mod-bar@v0.2.0
github.com/razonyang/hugo-mod-bar@v0.2.0 github.com/razonyang/hugo-mod-foo@v0.2.0

$ hugo mod get -u

$ git status
On branch main
nothing to commit, working tree clean

$ hugo mod graph
github.com/razonyang/hugo-mod-tests github.com/razonyang/hugo-mod-bar@v0.2.0
github.com/razonyang/hugo-mod-bar@v0.2.0 github.com/razonyang/hugo-mod-foo@v0.2.0
```

After modifying the `go.mod` manally, then it'll updates.

```sh
$ vim go.mod

$ git status
On branch main
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   go.mod
no changes added to commit (use "git add" and/or "git commit -a")

$ git diff
diff --git a/go.mod b/go.mod
index bb3d1e4..bb27506 100644
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,4 @@ go 1.20

 require (
        github.com/razonyang/hugo-mod-bar v0.2.0 // indirect
-       github.com/razonyang/hugo-mod-foo v0.2.0 // indirect
 )

$ hugo mod get -u
go: added github.com/razonyang/hugo-mod-foo v0.3.0
hugo: collected modules in 1891 ms

$ hugo mod graph
github.com/razonyang/hugo-mod-tests github.com/razonyang/hugo-mod-bar@v0.2.0
github.com/razonyang/hugo-mod-bar@v0.2.0 github.com/razonyang/hugo-mod-foo@v0.3.0
```

bep:
OK, then I think I understand.

I think what we need to do is to make the code in this PR run only when the `-u` flag. When that flag is not present we only want to update direct imports.

But we need a test for this.

I have created a new version of this test module which is included in some of the others:

[commonmod](https://github.com/gohugoio/hugo-mod-integrationtests/tree/main/commonmod)

If you:

1. Copy https://github.com/gohugoio/hugo/blob/5160c7efa5771d74b560f9c2ea761f09e08a9216/testscripts/commands/mod__disable.txt#L8 into ../unfinished and name it something sensible
2. Then run https://github.com/gohugoio/hugo/blob/5160c7efa5771d74b560f9c2ea761f09e08a9216/main_test.go#L46 until you're happy and then move the test back to commands.

Some related tips:

* I use this VS Code extension that allows me to "pin a test" and run it on every save: https://marketplace.visualstudio.com/items?itemName=windmilleng.vscode-go-autotest
* That extension only watches *.go files, so I made this hack to make it work with *.txt test scripts: https://github.com/gohugoio/hugo/blob/5160c7efa5771d74b560f9c2ea761f09e08a9216/watchtestscripts.sh

razonyang:
Thanks for the tip, I hope I'm doing it right.

bep:
OK, I have added a commonmod2 module in the repo above, which should give you a transitive dependency.

razonyang:
@bep could you please draft one more release for `commonmod` contains the newer dependencies (`commonmod2`)?

bep:
@razonyang https://github.com/gohugoio/hugo-mod-integrationtests/releases/tag/commonmod%2Fv1.0.1

razonyang:
Thanks, is it possible to run multiple commands in one test? The output doesn't include the info about `commonmod2`, I'm wondering if I can run a subsequent command to check the `go.mod` and `go.sum`, something like `hugo mod get -u && hugo mod graph`.

razonyang:
I figured it out by putting the subsequent command in a new line, the PR is ready to review.

bep:
@razonyang thanks, much appreciated, I will take a closer look at this tomorrow.

phuchau1989 (approved):

PR #11578, opened 2023-10-18
livereloadinject: Use more robust injection method

DominoPivot:
This PR addresses all the concerns I raised in issue https://github.com/gohugoio/hugo/issues/11562 and then some.

It changes the behavior of the livereloadinject transformer to read the HTML document from the start, consuming any comments and whitespace as well as the doctype, html start tag and head start tag in that order, stopping the moment another tag or text content is encountered, and injecting the livereload script there. Compared to the current code, it specifically:
- Allows mixed casing in tag names.
- Always injects at the start of the head even when tags are omitted, meaning the browser can download the script faster.
- Always attempts to inject, and never with a warning. Browsers already warn the user if the HTML is malformed.
- Requires whitespace before attributes, so `<header>` or `<head-foo>` are no longer mistaken for the head start tag.
- Explicitly skips `<!--comments-->` that may contain tags. Also treats `<?xml instructions ?>` as comments.
- Won't attempt to scan the document further if any other element or text is encountered. This should improve performance on pages without a head start tag, and prevent bugs due to scans inside `<title>`, `<script>` and `<template>` elements, or elements whose attribute values might contain `<` or `>`.

This is a different approach compared to PR https://github.com/gohugoio/hugo/pull/11565 where I would search the whole document for each potential insertion point, which left some of the bugs unresolved.

### DIFF
--- a/transform/livereloadinject/livereloadinject.go
+++ b/transform/livereloadinject/livereloadinject.go
@@ -25,12 +25,14 @@ import (
 	"github.com/gohugoio/hugo/transform"
 )
 
-var ignoredSyntax = regexp.MustCompile(`(?s)^(?:\s+|<!--.*?-->|<\?.*?\?>)*`)
-var tagsBeforeHead = []*regexp.Regexp{
-	regexp.MustCompile(`(?is)^<!doctype\s[^>]*>`),
-	regexp.MustCompile(`(?is)^<html(?:\s[^>]*)?>`),
-	regexp.MustCompile(`(?is)^<head(?:\s[^>]*)?>`),
-}
+var (
+	ignoredSyntax  = regexp.MustCompile(`(?s)^(?:\s+|<!--.*?-->|<\?.*?\?>)*`)
+	tagsBeforeHead = []*regexp.Regexp{
+		regexp.MustCompile(`(?is)^<!doctype\s[^>]*>`),
+		regexp.MustCompile(`(?is)^<html(?:\s[^>]*)?>`),
+		regexp.MustCompile(`(?is)^<head(?:\s[^>]*)?>`),
+	}
+)
 
 // New creates a function that can be used to inject a script tag for
 // the livereload JavaScript at the start of an HTML document's head.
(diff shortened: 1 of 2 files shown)


DominoPivot:
Force pushed a slight change because I realized there's no reason to skip whitespace and comments when we're already inside the head, and it leads to cleaner code.

transform/livereloadinject/livereloadinject.go
```diff
@@ -72,23 +57,9 @@ func New(baseURL url.URL) transform.Transformer {
 		c := make([]byte, len(b))
 		copy(c, b)
```
DominoPivot, line 58: I didn't write those lines, but isn't this pointless? From my (limited) understanding of Go:
- `make` wasn't passed a capacity, so future calls to `append(c, ...)` will need to reallocate
- `copy` was given the entirety of `b` even though we're about to append to `c[:idx]`, overwriting most of it

My intuition would be to write this lower in the function instead:

    c := make([]byte, idx, len(b)+len(script))
    copy(c, b)

Or alternatively:

    c := make([]byte, 0, len(b)+len(script))
    c = append(c, b[:idx]...)

transform/livereloadinject/livereloadinject.go
```diff
@@ -72,23 +57,9 @@ func New(baseURL url.URL) transform.Transformer {
...
-		i := idx
-		if match.appendScript {
-			i += bytes.Index(b[i:], []byte(">")) + 1
-		}
-
-		if match.warnRequired {
-			script = append(script, []byte(warnScript)...)
-		}
-
-		c = append(c[:i], append(script, c[i:]...)...)
```
DominoPivot, line 91: And again here, doesn't `append(script, c[i:]...)` cause another unnecessary reallocation? Wouldn't it be better to do:

    c := make([]byte, 0, len(b)+len(script))
    c = append(c, b[:idx]...)
    c = append(c, script...)
    c = append(c, b[idx:]...)

transform/livereloadinject/livereloadinject.go
bep: You are right, this would be better:

```go
c := make([]byte, idx, len(b)+len(script))
copy(c, b)
```

transform/livereloadinject/livereloadinject.go
bep: From my quick glance, I don't think you're right here, but it would be trivial to write a quick benchmark.

PR #11593, opened 2023-10-22
markdown: Pass emoji codes to yuin/goldmark-emoji 

### LINKED ISSUE #7332: Emoji are rendered within code and code blocks

<pre>
$ hugo version
Hugo Static Site Generator v0.71.1/extended linux/amd64 BuildDate: 2020-05-25T13:06:06Z
</pre>

Yes

### Problem Description

Emoji strings with code (single backticks) or code blocks (triple backticks) are rendered to images. They should remain as strings.

By comparison, GitHub does not transform emoji strings when they appear in code.

I came across this problem while trying to document the use of emoji strings. Not very important...

### LINKED ISSUE #11587: Emoji codes are not converted, and disappear, when using the RenderShortcodes method
Not sure if we could resolve this by adding the [yuin/goldmark-emoji](https://github.com/yuin/goldmark-emoji/) extension, which solves another problem too, but it's worth a test.

See <https://github.com/gohugoio/hugo/issues/7332#issuecomment-1608424761>.

Tested with `enableEmoji= true` in site configuration.

---
Update: I enabled the yuin/goldmark-emoji extension, and set `enableEmoji= false`. Fixes the problem. I think it makes sense to add that extension, disabled by default.

jmooring:
Fixes #7332
Fixes #11587


jmooring:
So, this works great, but right now you need to do two things in site config to use it:

```toml
enableEmoji = false
[markup.goldmark.extensions]
emoji = true # default is false
```

It would be better if `enableEmoji` were the primary on/off switch, with  goldmark/emoji enabled by default. The problem I'm having is in the page parser:

(https://github.com/gohugoio/hugo/blob/de4e466036026e9a5805155f00882b93267231b5/hugolib/page.go#L796-L801)

What I want to do is:

```go
if this_content_will_be_rendered_by_goldmark {
  rn.AddBytes(it) 
} else {
  if emoji := helpers.Emoji(it.ValStr(result.Input())); emoji != nil { 
    rn.AddReplacement(emoji, it) 
  } else { 
    rn.AddBytes(it) 
  } 
}
```
But I can't figure out how to determine this_content_will_be_rendered_by_goldmark.

bep:
>But I can't figure out how to determine this_content_will_be_rendered_by_goldmark.

So, the Emoji feature was added a long time ago. I'm tempted to make that into a Goldmark only thing. We need to reduce some of the complexity in Hugo, and, as the issues referenced above, what we have isn't really working so we might as well just cut it.

We should translate `enableEmoji` to the new Goldmark option.

jmooring:
> Goldmark only thing

I'm fine with that, as long as we understand that existing emoji in org, pandoc, rst, asciidoc, and html will no longer render (breaking change). I just tested the first four to verify that the renders don't already have this baked in.

I'll update the PR.

jmooring:
`enableEmoji` in site config now enables/disables the yuin/goldmark-emoji extension. No other config setting required. For markdown users, there is one expected change that is covered in the test cases:

```
{{< sc >}}:smiley:{{< /sc >}}
```

The shortcode must now do this:

```
{{ .Inner | .Page.RenderString }}
```

Instead of this:

```
{{ .Inner }}
```

Shortcodes called with the `{{% sc %}}` notation continue to work as they did before.

markup/goldmark/integration_test.go
```diff
@@ -578,3 +578,114 @@ a <!-- b --> c
...
+-- layouts/_default/single.html --
+{{ .Content }}
+`
+
+	b := hugolib.NewIntegrationTestBuilder(
+		hugolib.IntegrationTestConfig{
+			T:           t,
+			TxtarString: files,
+		},
+	).Build()
```
bep, line 688: A tip: I found that I repeated the above construct so often, so I added 2 convenience funcs named `Test` and `TestRunning` which you could rewrite the above to:

```go
b := hugolib.Test(t, files)
```

PR #11606, opened 2023-10-26
Add a simple skeleton when creating a new site

### LINKED ISSUE #6859: Warning when building site generated using hugo new site

v0.64.0-241DB8F7 linux/amd64 BuildDate: 2020-02-04T09:10:23Z

Yes.

### Platform

Pop!_OS 19.10 (based on Ubuntu 19.10)

### To reproduce

1. `hugo new site my-site`
2. `cd my-site`
3. `hugo`

### What should happen

Output should resemble the following, with no errors or warnings on a newly-generated site.

```sh
Building sites…

                   | EN  
-------------------+-----
  Pages            |  3  
  Paginator pages  |  0  
  Non-page files   |  0  
  Static files     |  0  
  Processed images |  0  
  Aliases          |  0  
  Sitemaps         |  1  
  Cleaned          |  0  

Total in 12 ms
```

### What actually happens

Output resembles the following with three cryptic warnings:

```sh
Building sites … WARN 2020/02/06 13:03:11 found no layout file for "HTML" for kind "taxonomyTerm": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
WARN 2020/02/06 13:03:11 found no layout file for "HTML" for kind "home": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
WARN 2020/02/06 13:03:11 found no layout file for "HTML" for kind "taxonomyTerm": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.

                   | EN  
-------------------+-----
  Pages            |  3  
  Paginator pages  |  0  
  Non-page files   |  0  
  Static files     |  0  
  Processed images |  0  
  Aliases          |  0  
  Sitemaps         |  1  
  Cleaned          |  0  

Total in 12 ms
```

### Proposed fix

Fix whatever is causing the warnings so that a new site, generated using the official command, builds without warnings or errors.

The minimum fix appears to be the following:

1. Add an empty _index.html_ file into the generated _layouts/_ folder
2. Add `disableKinds = ['taxonomyTerm']` to the generated _config.toml_ file.

With those two alterations, a site generated using `hugo new site` builds without warnings or errors.

The above is most likely __not__ also the ideal fix, which might include having the generated index.html file contain some useful Getting Started information for steps the person can take next to start customising the site.

carloseabade:
Fixes #6859


jmooring:
We specifically didn't do this with https://github.com/gohugoio/hugo/pull/11367 because it breaks this:

<https://gohugo.io/getting-started/quick-start/#create-a-site>

Users would have to delete layouts/index.html and the cruft in static.  #6859 is a good idea, but we need to agree on an approach before coding anything.

Issue #11599, opened 2023-10-24
Create padding image filter


jmooring:
Return an image resource that you can use with `images.Text` and `images.Overlay`.

```
images.Create WIDTH HEIGHT [BGCOLOR] [FORMAT]
```

- Width and height must be in the range [1,10000] (or something smaller, but put a limit on it).
- Background color is RGBA, with length of 3, 4, 6, or 8 (e.g., `#123`, `#123f`, `#112233`, `#112233ff`. If alpha not specified, opaque. Default is white.
- Format is one of `jpeg`, `webp`, etc. Default is `png`.

The way to handle this today is with a `1x1.png` in the assets directory that you grab as a resource and then resize. 

Reference: <https://discourse.gohugo.io/t/should-it-be-possible-to-pad-an-an-image-using-0-119-image-processing/46684/4>

---

EDIT: Rethinking this, the specific case referenced above, image padding, would probably be handled better with an image filter.

```
images.Pad D1 [D2] [D3] [D4] [BGCOLOR]
```

Where the dimensions use the same syntax as CSS padding (e.g., if you provide one number if affects all sides), and background color is as previously described.

bep:
>The way to handle this today is with a 1x1.png in the assets directory that you grab as a resource and then resize.

That said, I say that the common way today is to use a pretty background image (with some gradient or something) for your overlay images, which is trivial to make in most photo editors.

jmooring:
Agreed. The _primary_ use case behind this suggestions is padding an image, which today is `get+resize+get+overlay`.  

Instead of doing `create+get+overlay`, maybe just a filter so you can do `get+pad`.

jmooring:
Renaming this issue; the ability to pad an image is the primary use case. We can revisit image creation at a later date if needed.

PR #11621, opened 2023-10-28
resources/images: Create padding image filter

jmooring:
This addresses the primary use case described in https://github.com/gohugoio/hugo/issues/11599.

### DIFF
--- a/resources/images/filters.go
+++ b/resources/images/filters.go
@@ -16,6 +16,7 @@ package images
 
 import (
 	"fmt"
+	"image/color"
 
 	"github.com/gohugoio/hugo/common/hugio"
 	"github.com/gohugoio/hugo/common/maps"
@@ -30,6 +31,7 @@ const filterAPIVersion = 0
 
 type Filters struct{}
 
+// Process creates a filter that processes an image using the given specification.
 func (*Filters) Process(spec any) gift.Filter {
 	return filter{
 		Options: newFilterOpts(spec),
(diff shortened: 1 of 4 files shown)


jmooring:
Can also combine with the `.Colors` method...

```text
{{ with resources.Get "images/c.jpg" }}
  {{ with .Filter (images.Padding "20" (index .Colors 2)) }}
  {{ end }}
{{ end }}
```

resources/images/filters.go
```diff
@@ -110,6 +113,65 @@ func (*Filters) Text(text string, options ...any) gift.Filter {
 	}
 }
 
+// Padding creates a filter that resizes the image canvas without resizing the
+// image. The last argument is the canvas color, expressed as an RGB or RGBA
+// hexadecimal color. The default value is `ffffffff` (opaque white). The
+// preceding arguments are the padding values, in pixels, using the CSS
+// shorthand property syntax. Negative padding values will crop the image. The
+// signature is images.Padding V1 [V2] [V3] [V4] [COLOR].
+func (*Filters) Padding(args ...any) gift.Filter {
```
bep, line 122: I like the overall idea of the Padding func. But I think the variadic function setup becomes too complicated with 5 different arguments.

I suggest

* using a dict with top, right, bottom, left, color
* create a struct with these and use mapstructure.WeakDecode
* But make sure to pass the original dict (the one provided from the user) to newFilterOpts (this allows us to add fields to the options struct without breaking the hash)

resources/images/padding.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021 The Hugo Authors. All rights reserved.
+//
```
bep, line 2: 2023.

bep (changes requested):

resources/images/filters.go
jmooring: This is a polite request to reconsider. In my view the most common cases are:

```go-html-template
{{ $filter := images.Padding 20 }}
{{ $filter := images.Padding 20 40 }}
```

Also, the CSS padding shorthand should be familiar to most site and theme authors, and I can certainly add links to mozilla docs describing the notation.

I'm fine with the options map; it just seems like a little more work compared to some of the other filters, excluding text and overlay.

bep: Can you guarantee that there will be no more options coming in the future here?

bep: What if you could have both ... So we would check for the common cases (1 or 2 integers) as the special case.

bep: Yea, ok, we could introduce the dict if it gets more complicated.

resources/images/filters.go
```diff
@@ -110,6 +113,65 @@ func (*Filters) Text(text string, options ...any) gift.Filter {
...
+	if len(args) == 0 {
+		panic("the padding filter requires one or more arguments")
+	}
+
+	var top, right, bottom, left int
+	var ccolor color.Color = color.White // canvas color
+	var err error
+
+	vc := args[len(args)-1]
+	if reflect.TypeOf(vc).Kind() == reflect.String {
```
bep, line 132: I don't see a need to use reflect here, do a "Go type cast".

resources/images/filters.go
```diff
@@ -110,6 +113,65 @@ func (*Filters) Text(text string, options ...any) gift.Filter {
...
+	case 2:
+		top, right, bottom, left = vals[0], vals[1], vals[0], vals[1]
+	case 3:
+		top, right, bottom, left = vals[0], vals[1], vals[2], vals[1]
+	case 4:
+		top, right, bottom, left = vals[0], vals[1], vals[2], vals[3]
+	}
+
+	return filter{
+		Options: newFilterOpts(top, right, bottom, left, ccolor),
```
bep, line 164: Use vc instead of ccolor here. Probably not important, but will at least guarantee the most stable hash.

bep (changes requested):

resources/images/filters.go
bep, line 164: Just pass `args..` in here, which would make the most stable hash.

PR #11622, opened 2023-10-29
markup/goldmark: update the CJK extension to allow specifying line break styles

henry0312:
This commit follows https://github.com/yuin/goldmark/pull/411

### DIFF
--- a/markup/goldmark/goldmark_config/config.go
+++ b/markup/goldmark/goldmark_config/config.go
@@ -44,9 +44,10 @@ var Default = Config{
 		LinkifyProtocol: "https",
 		TaskList:        true,
 		CJK: CJK{
-			Enable:              false,
-			EastAsianLineBreaks: false,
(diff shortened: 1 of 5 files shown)


markup/goldmark/goldmark_config/config.go
```diff
@@ -44,9 +44,10 @@ var Default = Config{
 		LinkifyProtocol: "https",
 		TaskList:        true,
 		CJK: CJK{
-			Enable:              false,
-			EastAsianLineBreaks: false,
-			EscapedSpace:        false,
+			Enable:                   false,
+			EastAsianLineBreaks:      false,
+			EastAsianLineBreaksStyle: "simple",
```
henry0312, line 49: I placed the style specification settings at this level to maintain backward compatibility.

`"simple"` is the current newline style. If not specified, it defaults to `"simple"`.
Another option available is `"css3draft"`.

PR #11688, opened 2023-11-09
tpl: allow using page resources on the images page parameter for `opengraph`, `schema` and `twitter_cards` templates

razonyang:
Hi, I want to  specify the page resources on `images` parameter other than renaming the files, the new image parsing order as follows.

1. If the image is external, leave it as it is.
2. Use page resource if there is matched one.
3. In other cases, use `absURL` to transform the image URL.


bep:
Thanks for this, and I agree about the change, but we need to fix the code duplication, as I get dizzy trying to verify that this gets correct across all the templates.

I'll /cc @jmooring to get the naming right here, but I'm thinking that we

* Create a embedded partial in `partials/_funcs/get-page-images.html`
* Add partials to the exclude here: https://github.com/gohugoio/hugo/blob/26a8ec207f0fe7b28c61829a3f254a29c40569c1/tpl/tplimpl/template.go#L782 (which would allow people to override this "fun partial".

And then make the partial return a slice of this `dict`:

```
Image # may be nil
RelPermalink
Permalink
```

And the order we check would be:

1. `$.Params.images` and for each of these: For relative URLs first check `$.Resources.GetMatch`, else use `relURL` and `baseURL`.
2. Use the logic the looks for featured image in page resources (see opengraph.html)
3. Look in site.Params.images and pick the first and then apply the logic in 1.

And then use that partial to get the images to use in all of these templates.

razonyang:
@bep thanks for the hints, I modified the one of the `go:embed` directives to make the build pass, but I'm not sure if it's correct.

```
//go:embed embedded/templates/*
```

to

```
//go:embed all:embedded/templates/*
```

Otherwise I unable to embed the `partials/_funcs` folder:

```log
tpl/tplimpl/template.go:754:12: pattern embedded/templates/*: cannot embed directory embedded/templates/partials: contains no embeddable files
```

jmooring:
While I think this change is needed, my _opinion_ on embedded templates remains unchanged.

If it were up to me, I would deprecate most (all?) of them. Instead, I would point site and theme authors to examples in documentation. I would **not** provide a module/theme. That would imply a promise of maintenance and some degree of backwards compatibility.

There's a reason we have [42 open issues and pull requests](https://github.com/gohugoio/hugo/labels/Feature%3A%20Embedded%20Template) related to embedded templates. They are (by definition) opinionated and we are afraid of breaking something. A while back, while modifying the alias template to use `site.Language.LanguageCode`, I also made the code easier to read. This affected white space in the rendered file, and someone complained about the size of their diff file. You can't win on this stuff.

Having said all that, the naming seems fine, but I'd like to take this PR for a spin before merging.

jmooring:
```text
content/
├── posts/
│   ├── post-1/
│   │   ├── a.jpg
│   │   ├── cover.jpg
│   │   └── index.md
│   └── _index.md
└── _index.md
```

content/posts/post-1/index.md

```text
+++
title = 'Post 1'
images = ['a.jpg']
+++
```

With opengraph in v0.120.4

```text
<meta property="og:image" content="http://localhost:1313/a.jpg" />
```

With opengraph in this PR

```
<meta property="og:image" content="http://localhost:1313/posts/post-1/a.jpg" />
<meta property="og:image" content="http://localhost:1313/posts/post-1/cover.jpg" />
```

The existing template skips page resources if `.Page.Params.images` isn't empty.

I don't know if anyone will complain, but now I'm including images that I wasn't before.

razonyang:
@jmooring thanks for taking time reviewing this.

- [x] The existing template skips page resources if .Page.Params.images isn't empty.
- [x] Reduced code of `range` block...

> Also, is there any benefit to caching? Seems like it's likely to be different on each page.

Since those three templates use the same images, so I intend to cache the images for each pages (uses dot `.`/page as the variant), to avoid parsing multiple times. Please correct me if I'm wrong.

tpl/tplimpl/embedded/templates/partials/_funcs/get-page-images.html
```diff
@@ -0,0 +1,27 @@
...
+  {{- $imgs = $imgs | append $img }}
+{{- end }}
+{{/* Find featured image resources if the images parameter is empty. */}}
+{{- if not $imgs }}
+  {{- $images := $.Resources.ByType "image" -}}
+  {{- $featured := $images.GetMatch "*feature*" -}}
+  {{- if not $featured }}{{ $featured = $images.GetMatch "{*cover*,*thumbnail*}" }}{{ end -}}
+  {{- with $featured }}
+    {{- $imgs = $imgs | append (dict
+      "Image" .Name
```
bep, line 15: Image must either be a image object (with .Width, .Height etc.) or nothing at all (an empty string is fine and practical.

The idea about this is that people could use this partial in their own template and say:

```handlebars
{{ $i := partial "_funcs/get-page-images.html" }}
{{ with $i }}
  {{ with $i.Image }}
      // resize, apply image filters etc.
  {{ else }}
   // just use the .RelPermalink
  {{ end }}
{{ end }}
```

PR #11706, opened 2023-11-13
resources/resource: Fix GroupByParamDate with raw TOML dates

### LINKED ISSUE #11563: With unquoted TOML values, GroupByParamDate requires fully qualified date
The unexpected behavior is limited to unquoted TOML values.

`GroupByParamDate "eventDate" "2006-01"`works fine with all of these:

```text
title = 'something'
eventDate = 2022-08-13T13:57:43-07:00   # unquoted 
```

```text
title = 'something'
eventDate = '2022-08-13T13:57:43-07:00'   # quoted 
```

```text
title = 'something'
eventDate = '2022-08-13'  # quoted 
```

But pages with this parameter date format are not grouped:

```text
title = 'something'
eventDate = 2022-08-13  # unquoted 
```

`GroupByDate`, `GroupByExpiryDate`, `GroupByLastmod`, and `GroupByPublishDate` work fine any of the four date values.

In that unquoted TOML dates are the best choice[^1] for custom date params, this seems like something we should fix.

---

Reproducible example:

```text
git clone --single-branch -b hugo-github-issue-11563 https://github.com/jmooring/hugo-testing hugo-github-issue-11563
cd hugo-github-issue-11563
hugo server
```

[^1]: With unquoted TOML dates you can do `time.Time` comparisons wit the `where` function.

jmooring:
Closes #11563

### DIFF
--- a/resources/resource/resource_helpers.go
+++ b/resources/resource/resource_helpers.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/gohugoio/hugo/helpers"
+	"github.com/pelletier/go-toml/v2"
 
 	"github.com/spf13/cast"
 )
@@ -55,6 +56,10 @@ func getParam(r Resource, key string, stringToLower bool) any {
 		return cast.ToFloat64(v)
 	case time.Time:
 		return val
+	case toml.LocalDate:
+		return val.AsTime(time.UTC)
+	case toml.LocalDateTime:
+		return val.AsTime(time.UTC)
 	case []string:
 		if stringToLower {
 			return helpers.SliceToLower(val)
@@ -65,6 +70,5 @@ func getParam(r Resource, key string, stringToLower bool) any {
 	case map[any]any:
 		return v
 	}
-
 	return nil
 }
--- a/resources/integration_test.go
+++ b/resources/integration_test.go
@@ -81,10 +81,10 @@ func TestSVGError(t *testing.T) {
 	files := `
 -- config.toml --
 -- assets/circle.svg --
-<svg height="100" width="100"><circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" /></svg> 
+<svg height="100" width="100"><circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" /></svg>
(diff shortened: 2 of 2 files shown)


resources/resource/resource_helpers.go
```diff
@@ -55,6 +58,11 @@ func getParam(r Resource, key string, stringToLower bool) any {
 		return cast.ToFloat64(v)
 	case time.Time:
 		return val
+	case toml.LocalDate, toml.LocalDateTime:
+		if vt, ok := hreflect.AsTime(reflect.ValueOf(val), time.UTC); ok {
```
bep, line 62: Using reflect here isn't needed. You can just do `val.AsTime(time.UTC)`.

jmooring: I couldn't quite figure out how to do what you describe, but hopefully using `htime.ToTimeInDefaultLocationE(val, time.UTC)` is sufficient.

resources/resource/resource_helpers.go
```diff
@@ -55,6 +57,12 @@ func getParam(r Resource, key string, stringToLower bool) any {
 		return cast.ToFloat64(v)
 	case time.Time:
 		return val
+	case toml.LocalDate, toml.LocalDateTime:
+		val, err := htime.ToTimeInDefaultLocationE(val, time.UTC)
```
bep, line 61: This is nit picking, but but when you have a known type, converting it to another type (e.g. `interface{}` or `reflect.Value`) just so it can be converted back to the type you started with. I don't think it allocates any extra memory in this particular case, but I had to navigate into the ToTime... func to see what it does. I suggest you just call val.AsTime directly.

jmooring: Sorry about the iterations here. This was the self-inflicted wound:
```
case toml.LocalDate, toml.LocalDateTime:
```

<https://go.dev/ref/spec#Type_switches>

> In clauses with a case listing exactly one type, the variable has that type; otherwise, the variable has the type of the expression in the TypeSwitchGuard.

Which meant I had type `any` instead of the toml.Whatever type, which meant AsTime() was not an option. Split the case; all is good.

PR #11741, opened 2023-11-26
parser/metadecoders: Accumulate org keywords into arrays

augustfengd:
Hey all! What are your thoughts on this change?

The [keywords](https://orgmode.org/worg/org-syntax.html#Keywords) definition from the org syntax specification doesn't say we can't repeat keywords. 

This change will also enable terms in org mode to have space. 

Currently, repeated keywords are concatenated: `#+title: foo\n#+title: bar` would render as `foo bar`. With this change, it just doesn't render anything but hugo is not crashing either.

### DIFF
--- a/parser/metadecoders/decoder.go
+++ b/parser/metadecoders/decoder.go
@@ -242,9 +242,8 @@ func (d Decoder) unmarshalORG(data []byte, v any) error {
 		k = strings.ToLower(k)
 		if strings.HasSuffix(k, "[]") {
 			frontMatter[k[:len(k)-2]] = strings.Fields(v)
-		} else if k == "tags" || k == "categories" || k == "aliases" {
-			log.Printf("warn: Please use '#+%s[]:' notation, automatic conversion is deprecated.", k)
-			frontMatter[k] = strings.Fields(v)
+		} else if strings.Contains(v, "\n") {
+			frontMatter[k] = strings.Split(v, "\n")
 		} else if k == "date" || k == "lastmod" || k == "publishdate" || k == "expirydate" {
 			frontMatter[k] = parseORGDate(v)
 		} else {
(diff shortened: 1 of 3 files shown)


augustfengd:
Hey @jmooring, I've a created a [proposal](https://github.com/gohugoio/hugo/issues/11743). Thanks for the advice.

I would just note that multi-word strings are natively supported in org mode, otherwise something as `#+title: foo bar baz` would not work.

parser/metadecoders/decoder.go
```diff
@@ -242,9 +242,10 @@ func (d Decoder) unmarshalORG(data []byte, v any) error {
 		k = strings.ToLower(k)
 		if strings.HasSuffix(k, "[]") {
 			frontMatter[k[:len(k)-2]] = strings.Fields(v)
-		} else if k == "tags" || k == "categories" || k == "aliases" {
-			log.Printf("warn: Please use '#+%s[]:' notation, automatic conversion is deprecated.", k)
-			frontMatter[k] = strings.Fields(v)
+		} else if strings.HasSuffix(k, "[]") {
```
jmooring, line 245: You have a repeated condition here.

jmooring:
Test site:

```text
git clone --single-branch -b hugo-github-issue-11743 https://github.com/jmooring/hugo-testing hugo-github-issue-11743
cd hugo-github-issue-11743
hugo server
```

parser/metadecoders/decoder.go
augustfengd: Oh my, thanks for catching that. I've just fixed the commit.

augustfengd:
Amazing, thanks for the hint. I've just squashed the commits.

jmooring:
@bep I am in favor of merging this. The only users that will be negatively affected are those who have been ignoring a related warning for the past 4.5 years.

PR #11784, opened 2023-12-05
markup/goldmark: Render strikethrough and emojis in TOC entries

### LINKED ISSUE #7169: Strikethrough Markdown Not Reflected in ToC Heading Text

<pre>
$ hugo version
Hugo Static Site Generator v0.68.3-157669A0/extended darwin/amd64 BuildDate: 2020-03-24T12:11:24Z
</pre>

Yes

Given a heading that contains strikethrough markdown, and TableOfContents is being rendered, the title of the heading does not reflect the strikethrough markdown - whereas it does for strong and italic markdown.

Example of markdown:
`## Spice Up Your ~~Life~~ Python :snake:`

Appears in ToC as below.

incorrect-markdown-rendering-toc

Whereas it is rendered fine as a heading.

markdown-heading-render

### LINKED ISSUE #8087: Markdown .TableOfContents ignores strikethrough in headings
Strikethrough text in headings seems to be stripped out and only the text is output by `.TableOfContents`.

Tested in Hugo version: v0.79.1
Source format of content: markdown

I've a heading with strikethrough (e.g. `~~world~~`) like this

# Hello ~~world~~

which renders with a strikethrough in normal page rendering.  However on the TOC list I see

# Hello world

[Live example](https://legends2k.github.io/note/git_nuances/)

### LINKED ISSUE #11783: Strikethrough in markdown heading triggers runtime error

<pre>
$ hugo version
v0.121.0
</pre>

Yes (also tested with current `master`).

### Summary

Hugo crashes with a similar message to  
```
Error: error building site: render: failed to render pages: render of "term" failed: template: _internal/_default/rss.xml:67:22: executing "_internal/_default/rss.xml" at <.Summary>: error calling Summary: runtime error: index out of range [32] with length 20
```

when these two conditions are true:
- ~~table of contents (TOC) is generated or taxonomies are configured~~
- Markdown page contains a certain combination of ~~strike-through~~ markup

### Reproduction

The problem can easily be reproduced by cloning the following repo (instruction in README.md):
https://github.com/lyind/reproducer-2023-12-05

### Notes

Possibly related to these auto-closed issues:
- https://github.com/gohugoio/hugo/issues/7169
- https://github.com/gohugoio/hugo/issues/8087

lyind:
Configure the TOC (table of contents, `toc.go`) goldmark renderer instance to always enable the Strikethrough extension.

This allows handling `ast.KindStrikethrough` within an AST node avoiding a crash.

Fixes #7169
Fixes #8087
Fixes #11783

### DIFF
--- a/markup/goldmark/convert.go
+++ b/markup/goldmark/convert.go
@@ -18,14 +18,14 @@ import (
 	"bytes"
 
 	"github.com/gohugoio/hugo-goldmark-extensions/passthrough"
+	"github.com/yuin/goldmark/util"
+
 	"github.com/gohugoio/hugo/markup/goldmark/codeblocks"
 	"github.com/gohugoio/hugo/markup/goldmark/goldmark_config"
 	"github.com/gohugoio/hugo/markup/goldmark/images"
 	"github.com/gohugoio/hugo/markup/goldmark/internal/extensions/attributes"
 	"github.com/gohugoio/hugo/markup/goldmark/internal/render"
 
-	"github.com/gohugoio/hugo/markup/converter"
-	"github.com/gohugoio/hugo/markup/tableofcontents"
 	"github.com/yuin/goldmark"
 	emoji "github.com/yuin/goldmark-emoji"
 	"github.com/yuin/goldmark/ast"
@@ -34,6 +34,9 @@ import (
 	"github.com/yuin/goldmark/renderer"
 	"github.com/yuin/goldmark/renderer/html"
 	"github.com/yuin/goldmark/text"
+
+	"github.com/gohugoio/hugo/markup/converter"
+	"github.com/gohugoio/hugo/markup/tableofcontents"
 )
 
 const (
@@ -91,10 +94,17 @@ func newMarkdown(pcfg converter.ProviderConfig) goldmark.Markdown {
 		rendererOptions = append(rendererOptions, html.WithUnsafe())
 	}
 
+	tocRendererOptions := make([]renderer.Option, len(rendererOptions))
+	if rendererOptions != nil {
+		copy(tocRendererOptions, rendererOptions)
+	}
+	tocRendererOptions = append(tocRendererOptions,
+		renderer.WithNodeRenderers(util.Prioritized(extension.NewStrikethroughHTMLRenderer(), 500)),
+		renderer.WithNodeRenderers(util.Prioritized(emoji.NewHTMLRenderer(), 200)))
 	var (
 		extensions = []goldmark.Extender{
 			newLinks(cfg),
-			newTocExtension(rendererOptions),
+			newTocExtension(tocRendererOptions),
 		}
 		parserOptions []parser.Option
 	)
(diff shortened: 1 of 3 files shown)


jmooring:
@lyind There's a failing test, and you need to add conditions to test the issues that this PR addresses. Also, please capitalize the first letter of the commit message (after the `xxx/xxx:`  part) per the [commit message guidelines](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines).

Tolen-arch (approved):

lyind:
I seems like the rendering test fails, because the output is **more correct** now:
```diff
...
        line diff (-got +want):
            []string{
            	"<nav id=\"TableOfContents\">\n",
            	"  <ul>\n",
            	"    <li><a href=\"#a--b--c--d\">A &lt; B &amp; C &gt; D</a></li>\n",
            	strings.Join({
            		`    <li><a href="#a--b--c--d-divfoodiv">A &lt; B &amp; C &gt; D `,
            		"<",
          - 		"!-- raw HTML omitted -->foo<!-- raw HTML omitted --",
          + 		"div>foo</div",
...
```

But I might confuse "got" with "want", so will have another look at the code of that test case.

lyind:
@jmooring I applied the following changes to fix the test and be more aligned in general:
- Also pass rendering options (`HardWraps`, `XHTML`, `Unsafe`) to the TOC rendering extension
- Last test in TestEscapeToc() (`unsafe` renderer): expect `<div>` element to correctly be excluded from the TOC output

Is my assumption correct that the test expectations were slightly incorrect, in regard to the `<div>` and possibly similar block elements to be passed through?

jmooring:
> Also pass rendering options (HardWraps, XHTML, Unsafe) to the TOC rendering extension

Why do we need to do this? The `xhtml` and `unsafe` options are already honored when rendering a TOC entry, and you would never have a newline within a heading so the `hardWraps` setting is irrelevant.

Although I have not looked into the details, I find it curious that the strikethrough extension is not used when enabled, but the typographer extension _is_ used when enabled. Can you explain why that is?

jmooring:
Let's re-task this PR to fix these **open** issues:

- https://github.com/gohugoio/hugo/issues/7169
- https://github.com/gohugoio/hugo/issues/12022
- https://github.com/gohugoio/hugo/issues/11783

I've submitted a PR (https://github.com/gohugoio/hugo/pull/12023) that revises the test cases. The last two cases are commented out; they fail due to the open issues above.

My previous comment about **not** passing the rendering options (HardWraps, XHTML, Unsafe) stands.

lyind:
> The xhtml and unsafe options are already honored when rendering a TOC entry, and you would never have a newline within a heading so the hardWraps setting is irrelevant.

True. Just passed the options down out of "stubborn completeness", will change that again.

jmooring:
Functionally this is great! Thank you. 

The next step is a review by bep.

PR #11787, opened 2023-12-08
refactor: should omit nil check; len() for map[string]any is defined as zero

BugKillerPro:


phuchau1989 (approved):

bep:
@BugKillerPro can you get the commit message in line with the guidelines?

PR #11789, opened 2023-12-11
refactor: should omit nil check; len() for map[string]any is defined …

BugKillerPro:
should omit nil check; len() for map[string]any is defined as zero


bep:
Thanks, seems like this is already fixed.

Issue #11857, opened 2024-01-05
`hugo mod get` always updates modules to the latest version


ajschmidt8:
I use a specific version of [Docsy](https://github.com/google/docsy/) (`v0.7.1`) and currently have a `go.mod` file that looks like this:

```go.mod
module github.com/my/project

go 1.21.5

require (
	github.com/google/docsy v0.7.1 // indirect
	github.com/google/docsy/dependencies v0.7.1 // indirect
)
```

Whenever I run `hugo mod get` in my project, it automatically installs the latest version of these dependencies and updates the `go.mod` file accordingly.

This is not what I would expect to happen. I would expect that my specified versions of Docsy are respected and installed.

This is an issue because the latest version of Docsy is not compatible with my current source code (it has breaking changes).

I believe I've traced this issue down to this code block:

- https://github.com/gohugoio/hugo/blob/e40b9fbbcf28c07ad90ae1048e144888414074d7/modules/client.go#L347-L360

The `go` commands that ultimately get executed here are:

```sh
go get -d github.com/google/docsy/dependencies
go get -d github.com/google/docsy
```

Since the versions of these packages (`v0.7.1`) are omitted from the `go get` commands, it seems that `go` automatically updates them to the latest version.

This issue would be resolved if the `go` commands included the package versions like this:

```sh
go get -d github.com/google/docsy/dependencies@v0.7.1
go get -d github.com/google/docsy@v0.7.1
```

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.121.1-00b46fed8e47f7bb0a85d7cfc2d9f1356379b740+extended linux/amd64 BuildDate=2023-12-08T08:47:45Z VendorInfo=gohugoio
</pre>

### Does this issue reproduce with the latest release?

Yes

jmooring:
Related: https://github.com/gohugoio/hugo/issues/7475

bep:
This is in line with the documentation, see

https://gohugo.io/commands/hugo_mod_get/

If you want to run with what you currently have in `go.mod` then don't run any `hugo mod get` command.

bep:
Also note, that if you just want to download modules/confirm your setup without building you can do e.g.

```
hugo config
```

Or similar.

ajschmidt8:
> Also note, that if you just want to download modules/confirm your setup without building you can do e.g.
> 
> ```
> hugo config
> ```
> 
> Or similar.

[...]

PR #11861, opened 2024-01-05
Extend Documentation for `hugo config`

ajschmidt8:
Related to #11857.

This PR documents that the `hugo config` command downloads Hugo modules without updating them.

To my knowledge, this information isn't documented anywhere currently.

### DIFF
--- a/docs/content/en/commands/hugo_config.md
+++ b/docs/content/en/commands/hugo_config.md
@@ -9,7 +9,7 @@ Print the site configuration
 
 ### Synopsis
 
-Print the site configuration, both default and custom settings.
+Print the site configuration, both default and custom settings, and install modules from "go.mod".
 
 ```
 hugo config [command] [flags]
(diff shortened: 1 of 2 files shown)


jmooring:
Thanks for this. A few notes...

First, the CLI documentation (docs/content/en/commands/*.md) is generated from code using `hugo gen doc`. You can't update those pages directly.

Second, modules are downloaded when you run `hugo`, `hugo config`, `hugo convert`, `hugo list`, `hugo mod`, `hugo new`, and `hugo server`. Or to put it another way, the only sub-commands that do _not_ download modules are those that don't need them: `hugo completion`, `hugo env`, `hugo gen`, `hugo help`, and `hugo version`.

Finally, if you feel these details should be documented, please open an issue in the [documentation repository](https://github.com/gohugoio/hugoDocs).

Thanks.

Issue #10894, opened 2023-04-12
markup/goldmark: Enable pass-through of raw content blocks


janhuenermann:
**Continuing the discussion from https://github.com/gohugoio/hugo/pull/10888 and [the forum](https://discourse.gohugo.io/t/katex-in-hugo/43274).** 

Hey everyone,

Following up on the PR, I wanted to open an issue to discuss an improvement to LaTeX support in Hugo. Standard LaTeX in Markdown support has been requested by the community for a while, especially among the scientific blogging community.  However, Hugo only supports a special syntax that requires modifying existing Markdown files to work with Hugo. Platforms like GitHub already support the standard dollar-sign syntax, e.g. that's an inline expression: $x^2$.

Here's the proposal:
- add a Goldmark extension to Hugo that finds all inline and block LaTeX in Markdown file and emit them as plain text. This can be enabled using a config flag and is disabled by default.
- With this extension, users can then include the Katex client side scripts and stylesheets, enabling them to render LaTeX without modifying their existing Markdown files.
- This proposal doesn't require any additional dependencies and doesn't involve server side rendering of the LaTeX expressions (which needs a JavaScript engine and is potentially slow)

This is already implemented here: https://github.com/janhuenermann/hugo-katex/commit/ad382464769740c73235cbc3b20c4a5143408aa2

Looking forward to hearing your thoughts on this!

bep:
>LaTeX in Markdown file 

What is "LaTeX in Markdown"? If this is about adding yet another delimiter for this particular synta, why not use a shortcode (for inline latex) and code fences for block latex?

janhuenermann:
> > LaTeX in Markdown file
> 
> What is "LaTeX in Markdown"? If this is about adding yet another delimiter for this particular syntax, why not use a shortcode (for inline latex) and code fences for block latex?

Hey @bep,

[...]

PR #11866, opened 2024-01-06
markup/goldmark: Support passthrough extension

j2kun:
Fixes https://github.com/gohugoio/hugo/issues/10894

Creating as a draft so that it can be discussed in the issue, and while I work on adding additional tests. Implementation by @janhuenermann, tests by me.


j2kun:
TODO:

- test multi-line inline mathmode
- test multi-line displaymode mathmode
- test escaped `$` is not treated as mathmode

j2kun:
Trying the initial integration with my blog, I'm seeing some issues with the renderer. When I configure my personal block via https://github.com/j2kun/math-intersect-programming/commit/892f42db29a388ae18e7aab70f6b3d4a579b66a4:

```
[markup.goldmark.extensions.passthrough]
  enable = true
  blockDelimiters = [
    {open = "$$", close = "$$"},
    {open = "\\[", close = "\\]"},
  ]
  inlineDelimiters = [
    {open = "$", close = "$"},
    {open = "\\(", close = "$\\)"},
  ]
```

I see

```
ERROR render of "page" failed: "/home/j2kun/blog/themes/paperesque/layouts/_default/baseof.html:9:7": execute of template failed: template: _default/single.html:9:7: executing "_default/single.html" at <partial "meta.html" .>: error calling partial: execute of template failed: template: _internal/opengraph.html:2:98: executing "_internal/opengraph.html" at <.Summary>: error calling Summary: runtime error: invalid memory address or nil pointer dereference
Total in 206 ms
Error: error building site: render: failed to render pages: render of "page" failed: "/home/j2kun/blog/themes/paperesque/layouts/_default/baseof.html:9:7": execute of template failed: template: _default/single.html:9:7: executing "_default/single.html" at <partial "meta.html" .>: error calling partial: execute of template failed: template: _internal/opengraph.html:2:98: executing "_internal/opengraph.html" at <.Summary>: error calling Summary: runtime error: invalid memory address or nil pointer dereference
```

Based on the `opengraph.html`, this seems related to hugo, not the theme I'm using: https://github.com/gohugoio/hugo/blob/a541e3b4d48cac704185a644539d3501ff183b19/tpl/tplimpl/embedded/templates/opengraph.html

But I'm not sure how to narrow this down further. Could be an issue with my building hugo at head (a541e3b4d48cac704185a644539d3501ff183b19) + this PR

jmooring:
I may have some time to look at this tomorrow, but the config structure is different that what we (I) discussed here:
<https://github.com/gohugoio/hugo-goldmark-extensions/pull/2#issuecomment-1899076657>

Of the two options, I like the second one better.

Whether a delim is an opener or a closer should be inferred from its position in the array. There's an [example of this](https://gohugo.io/getting-started/configuration/#configure-minify) in the config structure for [tdewolff/minify](https://github.com/tdewolff/minify), e.g., 


```text
[minify.tdewolff.html]
    templateDelims = ['<?php', '?>']
```

markup/goldmark/integration_test.go
```diff
@@ -711,3 +711,76 @@ echo "hello";
...
+---
+## LaTeX test
+
+Inline equation that would be mangled by default parser: $a^*=x-b^*$
+
+-- layouts/_default/single.html --
+{{ .Content }}
+`
+
+	b := hugolib.NewIntegrationTestBuilder(
```
bep, line 738: There is a convience func that makes these tests shorter:

```go
b := hugolib.Test(t, files)
b.Assert...
```

jmooring:
I haven't tried to reproduce the error you encountered, but I noticed that your config is probably not what you intended:

```text
inlineDelimiters = [
  {open = "$", close = "$"},
  {open = "\\(", close = "$\\)"},  # should be {open = "\\(", close = "\\)"}
]
```

jmooring:
On the Hugo side of this (not the extension), the default config should include the four pairs of delimiters [described here](https://github.com/gohugoio/hugo-goldmark-extensions/pull/2#issuecomment-1899076657), with the ability to override to replace, remove, or add pairs of delimiters.

For example, if my markdown includes LaTeX mathematical expressions/equations, this should be all I need to do[^1]:

```yaml
markup:
  goldmark:
    extensions:
      passthrough:
        enable: true
```

If I want to avoid fighting with dollar signs in markdown (e.g., "the price is $2.00"), I should be able to do this:

```yaml
markup:
  goldmark:
    extensions:
      passthrough:
        enable: true
        delimiters:
          inline:
            - ['\(','\)'] 
          block:
            - ['\[','\]']
```

[^1]: Assumes you are loading KaTeX or MathJax.

jmooring:
@j2kun Here is one of several posts that is triggering the error:

```text
content/posts/2017-08-14-notes-on-math-and-gerrymandering.md
```

Specifically, this is the bit that is causing the problem:

```text
\[Soapbox\] ... \[/Soapbox\]
```

Which can be reduced to:

```text
\[x\] \[x\]
```

The error is not specific to your site or theme. It can be reproduced in a test case with:

```
---
title: foo
---

\[x\] \[x\]
```

markup/goldmark/integration_test.go
j2kun: Done

j2kun:
I made the changes requested to the hugo extension:

- Changed config structure
- Added default delimiters for LaTeX math mode (and tested overrides)
- Used simpler test invocation

Fixing the double-block bug in the parser next

j2kun:
I think it should be good now, tested against https://github.com/gohugoio/hugo-goldmark-extensions/pull/2/commits/2ce467c143ef766bf03b6e7a0709e388fa877ad3 with my whole blog and at least it doesn't crash. I will have to do some more investigating on my blog posts, because the Wordpress->hugo export chain I used did some funky stuff with my typeset math (e.g., all backslashes are escaped, all brackets are escaped, which led to that funky `\[ Soapbox \]` line, while my actual block math mode was converted to `\\\[ ... \\\]`)

jmooring:
I wanted to disable the `$` delimiters, both inline and block, so did this in site config:

```text
[markup.goldmark.extensions.passthrough]
  enable = true

[markup.goldmark.extensions.passthrough.delimiters]
  block = [['\[', '\]']]
  inline = [['\(', '\)']]
```

But that has no effect when testing. When I type `hugo config | grep passthrough -A3` I see:

```text
[markup.goldmark.extensions.passthrough]
  enable = true

  [markup.goldmark.extensions.passthrough.delimiters]
    block = [['\[', '\]'], ['\[', '\]']]
    inline = [['\(', '\)'], ['\(', '\)']]
```

jmooring:
My previous test was invalid... this works great to disable the `$` inline delimiter. However, the stored config is still incorrect.

site config

```text
[markup.goldmark.extensions.passthrough.delimiters]
inline = [['\(', '\)']]
```

command line 

```
hugo config | grep passthrough -A3
```

result  

```text
[markup.goldmark.extensions.passthrough]
  enable = true

  [markup.goldmark.extensions.passthrough.delimiters]
    block = [['$$', '$$'], ['\[', '\]']]
    inline = [['\(', '\)'], ['\(', '\)']]
```

jmooring:
@j2kun I think we should remove the single `$` inline delimiter from the default configuration:

```text
[markup.goldmark.extensions.passthrough.delimiters]
block = [['$$', '$$'], ['\[', '\]']]
inline = [['\(', '\)']]
```

Neither KaTeX.js nor MathJax.js enables the single `$` inline delimiter by default. Our config would be consistent with theirs, and out-of-the-box users would not have to escape single `$` used in currency, etc.

markup/goldmark/integration_test.go
```diff
@@ -711,3 +711,139 @@ echo "hello";
...
+func TestPassthroughOverriddenDefaultFences(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- config.toml --
+[markup.goldmark.extensions.passthrough]
+enable = true
+[markup.goldmark.extensions.passthrough.delimiters]
+inline = [['(((', ')))']]
+block = [['%!%'%!%']]
```
jmooring, line 817:
```suggestion
block = [['%!%', '%!%']]
```

jmooring:
Related to a previous comment...

I expected that I would be able to disable inline parsing with this:

```text
[markup.goldmark.extensions.passthrough.delimiters]
inline = []
```

But the config is unchanged.

```text
$ hugo config | grep -A3 passthrough

[markup.goldmark.extensions.passthrough.delimiters]
block = [['$$', '$$'], ['\[', '\]']]
inline = [['$', '$'], ['\(', '\)']]
```

j2kun:
> I think we should remove the single `$` inline delimiter from the default configuration:

If the goal is to have an extension that is agnostic of LaTeX, then my thought is to have no default configured delimiters. Make the user explicitly set them always. It would cause fewer surprises for people who want to use it for something besides TeX math mode, and would result in less code on the hugo side.

If you disagree, I'll go ahead and fix both the issues above (defaults not cleared and remove `$` as a default inline fence).

jmooring:
> then my thought is to have no default configured delimiters

You're right; thank you. We have to document this functionality anyway, including how to add and remove delimiter pairs. Please proceed.

markup/goldmark/integration_test.go
```diff
@@ -711,3 +711,139 @@ echo "hello";
...
+{{ .Content }}
+`
+
+	b := hugolib.Test(t, files)
+	b.AssertFileContent("public/p1/index.html", `
+		$$a^*=x-b^*$$
+	`)
+}
+
+func TestPassthroughUseDefaultFences(t *testing.T) {
```
jmooring, line 772: Failing test. We no longer have default fences. I think you can remove this test.

markup/goldmark/integration_test.go
```diff
@@ -711,3 +711,139 @@ echo "hello";
...
+		$a^*=x-b^*$
+	`)
+	b.AssertFileContent("public/p1/index.html", `
+\[
+a^*=x-b^*
+\]
+	`)
+}
+
+func TestPassthroughOverriddenDefaultFences(t *testing.T) {
```
jmooring, line 808: Rename test. We no longer have default fences.

jmooring (changes requested):

jmooring:
Also, please squash the commits with a new commit message: 

```text
markup/goldmark: Support passthrough extension

Fixes #10894
```

We capitalize the first word after the "foo/bar: " bit per:
https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines

markup/goldmark/goldmark_config/config.go
```diff
@@ -160,3 +190,5 @@ type ParserAttribute struct {
 	// Enables custom attributeds for blocks.
 	Block bool
 }
+
+// FIXME: add delimiters struct and passthrough config
```
jmooring, line 194:
```suggestion
```

j2kun:
Address all comments.

jmooring (approved):

bep:
This looks solid, great work!

PR #11880, opened 2024-01-11
deps: Update github.com/tdewolff/minify/v2 v2.20.9 => v2.20.13

jtatum:
KeepConditionalComments was deprecated in the upstream library and replaced with KeepSpecialComments. This new option reflects that both conditional comments and Server Side Include comments can be optionally stripped by the minifier. As with KeepConditionalComments, the minifier is configured not to strip them by default.

### DIFF
--- a/docs/data/docs.yaml
+++ b/docs/data/docs.yaml
@@ -1265,7 +1265,7 @@ config:
         precision: 0
       html:
         keepComments: false
-        keepConditionalComments: true
+        keepSpecialComments: true
         keepDefaultAttrVals: true
         keepDocumentTags: true
         keepEndTags: true
--- a/go.mod
+++ b/go.mod
@@ -61,8 +61,8 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/fsync v0.9.0
 	github.com/spf13/pflag v1.0.5
-	github.com/tdewolff/minify/v2 v2.20.9
-	github.com/tdewolff/parse/v2 v2.7.6
+	github.com/tdewolff/minify/v2 v2.20.13
+	github.com/tdewolff/parse/v2 v2.7.8
 	github.com/yuin/goldmark v1.6.0
 	github.com/yuin/goldmark-emoji v1.0.2
 	go.uber.org/atomic v1.11.0
@@ -143,7 +143,7 @@ require (
 	golang.org/x/crypto v0.17.0 // indirect
 	golang.org/x/mod v0.14.0 // indirect
 	golang.org/x/oauth2 v0.15.0 // indirect
-	golang.org/x/sys v0.15.0 // indirect
+	golang.org/x/sys v0.16.0 // indirect
 	golang.org/x/time v0.5.0 // indirect
 	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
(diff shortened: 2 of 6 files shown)


jmooring (changes requested):
This change is not backward compatible. If you have this in your site config:

```text
[minify.tdewolff.html]
keepConditionalComments = false
```

The conditional comments will still be present.

Also, with this:

```text
[minify.tdewolff.html]
keepConditionalComments = true
```

...the console is flooded with the message below, once for every page:

> DEPRECATED: KeepConditionalComments is replaced by KeepSpecialComments

jtatum:
> This change is not backward compatible. If you have this in your site config:
> 
> 
> 
> ```text
> 
> [minify.tdewolff.html]
> 
> keepConditionalComments = false
> 
> ```
> 
> 
> 
> The conditional comments will still be present.
> 
> 
> 
> Also, with this:
> 
> 
> 
> ```text
> 
> [minify.tdewolff.html]
> 
> keepConditionalComments = true
> 
> ```
> 
> 
> 
> ...the console is flooded with the message below, once for every page:
> 
> 
> 
> > DEPRECATED: KeepConditionalComments is replaced by KeepSpecialComments
> 
> 

@jmooring The deprecated message comes from the upstream library, which essentially collapsed the KeepConditionalComments option into the KeepSpecialComments option. We can add a special case to automatically set KeepSpecialComments false if KeepConditionalComments is false? Would that make sense? It's a little more than what the user requested...

jtatum:
@jmooring I think this should do it, and I validated that both keepconditionalcomments = false and = true work in config.

jtatum:
> When both are present in config, the newer one must win, e.g.
> 
> ```
> keepConditionalComments = false
> keepSpecialComments = true
> ```

If anything, this should be an error. The upstream library does not have separate code for keepSpecialComments vs keepConditionalComments - they're treated as the same option. Not sure I see the value in handling setting both? Is there precedent for doing this, or raising a config error like this to the user?

jmooring:
This now works as expected. Couple of other things...

1\) While the body's open, do the same thing for decimal/precision handling (currently the old one wins), something like:

<details>
<summary>minifiers/config.go</summary>

```text
for _, key := range []string{"css", "svg"} {
	if v, found := tdm[key]; found {
		vm := maps.ToStringMap(v)
		ko := "decimal"
		kn := "precision"
		if vv, found := vm[ko]; found {
			if _, found = vm[kn]; !found {
				vvi := cast.ToInt(vv)
				if vvi > 0 {
					vm[kn] = vvi
				}
			}
			delete(vm, ko)
		}
	}
}

if v, found := tdm["html"]; found {
	vm := maps.ToStringMap(v)
	ko := "keepconditionalcomments"
	kn := "keepspecialcomments"
	if vv, found := vm[ko]; found {
		if _, found = vm[kn]; !found {
			vm[kn] = cast.ToBool(vv)
		}
		delete(vm, ko)
	}
}
```
</details>


2\) Change the commit message and PR title to "deps: Update github.com/tdewolff/minify/v2 v2.20.9 => v2.20.14". This is consistent with how we've handled these before.

3\) Open an a new issue about deprecating all three keys.

jmooring (approved):

jmooring:
@jtatum Thanks, this works great. I'm sure bep will review/merge this when convenient.

jtatum:
> @jtatum Thanks, this works great. I'm sure bep will review/merge this when convenient.

Thank you for all the mentorship! I appreciate your time.

Issue #5139, opened 2018-08-30
Add role based output (think membership sections, RBAC)


bep:
Note that this proposal isn't a "ready to implement" one. These are some initial thoughts. And these thoughts skip the parts that live outside of Hugo (authentication etc.).

The current options for this with Hugo include

* Something like `htaccess` with Apache. Configuration can be managed and generated by Hugo, but you need to map that model yourself.
* Some third party service like [Google CLOUD IDENTITY-AWARE PROXY
](https://cloud.google.com/iap/) or [Netlify Access Control](https://www.netlify.com/docs/visitor-access-control/) or you build your own using the mechanisms in [AWS](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/PrivateContent.html)

The downsides of the above are

* It can be very complex and potentially very expensive
* It only allows you to allow or restrict access to a set of resources, i.e. URLs. E.g.: "You have to have the role MEMBER to view pages below /member", "You have to be ADMIN to view pages below /admin" etc.

If you want to, say, show a different home page for members, you will have to go the cumbersome route of somehow building your pages with secured server-side services. Which partially removes the static part and you end up with some ... **[jam](https://jamstack.org/)**.

This proposal is simple: 

* Add **Role** as a new concept in Hugo
* There should be a **Role** [hierarchy](https://docs.spring.io/spring-security/site/docs/5.0.7.RELEASE/reference/htmlsingle/#authz-hierarchical-roles)
* It should be possible to override templates for a given role, e.g. `home.member.html`.
* It should be possible to do role membership checking in the template: `{{ if .Site.HasRole "member" }}{{ end }}`
* A way to restrict output for roles.

The above is simplified. There are some questions on how to do this effectively with many roles to avoid duplication etc., but in it's simplest form with 2 roles you may end up with something like this:

```bash
site
└── public
    ├── member
    │   └── index.html
    └── user
        └── index.html
```

The above will solve both `authorization`  (_you need to be a member to have access to resource x_) and `presentation` (_if not a member, show the registration link_).

[...]

PR #11894, opened 2024-01-17
all: Rework page store, add a dynacache to enable bigger data/content, and some general spring cleaning

### LINKED ISSUE #7425: Make transform.Unmarshal scale to large(r) data sets
`transform.Unmarshal` caches the result using an unbounded custom "named cache", which is great in the common case -- but not fantastic if you have hundreds of gigabyte of JSON. 

So, we need to add a LRU/size eviction policy. I have looked at this before, which strikes me as both simple and with the right feature set.

https://github.com/karlseguin/ccache

### LINKED ISSUE #7436: Resource transformation key conflict when multiple languages in multihost mode and each with its own content dir
Which would mean that:

```
{{ .Resources.GetMatch "*.css" | minify }}
```

Would give a cached and wrong result for the other languages if the CSS filename is the same.

This would not be the case if you have all content living in the same content folder, e.g.:

```bash
index.md
style.en.css
style.fr.css
```

### LINKED ISSUE #7437: Language code in resource filename kept in RelPermalink
As in the example below:

```bash
index.md
style.en.css
style.fr.css
```

The language indicators tells Hugo what language/site they belong to, but currently these would be published to `https://example.com/style.en.css` and `https://example.com/style.fr.css`.

The above was never intended, and probably the reason why we did not discover #7436

### LINKED ISSUE #7544: ESBuilder not rebuilding when an imported JSON content changes
When an imported JS file changes, ESBuilder rebuilds as excepted, however it doesn't rebuild when an imported JSON changes (for example: `import data from "../../data/data.json";`)

<pre>
v0.74.3
</pre>

Yes

### LINKED ISSUE #7882: Fingerprint not recalculated in server mode for page resources
I'm not sure what's happening, but I got this failing test when working the memory caches.

### LINKED ISSUE #8307: FirstSection returns wrong value for taxonomy nodes
Working on something else, I have a failing test that looks like it should not. `mycategory.FirstSection` shold return the `categories` branch, now it returns the home page.

### LINKED ISSUE #8498: disableKinds with taxonomy should be recursive (also disable term)
While doing some restructuring work, I have a failing test that is a little head-scratching.

I have `disableKind = ["taxonomy"]`, yet I have an assertion about a term page that should exist.

I know changing this may potentially be breaking, but I suspect that I applied the above logic before we got the build options, which would allow you to disable rendering of, say, the `categories` list page, while still keep all the terms.

### LINKED ISSUE #8927: Add outputFormat.Ugly
I'm in the process of getting the 404 and some other similar pages into the same data structure as the rest, and in its default setup, 404 pages is _ugly_, that is, they get rendered to `/404.html`. 

This has been asked for by others before. The current workaround for similar stuff is to set the global `uglyURLs=true` and then set `noUgly=true` on the `OutputFormat`. This isn't particular practical.

### LINKED ISSUE #9192: Remove the purell library, SanitizeURL etc.
There is an undocumented `safe.SanitizeURL` template func that needs to be deprecated/removed. Other than that I see only one usage of this, and that looks like a mistake.

### LINKED ISSUE #9324: Wrong Permalink for resource in multihost mode when baseURL in one language has a sub path
I'm doing some  code simplification and have a test failure that looks like a bug.

I have a case with 
* Multiple language, each with it's own `baseURL`
* The English language also has a sub-path (`baseURL = "https://example.com/docs"`), the others not
* Still, the resources for French gets prefixed with the `/docs` path.

### LINKED ISSUE #9343: Create TranslationBaseName from filename only
I'm  in the last leg of some improvements in Hugo which requires some simplification in the _content path_ logic . This is also looking a little to the future where content will not just be files, and the task of finding the _cannonical path_ of a piece of content should be a simple task.

Given this content:

```
content
├── foo.en.md
├── foo.jp.md
└── foo.no.md
```

Currently, the value of `.Page.File.TranslationBaseName`, depends on the language configuration. If only `en` is configured, then `TranslationBaseName` for `foo.jp.md` will be `foo.jp`.

In the new setup, every part of the _base filename_ after the first "." separator will be considered _identifiers_:

```
 foo.no.md =>  foo + IDENTIFIER_LANGUAGE + IDENTIFIER_EXTENSION
```

Note that the `Page` backed by a `foo.unknownlanguage.md` will still be assigned the `defaultContentLanguage`.

I'm sure there will be some subtle site breakage in the above, but I'm also sure that we need to do this.

Shout if you disagree.

### LINKED ISSUE #10104: Preserve triple hyphens etc. in URLs.
This relates to #7288 -- which was fixed.

I create this just to avoid some yelling about "broken URLs" from @kaushalmodi and friends.

I'm doing some ground work to simplify stuff and also make stuff faster and more effective.

So with #7288 the issue was 

>content/foo--bar.md, the page URL is expected to be <site>/foo--bar/.

With that it does not make much sense to say that `content/foo---bar.md` becomes `foo-bar/`.

We have a test case for the above, but I will change that so that any number of subsequent hyphens are preserved...

### LINKED ISSUE #10380: Default title for nested taxonomies should be the base name
Currently it's possible to do:

```toml
[taxonomies]
"news/category" = "news/categories"
```

And the taxonomy pages will get the paths as expected. The default title, however (when no `_index.md` file is found) is set to `News/Categories`, which I suspect is not a great title.

I'm doing some simplification work in this department now, and had a failing test above, so I'll create an issue to mark that I'll change it so the title becomes the base name (e.g. `Categories`).

### LINKED ISSUE #10694: Build options are ignored when generating the taxonomies map
With this in site configuration:

```text
[[cascade]]

[cascade._build]
render = "never"
list = "never"
publishResources = false

[cascade._target]
path = '/hidden/**'
```

The taxonomies map unexpectedly contains:
- Taxonomies that are only used in the "hidden" section
- Terms that are only used in the "hidden" section
- Weighted pages from the "hidden" section 

This results in unexpected entries in:
- Taxonomy pages
- Term pages
- RSS feeds
- Sitemap

Example:

```text
git clone --single-branch -b hugo-github-issue-10694 https://github.com/jmooring/hugo-testing hugo-github-issue-10694
cd hugo-github-issue-10694
hugo && grep -ri hidden public/*
```

Workaround:

```text
[[module.mounts]]
source = 'content'
target = 'content'
excludeFiles = '/hidden/**.*'
```

### LINKED ISSUE #11439: Remove the "no resources in sub folders" restriction for branch bundles
I'm working on some cool stuff that requires some simplifications in how we process `/content` files.
 
Currently

* resources for branch bundles (e.g. sections, home page) need to live in the  root folder of that bundle.
* every non-content-file inside `/content` that's not part of a bundle gets copied to /public as it would live in `/static`.

```bash
content
├── _index.md
├── foo.png
└── posts
    ├── _index.md
    ├── baz.png
    ├── images
    │   ├── image1.png
    │   ├── image2.png
    │   └── image3.png
    ├── p1.md
    └── p2
        ├── bar.png
        └── index.md
```

Given the above example, the images in `posts/images` will with this proposal be part of the `posts` section bundle.

That will in some (rare?) setups be a breaking change but I would argue that

1. It feels more natural and is easier to understand/document.
2. It will probably solve more problems that it breaks.

/cc @jmooring

### LINKED ISSUE #11453: Don't duplicate page resource output per language
We're planning on some big improvements for the next Hugo. A part of that is to make Hugo simpler and more effective in the way we handle content, also to pave away for features going forward. And with that comes some changes which some may consider breaking, more would consider bug fixes.

Given this example content folder and 2 languages defined:

```bash
content
└── mybundle
    ├── data.en.json
    ├── index.en.md
    ├── index.no.md
    └── sunset.en.jpg
```

In the current version of Hugo we would duplicate the output so you would get:

```
public
├── en
│   └── mybundle
│       ├── data.json
│       ├── index.html
│       └── sunset.jpg
└── no
    └── mybundle
        ├── data.json
        ├── index.html
        └── sunset.jpg
```

The original implementation of the above preceded render hooks and the main motivation was that you should be able to have links that worked both on GitHub and when rendered in Hugo, even across languages.

We would still be able to apply language specific params/name/title per resource, but `.RelPermalink` would point to the same file. If you need a different file for a given language, you can easily create a language specific version (e.g. `data.no.json`).

There's several reasons why the above is a great idea:

* It reduces duplication, which is especially true for bigger files (e.g. images).
* It allows us to to relatively cheaply add new build dimensions in addition to languages (e.g. roles)
* Also, when we "stop caring" about where these objects lives on the file system (stop hard coding URLs), we can optimise further by de-duplication (e.g. publish all images to one folder named by their content hash).

### LINKED ISSUE #11455: Some breaking changes planned for Hugo 0.123.0
We're planning some big improvements for the next Hugo, and with that comes some changes which some may consider breaking, but I suspect the majority would consider these to be bug fixes/improvements.

**This list is a work in progress.** 

- [x] #11654
- [x] #11453
- [x] #11663
- [x] #11439
- [x] #11556
- [x] #11669
- [x] #11807
- [x] #11808
- [x] #11809
- [x] #11815 
- [x] #10104
- [x] #9192
- [x] #11860

Note to self: Merge #11549 and this issue at some point.

### LINKED ISSUE #11457: Rename site.LastChange => site.Lastmod
To get it in line with the reset. I always get this wrong when taking it from memory. We should keep the old for backwards compability.

### LINKED ISSUE #11466: cascade title not correctly applied to term pages
I'm reimplementing how _cascade and aggregated values (dates) are applied to the page tree, and I have a test failure re. cascade that is obviously failing for the wrong reason.

### LINKED ISSUE #11540: Page bundles linked by translationKey do not share page resources
```text
content/
├── en/
│   ├── foo/
│   │   └── foo-1/
│   │       ├── a.jpg
│   │       └── index.md   <-- translationKey = 'baz'
│   └── _index.md
└── pl/
    ├── bar/
    │   └── bar-1/
    │       └── index.md   <-- translationKey = 'baz'
    └── _index.md
```

The a.jpg image is not available as a page resource when rendering content/pl/bar/bar-1.md.

Although the documentation doesn’t specifically state that page resources are shared when using `translationKey` in front matter, it is certainly implied.

Reproducible example:

```text
git clone --single-branch -b hugo-forum-topic-46581 https://github.com/jmooring/hugo-testing hugo-forum-topic-46581
cd hugo-forum-topic-46581
hugo server
```

Reference: <https://discourse.gohugo.io/t/46581>

### LINKED ISSUE #11549: Mini issue tracker for Hugo 0.123.0
- [x] #10169
- [x] #10364
- [x] #10482
- [x] #10630
- [x] #10656
- [x] #10694
- [x] #10918
- [x] #11262
- [x] #11439
- [x] #11453
- [x] #11457
- [x] #11466
- [x] #11540
- [x] #11551
- [x] #11556
- [x] #11654
- [x] #11661
- [x] #11663
- [x] #11664
- [x] #11669
- [x] #11671
- [x] #11807
- [x] #11808
- [x] #11809
- [x] #11815 
- [x] #11840
- [x] #11853
- [x] #11860
- [x] #11883
- [x] #11904
- [x] #7388
- [x] #7425
- [x] #7436
- [x] #7544
- [x] #7882
- [x] #7960
- [x] #8255
- [x] #8307
- [x] #8863
- [x] #8927
- [x] #9192
- [x] #9324

### LINKED ISSUE #11551: Add outputformat.Root
For output formats which path should always start at the root and not in any language sub folder (e.g. robots.txt).

### LINKED ISSUE #11556: Remove symlink support for the input files
This relates to #11543 and others, a part of a major overhaul/simplification/improvement in how Hugo process content (where content includes all files being read and processed as part of the build process). 

This work is motivated by several things:

* We need to improve on the partial builds.
* We need Hugo to handle larger data sets and use less memory.
* We need to simplify things to make it possible to add the features planned, but also to make the code easier to maintain.

Working on the above, I have so far just ignored our support for symlinked directories, but now is the time for me to consider putting that support back into my development branch or not.

My main argument for removing it is that it should not really be needed anymore as we got virtual file mounts, e.g.:

```toml
[[module.mounts]]
source = "/foo/bar"
target = "content"
lang = "en"
```

In contrast to symlinks, these mounts are OS and file system agnostic (even work for files/dirs mounted from GitHub), and Hugo knows about them, so we can react to configuration changes.

Some other issues with symlinks are:

* They're very hard to reason about when files changes and we need to rebuild, and this becomes almost impossible on structural changes (symlinks that is deleted, renamed etc.)
*  There are security concerns, which is why Hugo only allows symlinks in the main project (not in themes/modules).
* They behave differently on different OSes (on Windows, you need to be an admin to create them).
* All this symlink checking also has a performance cost.

But most importantly, I don't want us to spend _many hours_ implementing and maintaining a feature that I think is very rarely used (I have never used it).

### LINKED ISSUE #11654: Don't duplicate page resource output per output format path
This relates to #11453.

For a given output format definition you can set a `path` (e.g. `amp`) which will render the output using that as the base path.

To enable easy guessing of relative URLs of bundled resources (e.g. JSON files), we have duplicated the output per sub path. This is both a waste of resources and is rather complicated.

This setup should be rare, and is not breaking if people use Hugo's API to resolve URLs (e.g. `(.Resources.Get "foo.txt").RelPermalink)`.

### LINKED ISSUE #11661: `fingerprint` resources not working on multihost except the default site
```go
{{ range .Page.Resources.Match "sample*" }}
  {{ $img := . | fingerprint }}
  <a href="{{ $img.RelPermalink }}" target="_blank">{{ .RelPermalink }}</a>
  <img src="{{ $img.RelPermalink }}" style="width: auto; height: 200px;">
{{ end }}
```

```toml
[languages.en]
weight = 1

[languages.zh-hans]
weight = 2
baseURL = "https://zh-hans.example.org/"
```

Content strucutre

```sh
$ tree content/hello
├── index.md
├── index.zh-hans.md
├── sample.gif
└── sample.jpg
```

The fingerprint resources won't be published to sites folder except the default site.

```sh
$ hugo

$ ls public/en/hello  
index.html                                                                   sample.gif
sample.862081bdf97cf8ca7ff865d76807d567830c9a0fa00ebd185be19eec4f09e7a2.gif  sample.jpg
sample.e2298d882fc6cabe5eda75d0efb87f4c42b506cff19fa60aed8104d07a585013.jpg

$ ls public/zh-hans/hello              
index.html  sample.gif  sample.jpg
```

image

<pre>
$ hugo version
hugo v0.120.3-a4892a07b41b7b3f1f143140ee4ec0a9a5cf3970+extended linux/amd64 BuildDate=2023-11-01T17:57:00Z VendorInfo=snap:0.120.3
</pre>

Yes

### LINKED ISSUE #11663: Don't create bundle header for header-less page resource content (e.g. md)
Given these files:

```
-- hugo.toml --
baseURL = "https://example.com"
disableKinds = ["taxonomy", "term"]
defaultContentLanguage = "en"
defaultContentLanguageInSubDir = true
[languages]
[languages.nn]
weight = 2
[languages.en]
weight = 1
-- content/section/mybundle/index.md --
---
title: "Mybundle"
---
-- content/section/mybundle/bundledpage.md --
---
title: "Bundled page en"
---
-- content/section/mybundle/bundledpage.nn.md --
---
title: "Bundled page nn"
---

Bundled page: {{ .RelPermalink}}|Len resources: {{ len .Resources }}|
```

In the current Hugo we have some logic that clones the `content/section/mybundle/index.md` for the `nn` language to make the bundled page `content/section/mybundle/bundledpage.nn.md` work.

This logic isn't something we can/want to maintain, especially when we consider getting content from other data sources. 

Ref. #6136

### LINKED ISSUE #11664: Wrong RelPermalink for content/mysect/index/index.md
The above will by default give `/mysect/`, which doesn't make much sense.

### LINKED ISSUE #11669: Always add a trailing slash if missing to the configured baseURL
E.g. 

* `https://example.com` => `https://example.com/`
* `https://example.com/docs` =>  `https://example.com/docs/`

This is an example of why it can be problematic to expose configuration as API (`site.BaseURL`).

Note that to me the important part is to standardise it, so we could do 

*  `https://example.com/` => `https://example.com`

If that's considered less breaking.

Note that I strongly recommend not using `site.BaseURL´ to construct URLs.

/cc @jmooring

### LINKED ISSUE #11671: Renamed content directories not added to the watcher

### LINKED ISSUE #11807: Term titles (when no content file) should follow the scheme of the other page kinds
I'm in the process of wrapping up some consolidation etc. work and I have this one failing test that didn't make much sense to me, but I notice that the titles for taxonomy terms ( e.g. `blue` in the `tags` taxonomy) are constructed by using the value as defined in front matter.

This may be considered a slightly breaking change, but we need to consolidate this, and I suspect most people would expect that the title case config would work also for these (https://gohugo.io/getting-started/configuration/#configure-title-case).

### LINKED ISSUE #11808: Nested sections (with no content file) should get their title from the base name
Also see #11807

A test is failing me, reminding me that some Hugo users creates a nested taxonomy structure, e.g:

```toml
[taxonomies]
"news/tag" = "news/tags"
"news/category" = "news/categories"
```

Which is all good and useful.

But in the current version of Hugo, if you don't have a content file backing e.g. `content/news/categories/_index.md`, the title is constructed using the full path (e.g. `News/Categories`).

To consolidate things and make this behave in a similar way for all of these different list kinds, we will change the above to use the branch base name, e.g. `content/news/categories/_index.md` => `categories` => title = `Categories` (depending on the configured title case).

### LINKED ISSUE #11809: Don't make draft/future/expired pages of branch nodes recursively disabled
This will be a breaking change for some, but it's meant well -- the current behaviour of special handling taxonomy pages is confusing and less flexible than it could be.

Currently, when we draft a taxonomy page (or e.g. mark it to be published in the future or it's expired):

`categories/_index.md`:

```md
---
title: "The Categories"
draft: true
---

Content.
```

We hide the entire tree from `/categories` and down.

The original intent/semantics of `draft` was to delay publishing of that particular page/content, so with this change we make kind=taxonomy work the same as sections. So, we say:

* For regular pages (leaf nodes) and term pages, the `Page` it's just removed/ignored.
* For home, section, taxonomy
    * we keep the Page node to make the page tree fully navigable.
    * the Page is not rendered and not in any page collection.
    * but you can reach it with `site.GetPage` (and the taxonomy in `site.Taxonomies`).

### LINKED ISSUE #11815: CurrentSection should navigate sideways for all branch nodes (also kind=taxonomy)
This issue is another part in an ongoing cleanup/consolidating job.

Given the below content tree, assuming that `categories` is defined as a taxonomy):

```bash
content
├── categories
│   └── _index.md
└── posts
    ├── _index.md
    └── p1.md
```

Current behaviour:

* (site.GetPage "posts").CurrentSection == (site.GetPage "posts")
* (site.GetPage "posts/p1").CurrentSection == (site.GetPage "posts")
* (site.GetPage "categories").CurrentSection == site.Home

So, `categories` isn't defined as a `section`, but having it behave differently in this context, is confusing.

This also means that calling `.FirstSection` on a `term` page will return the root `taxonomy` node (and not the home page).

### LINKED ISSUE #11840: Resource language fallback should be the default content language (if possible)
```handlebars
func TestBundleResourceLanguageBestMatch(t *testing.T) {
	files := `
-- hugo.toml --
defaultContentLanguage = "fr"
defaultContentLanguageInSubdir = true
[languages]
[languages.en]
weight = 1
[languages.fr]
weight = 2
[languages.de]
weight = 3
-- layouts/index.html --
{{ $bundle := site.GetPage "bundle" }}
{{ $r := $bundle.Resources.GetMatch "*.txt" }}
{{ .Language.Lang }}: {{ $r.RelPermalink }}|{{ $r.Content }}
-- content/bundle/index.fr.md --
---
title: "Bundle Fr"
---
-- content/bundle/index.en.md --
---
title: "Bundle En"
---
-- content/bundle/index.de.md --
---
title: "Bundle De"
---
-- content/bundle/data.fr.txt --
Data fr
-- content/bundle/data.en.txt --
Data en

`
	b := Test(t, files)

	b.AssertFileContent("public/fr/index.html", "fr: /fr/bundle/data.fr.txt|Data fr")
	b.AssertFileContent("public/en/index.html", "en: /en/bundle/data.en.txt|Data en")
	b.AssertFileContent("public/de/index.html", "de: /de/bundle/data.en.txt|Data en")
}

```

@jmooring if I could borrow your eyes for a minute.

The above test passes in the master branch, which surprises me a little. I guess most people have their `defaultContentLanguage` defined as their first language.

The surprise is this:

```bash
b.AssertFileContent("public/de/index.html", "de: /de/bundle/data.en.txt|Data en")
```

What fallback text file would you expect in the de bundle?

### LINKED ISSUE #11860: Pull the default content language to the top in site.Sites
The reason is rather technical, but I have rearranged the sites slice internally to match the most common sort order for these (for quick indexing), but to sum up:

Currently when you have languages defined with a sort order (weight, name etc.) that does not put the `defaultContentLanguage` on top, you get:

* site.Languages => sorted by weight, name etc.
* site.Sites => sorted by weight, name etc.

After this fix:

* site.Languages => sorted by weight, name etc.
* site.Sites => default content language at the top, then sorted by weight, name etc.

### LINKED ISSUE #11883: GetPage returns a page when I think it should return nil
I understand that `.GetPage` on both `Page` and `Site` have some fuzziness when finding a page, but I was surprised by this.

```text
content/
├── p1/
│   ├── index.md
│   └── p1.xyz
└── _index.md
```

layouts/_default/home.html

```text
{{ with .Page.GetPage "p1.xyz" }}
  .Page.GetPage: {{ .RelPermalink }}
{{ end }}

{{ with .Site.GetPage "p1.xyz" }}
  .Site.GetPage: {{ .RelPermalink }}
{{ end }}
```

public/index.html

```text
.Page.GetPage: /p1/

.Site.GetPage: /p1/
```

I would expect both methods to return nil because p1.xyz is a page resource, not a page.

bep:
This PR is big. I could/should probably try to split it, but the pieces are in many ways related, and splitting it would be hard.

There are some breaking changes in this commit, see https://github.com/gohugoio/hugo/issues/11455.

Closes https://github.com/gohugoio/hugo/issues/11455
Closes https://github.com/gohugoio/hugo/issues/11549

## What Changed

This fixes a set of bugs (see issue list) and it is also paying some technical debt accumulated over the years. We now build with [Staticcheck](https://github.com/dominikh/go-tools) enabled in the CI build.

The performance should be about the same as before for regular sized Hugo sites, but it should perform and scale much better to larger data sets, as objects that uses lots of memory (e.g. rendered Markdown, big JSON files read into maps with [transform.Unmarshal](https://gohugo.io/functions/transform/unmarshal/) etc.) will now get automatically garbage collected if needed. Performance on partial rebuilds when running the server in fast render mode should be the same, but the change detection should be much more accurate.

A list of the notable new features:

1. A new [dependency tracker](#dependency-tracker) that covers (almost) all of Hugo's API and is used to do fine grained partial rebuilds when running the server.
1. A new and simpler tree [document store](#document-store) which allows fast lookups and prefix-walking in all dimensions (e.g. language) concurrently.
1. You can now configure an upper [memory limit](#memory-limit) allowing for much larger data sets and/or running on lower specced PCs.
1. We have lifted the  "no resources in sub folders" restriction for branch bundles (e.g. sections).

### Memory Limit

Hugos will, by default, set aside a quarter of the total system memory, but you can set this via the OS environment variable `HUGO_MEMORYLIMIT` (in gigabytes). This is backed by a partitioned LRU cache used throughout Hugo. A cache that gets dynamically resized in low memory situations, allowing [Go's Garbage Collector](https://tip.golang.org/doc/gc-guide) to free the memory.

### Dependency Tracker

Hugo has had a rule based coarse grained approach to server rebuilds that has worked mostly pretty well, but there have been some surprises (e.g. stale content). This is now revamped with a new dependency tracker that can quickly calculate the delta given a changed resource (e.g. a content file, template, JS file etc.). This handles transitive relations, e.g. `$page -> js.Build -> JS import`, or `$page1.Content -> render hook -> site.GetPage -> $page2.Title`, or `$page1.Content -> shortcode -> partial -> site.RegularPages -> $page2.Content -> shortcode ...`, and should also handle changes to aggregated values (e.g. `site.Lastmod`) effectively.

This covers all of Hugo's API with 2 known exceptions (a list that may not be fully exhaustive):

1. Changes to files loaded with template func [os.ReadFile](https://gohugo.io/functions/os/readfile/) may not be handled correctly. We recommend loading resources with [resources.Get](https://gohugo.io/functions/resources/get/)
1. Changes to Hugo objects (e.g. `Page`) passed in the template context to [lang.Translate](https://gohugo.io/functions/lang/translate/) may not be detected correctly. We recommend having simple i18n templates without too much data context passed in other than simple types such as strings and numbers.

Note that the [cachebuster](https://gohugo.io/getting-started/configuration/#configure-cache-busters) configuration (when A changes then rebuild B) works well with the above, but we recommend that you revise that configuration, as it in most situations should not be needed. One example where it is still needed is with TailwindCSS and using changes to `hugo_stats.json` to trigger new CSS rebuilds.

### Document Store

Previously, a little simplified, we split the document store (where we store pages and resources) in a tree per language. This worked pretty well, but the structure made some operations harder than they needed to be. We have now restructured it into one Radix tree for all languages. Internally the language is considered to be a dimension of that tree, and the tree can be viewed in all dimensions concurrently. This makes some operations re. language simpler (e.g. finding translations is just a slice range), but the idea is that it should also be relatively inexpensive to add more dimensions if needed (e.g. [role](https://github.com/gohugoio/hugo/issues/5139)).

Closes #11455
Closes #11549

Fixes #10104
Fixes #10380
Fixes #10694
Fixes #11439
Fixes #11453
Fixes #11457
Fixes #11466
Fixes #11540
Fixes #11551
Fixes #11556
Fixes #11654
Fixes #11661
Fixes #11663
Fixes #11664
Fixes #11669
Fixes #11671
Fixes #11807
Fixes #11808
Fixes #11809
Fixes #11815
Fixes #11840
Fixes #11860
Fixes #11883
Fixes #7425
Fixes #7436
Fixes #7437
Fixes #7544
Fixes #7882
Fixes #8307
Fixes #8498
Fixes #8927
Fixes #9192
Fixes #9324
Fixes #9343


bep:
- [ ] Create an issue re `../p2`, see https://gist.github.com/jmooring/0b88af8890471cf26ed2096d5913efc7?permalink_comment_id=4835239#gistcomment-4835239

bep:
@jmooring I will let this PR sit as is over the weekend with the goal to merge it into the main branch early next week. This will obviously not be released right away (and even then probably a beta or 2), so there are plenty of changes to fix thing/make adjustments before it goes live.

PR #11924, opened 2024-01-26
docs: Use img shortcode from theme

jmooring:


bep:
Why not do this edit in the docs repo?

jmooring:
I may not understand the process of pull/push between the docs repo and this repo. In this case, the "img" shortcode was moved from the docs repo to the theme repo a while ago. But it returned to the docs repo this morning after the merge.

bep:
>I may not understand the process of pull/push between the docs repo and this repo.

* hugoDocs is master
* the only change going from hugo -> hugoDocs is edits done in this repo (which should be the edits that reflects new features, changes to code)
* the above looks like a bad merge situation; I try my best, but when there's a lot of merge conflicts, I may glitch.

bep:
Just to be clear: The only motivation behind the `/docs` subtree construct is to provide a versioned snapshot of the documentation in the GitHub source tarballs (which is also why we vendor the theme).

bep:
I'm closing this. This will eventually be synced.

PR #11997, opened 2024-02-06
Fix typos

coliff:


output/outputFormat.go
```diff
@@ -72,10 +72,10 @@ type Format struct {
 	// Setting this will make this output format control the value of
 	// .Permalink and .RelPermalink for a rendered Page.
 	// If not set, these values will point to the main (first) output format
-	// configured. That is probably the behaviour you want in most situations,
+	// configured. That is probably the behavior you want in most situations,
```
bep, line 75: I wrote this; I write "colour" and "behaviour", both of which is correct, no?

output/outputFormat.go
```diff
@@ -72,10 +72,10 @@ type Format struct {
...
 	// as you probably don't want to link back to the RSS version of a page, as an
 	// example. AMP would, however, be a good example of an output format where this
-	// behaviour is wanted.
+	// behavior is wanted.
```
bep, line 78: See above.

tpl/internal/resourcehelpers/helpers.go
```diff
@@ -38,7 +38,7 @@ func ResolveIfFirstArgIsString(args []any) (resources.ResourceTransformer, strin
 	return v2, v1, ok2
 }
 
-// This roundabout way of doing it is needed to get both pipeline behaviour and options as arguments.
+// This roundabout way of doing it is needed to get both pipeline behavior and options as arguments.
```
bep, line 41: See above.

watcher/filenotify/poller.go
```diff
@@ -201,7 +201,7 @@ func (r *recording) record(filename string) error {
 	r.FileInfo = fi
 
 	// If fi is a dir, we watch the files inside that directory (not recursively).
-	// This matches the behaviour of fsnotity.
+	// This matches the behavior of fsnotity.
```
bep, line 204: See above.

bep (changes requested):

PR #12004, opened 2024-02-06
GitHub workflows: bump actions to latest versions

deining:
This PR bumps GitHub workflow actions to their latest versions, thus avoiding deprecation warnings as seen e.g. [here](https://github.com/gohugoio/hugo/actions/runs/7803541200).

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -12,7 +12,7 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      - uses: dessant/lock-threads@08e671be8ac8944d0e132aa71d0ae8ccfb347675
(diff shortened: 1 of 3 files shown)


coliff (approved):

bep:
We use explicit commits for these for security reasons. Also, since veryfying commit hashes is the same work as actually doing it, I'll put this on my list to do myself.

deining:
> We use explicit commits for these for security reasons.

I understand.

> I'll put this on my list to do myself.

What about letting `dependabot` do the job in the future?

Adding that section to `.github/dependabot.yml` should do the job:

```
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      # Check for updates to GitHub Actions every weekday
      interval: "daily"
```

PR #12008, opened 2024-02-07
Make HTML behave exactly like other content formats

bep:
This currently only contains a test that illustrates the issue.

Without digging down trying to find the reason why this has changed, I think we agree that the current behaviour is less than useful, and that if we're going to break anything, we might as well break it in the direction of useful.

What if we say that we start **treating all content files the same (Markdown, HTML, Asciidoc etc.)**.

Does that sound like a plan?

### DIFF
--- a/helpers/general.go
+++ b/helpers/general.go
@@ -328,7 +328,7 @@ func PrintFs(fs afero.Fs, path string, w io.Writer) {
 	}
 
 	afero.Walk(fs, path, func(path string, info os.FileInfo, err error) error {
-		fmt.Println(path)
+		fmt.Fprintln(w, filepath.ToSlash(path))
 		return nil
 	})
 }
--- a/hugolib/integrationtest_builder.go
+++ b/hugolib/integrationtest_builder.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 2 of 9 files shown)


bep:
@jmooring this should be a controlled change as I have just "removed lots of code". I have added some test cases, but it would be good if you could have a look.

jmooring:
Using this example...

```text
git clone --single-branch -b hugo-github-issue-11999 https://github.com/jmooring/hugo-testing hugo-github-issue-11999
cd hugo-github-issue-11999
rm -rf
```

This content...

```text
content/
└── posts/
    └── post-1/
        ├── foo.jpg
        ├── index.md
        ├── with-front-matter.html
        ├── with-front-matter.md
        ├── without-front-matter-begins-with-<.html
        ├── without-front-matter-does-not-begin-with-<.html
        └── without-front-matter.md
```

Produces these results...

File|Page resource|Copied to pubic
:--|:--|:--
foo.jpg|:heavy_check_mark:|:heavy_check_mark:
with-front-matter.html|:x:|:x:
with-front-matter.md|:heavy_check_mark:|:x:
without-front-matter-begins-with-<.html|:heavy_check_mark:|:x:
without-front-matter-does-not-begin-with-<.html|:heavy_check_mark:|:x:
without-front-matter.md|:heavy_check_mark:|:x:

I think the second one should be a page resources to be consistent with how we handle the other content formats.

bep:
```
├── with-front-matter.html
├── with-front-matter.md
```

I will have a look (but I do suspect you would see the same if you add `with-front-matter.asciidoc` etc. -- as in, the same base name).

jmooring:
When we have two images (non-content resources) with the same base name there is no collision; both are available as resources, and both are copied to public.

But when we have two content files (content resources) with the same base name, only one is available as a resource. Not sure which one wins.

bep:
I need to take that issue (which is not related to this particular issue) in https://github.com/gohugoio/hugo/issues/12013 -- we currently pick one of the versions (based on a consistent sort order), which makes sense for regular pages, possibly not for page resources ... but it's hard to distinguish the two at this point, so I suspect I need to print a warning about it -- I don't see why you would need two content files with the same name.

Other than that ... this is OK?

jmooring:
> Other than that ... this is OK?

Yes!

PR #12016, opened 2024-02-08
Add images.Dither filter

### LINKED ISSUE #8598: Add image dithering filter to build low-tech websites

jmooring:
Closes #8598

I understand we may not want to add another dependency, but the package is light, fast, and feature-rich. I expected ordered dithering to be fast, which it is, but was pleasantly surprised to find that error diffusion isn't too bad either. It's pretty slick.

And you can define the dithering palette on a per-image basis. The example below sets the dithering palette to the dominant colors:

```text
{{ with resources.Get "original.jpg" }}
  {{ with .Resize "480x" }}
    {{ $opts := dict "method" "ClusteredDot4x4" "colors" .Colors }}
    {{ $filter := images.Dither $opts }}
    {{ with images.Filter $filter . }}
    {{ end }}
  {{ end }}
{{ end }}
```
original_hu2da4d17f85aacea80976f9621a89ee6c_41138_427c7dec61ba15f974deeae5318a6526


bep:
>and feature-rich

If the use case is to build "low tech" looking use case,

* Why do we need "feature-rich" dithering? 
* Why isn't the FloydSteinberg drawer in the draw package we already use good enough to create this "low tech" look?

This seem like hitting the nail with a hammer to solve a, I suspect, niche use case (I haven't seen many people asking for this).

jmooring:
The examples that I have seen referenced by those who have expressed interest have been dithered using ordered patterns (e.g. cluster dot) not error diffusion (e.g, FloydSteinberg). This includes both.

To achieve the desired effect, you may need to use different ordered patterns depending on image size. For example, an 8x8 pattern on a 480x270 image looks terrible, but looks pretty good at 1920x1080.

The examples I have seen have varying palettes, so passing in a custom palette is desirable. Although not required by the examples that I have seen, creating a palette from the dominant colors returned by `.Colors` is pretty slick.

Finally, what triggered this PR was an article where someone created an external image processor/server to dither, and then get those images with `resources.GetRemote`. Compared to the image filter API that you created, that approach is just ugly. To me this is a perfect fit for our library of filters... create a slice, go left to right, you're done. It's remarkably elegant.

Yes, this is a hammer. But it's a really nice hammer.

bep:
>Finally, what triggered this PR was an article where someone created an external image processor/server to dither, 

Do you have a link to that article?

bep:
OK; I tested it now, and I'll admit, it is pretty cool

jmooring:
> https://davideisinger.com/journal/encrypt-and-dither-photos-in-hugo/

bep (changes requested):

resources/images/dither.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/makeworld-the-better-one/dither/v2"
+)
+
+var _ gift.Filter = (*ditherFilter)(nil)
+
+type ditherFilter struct {
+	ditherer *dither.Ditherer
+}
+
+var methodsErrorDiffusion = map[string]dither.ErrorDiffusionMatrix{
```
bep, line 30: This is nit-picking, but since these are package variables, it would be good if they had more specific names, e.g. `ditherMethodsErrorDiffusion`. The same applies to the one below.

Other than that, this looks good.

resources/images/filters.go
```diff
@@ -174,6 +176,68 @@ func (*Filters) Padding(args ...any) gift.Filter {
 	}
 }
 
+// Dither creates a filter that dithers an image.
+func (*Filters) Dither(options ...any) gift.Filter {
+	type ditherOptions struct {
```
bep, line 181: An anonomous struct (ditherOptions := struct{}) would probably be a better choice here.

jmooring:
I also adjusted where case folding occurs to improve one of the error messages, and made some minor changes to the documentation page.

resources/images/filters.go
```diff
@@ -174,6 +176,66 @@ func (*Filters) Padding(args ...any) gift.Filter {
...
+	}{
+		colors:     []string{"000000ff", "ffffffff"},
+		method:     "floydsteinberg",
+		serpentine: true,
+		strength:   1.0,
+	}
+
+	var opt maps.Params
+	if len(options) > 0 {
+		opt = maps.MustToParamsAndPrepare(options[0])
```
bep, line 195: I was going to hit the merge button, but I'm going to request one more improvement.

In most similar places, we use `mapstructure.WeakDecode` to read from a map into a struct. That works great in most simple cases (and is what we use to read the main config).

I suspect you need to make the struct fields exported (first upper char) to make this work.

This is nit-picking, but I have spent a lot of time cleaning up old code trying, when possible, to do it one way, me may as well continue that "new year's resolution".

Note that I don't mind the current `switch` approach in situations where mapstructure doesn't work/is too cumbersome, but I don't think this is the case.

resources/images/filters.go
```diff
@@ -174,6 +177,54 @@ func (*Filters) Padding(args ...any) gift.Filter {
...
+		Serpentine bool
+		Strength   float32
+	}{
+		Colors:     []string{"000000ff", "ffffffff"},
+		Method:     "floydsteinberg",
+		Serpentine: true,
+		Strength:   1.0,
+	}
+
+	err := mapstructure.WeakDecode(options[0], &ditherOptions)
```
bep, line 194: Add a check for empty options slice and we should be all good.

bep: I guess empty options should give the defaults.

jmooring: I tested for an empty map, but not for zero args. Thanks.

PR #12034, opened 2024-02-13
Fix typos

coliff:
and fix a few URLs

### DIFF
--- a/hugolib/shortcode.go
+++ b/hugolib/shortcode.go
@@ -206,7 +206,7 @@ type shortcode struct {
 	// Before Hug0 0.55 we didn't send any shortcode output to the markup
 	// renderer, and this flag told Hugo to process the {{ .Inner }} content
 	// separately.
-	// The old behaviour can be had by starting your shortcode template with:
+	// The old behavior can be had by starting your shortcode template with:
 	//    {{ $_hugo_config := `{ "version": 1 }`}}
 	doMarkup bool
 
@@ -422,7 +422,7 @@ func doRenderShortcode(
 			}
 		}
 
-		// Pre Hugo 0.55 this was the behaviour even for the outer-most
+		// Pre Hugo 0.55 this was the behavior even for the outer-most
 		// shortcode.
 		if sc.doMarkup && (level > 0 || sc.configVersion() == 1) {
 			var err error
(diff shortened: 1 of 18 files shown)


bep:
All of these looks good, but could you resolve the merge conflict (as in: rebase against master)? I have previously tested the "Resolve conflicts" button with limited success.

coliff:
Ok good to go now

hugolib/shortcode.go
```diff
@@ -206,7 +206,7 @@ type shortcode struct {
 	// Before Hug0 0.55 we didn't send any shortcode output to the markup
 	// renderer, and this flag told Hugo to process the {{ .Inner }} content
 	// separately.
-	// The old behaviour can be had by starting your shortcode template with:
+	// The old behavior can be had by starting your shortcode template with:
```
bep, line 209: I know I have merged a similar "typo fix" before, but as these comments are mostly written by me, and I write British English, I don't think we can call the above a "typo". It's listed first in the Oxford dictionary (even as the first option):

https://www.oed.com/search/dictionary/?scope=Entries&q=behaviour

hugolib/shortcode.go
```diff
@@ -422,7 +422,7 @@ func doRenderShortcode(
 			}
 		}
 
-		// Pre Hugo 0.55 this was the behaviour even for the outer-most
+		// Pre Hugo 0.55 this was the behavior even for the outer-most
 		// shortcode.
```
bep, line 426: See above.

resources/transform_test.go
```diff
@@ -185,7 +185,7 @@ func TestTransform(t *testing.T) {
 
 		spec := newTestResourceSpec(specDescriptor{c: c})
 
-		// Two transformations with same id, different behaviour.
+		// Two transformations with same id, different behavior.
```
bep, line 188: See above.

bep (changes requested):

hugolib/shortcode.go
coliff: We could revert that if you prefer, but most English technical documentation is written as American English. I thought it’d be best for it to be consistent.

bep: I think you would find that most English technical documentation is written in the language of the people who created it. But OK, I guess one could argue that _some_ of the comments in this repo is also used in the HugoDocs, which we have said should be in Simple English.

PR #12035, opened 2024-02-14
ci: Dependabot check for GitHub workflow actions

deining:
This PR follows up on #12004: it automates dependency checking for GitHub actions to dependabot and delegates the automatable work to this bot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,10 @@
 # See https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem
 version: 2
 updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


bep:
This is not something we're adding. We're pinning our actions to explicit and vetted commits (not versions), which isn't something this tool can doo.

deining:
> We're pinning our actions to explicit and vetted commits (not versions),

I understand this is for security reasons and I deem this a good practice.

> which isn't something this tool can do.

It actually can. For example have a look at this [PR](https://github.com/remkop/picocli/pull/2120/commits/66ece2771c10def40802f2287fde058c407b15aa) authored by dependabot: versions are addressed by commit hash, not by versions.
As stated [here](https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/), dependabot even updates comments in GitHub Actions workflows referencing action versions.

PR #12056, opened 2024-02-17
Fix handling of draft term pages

### LINKED ISSUE #12055: Draft term pages not handled correctly
hugo v0.123.0-DEV-9cc65757a

This test fails:

```go
func TestFoo(t *testing.T) {
	t.Parallel()

	files := `
-- layouts/_default/list.html --
|{{ .Title }}|
-- content/p1.md --
---
title: p1
tags: [a]
---
-- content/tags/a/_index.md --
---
title: tag-a-title-override
draft: true
---
  `

	b := hugolib.Test(t, files)

	b.AssertFileContent("public/tags/a/index.html", "|a|")
}
```

Want:

```
|a|
```

Got:

```text
|tag-a-title-override|
```

bep:
By just removing the term page and all of its page entries.

Which is a little of a coin flip, we could possibley also do like we handle the taxonomy nodes, e.g. disable them.

The reason this didn't just work is the way we need to wait until the end of the chain before we can apply the fully cascaded metadata to these nodes (e.g. the draft flag).

Fixes #12055

### DIFF
--- a/hugolib/content_map_page.go
+++ b/hugolib/content_map_page.go
@@ -1415,8 +1415,10 @@ func (sa *sitePagesAssembler) applyAggregatesToTaxonomiesAndTerms() error {
 					if err := p.setMetaPost(cascade); err != nil {
 						return false, err
 					}
-
-					if err := sa.pageMap.treeTaxonomyEntries.WalkPrefix(
+					if !p.s.shouldBuild(p) {
+						sa.pageMap.treePages.Delete(s)
+						sa.pageMap.treeTaxonomyEntries.DeletePrefix(paths.AddTrailingSlash(s))
+					} else if err := sa.pageMap.treeTaxonomyEntries.WalkPrefix(
 						doctree.LockTypeRead,
 						paths.AddTrailingSlash(s),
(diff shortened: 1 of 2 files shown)


bep:
@jmooring what do you think about this?

jmooring:
@bep I think this is OK. The following now do the same thing:

```text
+++
title = 'My Tag A'
[build]
render = 'never'
list = 'never'
+++
```

```text
+++
title = 'My Tag A'
draft = true
+++
```

We don't render the page, and the term is not a member of the taxonomy.

jmooring:
This PR fixed #12055, but I need to correct my previous statement. Assuming "tag-a" is assigned to one or more content pages, these are **not** equivalent...

content/tags/tag-a/_index.md

```text
+++
title = 'My Tag A'
draft = true
+++
```

```text
+++
title = 'My Tag A'
[build]
render = 'never'
list = 'never'
+++
```

With the former, the term is excluded from the "tags" taxonomies object. This is expected/desired.

With the latter, the term is still included in the "tags" taxonomies object. This is a separate issue.

PR #12109, opened 2024-02-22
fix(commands/mod): ignore invalid module path

razonyang:
Related topic: https://discourse.gohugo.io/t/theme-error-with-hugo-mod-u/48410


jmooring:
<https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines>. Something like:

```
commands/mod: Ignore invalid module path
```

PR #12179, opened 2024-02-29
Add hugo.IsMultiHost

razonyang:
It would(may) be helpful for generating resources those will be shared between multihosts, take search index as an example.

```go
{{ $pages := slice }}
{{ range site.Sites.RegularPages }}
  {{ $pages = $pages | append (dict
    "title" .Title
    "url" (cond hugo.IsMultiHost .Permalink .RelPermalink))
  }}
{{ end }}
```


common/hugo/hugo.go
```diff
@@ -111,11 +111,17 @@ func (i HugoInfo) Deps() []*Dependency {
 	return i.deps
 }
 
+// IsMultiHost reports whether is a multihost setup.
```
jmooring, line 114:
```suggestion
// IsMultiHost reports whether each configured language has a unique baseURL.
```

PR #12243, opened 2024-03-14
add pubcrank.com editor

pizzapanther:
I've built an editor for static site generators like Hugo and added it in the docs.

https://www.pubcrank.com/


jmooring:
Please submit this to the [documentation repository](https://github.com/gohugoio/hugoDocs).

PR #12281, opened 2024-03-19
Update sitemap.xml to allow content to be excluded

kolappannathan:
With only one line change as mentioned in https://github.com/gohugoio/hugo/pull/6370#issuecomment-2007667847

**Related**:
- https://github.com/gohugoio/hugo/issues/653
- https://github.com/gohugoio/hugo/pull/6370

PR #12282, opened 2024-03-19
tpl/tplimpl: Allow content exclusion in sitemap.xml  Closes #653

### LINKED ISSUE #653: Means to exclude content from sitemap.xml

kolappannathan:
With only one line change as mentioned in https://github.com/gohugoio/hugo/pull/6370#issuecomment-2007667847

**Related**:
- https://github.com/gohugoio/hugo/issues/653
- https://github.com/gohugoio/hugo/pull/6370


jmooring:
The commit message is still incorrect. See <https://github.com/gohugoio/hugo/pull/12281#issuecomment-2007808649>

It should be on three lines. Make the change to the commit  message (git commit --amend) and force push (git push -f).

kolappannathan:
> The commit message is still incorrect. See [#12281 (comment)](https://github.com/gohugoio/hugo/pull/12281#issuecomment-2007808649)
> 
> It should be on three lines. Make the change to the commit message (git commit --amend) and force push (git push -f).

Updated commit message.

jmooring:
bep, please merge. This allows sitemap exclusion on a page-by-page basis:

```text
[sitemap]
exclude = true
changeFreq = "hourly"
priority = 0.7
```

This obviously fits nicely with the existing front matter values, and the change is not invasive (limited to the embedded template).

XhmikosR:
May I suggest this so that the output doesn't have extra whitespace?

```xml
{{ printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml">
  {{- range where .Data.Pages "Params.sitemap.exclude" "ne" true -}}{{ if .Permalink }}
  <url>
    <loc>{{ .Permalink }}</loc>{{ if not .Lastmod.IsZero }}
    <lastmod>{{ safeHTML (.Lastmod.Format "2006-01-02T15:04:05-07:00") }}</lastmod>{{ end }}{{ with .Sitemap.ChangeFreq }}
    <changefreq>{{ . }}</changefreq>{{ end }}{{ if ge .Sitemap.Priority 0.0 }}
    <priority>{{ .Sitemap.Priority }}</priority>{{ end }}{{ if .IsTranslated }}{{ range .Translations }}
    <xhtml:link rel="alternate" hreflang="{{ .Language.LanguageCode }}" href="{{ .Permalink }}"/>{{ end }}
    <xhtml:link rel="alternate" hreflang="{{ .Language.LanguageCode }}" href="{{ .Permalink }}"/>{{ end }}
  </url>{{ end }}{{ end }}
</urlset>

```

kolappannathan:
@XhmikosR May be we can have the white space removal as a separate PR? I want to keep the PR minimal in changes so that it is merged faster.

XhmikosR:
Not my choice to make, but it's a cosmetic change anyway. Feel free to submit it separately, I just thought now it'd be a good time

tpl/tplimpl/embedded/templates/_default/sitemap.xml
```diff
@@ -1,7 +1,7 @@
 {{ printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
   xmlns:xhtml="http://www.w3.org/1999/xhtml">
-  {{ range .Data.Pages }}
+  {{ range where .Data.Pages "Params.sitemap.exclude" "ne" true }}
```
bep, line 4: * Add `Exclude` to the `SitemapConfig` struct.
* Change the above where clause to use `.Sitemap.Exclude`
* We also need a test for this (I guess `tplimpl_integration_test.go` would be a OK placement)

bep (changes requested):

jmooring:
OK, we're back to https://github.com/gohugoio/hugo/pull/6370.

tpl/tplimpl/embedded/templates/_default/sitemap.xml
bep: Make that `.Sitemap.Disable` -- which is in line with what we have elsewhere.

PR #12298, opened 2024-03-22
hugolib: Conditionally suppress .Site.Author deprecation notice

### LINKED ISSUE #12297: .Site.Author deprecation is logged when .Site.Author not called in template
Reference: <https://discourse.gohugo.io/t/info-deprecated-site-author/48978/>

jmooring:
Suppress the .Site.Author deprecation notice unless the Author key is present and not empty in the site configuration.

Closes #12297


jmooring:
This may not be the most elegant solution to the underlying problem, but I think it's the simplest.

PR #12311, opened 2024-03-25
resources: Use different cache key when copying resources

### LINKED ISSUE #10412: resources.Copy to original resource path points to wrong resource
```text
assets/
└── a.jpg    # image is 600x400
```

#### This works as expected...

```text
{{ with resources.Get "a.jpg" }}
  {{ with .Resize "200x" | resources.Copy "foo.jpg" }}
    <img src="{{ .RelPermalink }}" width="{{ .Width }}" height="{{ .Height }}">
  {{ end }}
{{ end }}
```

```html
<img src="/foo.jpg" width="200" height="133">
```

#### But if you try to copy the resource to the original path (a.jpg)...

```text
{{ with resources.Get "a.jpg" }}
  {{ with .Resize "200x" | resources.Copy "a.jpg" }}
    <img src="{{ .RelPermalink }}" width="{{ .Width }}" height="{{ .Height }}">
  {{ end }}
{{ end }}
```

```html
<img src="/a.jpg" width="600" height="400">
```

Note that the `width` and `height` attributes are from the original image, not the resized image.

Why would you want to copy a resource to its original path? To have something other than this:

```html
<img src="/a_hu7ea6bbdd4d289c929adf1f5f39c46d1e_158382_200x0_resize_q75_box.jpg" width="200" height="133">
```

### LINKED ISSUE #12310: Improve resources.Copy to prevent nil pointer when using unusual construct
Reference: <https://discourse.gohugo.io/t/nil-pointer-evaluating-resource-resource-relpermalink-when-using-resources-copy/49030/14>

Error:

> Error: error building site: render: failed to render pages: render of "page" failed: "/home/jmooring/code/hugo-testing/layouts/_default/single.html:4:5": execute of template failed: template: _default/single.html:4:5: executing "_default/single.html" at <$r.RelPermalink>: nil pointer evaluating resource.Resource.RelPermalink

Notes:

1. I have no idea if this a bug or an enhancement. 
2. If it's a bug, it's been there forever; this isn't anything new.
3. The construct below is unusual. The example reported in the forum topic makes sense, while the test case below is simply a minimal reproducer.
4. The OP has at least two ways to get around this.

Minimal reproducible example:

```text
git clone --single-branch -b hugo-forum-topic-49030 https://github.com/jmooring/hugo-testing hugo-forum-topic-49030
cd hugo-forum-topic-49030
hugo
```

Test case:

```go
func TestFoo(t *testing.T) {
	t.Parallel()

	files := `
-- hugo.toml --
disableKinds = ['home','section','rss','sitemap','taxonomy','term']
-- assets/a.txt --
This was assets/a.txt
-- content/p1.md --
---
title: p1
---
{{< sc >}}
-- layouts/shortcodes/sc.html --
{{ $noop := resources.Get "/p1/b.txt" }}
{{ $noop := .Content }}
{{ $r := resources.Get "a.txt" }}
{{ $r = resources.Copy "/p1/b.txt" $r }}
{{ $r.RelPermalink }}
{{/* In the above, no errors you remove line 1. */}}
`

	b.Assert(err, qt.IsNil)
	b.AssertFileContent("public/p1/b.txt", "This was assets/a.txt")
}

```

In the above, `resources.Get "/p1/b.txt"` is not a typographical error. Again, the example reported in the forum topic makes sense; the above is a minimal reproducer.

My best guess is that we cache a nil resource when calling the shortcode, and resources.Copy doesn't overwrite it because... it's cached.

jmooring:
Closes #10412
Closes #12310

### DIFF
--- a/resources/resource_factories/create/create.go
+++ b/resources/resource_factories/create/create.go
@@ -57,7 +57,7 @@ func New(rs *resources.Spec) *Client {
 
 // Copy copies r to the new targetPath.
 func (c *Client) Copy(r resource.Resource, targetPath string) (resource.Resource, error) {
-	key := dynacache.CleanKey(targetPath)
+	key := dynacache.CleanKey(targetPath) + "__copy"
 	return c.rs.ResourceCache.GetOrCreate(key, func() (resource.Resource, error) {
 		return resources.Copy(r, targetPath), nil
 	})
@@ -66,7 +66,7 @@ func (c *Client) Copy(r resource.Resource, targetPath string) (resource.Resource
 // Get creates a new Resource by opening the given pathname in the assets filesystem.
 func (c *Client) Get(pathname string) (resource.Resource, error) {
 	pathname = path.Clean(pathname)
-	key := dynacache.CleanKey(pathname)
+	key := dynacache.CleanKey(pathname) + "__get"
 
 	return c.rs.ResourceCache.GetOrCreate(key, func() (resource.Resource, error) {
 		// The resource file will not be read before it gets used (e.g. in .Content),
--- a/resources/resources_integration_test.go
+++ b/resources/resources_integration_test.go
@@ -227,3 +227,50 @@ eventDate: 2023-11-01T07:00:00-08:00
 
 	b.AssertFileContent("public/index.html", "2023-11|p9|p8|p7|2023-10|p6|p5|p4|2023-09|p3|p2|p1|")
 }
+
+// Issue 10412
+func TestImageTransformThenCopy(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+disableKinds = ['page','rss','section','sitemap','taxonomy','term']
+-- assets/pixel.png --
+iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAAAABJRU5ErkJggg==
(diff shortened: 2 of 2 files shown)


resources/resource_cache.go
```diff
@@ -68,6 +68,10 @@ func (c *ResourceCache) Get(ctx context.Context, key string) (resource.Resource,
 }
 
 func (c *ResourceCache) GetOrCreate(key string, f func() (resource.Resource, error)) (resource.Resource, error) {
+	// Do nothing if resource is nil.
+	if r, _ := f(); r == nil {
```
bep, line 72: The main motivation behind this cache is hav `f` being called once, which this breaks.

I suggest that we instead fix this by adding a prefix to the cache key in `Copy` (and also do the same in `Get` to make it symmetrical).

```go
func (c *Client) Copy(r resource.Resource, targetPath string) (resource.Resource, error) {
	key :=  dynacache.CleanKey(targetPath) + "__copy"
	return c.rs.ResourceCache.GetOrCreate(key, func() (resource.Resource, error) {
		return resources.Copy(r, targetPath), nil
	})
}

```

Or something.

jmooring:
Using a different key for Copy may also address https://github.com/gohugoio/hugo/issues/10412.

resources/resource_cache.go
jmooring: Added suffix to Get and Copy, and added test case for https://github.com/gohugoio/hugo/issues/10412.

PR #12312, opened 2024-03-26
hugolib: Fix regression in summary generation behavior since v0.123

curegit:
Since v0.123, Hugo will use the automatic summary generation when the manual summary results in an empty string, even if there is a `<!--more-->` summary divider.

This PR fixes that regression to allow empty summary, restoring the intended functionality and flexibility provided by the manual summary divider.

Note about implementation in v0.122:
<https://github.com/gohugoio/hugo/blob/v0.122.0/hugolib/page__per_output.go#L791>

### DIFF
--- a/hugolib/page__content.go
+++ b/hugolib/page__content.go
@@ -770,7 +770,7 @@ func (c *cachedContent) contentPlain(ctx context.Context, cp *pageContentOutput)
 			result.readingTime = (result.wordCount + 212) / 213
 		}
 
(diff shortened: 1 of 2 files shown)


bep:
Thanks for this. This looks correct, but could you add a test for this so we don't break it again in the future?

curegit:
@bep Thank you for the reply. I'm going to write a test.

curegit:
@bep

I added the test.
Could you please review it?

PR #12325, opened 2024-04-01
chore: fix function names

stayweek:


resources/images/smartcrop.go
```diff
@@ -81,7 +81,7 @@ func (p *ImageProcessor) smartCrop(img image.Image, width, height int, filter gi
 	return img.Bounds().Intersect(rect), nil
 }
 
-// Calculates scaling factors using old and new image dimensions.
+// calcFactorsNfnt scaling factors using old and new image dimensions.
```
bep, line 84: This doesn't make sense. You should fix the function name, not the comment.

bep (changes requested):

PR #12328, opened 2024-04-01
tpl/tplimpl: Improve embedded opengraph template

### LINKED ISSUE #8296: Opengraph template locale-  fallback to site's locale param
Currently the built-in Opengraph template only adds the `<meta property="og:locale" content="" />` meta tag to a page if the locale param is in the frontmatter of the page. It'd be great if no param was specified in the frontmatter then it used the locale param in the sites config instead. It could make maintenance easier (if you have a 200-page site in a single language you only need to define the locale in the site config).

Maybe something like this?

```go
{{- with .Params.locale }}<meta property="og:locale" content="{{ . }}" />
{{else}}
{{- with .Site.Params.locale }}<meta property="og:locale" content="{{ . }}" />
{{ end }}
```

cc @djatwood 

REF:
https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/opengraph.html

### LINKED ISSUE #8698: Add plainify to Title - in internal template opengraph.html

### LINKED ISSUE #9818: Handle Markdown for .Title and .Description in opengraph properties

jmooring:
Changes:

- Add tags per documentation
- Prefer site.Title over site.Params.title
- Plainify titles, tags, and descriptions
- Add fallback values for locale
- Fix pages related by series
- Improve readability

Closes #8296
Closes #8698
Closes #8991
Closes #9818
Closes #9866
Closes #10647

### DIFF
--- a/tpl/tplimpl/embedded/templates/opengraph.html
+++ b/tpl/tplimpl/embedded/templates/opengraph.html
@@ -1,44 +1,72 @@
-<meta property="og:title" content="{{ .Title }}">
-<meta property="og:description" content="{{ with .Description }}{{ . }}{{ else }}{{if .IsPage}}{{ .Summary }}{{ else }}{{ with .Site.Params.description }}{{ . }}{{ end }}{{ end }}{{ end }}">
-<meta property="og:type" content="{{ if .IsPage }}article{{ else }}website{{ end }}">
 <meta property="og:url" content="{{ .Permalink }}">
 
-{{- $images := partial "_funcs/get-page-images" . -}}
-{{- range first 6 $images -}}
-  <meta property="og:image" content="{{ .Permalink }}">
-{{- end -}}
-
-{{- if .IsPage }}
-{{- $iso8601 := "2006-01-02T15:04:05-07:00" -}}
-<meta property="article:section" content="{{ .Section }}">
-{{ with .PublishDate }}<meta property="article:published_time" {{ .Format $iso8601 | printf "content=%q" | safeHTMLAttr }}>{{ end }}
-{{ with .Lastmod }}<meta property="article:modified_time" {{ .Format $iso8601 | printf "content=%q" | safeHTMLAttr }}>{{ end }}
-{{- end -}}
-
-{{- with .Params.audio }}<meta property="og:audio" content="{{ . }}">{{ end }}
-{{- with .Params.locale }}<meta property="og:locale" content="{{ . }}">{{ end }}
-{{- with .Site.Params.title }}<meta property="og:site_name" content="{{ . }}">{{ end }}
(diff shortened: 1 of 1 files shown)


tpl/tplimpl/embedded/templates/opengraph.html
```diff
@@ -1,44 +1,72 @@
...
+
+{{- with .Params.videos }}
+  {{- range . | first 6 }}
+    <meta property="og:video" content="{{ . | absURL }}">
+  {{- end }}
+{{- end }}
+
+{{- range .GetTerms "series" }}
+  {{- range .Pages | first 7 }}
+    {{- if not ($.Eq .) }}
```
bep, line 56: I would prefer if we use `eq` and similar for equality checks. It's not like `.Eq` is going away, but it was added plainly to make `eq` work with wrapped page types.

bep: My take on this is that

* `eq` can be used without on (almost) any Hugo object/type.
* `.Eq` is only available on some types, so it's less portable and will break if people expect it to work on ... say strings.

PR #12332, opened 2024-04-02
tpl/strings: Create strings.Diff template function

### LINKED ISSUE #12330: Create a strings.Diff template function
No new dependencies; this would call <https://pkg.go.dev/github.com/rogpeppe/go-internal@v1.12.0/diff#Diff>. 

Related: <https://discourse.gohugo.io/t/functional-testing-harness-for-templates-partials/49094>

It's fairly trivial to create a Hugo site to test partial and shortcode output against gold files, and then compare the content to report pass/fail. But it would be handy to report the diff as well. Something like:

```text
{{ $diff := strings.Diff "want" $want "got" $got }}
{{ transform.Highlight $diff "diff" }}
```

The above would look like:

image

jmooring:
Closes #12330

### DIFF
--- a/tpl/strings/strings.go
+++ b/tpl/strings/strings.go
@@ -27,6 +27,7 @@ import (
 	"github.com/gohugoio/hugo/deps"
 	"github.com/gohugoio/hugo/helpers"
 	"github.com/gohugoio/hugo/tpl"
+	"github.com/rogpeppe/go-internal/diff"
 
 	"github.com/spf13/cast"
 )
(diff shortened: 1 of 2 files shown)


tpl/strings/strings.go
```diff
@@ -172,6 +173,15 @@ func (ns *Namespace) ContainsNonSpace(s any) bool {
 	return false
 }
 
+// Diff returns an anchored diff of the two texts old and new in the “unified
+// diff” format. If old and new are identical, Diff returns a nil slice.
```
bep, line 177: The return value is never nil.

PR #12334, opened 2024-04-02
Fix misplaced end statment

learn-as-i-learn:
Looks like an {{ end }} statement got misplaced.


jmooring:
The existing code is fine. Your change would break it.

Issue #2354, opened 2016-08-12
tpl: Handle all errors from cast in the template funcs


bep:

bep:
Note: This is less important where the value is just returned to the user (like in safeHTML etc.), but uniformity is good.

PR #12335, opened 2024-04-03
errors: Return error from cast.ToStringE() consistently

seiyab:
I guess some errors are mistakenly discarded in https://github.com/gohugoio/hugo/issues/2354.

I didn't add any tests because I cannot find tests even for similar methods so maybe you consider such tests are too much. Ask me write ones if you consider they are necessary. To write test, I need your instructions, say, where to write test, how big (integrated) should they be, how much should tests cover and so on.

### DIFF
--- a/tpl/lang/lang.go
+++ b/tpl/lang/lang.go
@@ -58,7 +58,7 @@ func (ns *Namespace) Translate(ctx context.Context, id any, args ...any) (string
 
 	sid, err := cast.ToStringE(id)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
 
 	return ns.deps.Translate(ctx, sid, templateData), nil
@@ -165,7 +165,7 @@ func (ns *Namespace) FormatNumberCustom(precision, number any, options ...any) (
 			// custom delimiter
 			s, err := cast.ToStringE(options[1])
 			if err != nil {
-				return "", nil
+				return "", err
 			}
 
 			delim = s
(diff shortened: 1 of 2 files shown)


jmooring:
I think this PR is OK. Returning `nil` instead of `err` seems like an oversight. For example, `AbsLangURL` and `RelLangURL` return the `err`, but their counterparts do not... so I don't think this was by-design. And the code hasn't been touched in 7 years.

I checked the tpl/strings package, and we are mostly consistent with cast error checking. There are a couple of cases where we are not... I will submit a separate PR for those.

seiyab:
Thank you for your review and merging.

> I checked the tpl/strings package, and we are mostly consistent with cast error checking. There are a couple of cases where we are not... I will submit a separate PR for those.

Oh, I haven't been aware of them. Thank you for checking.

To be honest, I know similar suspicious codes more. I didn't include them in this PR because I'm not sure they are bug yet. I list up similar ones here anyway.
https://github.com/gohugoio/hugo/blob/02d5ec14f3fea20fa05aa2f0c2b9698620633edc/hugolib/page__ref.go#L50-L53
https://github.com/gohugoio/hugo/blob/02d5ec14f3fea20fa05aa2f0c2b9698620633edc/tpl/collections/collections.go#L301-L309
https://github.com/gohugoio/hugo/blob/master/modules/npm/package_builder.go#L90-L93
https://github.com/gohugoio/hugo/blob/02d5ec14f3fea20fa05aa2f0c2b9698620633edc/tpl/os/os.go#L64-L67
https://github.com/gohugoio/hugo/blob/02d5ec14f3fea20fa05aa2f0c2b9698620633edc/helpers/general.go#L312-L314

Ask me to open an issue if necessary. I'm sorry but I'm not eager to investigate and fix them by myself.

Issue #8232, opened 2021-02-13
Hash sign (#) in taxonomy terms not properly sanitized


michaelroland:
### What version of Hugo are you using (`hugo version`)?

<pre>
Hugo Static Site Generator v0.80.0/extended windows/amd64 BuildDate: unknown
</pre>
<pre>
Hugo Static Site Generator v0.80.0-792EF0F4/extended linux/amd64 BuildDate: 2020-12-31T13:46:18Z
</pre>

### Does this issue reproduce with the latest release?

Yes, tested with the latest (0.80.0) version on Windows and Linux.

### Issue

When a taxonomy term contains one or more `#` (hash sign) characters, the first hash sign character is not sanitized (escaped to `%23`) in URLs generated for the term page. Consequently, the browser will treat the path segement following the hash sign as fragment identifier and not as part of the path. If a taxonomy contains more than one hash sign, only the first one is not sanitized, all subsequent hash signs are properly sanitized to `%23`.

### Example

- `testtaxonomy = [ "Testterm#" ]`
  - generates a directory named "testterm#"
  - URL (.RelPermalink) becomes "/testtaxonomy/testterm#/"
- `testtaxonomy = [ "Test#term#" ]`
  - generates a directory named "test#term#"
  - URL (.RelPermalink) becomes "/testtaxonomy/test#term%23/"
- `testtaxonomy = [ "Test#te#rm#" ]`
  - generates a directory named "test#te#rm#"
  - URL (.RelPermalink) becomes "/testtaxonomy/test#te%23rm%23/"

### Expected behavior

All occurrences of `#` should be escaped in the resulting URLs or, better yet, `#` characters should be removed from the generates filenames at all. Considering that other potentially problematic characters (such as `!?§$%&*=,:;|(){}[]<>"`) are already removed from the generated filenames, I'm rather surprised that `#` is not. I therefore believe that this would be the more logical solution. In my opinion, the same applies to `.` (dot), `+` (plus), `~` (tilde), `_` (underscore), [and potentially any other non-word character that I missed] which are currently kept as-is. This would have the benefit of also mitigating issues like #8153.

david-bakin:
See also related #4926.  It would be sweet if I could have a tag `C#`...

michaelroland:
This issue is still observable in the lastest release. Tested with

```
[...]

PR #12347, opened 2024-04-05
Strip hash sign (#) from file paths/URLs

### LINKED ISSUE #4926: `#` not escaped in Permalink when in title value

bep:
The general way Hugo does this now is:

* Sanitize the file paths so the work as URLs
* When we create the final RelPermalink/Permalink, we use Go's `url.Parse` to escape it so it work for the browser.

So, leaving anything in the first step that does not work with the second step, just doesn't work.

It's a little bit odd that `url.Parse` silently truncates this URL without any error, but that's for another day.

I have another better test coverage for this.

Fixes #12342
Fixes #4926
See #8232

### DIFF
--- a/common/paths/path.go
+++ b/common/paths/path.go
@@ -318,7 +318,7 @@ func isAllowedPathCharacter(s string, i int, r rune) bool {
 	}
 	// Check for the most likely first (faster).
 	isAllowed := unicode.IsLetter(r) || unicode.IsDigit(r)
-	isAllowed = isAllowed || r == '.' || r == '/' || r == '\\' || r == '_' || r == '#' || r == '+' || r == '~' || r == '-' || r == '@'
+	isAllowed = isAllowed || r == '.' || r == '/' || r == '\\' || r == '_' || r == '+' || r == '~' || r == '-' || r == '@'
 	isAllowed = isAllowed || unicode.IsMark(r)
 	isAllowed = isAllowed || (r == '%' && i+2 < len(s) && ishex(s[i+1]) && ishex(s[i+2]))
 	return isAllowed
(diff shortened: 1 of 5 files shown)


jmooring:
The first time I ran across this issue was when someone tagged their content with the programming language name c#. If we eat the hash signs, this causes page collisions:

```
---
title: p1
tags: [c, c#, c++]
---
````

c++ works fine, but the other two collide.

preechapon-cmd (approved):

PR #12353, opened 2024-04-07
completion: Improve existing argument completions, add many more

scop:
Do not offer filenames to arguments not taking one, complete arguments of options taking resource kinds, directory names, --logLevel, release --step, config and new --format.

Set argument counts and types on Cobra command level. As part of this, remove internal checks on argument counts. They are redundant with cobra.Args settings, and although they would give a chance to output a more specific error message, using cobra.Args makes completions suggest the correct number of positional arg completions.

As an internal refactoring, use higher level functions to set flag completions.  SetAnnotation works, but is more verbose than alternatives, and uses bash specific wording.

While at it, move setting completions next to flag definitions consistently.

Remove superfluous --destination completer setting, which is already set elsewhere.


commands/new.go
```diff
@@ -60,8 +61,16 @@ Ensure you run this within the root directory of your site.`,
...
 				},
 				withc: func(cmd *cobra.Command, r *rootCommand) {
+					cmd.ValidArgsFunction = func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+						if len(args) != 0 {
+							return []string{}, cobra.ShellCompDirectiveNoFileComp
+						}
+						return []string{}, cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveFilterDirs
+					}
 					cmd.Flags().StringVarP(&contentType, "kind", "k", "", "content type to create")
+					_ = cmd.RegisterFlagCompletionFunc("kind", cobra.FixedCompletions(kinds.AllKinds, cobra.ShellCompDirectiveNoFileComp))
```
bep, line 71: I have not looked at all usage of AllKinds in here, but I would say that for completion, I think the current best approach would be to:

* Use AllKindsInPages in the above case and similar.
* Use AllKinds for the `disableKinds`.

scop: Cheers, changed this one to `AllKindsInPages`. `AllKinds` was already in use in `disableKinds`.

PR #12363, opened 2024-04-11
resources/page: Add taxonomies Page method

### LINKED ISSUE #12316: Add Page method/field to taxonomy object
We can get a reference to a _term_ page with:

```text
{{ $termPage := site.Taxonomies.tags.foo.Page }}
```

It would be convenient if we could get a reference to a _taxonomy_ page with:

```text
{{ $taxonomyPage := site.Taxonomies.tags.Page }}
```

That may not be possible, despite the fact we lowercase terms internally (e.g., there shouldn't be a collision between `tags.page` (a term) and `tags.Page` (a method/field).

The alternative has always seemed less elegant compared to how you can get a reference to a term page:

```text
{{ site.GetPage "/tags" }}
```

jmooring:
Closes #12316

### DIFF
--- a/resources/page/taxonomy.go
+++ b/resources/page/taxonomy.go
@@ -112,6 +112,14 @@ func (i Taxonomy) ByCount() OrderedTaxonomy {
 	return ia
 }
 
+// Page returns the taxonomy page or nil if the taxonomy has no terms.
+func (i Taxonomy) Page() Page {
+	for _, v := range i {
+		return v.Page().Parent()
+	}
+	return nil
+}
+
 // Pages returns the Pages for this taxonomy.
 func (ie OrderedTaxonomyEntry) Pages() Pages {
 	return ie.WeightedPages.Pages()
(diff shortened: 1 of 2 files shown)


resources/page/taxonomy.go
```diff
@@ -112,6 +112,19 @@ func (i Taxonomy) ByCount() OrderedTaxonomy {
...
+func (i Taxonomy) Page() Page {
+	var first WeightedPages
+	for _, v := range i {
+		first = v
+		break
+	}
+	if first == nil {
+		return nil
+	}
+	return first.Page().Parent()
```
bep, line 125: I know I was the one who suggested the above construct, but looking at it now, I suspect it would be simpler to reason about if it looked like:

```go
func (i Taxonomy) Page() Page {
  for _, v := range i {
      return v.Page().Parent()
  }
  return nil
}
```

jmooring: I should have seen that. Thanks.

PR #12404, opened 2024-04-21
improve: use errors.New to replace fmt.Errorf with no parameters will much better

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better


bep:
Two of these edits are done in a forked version of Go's template packages, and we don't edit those unless needed. Also, I'm not convinced this is an improvement, but thanks.

PR #12405, opened 2024-04-21
SECURITY.md: Update link to security model

ejona86:
The security model was moved in https://github.com/gohugoio/hugoDocs/pull/2495


jmooring:
I've added an alias (redirect) on the docs site.
<https://gohugo.io/about/security-model/>

ejona86:
Oh, good. I just clicked the link again and saw it worked. "Did I do it wrong before?" I'll leave this open since you might want to avoid the redirect, but I'm fine with this being closed.

Issue #10900, opened 2023-04-17
Code readability is not strong, high barriers to entry for newcomers


ZhangChengLin:
https://github.com/gohugoio/hugo/blob/9906c1ae52e44f2e8ed45873ea36cd83a9e9bcc0/tpl/tplimpl/embedded/templates/_default/rss.xml#L29-L36

Whether the code can increase the readability, this will not affect the later compression, it should be updated and maintained.

The official documentation feels outdated, so you can only look at the source code. Sometimes the source code is not readable, but Hugo is really easy to use. When can you consider some basic issues.

It's just a personal feeling, if something is wrong, please forgive me, I hope Hugo will get better and better.

bwc69151:
I disagree that newcomers are looking at code readability but rather at more examples or beginner-friendly language in the docs.

jmooring:
Many (all?) of the embedded templates were created before the release of Go 1.6, which introduced syntax to remove white space&nbsp;(`{{- -}}`). Tracking progress on this issue...

#### Partials

- [ ] [tpl/tplimpl/embedded/templates/_partials/disqus.html](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_partials/disqus.html)
- [x] tpl/tplimpl/embedded/templates/_partials/twitter_cards.html
- [x] tpl/tplimpl/embedded/templates/_partials/_funcs/get-page-images.html
- [x] tpl/tplimpl/embedded/templates/_partials/alias.html
- [x] tpl/tplimpl/embedded/templates/_partials/google_analytics.html
- [x] tpl/tplimpl/embedded/templates/_partials/opengraph.html
- [x] tpl/tplimpl/embedded/templates/_partials/pagination.html
- [x] tpl/tplimpl/embedded/templates/_partials/schema.html

#### Shortcodes

- [ ] [tpl/tplimpl/embedded/templates/_shortcodes/highlight.html](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_shortcodes/highlight.html)
- [ ] [tpl/tplimpl/embedded/templates/_shortcodes/param.html](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_shortcodes/param.html)
- [x] tpl/tplimpl/embedded/templates/_shortcodes/comment.html
- [x] tpl/tplimpl/embedded/templates/_shortcodes/details.html
[...]

Issue #8910, opened 2021-08-20
Manual content summaries insert with `<p>` tag


jstamant:
While designing my own theme, I came across a curious issue:

When using the `{{ .Summary }}` page variable to manually generate a summary in combination with the `<!--more-->` tag in a post, Hugo inserts the summary wrapped with a `<p>` tag.

Here's what I'm expecting:

When using the `{{ .Summary }}` page variable to automatically generate summaries, Hugo simply inserts the summary text in the template - not wrapped with any HTML tags.

<pre>
$ hugo version
hugo v0.87.0+extended linux/amd64 BuildDate=unknown
</pre>

I am using the latest release. So, yes, this issue is reproducible in the latest release.

#### `summary.html` template file
```handlebars
<p class="excerpt">
  {{ .Summary }}
</p>
{{- if .Truncated }}
<p>
  <a href="{{ .RelPermalink }}">Read more...</a>
</p>
{{- end }}
```

### Reproducible erroneous behavior

See this example with manual summary generation:

#### `test.md`
```markdown
---
title: "Test"
date: 2021-08-08T13:33:29-05:00
draft: false
---

Lorem ipsum dolor sit amet.<!--more-->
```
#### Result
```html
<p class="excerpt">
  <p>Lorem ipsum dolor sit amet.</p>
</p>
<p>
  <a href="/posts/test/">Read more...</a>
</p>
```

### Expected behavior

Now compare this to what I'm expecting:

#### `test.md`
```markdown
---
title: "Test"
date: 2021-08-08T13:33:29-05:00
draft: false
---

Lorem ipsum dolor sit amet.
```
#### Result
```html
<p class="excerpt">
  <p>Lorem ipsum dolor sit amet.</p>
</p>
<p>
  <a href="/posts/test/">Read more...</a>
</p>
```

jmooring:
The built-in mechanisms for summary assignment and rendering have [created confusion](https://discourse.gohugo.io/search?q=summary) for several years.

<https://gohugo.io/content-management/summaries>

Method|Description|Transformation|Single Lines Wrapped with `<p>`
:--|:--|:--|:--
A|Automatic Summary|Stripped to plain text|false
B|Manual Summary|Markdown processor|true
C|Frontmatter Summary|Markdown processor|false

[...]

PR #12425, opened 2024-04-24
tpl/tplimpl: Fix double-escaping in opengraph template

### LINKED ISSUE #12418: Twitter `og:description` doubly escaped by Hugo 0.125.x
If you build this [small repro (test-site-2024-04-23.zip)](https://drive.google.com/file/d/1VIOiEZ8tdyI2CR1pTkoT9dqw_r5kdBdu) using Hugo 0.124.1 and 0.125.2, you'll see the following difference in the `content` of `<meta property="og:description">` for http://localhost:1313/docs/page:

```html
0.124.1: <meta property="og:description" content="Test line with a single quote: can&rsquo;t." />
0.125.2: <meta property="og:description" content="Test line with a single quote: can&amp;rsquo;t.">
```

Notice that 0.125.2 doubly escapes the `content`, so that:

- `can&rsquo;t` becomes
- `can&amp;rsquo;t`

This seems wrong to me. WDYT?

---

Below I include a bit more header context.

Hugo 0.124.1:

```html
<meta name="description" content="Test line with a single quote: can&amp;rsquo;t.">
<meta property="og:title" content="Some page" />
<meta property="og:description" content="Test line with a single quote: can&rsquo;t." />
...
<meta name="twitter:description" content="Test line with a single quote: can&rsquo;t."/>
```

Hugo 0.125.2:

```html
<meta name="description" content="Test line with a single quote: can&amp;rsquo;t.">
<meta property="og:url" content="http://localhost:1313/docs/page/">
  <meta property="og:site_name" content="My New Hugo Site">
  <meta property="og:title" content="Some page">
  <meta property="og:description" content="Test line with a single quote: can&amp;rsquo;t.">
  ...
<meta name="twitter:description" content="Test line with a single quote: can&rsquo;t.">
```

jmooring:
- Do not escape HTML entities
- Use consistent formatting for title and description with opengraph, schema, and twitter_cards
- Improve readability of twitter_cards

Closes #12418

### DIFF
--- a/tpl/tplimpl/embedded/templates/opengraph.html
+++ b/tpl/tplimpl/embedded/templates/opengraph.html
@@ -4,11 +4,11 @@
   <meta property="og:site_name" content="{{ . }}">
 {{- end }}
 
-{{- with or .Title site.Title site.Params.title | plainify}}
+{{- with or .Title site.Title site.Params.title | plainify }}
   <meta property="og:title" content="{{ . }}">
 {{- end }}
 
-{{- with or .Description .Summary site.Params.description | plainify }}
+{{- with or .Description .Summary site.Params.description | plainify | htmlUnescape | chomp }}
   <meta property="og:description" content="{{ . }}">
 {{- end }}
 
@@ -18,7 +18,9 @@
 
 {{- if .IsPage }}
   <meta property="og:type" content="article">
-  <meta property="article:section" content="{{ .Section }}">
+  {{- with .Section }}
+    <meta property="article:section" content="{{ . }}">
+  {{- end }}
   {{- $ISO8601 := "2006-01-02T15:04:05-07:00" }}
   {{- with .PublishDate }}
     <meta property="article:published_time" {{ .Format $ISO8601 | printf "content=%q" | safeHTMLAttr }}>
--- a/tpl/tplimpl/tplimpl_integration_test.go
+++ b/tpl/tplimpl/tplimpl_integration_test.go
@@ -305,3 +305,109 @@ title: p2
 		"<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>\n<urlset xmlns=\"http://www.sitemaps.org/schemas/sitemap/0.9\"\n  xmlns:xhtml=\"http://www.w3.org/1999/xhtml\">\n  <url>\n    <loc>/p2/</loc>\n  </url>\n</urlset>\n",
 	)
 }
+
+// Issue 12418
+func TestOpengraph(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+capitalizeListTitles = false
+disableKinds = ['rss','sitemap']
+languageCode = 'en-US'
+[markup.goldmark.renderer]
+unsafe = true
+[params]
+description = "m <em>n</em> and **o** can't."
+[params.social]
+facebook_admin = 'foo'
+[taxonomies]
+series = 'series'
+tag = 'tags'
+-- layouts/_default/list.html --
+{{ template "_internal/opengraph.html" . }}
+-- layouts/_default/single.html --
+{{ template "_internal/opengraph.html" . }}
+-- content/s1/p1.md --
+---
+title: p1
+date: 2024-04-24T08:00:00-07:00
+lastmod: 2024-04-24T11:00:00-07:00
+images: [a.jpg,b.jpg]
+audio: [c.mp3,d.mp3]
+videos: [e.mp4,f.mp4]
+series: [series-1]
+tags: [t1,t2]
+---
+a <em>b</em> and **c** can't.
+-- content/s1/p2.md --
+---
+title: p2
+series: [series-1]
+---
+d <em>e</em> and **f** can't.
+<!--more-->
+-- content/s1/p3.md --
+---
+title: p3
+series: [series-1]
+summary: g <em>h</em> and **i** can't.
+---
+-- content/s1/p4.md --
+---
+title: p4
+series: [series-1]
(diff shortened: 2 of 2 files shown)


tpl/tplimpl/embedded/templates/opengraph.html
```diff
@@ -1,14 +1,14 @@
 <meta property="og:url" content="{{ .Permalink }}">
 
-{{- with or site.Title site.Params.title | plainify }}
+{{- with or site.Title site.Params.title | .RenderString | plainify | htmlUnescape }}
```
bep, line 3: I suggest we focus on fixing #12418 and not add other not obviously related.

I assume the HTML gets escaped the second time in

```html
content="{{ . }}"
```

Where is the first escape?

Also, a test for #12418 would help.

jmooring:
Where the escapes coming from?

Front matter:

```
description = "This is <em>emphasized</em> and this is **bold**, but it's been stripped."
```

Passing `description` through `.RenderString` renders the single quote to `&rsquo;`.
Passing that through `plainify` renders `&rsquo;` to `&amp;rsquo;`.

I can split this into three PR's each with their own issue:

1. Fix #12418
2. Fix the same problem with `Page.Title`, `Site.Title`, and `Site.Params.title` in the opengraph template  (#8698)
3. Fix the same problems in schema.html and twitter_cards.html, and improve readability of twitter_cards.html at the same time

I will add a test for #12418, and add additional conditions to that test with each subsequent PR.

Believe me, I am as fatigued by this as you are.

bep:
>Passing description through .RenderString renders the single quote to &rsquo;.

`.RenderString` wasn't in the template before this PR, was it? Where is the double escape in the below?

```
{{- with or .Description .Summary site.Params.description | plainify }}
  <meta property="og:description" content="{{ . }}">
{{- end }}
```

Using Markdown syntax in `title:` e.g. sounds foreign to me and adding `.RenderString` all over the place is something we need to discuss. I have been known to overthink things in the performance department, but `.RenderString` isn't particulary light weight.

I suggest that we fix the bug with as little change as possible. People can create their own templates if they need special things.

tpl/tplimpl/embedded/templates/opengraph.html
```diff
@@ -1,14 +1,14 @@
...
   <meta property="og:site_name" content="{{ . }}">
```
chalin, line 4: Nit: could we unindent all of the "  <meta...>" lines (lines 4, 8, 12 here, and the other file) to match the first line of this file that is not indented.

Otherwise, as illustrated in #12418, we get HTML generated like this:


```html
<meta name="description" content="Test line with a single quote: can&amp;rsquo;t.">
<meta property="og:url" content="http://localhost:1313/docs/page/">
  <meta property="og:site_name" content="My New Hugo Site">
  <meta property="og:title" content="Some page">
  <meta property="og:description" content="Test line with a single quote: can&amp;rsquo;t.">
  ...
<meta name="twitter:description" content="Test line with a single quote: can&rsquo;t.">
```

chalin (commented):
Thanks for fixing this @jmooring!
See inline for one nit.

tpl/tplimpl/embedded/templates/opengraph.html
jmooring: Given that you (almost always?) minify in production this definitely seems like a nit. I'd prefer indented code to improve template readability per #10900.

chalin: FYI, my use case is this: I diff non-minified generated site files of the websites I work on so that I can catch bugs like #12418 (it's impossible to diff minified site files).

In the case of this PR, we can have our cake and eat it too: satisfy #10900 and use appropriate whitespace trimming directives. My 2 cents. Thanks for giving this some thought.

jmooring: Per bep's comment above...

> People can create their own templates if they need special things.

My view on many of the embedded templates is that they are minimal examples, and every time we touch one somebody gets upset. I'd rather eliminate many of them and push responsibility back to site and theme authors. But I guess that train left the station too long ago.

If I were you I would override the opengraph, schema, and twitter_card templates. Based on bep's review, most of the changes in this PR will be eliminated, focusing instead on this one problem (double-escaping HTML entities in the description field of the opengraph template).

jmooring:
@bep 

> Where is the double escape in the below?

When `og:description` is derived from .Summary, either automatic or with summary divider, e.g.

```
---
title: p1
---

Test line with a single quote: can't.
<!--more-->
```

There are three test cases that I need to include, with examples here:
<https://github.com/gohugoio/hugo/issues/12418#issuecomment-2072742073>

We've got way to many ways to define Summary, each with their own idiosyncrasies:
<https://github.com/gohugoio/hugo/issues/8910#issuecomment-903158600>

tpl/tplimpl/embedded/templates/opengraph.html
chalin: I totally get it. No worries, I'll roll with what you provide, thanks!

bep:
@jmooring OK, 

So,

```handlebars
{{- with or site.Title site.Params.title | plainify | htmlUnescape }}
```

Etc. should work (??) and is perfectly fine.

Adding `.RenderString` to the mix needs a discussion (it will certainly slow down asciidoc builds ...)

jmooring:
OK, we pass markdown through un-rendered. To handle the manual summary divider in one of the (now) 5 test cases, we need to  chomp the trailing newline:

```
{{- with or site.Title site.Params.title | plainify | htmlUnescape | chomp }}
```

Hopefully that's OK.

PR #12444, opened 2024-04-28
Enable cross-site translations

### LINKED ISSUE #7844: Add .Site.Language.I18n method (or .Translate?)
Following the support question in [this forum thread](https://discourse.gohugo.io/t/get-i18n-string-from-arbitrary-language-not-current-pages-language/28811): it would be nice to have `site.Translation` method to leverage `i18n` information from the `.Translations` range.

jmooring:
This introduces two methods to enable cross-site translations:

- Site.Translate returns a translated string in the language of the given site.

  {{ range .AllTranslations }} {{ .Site.Translate "cat" 6 }} {{ end }}

- Sites.GetSite returns the site corresponding to the given language. This method simplifies cross-site translations.

  {{ with .Sites.GetSite "en" }} {{ .Translate "dog" 7 }} {{ end }}

Closes #7844

### DIFF
--- a/resources/page/site.go
+++ b/resources/page/site.go
@@ -21,6 +21,7 @@ import (
 	"github.com/gohugoio/hugo/config/privacy"
 	"github.com/gohugoio/hugo/config/services"
 	"github.com/gohugoio/hugo/identity"
+	"github.com/spf13/cast"
 
 	"github.com/gohugoio/hugo/config"
 
@@ -145,6 +146,20 @@ func (s Sites) First() Site {
 		return nil
 	}
 	return s[0]
+}
+
+// GetSite returns the site corresponding to the given language.
+func (s Sites) GetSite(lang string) (Site, error) {
+	slang, err := cast.ToStringE(lang)
+	if err != nil {
+		return nil, err
+	}
+	for _, site := range s {
+		if site.Language().Lang == slang {
+			return site, nil
+		}
+	}
+	return nil, nil
 }
 
 // Some additional interfaces implemented by siteWrapper that's not on Site.
--- a/langs/i18n/i18n_integration_test.go
+++ b/langs/i18n/i18n_integration_test.go
@@ -126,3 +126,43 @@ title: home_es
 	b.AssertFileContent("public/es/index.html", `home_es_gato`)
 	b.AssertFileContent("public/fr/index.html", `home_fr_gato`)
 }
+
+// Issue 7844
+func TestSiteLanguageTranslate(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+disableKinds = ['page','rss','section','sitemap','taxonomy','term']
+defaultContentLanguage = 'de'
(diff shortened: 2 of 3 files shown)


hugolib/site_new.go
```diff
@@ -558,3 +559,9 @@ func (s *Site) AllPages() page.Pages {
 func (s *Site) AllRegularPages() page.Pages {
 	return s.h.RegularPages()
 }
+
+// Translate returns a translated string for id.
+func (s *Site) Translate(id string, args ...any) (string, error) {
+	ctx := context.Background()
```
bep, line 565: Add `ctx` as the first arg in `Translate`. It will be "auto injected".

That said. I don't remember the original issue, but looking at the below I would say that this methods belongs on the `Language` struct.

You would need to say `.Site.Language.Translate` ... which is a little longer ... but cleaner.

hugolib/site_new.go
```diff
@@ -558,3 +559,9 @@ func (s *Site) AllPages() page.Pages {
...
+	return lang.New(s.Deps, langs.GetTranslator(s.Language())).Translate(ctx, id, args...)
```
bep, line 566: The above is propbably simpler to reason about when moving the method, but we do not want to create a new language for each translated string.

resources/page/site.go
```diff
@@ -147,6 +152,19 @@ func (s Sites) First() Site {
...
+	slang, err := cast.ToStringE(lang)
+	if err != nil {
+		return nil, err
+	}
+	for _, site := range s {
+		if site.Language().Lang == slang {
+			return site, nil
+		}
+	}
+	return nil, fmt.Errorf("unable to find site with lang %q", lang)
```
bep, line 165: I think it matches _most_ of the similar situations to return nil in this case. That way you can try to translate arbitrary languages by wrapping it in `with`.

bep (changes requested):

resources/page/site.go
jmooring: Thanks for the review. It looks like you have recently implemented something similar to site.GetSites while working on the pages from data stuff, so I was going to hold off until that was finalized. It looked like it I might get hugo.GetSite for free. Let me know if that's not accurate.

bep: No, I haven't touched anything similar in that branch. What I was concerned about in that setting was to make sure that when you do `i18n` and similar, you would get the _current_ site/language's translation, the same as you do today in `single.html` and friends.

bep: That said #12440 is shaping up nicely. I need to take it the final mile (resources support is missing, for one), but I have tested it out on a "real project" and it's very nice to work with (especially after I figured out how to map the `gotmpl` extension to `html` in VSCode). Having it in the templates gives the unexpected benefit of partial rebuilds "for free", so edits to assets included via `resources.Get` etc. just works. I also think this finally can make Hugo's OpenAPI integration generally useful.

jmooring:
I am unable to figure this out in a reasonable time frame, so I am going to punt for now.

PR #12459, opened 2024-05-04
markup/goldmark: Support extras extension

jmooring:
Enables inclusion of these HTML elements in Markdown:

- Inserted Text (++inserted++)
- Mark Text (==marked==)
- Subscript (H\~2\~O)
- Superscript (1^st^)

Default configuration:

```yaml
markup:
  goldmark:
    extensions:
      extras:
        insert:
          enable: false
        mark:
          enable: false
        subscript:
          enable: false
        superscript:
          enable: false
```


jmooring:
@bowman2001 Building from this PR should just work. Note the configuration structure in the comment above; one extension with four options.

See documentation preview:
<https://deploy-preview-2555--gohugoio.netlify.app/getting-started/configuration-markup/#extras-extension>

bowman2001:
Thanks a lot! Your integration test is way more concise than mine was.

markup/goldmark/convert.go
```diff
@@ -194,6 +195,30 @@ func newMarkdown(pcfg converter.ProviderConfig) goldmark.Markdown {
 		))
 	}
 
+	if cfg.Extensions.Extras.Insert.Enable {
```
bep, line 198: The `extras` has the same structure as Hugo's, so I would just do

```go
extensions = append(extensions, extras.New(
			extras.Config{
                            Mark: extras.MarkConfig{Enable: cfg.Extensions.Extras.Mark.Enable},
                            ...
},
))
```

If all are disabled, you'd end up with a `noop` extension.

bep: OK, I see I was a little unclear about this. The current setup allows for setting up 1 extension in one go without any conditionals:

```go
extensions = append(extensions, extras.New(
	extras.Config{
	   Insert: extras.InsertConfig{Enable: cfg.Extensions.Extras.Insert.Enable},
           Mark: extras.MarConfig{Enable: cfg.Extensions.Extras.Mark.Enable},
           // ... the rest of the extras.
	},
))
```

The above is, in my head,

1. Easier to reason about.
2. It would allow potential upstream improvements, maybe the implementation could be sped up by making it into one parser instance?

jmooring: Gotcha.

bep:
This looks good now, I will merge this early next week and get it into the same release as the "pages from data" feature. I'm just holding it just in case I need to do a patch release for some other reason.

PR #12494, opened 2024-05-14
Use UTC time to generate lastmod time in sitemap

xuhdev:


xuhdev:
Actually I found that the merely changes the format without performing the actual timezone conversion. I'm in UTC-7 so it could be confusing.

jmooring:
I don't understand how the existing code is incorrect. The resulting date/time stamp conforms to https://www.sitemaps.org/protocol.html.

xuhdev:
Yes, it conforms. But the time is incorrect. (I thought this PR fixed the issue but it turned out to be a "fake" fix.)

I tried to replace `-07` with different timezones in the sitemap template, but I always get the exact same time + the specified timezone, while I would expect some timezone conversion to take place.

jmooring:
This is a support issue, not a bug to be fixed. Please use the forum (https://discourse.gohugo.io/) for questions and troubleshooting. Thanks.

xuhdev:
After having dug for a while, I still think this is more likely a bug than a support question. (While this PR certainly didn't clarify.) I opened #12495.

PR #12505, opened 2024-05-16
[editorial] google_analytics.html: trim trailing whitespace from tpl actions

chalin:
"Editorial" change only: from `{{ ... }}` to `{{- ... }}`, to trim template action trailing whitespace.


jmooring:
To be consistent with other embedded templates can we trim leading instead? For example:
<https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/schema.html>

chalin:
Sure whatever convention you prefer.

jmooring:
Trim leading please... thanks.

chalin:
Ok, done ✨

jmooring (approved):
LGTM

PR #12516, opened 2024-05-17
Support for .textbundle leaf nodes

markphilpot:
I use Ulysses external folder support for my writing, and they use textbundle format for their "posts" -- https://textbundle.org/

The spec uses `text.md` as the markdown container within the `.textbundle` directory and `assets/*` for relative images. With this change, hugo recognizes the `text.md` leaf inside the textbundle and the paths to relative images work correctly.

If there are any tests that should be updated for this, I'd be happy to update them (but some help find where would be appreciated).

### DIFF
--- a/hugofs/files/classifier.go
+++ b/hugofs/files/classifier.go
@@ -77,7 +77,7 @@ func IsIndexContentFile(filename string) bool {
 
 	base := filepath.Base(filename)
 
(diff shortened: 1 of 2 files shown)


markphilpot:
I see the code base has significantly evolved since I made my local change... I will attempt to get this support working in the latest version then resubmit.

jmooring:
Please see <https://github.com/gohugoio/hugo?tab=readme-ov-file#contributing>

PR #12517, opened 2024-05-17
Support for textbundle leaf nodes

markphilpot:
Recognize the `text.md` inside a `.textbundle` file as a leaf node so the relative images in the `assets/` directory are correctly handled.

textbundle spec: https://textbundle.org/

There are probably some tests that I need to update, but if someone can direct me in the right direction that would be appreciated.


jmooring:
Please see <https://github.com/gohugoio/hugo?tab=readme-ov-file#contributing>

PR #12529, opened 2024-05-22
hugolib: Allow override of sitemap file name

### LINKED ISSUE #12525: sitemap file name not changing when custom filename is specified
If I am not doing anything wrong, it seems that the newer version is not generating a sitemap with a custom name. I have copied the [configuration from the docs](https://gohugo.io/templates/sitemap-template/#configuration) and changed the `filename` but it is always generating the sitemap with the default `sitemap.xml` name.

this is my config
image

and when I run `hugo` to build the site, the sitemap in the `public` folder is `sitemap.xml` instead of the `sitemap-1.xml` in the configuration.

image

I know it was working in some versions ago, but I am not sure which one it was working. Also, I can be misconfiguring stuff, so I am sorry I am doing something wrong.

hugo version
<pre>
 v0.126.1+extended
</pre>

yes

Thank you for the great project! I love hugo <3

jmooring:
Closes #12525

### DIFF
--- a/hugolib/content_map_page.go
+++ b/hugolib/content_map_page.go
@@ -1881,11 +1881,19 @@ func (sa *sitePagesAssembler) addStandalonePages() error {
 	}
 
 	if sitemapEnabled {
-		addStandalone("/_sitemap", kinds.KindSitemap, output.SitemapFormat)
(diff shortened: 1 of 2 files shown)


jmooring:
This fix is imperfect in that the file extension of the overriding filename is ignored, but I think that's a good thing.

PR #12563, opened 2024-06-02
tpl/inflect: Add the SI function to format numbers with SI notation

razonyang:
https://en.wikipedia.org/wiki/Metric_prefix

One of use case of mine: a repository card.

image

Not sure if I put this function into the right place/package.

### DIFF
--- a/tpl/inflect/inflect.go
+++ b/tpl/inflect/inflect.go
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"strings"
 
+	_humanize "github.com/dustin/go-humanize"
 	_inflect "github.com/gobuffalo/flect"
 	"github.com/spf13/cast"
 )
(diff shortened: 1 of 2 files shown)


jmooring:
How would I use this with the [`lang.FormatSomething`](https://gohugo.io/functions/lang/) functions? Each of these functions returns a string. I imagine in some locales you might want `12,3 kb` instead of `12.3 k`.

Maybe this should be `lang.FormatNumberSI`, and possibly create derivatives of `lang.FormatNumberCustom` and `lang.FormatCurrency` as well.

I suggest closing this PR and opening a proposal to work out the details.

razonyang:
> I imagine in some locales you might want 12,3 kb instead of 12.3 kb.

Haven't thought about this.

Done, https://discourse.gohugo.io/t/function-to-format-numbers-with-si-notation/50131, close for now.

PR #8700, opened 2021-06-27
Publish tagged as Github Packages

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://github.com/gliptak/hugo/actions/runs/975239969

https://github.com/gliptak/hugo/pkgs/container/hugo

```
docker run ghcr.io/gliptak/hugo:latest version
hugo v0.85.0-DEV linux/amd64 BuildDate=2021-06-27T01:03:38Z
```

#5056

PR #12573, opened 2024-06-07
ci: add image build

### LINKED ISSUE #10760: Create and publish official Docker images
The latest docker image on https://hub.docker.com/r/klakegg/hugo is 0.111.3 which was pushed 2 weeks ago but was released on 12 March.
When searching for any higher version like 0.112.0 or 0.114.0 no images/tags can be found.

Same issue as #10324 reported some months ago.

morremeyer:
## Description

This adds a Docker image build running in GitHub actions for `linux/amd64` and `linux/arm64` images.

The build runs for every PR that changes either the workflow or the Dockerfile to verify the image still can be built correctly, but does not push the image for these runs.

For every tag, the images are built and pushed to the GitHub OCI registry.

Resolves #10760.

## Testing

I have run tests with a slightly modified version (to enable testing) in my fork, see the [test-add-image-build](https://github.com/morremeyer/hugo/tree/ci/test-add-image-build) branch.

This builds the image with [this workflow](https://github.com/morremeyer/hugo/actions/runs/9416950944) and results in the package being pushed to https://github.com/morremeyer/hugo/pkgs/container/hugo.

## Continuous testing

I am aware of https://github.com/gohugoio/hugo/pull/8700#issuecomment-869981438 and happy to add automated testing of the images if there is a clear requirement for what should be tested.

However, as with the Dockerfile being a convenience, the same could also be applied to the resulting image in terms of guarantees.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,0 +1,127 @@
+name: Build Docker image
+
+on:
+  push:
+
+  # Run on pull requests to verify that the image still builds correctly
+  # The image is not pushed on pull requests
+  pull_request:
+    paths:
+      - Dockerfile
+      - .github/workflows/image.yml
+
+permissions:
+  packages: write
+
+env:
+  REGISTRY_IMAGE: ghcr.io/gohugoio/hugo
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - linux/amd64
+          - linux/arm64
+    steps:
+      - name: Prepare
+        run: |
+          platform=${{ matrix.platform }}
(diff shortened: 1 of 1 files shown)


morremeyer:
This is now ready for review, no more force-pushes here.

bep:
@morremeyer thanks for this. I just wanted to chime in and say that I have seen this, and it looks good. I just need to think about this, as I'm a little stretched for time, and 1) A setup like this will eventually have to be updated/fixed when GitHub changes its API etc. and 2) This will open the gate for people wanting a more advanced `Docerfile` setup (e.g. Dart Sass, asciidoc etc. installs).

phantasy36:
> 现在已经准备好进行审查了，这里不再有强制推动。

What is the name of the repository to pull?

bep:
Sorry for the delay; I'm going to test and complete this in #12734.

PR #12588, opened 2024-06-10
Add instructions on readme to build from the latest dev code, rather then just building from the latest release

godalming123:
Add instructions on readme to build from the latest dev code, rather then just building from the latest release.


bep:
Thanks, but there's already instructions about how to do this in the paragraph above.

Issue #12438, opened 2024-04-26
PublishDate does not roll-up to section, taxonomy, or term pages


schu:
### What version of Hugo are you using (`hugo version`)?

<pre>
hugo v0.125.4+extended darwin/arm64 BuildDate=2024-04-25T13:27:26Z VendorInfo=brew
</pre>

### Does this issue reproduce with the latest release?

Yes.

### Description

After upgrading to v0.125 I see changes like the following for `categories/index.xml` and `tags/index.xml` files:

```
-      <pubDate>Sat, 27 Oct 2018 08:30:00 +0000</pubDate>
+      <pubDate>Mon, 01 Jan 0001 00:00:00 +0000</pubDate>
```

For subindex files like `categories/foo/index.xml` the `pubDate` is set to the expected value, i.e. the generated output is identical to versions < v0.125.

From looking at the [changelog](https://github.com/gohugoio/hugo/releases/tag/v0.125.0) and [commit f0a26cf58e9bfbe55c091d51a4133bc22cad5d1f](https://github.com/gohugoio/hugo/commit/f0a26cf58e9bfbe55c091d51a4133bc22cad5d1f) it's not clear to me what needs to be changed in order to retain the previous (default) selection of values for `pubDate`? Am I missing something?

Also, with above hugo version `{{ hugo.Version }}` doesn't seem to be set, resulting in the following output for the `generator` tag:

```
-    <generator>Hugo -- gohugo.io</generator>
+    <generator>Hugo</generator>
```

jmooring:
@schu

The version string was added to the generator tag in v0.125.0 to address #11692, and was removed in v0.125.3 to address https://github.com/gohugoio/hugo/pull/12300#issuecomment-2068869567. We should have closed #11692 as "won't fix".

I'll look into the zero pubdate  problem.

jmooring:
The `Date` and `Lastmod` "roll-up" to section pages and term pages. The `PublishDate` does not, and it should.

Applicable to:
- Automatic section and term pages
- Section and term pages without `Date` or `PublishDate`

bep:
>I think the PublishDate should roll-up to the latest date before now(). 

Yea, that sounds reasonable.

schu:
Is there a workaround that I could apply to get back the old behavior?

Or is there a chance to revert to the old behavior until a working fix is found?

jmooring:
Override the embedded RSS template, and change this line:

[...]

PR #12629, opened 2024-06-28
tpl/tplimpl: Fall back from publish date to date in RSS template

jmooring:
Improves #12438


bep:
Just to be clear, if we (I) fix #12438 -- then we can close this?

jmooring:
Closing in favor of https://github.com/gohugoio/hugo/pull/12633.

PR #12631, opened 2024-06-29
create/skeletons: Fix languageCode region subtag

CyrusYip:
Region subtag should be in ALLCAPS

### DIFF
--- a/create/skeletons/theme/hugo.toml
+++ b/create/skeletons/theme/hugo.toml
@@ -1,5 +1,5 @@
 baseURL = 'https://example.org/'
-languageCode = 'en-us'
+languageCode = 'en-US'
 title = 'My New Hugo Site'
 
 [[menus.main]]


CyrusYip:
Reference:

> - [Region subtag](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#region_subtag)
>
>     Optional. This subtag defines a dialect of the base language from a particular location, and is either 2 letters in ALLCAPS matching a country code, or 3 numbers matching a non-country area. For example, `es-ES` is for Spanish as spoken in Spain, and `es-013` is Spanish as spoken in Central America. "International Spanish" would just be `es`.
> —[lang - HTML: HyperText Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)

jmooring (approved):
LGTM.

PR #12642, opened 2024-07-05
Expose commit body

### LINKED ISSUE #10905: Full Commit Message in .GitInfo
**Current Behavior**

The `.GitInfo` object hat a `Subject` field which contains the first line of the commit message

**Proposed Bahavior**

Add another field `Message` with the full commit message.

**Reasoning**

We implemented a content management system based on Git and Hugo. For the meta information it would be useful to present the full commit message, e.g. as a tooltip on top of the subject.

PeskyPotato:
Fixes #10905
 
## Changes
This PR adds `Body` to the GitInfo struct type to expose the commit message's body as is available in [bem/gitmapv1.4.0](https://github.com/bep/gitmap/releases/tag/v1.4.0). If no commit body is present an empty string is returned.

Docs have been updated with the new parameter, I've included an email of the output as well.

## Tests
No tests were added for this feature. I ran `mage check` after installing asciidoctor and there was no output, I assume this is good?

`go test -run ''` passes all tests.

I tested this against a local test site and it works as expected.


PeskyPotato:
source and docs commit merged into one.

source/fileInfo.go
```diff
@@ -175,6 +175,8 @@ type GitInfo struct {
 	AuthorDate time.Time `json:"authorDate"`
 	// The commit date.
 	CommitDate time.Time `json:"commitDate"`
+	// The commit message's Body.
+	Body string `json:"Body"`
```
bep, line 179: Can you make it `json:"body"` (which is in line with the others)

bep (changes requested):

source/fileInfo.go
PeskyPotato: Done, will combine the commits and push now.

Issue #11553, opened 2023-10-11
Slice of slices changed behaviour


raoulb:
I have an old partial that splits a slice of pages into a slice of slices of pages (in my case pairs).

This worked well with Hugo up to v0.113.0 but broke with v0.114 up to the most recent 0.119.0.
For me this breaks a production site.

A minimal example can be found here:
* https://github.com/raoulb/hugo_issue_slices

This issue might be related (or equivalent?) to #11131.

BTW: Such a function to split slices into even pieces appears to be a good candidate for a Hugo builtin functionality anyway.

raoulb:
Any updates on this?

cmahnke:
@jmooring Looks like a duplicate to #11131 to me.

Issue #11131, opened 2023-06-18
Unexpected results with append template function


jmooring:
Testing https://github.com/gohugoio/hugo/commit/12dc9a6e4acd5280a3e8b1658cbb96669fa97057:

PASS

```
{{ $s := slice | append "a" }}

[
  "a"
]
```

FAIL

```
{{ $s := slice | append (slice "a") }}

[
  "a"
]
```

PASS

```
{{ $s := slice | append (slice "a") (slice "b") }}

[
  [
    "a"
  ],
  [
    "b"
  ]
]
```

bep:
First, I don't see the relevance to this with https://github.com/gohugoio/hugo/commit/12dc9a6e4acd5280a3e8b1658cbb96669fa97057


```bash
{{ $s := slice | append (slice "a") }}
[
  "a"
]
```

I just checked, and the above was how it behaved also in v0.112.0 and it's not something we can change without breaking lots of sites, and I actually think the above behaviour is the most logical. It's not like we can add a dimension to the target slice.

```
{{ $s := slice | append (slice "a") (slice "b") }}

[
  [
    "a"
  ],
  [
    "b"
  ]
]
```

OK, the above doesn't look right, but it's not obvious what the right answer is. Again, I don't think we could/should expand the dimensions in the target slice. But again, this is how it behaved in older Hugo versions, so it's nothing new, right?

jmooring:
> I don't see the relevance to this with https://github.com/gohugoio/hugo/commit/12dc9a6e4acd5280a3e8b1658cbb96669fa97057

Sorry, what I meant was, I built Hugo at that commit for testing. But...

Let's set aside the above for a moment. I think I have/had a fundamental misunderstanding of how the `append` function is supposed to work. If I start with an empty slice, I thought I could add additional elements including scalars, slices, and maps. The root of my misunderstanding is that both of these produce the same result:

```
{{ $s := slice }}
{{ $s = $s | append 1 2 }}  # I want to append 2 scalar values, each as their own element

{{ $s := slice }}
{{ $s = $s | append (slice 1 2) }} # I want to append a slice of 2 scalar values as a single element

[
  1,
  2
]
```

The above is true for both v0.113.0 and v0.114.0. We didn't break anything, it just doesn't work the way I expected/wanted. And, as you pointed out, changing that behavior will break lots of sites.

So, to create a slice of slices (with both v0.113.0 and v0.114.0) I have to do this:

```
{{ $s := slice }}
[...]

PR #12661, opened 2024-07-16
collections: Add function to partition page slices

raoulb:
Another thing I think is missing from the standard palette of template functions: partition a slice (of pages) into a set of equal chunks (plus a possible remainder).

This implements the proposal from issue [#11553](https://github.com/gohugoio/hugo/issues/11553).

Help wanted to discuss the implementation details. I'll also add docs and tests.

### DIFF
--- a/common/collections/collections.go
+++ b/common/collections/collections.go
@@ -19,3 +19,9 @@ package collections
 type Grouper interface {
 	Group(key any, items any) (any, error)
 }
+
+// Partitioner defines a very generic way to partition sets of items into
+// chunks of given maximal length.
+type Partitioner interface {
+	Partition(n any, items any) (any, error)
+}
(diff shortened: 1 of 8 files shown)


bep:
Thanks for this. Issue #11553 is about a bug in `slice`. This PR does not fix that particular bug. I don't say this isn't a great idea, but we cannot start the discussion with a implemention. Please opne a new proposal if you want to discuss this.

raoulb:
This PR does not directly address the change in behavior of `append` combined with `slice` as shown in bug #11131 (which broke several sites). For that, one should first decide if this change is actually a bug or just the new correct behavior. (Last time I've read that report it appeared to me tendency was towards the latter.)

However, the proposed new `partition` functionality here directly addresses my own bug report #11553 which is very likely related to the first one.

I just updated the branch to include some unit tests. This code is now no longer WiP from my point of view and I'd like to know what needs to be changed to get this merged. (Should I open a fresh PR?)

In any case, I propose to add to Hugo the partition functionality with the semantics as documented and implemented by this code.

PR #12681, opened 2024-07-27
math: Add exp and log2 / log10 functions

raoulb:
Hugo already has the natural `log` function. So it seems like a gap missing out on its inverse, the `exp` function.

We also add a `log2` (binary logarithm) and `log10` (decimal logarithm) function which both can be very useful at times. By including explicit functions, template authors do not have to implement change of base themselves. (We do not do change of base but rather use functions provided by the golang math lib.)

### DIFF
--- a/docs/content/en/functions/math/E.md
+++ b/docs/content/en/functions/math/E.md
@@ -1,0 +1,19 @@
+---
+title: math.E
+description: Returns the mathematical constant e.
+categories: []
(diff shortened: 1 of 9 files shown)


bep:

raoulb:
I don't know what you have in mind with an "accepted proposal" for pure and simple functionality like this ... the proposal is the PR ...

PR #12682, opened 2024-07-27
math: Add golden ratio constant

raoulb:
This PR add the golden ratio constant. This constant is often used in graphical contexts to get pleasing aesthetics. In will be useful for subdividing layouts, partition page space or fitting images and possibly more.


bep:

PR #12683, opened 2024-07-27
math: Add cube root function

raoulb:
The final PR of this series adds the cube root function. While it could be computed via fractional powers, providing a dedicated function has benefits. F.e. it is faster and avoids corner cases. Equally important, it is much clearer to write and read. (Which is especially important as go templates can quickly become convoluted.)


bep:

PR #12695, opened 2024-07-31
resources/page: Expand parmalinks tokens in `url`

### LINKED ISSUE #9714: Allow permalinks configuration syntax to `url` Front Matter.
### Currently

Currently, batch permalink settings can only happen through the site's config's `permalinks` object. For every setting, it takes a content directory as key and a string that can include "template strings" syntax as value.

To set the permalink of the `/recipe` directory using the slug:

```yaml
permalinks:
  recipe: /delicious-recipe/:slug
```

That is limiting to batch editing content files of a root directory. The recent `:sections` templates brought more power to the desired permalink but did not fully resolve the limited "targeting" of the batch. For example, if your content files are in a subdirectory `content/cooking/recipe`, you would have no way (at least I did not find one) to target the files in the directory to batch assign a permalink structure.

### The proposal

Allow the `url` Front Matter key to use the same syntax as permalink's configuration.

```yaml
---
# content/cooking/recipe/_index.md
title: Delicious Recipes
url: /delicious-recipe
cascade:
  url: /delicious-recipe/:slug
---
```

n1xx1:
This change allows to use permalink tokens in `url` front matter fields. This should be useful to target more specific pages instead of using a global permalink configuration. It's expected to be used with `cascade`.

Fixes #9714.

### DIFF
--- a/hugolib/page__paths.go
+++ b/hugolib/page__paths.go
@@ -141,6 +141,21 @@ func createTargetPathDescriptor(p *pageState) (page.TargetPathDescriptor, error)
 	desc.PrefixFilePath = s.getLanguageTargetPathLang(alwaysInSubDir)
 	desc.PrefixLink = s.getLanguagePermalinkLang(alwaysInSubDir)
 
+	if desc.URL != "" && strings.IndexByte(desc.URL, ':') >= 0 {
+		// Attempt to parse and expand an url. Probably it's a good
+		// idea to cache the parsed url to avoid regexp matching and memory allocations.
+
+		opath, err := d.ResourceSpec.Permalinks.ExpandPattern(desc.URL, p)
+		if err != nil {
+			return desc, err
+		}
+
+		if opath != "" {
+			opath, _ = url.QueryUnescape(opath)
+			desc.URL = opath
+		}
+	}
+
 	opath, err := d.ResourceSpec.Permalinks.Expand(p.Section(), p)
 	if err != nil {
 		return desc, err
--- a/hugolib/page_permalink_test.go
+++ b/hugolib/page_permalink_test.go
@@ -59,6 +59,8 @@ func TestPermalink(t *testing.T) {
 
 		// test URL overrides
 		{"x/y/z/boofar.md", "", "", "/z/y/q/", false, false, "/z/y/q/", "/z/y/q/"},
+		// test URL override with expands
+		{"x/y/z/boofar.md", "", "test", "/z/:slug/", false, false, "/z/test/", "/z/test/"},
 	}
 
 	for i, test := range tests {
--- a/resources/page/permalinks.go
+++ b/resources/page/permalinks.go
@@ -102,6 +102,16 @@ func NewPermalinkExpander(urlize func(uri string) string, patterns map[string]ma
 	return p, nil
 }
 
+// ExpandPattern expands the path in p with the specified expand pattern.
+func (l PermalinkExpander) ExpandPattern(pattern string, p Page) (string, error) {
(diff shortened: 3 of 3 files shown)


hugolib/page__paths.go
```diff
@@ -141,6 +141,21 @@ func createTargetPathDescriptor(p *pageState) (page.TargetPathDescriptor, error)
 	desc.PrefixFilePath = s.getLanguageTargetPathLang(alwaysInSubDir)
 	desc.PrefixLink = s.getLanguagePermalinkLang(alwaysInSubDir)
 
+	if desc.URL != "" && strings.IndexByte(desc.URL, ':') >= 0 {
+		// Attempt to parse and expand an url. Probably it's a good
+		// idea to cache the parsed url to avoid regexp matching and memory allocations.
+
+		opath, err := d.ResourceSpec.Permalinks.ExpandPattern(desc.URL, p)
```
bep, line 148: You are right about the "cache" remark.

1. The parser isn't fast.
2. I predict that 90% of the use of this new feature will be via cascade, so it should be cache friendly.

* You can get a thread safe cache with `maps.NewCache`
* I suggest you use the `GetOrCreate` method there; I also suggest you rename the `ExpandPattern` to `GetOrExpandPattern` to reflect this.

hugolib/page_permalink_test.go
```diff
@@ -59,6 +59,8 @@ func TestPermalink(t *testing.T) {
 
 		// test URL overrides
 		{"x/y/z/boofar.md", "", "", "/z/y/q/", false, false, "/z/y/q/", "/z/y/q/"},
+		// test URL override with expands
+		{"x/y/z/boofar.md", "", "test", "/z/:slug/", false, false, "/z/test/", "/z/test/"},
```
bep, line 63: It's a little hard to see that this new feature works without a integration test.

I suggest that you create one in `page/permalinks_integration.test` -- look for examples in test files with similar names. They use the `hugolib.Test` function.

bep (changes requested):

hugolib/page_permalink_test.go
n1xx1: I did leave this test and implemented a new one in permalinks_integration_test.

resources/page/permalinks.go
```diff
@@ -139,7 +144,12 @@ func init() {
 	}
 }
 
-func (l PermalinkExpander) parsePattern(pattern string) (func(Page) (string, error), error) {
+func (l PermalinkExpander) getOrParsePattern(pattern string) (func(Page) (string, error), error) {
+	expander, ok := l.patternCache.Get(pattern)
```
bep, line 148: This is nit picking, but the cache instance has a `GetOrCreate` method, so I suggest you make it

```go
return l.patternCache.GetOrCreate(pattern, parsePattern)
```

n1xx1: I didn't like GetOrCreate because it needs to handle errors, but I can do something like this if it's what you're looking for:

```
	var err error
	expander := l.patternCache.GetOrCreate(pattern, func() func(Page) (string, error) {
		if !l.validate(pattern) {
			err = &permalinkExpandError{pattern: pattern, err: errPermalinkIllFormed}
			return nil
		}
		// ...
	})
	return expander, err
```

bep: GetOrCreate is used 2 other places, just change the signature to make it return an error.

bep: As I said, 

>GetOrCreate is used 2 other places, just change the signature to make it return an error.

But it's not important, I can fix it later.

bep (approved):

PR #12703, opened 2024-08-01
Allow nested params when using Pages.GroupByParam and Pages.GroupByParamDate

### LINKED ISSUE #12289: Support nested params when using Pages.GroupByParam
<pre>
$ hugo version
hugo v0.113.0+extended linux/amd64 BuildDate=2024-01-23T04:15:35Z VendorInfo=ubuntu:0.113.0-3ubuntu0.1
</pre>

I am not sure

### Description
When using Pages.GroupByParam hugo will fail to build if the param used is a nested like the following example.

Frontmatter:
<pre>
{
   "author": {
       "email": "jsmith@example.org",
       "name": "John Smith"
   },
}
</pre>

Grouping fails
<pre>
{{ range .Pages.GroupByParam "author.name" }}
</pre>

But sorting works
<pre>
{{ range .Pages.ByParam "author.name" }}
</pre>

### Error
<pre>
execute of template failed at <.Pages.GroupByParam>: error calling GroupByParam: there is no such a param
</pre>

I reported this issue on discourse.gohugo.io and was asked to submit this ticket.
https://discourse.gohugo.io/t/how-to-use-groupbyparam-with-param-subkeys/48929

n1xx1:
Basically just use `maps.GetNestedParam` in `resource.GetParam` and `resource.GetParamToLower`. It only affects the `Pages.GroupByParam` and `Pages.GroupByParamDate` methods since it's only used there. There is no need to check for an error because `maps.GetNestedParam` never returns one. The tests are the same as the not nested version but using a nested field.

Fixes #12289.

### DIFF
--- a/resources/page/pagegroup_test.go
+++ b/resources/page/pagegroup_test.go
@@ -55,6 +55,11 @@ func preparePageGroupTestPages(t *testing.T) Pages {
 		p.params["custom_param"] = src.param
 		p.params["custom_date"] = cast.ToTime(src.date)
 		p.params["custom_string_date"] = src.date
+		p.params["custom_object"] = map[string]any{
+			"param":       src.param,
+			"date":        cast.ToTime(src.date),
+			"string_date": src.date,
+		}
 		pages = append(pages, p)
 	}
 	return pages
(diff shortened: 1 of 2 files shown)


resources/page/pagegroup_test.go
```diff
@@ -372,6 +396,24 @@ func TestGroupByParamDate(t *testing.T) {
...
+	t.Parallel()
+	pages := preparePageGroupTestPages(t)
+	expect := PagesGroup{
+		{Key: "2012-04", Pages: Pages{pages[4], pages[2], pages[0]}},
+		{Key: "2012-03", Pages: Pages{pages[3]}},
+		{Key: "2012-01", Pages: Pages{pages[1]}},
+	}
+
+	groups, err := pages.GroupByParamDate("custom_object.date", "2006-01")
+	if err != nil {
```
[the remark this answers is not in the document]
n1xx1: I've never used quicktest or go-cmp, but I think it's correct. Let me know if it's ok or you need something else instead.

resources/page/pagegroup_test.go
```diff
@@ -75,15 +81,13 @@ func TestGroupByWithFieldNameArg(t *testing.T) {
...
 
 	groups, err := pages.GroupBy(context.Background(), "Weight")
-	if err != nil {
-		t.Fatalf("Unable to make PagesGroup array: %s", err)
-	}
-	if !reflect.DeepEqual(groups, expect) {
-		t.Errorf("PagesGroup has unexpected groups. It should be %#v, got %#v", expect, groups)
-	}
+	c.Assert(err, qt.IsNil)
+	c.Assert(groups, comparePageGroup, expect)
```
bep, line 85: I suspect qt.DeepEquals should do the job here?

n1xx1: It doesn't, go-cmp can't compare unexported fields, so you can't compare `Page`s (I tried before making a custom comparer 😅)

n1xx1: I'm just now noticing that you can use `cmpopts.IgnoreUnexported` or `AllowUnexported`, oh well. I can fix it or leave it like this, let me know.

bep: No, leave it, it makes sense ...

PR #12731, opened 2024-08-08
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


yarikoptic:
meanwhile I have signed SLA, rebased, redid codespell fixes with newer codespell (to address conflict + account new found on CI but not locally typo), and fixed indentation after codespell fix which lead to some linting CI jobs to fail.

bep:
You need to include the config setup ONLY.

bep:
I'm closing this for now. This may be interesting some time in the future, but as for now it creates too much noise.

PR #12732, opened 2024-08-08
Very minor: use plural "Actors" since multiple and would be consistent with other examples

yarikoptic:


jmooring:
Please submit documentation edits to the documentation repository:
<https://github.com/gohugoio/hugoDocs>

PR #12734, opened 2024-08-09
github: Build and publish Docker image on release

### LINKED ISSUE #10760: Create and publish official Docker images

bep:
Fixes #10760

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,0 +1,127 @@
+name: Build Docker image
+
+on:
+  push:
+
+  # Run on pull requests to verify that the image still builds correctly
+  # The image is not pushed on pull requests
+  pull_request:
+    paths:
+      - Dockerfile
+      - .github/workflows/image.yml
+
+permissions:
+  packages: write
+
+env:
+  REGISTRY_IMAGE: ghcr.io/gohugoio/hugo
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - linux/amd64
+          - linux/arm64
+    steps:
+      - name: Prepare
+        run: |
+          platform=${{ matrix.platform }}
+          echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
+
+      - name: Checkout
+        uses: actions/checkout@v4.1.6
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5.5.1
+        with:
+          images: ${{ env.REGISTRY_IMAGE }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3.0.0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3.3.0
+
+      - name: Login to GHCR
+        # Login is only needed when the image is pushed
+        if: ${{ startsWith(github.ref, 'refs/tags') }}
+        uses: docker/login-action@v3.2.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push by digest
+        id: build
+        uses: docker/build-push-action@v5.3.0
+        with:
+          context: .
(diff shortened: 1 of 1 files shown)


bep:
@morremeyer a question;

* The build above gets a 403 error ...
* The `GITHUB_TOKEN` needed for this; the token that gets "auto injected" is not good enough for this?
* Reading this, it says I need to create a "personal token" (https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication); is that your experience with this as well?

.github/workflows/image.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf # v3.2.0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@988b5a0280414f521da01fcc63a27aeeb4b104db # v3.6.1
+
+      - name: Login to GHCR
+        # Login is only needed when the image is pushed
+        if: ${{ startsWith(github.ref, 'refs/tags') }}
```
morremeyer, line 51:
```suggestion
```

.github/workflows/image.yml
```diff
@@ -0,0 +1,127 @@
...
+          name: digests-${{ env.PLATFORM_PAIR }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    if: ${{ startsWith(github.ref, 'refs/tags') }}
```
morremeyer, line 86:
```suggestion
```

bep:
>If you want, I can go through the different scenarios on my fork to verify them.

No, that's fine. I need to understand this setup myself, and a little trying-and-failing helps ...

morremeyer:
Feel free to ping me for anything related to the image in the future, too - I'll use the image myself, so this will be a nice way to contribute back.

.github/workflows/image.yml
```diff
@@ -0,0 +1,122 @@
...
+        with:
+          images: ${{ env.REGISTRY_IMAGE }}
+
+          flavor: |
+            latest=false
+
+          # Tag the image with the v-prefixed semver version on tags, tag with the ref on PRs
+          tags: |
+            type=semver,pattern={{raw}}
+            type=ref,event=pr
```
morremeyer, line 104:
```suggestion
```

This breaks the inspection step, since there is no `type=ref,event=branch` in here. The defaults from https://github.com/docker/metadata-action#typeref should be fine here.

bep:
@morremeyer I had to add this commit to make the build green: 4b516c3c6fc671d1204ae60609f67ddf7095682d ... makes sense to you?

I'm doing a Hugo release on either Monday or Tuesday, in which we'll see if the publishing part works ...

.github/workflows/image.yml
```diff
@@ -0,0 +1,123 @@
+name: Build Docker image
+
+on:
+  push:
+    paths:
+      - Dockerfile
+      - .github/workflows/image.yml
```
morremeyer, line 7:
```suggestion
on:
  push:
    tags:
      - '*'
    paths:
      - Dockerfile
      - .github/workflows/image.yml
```

Run for tags, too

morremeyer:
That works, but we might end up with layers/images that are never published, since we build on push of changes to the Dockerfile or the workflow and then never merge these images, if it's not a tag.

My suggestion in https://github.com/gohugoio/hugo/pull/12734#discussion_r1711245074 should fix that while still merging for all relevant events.

I also realized that we need to run for tags, not only for pushes, so I opened a second suggestion for that with https://github.com/gohugoio/hugo/pull/12734#discussion_r1711395507.

bep:
>I also realized that we need to run for tags, not only for pushes,

Good catch, added ...

preechapon-cmd (approved):

PR #12754, opened 2024-08-12
ci: run on all tags

### LINKED ISSUE #12753: The new Docker package workflow not triggered by release tags
https://github.com/gohugoio/hugo/blob/caa90e6a12061893c01bea77030f1e90970cc25c/.github/workflows/image.yml#L6

Not sure what's wrong.

morremeyer:
Run on all tags, check files only for PRs.

Should resolve #12753.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -4,9 +4,11 @@ on:
   push:
     tags:
       - "*"
+  pull_request:
     paths:
       - Dockerfile
       - .github/workflows/image.yml
+
 permissions:
   packages: write
 


bep:
This looks like guessing to me.

We don't want this to be run on `pull_request`. I do understand that I added that line at some point (and removed), but that was a mistake.

So, the intention of the current setup is right:

Build image if either

1. Docker file or image.yml changed
2. A v* tag is created.

What's missing is 2. I could probably force a build by changing the image.yml file, but I doubt it would pick up the tag...

To conclude., we need someone who _knows_.

bep:
And a double note: This isn't me pointing fingers at anyone; I have spent too may hours in the mysteries of GitHub actions to blame anyone of not understanding it to the point ...

theory:
Maybe [this](https://stackoverflow.com/a/77703560/79202)?

``` yaml
on:
  push:
    paths:
      - Dockerfile
      - .github/workflows/image.yml
```

theory:
> I have spent too may hours in the mysteries of GitHub actions to blame anyone of not understanding it to the point ...

I FEEL THIS IN MY BONES.

morremeyer:
@bep so for 1, you want to build it on every push that changes the files, even to master?

bep:
>@bep so for 1, you want to build it on every push that changes the files, even to master?

I don't mind doing that. My take on this is:

* I don't want to build this on every PR because that will be a waste of bot time and storage
* These 2 files changes very infrequently, so I would not mind building then to verify that it "still works"
* But the most important part is release tags on the master branch.

I will do a little investigation on my own on this. I'm doing a patch release later today.

morremeyer:
I agree with that take. 

With the current changes, it should build:

* on every tag
* For every PR that changes the two files

That is as you described it in https://github.com/gohugoio/hugo/pull/12754#issuecomment-2284920836.

The `on.pull_request` is restricted by the `on.pull_request.paths` to only run on PRs that change these paths.

The issue now might be - and I don't have any experience with that so far - that this PR here is coming from my fork, which does not have push access to the registry for the **gohugoio** organization, and rightly so.

I think that might make it impossible to use the matrix workflow (I adapted it from https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners) to speed up builds.

Maybe a better way is to instead:

* not use matrix builds
* use the docker/build-push-action and set `push` to true only for tags

With that, the build itself would be tested, but the image only pushed for tags.

Sorry that this change is causing so many difficulties right now, this is not what I intended.

bep:
>The issue now might be - and I don't have any experience with that so far - that this PR here is coming from my fork, which does not have push access to the registry for the gohugoio organization, and rightly so.

The GITHUB_TOKEN with write access is not available to forks, which is a good thing! I tested this in a PR in a branch in the Hugo repo, and that works.

We will never get a setup that tests this in PRs, and that is fine; as long as we get it to work on the `master` branch, I'm fine.

Let's move this discussion to https://github.com/gohugoio/hugo/pull/12758

PR #12769, opened 2024-08-14
Fix: swap GITHUB_TOKEN for GH_PAT

webgtx:
You can't use default `GITHUB_TOKEN` if you want to push container to ghcr, you need to use [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) instead.

Don't forget to add a new secret, and configure scopes for your personal token.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -102,7 +102,8 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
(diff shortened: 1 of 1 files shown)


bep:
Thanks, but your statement is not true. I have pushed several images with the default token; the problem is that the workflow does not trigger on tags produced by the GitHub release UI.

webgtx:
@bep You're right, sorry for the false PR. If think you need `PAT` only for private registries/repositories

PR #12852, opened 2024-09-15
libsass: Resolve directory paths to directory index files

### LINKED ISSUE #12851: libsass: Unable to resolve directory paths to directory index files
Similar to https://github.com/gohugoio/hugo/issues/12849, but for LibSass.

When implementing https://github.com/gohugoio/hugo/pull/12850 I didn't notice that LibSass 3.6.0 and later support index files with `@import` rules.

See <https://sass-lang.com/documentation/at-rules/import/#index-files>.

jmooring:
Closes #12851


jmooring:
This PR also also sets the order of precedence for both Dart Sass and LibSass to resolve to a file before resolving to a directory. The previous Dart Sass order of precedence was flawed.

Issue #12677, opened 2024-07-25
`--watch` not working with helix-editor


urbantrout:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.129.0+extended darwin/amd64 BuildDate=2024-07-17T13:29:16Z VendorInfo=brew
</pre>

### Steps to reproduce:
1. Run `hugo server` or `hugo --watch`
2. Change files in [Helix](https://helix-editor.com/)

### Expected result
Hugo reacts to file changes and rebuilds pages

### Actual result
Hugo reacts to file changes and builds pages, but not all of them and the output is the same as the one without the changes.

This only happens in [Helix editor](https://helix-editor.com/). In `vim` everything works as expected.

bep:
Could you try to run

```
hugo server --logLevel info
```

And note the lines with `livereload` in them when you change something.

You could also try `hugo server --poll 1s`. If that works, the Helix must doing some non-standard way of saving files that does not trigger the expected fs events.

urbantrout:
This is the output with `--logLevel info`

```
Change detected, rebuilding site (#1).
2024-07-25 16:35:34.441 +0200
Template added /_default/baseof.htmlc3gzwd.bck
Template added /_default/baseof.html
INFO  build:  step process substep resolve content adapter change set changes 2 checked 2 matches 0 duration 36.901µs
INFO  build:  step process substep resolve page output change set changes 2 checked 448 matches 0 duration 3.843363ms
INFO  build:  step process substep gc dynacache duration 196.695µs
INFO  build:  step process substep rebuild templates duration 8.210313ms
INFO  build:  step process duration 13.327644ms
INFO  build:  step assemble duration 777ns
INFO  build:  step render substep pages site de outputFormat html duration 20.021809ms
INFO  build:  step render pages 0 content 0 duration 20.075914ms
INFO  build:  step render deferred count 0 duration 917ns
INFO  build:  step postProcess duration 9.17µs
INFO  build:  duration 33.51625ms
Web Server is available at http://localhost:1313/ (bind address 0.0.0.0)
Total in 33 ms
INFO  livereload: build changed 0 files
```

urbantrout:
adding `--poll 1s` fixes the issue.

urbantrout:
This is not a hugo issue but a helix issue.

bep:
>This is not a hugo issue but a helix issue.

Well, that's debatable.

Looking at these 2 lines:

```
[...]

PR #12856, opened 2024-09-17
Add exclusion for helix .bck files

David-Else:
This should fix https://github.com/gohugoio/hugo/issues/12677, see https://github.com/gohugoio/hugo/issues/12677#issuecomment-2308489571 by @jmooring


bep:
>This should fix #12677


Can someone test this and confirm. I'm not a fan of "I think this should work" when it can be easily tested before we merge it.

David-Else:
I just tested it by building my branch (first time at Go), and it works great, you merged before I had time to get back to you :)

## Before
Screenshot from 2024-09-18 09-53-19

## After
Screenshot from 2024-09-18 09-59-10

PR #12869, opened 2024-09-20
fix js.Build sourcemap comment

BrianLeishman:
https://esbuild.github.io/api/#sourcemap

esbuild sitemap option when set to "external" doesn't actually include the sitemap comment in the file (maybe it used to?). The "linked" option does do this, but I think we want to keep using external here since we want to generate our own source map comment still. 

In this PR I update the code to simply add our source map comment vs trying to replace the existing one (as one doesn't exist.)

I also update the `Sourcefile` esbuild option to be the name of the given file in "assets" so that the source map will specify the actual file, vs just saying "stdin".


BrianLeishman:
Created a small test repo just to showcase a forced console error with the correct file name and line number here: https://github.com/BrianLeishman/hugo-milkdown-esbuild

```shell
git clone https://github.com/BrianLeishman/hugo-milkdown-esbuild --recurse-submodules
cd hugo-milkdown-esbuild
npm ci
hugo server
```

image

bep:
Thanks for this. This _looks_ good, but it has worked before (I remember testing it), and the reason it broke is because we don't have a proper test for it. So, we need that. If you add a test in the `js_integration_test.go`, I will merge.

BrianLeishman:
Sounds good! I did just realize though a few minutes ago that I am using the absolute path "assets" to determine the physical location of the input file, and I don't think that's right, but I'm not familiar enough yet with the hugo source to know how to resolve the real path (relative of the workspace) of a file in "assets".

I'll work on the test and the better file resolution now.

BrianLeishman:
Both of those changes should be good to go!

## Resolve real location of file in assets fs on the os fs

Tested this by moving my "ts" folder to the root of the workspace, and mounting into "assets" view the module mounting config in hugo.toml, getting the correct path, and then testing again by removing the mount and placing directly in assets. 

This was easier than expected since there was an example in the same file.

## Source comment test

With the new comment assertion, the test fails if I undo my changes to the comment output, but passes with this update.

bep:
This should be fixed with my related work in this area. I suspect ESBuild changed their way of doing this; there's now a `linked` option, which I guess is what most people would want. I have added tests ...

BrianLeishman:
I saw! I'm excited to try and switch to this new update from my fork once we get some time. Thank you for all this work, truly amazing.

PR #12872, opened 2024-09-21
tpl: Add bitwise manipulation functions

ElementW:
In evaluating to what system we could migrate the [3dbrew.org wiki](https://3dbrew.org), we are considering using Hugo. Unfortunately it lacks bitwise operators, making template-based formatting of certain fields (like IPC command headers) more difficult than needs be. This PR implements said bitwise operators.

Note I'm not sure what version to use for `{{< new-in >}}`, I assume the next release (0.135.0) is correct. Please correct me if this is wrong. 

---

Adds a `bit` package, available to be used from templates to perform bitwise arithmetic.

The exposed functions/operations are:
- And
- Clear (ANDN)
- Extract (last `l` bits of `n >> s`)
- LeadingZeros
- Not
- OnesCount (popcount)
- Or
- ShiftLeft
- ShiftRight
- TrailingZeros
- Xnor
- Xor

Most of them have short aliases, prefixed with `b` if they would clash with existing functions (like boolean logic functions).

### DIFF
--- a/docs/content/en/functions/bit/And.md
+++ b/docs/content/en/functions/bit/And.md
@@ -1,0 +1,22 @@
+---
+title: bit.And
+description: Bitwise ANDs two or more numbers together.
+categories: []
+keywords: []
+action:
+  aliases: [band]
+  related:
+    - functions/bit/Clear
+    - functions/bit/Or
+    - functions/bit/Xnor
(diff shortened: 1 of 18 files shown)


bep:

Please create an issue with label proposal first.

PR #12901, opened 2024-10-02
tpl/tplimpl: trim descriptions rather than just chomp

chalin:
This PR proposes that we trim whitespace from both ends of descriptions rather than just chomp.

A bit of context for this change: since upgrading to Hugo 0.135.0, opentelemetry.io many page descriptions have been padded with (IMHO unnecessary) whitespace. If the Hugo templates are `chomp`ing the description text, it might as well just `trim` both ends of any whitespace, right? :) 

/cc @svrnm @theletterf

### DIFF
--- a/tpl/tplimpl/embedded/templates/opengraph.html
+++ b/tpl/tplimpl/embedded/templates/opengraph.html
@@ -8,8 +8,8 @@
   <meta property="og:title" content="{{ . }}">
 {{- end }}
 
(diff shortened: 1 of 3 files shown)


theletterf:
Zen and the Art of Hugo Bonsai Pruning! Like it!

jmooring (approved):
LGTM

PR #12903, opened 2024-10-03
Updated Go version to mitigate existing vulnerabilities 

offgrid88:
Updated Go version to mitigate existing vulnerabilities  in the late versions as stated by @floryn90 
to golang:1.23.2
go version 1.23.0 seems to be affected by:

    https://pkg.go.dev/vuln/GO-2024-3106
    https://pkg.go.dev/vuln/GO-2024-3107


bep:
Thanks. I will handle this in #12920

Issue #12885, opened 2024-09-27
Improve Docker image


jmooring:
The purpose of this issue is to solicit community feedback and hopefully reach some consensus.

As of v0.135.0 I can do things like[^1]:

```text
docker run -v .:/site -u $(id -u):$(id -g) ghcr.io/gohugoio/hugo:v0.135.0 config
docker run -v .:/site -u $(id -u):$(id -g) ghcr.io/gohugoio/hugo:v0.135.0 env
docker run -v .:/site -u $(id -u):$(id -g) ghcr.io/gohugoio/hugo:v0.135.0 list all
docker run -v .:/site -u $(id -u):$(id -g) ghcr.io/gohugoio/hugo:v0.135.0 new foo.md
docker run -v .:/site -u $(id -u):$(id -g) ghcr.io/gohugoio/hugo:v0.135.0 server
```

But I can't build the site like this:

```text
docker run -v .:/site -u $(id -u):$(id -g) ghcr.io/gohugoio/hugo:v0.135.0 
```

Instead, I have to pass an empty entry point:

```text
docker run --rm -v .:/site -u $(id -u):$(id -g) --entrypoint "" ghcr.io/gohugoio/hugo:v0.135.0 hugo
```

Or pass in a CLI flag to override the default `--help` flag defined in the Dockerfile. For example:

```text
docker run --rm -v .:/site -u $(id -u):$(id -g) ghcr.io/gohugoio/hugo:v0.135.0 --minify
```

Additionally, the current version:

1. Is not the extended version
2. Does not support Hugo modules
3. Does not support AsciiDoc, Pandoc, or reStructuredText (which is probably fine)
4. Does not support PostCSS
5. Does not support GitInfo
6. Does not write to host cache directory (`cacheDir` / `HUGO_CACHEDIR` / `os.CacheDir`)
7. I'm not sure if css.TailwindCSS works

Now that we can build an image, it would be good to define what we actually want it to do.

Some time ago I created a repository to test docker images:
<https://github.com/jmooring/hugo-feature-test>

Our Docker image doesn't need to pass that test, but we should consider what we want to support.

There's an open issue about how to document usage:
<https://github.com/gohugoio/hugoDocs/issues/2727>

[^1]: The `-u $(id -u):$(id -g)` flag is required to set correct permissions on published files. Without this, owner and group were set to root during my Linux testing.

bep:
[...]

PR #12921, opened 2024-10-09
Misc Docker + Go version upgrade

bep:
As to the Docker image adjustments, I have tested OK:

* [x] css.TailwindCSS
* [x] PostCSS
* [x] Libsass (extended build is now default)
* [x] Dart Sass
* [x] The `hugoDocs` (Go modules, Git etc.) and misc others Go modules setups.
* [x] It sets `ENV HUGO_CACHEDIR=/cache` allowing the Hugo cache to survive builds.

On MacOS, an example build line would be:

```
docker run --rm -v .:/project -v $HOME/Library/Caches/hugo_cache:/cache imagename build
```

Starting the server:

```
docker run --rm -v .:/project -v $HOME/Library/Caches/hugo_cache:/cache imagename server --bind="0.0.0.0"
```

This is the entry point file:

```sh
#!/bin/sh

# Check if a custom hugo-docker-entrypoint.sh file exists.
if [ -f hugo-docker-entrypoint.sh ]; then
  # Execute the custom entrypoint file.
  sh hugo-docker-entrypoint.sh "$@"
  exit $?
fi

# Check if a package.json file exists.
if [ -f package.json ]; then
  # Check if node_modules exists.
  if [ ! -d node_modules ]; then
    # Install npm packages.
    # Note that we deliberately do not use `npm ci` here, as it would fail if the package-lock.json file is not up-to-date,
    # which would be the case if you run the container with a different OS or architecture than the one used to create the package-lock.json file.
    npm i
  fi
fi

exec "hugo" "$@"
```

If a custom `hugo-docker-entrypoint.sh` script exists in the root of the Hugo project, that script will be executed instead of the default entrypoint script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ parameters:
 defaults: &defaults
   resource_class: large
   docker:
-      - image: bepsays/ci-hugoreleaser:1.22300.20000
+      - image: bepsays/ci-hugoreleaser:1.22300.20200
 environment: &buildenv
       GOMODCACHE: /root/project/gomodcache
 version: 2
@@ -60,7 +60,7 @@ jobs:
     environment:
         <<: [*buildenv]
     docker:
-    - image: bepsays/ci-hugoreleaser-linux-arm64:1.22300.20000
+    - image: bepsays/ci-hugoreleaser-linux-arm64:1.22300.20200
     steps:
       - *restore-cache
       - &attach-workspace
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -3,6 +3,7 @@ name: Build Docker image
 on:
   release:
     types: [published]
+  pull_request:
 permissions:
   packages: write
 
@@ -12,18 +13,8 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        platform:
-          - linux/amd64
-          - linux/arm64
-    steps:
-      - name: Prepare
-        run: |
-          platform=${{ matrix.platform }}
-          echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
 
+    steps:
       - name: Checkout
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
@@ -32,9 +23,6 @@ jobs:
         uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81 # v5.5.1
         with:
           images: ${{ env.REGISTRY_IMAGE }}
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf # v3.2.0
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@988b5a0280414f521da01fcc63a27aeeb4b104db # v3.6.1
@@ -47,69 +35,14 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and push by digest
+      - name: Build and push
         id: build
         uses: docker/build-push-action@16ebe778df0e7752d2cfcbd924afdbbd89c1a755 # v6.6.1
         with:
           context: .
-          push: ${{ startsWith(github.ref, 'refs/tags') }}
-          platforms: ${{ matrix.platform }}
+          provenance: mode=max
+          sbom: true
+          push: ${{ github.event_name != 'pull_request' }}
+          platforms: linux/amd64,linux/arm64
+          tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          outputs: type=image,name=${{ env.REGISTRY_IMAGE }},push-by-digest=true,name-canonical=true,push=true
-
-      - name: Export digest
-        run: |
-          mkdir -p /tmp/digests
-          digest="${{ steps.build.outputs.digest }}"
-          touch "/tmp/digests/${digest#sha256:}"
-
-      - name: Upload digest
-        uses: actions/upload-artifact@834a144ee995460fba8ed112a2fc961b36a5ec5a # v4.3.6
-        with:
-          name: digests-${{ env.PLATFORM_PAIR }}
-          path: /tmp/digests/*
-          if-no-files-found: error
-          retention-days: 1
-
-  merge:
-    runs-on: ubuntu-latest
-    needs:
-      - build
-    steps:
-      - name: Download digests
-        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
-        with:
-          path: /tmp/digests
-          pattern: digests-*
-          merge-multiple: true
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@988b5a0280414f521da01fcc63a27aeeb4b104db # v3.6.1
-
-      - name: Docker meta
-        id: meta
-        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81 # v5.5.1
-        with:
-          images: ${{ env.REGISTRY_IMAGE }}
-
-          flavor: |
-            latest=false
-
-      - name: Login to GHCR
-        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Create manifest list and push
-        if: ${{ startsWith(github.ref, 'refs/tags') }}
-        working-directory: /tmp/digests
-        run: |
-          docker buildx imagetools create $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
-            $(printf '${{ env.REGISTRY_IMAGE }}@sha256:%s ' *)
-
-      - name: Inspect image
-        if: ${{ startsWith(github.ref, 'refs/tags') }}
-        run: |
-          docker buildx imagetools inspect ${{ env.REGISTRY_IMAGE }}:${{ steps.meta.outputs.version }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,17 @@
 # Twitter:      https://twitter.com/gohugoio
 # Website:      https://gohugo.io/
 
-FROM golang:1.22.6-alpine AS build
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.5.0 AS xx
+
+FROM --platform=$BUILDPLATFORM golang:1.22.6-alpine AS build
+
+# Set up cross-compilation helpers
+COPY --from=xx / /
+RUN apk add clang lld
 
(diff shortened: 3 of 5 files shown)


bep:
@jmooring I have spent some hours today trying to get up to speed re. the current Docker "scene"; I'm reasonably happy but the `Dockerfile` setup, and it should support all but 3) in your list in #12885.

I wouldn't mind if we as the next step added a `hugo-extras` (or something) with Asciidoc etc. added, but I think it would be good to have **one** base image that covered 90% of the use cases.

On my MacBook Pro M1 the image size increases from 135MB to to 406MB, which I suspect some will complain about, but it's not dramatic enough for me to even consider breaking it up into multiple setups.

What do you think? The entrypoint/npm install setup is a little opinionated, but I think it should work. I have tested it would most of my Hugo projects with a `package.json` and it seems to work fine (even with the new Tailwind). 

What do you think?

dvdksn:
I'm taking a look at the build failure, looks like something related to cross-comp with the new Dockerfile

dvdksn:
OK so here are some changes that I came up with;

- Currently, the `final` image is based on the `base` stage which is pinning the platform to `$BUILDPLATFORM` - we need the final image to use the target platform, so I've decoupled base and final.
  - I added a GO_VERSION arg to make sure they're both using the same golang version.
  - It seems like gcc/g++ are also runtime dependencies for libsass and libwebp, so we need to install them in both the build and final stages.
- I added clang and lld back for the C cross-compilation. I don't know too much about how this works but they're [recommended for use with xx](https://github.com/tonistiigi/xx?tab=readme-ov-file#cc)
- Re-enabled CGO
- Fixed the `install_runtimedeps_default.sh` script to use TARGETARCH instead of BUILDARCH, since we need different versions in the runtime stage.

```diff
diff --git a/Dockerfile b/Dockerfile
index 6da83a3cc..921215c59 100755
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,15 @@
 # Twitter:      https://twitter.com/gohugoio
 # Website:      https://gohugo.io/
 
+ARG GO_VERSION="1.23.2"
+
 FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.5.0 AS xx
-FROM --platform=$BUILDPLATFORM golang:1.23.2-alpine AS base
-FROM base AS build
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS build
 
 # Set up cross-compilation helpers
 COPY --from=xx / /
-
-# gcc/g++ are required to build libsass and libwebp libraries for the extended version.
+RUN apk add clang lld
+ARG TARGETPLATFORM
 RUN xx-apk add --no-scripts --no-cache gcc g++
 
 # Optionally set HUGO_BUILD_TAGS to "none" or "nodeploy" when building like so:
@@ -20,7 +21,7 @@ ARG HUGO_BUILD_TAGS="extended"
 ENV GOPROXY=https://proxy.golang.org
 ENV GOCACHE=/root/.cache/go-build
 ENV GOMODCACHE=/go/pkg/mod
-ARG TARGETPLATFORM
+ENV CGO_ENABLED=1
 
 WORKDIR /go/src/github.com/gohugoio/hugo
 
@@ -36,7 +37,10 @@ RUN --mount=target=. \
     xx-verify /usr/bin/hugo
 EOT
 
-FROM base AS final
+FROM golang:${GO_VERSION}-alpine AS final
+
+# libsass and libwebp also require gcc/g++ as runtime dependencies
+RUN apk add --no-scripts --no-cache gcc g++
 
 COPY --from=build /usr/bin/hugo /usr/bin/hugo
 
@@ -60,8 +64,8 @@ VOLUME /project
 WORKDIR /project
 USER hugo:hugo
 ENV HUGO_CACHEDIR=/cache
-ARG BUILDARCH
-ENV BUILDARCH=${BUILDARCH}
+ARG TARGETARCH
+ENV ARCH=${TARGETARCH/amd64/x64}
 
 COPY scripts/docker scripts/docker
 
diff --git a/scripts/docker/install_runtimedeps_default.sh b/scripts/docker/install_runtimedeps_default.sh
index 0b6c2c617..4ca88fd67 100755
--- a/scripts/docker/install_runtimedeps_default.sh
+++ b/scripts/docker/install_runtimedeps_default.sh
@@ -4,17 +4,10 @@ set -ex
 
 export DART_SASS_VERSION=1.79.3
 
-# If $BUILDARCH=arm64, then we need to install the arm64 version of Dart Sass,
-# otherwise we install the x64 version.
-ARCH="x64"
-if [ "$BUILDARCH" = "arm64" ]; then
-    ARCH="arm64"
-fi
-
 cd /tmp
 curl -LJO https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz 
 ls -ltr
 tar -xf dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz
 rm dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz && \
 # The dart-sass folder is added to the PATH by the caller.
-mv dart-sass /var/hugo/bin
\ No newline at end of file
+mv dart-sass /var/hugo/bin
```

dvdksn:
I noticed you pushed some more changes, and I also noticed that the way dart-sass is installed could be improved, e.g. we install `curl` in the runtime image which isn't very good, we only want the dart-sass binary. So I made installing dart-sass a separate stage and copy the artifact from there. 

```diff
diff --git a/Dockerfile b/Dockerfile
index 6da83a3cc..4455ac8d4 100755
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,15 @@
 # Twitter:      https://twitter.com/gohugoio
 # Website:      https://gohugo.io/
 
+ARG GO_VERSION="1.23.2"
+
 FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.5.0 AS xx
-FROM --platform=$BUILDPLATFORM golang:1.23.2-alpine AS base
-FROM base AS build
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS build
 
 # Set up cross-compilation helpers
 COPY --from=xx / /
-
-# gcc/g++ are required to build libsass and libwebp libraries for the extended version.
+RUN apk add clang lld
+ARG TARGETPLATFORM
 RUN xx-apk add --no-scripts --no-cache gcc g++
 
 # Optionally set HUGO_BUILD_TAGS to "none" or "nodeploy" when building like so:
@@ -20,7 +21,7 @@ ARG HUGO_BUILD_TAGS="extended"
 ENV GOPROXY=https://proxy.golang.org
 ENV GOCACHE=/root/.cache/go-build
 ENV GOMODCACHE=/go/pkg/mod
-ARG TARGETPLATFORM
+ENV CGO_ENABLED=1
 
 WORKDIR /go/src/github.com/gohugoio/hugo
 
@@ -36,48 +37,57 @@ RUN --mount=target=. \
     xx-verify /usr/bin/hugo
 EOT
 
-FROM base AS final
+FROM alpine:latest AS dart-sass
+ARG TARGETARCH
+ARG ARCH=${TARGETARCH/amd64/x64}
+ARG DART_SASS_VERSION=1.79.3
+WORKDIR /tmp
+RUN apk add --no-cache curl
+RUN <<EOT
+    set -ex
+    curl -LJO https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz
+    tar -xf dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz
+EOT
 
-COPY --from=build /usr/bin/hugo /usr/bin/hugo
+FROM golang:${GO_VERSION}-alpine AS final
+
+# libsass and libwebp also require gcc/g++ as runtime dependencies
+RUN apk add --no-scripts --no-cache
 
 # libc6-compat & libstdc++ are required for extended libraries (libsass, libwebp).
 RUN apk add --no-cache \
+    g++ \
+    gcc \
+    git \
     libc6-compat \
     libstdc++ \
-    git \
-    runuser \
-    curl \
     nodejs \
-    npm
+    npm \
+    runuser
 
-RUN mkdir -p /var/hugo/bin && \
-    addgroup -Sg 1000 hugo && \
-    adduser -Sg hugo -u 1000 -h /var/hugo hugo && \
-    chown -R hugo: /var/hugo && \
-    runuser -u hugo -- git config --global --add safe.directory /project
+COPY --from=dart-sass /tmp/dart-sass /var/hugo/bin/dart-sass
+COPY --from=build /usr/bin/hugo /usr/bin/hugo
+ENV PATH="/var/hugo/bin:/var/hugo/bin/dart-sass:$PATH"
 
+RUN <<EOT
+    mkdir -p /var/hugo/bin /cache
+    addgroup -Sg 1000 hugo
+    adduser -Sg hugo -u 1000 -h /var/hugo hugo
+    chown -R hugo: /var/hugo /cache
+    # For the Hugo's Git integration to work.
+    runuser -u hugo -- git config --global --add safe.directory /project && \ 
+    # See https://github.com/gohugoio/hugo/issues/9810
+    runuser -u hugo -- git config --global core.quotepath false
+EOT
+
+USER hugo:hugo
 VOLUME /project
 WORKDIR /project
-USER hugo:hugo
 ENV HUGO_CACHEDIR=/cache
-ARG BUILDARCH
-ENV BUILDARCH=${BUILDARCH}
-
-COPY scripts/docker scripts/docker
-
-# Install default dependencies.
-RUN scripts/docker/install_runtimedeps_default.sh
-
-COPY scripts/docker/entrypoint.sh /entrypoint.sh
-
-ENV PATH="/var/hugo/bin:/var/hugo/bin/dart-sass:$PATH"
 
 RUN sass --version
 RUN hugo version
 
-# Expose port for live server
-EXPOSE 1313
-
+COPY scripts/docker/entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
 CMD ["--help"]
-
diff --git a/scripts/docker/install_runtimedeps_default.sh b/scripts/docker/install_runtimedeps_default.sh
deleted file mode 100755
index 0b6c2c617..000000000
--- a/scripts/docker/install_runtimedeps_default.sh
+++ /dev/null
@@ -1,20 +0,0 @@
-#!/bin/sh
-
-set -ex
-
-export DART_SASS_VERSION=1.79.3
-
-# If $BUILDARCH=arm64, then we need to install the arm64 version of Dart Sass,
-# otherwise we install the x64 version.
-ARCH="x64"
-if [ "$BUILDARCH" = "arm64" ]; then
-    ARCH="arm64"
-fi
-
-cd /tmp
-curl -LJO https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz 
-ls -ltr
-tar -xf dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz
-rm dart-sass-${DART_SASS_VERSION}-linux-${ARCH}.tar.gz && \
-# The dart-sass folder is added to the PATH by the caller.
-mv dart-sass /var/hugo/bin
\ No newline at end of file
```

bep:
@dvdksn thanks for the input; one of my goals of this work is to remove (at least most of) the guess work from the Docker setup (e.g. "It seems like gcc/g++ are also runtime dependencies for libsass and libwebp"; we need a libc implementation, but not the full grown compiler).

dvdksn:
OK, then I think you can just remove

```
    g++ \
    gcc \
```

from the Dockerfile above and it should work

bep:
OK, I guess I need to pull out my non-ARM macbook to get this up and running instead of guessing ...

jmooring:
@bep Support for asciidoc, pandoc, and rst can wait for another day. I think this covers > 95% of use cases. I don't have any opinion on size... I suppose if someone wants something thinner they can grow their own, as many have already done.

PR #12927, opened 2024-10-12
fixed #12926 uglyurls parse to map[string]bool

### LINKED ISSUE #12926: uglyurls not support map[string]bool
uglyurls support is configured as map[string]bool according to the following link 

- https://github.com/gohugoio/hugo/blob/9b635522e28e507c49e558e537157c932398981c/config/allconfig/allconfig.go#L335-L344 
- https://github.com/gohugoio/hugo/blob/9b635522e28e507c49e558e537157c932398981c/config/allconfig/alldecoders.go#L413-L428

but not work

because v is `maps.Params[_merge:none post:true]` in `p.c.UglyURLs = cast.ToStringMapBool(v)`, `none` should be clean

akkuman:
fix https://github.com/gohugoio/hugo/issues/12926

### DIFF
--- a/config/allconfig/alldecoders.go
+++ b/config/allconfig/alldecoders.go
@@ -419,6 +419,8 @@ var allDecoderSetups = map[string]decodeWeight{
 				p.c.UglyURLs = vv
 			case string:
 				p.c.UglyURLs = vv == "true"
+			case maps.Params:
+				p.c.UglyURLs = cast.ToStringMapBool(maps.CleanConfigStringMap(vv))
 			default:
 				p.c.UglyURLs = cast.ToStringMapBool(v)
 			}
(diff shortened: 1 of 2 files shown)


config/allconfig/alldecoders.go
```diff
@@ -419,6 +419,8 @@ var allDecoderSetups = map[string]decodeWeight{
 				p.c.UglyURLs = vv
 			case string:
 				p.c.UglyURLs = vv == "true"
+			case maps.Params:
+				p.c.UglyURLs = cast.ToStringMapBool(maps.CleanConfigStringMap(vv))
```
bep, line 423: We need a failing test case for this.

Also, I suspect you could handle this case in the `default`, but start with a test.

akkuman: @bep I've added test cases

PR #12947, opened 2024-10-15
create/skeletons: Add delimiters to archetype front matter

### LINKED ISSUE #12945: TOML and YAML archetypes missing delimiters
JSON archetypes are fine.

This is an embarrassing bug, introduced in https://github.com/gohugoio/hugo/pull/12928, which would have been caught had I created some decent test cases. 

We'll need a patch release on this one.

jmooring:
Fixes #12945

### DIFF
--- a/create/skeletons/skeletons.go
+++ b/create/skeletons/skeletons.go
@@ -123,7 +123,7 @@ func newSiteCreateArchetype(fs afero.Fs, createpath string, format string) (err 
 	}
 
 	var buf bytes.Buffer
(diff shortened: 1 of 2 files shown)


bep:
There's a Windows test failure. I think you can find a. `unixtodos` command in some of the golden text scripts that makes this ... work.

jmooring:
Just saw your previous comment. We now skip the comparison to golden on Windows. Let me know if that's not acceptable, but the newline difference is, in my opinion, immaterial to the underlying bug.

bep:
Thanks ... I'll do a patch release tomorrow.

PR #12959, opened 2024-10-17
docker: fix dart-sass arch mismatch, /cache permissions

### LINKED ISSUE #12956: Check cache dir creation permissions vs Docker image
https://github.com/gohugoio/hugoDocs/pull/2741

>Error: failed to load modules: failed to download modules: failed to execute 'go [mod download -modcacherw]': failed to execute binary "go" with args [mod download -modcacherw]: go: github.com/jmooring/hugo-content@v0.1.0: mkdir /cache/modules: permission denied
*errors.errorString

### LINKED ISSUE #12957: Dockerfile Dart Sass permission issue on MacOS
```
Watching for changes in /project/{archetypes,assets,content,layouts,package.json,postcss.config.js,static}
Watching for config changes in /project/hugo.toml, /project/config/_default, /project/go.mod
Start building sites … 
hugo v0.136.0+extended linux/arm64 BuildDate=unknown VendorInfo=docker

ERROR TOCSS-DART: failed to transform "/sass/main.scss" (text/x-scss): got unexpected EOF when executing "sass". The user running hugo must have read and execute permissions on this program. With execute permissions only, this error is thrown.
ERROR render: failed to render pages: render of "page" failed: "/project/layouts/_default/dart-sass.html:25:23": execute of template failed: template: _default/dart-sass.html:25:23: executing "main" at <.Content>: error calling Content: TOCSS-DART: failed to transform "/temp/sass" (application/octet-stream): "<stream>:1:1": connection is shut down
Built in 593 ms
Error: error building site: TOCSS-DART: failed to transform "/temp/sass" (application/octet-stream): "<stream>:1:1": connection is shut down
````

### PR DESCRIPTION
**Pull request by dvdksn**
- Closes #12956
- Closes #12957

The dart-sass setup used the `$BUILDARCH` variable to fetch the dart-sass
binary, which resulted in the wrong architecture being downloaded when building
for another architecture. The first commit changes it to use `$TARGETARCH`
instead. It also simplifies/inlines the dart-sass setup.

The second commit fixes the permission issue with `/cache` being owned by root.

### CODE REVIEW

**Comment by dvdksn** [comment]
@bep can you try building a dart-sass project with this version? I pushed an image of this build to `davidkarlsson416/hugo:12959`

```bash
docker run -v .:/project -p 1313:1313 davidkarlsson416/hugo:12959
```

Dockerfile
```diff
@@ -45,6 +45,15 @@ RUN --mount=target=. \
     xx-verify /usr/bin/hugo
 EOT
 
+# dart-sass downloads the dart-sass runtime dependency
+FROM alpine:${ALPINE_VERSION} AS dart-sass
+ARG TARGETARCH
+ARG DART_SASS_VERSION=1.79.3
```
bep, line 51: Can you make these Dart `ARG` vars into ENV. I understand the ... flexibility, but it just adds work to the documentation/testing pile.

Other than that this looks/works great.

bep:
>docker run -v .:/project davidkarlsson416/hugo:12959 build

Works great, thanks.

dvdksn:
@bep I saw your comment about `ARG`/`ENV`. I pushed a small edit to the first commit to make DART_SASS_VERSION work like the other version args. It makes it easier to maintain, and I like to stick with `ENV` only when I actually want to set an env var in the runtime container or use it to configure a build tool, like `GOOS`. Hope it makes sense.

bep:
>t makes it easier to maintain, and I like to stick with ENV only when I actually want to set an env var in the runtime container or use it to configure a build tool, like GOOS. Hope it makes sense.

@dvdksn yea, I removed that comment as I saw that we used ARG for the go versions. We can handle this later, but your argument only makes sense if Docker had a concept of "private ARGs". 

As I understand it:

* ARG can be set in argument to `docker build` and is such part of the "build API"
* ENV is visible in the OS environment of the current shell, but not from the "outside" (e.g. not when running `hugo`,  not as arguments to `docker build`).

dvdksn:
Yes I get it, the ability to set these args on the command line is not desirable. But `ARG` is the better way to go for build-time variables, even if you don't plan on overriding them.

bep:
>But ARG is the better way to go for build-time variables, even if you don't plan on overriding them.

How is it better? My argument is above is that the only practical difference in the "dart sass version" case above is that ARG can be set from the outside, which is a negative; what are the positive sides of ARG that outweighs the negative?

dvdksn:
In this specific case, no real benefit; it's just more idiomatic. I think it's better to use `ENV` if you explicitly want an environment variable to configure a build step, or if you want to set an environment variable that persists in the final image and in containers created from it. And use `ARG` for variables that are only used to influence build execution.

jmooring:
<https://www.docker.com/blog/docker-best-practices-using-arg-and-env-in-your-dockerfiles/#tldr>

Also note:

> Both ARG and ENV can be overridden from the command line in docker run and docker compose

bep:
>Both ARG and ENV can be overridden from the command line in docker run and docker compose

I have not found any example/documentation about how to pass ENV variables from the command line to `docker build`.

dvdksn:
> I have not found any example/documentation about how to pass ENV variables from the command line to docker build.

You can't. You can't change the value of `ENV` outside the Dockerfile itself (unless you use `ENV foo=$ARG`...)

jmooring:
Elsewhere in that article the author provides this example (which I have not tested):

```text
$ docker run --rm -e THEENV=bar envtest
```

bep:
> docker run --rm -e THEENV=bar envtest

Sure, but the discussion above is about `docker build` and I see no `-e` option there.

jmooring:
Gotcha. When reading the article I didn't notice that `docker build` was excluded.

PR #12966, opened 2024-10-19
resources: Address Dart Sass deprecation of global built-in functions

### LINKED ISSUE #12961: Dart Sass v1.80.1 deprecation warning
Testing Dart Sass v1.80.1 

> WARN  Dart Sass: hugo:vars:0:17: Global built-in functions are deprecated and will be removed in Dart Sass 3.0.0.
Use string.unquote instead.

Why we're getting this is unclear at the moment.

jmooring:
See https://github.com/sass/dart-sass/releases/tag/1.80.0

Fixes #12961


jmooring:
<https://github.com/gohugoio/hugo/issues/12961#issuecomment-2420360194>

> I guess adding some kind of identifier as the first arg to CreateVarsStyleSheet would solve it.

To avoid hardcoded the identifiers, I moved the constant declarations from the `tpl/css` package to the `sass` package, which was previously an "internal" package. I had to make it "not internal" in order to export the constants.

Issue #12971, opened 2024-10-20
docker: npm errors (windows)


jmooring:
This is a regression from v0.136.1.

<details>
<summary>error log</summary>

```text
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/nanoid/non-secure/index.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/postcss/lib/list.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/fastq/test/example.ts'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/nanoid/url-alphabet/index.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/postcss/lib/map-generator.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/postcss/lib/no-work-result.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/@nodelib/fs.stat/out/providers/sync.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/@nodelib/fs.walk/out/readers/reader.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/nanoid/non-secure/package.json'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/postcss/lib/node.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/@nodelib/fs.scandir/out/providers/sync.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/nanoid/url-alphabet/package.json'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/postcss/lib/parse.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/fast-glob/out/utils/errno.js'
npm warn tar TAR_ENTRY_ERROR ENOENT: no such file or directory, open '/project/node_modules/postcss/lib/parser.js'
[...]

Issue #12970, opened 2024-10-20
docker: npm errors (linux)


jmooring:
This is a regression from v0.136.1.

<details>
<summary>error log</summary>

```text
npm error code EACCES
npm error syscall mkdir
npm error path /var/hugo/.npm
npm error errno -13
npm error
npm error Your cache folder contains root-owned files, due to a bug in
npm error previous versions of npm which has since been addressed.
npm error
npm error To permanently fix this problem, please run:
npm error   sudo chown -R 1000:1000 "/var/hugo/.npm"
npm error Log files were not written due to an error writing to the directory: /var/hugo/.npm/_logs
npm error You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
Watching for changes in /project/{archetypes,assets,content,layouts,package.json,postcss.config.js,static}
Watching for config changes in /project/config/_default, /project/go.mod
Start building sites … 
hugo v0.136.2+extended linux/amd64 BuildDate=unknown VendorInfo=docker

ERROR render of "page" failed: "/project/layouts/_default/tailwindcss.html:16:23": execute of template failed: template: _default/tailwindcss.html:16:23: executing "main" at <.Content>: error calling Content: TAILWINDCSS: failed to transform "/temp/css" (application/octet-stream): npm error code EACCES
npm error syscall mkdir
npm error path /var/hugo/.npm
npm error errno EACCES
npm error
npm error Your cache folder contains root-owned files, due to a bug in
npm error previous versions of npm which has since been addressed.
npm error
npm error To permanently fix this problem, please run:
npm error   sudo chown -R 1000:1000 "/var/hugo/.npm"
npm error Log files were not written due to an error writing to the directory: /var/hugo/.npm/_logs
npm error You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
ERROR POSTCSS: failed to transform "/temp/css" (application/octet-stream): npm error code EACCES
npm error syscall mkdir
npm error path /var/hugo/.npm
npm error errno EACCES
npm error
npm error Your cache folder contains root-owned files, due to a bug in
npm error previous versions of npm which has since been addressed.
npm error
npm error To permanently fix this problem, please run:
npm error   sudo chown -R 1000:1000 "/var/hugo/.npm"
npm error Log files were not written due to an error writing to the directory: /var/hugo/.npm/_logs
npm error You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
[...]

PR #12975, opened 2024-10-21
docker: fix permission issues in Dockerfile

dvdksn:
I think the `--link` option might've caused a permission issue for the non-root `hugo` user.

Related issues:

- #12970
- #12971

Also noticed that the `/project` workdir was created before we assumed the `hugo:hugo` user, which resulted in that directory being owned by `root`. I reordered the instructions to make sure `/project` is owned by the correct user.


dvdksn:
@jmooring @bep could you try with this build? I ran the reproducer in the linked issue (on macos) and it seems to work for me.

bep:
Works for me.

bep:
@dvdksn could you get the commit message(s) in line with the guidelines?

PR #13011, opened 2024-11-04
tpl/tplimpl: Create an embedded comment shortcode

### LINKED ISSUE #13010: Embedded "comment" shortcode
Although I dislike embedding opinionated shortcodes, given the changes in v0.137.0 (raw HTML warnings) a comment shortcode might make sense:

```text
{{- $noop := .Inner -}}
```

Obviously this is not opinionated.

I'm in the process of writing a tips and trick article for the forum, and part of that document explains what to do instead of using HTML comments in Markdown. So instead of each site and theme author having to reinvent the wheel, perhaps we should embed one.

jmooring:
Closes #13010


tpl/tplimpl/tplimpl_integration_test.go
```diff
@@ -584,3 +584,19 @@ title: p5
...
+func TestCommentShortcode(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+disableKinds = ['page','rss','section','sitemap','taxonomy','term']
+-- layouts/index.html --
+{{ .Content }}
+-- content/_index.md --
+a{{< comment >}}b{{< /comment >}}c
```
bep, line 597: This is fine for the test, but in the docs/examples I would recommend that we use the `{{%` separator. It would make not no difference for the end result, but would save a few CPU cycles ...

jmooring: Understood. I'll leave the test as-is.

PR #13100, opened 2024-12-01
tpl/tplimpl: Add details shortcode

### LINKED ISSUE #13090: Add Details Shortcode for collapsable sections within markdown content
Hello, 

I would like to propose adding a new shortcode: 
- tpl/tplimpl/embedded/templates/shortcodes/details.html

The Details shortcode would allow users to use the [Details html element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) within their markdown files. I realize this could already be done by using a rawhtml shortcode but I believe this is a common enough element to warrant inclusion in the project. So common that [github implemented it in their markdown](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections) which lets me show you what I'm talking about directly within this proposal: 

<details>

<summary>Details</summary>

### You can add a header

You can add text within a collapsed section. 

You can add an image or a code block, too.

```ruby
   puts "Hello World"
```

</details>

To implement the details element, I have developed the following code:
```js
{{ $summary := .Get "summary" | default (.Get 0) | default "Details:" | markdownify }}
{{ $openParam := .Get "open" | default (.Get 1) | default false }}
{{ $isOpen := not (or (eq $openParam false) (and (eq (printf "%T" $openParam) "string") (eq (lower $openParam) "false"))) }}
{{ $altSummary := .Get "altSummary" | default (.Get 2) | default $summary | markdownify }}
{{ $name := .Get "name" | default (.Get 3) | default "" }}
<details {{ if $isOpen }}open{{ end }}{{ with $name }} name="{{ . }}"{{ end }}>
    <summary onclick="this.innerHTML = this.parentNode.open ? '{{ $summary }}' : '{{ $altSummary }}';">
        {{ if $isOpen }}
            {{ $altSummary }}
        {{ else }}
            {{ $summary }}
        {{ end }}
    </summary>
    {{ .Inner | markdownify }}
</details>
```

The code has 4 arguments that can be passed:  
1. `summary` (str). Default = "Details:". This is the text that appears for users to click on.
2. `open` (bool). Default = `false`. If this argument is specified, and a value other than `false` (bool) or "false" (str) is provided, then the element will be expanded upon page load.  
3. `altSummary` (str). Default = `summary`. This is the text that appears when the details tag is open. If not provided it will be the same as the summary text.
4. `name` (str). Default = "" (empty string). This attribute allows you to connect multiple details elements so that only one can be open at a time. If a user opens a named element then all other same-named elements will be closed. 

Notes:
- The arguments are named and positional. 
- The details shortcode generates a standard HTML details element that works for basic  expand/collapse  functionality without additional CSS or JS. The shortcode does include a javascript `onclick` handler for the summary/altSummary text switching feature but the code is self-contained to the shortcode's .html file. 
- The `open` behavior matches [html's implementation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#open), with the exception it also accepts "false" as a str to keep the element closed on page load. 
- the `name` argument is an html argument and the shortcode does not deviate from the expected behavior. 
- The `altSummary` is an added feature over existing html behavior. I added this since I wanted to express when the details elements were expanded/collapsed, such as toggling between "10 cells collapsed" and "10 cells expanded". 

Discourse [29546](https://discourse.gohugo.io/t/detail-html-tag/29546) has had 2.2k views in a few years demonstrating it is a popular query, and my proposal has additional features to align with html's implementation. 

I cannot fully demonstrate my shortcode within github discussion. If you would like to see examples of its usage, You can see the usage of the details shortcode within hugo in my [blog post here, ](https://racedorsey.com/posts/2024/hugo-details-shortcode/). 

This is would be my first contribution to the project. I have signed the CLA and am prepared to update documentation if my proposal is approved. Thanks for your time!

racehd:
- Add new shortcode to render details HTML element.
- Implement integration tests to check: default state, custom summary, open state, safeHTML sanitization, allowed attributes, and localization of default summary text.
- Update docs to include details shortcode.

Closes #13090

### DIFF
--- a/docs/content/en/content-management/shortcodes.md
+++ b/docs/content/en/content-management/shortcodes.md
@@ -94,6 +94,40 @@ Example usage:
 
 Although you can call this shortcode using the `{{</* */>}}` notation, computationally it is more efficient to call it using the `{{%/* */%}}` notation as shown above.
 
+### details
+
+{{< new-in 0.140.0 >}}
+
+{{% note %}}
+To override Hugo's embedded `details` shortcode, copy the [source code] to a file with the same name in the layouts/shortcodes directory.
+
+This may be useful if you are wanting access to more global HTML attributes. 
+
+[source code]: {{% eturl details %}}
+{{% /note %}}
+
+Use the `details` shortcode to generate a collapsible details HTML element. For example:
+
+```text
+{{</* details summary="Custom Summary Text" */>}}
+Showing custom `summary` text.
+{{</* /details */>}}
+```
+
+Additional examples can be found in the source code. The `details` shortcode can use the following named arguments:
+
+summary
+: (`string`) Optional. Specifies the content of the child summary element. Default is "Details"
+
+open
+: (`bool`) Optional. Whether to initially display the contents of the details element. Default is `false`.
+
+name
+: (`string`) Optional. The value of the element's name attribute.
+
+class
+: (`string`) Optional. The value of the element's class attribute.
+
 ### figure
 
 {{% note %}}
--- a/tpl/tplimpl/embedded/templates/shortcodes/details.html
+++ b/tpl/tplimpl/embedded/templates/shortcodes/details.html
@@ -1,0 +1,68 @@
+{{- /*
+Renders an HTML details element.
+
+@param {string} [summary] The content of the child summary element.
+@param {bool} [open=false] Whether to initially display the contents of the details element.
+@param {string} [class] The value of the element's class attribute.
+@param {string} [name] The value of the element's name attribute.
+
+@reference https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details
+
+@examples
+
+    {{< details >}}
+    A basic collapsible section.
+    {{< /details >}}
+
+    {{< details summary="Custom Summary Text" >}}
+    Showing custom `summary` text.
+    {{< /details >}}
+
+    {{< details summary="Open Details" open=true >}}
+    Contents displayed initially by using `open`.
+    {{< /details >}}
+
+    {{< details summary="Styled Content" class="my-custom-class" >}}
+    Content can be styled with CSS by specifying a `class`.
+
+    Target details element:
+
+    ```css
+    details.my-custom-class { }
+    ```
+
+    Target summary element:
+
+    ```css
+    details.my-custom-class > summary > * { }
+    ```
+
+    Target inner content:
+
+    ```css
+    details.my-custom-class > :not(summary) { }
+    ```
+    {{< /details >}}
+
+    {{< details summary="Grouped Details" name="my-details" >}}
+    Specifying a `name` allows elements to be connected, with only one able to be open at a time.
+    {{< /details >}}
+
+*/}}
+
+{{- /* Get arguments. */}}
+{{- $summary := or (.Get "summary") (T "shortcodes.details") "Details" }}
+{{- $class := or (.Get "class") "" }}
+{{- $name := or (.Get "name") "" }}
+{{- $open := false }}
+{{- if in (slice "false" false 0) (.Get "open") }}
+    {{- $open = false }}
+{{- else if in (slice "true" true 1) (.Get "open")}}
+    {{- $open = true }}
+{{- end }}
+
+{{- /* Render. */}}
+<details{{- if $open }} open{{ end }}{{- if $name }} name="{{ $name }}"{{- end }}{{- if $class }} class="{{ $class }}"{{- end }}>
+    <summary>{{ $summary | .Page.RenderString }}</summary>
+    {{ .Inner | .Page.RenderString (dict "display" "block") -}}
+</details>
--- a/docs/data/embedded_template_urls.toml
+++ b/docs/data/embedded_template_urls.toml
@@ -25,6 +25,7 @@
 
 # Shortcodes
 'comment' = 'shortcodes/comment.html'
+'details' = 'shortcodes/details.html'
 'figure' = 'shortcodes/figure.html'
 'gist' = 'shortcodes/gist.html'
 'highlight' = 'shortcodes/highlight.html'
--- a/tpl/tplimpl/tplimpl_integration_test.go
+++ b/tpl/tplimpl/tplimpl_integration_test.go
@@ -600,3 +600,118 @@ a{{< comment >}}b{{< /comment >}}c
 	b := hugolib.Test(t, files)
 	b.AssertFileContent("public/index.html", "<p>ac</p>")
 }
+
+func TestDetailsShortcode(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+disableKinds = ['rss','section','sitemap','taxonomy','term']
+defaultContentLanguage = "en"
+[languages]
+  [languages.en]
+    weight = 1
+  [languages.es]
+    weight = 2
+-- i18n/en.toml --
+[shortcodes.details]
+other = "Details"
+-- i18n/es.toml --
+[shortcodes.details]
+other = "Detalles"
+-- layouts/_default/single.html --
+{{ .Content }}
+-- content/d1.md --
+---
+title: Default State Test
+---
+{{< details >}}
+Basic example without summary
+{{< /details >}}
+-- content/d2.md --
+---
+title: Custom Summary Test
+---
+{{< details summary="Custom Summary" >}}
+Example with custom summary text
+{{< /details >}}
+-- content/d3.md --
+---
+title: Open State Test
+---
+{{< details summary="Test Open State" open="true" >}}
+Example with open state
+{{< /details >}}
+-- content/d4.md --
+---
+title: Attributes Test
+---
+{{< details summary="Test Attribute sanitization" style="color: red; font-weight: bold; background-color: #eee" onclick="alert('test')" >}}
+Example testing attribute sanitization
+{{< /details >}}
+-- content/d5.md --
+---
+title: Class Test
+---
+{{< details class="custom-class" >}}
+Example with allowed class attribute
+{{< /details >}}
+-- content/d6.md --
+---
+title: Name Test
+---
+{{< details name="custom-name" >}}
+Example with allowed name attribute
+{{< /details >}}
+-- content/d7.es.md --
+---
+title: Localization Test
+---
+{{< details >}}
+Localization example without summary
+{{< /details >}}
+`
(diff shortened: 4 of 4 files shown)


docs/content/en/content-management/shortcodes.md
```diff
@@ -94,6 +94,33 @@ Example usage:
...
+{{% note %}}
+To override Hugo's embedded `details` shortcode, copy the [source code] to a file with the same name in the layouts/shortcodes directory.
+
+[source code]: {{% eturl details %}}
+{{% /note %}}
+
+The `details` shortcode allows you to generate a collapsible details HTML element. Example usage:
+
+```text
+    {{</* details summary="Custom Summary Text" */>}}
```
jmooring, line 108: Do not indent example

docs/content/en/content-management/shortcodes.md
```diff
@@ -94,6 +94,33 @@ Example usage:
 
 Although you can call this shortcode using the `{{</* */>}}` notation, computationally it is more efficient to call it using the `{{%/* */%}}` notation as shown above.
 
+### details
+
+{{% note %}}
+To override Hugo's embedded `details` shortcode, copy the [source code] to a file with the same name in the layouts/shortcodes directory.
+
+[source code]: {{% eturl details %}}
```
jmooring, line 102: Add the following to data/embedded_template_urls.toml:

```text
'details' = 'shortcodes/details.html'
```

docs/content/en/content-management/shortcodes.md
```diff
@@ -94,6 +94,33 @@ Example usage:
...
+
+Additional examples can be found in source code. The `details` shortcode can use the following named arguments:
+
+summary
+: (`string`) Optional. Specifies the content of the child summary element. Default is "Details"
+
+open
+: (`bool`) Optional. Whether to initially display the contents of the details element. Default is `false`.
+
+[...params]
```
jmooring, line 121:
```suggestion
```

docs/content/en/content-management/shortcodes.md
```diff
@@ -94,6 +94,33 @@ Example usage:
...
+: (`object`) Optional. Additional HTML attributes passed directly to the details element. For instance, can specify `name`, `class`, and other global HTML attributes.
```
jmooring, line 122:
~~~suggestion
You can also pass global HTML attributes when calling the shortcode. For example:

```text
{{</* details summary="Custom Summary Text" class="my-class" */>}}
Showing custom `summary` text.
{{</* /details */>}}
```
~~~

docs/content/en/content-management/shortcodes.md
```diff
@@ -94,6 +94,33 @@ Example usage:
...
 
+### details
+
+{{% note %}}
+To override Hugo's embedded `details` shortcode, copy the [source code] to a file with the same name in the layouts/shortcodes directory.
+
+[source code]: {{% eturl details %}}
+{{% /note %}}
+
+The `details` shortcode allows you to generate a collapsible details HTML element. Example usage:
```
jmooring, line 105:
```suggestion
Use the `details` shortcode to generate a collapsible details HTML element. For example:
```

docs/content/en/content-management/shortcodes.md
jmooring, line 121: This isn't a named parameter. See below.

jmooring (changes requested):

jmooring (commented):
The AssertFileContent method iterates over the values and looks for a match. So it tests if the result contains the value, but it does not test if they are in the given order. I would try to combine these into a single string, possibly using `AssertFileContentEquals` instead.

Also, I think our convention with multiple values is to place the  closing parenthesis on its own line, e.g., 

```text
	b.AssertFileContent("foo",
		"bar",
		"baz",
	)

**Comment by racehd** [comment]
> The AssertFileContent method iterates over the values and looks for a match. So it tests if the result contains the value, but it does not test if they are in the given order. I would try to combine these into a single string, possibly using `AssertFileContentEquals` instead.

Thank you for the review! I had noticed AssertFileContent did not check the order. I previously had tried AssertFileContentEquals but it will also check for newlines and whitespace which I thought may make it more prone to triggering false positives in the future. 

The following would pass the checks: 
<details><summary>AssertFileContentEquals</summary>

```
	// Test1: default state (closed by default)
	b.AssertFileContentEquals("public/d1/index.html",
	"\n<details>\n    <summary>Details</summary>\n    <p>Basic example without summary</p>\n</details>\n",
	)
	content1 := b.FileContent("public/d1/index.html")
	c := qt.New(t)
	c.Assert(content1, qt.Not(qt.Contains), "open")

	// Test2: custom summary
	b.AssertFileContentEquals("public/d2/index.html",
	"\n<details>\n    <summary>Custom Summary</summary>\n    <p>Example with custom summary text</p>\n</details>\n",
	)

	// Test3: open state
	b.AssertFileContentEquals("public/d3/index.html",
	"\n<details open>\n    <summary>Test Open State</summary>\n    <p>Example with open state</p>\n</details>\n",
	)

	// Test4: safeHTML sanitization
	b.AssertFileContentEquals("public/d4/index.html",
	"\n<details style=\"color: red;\">\n    <summary>Test Attribute sanitization</summary>\n    <p>Example testing attribute sanitization</p>\n</details>\n",
	)
	content4 := b.FileContent("public/d4/index.html")
	c.Assert(content4, qt.Not(qt.Contains), "onclick")
	c.Assert(content4, qt.Not(qt.Contains), "alert")

	// Test5: class attribute
	b.AssertFileContentEquals("public/d5/index.html",
	"\n<details class=\"custom-class\">\n    <summary>Details</summary>\n    <p>Example with allowed class attribute</p>\n</details>\n",
	)

	// Test6: localization
	b.AssertFileContentEquals("public/es/d6/index.html",
	"\n<details>\n    <summary>Detalles</summary>\n    <p>Localization example without summary</p>\n</details>\n",
	)
```

</details>

Alternatively, I could write a helper function to ignore whitespace and make it a little more robust. Do you have a preference on this?

**Comment by jmooring** [comment]
> Do you have a preference on this?

I think what you have is fine.

File: tpl/tplimpl/embedded/templates/shortcodes/details.html [hunk 6]
```diff
@@ -0,0 +1,72 @@
...
+{{- else if in (slice "true" true 1) (.Get "open")}}
+    {{- $open = true }}
+{{- end }}
+
+{{- /* Render. */}}
+<details
+    {{- if $open }} open{{ end }}
+    {{- range $k, $v := .Params }}
+        {{- if not (or (in (slice "open" "summary") $k) (strings.HasPrefix $k "on")) }}
+        {{- printf " %s=%q" $k $v | safeHTMLAttr }}
```
jmooring, line 66:
```suggestion
            {{- printf " %s=%q" $k $v | safeHTMLAttr }}
```

docs/content/en/content-management/shortcodes.md
```diff
@@ -94,6 +94,38 @@ Example usage:
 
 Although you can call this shortcode using the `{{</* */>}}` notation, computationally it is more efficient to call it using the `{{%/* */%}}` notation as shown above.
 
+### details
+
```
jmooring, line 98:
```suggestion

{{< new-in 0.140.0 >}}

```

docs/content/en/content-management/shortcodes.md
```diff
@@ -94,6 +94,38 @@ Example usage:
...
+Use the `details` shortcode to generate a collapsible details HTML element. For example:
+
+```text
+{{</* details summary="Custom Summary Text" */>}}
+Showing custom `summary` text.
+{{</* /details */>}}
+```
+
+Additional examples can be found in source code. The `details` shortcode can use the following named arguments:
```
jmooring, line 113:
```suggestion
Additional examples can be found in the source code. The `details` shortcode can use the following named arguments:
```

jmooring (changes requested):
Just three minor changes.

jmooring:
Also, please squash commits.

jmooring:
Please clean up the commit message and force push. It should look something like this:

```text
- Add new shortcode to render details HTML element.
- Implement integration tests to check: default state, custom summary, open state, safeHTML sanitization, allowed attributes, and localization of default summary text.
- Update docs to include details shortcode.

Closes # 13090
```

You don't need all the messages from the squashed commits.

racehd:
Got it, thanks again

jmooring (approved):

jmooring:
This looks and works great! I recommend we merge this after the tests go green.

bep:
Two comments from me:

We cannot have this:

```handlebars
 {{- printf " %s=%q" $k $v | safeHTMLAttr }}
```

I understand it's ... usefulness, but it would violate Hugo's security policy.

Also:

```
{{- $summary := or (.Get "summary") (T "details") "Details" }}
```

I'm not sure if we're using the pattern above elsewhere, but I think we need to need to namespace the translation keys. I'm not sure what, but `details` seems too generic.

Also, for the future, it would be great if we could wait with the implementation until the issue goes from `Proposal` to `Enhancement`. I understand the desire for speed/momentum, but ...

racehd:
Thank you for the feedback. I have made the following modifications:
- Removed `safeHTMLattr`
- Includes namespaced translation key `shortcodes.details`. I found no examples of other shortcode translation keys within Hugo already. 
- Updated to use explicit argument passing for HTML attributes instead of a [,,,params] approach. This felt like it would be less confusing to end-users since `style` would not work without `safeCSS` if a user had multiple style changes (this is also why I did not add `style` as an allowed attribute) 
- Minor formatting of details shortcode. Since attributes are being passed conditionally this inserts a newline to output if each argument is in a new line. To avoid this, i put the allowed attributes on one line to make formatting more predictable for tests. `<details{{- if $open }} open{{ end }}{{- if $name }} name="{{ $name }}"{{- end }}{{- if $class }} class="{{ $class }}"{{- end }}>`

jmooring:
@racehd Let's put this on hold for a bit. There's some other work in progress that will affect how we want to handle this. Note that it is desirable to allow some other attributes like `id`, `title`, `data-foo`, etc. Example use of the `data` attribute:

````text
{{< details data-copy=true >}}
```css
body {color: red;}
```
{{< /details >}}
````

Then add a copy button via JS to all `details` element with the `data-copy` attribute.

And in the future, whether added to the embedded template or extended by site and theme authors:

```text
{{< details data-file="examples/something.go" />}}
```

PR #13120, opened 2024-12-09
Add asciidoctor to container

xoxys:
It would be helpful to have `asciidoctor` in the official container image to support the Hugo asciidoc feature.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,8 @@ RUN apk add --no-cache \
     git \
     runuser \
     nodejs \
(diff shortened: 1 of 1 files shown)


bep:
I'm reluctant to do that. I suggest you start by creating a proposal issue.

xoxys:
You have already added the heavy `node` package, but don't want to add the 1.4MB `asciidoctor` package as well? Alright, in that case the official Hugo image is useless for all your users using asciidoc.

PR #13140, opened 2024-12-13
tpl/collections: Allow querify to accept a map argument

### LINKED ISSUE #13131: tpl/collections: Allow querify to accept a map argument
The [`collections.Querify`](https://gohugo.io/functions/collections/querify/) template function accepts either of the following argument forms:

```text
{{ collections.Querify "a" 1 "b" 2 }}         --> a=1&b=2
{{ collections.Querify (slice "a" 1 "b" 2) }} --> a=1&b=2
```

But it would convenient if we could do:

```text
{{ collections.Querify (dict "a" 1 "b" 2) }}
```

For example, in our [embedded youtube shortcode] we:

[embedded youtube shortcode]: https://github.com/gohugoio/hugo/blob/989b299b55d3fb48f8447ea69e85e26ee9dc7795/tpl/tplimpl/embedded/templates/shortcodes/youtube.html#L76-L94

1. Build a map of parameters
2. Convert the map into a slice with this construct:

    ```text
    {{ range $k, $v := $params }}
      {{ $s = $s | append $k }}
      {{ $s = $s | append $v }}
    {{ end }}
    ```
3. Pass the slice to `collections.Querify`

It would be more elegant if we could remove step 2 above.

jmooring:
Closes #13131

### DIFF
--- a/tpl/collections/querify.go
+++ b/tpl/collections/querify.go
@@ -1,0 +1,125 @@
+// Copyright 2024 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collections
+
+import (
+	"errors"
+	"net/url"
+
+	"github.com/gohugoio/hugo/common/maps"
+	"github.com/spf13/cast"
+)
+
+var (
+	errWrongArgStructure = errors.New("expected a map, a slice with an even number of elements, or an even number of scalar values, and each key must be a string")
+	errKeyIsEmptyString  = errors.New("one of the keys is an empty string")
+)
+
+// Querify returns a URL query string composed of the given key-value pairs,
+// encoded and sorted by key.
+func (ns *Namespace) Querify(params ...any) (string, error) {
+	if len(params) == 0 {
+		return "", nil
+	}
+
+	if len(params) == 1 {
+		switch v := params[0].(type) {
+		case map[string]any: // created with collections.Dictionary
+			return mapToQueryString(v)
+		case maps.Params: // site configuration or page parameters
+			return mapToQueryString(v)
+		case []string:
+			return stringSliceToQueryString(v)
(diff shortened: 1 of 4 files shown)


jmooring:
These are now equivalent:

```text
1. {{ collections.Querify "a" 1 "b" 2 }}
2. {{ collections.Querify (slice "a" 1 "b" 2) }}
3. {{ collections.Querify (dict "a" 1 "b" 2) }}
````

Benchmark testing indicates:
- Worse performance when passing a sequence of scalar values
- No change when passing a slice

I'm guessing that the performance degradation is primarily due to increased error checking and some deduplication of code. This function isn't widely used, so I suspect this won't be a problem for anyone.

```text
goos: linux
goarch: amd64
pkg: github.com/gohugoio/hugo/tpl/collections
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
                │     old     │                 new                 │
                │   sec/op    │    sec/op     vs base               │
Querify-12        670.6n ± 2%   905.4n ± 15%  +35.02% (p=0.002 n=6)
QuerifySlice-12   556.6n ± 5%   552.7n ±  4%        ~ (p=0.589 n=6)
QuerifyMap-12                   719.8n ±  4%
geomean           610.9n        711.5n        +15.79%

                │    old     │                 new                 │
                │    B/op    │    B/op     vs base                 │
Querify-12        128.0 ± 0%   224.0 ± 0%  +75.00% (p=0.002 n=6)
QuerifySlice-12   152.0 ± 0%   152.0 ± 0%        ~ (p=1.000 n=6) ¹
QuerifyMap-12                  128.0 ± 0%
geomean           139.5        163.3       +32.29%
¹ all samples are equal

                │    old     │                new                 │
                │ allocs/op  │ allocs/op   vs base                │
Querify-12        8.000 ± 0%   8.000 ± 0%       ~ (p=1.000 n=6) ¹
QuerifySlice-12   8.000 ± 0%   8.000 ± 0%       ~ (p=1.000 n=6) ¹
QuerifyMap-12                  7.000 ± 0%
geomean           8.000        7.652       +0.00%
¹ all samples are equal
```

tpl/collections/querify.go
```diff
@@ -0,0 +1,125 @@
...
+				return "", err
+			}
+			return stringSliceToQueryString(s)
+		default:
+			return "", fmt.Errorf("%s", msgExpectedParams)
+		}
+	}
+
+	if len(params)%2 != 0 {
+		return "", fmt.Errorf("%s", msgExpectedParams)
```
bep, line 56: This is a little bit of a "style" thing, but if the error message never changes, I suggest we replace the `msgExpectedParams` with global error variables (using `errors.New`) and return those.

bep (changes requested):

PR #13180, opened 2024-12-20
Refactor: filepath.Walk to filepath.WalkDir

Vkanhan:
filepath.Walk calls os.Lstat for every file or directory to retrieve os.FileInfo, which can be slower.

filepath.WalkDir avoids unnecessary system calls since it provides a fs.DirEntry, which includes file type information without requiring a stat call.


bep:
Thanks for this, but this code is only used for code generation at build time, so the performance gains here doesn't matter.

PR #13190, opened 2024-12-23
Fix docstring to match function name

jdbaldry:


bep:
Thanks for this; could you push the "sign CLA" button above?

jdbaldry:
All signed! Merry Christmas :)

PR #13200, opened 2024-12-29
config/allconfig: Throw error when output format is not defined

### LINKED ISSUE #13199: config/allconfig: Error not thrown when output format is not defined
Reference: <https://discourse.gohugo.io/t/seems-hugo-no-longer-throw-an-error-if-output-format-is-invalid/52912>

This worked as expected through v0.122.0. Failing test:

```go
// Issue 13199
func TestInvalidOutputFormat(t *testing.T) {
	t.Parallel()

	files := `
-- hugo.toml --
[outputs]
home = ['html','foo']
-- layouts/index.html --
x
`

	b.Assert(err.Error(), qt.Contains, `failed to create config: unknown output format "foo" for kind "home"`)
}
```

jmooring:
Fixes #13199

### DIFF
--- a/config/allconfig/allconfig.go
+++ b/config/allconfig/allconfig.go
@@ -734,7 +734,7 @@ func (c *Configs) Validate(logger loggers.Logger) error {
 
 // transientErr returns the last transient error found during config compilation.
 func (c *Configs) transientErr() error {
-	for _, l := range c.LanguageConfigSlice {
+	for _, l := range c.LanguageConfigMap {
 		if l.C.transientErr != nil {
 			return l.C.transientErr
 		}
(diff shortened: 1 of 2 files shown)


config/allconfig/allconfig.go
```diff
@@ -734,7 +734,7 @@ func (c *Configs) Validate(logger loggers.Logger) error {
 
 // transientErr returns the last transient error found during config compilation.
 func (c *Configs) transientErr() error {
-	for _, l := range c.LanguageConfigSlice {
```
jmooring, line 737: `(allconfig.Configs).LanguageConfigSlice` is empty at this point. It is populated by `(allconfig.Configs).Init` which isn't called until _after_ we've checked for transient errors.

PR #13203, opened 2024-12-29
Open external links in new tab/window

TheGroundZero:
Add `rel="external"` and `target="_blank"` to external links.
Based on https://gohugo.io/render-hooks/links/#examples


jmooring:
This is an opinionated change that some (most) won't want. Please override the template as needed.

TheGroundZero:
Using `rel=external` does [add semantics](https://www.w3schools.com/TAGS/att_a_rel.asp) to URLs and helps with SEO.
Opening external links in new tabs is indeed a personal preference. My preference is not to loose track of the origin page when clicking on an external link.

PR #13212, opened 2025-01-02
tpl/images: Add images.QR function

### LINKED ISSUE #13205: Add QR code generation to image functions for offline permalinks
This might sound a bit odd for a website, but the idea is to add a QR code of the permalink of a page to that page itself, possibly only displayed by a `@media print` css rule. Then when someone encounters the printed version, they can easily go to the up to date online version.

jmooring:
Closes #13205

Note that this uses `(create.Client).FromString` under the covers, which seemed a lot simpler than creating a `FromBytes` or `FromImage` method.

Site and theme authors should test the rendered image per the documentation:

> Always test the rendered QR code after resizing, both on-screen and in print.


tpl/images/images.go
```diff
@@ -117,3 +122,65 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+		"high":     qr.H,
+	}
+	qrDefaultErrorCorrectLevel = "medium"
+)
+
+// QR encodes the given text to a QR code using the given error correction
+// level, returning an image resource.
+func (ns *Namespace) QR(args ...any) (images.ImageResource, error) {
+	if len(args) == 0 || len(args) > 2 {
+		return nil, errors.New("requires one or two arguments")
```
bep, line 140: This looks good in general. But re. 2 positional args vs a map is something I have thought about lately ... Don't make the change if you somehow disagree, but my take on this is:

* A map is a better choice for multiple args, and especially if it may become more in the future.
* The decoding becomes much simpler, you just create a default struct and use `mapstruct.WeakDecode`.
* It is in my head simpler to document, which becomes even more truthful if there comes more options in the future.

tpl/images/images.go
```diff
@@ -117,3 +122,65 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+	targetPath := fmt.Sprintf("qr_%v.png", hashing.XxHashFromStringHexEncoded(text+levels))
+
+	r, err := ns.createClient.FromString(targetPath, string(png))
+	if err != nil {
+		return nil, err
+	}
+
+	ir, ok := r.(images.ImageResource)
+	if !ok {
+		return nil, errors.New("bug: resource is not an image resource")
```
bep, line 182: I prefer panics for bugs. In this case it doesn't matter much, as it will look like an error, but at least I wouldn't look for a corresponding test ...

bep (changes requested):

tpl/images/images.go
jmooring: I'll make both changes.

tpl/images/images.go
```diff
@@ -117,3 +124,69 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+	"high":     qr.H,
+}
+
+const qrDefaultErrorCorrectionLevel = "medium"
+
+// QR encodes the given text to a QR code using the given options, returning an
+// image resource.
+func (ns *Namespace) QR(args ...any) (images.ImageResource, error) {
+	if len(args) == 0 || len(args) > 2 {
+		return nil, errors.New("requires one or two arguments")
```
bep, line 146: We talked about making it simpler (to document etc.); I'm not sure allowing both positional args and a map accomplishes that.

Make it 1 arg (`opts any`).

jmooring: Make sense to me. Done.

tpl/images/images.go
```diff
@@ -117,3 +124,60 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+		return nil, errors.New("error correction level must be one of low, medium, quartile, or high")
+	}
+
+	code, err := qr.Encode(qrOptions.Text, level)
+	if err != nil {
+		return nil, err
+	}
+
+	hash := hashing.XxHashFromStringHexEncoded(qrOptions.Text + qrOptions.Level)
+	targetPath := filepath.Join(qrOptions.TargetDir, fmt.Sprintf("qr_%s.png", hash))
```
bep, line 170: This all looks good now. Just one tiny detail: FromString expects Unix paths; I'm assuming that Windows paths will somehow work as well, but I suggest that you change the above to use `path.Join...`.

jmooring: Will do.

I'm not thrilled about what you currently need to do to scale these down to something other that [8 image pixels per QR code module](https://github.com/rsc/qr/blob/master/qr.go#L77). See last example in documentation. So I may add another option if I'm satisfied with the result. Great idea to pass a map instead of scalars... I get it. It makes additions almost trivial.

docs/content/en/functions/images/QR.md
```diff
@@ -0,0 +1,96 @@
+---
+title: images.QR
+description: Encodes text to a QR code using the given error correction level, returning an image resource.
+keywords: []
+action:
+  aliases: []
+  related: []
+  returnType: images.ImageResource
+  signatures: ['images.QR TEXT [OPTIONS]']
```
bep, line 9: OPTIONS only.

tpl/images/images.go
bep: Yea, going forward, I see 3 variants for template funcs args:

1. One single argument.
2. Multiple arguments where one of them is a obvious candidate for "piping", e.g. `{{ $resource | js.Build (dict "other" "option" }}`
3. Multiple arguments where there's no obvious pipe candidate: One options map.

tpl/images/images.go
```diff
@@ -117,3 +124,60 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+	if !ok {
+		return nil, errors.New("error correction level must be one of low, medium, quartile, or high")
+	}
+
+	code, err := qr.Encode(qrOptions.Text, level)
+	if err != nil {
+		return nil, err
+	}
+
+	hash := hashing.XxHashFromStringHexEncoded(qrOptions.Text + qrOptions.Level)
```
bep, line 169: I suggest we create the hash from the input `opts`, e.g. `hash := hashing.HashString(opts)`. That way:

1. It will change whenever the user changes the options.
2. We don't have to worry about it when adding new options.

This is the pattern we use elsewhere (this becomes even more important if we want to cache to disk).

jmooring: Will do. The scaling thing works great, so I'm adding that option. Also, this is useful enough that I'd like to add an embedded shortcode. I know I've pushed back on adding more shortcodes in the past, but I think this is a non-opinionated addition. Please shout if you disagree.

tpl/images/images.go
```diff
@@ -117,3 +124,69 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+
+const (
+	qrDefaultErrorCorrectionLevel = "medium"
+	qrDefaultScale                = 4
+)
+
+// QR encodes the given text into a QR code using the specified options,
+// returning an image resource.
+func (ns *Namespace) QR(opts any) (images.ImageResource, error) {
+	qrOptions := qrOptions{
```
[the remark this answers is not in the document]
bep: Also, see https://github.com/gohugoio/hugo/blob/723e3f4342dbf7b64350eb1db2d96dd53b919a0d/resources/resource_factories/create/create.go#L291

Using the newly added `FromOpts` method avoids having to re-create the same QR for the same input over and over again.

* Set `TargetPathHasHash=true`
* Use `NewReadSeekerNoOpCloserFromBytes`.
* Create the QR in the outer func.

tpl/tplimpl/embedded/templates/shortcodes/qr.html
```diff
@@ -0,0 +1,80 @@
...
+
+{{- /* Render image. */}}
+{{- if not $errors }}
+  {{- $opts := dict "text" $text "level" $level "scale" $scale "targetDir" $targetDir }}
+  {{- with images.QR $opts -}}
+    <img src="{{ .RelPermalink }}" width="{{ .Width }}" height="{{ .Height }}"
+      {{- with $alt }} alt="{{ $alt }}" {{- end }}
+      {{- with $class }} class="{{ $class }}" {{- end }}
+      {{- with $id }} id="{{ $id }}" {{- end }}
+      {{- with $style }} style="{{ $style | safe.CSS }}" {{- end }}
```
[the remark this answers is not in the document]
jmooring: Yeah, the only reason I added the `style` attribute was so that I could set `width: initial` on the docs site because I couldn't figure out how to override with tachyons, and was too lazy to recompile the theme CSS. But I agree and will remove.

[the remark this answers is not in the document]
bep: That theme will not live much longer, so if you need to add some `<styles>` sections to the templates, that would work.

jmooring:
I'll handle styling of the example images on the docs site, so if you build the docs from this repo the examples will be stretched.

tpl/images/images.go
```diff
@@ -117,3 +125,79 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+	code.Scale = opts.Scale
+	content := code.PNG()
+
+	targetPath := path.Join(opts.TargetDir, fmt.Sprintf("qr_%s.png", hashing.HashString(opts)))
+
+	r, err := ns.createClient.FromOpts(
+		create.Options{
+			TargetPath:        targetPath,
+			TargetPathHasHash: true,
+			CreateContent: func() (func() (hugio.ReadSeekCloser, error), error) {
```
bep, line 186: One motivation behind the new `FromOpts` method was to make sure we create a given QR only once for some given input, so we need to move  `code, err := qr.Encode(opts.Text, level) ... `. down here.

tpl/images/images.go
```diff
@@ -117,3 +125,77 @@ func (ns *Namespace) Filter(args ...any) (images.ImageResource, error) {
...
+	}
+
+	targetPath := path.Join(opts.TargetDir, fmt.Sprintf("qr_%s.png", hashing.HashString(opts)))
+
+	r, err := ns.createClient.FromOpts(
+		create.Options{
+			TargetPath:        targetPath,
+			TargetPathHasHash: true,
+			CreateContent: func() (func() (hugio.ReadSeekCloser, error), error) {
+				f := func() (hugio.ReadSeekCloser, error) {
```
bep, line 179: The double func signature is a little odd, I agree, but the motivation is so we can do:

```go
CreateContent: func() (func() (hugio.ReadSeekCloser, error), error) {
	code, err := qr.Encode(opts.Text, level)
	if err != nil {
		return nil, err
	}
	code.Scale = opts.Scale
	png := code.PNG())
	return func() (hugio.ReadSeekCloser, error) {
		return hugio.NewReadSeekerNoOpCloserFromBytes(code.PNG()), nil
	}
},
```

This way we `Encode` once, and reuse the PNG bytes if needed. The `ReadSeekerCloser` interface is what a `File` would implement.

bep (approved):

mrusme:
Just in case anyone is stumbling upon this feature -- thank you for it btw! -- and is thinking _"dammit, no way to change the colors of the QR code to adjust for dark/light themes!"_, I got you covered:

```html
      <div id="qrcode">
      {{ with (images.QR .Permalink (dict "level" "high" "scale" 3)) }}
        <img src="{{ .RelPermalink }}"
            width="{{ .Width }}"
            height="{{ .Height }}"
            alt="QR code linking to {{ $.Permalink }}"
            class="qrcode-light"
            loading="lazy">
        {{ with . | images.Filter images.Invert }}
          <img src="{{ .RelPermalink }}"
              width="{{ .Width }}"
              height="{{ .Height }}"
              alt="QR code linking to {{ $.Permalink }}"
              class="qrcode-dark"
              loading="lazy">
        {{ end }}
      {{ end }}
      </div>
```

```css
@media (prefers-color-scheme: dark) {
  #qrcode > .qrcode-light {
    display: none;
  }
}

@media (prefers-color-scheme: light) {
  #qrcode > .qrcode-dark {
    display: none;
  }
}
```

PR #13223, opened 2025-01-06
images.Text: Add option for horizontal alignment

### LINKED ISSUE #10849: Add support for aligning text when adding to images using images.Text
If we could left/center/right align text within a textbox (and maybe support word-wrap?) this would make the text rendering a lot more useful.  Often when we are rendering images dynamically, to match some designer-specified template, they would include left/centre/right alignment, which I don't think we can currently do.

jlskuz:
**_Please note that this is my first contribution to Hugo and also the first time I wrote something in Go. Please let me know if I missed anything and I will try my best to correct it._**

Add an "alignx" option to the images.Text to control whether the value of the "x" option is the left border of the text (current behaviour), the center of each line or the right border.

Fixes #10849

Example with `"alignx" "left"`:

grafik

Example with `"alignx" "center"`:

grafik

Example with `"alignx" "right"`:

grafik

### DIFF
--- a/resources/images/filters.go
+++ b/resources/images/filters.go
@@ -69,6 +69,7 @@ func (*Filters) Text(text string, options ...any) gift.Filter {
 		size:        20,
 		x:           10,
 		y:           10,
+		alignx:      "left",
 		linespacing: 2,
 	}
 
@@ -87,6 +88,12 @@ func (*Filters) Text(text string, options ...any) gift.Filter {
 				tf.x = cast.ToInt(v)
 			case "y":
 				tf.y = cast.ToInt(v)
+			case "alignx":
+				tf.alignx = cast.ToString(v)
+				if tf.alignx != "left" && tf.alignx != "center" && tf.alignx != "right" {
+					panic("alignx must be one of left, center, right")
+				}
+
 			case "linespacing":
 				tf.linespacing = cast.ToInt(v)
 			case "font":
--- a/resources/images/text.go
+++ b/resources/images/text.go
@@ -35,6 +35,7 @@ type textFilter struct {
 	text        string
 	color       color.Color
 	x, y        int
+	alignx      string
 	size        float64
 	linespacing int
 	fontSource  hugio.ReadSeekCloserProvider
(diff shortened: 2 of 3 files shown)


resources/images/filters.go
```diff
@@ -87,6 +88,8 @@ func (*Filters) Text(text string, options ...any) gift.Filter {
 				tf.x = cast.ToInt(v)
 			case "y":
 				tf.y = cast.ToInt(v)
+			case "alignx":
+				tf.alignx = cast.ToString(v)
```
bep, line 92: Add some validation here. This method does not return error, so you need to panic, e.g. `panic("must be one of left, ...")`

resources/images/text.go
```diff
@@ -77,33 +78,67 @@ func (f textFilter) Draw(dst draw.Image, src image.Image, options *gift.Options)
...
 
-	// Draw text, consider and include linebreaks
 	maxWidth := dst.Bounds().Dx() - 20
-	fontHeight := face.Metrics().Ascent.Ceil()
 
-	// Correct y position based on font and size
-	f.y = f.y + fontHeight
+	// Default left alignment
+	availableWidth := maxWidth - f.x
+	if f.alignx == "center" {
```
bep, line 85: I would prefer a switch here and use all 3 values, e.g.

```go
var availableWidth int
switch f.alignX {
  case "left":
  case "center":
  case "right":
}
```

resources/images/text.go
```diff
@@ -77,33 +78,67 @@ func (f textFilter) Draw(dst draw.Image, src image.Image, options *gift.Options)
...
 		y = y + fontHeight + f.linespacing
-		d.Dot = fixed.P(f.x, y)
 	}
 }
 
 func (f textFilter) Bounds(srcBounds image.Rectangle) image.Rectangle {
 	return image.Rect(0, 0, srcBounds.Dx(), srcBounds.Dy())
 }
+
+func calcX(alignX string, x int, strWidth fixed.Int26_6) (rightX int) {
```
bep, line 133: Named return values in Go is, I think, in general frowned upon, and only used if you really need it (the one case I can come up with is when you need to change the return value in defer/recover).

Also, if `calcX` isn't in general useful, I would just pull it into the method, e.g.: `calcX := func(alignX string, x int, strWidth fixed.Int26_6) (rightX int) {...}`

resources/images/text.go
```diff
@@ -77,33 +78,67 @@ func (f textFilter) Draw(dst draw.Image, src image.Image, options *gift.Options)
...
+
+	if alignX == "center" {
+		rightX = x - (strWidth.Ceil() / 2)
```
bep, line 136: See my comment above about switch.

jlskuz:
@bep Thanks for the review all your comments should be addressed now.

jmooring (commented):
This work fine, you just have to get the image before setting the options in order to calculate x offset...

```text
{{ $fontPath := "https://github.com/google/fonts/raw/main/ofl/lato/Lato-Regular.ttf" }}
{{ $imagePath := "images/original.jpg" }}
{{ $text := "A kitten!" }}

{{ with try (resources.GetRemote $fontPath) }}
  {{ with .Err }}
    {{ errorf "%s" . }}
  {{ else with .Value }}
    {{ $font := . }}
    {{ with resources.Get $imagePath }}
      {{ $opts := dict
        "color" "#fbfaf5"
        "font" $font
        "linespacing" 8
        "size" 120
        "x" (div .Width 2 | int)
        "y" 50
        "alignx" "center"
      }}
      {{ $filter := images.Text $text $opts }}
      {{ with . | images.Filter $filter }}
      {{ end }}
    {{ else }}
      {{ errorf "%s" "Unable to get resource %q" $imagePath }}
    {{ end }}
  {{ else }}
    {{ errorf "Unable to get resource %q" $fontPath }}
  {{ end }}
{{ end }}
```

bep:
>Is there an easy way to re-trigger the tests myself? I fixed them.

No.

There is a restriction in GitHub for first time contributors of a repository, it requires someone to Approve each PR build (which I guess makes sense, else spam bots would go bananas creating new CI builds).

Once this is merged, any new PR or push to a PR branch from you will trigger a build without any ceremony.

PR #13232, opened 2025-01-08
Development guide: add ./check_gofmt.sh to test step

jlskuz:


bep:
So, this is correct, but I'm going to close it. We had a much more involved guide before (using `mage check`). My arguments against this particular patch is:

1. It's not portable (e.g. Windows command prompt)
2. This will be caught in CI checks.
3. For Go developers this should "go without saying", and most have this set up as part of their dev environment (I have it on save in VS Code); and if you haven't, the  CI failure may be a good reminder to set it up.

PR #13276, opened 2025-01-17
Add loading attribute

nfriedli:
All images shortcodes should have a loading attribute (for example for lazy).

### DIFF
--- a/tpl/tplimpl/embedded/templates/shortcodes/qr.html
+++ b/tpl/tplimpl/embedded/templates/shortcodes/qr.html
@@ -9,6 +9,7 @@ Encodes the given text into a QR code using the specified options and renders th
 @param {string} [class] The class attribute of the img element.
 @param {string} [id] The id attribute of the img element.
 @param {string} [title] The title attribute of the img element.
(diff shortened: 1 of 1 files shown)


jmooring:
This is fine, but the images are like 1k...

nfriedli:
I agree with you. In the vast majority of cases, the images are tiny.

But I think it's a good idea to add this option for 2 reasons: 

1. To be consistent with the `figure.html` shortcode, which features the loading attribute. This way, users really have control over image loading, with the shortcode provided by default.
2. Because I can think of lots of other relevant uses, notably for creating deliberately large QR codes, with lots of text. I've just made a `{{< qr level=“high” scale=16 >}}`. And as I'll have several on the page, for poetry, it would be useful.

----

**Personal note.** I find the `qr.html` shortcode very, very well designed.

The variable harvesting part is great for overloading the default configuration. I wanted to create variables in the configuration, but it's so clear that I did it in my local shortcode, for example:

```
{{- $targetDir := or (.Get “targetDir”) “images/qr” }}
{{- $class := or (.Get “class”) “qr” }}
{{- $loading := or (.Get “loading”) “lazy” }}

```

Many thanks for your work, @jmooring & @bep.

PR #13277, opened 2025-01-17
tpl/tplimpl: Prevent overloading of embedded comment shortcode

jmooring:
Creates a mechanism to prevent loading of user space templates that
match a list of reserved paths. The primary intent is to prevent users
from overloading specific embedded templates. For example, with the
string "shortcodes/comment" in the list of reserved paths, we do not
load any of the following from user space:

  - layouts/shortcodes/comment.html
  - layouts/shortcodes/comment.html.html
  - layouts/shortcodes/comment.en.html.html
  - layouts/shortcodes/comment.json"
  - layouts/shortcodes/comment.json.json"
  - layouts/shortcodes/comment.en.json.json"

### DIFF
--- a/common/paths/path.go
+++ b/common/paths/path.go
@@ -428,3 +428,16 @@ func ToSlashPreserveLeading(s string) string {
 func IsSameFilePath(s1, s2 string) bool {
 	return path.Clean(ToSlashTrim(s1)) == path.Clean(ToSlashTrim(s2))
 }
+
+// ToSlashNoExtensions returns the result of replacing each separator character
+// in path s with a slash ('/') character, then removing all file extensions.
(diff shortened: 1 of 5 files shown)


bep:
I have thought about this a little, and I understand the problem we're trying to solve, but I suspect a simpler way to go about this is to introduce a reserved namespace (e.g. `h:comment`) -- and then deprecate the old. I have had some ideas/use cases for a internal Hugo shortcode namespace before (cannot think of an example now, but I think they were great ideas).

We should think a little about what the namespace should look like, though.

jmooring:
Closing in favor of <https://github.com/gohugoio/hugo/issues/13278>.

PR #13281, opened 2025-01-19
tpl/tplimpl: Fix context in shortcode error messages

### LINKED ISSUE #13279: Fix context issue with some embedded templates
Reference: <https://discourse.gohugo.io/t/problems-with-0-141-0-and-tweet-x-embed/53170/10>

jmooring:
Fixes #13279

OK, so this PR does a bit more than advertised, primarily moving the shortcode integration tests into their own file, and adding a few more.

For shortcodes that render more than a few lines, these tests compare content hashes. It may take a little more time to figure out why one of these tests fails in the future, but this does a much better job of detecting the  failure.


tpl/tplimpl/shortcodes_integration_test.go
```diff
@@ -0,0 +1,328 @@
...
+title: home
+---
+{{< instagram CxOWiQNP2MO >}}
+-- layouts/index.html --
+{{- .Content | hash.XxHash -}}
+`
+
+	// Regular mode
+	b := hugolib.Test(t, files)
+	b.AssertFileContentEquals("public/index.html", "a7937c49665872d3")
```
bep, line 107: So, I like the concept of using hashes here, but I suspect I'm going to get bitten by the lack of output at some point. 

I suggest that we print both the hash and the content, e.g.

```
Hash: {{ .Content | hash.XxHash }}
Content: {{ .Content }}
```

And then do:

```
b.AssertFileContent("public/index.html", "a7937c49665872d3")
```

The test will be as robust as before, but we will get better output on failures.

PR #13342, opened 2025-02-03
parser/pageparser: enforce non-alphanumeric parameters must be quoted

guilhas07:
## Problem:
Before unquoted named parameters would be parsed until a space was
found, allowing for non-alphanumeric parameters.

## Solution:
Parse until a space, or a non-alphanumeric character is found.

### DIFF
--- a/parser/pageparser/pagelexer_shortcode.go
+++ b/parser/pageparser/pagelexer_shortcode.go
@@ -139,7 +139,13 @@ func lexShortcodeParam(l *pageLexer, escapedQuoteStart bool) stateFunc {
 }
 
 func lexShortcodeParamVal(l *pageLexer) stateFunc {
-	l.consumeToSpace()
+	for {
+		r := l.next()
+		if r == eof || isSpace(r) || (!isAlphaNumericOrHyphen(r) && r != '.') {
+			l.backup()
+			break
+		}
+	}
 	l.emit(tScParamVal)
 	return lexInsideShortcode
 }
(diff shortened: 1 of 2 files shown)


jmooring:
This a breaking change.

With v0.143.0 this works fine:

```text
{{< foo param=https://example.ork/foo/?a=b&c=d#bar >}}
```

But with this PR:

> Error: ... unrecognized character in shortcode action: U+003A ':'. Note: Parameters with non-alphanumeric args must be quoted

jmooring:
I suggest you create a topic in the [forum](https://discourse.gohugo.io/) to discuss the particular problem you are having. If it's a bug or opportunity for improvement, we'll create an issue.  Please do not submit PRs without a corresponding issue that's been approved by the project lead.

guilhas07:
> This a breaking change.
> 
> With v0.143.0 this works fine:
> 
> ```
> {{< foo param=https://example.ork/foo/?a=b&c=d#bar >}}
> ```
> 
> But with this PR:
> 
> > Error: ... unrecognized character in shortcode action: U+003A ':'. Note: Parameters with non-alphanumeric args must be quoted

As you  can see in the error, which was _not_ added by me, `Parameters with non-alphanumeric args must be quoted`: 
https://github.com/gohugoio/hugo/blob/377287a614821aa024385772db64721b0fd24fc1/parser/pageparser/pagelexer_shortcode.go#L339

That is why I made this change. This makes unnamed and named parameters behaviour consistent. 
This is currently not valid in `v0.143.0`:
```
{{< foo https://example.ork/foo/?a=b&c=d#bar >}}
```
Maybe I should have been more obvious in the pull request/commit description.

> Please do not submit PRs without a corresponding issue that's been approved by the project lead.

I submitted a similar fix for this part of the code which was approved #13299 and noticed multiple PRs from non-members that didn’t seem to be associated with any approved issues so I made this honest mistake. See issue #13343

Issue #11722, opened 2023-11-21
Add permalinks tokens contentbasename and contentbasenameorslug


hacdias:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.120.4-f11bca5fec2ebb3a02727fb2a5cfb08da96fd9df+extended darwin/arm64 BuildDate=2023-11-08T11:18:07Z VendorInfo=brew
</pre>

### Does this issue reproduce with the latest release?

1. Create a `categories` taxonomy.
2. Create file `/categories/a/_index.md` and `/categories/b/_index.md`
3. Create posts for those categories.
4. Add the following to your config:

```
[permalinks]
  categories = '/:filename/'
```

All categories will be concurrently built to `/index.html`, which leads me to believe that `:filename` is empty for them.

jmooring:
> /categories/a/_index.md and /categories/b/_index.md

Both filenames are _index.md. Not sure why you would want to do this. Seems like you want this instead:

```
[permalinks.term]
categories = '/:slug/'
```

hacdias:
@jmooring for other files, `:filename` is filled up with their directory name. I have the following rule:

```
posts = '/:year/:month/:day/:filename/'
```

And the post at `posts/2023/distributed-tracing-with-opentelemetry/index.md` will be published at: 

`/2023/04/18/distributed-tracing-with-opentelemetry/` (using the date in the frontmatter).

If it works for posts, why doesn't it work for categories?

I want the name attributed to that entry. Whether it is a directory entry of just a markdown file, it should not matter.

bep:
Assuming you have files backing these categories, I don't see how they should behave differently.

But I see this in your config, which does not match your expectations:


```toml
[permalinks]
  tags = '/tags/:title/'
  categories = '/:title/'
  posts = '/:year/:month/:day/:filename/'
```

Also, I suspect that `:slugorfilename` may be a better option.

hacdias:
@bep 

> But I see this in your config, which does not match your expectations:

I want to change to `categories = '/:filename/'`. But if I do that, it will all concurrently build to `/index.html` instead of to `/:filename/index.html`.

jmooring:
https://github.com/gohugoio/hugo/blob/5887230b7d8bac2c5429f8cb51af1e75e3cdf085/resources/page/permalinks.go#L259-L271

[...]

PR #13382, opened 2025-02-12
feat: add :contentbasename permalink token

hacdias:
Adds changes described in https://github.com/gohugoio/hugo/issues/11722#issuecomment-2093337853, except for the deprecation of `:filename`.

### DIFF
--- a/resources/page/permalinks_test.go
+++ b/resources/page/permalinks_test.go
@@ -46,6 +46,8 @@ var testdataPermalinks = []struct {
 	{"/:sections[0]/:sections[last]/", true, "/a/c/"},               // Sections
 	{"/\\:filename", true, "/:filename"},                            // Escape sequence
 	{"/special\\::slug/", true, "/special:the-slug/"},               // Escape sequence
+	{"/:contentbasename/", true, "/index/"},                         // Content base name
+	{"/:contentbasenameorslug/", true, "/index/"},                   // Content base name or slug
 
 	// Failures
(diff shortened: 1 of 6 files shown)


resources/page/permalinks_test.go
```diff
@@ -46,6 +46,8 @@ var testdataPermalinks = []struct {
 	{"/:sections[0]/:sections[last]/", true, "/a/c/"},               // Sections
 	{"/\\:filename", true, "/:filename"},                            // Escape sequence
 	{"/special\\::slug/", true, "/special:the-slug/"},               // Escape sequence
+	{"/:contentbasename/", true, "/index/"},                         // Content base name
+	{"/:contentbasenameorslug/", true, "/index/"},                   // Content base name or slug
```
hacdias, line 50: I noticed that, for the test files, the `component` is not defined for paths, meaning that here we just get the filename (without extension) instead of the parent file's name as we'd get inside the `content` directory.

hacdias:
Tests seem to be failing with unrelated errors:

> ERROR Unable to get remote resource: https://gohugo.io/img/hugo.png

bep:
@jmooring No rush (I'm about to log off for the weekend ...), but when you have a minute, could you have a look at f6e8d3472592aee061be03cf44a5bfdfabcfd4d8

The test deliberately fails, but should get the point across: I think we need to make the pages added via content adapters behave the same way re these new tokens.

bep:
I will wrap this up in #/13402

PR #13395, opened 2025-02-13
fix server.redirects.fromRe being ignored unless server.redirects.from set

p-weston:
Tiny fix: server.redirects.fromRe was being ignored unless server.redirects.from was also set.


bep:
Thanks for this, good catch.

PR #13415, opened 2025-02-18
images.Text: Add option for vertical alignment

### LINKED ISSUE #13414: images.Text: Add support for vertical alignment of text
It would be useful to be able to choose a vertical alignment (relative to the ``y`` offset) for text added using ``images.Text``. This becomes useful in the following scenarios when the number of lines of text is not fixed.
- Centering text vertically.
- Creating an Open Graph image with a subtitle at a fixed distance below the title. This can be achieved by making the title "bottom" aligned and the subtitle "top" aligned and having the same ``y`` offset for both texts.

It is not currently possible to implement these using only the ``y`` option which sets a vertical offset for the top of the text relative to the top of the image.

Vertical alignment would complement the horizontal alignment feature (``alignx``) mentioned in #10849 and added in #13223.

pranshugaba:
Add option ``aligny`` to specify the vertical alignment of the text with respect to the ``y`` offset from the top of the image. Possible values of ``aligny`` are ``top`` (default), ``center``, and ``bottom``.

The height of the block of text is measured from the top of the first line to the baseline of the last line.

- ``top``: (Current behaviour) The top of the first line of the block of text is at an offset of ``y`` from the top of the image.

- ``center``: The vertical center of the block of text is at an offset of ``y`` from the top of the image.

- ``bottom``: The baseline of the last line of the text is at an offset of ``y`` from the top of the image.

Resolves #13414

Examples:
- Vertically center-aligned text with ``aligny`` set to ``center``
center-1
center-2
center-3
- Title is ``bottom``-aligned and subtitle is ``top``-aligned
bottom-1
bottom-2

### DIFF
--- a/docs/content/en/functions/images/Text.md
+++ b/docs/content/en/functions/images/Text.md
@@ -21,6 +21,9 @@ alignx
  {{< new-in 0.141.0 />}}
 : (`string`) The horizontal alignment of the text relative to the horizontal offset, one of `left`, `center`, or `right`. Default is `left`.
 
+aligny
+: (`string`) The vertical alignment of the text relative to the vertical offset, one of `top`, `center`, or `bottom`. Default is `top`.
+
 color
 : (`string`) The font color, either a 3-digit or 6-digit hexadecimal color code. Default is `#ffffff` (white).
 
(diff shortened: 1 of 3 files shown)


bep:
Thanks for this and sorry for the delay. I'm going to rebase and wrap this up in #13647.

Issue #13422, opened 2025-02-20
Term name includes plural taxonomy name as prefix


jasonccox:
When a taxonomy name includes a space, its terms which only have a single associated page have the plural taxonomy name prepended to their name when rendering.

This bug appears to have been introduced in https://github.com/gohugoio/hugo/pull/13064.

### Setup

`hugo.toml`:
```
[taxonomies]
"book author" = "book authors"
```

`content/posts/post1.md`:
```
---
title: My Post
tags: ["Tag One"]
book authors: ["Author One"]
---
```

`content/posts/post2.md`:
```
---
title: My Post
tags: ["Tag One", "Tag Two"]
book authors: ["Author One", "Author Two"]
---
```

`layouts/_default/taxonomy.html`
```
{{ range .Data.Terms.ByCount }}
Term: {{ .Term }}
Count: {{ .Count }}
Name: {{ .Name }}
Page title: {{ .Page.Title }}

{{ end }}
```

### Expected output in `public/book-authors/index.html`

```
Term: author one
Count: 2
Name: author one
Page title: Author One

Term: author two
Count: 1
Name: author two
Page title: Author Two

```

### Actual output in `public/book-authors/index.html`

```
Term: author one
Count: 2
Name: author one
Page title: Author One

Term: book authors/author two
Count: 1
Name: book authors/author two
Page title: Book Authors/Author Two

```

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.144.2+extended+withdeploy linux/amd64 BuildDate=unknown
</pre>

### Does this issue reproduce with the latest release?

Yes

jmooring:
A variation of this problem:
<https://discourse.gohugo.io/t/tags-being-added-to-beginning-of-term-title/53970>

PR #13423, opened 2025-02-20
Always strip taxonomy name prefix from term name

jasonccox:
If the taxonomy name differs when normalized versus unnormalized (e.g., because it has a space in it which is converted to a dash when normalized), the plural tree key will not match as a prefix to the term's base path (which is normalized). Use the non-normalized taxonomy name instead to ensure that the prefix always matches and is trimmed.

Fixes: gohugoio/hugo#13422


jmooring:
I have verified that this change addresses both of the cases described in #13422.

jasonccox:
Looks like this got fixed in https://github.com/gohugoio/hugo/pull/14518 instead.

Issue #13435, opened 2025-02-22
Container/Docker Images: Publish Version "withdeploy"


margau:
With #12994, the deploy functionality was hidden behind a feature flag.
The containers published on GHCR are currently built without the deploy feature flag, as far as I can tell.

I'd propose to either build and publish these containers with the deployment option by default, or build separate containers containing the "withdeploy"-version.

bep:
>I'd propose to either build and publish these containers with the deployment option by default.

Lets do that. PR welcome.

margau:
First try in #13435

bep:
This should be fixed.

PR #13436, opened 2025-02-22
build docker images with deploy capability

margau:
This PRs adds one variant of the functionality proposed in #13435: 

The Image is built using the "withdeploy" build tag, enabling deploy capability in all containers.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -46,4 +46,4 @@ jobs:
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          build-args: HUGO_BUILD_TAGS=withdeploy
+          build-args: HUGO_BUILD_TAGS=extended,withdeploy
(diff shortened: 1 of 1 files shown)


margau:
This change is currently untested, as I have no good way to properly test if the pipeline working as expected inside the fork.

bep:
This does not look like my comment in https://github.com/gohugoio/hugo/issues/13435#issuecomment-2676324034 ...

I would expect the change set to be ... smaller (e.g. add a tag to the existing build command).

margau:
> This does not look like my comment in [#13435 (comment)](https://github.com/gohugoio/hugo/issues/13435#issuecomment-2676324034) ...
> 
> I would expect the change set to be ... smaller (e.g. add a tag to the existing build command).

Sorry, misunderstood the comment - adjusted the PR accordingly, it just adds the build arg now - no separate image variants.

.github/workflows/image.yml
```diff
@@ -46,3 +46,4 @@ jobs:
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: HUGO_BUILD_TAGS=withdeploy
```
bep, line 49: The default value is `extended`, so this needs to be `extended,withdeploy`

bep (changes requested):

PR #13446, opened 2025-02-25
tpl: HTTPS the instagram Shortcode JS

coliff:
This JavaScript will always load over HTTPS so it's best to specify that. Protocol-less loaded resources are an anti-pattern (REF: https://www.paulirish.com/2010/the-protocol-relative-url/)


bep:
Can you fix the failing test?

coliff:
I'm not certain on what exactly to change. I see in the `shortcodes_integration_test.go` file that for the test the file content has a hash `a7937c49665872d3` so testing if changing that to `6e93404b93277876` works.

PR #13455, opened 2025-02-26
all: Typo fixes

coliff:


bep:
This is good, but it would be even better if you could get the commit message title in line with the guidelines, especially since the commit itself is about language correctness.

bep:
>hugolib folder, but did you mention at some point a while back that changes

That was back when I had a "full rewrite" of that package living in a branch, which would have made spelling fixes a real pain to merge in.

Also, to fix the commit message you need to amend your commit (the PR title is a different title). I could certainly do it when merging this, but it would save some future hassle to get the setup ... right.

coliff:
ok sounds good. I'll add those others I found in `/hugolib` and amend the commit to match the PR title `all: Typo fixes` - is that ok?

PR #13531, opened 2025-03-26
Fix validator errors

gideonstar-git:
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#allowfullscreen 
* Validator complains about mozallowfullscreen and webkitallowfullscreen, allow="fullscreen" is the right way to do it


bep:
Thanks for this. Can you make the TestVimeoShortcode go green by updating the expected hash by what's printed when running test.

bep:
To save some time, I'm going to merge #13532 instead, but thanks for this, much appreciated.

Issue #13146, opened 2024-12-14
XMP and IPTC image metadata support


LNA-DEV:
Hej. First of all thanks for this amazing project!

As said in #12651 the new metadata package supports XMP and IPTC metadata. Would be really nice to have this feature therefore I created this issue.

Is there any update / plan on this topic?

bep:
>Is there any update / plan on this topic?

Not really. The main challenge with this, as I remember back when I implemented the new library, was how to represent all these tags in a data structure. This gets incredible confusing when you start looking into it; grouping the tags by their source (EXIF, XMP, IPTC) may be fine in Adobe Lightroom, but I'm not sure it's very user friendly from Hugo ... But as I said, I have not thought about this lately.

LNA-DEV:
I fully get your point. It really gets complicated if you think about it. 

I thought maybe we could do some sort of guessing logic (for the most-used properties) **and** include a raw section so the user can, if they want to, fall back to it. But I’m not sure if a guessing logic of any sort is a good idea. You could do something like checking EXIF first and if null, checking IPTC and if that’s also null, taking the XMP property. But this could get a bit confusing if there are different values set in those sources. If well-documented and communicated, though, this might be a reasonable compromise. 

But maybe you or someone else has a better idea^^

Anyway, I’m looking forward to it :smile:

LNA-DEV:
I have experimented with this a bit today. I think it would be most useful if we do something like I did in this pull request: https://github.com/gohugoio/hugo/pull/13539

This way existing pages still work the same but the user can find additional tags in the Exif object under Tags. (I am not sure here if IPTC and XMP should be place together with EXIF under a own Metadata object. But I think this could lead to confusion and could break support for existing sites. So I did not do it that way.)

Under Tags are all tags from every metadata category with the priority Exif > IPTC > XMP but if someone wants a specific tag from a specific category this is still possible because of the "raw" lists of tags.

solonovamax:
> This way existing pages still work the same but the user can find additional tags in the Exif object under Tags. (I am not sure here if IPTC and XMP should be place together with EXIF under a own Metadata object. But I think this could lead to confusion and could break support for existing sites. So I did not do it that way.)

honestly, I think something like the following would work (and doesn't have any backwards compat issues):
```go
// ExifInfo holds the decoded Exif data for an Image.
type ExifInfo struct {
	// GPS latitude in degrees.
	Lat float64
	// GPS longitude in degrees.
	Long float64
[...]

PR #13539, opened 2025-03-29
exif: Added IPTC and XMP metadata to the EXIF object

LNA-DEV:
Under the Exif Tags are now all tags from every metadata category with the priority Exif > IPTC > XMP but if someone wants a specific tag from a specific category this is still possible because of the "raw" lists of tags.

#13146

### DIFF
--- a/resources/images/exif/exif.go
+++ b/resources/images/exif/exif.go
@@ -37,8 +37,17 @@ type ExifInfo struct {
 	// Image creation date/time.
 	Date time.Time
 
-	// A collection of the available Exif tags for this Image.
+	// A collection of the available Exif, Iptc and Xmp tags for this Image.
 	Tags Tags
+
+	// A collection of the available Exif tags for this Image.
+	Exif Tags
+
+	// A collection of the available Iptc tags for this Image.
+	Iptc Tags
+
+	// A collection of the available Xmp tags for this Image.
+	Xmp Tags
 }
 
 type Decoder struct {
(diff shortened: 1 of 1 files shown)


resources/images/exif/exif.go
```diff
@@ -210,17 +218,44 @@ func (d *Decoder) Decode(filename string, format imagemeta.ImageFormat, r io.Rea
...
+		if !d.shouldInclude(k) {
+			continue
+		}
+		tagsXmp[k] = v.Value
+		tags[k] = v.Value
+	}
+
+	tagsIptc := make(map[string]any)
+	for k, v := range tagInfos.IPTC() {
+		if d.shouldExclude(k) {
```
bep, line 236: I need to think harder about this full PR, but

* I think these shouldInclude/exclude checks is something that's mistakingly left over from the old setup.
* We apply the same filter much closer to the source in the `Decode` method (see https://github.com/gohugoio/hugo/blob/6f14dbe24c2c951ff5cf70986170d87aa56cb03e/resources/images/exif/exif.go#L150)

LNA-DEV: I tested it and it seems like you are right. Still works as expected :+1:

shombando:
Just wanted to express interest in and support for this functionality. Image metadata between exif, iptc, xmp is a hot mess in the standards and how applications use them, and I really appreciate you tackling this thorny but useful issue. Thanks for your work!

LNA-DEV:
#14378 fixed it

PR #13542, opened 2025-03-30
deps: Upgrade github.com/gohugoio/hugo-goldmark-extensions/passthrough v0.3.0 => v0.3.1

jmooring:


bep:
In the future, it would be better if you created these release tags via a the GitHub release feature, so they show up here:

https://github.com/gohugoio/hugo-goldmark-extensions/releases

If this is a permission issue, let me know and I'll have a look.

PR #13551, opened 2025-04-01
Docs: Update quick-start.md

saraswativ23:
Add "sequentially" to "Run these commands to create a Hugo site" so it becomes "Run these commands sequentially to create a Hugo site".


jmooring:
Thank you, but no.

PR #13585, opened 2025-04-11
common/paths: support `.gotmpl` outer filetype extension

### LINKED ISSUE #10449: Support `.gohtml` file extension for templates
For partials and shortcodes it is already possible with the current version of Hugo to use the file extension `.gohtml` instead of `.html`. It would be great if the file extension `.gohtml` could also be used for templates.

The advantage of `.gohtml` instead of `.html` is better syntax highlighting for Golang in many IDEs like IntelliJ, without the need for manual hacks.

jenbroek:
This allows for text editors to (more easily) detect both the outer (e.g. `gotmpl`) and inner (.e.g `html`) syntax used in templates.

A `.gotmpl` extension is considered an outer filetype and ignored in template resolution when:
* It is not recognized as a Kind, Language, or OutputFormat
* It is not the last extension (as this signifies a content adapter)

For example:
* `layouts/home.gotmpl.html` -> same as `layouts/home.html`
* `layouts/_shortcodes/foo.gotmpl.html` -> `{{< foo >}}`
* `layouts/_partials/foo.gotmpl.html` -> `{{ partial "foo.html" }}

Note that if both an e.g. `layouts/home.html` and `layouts/home.gotmpl.html` exist, the latter takes precedence. Same with shortcode templates.

When using the embedded webserver, later writes to e.g. `layouts/home.html` *will* reflect in the stored template.

Closes #10449

---

If this is a worthwhile change, any feedback is appreciated 🙂

I thought on whether to have `Path()` remove the outer filetype extension or instead add a separate method `PathNoOuterftypeExt()` and update references to `Path()` where needed. Let me know if a separate method is preferrable.

### DIFF
--- a/common/paths/pathparser.go
+++ b/common/paths/pathparser.go
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	identifierBaseof = "baseof"
+	identifierBaseof           = "baseof"
+	identifierGotmplOuterftype = "gotmpl"
 )
 
 // PathParser parses a path into a Path.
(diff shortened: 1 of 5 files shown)


jmooring:
https://github.com/gohugoio/hugo/issues/10449 is a proposal. The proposal has not been accepted. 

From this project's README.md:

From this project's CONTRIBUTING.md:

> A non-trivial code change without an issue reference with one of the labels bug or enhancement will not be merged.

From https://gohugo.io/contribute/development/:

PR #13623, opened 2025-04-18
Capitalize UTF in sitemap.xml and sitemapindex.xml

nocanoa:
Hey,

I was using the default sitemap.xml and google search console kept giving me a error but not saying what.
After some testing I concluded that the `utf-8` was not uppercase, after changint that google search console concluded whit no errors.

Also cheked other sitemaps and all have it uppercase. Its also shown as `encoding="UTF-8"` on https://www.sitemaps.org/protocol.html

Thanks ✌️

### DIFF
--- a/tpl/tplimpl/embedded/templates/rss.xml
+++ b/tpl/tplimpl/embedded/templates/rss.xml
@@ -30,7 +30,7 @@
 {{- if ge $limit 1 }}
 {{- $pages = $pages | first $limit }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


nocanoa:
@jmooring Also found it weird, that it would be the uppercase, just did some more testing, and it seems that google doesn't like the name itself /sitemap.xml because if i give it any other name /sitemap-website.xml ; /sitemap-site.xml etc it succeeds.

So is it related to the file name it self?

jmooring:
> So is it related to the file name it self?

No. Maybe there's an invalid cached version of your sitemap.xml file somewhere. 

Regardless, let's handle this as a support issue. Please create a forum [topic](https://discourse.gohugo.io/).

PR #13695, opened 2025-05-07
tpl/math: Add MaxInt64 function

### LINKED ISSUE #13693: tpl/math: Add MaxInt64 and MinInt64 functions
Now that we have a `break` statement it would be handy (in some rare scenarios) if we could do:

```text
{{ range }} <----------- do this forever until break
  {{ if $someBoolean }}
    {{ break }}
  {{ end }}
{{ end }}
```

But you can't do that... `range` requires an argument.

The text/template package in Go 1.22 and later supports range over int, so you can now do this:

```text
{{ range 1000000000 }}  <----------- do this many times until break
  {{ if $someBoolean }}
    {{ break }}
  {{ end }}
{{ end }}
```

It would be cleaner if we could do this:

```text
{{ range MaxInt64 }}  <----------- do this 9223372036854775807 times until break
  {{ if $someBoolean }}
    {{ break }}
  {{ end }}
{{ end }}
```

It's not infinite, but it's close enough for most cases that I can envision.

```go
// MaxInt64 returns the maximum value for a signed 64-bit integer.
func (ns *Namespace) MaxInt64() int64 {
	return math.MaxInt64
}

// MinInt64 returns the minimum value for a signed 64-bit integer.
func (ns *Namespace) MinInt64() int64 {
	return math.MinInt64
}
```

jmooring:
Closes #13693

### DIFF
--- a/tpl/math/init.go
+++ b/tpl/math/init.go
@@ -115,6 +115,13 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.MaxInt64,
+			nil,
+			[][2]string{
+				{"{{ math.MaxInt64 }}", "9223372036854775807"},
+			},
+		)
+
 		ns.AddMethodMapping(ctx.Min,
 			nil,
 			[][2]string{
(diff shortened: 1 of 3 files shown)


Aravind-blip (commented):
Clean and well-scoped feature addition. MaxInt64 is correctly wired to templates, returns the proper type, and includes dedicated tests. Small change, but complete and verifiable.

PR #13726, opened 2025-05-21
Add hosting support for ENS & IPFS

rrbin666:


bep:
@jmooring I will look at this later. I guess our unwritten "notability rules" would exclude this hosting provider, but I have talked to these people via mail, and I'm mildly curious about this web3 "thing".

bep:
OK, I didn't notice which repo we're in. @rrbin666 you need to move this PR to https://github.com/gohugoio/hugoDocs/pulls

The /docs folder in this repo is a Git sub tree. We do two-way-merges, but this repo is only for changes to docs as part of a code change.

PR #13742, opened 2025-05-26
parser/pageparser: Add coverage for all IsX methods of Item

pixel365:
This PR adds unit tests for all `IsX` boolean methods of the `Item` struct in the `parser/pageparser` package.


bep:
Thanks for this, this looks great in isolation, but I need to think a little about whether I want to merge these. If this package had been a standalone library, these would add great value. But running `go test -coverprofile=cover.out -coverpkg=./... ./...` from the root of this project, all of the Is* methods are covered (in realistic tests). The ToString method is not covered, but that is a "development tool".

PR #13746, opened 2025-05-27
common/collections: Increase test coverage

pixel365:
This PR improves test coverage for the `common/collections` package.

### DIFF
--- a/common/collections/stack_test.go
+++ b/common/collections/stack_test.go
@@ -1,0 +1,77 @@
+package collections
+
+import (
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+)
+
+func TestNewStack(t *testing.T) {
+	t.Parallel()
+	c := qt.New(t)
+
+	s := NewStack[int]()
+
+	c.Assert(s, qt.IsNotNil)
+}
+
+func TestStackBasic(t *testing.T) {
+	t.Parallel()
+	c := qt.New(t)
+
+	s := NewStack[int]()
+
+	c.Assert(s.Len(), qt.Equals, 0)
+
+	s.Push(1)
+	s.Push(2)
+	s.Push(3)
+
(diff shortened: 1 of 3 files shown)


common/collections/stack_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	s := NewStack[string]()
+	s.Push("a")
+	s.Push("b")
+
+	got := s.Drain()
+
+	c.Assert(len(got), qt.Equals, 2)
+	c.Assert(got[0], qt.Equals, "a")
+	c.Assert(got[1], qt.Equals, "b")
```
bep, line 60: Nit-picking, but the 3 lines above can be collapsed into one using DeepEquals.

common/collections/stack_test.go
```diff
@@ -0,0 +1,89 @@
...
+	s.Push(1)
+	s.Push(2)
+	s.Push(3)
+	s.Push(4)
+
+	got := s.DrainMatching(func(v int) bool { return v%2 == 0 })
+
+	c.Assert(got, qt.DeepEquals, []int{4, 2})
+
+	wantRemaining := []int{1, 3}
```
bep, line 78: I suggest you just use `c.Assert(s.Drain(), qt.DeepEquals([]int{4, 2})`

common/collections/stack_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+)
+
+func TestNewStack(t *testing.T) {
+	s := NewStack[int]()
+
+	if s == nil {
+		t.Fatal("expected non-nil Stack")
```
bep, line 13: Use `qt` to assert.

bep (changes requested):

PR #13791, opened 2025-06-08
tpl: add full post content in rss

vaaleyard:
If I have to do anything else please say and I will happy to help.


jmooring:
The corresponding proposal, https://github.com/gohugoio/hugo/issues/4071, has not been accepted by the project lead. For now, just override the RSS template. If you need assistance with that, please create a new topic in the [forum](https://discourse.gohugo.io/).

PR #13796, opened 2025-06-11
deps: Upgrade github.com/spf13/cast v1.8.0 => v1.9.2

jmooring:
Notes about test changes:

- The `math.Mod` and `math.ModBool` template functions now handle floats represented as strings. See [spf13/cast/pull/261](https://github.com/spf13/cast/pull/261). 

- Although I didn't spend a lot time digging through our code, I was unable to determine which `spf13/cast` change was responsible for the required change to our `TestLoadConfigYamlEnvVar` test, where we now get an empty slice instead of `nil`. I verified that this change does not affect which files are published.  Or to put it another way, this still works great:

  ```text
  HUGO_OUTPUTS="section: []" hugo
  ```

### DIFF
--- a/tpl/math/math_test.go
+++ b/tpl/math/math_test.go
@@ -258,7 +258,7 @@ func TestMod(t *testing.T) {
 		{int32(3), int32(2), int64(1)},
 		{int64(3), int64(2), int64(1)},
 		{"3", "2", int64(1)},
-		{"3.1", "2", false},
+		{"3.1", "2", int64(1)},
 		{"aaa", "0", false},
 		{"3", "aaa", false},
 	} {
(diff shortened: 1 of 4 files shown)


tpl/math/math_test.go
```diff
@@ -258,7 +258,7 @@ func TestMod(t *testing.T) {
 		{int32(3), int32(2), int64(1)},
 		{int64(3), int64(2), int64(1)},
 		{"3", "2", int64(1)},
-		{"3.1", "2", false},
+		{"3.1", "2", int64(1)},
```
bep, line 261: While `1` may seem more correct than an error, I would expect a float (1.1) as the result here?

bep: Never mind, I see that Mod returns an int64.

PR #13805, opened 2025-06-17
command: `hugo gen jsonschemas`

### LINKED ISSUE #10984: Add a JSON config schema
The ones I find in VS Code are incredibly outdated:

image

bmackinney:
Adds a hugo gen jsonschemas subcommand that generates JSON Schema files for Hugo's configuration options and page front matter using Go reflection.

- 32 schema files: 30 per-section schemas + root hugo-config.schema.json (with $ref to sections) + hugo-content.schema.json for front matter
- Reflection-driven: walks allconfig.Config fields at generation time — no hard-coded lists
- Doc links: crawls generated Hugo docs HTML to attach documentation URLs to schema properties

Fixes gohugoio#10984

> AI Assistance Notice: This PR was written with substantial assistance from GitHub Copilot and OpenCode (Big Pickle model). The contributor is not a Go developer and relied on AI assistance for implementation.


bmackinney:
@bep this is ready for you to review.

bep:
@bmackinney thanks for this. We have been doing a fair bit of work lately to get rid of/clean up some legacy APIs -- so this has not been a priority. After the next minor Hugo version I think we're more or less in a "stable API" situation, and I will have a closer look at this. And yes, I do agree that, especially for config, the VS code situation is less than optimal.

bep:
I will test this myself, but one quick question: The lower casing of the first char in `Page` methods/properties, is that correct (or: does that work?).

bmackinney:
It works and is correct for json schemas, but I'm going to improve the camel function to handle acronyms better (ID → id instead of iD, JS → js instead of jS).

bep:
@bmackinney I have tried to test this in a Netlify docs branch, but I have 2 issues:

* It would be good for testing if either the `$ref` URLs were relative (not sure if that's possible) or if I could somehow set `--baseURL` when running the `hugo getn jsonschemas`. For testing I have done a "search and replace", which I guess works, but that's a little ... cumbersome.
* But after doing that, I get "Error when resolving ...." e.g.  `hugo-config-params.schema.json`. It's referenced, but the JSON file is not generated.

bmackinney:
@bep it's ready to re-test.

bep:
@jmooring I will do a more thorough test/review of this setup., but I would appreciated your input on structure/naming. I'm pretty convinced that this is worth adding, especially considering the poor state of the existing "Hugo schemas" out there. 

The structure of this PR is currently:

* config
    * markup
    * ... other config sections 
* page (which is page front matter)

I would think that we at least would want to have an option in the future to add JSON schemas for Hugo's template APIs, e.g. `page` (which conflicts with the above), `site`, `hugo`.

I'm not an JSON schema expert, but I suspect that it would be great if the structure somehow followed the documentation about the same topics.

NoClueMike (approved):

bmackinney:
@bep I renamed the hugo-pages.schema.json to hugo-content.schema.json. I think the rest might be ready to go. it catches version and role now as well.

PR #13816, opened 2025-06-22
markup/goldmark: Change link and image render hook enablement to enums

### LINKED ISSUE #13535: Allow embedded link and image render hooks to override others
#### Example

Let's say that a theme provides image and link render hooks that are not compatible with multilingual single-host sites. Enabling the embedded image and link render hooks in the config does nothing, as designed:

```text
[markup.goldmark.renderHooks.image]
enableDefault = true

[markup.goldmark.renderHooks.link]
enableDefault = true
```

Today, site authors have to override the theme's files in the usual fashion, copying code from:

- https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_default/_markup/render-image.html
- https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_default/_markup/render-link.html

#### Proposal

Provide a config option to ALWAYS use the embedded hook, overriding those provided by themes, modules, and in the project root. For example:

```text
[markup.goldmark.renderHooks.image]
enableDefault = true
override = true

[markup.goldmark.renderHooks.link]
enableDefault = true
override = true
```

I'm not in love with the option name.

In addition to the example above, perhaps the most common reason to enable the embedded image and link render hooks is to handle sites that are served from subdirectories (e.g., almost every GitHub Pages site).

jmooring:
Closes #13535

How this is supposed to work:

- [`renderHooks.image.useEmbedded`](https://deploy-preview-3124--gohugoio.netlify.app/configuration/markup/#renderhooksimageuseembedded)
- [`renderHooks.link.useEmbedded`](https://deploy-preview-3124--gohugoio.netlify.app/configuration/markup/#renderhookslinkuseembedded)

### DIFF
--- a/tpl/tplimpl/render_hook_integration_test.go
+++ b/tpl/tplimpl/render_hook_integration_test.go
@@ -192,3 +192,115 @@ iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAA
 		`<img src="/dir/p1/pixel.png" alt="alt4" id="&#34;&gt;&lt;script&gt;alert()&lt;/script&gt;">`,
 	)
 }
+
+// Issue 13535
+func TestEmbeddedLinkAndImageRenderHookConfig(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+disableKinds = ['home','rss','section','sitemap','taxonomy','term']
+
+[markup.goldmark]
+duplicateResourceFiles = false
+
+[markup.goldmark.renderHooks.image]
+#KEY_VALUE
+
+[markup.goldmark.renderHooks.link]
+#KEY_VALUE
+
+#LANGUAGES
+-- content/s1/p1/index.md --
+---
+title: p1
+---
+[p2](p2)
+
+[a](a.txt)
+
+![b](b.jpg)
+-- content/s1/p1/a.txt --
+-- content/s1/p1/b.jpg --
+-- content/s1/p2.md --
+---
+title: p2
+---
+-- layouts/all.html --
+{{ .Content }}
+`
+
+	const customHooks string = `
+-- layouts/s1/_markup/render-link.html --
+custom link render hook: {{ .Text }}|{{ .Destination }}
+-- layouts/s1/_markup/render-image.html --
+custom image render hook: {{ .Text }}|{{ .Destination }}
+`
+
+	const languages string = `
+[languages.en]
+[languages.fr]
+`
+
+	const (
+		fileToCheck         = "public/s1/p1/index.html"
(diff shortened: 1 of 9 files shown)


bep:
I will have a closer look at this tomorrow. A side note, I'm planning on doing a `v0.148.0` release on Monday with this and the  other recent changes.

tpl/tplimpl/render_hook_integration_test.go
```diff
@@ -192,3 +193,112 @@ iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAA
...
+			f = strings.ReplaceAll(f, "#LANGUAGES", languages)
+		}
+		if tt.hasCustomHooks {
+			f = f + customHooks
+		}
+		f = strings.ReplaceAll(f, "#KEY_VALUE", tt.keyValuePair)
+
+		b := hugolib.Test(t, f)
+		b.AssertFileContent(fileToCheck, tt.want)
+		fmt.Printf("Test %s: pass\n", tt.id)
```
bep: Remove printf.

tpl/tplimpl/templatestore.go
```diff
@@ -1048,7 +1052,9 @@ func (s *TemplateStore) insertEmbedded() error {
 					return err
 				}
 			} else {
-				ti, err = s.insertTemplate(pi, nil, SubCategoryEmbedded, false, s.treeMain)
+				replace := (name == path.Join(containerMarkup, "render-image.html") && s.opts.RenderHooks.Image.UseEmbedded == "always") ||
+					(name == path.Join(containerMarkup, "render-link.html") && s.opts.RenderHooks.Link.UseEmbedded == "always")
+				ti, err = s.insertTemplate(pi, nil, SubCategoryEmbedded, replace, s.treeMain)
 				if err != nil {
```
bep: I suspect that there's a missing (and failing) test case for this. I'm not sure it's worth spending time fixing (because I don't think the fix is trivial), but, when running the server and `useEmbedded=auto` for image hooks and the user edits the hook template in the project.

bep: OK, thinking a little more about the above: What about `layouts/mysection/_markup/render-image.html`? I think this logic needs to be moved to where the user provided templates gets inserted.

jmooring:
> layouts/mysection/_markup/render-image.html

I did not consider higher-specificity user-defined templates when conceptualizing how this would work. To clarify, when `renderHooks.link.useEmbedded` is set to `always`, we must _always_ use the embedded hook even if a higher-specificity user-defined template exists. I will fix this and add tests.

The `auto`, `always`, `fallback`, and `never` strings are used in enough places that I'm going to convert these to constants when used in a goldmark context.

PR #13817, opened 2025-06-23
Deploy, walkLocal performance

davidejones:
### What is the purpose?

During hugo deploy we run `walkLocal` which iterates over the filesystem. At larger scale deploys I've noticed this can be slow. 
This PR attempts to improve the performance by using a worker pool

### Why?

To help improve deploy times 

### Example

Before my changes (just log changes)

After my changes

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -36,6 +36,7 @@ import (
 	"github.com/dustin/go-humanize"
 	"github.com/gobwas/glob"
 	"github.com/gohugoio/hugo/common/loggers"
+	"github.com/gohugoio/hugo/common/para"
 	"github.com/gohugoio/hugo/config"
 	"github.com/gohugoio/hugo/deploy/deployconfig"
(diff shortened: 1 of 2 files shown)


deploy/deploy.go
```diff
@@ -487,7 +487,71 @@ func knownHiddenDirectory(name string) bool {
...
+
+	// Start workers
+	for i := 0; i < workerCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for path := range jobs {
+				info, err := fs.Stat(path)
+				if err != nil {
+					errs <- err
```
bep: I think the behaviour of the old code was to stop early on errors, and to me it makes sense to keep that.

Also see the `para.Workers` type in this repo that I think makes this pattern easier to read.

davidejones: Thanks! i've tried updating this now

PR #13865, opened 2025-07-22
markup/org: add custom org link resolve function for `denote`

theSuess:
The go-org library used by hugo to parse `.org` files now supports custom link resolvers. This PR uses the new functionality to add support for [denote](https://protesilaos.com/emacs/denote) style links.

This can be used to publish notes as-is using hugo.

I've written about the details in a blog post [here](https://dominik.suess.wtf/posts/2025/01/02/denote-hugo/).

I don't know if this functionality is too specific to my use case, so I'm happy to discuss alternatives which could be more modular/adaptable to other systems.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,7 +54,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20
 	github.com/mitchellh/mapstructure v1.5.1-0.20231216201459-8508981c8b6c
 	github.com/muesli/smartcrop v0.3.0
-	github.com/niklasfasching/go-org v1.8.0
+	github.com/niklasfasching/go-org v1.9.1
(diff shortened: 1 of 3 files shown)


jmooring:
@theSuess In this project's README in the "Contributing" section:

...and there isn't a proposal. Please create one.

Although it's not my call, I do not object to making this change. With a more mainstream markup language (e.g.,  Markdown) I would want to abstract this a bit (e.g., add a config settting or hook), but in this case I think hardcoding the protocol is fine.

theSuess:
`CONTRIBUTING.md` doesn't mention proposals at all, so I didn't realize this was needed. I've got a better idea on how to implement this anyway - will open a proposal for that once I know if it's feasible. Thanks for getting back to me on this!

PR #13866, opened 2025-07-23
refactor: use slices.Equal to simplify code

minxinyi:
In the Go 1.21 standard library, a new function has been introduced that enhances code conciseness and readability. It can be find  [here](https://pkg.go.dev/slices@go1.21.1#Equal).

### DIFF
--- a/navigation/menu_cache.go
+++ b/navigation/menu_cache.go
@@ -28,7 +28,7 @@ func (entry menuCacheEntry) matches(menuList []Menu) bool {
 		return false
 	}
 	for i, m := range menuList {
-		if !menuEqual(m, entry.in[i]) {
+		if !slices.Equal(m, entry.in[i]) {
 			return false
 		}
 	}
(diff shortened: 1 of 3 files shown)


navigation/menu_cache.go
```diff
@@ -48,17 +48,7 @@ type menuCache struct {
...
-
-	for i := range m1 {
-		if m1[i] != m2[i] {
-			return false
-		}
-	}
-
-	return true
+	return slices.Equal(m1, m2)
 }
```
bep: Thanks for this. This would be even better if you got rid of the menuEqual and compareStringSlices funcs (they're only used in a few places.

minxinyi: Modified. Please review again.

dilyanpalauzov:
navigation/menu_cache_test.go uses the now deleted `menuEqual()`.

minxinyi:
> navigation/menu_cache_test.go uses the now deleted `menuEqual()`.

Oh, Sorry. 

Thank you for your review; I have searched and replaced all instances.

minxinyi:
It seems that the CI error is unrelated to this change.

PR #13882, opened 2025-07-30
restore older uglyurls behavior with configuration setting

n1xx1:
After #13835 UglyURLs are now working as intended (?) but it did break my site that was relying on uglyUrls and [this](https://vercel.com/docs/project-configuration#cleanurls) on vercel. With this change there is a new configuration key (`sectionUglyURLs`, let me know if you want a different name) to restore the previous behavior. I've also added the needed tests for the feature.


jmooring:
Thanks for this, but a configuration setting to restore buggy behavior isn't going to happen. In the future, per our contribution guidelines, please create a proposal before spending time on a PR.

PR #13885, opened 2025-07-31
implement update function for slices and maps

### LINKED ISSUE #7232: Add `update` template funcs to change maps and slices
I once wrote a proposal (https://github.com/gohugoio/hugo/issues/5604) for a complex data manipulation API which I now find unnecessary except for the ability to build on maps.

For now the only way to build on maps is either using `.Scratch` (SetInMap) or the [merge](https://gohugo.io/functions/merge/#readout) function which tends to be much slower.

<del>A problematic data race (https://github.com/gohugoio/hugo/issues/7230#issuecomment-622334415) was recently surfaced on `.Scratch.SetInMap`. Applying a patch for it might slow the build of other more common usage.</del>

This proposal explores building on maps without Scratch.

## Proposal

Introduce a global function to build on maps

### 1. Use the existing `append` function 
It is currently reserved for slices. This means the improved `append` would test for the input data type, and upon finding a `map`, expect 3 arguments instead of 2:
- initial map
- key
- value

If key already exists, it would be overwritten.

```
{{ $data := dict }}
{{ $data = append $data "key" "value" }}
```

### 2. A new function, yet to be named 
It would take 3 parameter
- initial map
- key
- value

If key already exists, it would be overwritten.

```
{{ $data := dict }}
{{ $data = appendToMap $data "key" "value" }}
```
### 3. `keyVals`

I know we use a [`keyVals` function](https://gohugo.io/content-management/related/#relatedto-keyvals-keyvals2-) (undocumented outside of Related Content) I have no idea if this could be useful in this context, but some might chime in.

## Also
This would remove the use of `.Scratch.SetInMap` and allow the future retirement of `.Scratch`. This was a great feature, but a workaround to now lifted Go Template limitations.

n1xx1:
Fixes #7232.

I didn't add an alias, for now, because there was some discussion about using `update`, `insert`, `put` or something else.

```
{{ $dict := dict "a" 1 "b" 2 }}
{{ $dict | collections.Update "c" 3 }}
{{/* $dict is { "a": 1, "b": 2, "c": 3 } */}}
```

```
{{ $slice := slice 1, 2, 3 }}
{{ $slice | collections.Update 1 100 }}
{{/* $slice is [1, 100, 3] */}}
```

### DIFF
--- a/tpl/collections/init.go
+++ b/tpl/collections/init.go
@@ -202,6 +202,11 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.Update,
+			[]string{},
+			[][2]string{},
+		)
+
 		return ns
 	}
 
--- a/tpl/collections/update.go
+++ b/tpl/collections/update.go
@@ -1,0 +1,64 @@
+package collections
+
+import (
+	"fmt"
+	"github.com/gohugoio/hugo/common/types"
+	"github.com/spf13/cast"
+	"reflect"
+)
+
+func (ns *Namespace) Update(index any, value any, mapOrSlice any) (string, error) {
+	vmapOrSlice := reflect.ValueOf(mapOrSlice)
+
+	vmapOrSlice, isNil := indirect(vmapOrSlice)
+	if isNil {
+		return "", fmt.Errorf("can't update a nil value")
+	}
+
+	var err error
+	if vmapOrSlice.Kind() == reflect.Map {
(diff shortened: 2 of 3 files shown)


bep:
Thanks for this. I will close this for now. We need to have an open issue that describes how this should look. I don't think we can add template functions that modifies slices and maps, because that is 1) Not thread safe and 2) Would create many potential surprises (e.g. if you modify `site.RegularPages`). But anyhow, it needs to start with a proposal.

PR #13888, opened 2025-08-03
Add support for "format" option in transform.Unmarshal

### LINKED ISSUE #13887: transform.Unmarshal should allow to specify a format
Currently `transform.Unmarshal` allows to either provide a string or a resource. In the first case the format is inferred from the string, but it may be more useful to be able to specify it. For example json has precedence over yaml, so something like `{p: [a, b]}` would fail since it’s not valid json but is valid yaml. My current solution for this is to make a temporary resource with a .yaml extension and pass that to Unmarshal. This solution is needlessly complicated and more expensive.

I propose to add a "format" key to the option dict so that it can be used like this:

```
{{ $opts := dict "format" "yaml" }}
{{ $parsed := "{p: [a, b]}” | unmarshal $opts }}
```

n1xx1:
Fixes #13887.

The "format" option works with string and resources (in the rare case you want to override the mime type). If it's not recognized the unmarshal fails.

### DIFF
--- a/tpl/transform/unmarshal.go
+++ b/tpl/transform/unmarshal.go
@@ -72,9 +72,17 @@ func (ns *Namespace) Unmarshal(args ...any) (any, error) {
 		}
 
 		v, err := ns.cacheUnmarshal.GetOrCreate(key, func(string) (*resources.StaleValue[any], error) {
-			f := metadecoders.FormatFromStrings(r.MediaType().Suffixes()...)
-			if f == "" {
-				return nil, fmt.Errorf("MIME %q not supported", r.MediaType())
+			var f metadecoders.Format
+			if decoder.Format != "" {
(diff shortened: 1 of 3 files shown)


tpl/transform/unmarshal.go
```diff
@@ -117,12 +125,20 @@ func (ns *Namespace) Unmarshal(args ...any) (any, error) {
 		return nil, nil
 	}
 
-	key := hashing.MD5FromStringHexEncoded(dataStr)
+	key := hashing.MD5FromStringHexEncoded(dataStr) + decoder.Format
```
bep: This looks like a caching bug in the old code (the code duplication here is slightly unfortunate, but lets leave that for now).

I suggest that you do similar to the above:

```go
if decoder != metadecoders.Default {
	key += decoder.OptionsKey()
}
```

PR #13892, opened 2025-08-07
partialCached now considers the key for deadlock prevention

### LINKED ISSUE #13889: Regression in infinite recursion prevention of partialCached

<pre>
$ hugo version
hugo v0.148.2-40c3d8233d4b123eff74725e5766fc6272f0a84d+extended windows/amd64 BuildDate=2025-07-27T12:43:24Z VendorInfo=gohugoio
</pre>

Yes

### Description

With the recursion prevention in #13631, valid recursive partials now fail. This is a simple repro: https://github.com/n1xx1/repro-hugo-infinite-recursion

If in the repro partialCached here (https://github.com/n1xx1/repro-hugo-infinite-recursion/blob/main/layouts/_shortcodes/include.html#L3) is replaced with partial, it works just fine.

Maybe adding the key to the check if it's a circular call could prevent the issue while keeping the detection.

n1xx1:
Fixes #13889.

### DIFF
--- a/tpl/partials/partials.go
+++ b/tpl/partials/partials.go
@@ -126,7 +126,7 @@ func (ns *Namespace) include(ctx context.Context, name string, dataList ...any) 
 	if err != nil {
 		return includeResult{err: err}
 	}
-	return ns.doInclude(ctx, v, dataList...)
+	return ns.doInclude(ctx, "", v, dataList...)
 }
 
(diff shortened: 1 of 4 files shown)


bep:
This looks great; a heads up re. commit messages for the future: https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines

n1xx1:
My bad, I will use the correct messages next time.

PR #13908, opened 2025-08-14
resources/page: Add :sectionslug and :sectionslugs permalink tokens

### LINKED ISSUE #13788: Implement a :sectionslug and a :sectionslugs permalink token
The URL path elements in a published Hugo site typically mirror the directory and file names of the content hierarchy within the `content/` directory, however, this is not always desirable.   There are legitimate reasons for wanting these structures to differ. 

It is easy to modify the final filename element of the URL using the `slug` front matter parameter but it is not possible to modify the naming of individual sections in a similar way. 

We propose adding `:sectionslug` and `:sectionslugs` permalink tokens (in symmetry to the existing [`:section`](https://gohugo.io/configuration/permalinks/#section) and [`:sections`](https://gohugo.io/configuration/permalinks/#sections) tokens) that modifies a section's URL path element by using the `slug` or `title` parameters as defined in a section's front matter (in `_index.md` file).  `:sectionslugs` will traverse the hierarchy iteratively to create the URL path, resulting in a slugified path

> **NOTE**: The [`url`](https://gohugo.io/content-management/urls/#url) field in the front matter can currently be used to override the entire path.  This approach becomes error prone when handling nested content section hierarchies as care must be taken to propagate any changes to a section's `url` to its subsections. 

## Example `:sectionslugs`

Given the content structure:

```
content/                           
└── 01-parent-section/
     ├── _index.md   # slug: my-parent-section-slug ; title: "My Parent Section Title"
     ├── 01-article.md			
     └── 02-child-section/
         ├── _index.md   # title: "My Child Section"
         └── 01-nested-article.md # title: "My Nested Article"
```

And the following permalink configuration using the proposed `:sectionslugs` token:

```
# config.yaml
...
permalinks:
  page: 
    01-parent-section: /:sectionslugs/:slug
  section:
	01-parent-section: /:sectionslugs/:slug
```

The published site structure would be:

```
public/                           
└── my-parent-section-slug/
   ├── index.html   
   └── my-child-section/ 
       ├── index.html
       └── my-nested-article/ 
           └── index.html
```

We have already written and tested the code for this on multiple Hugo sites, and can make it available on request.

See the forum discussion [here](https://discourse.gohugo.io/t/support-a-sectionslug-permalink-token/55005/3)

DustinFischer:
Add new permalink tokens allowing slugified sections in the URL path, with fallback behavior and slice syntax support.

- `:sectionslug`: the content's section using the section's `slug` frontmatter parameter, else falling back to its `title` parameter.
- `:sectionslugs`: the content's sections hierarchy, using each section's respective `slug` frontmatter parameter, else falling back to its `title` parameter.

Includes integration tests and documentation updates.

Closes https://github.com/gohugoio/hugo/issues/13788

### DIFF
--- a/resources/page/permalinks.go
+++ b/resources/page/permalinks.go
@@ -317,12 +317,8 @@ func (l PermalinkExpander) pageToPermalinkSection(p Page, _ string) (string, err
 
 // pageToPermalinkSectionSlug returns the URL-safe form of the first section's slug or title
 func (l PermalinkExpander) pageToPermalinkSectionSlug(p Page, attr string) (string, error) {
-	if p.Section() == "" {
-		return "", nil
-	}
-
-	sectionPage, err := p.GetPage("/" + p.Section())
-	if err != nil {
+	sectionPage := p.FirstSection()
+	if sectionPage == nil || sectionPage.IsHome() {
 		return "", nil
 	}
 	return l.pageToPermalinkSlugElseTitle(sectionPage, attr)
@@ -369,18 +365,16 @@ func (l PermalinkExpander) translationBaseName(p Page) string {
 func (l PermalinkExpander) withSectionPagesFunc(f func(Page, string) (string, error), join func(...string) string) func(p Page, s string) (string, error) {
 	return func(p Page, s string) (string, error) {
 		var entries []string
-		sectionEntries := p.CurrentSection().SectionsEntries()
+		currentSection := p.CurrentSection()
 
-		for i := range sectionEntries {
-			sectionPath := "/" + path.Join(sectionEntries[:i+1]...)
-			section, err := p.GetPage(sectionPath)
-			if err != nil {
-				return "", err
-			}
-			if section == nil {
-				return "", nil
-			}
+		// Build section hierarchy: ancestors (reversed to root-first) + current section
+		sections := currentSection.Ancestors().Reverse()
+		sections = append(sections, currentSection)
 
+		for _, section := range sections {
+			if section.IsHome() {
+				continue
+			}
 			entry, err := f(section, s)
 			if err != nil {
 				return "", err
--- a/resources/page/permalinks.go
+++ b/resources/page/permalinks.go
@@ -62,6 +62,14 @@ func (p PermalinkExpander) callback(attr string) (pageToPermaAttribute, bool) {
 		}, true
 	}
 
+	if strings.HasPrefix(attr, "sectionslugs[") {
+		fn := p.toSliceFunc(strings.TrimPrefix(attr, "sectionslugs"))
+		sectionSlugsFunc := p.withSectionPagesFunc(p.pageToPermalinkSlugElseTitle, func(s ...string) string {
+			return path.Join(fn(s)...)
+		})
+		return sectionSlugsFunc, true
+	}
+
 	// Make sure this comes after all the other checks.
 	if referenceTime.Format(attr) != attr {
 		return p.pageToPermalinkDate, true
@@ -87,7 +95,9 @@ func NewPermalinkExpander(urlize func(uri string) string, patterns map[string]ma
 		"weekdayname":           p.pageToPermalinkDate,
 		"yearday":               p.pageToPermalinkDate,
 		"section":               p.pageToPermalinkSection,
+		"sectionslug":           p.pageToPermalinkSectionSlug,
 		"sections":              p.pageToPermalinkSections,
+		"sectionslugs":          p.pageToPermalinkSectionSlugs,
 		"title":                 p.pageToPermalinkTitle,
 		"slug":                  p.pageToPermalinkSlugElseTitle,
 		"slugorfilename":        p.pageToPermalinkSlugElseFilename,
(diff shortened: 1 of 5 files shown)


resources/page/permalinks.go
```diff
@@ -305,10 +315,29 @@ func (l PermalinkExpander) pageToPermalinkSection(p Page, _ string) (string, err
 	return p.Section(), nil
 }
 
+// pageToPermalinkSectionSlug returns the URL-safe form of the first section's slug or title
+func (l PermalinkExpander) pageToPermalinkSectionSlug(p Page, attr string) (string, error) {
+	if p.Section() == "" {
+		return "", nil
+	}
+
+	sectionPage, err := p.GetPage("/" + p.Section())
```
bep: p.Section() returns the first level (that method was added before we had a concept of nested section), which is, I'm guessing, not what we want here. Also, I don't see why this is needed -- isn't `p` the section you want the title from?

resources/page/permalinks.go
```diff
@@ -333,6 +362,36 @@ func (l PermalinkExpander) translationBaseName(p Page) string {
...
+// withSectionPagesFunc returns a function that builds permalink attributes from section pages.
+// It applies the transformation function f to each ancestor section (Page), then joins the results with the join function.
+//
+// Current use is create section-based hierarchical paths using section slugs.
+func (l PermalinkExpander) withSectionPagesFunc(f func(Page, string) (string, error), join func(...string) string) func(p Page, s string) (string, error) {
+	return func(p Page, s string) (string, error) {
+		var entries []string
+		sectionEntries := p.CurrentSection().SectionsEntries()
+
+		for i := range sectionEntries {
```
bep: While this would work, it's much more effective (and simpler) to just do `for _, section := range p.Ancestors().Reverse() {}`

resources/page/permalinks.go
DustinFischer: The intention with this change was to keep the `:sectionslug` token symmetrical with the `:section` token which puts the first level section in the path (here's a related [discussion](https://discourse.gohugo.io/t/support-a-sectionslug-permalink-token/55005/2) from the original forum post). 

For example, if we had content structure:
* `content/a/_index.md` with `slug: section-a-slug` 
* `content/a/b/myarticle.md` with `my-article-slug`


If `p` is the Page for `a/b/myarticle.md` then the expected output for the following url formats would be :
* `/:section/:slug` -> `/a/my-article-slug`
* `/:sectionslug/:slug` -> `/section-a-slug/my-article-slug`

resources/page/permalinks.go
DustinFischer: I had considered and tried this approach initially. The issue was that `p.Ancestors()` includes the _home_ page if present, so the path element for the _home_ page would be included in the published site structure as well. 

I tried to keep the behaviour of `:sectionslugs`as similar as possible to `:sections` which never includes the home page in the path. I suppose I could add a check on `p.IsSection()` to filter out the home page as well? In that scenario would it be necessary to look at performance considerations? It seemed to me like the `GetPage()` approach would require fewer tree lookup operations compared to using `Ancestors()` combined with `IsSection()`.

resources/page/permalinks.go
bep: Yea, OK, i read this a little fast then. But then I suggest we do `sectionPage := p.FirstSection`.

resources/page/permalinks.go
bep: I'm pretty `p.Ancestors()` is at least as performant as using GetPage, and it has the upside of making the code much simpler, which is always a good thing. I suggest you check for `.IsHome` and break.

resources/page/permalinks.go
DustinFischer: Yes that's a good point. I'll make the change.

resources/page/permalinks.go
DustinFischer: Sure. I'm happy to make this change. I agree it will simplify code and improve readability.

DustinFischer:
@bep I've made the changes you suggested:
- Use `FirstSection()` instead of `GetPage` lookup for `:sectionslug`
- Use `Ancestors().Reverse()`  instead of `GetPage` lookup for `:sectionslugs`

resources/page/permalinks.go
```diff
@@ -317,12 +317,8 @@ func (l PermalinkExpander) pageToPermalinkSection(p Page, _ string) (string, err
 
 // pageToPermalinkSectionSlug returns the URL-safe form of the first section's slug or title
 func (l PermalinkExpander) pageToPermalinkSectionSlug(p Page, attr string) (string, error) {
-	if p.Section() == "" {
-		return "", nil
-	}
-
-	sectionPage, err := p.GetPage("/" + p.Section())
-	if err != nil {
+	sectionPage := p.FirstSection()
```
torabisu, line 320: Looking at the PR comments, agree with @bep, this is way cleaner

resources/page/permalinks.go
```diff
@@ -369,18 +365,16 @@ func (l PermalinkExpander) translationBaseName(p Page) string {
...
-			sectionPath := "/" + path.Join(sectionEntries[:i+1]...)
-			section, err := p.GetPage(sectionPath)
-			if err != nil {
-				return "", err
-			}
-			if section == nil {
-				return "", nil
-			}
+		// Build section hierarchy: ancestors (reversed to root-first) + current section
+		sections := currentSection.Ancestors().Reverse()
```
torabisu, line 371: @DustinFischer again more cleaner and readable

bep:
This looks great, thanks! I will merge once the test goes green. A side note: I plan on doing a Hugo release on Monday, mostly to get a Hugo release with Go 1.25 out there, but this new feature will also be very useful to many.

DustinFischer:
Thanks @bep, that's great news! And thanks for the considerate review and feedback. Hoping that it will be a useful enhancement to many users. I'll keep an eye out for the release today.

PR #13913, opened 2025-08-18
tpl: Fix incorrect template selection order

### LINKED ISSUE #13877: Template selection considers last media type suffix instead of first
This is not a recent regression; the behavior exists in v0.146.0 and later.

While updating the docs for the new template selection logic, I ran across this:

> Note that if the output format’s Media Type has more than one suffix defined, only the **first** is considered.

I wanted to verify that the behavior had not changed, and found that with v0.146.0 and later, only the **last** is considered.

site config

```toml
[mediaTypes.'text/html']
suffixes = ['b','a','c']

[outputFormats.html]
mediaType = 'text/html'
```

project structure

```text
./
├── content/
│   └── p1.md
├── layouts/
│   └── _default/
│       ├── page.html.a
│       ├── page.html.b
│       └── page.html.c
└── hugo.toml
```

With the above, `layouts/_default/page.html.c` is used instead of `layouts/_default/page.html.b`.

Test case to follow.

yuhui-liu:
closes  #13877

### DIFF
--- a/tpl/tplimpl/templatestore.go
+++ b/tpl/tplimpl/templatestore.go
@@ -1212,6 +1212,12 @@ func (s *TemplateStore) insertTemplate2(
 		replace = subCategory == SubCategoryInline && nkExisting.subCategory == SubCategoryInline
 	}
 
+	if !replace && existingFound && fi != nil && nkExisting.Fi != nil {
+		// Fix issue #13877.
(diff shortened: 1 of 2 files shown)


jmooring:
Addressed with https://github.com/gohugoio/hugo/pull/14602.

PR #13915, opened 2025-08-19
refactor to use reflect.TypeFor

cuiweixie:


bep:
This change is fine, but for the future I would appreciate if you could tone down the very lengthy (I assume) AI generated PR descriptions, especially considering the simple nature of this change.

cuiweixie:
> This change is fine, but for the future I would appreciate if you could tone down the very lengthy (I assume) AI generated PR descriptions, especially considering the simple nature of this change.

okay

PR #13969, opened 2025-09-05
resources/page: Fix truncated summary logic

### LINKED ISSUE #13967: .Page.Truncated sometimes returns an incorrect value with manual summaries
In the test below, the first and last assertion fail because the value of `.Truncated` is the opposite of what it should be.

```go
func TestIssue13967(t *testing.T) {
	t.Parallel()

	files := `
-- hugo.toml --
disableKinds = ['home','rss','section','sitemap','taxonomy','term']
-- layouts/all.html --
Title: {{ .Title }}|Summary: {{ .Summary }}|Truncated: {{ .Truncated }}|
-- content/p1.md --
---
title: p1
---
<!--more--> one two three
-- content/p2.md --
---
title: p2
---
one <!--more--> two three
-- content/p3.md --
---
title: p3
---
one two <!--more--> three
-- content/p4.md --
---
title: p4
---
one two three <!--more-->
`
	b := hugolib.Test(t, files)

	b.AssertFileContent("public/p1/index.html", `Title: p1|Summary: |Truncated: true|`) // fails
	b.AssertFileContent("public/p2/index.html", `Title: p2|Summary: <p>one</p>|Truncated: true|`)
	b.AssertFileContent("public/p3/index.html", `Title: p3|Summary: <p>one two</p>|Truncated: true|`)
	b.AssertFileContent("public/p4/index.html", `Title: p4|Summary: <p>one two three</p>|Truncated: false|`) // fails
}
```

Reference: <https://discourse.gohugo.io/t/problem-with-the-truncated-method-returning-true-when-it-shouldnt/55861>

### LINKED ISSUE #13968: .Page.Truncated sometimes returns an incorrect value with automatic summaries
In the test below, all but the last assertion fail because the value of .Truncated is the opposite of what it should be.

```go
func TestIssue13968(t *testing.T) {
	t.Parallel()

	files := `
-- hugo.toml --
summaryLength = SUMMARY_LENGTH
-- layouts/all.html --
Title: {{ .Title }}|Summary: {{ .Summary }}|Truncated: {{ .Truncated }}|
-- content/_index.md --
---
title: home
---
one two three
`

	tests := []struct {
		summaryLength int
		want          string
	}{
		{0, "Title: home|Summary: |Truncated: true|"},                      // fails
		{1, "Title: home|Summary: <p>one two three</p>|Truncated: false|"}, // fails
		{2, "Title: home|Summary: <p>one two three</p>|Truncated: false|"}, // fails
		{3, "Title: home|Summary: <p>one two three</p>|Truncated: false|"}, // fails
		{4, "Title: home|Summary: <p>one two three</p>|Truncated: false|"},
	}

	for _, tt := range tests {
		f := strings.ReplaceAll(files, "SUMMARY_LENGTH", strconv.Itoa(tt.summaryLength))
		b := hugolib.Test(t, f)
		b.AssertFileContent("public/index.html", tt.want)
	}
}
```

Reference: <https://discourse.gohugo.io/t/problem-with-the-truncated-method-returning-true-when-it-shouldnt/55861>

jmooring:
Fixes #13967
Fixes #13968

### DIFF
--- a/resources/page/page_markup.go
+++ b/resources/page/page_markup.go
@@ -104,7 +104,7 @@ func (s HtmlSummary) trimSpace(ss string) string {
 
 func (s HtmlSummary) Content() string {
 	if s.Divider.IsZero() {
-		return s.source
+		return s.trimSpace(s.source)
 	}
 	ss := s.source[:s.Divider.Low]
 	ss += s.source[s.Divider.High:]
@@ -139,7 +139,7 @@ func (s HtmlSummary) ContentWithoutSummary() string {
 }
 
 func (s HtmlSummary) Truncated() bool {
-	return s.SummaryLowHigh.High < len(s.source)
+	return s.Summary() != s.Content()
 }
 
 func (s *HtmlSummary) resolveParagraphTagAndSetWrapper(mt media.Type) tagReStartEnd {
--- a/hugolib/page__content.go
+++ b/hugolib/page__content.go
@@ -645,8 +645,9 @@ func (c *cachedContentScope) contentRendered(ctx context.Context) (contentSummar
 			}
 			html := cp.po.p.s.ContentSpec.TrimShortHTML(b.Bytes(), cp.po.p.m.pageConfig.Content.Markup)
 			rs.Value.summary = page.Summary{
-				Text: helpers.BytesToHTML(html),
-				Type: page.SummaryTypeFrontMatter,
+				Text:      helpers.BytesToHTML(html),
+				Type:      page.SummaryTypeFrontMatter,
+				Truncated: rs.Value.summary.Truncated,
 			}
 			rs.Value.contentWithoutSummary = rs.Value.content
 		}
--- a/resources/page/page_markup_integration_test.go
+++ b/resources/page/page_markup_integration_test.go
@@ -14,6 +14,8 @@
 package page_test
 
 import (
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/gohugoio/hugo/hugolib"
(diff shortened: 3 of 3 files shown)


resources/page/page_markup.go
```diff
@@ -139,7 +139,7 @@ func (s HtmlSummary) ContentWithoutSummary() string {
 }
 
 func (s HtmlSummary) Truncated() bool {
-	return s.SummaryLowHigh.High < len(s.source)
+	return strings.TrimSpace(s.Summary()) != strings.TrimSpace(s.Content())
```
bep: While this may be more correct, it's also very costly (mostly memory). I suggest you start out trying to make `s.SummaryLowHigh.High` move past any whitepsace (or something).

jmooring: `s.source` looks like this:

```text
<p>one two three</p>
<p>HUGOMORE42</p>
```

So, is this acceptable (less costly)?

```text
return s.SummaryLowHigh.High < len(strings.TrimSpace(s.Content()))
```

bep: OK, I read your code a little too fast. Looking at the `HtmlSummary` struct now, I see that we're already doing "trim spacing" in both Content and Summary, so 

1. This would not make it any "worse" -- and we could fix that (by storing the values) if that becomes a problem.
2. But since both Summary() and Content() are trimmed, you don't need to trim them again.

jmooring: `Content()`  wasn't trimming whitespace when a summary divider was missing. This was causing four assertions in TestIssue13968 to fail. I've added a trimming step inside that conditional to resolve the issue, and removed the trimming from the comparison.

PR #13982, opened 2025-09-13
markup/goldmark: Enhance footnote extension with auto-prefixing option

### LINKED ISSUE #8045: Support `footnoteAnchorPrefix` for `goldmark`
It looks like goldmark has implemented the `footnoteAnchorPrefix` feature https://github.com/yuin/goldmark/commit/9e0189df270cb443c204b51e89e2b0749b98370c. It would be great if hugo can also support this option.

Also see discussions in [Goldmark: How to configure the footnote anchor prefix?
](https://discourse.gohugo.io/t/goldmark-how-to-configure-the-footnote-anchor-prefix/29639)

jmooring:
This commit introduces a new option, `enableAutoIDPrefix`, to the Goldmark footnote extension. When enabled, it prepends a unique prefix to footnote IDs, preventing clashes when multiple documents are rendered together. This prefix is unique to each logical path, which means that the prefix is not unique across content dimensions such as language. This change also refactors the extension's configuration from a boolean to a struct.

Closes #8045

### DIFF
--- a/markup/goldmark/convert.go
+++ b/markup/goldmark/convert.go
@@ -176,8 +176,15 @@ func newMarkdown(pcfg converter.ProviderConfig) goldmark.Markdown {
 		extensions = append(extensions, extension.DefinitionList)
 	}
 
-	if cfg.Extensions.Footnote {
-		extensions = append(extensions, extension.Footnote)
+	if cfg.Extensions.Footnote.Enable {
+		if cfg.Extensions.Footnote.EnableAutoIDPrefix {
+			extensions = append(extensions, extension.NewFootnote(extension.WithFootnoteIDPrefixFunction(func(n ast.Node) []byte {
+				documentID := n.OwnerDocument().Meta()["documentID"].(string)
+				return []byte("h" + documentID)
+			})))
+		} else {
+			extensions = append(extensions, extension.Footnote)
+		}
 	}
 
 	if cfg.Extensions.CJK.Enable {
@@ -262,6 +269,7 @@ func (c *goldmarkConverter) Parse(ctx converter.RenderContext) (converter.Result
 		reader,
 		parser.WithContext(pctx),
 	)
+	doc.OwnerDocument().AddMeta("documentID", c.ctx.DocumentID)
 
 	return parserResult{
 		doc: doc,
--- a/docs/data/docs.yaml
+++ b/docs/data/docs.yaml
@@ -1238,7 +1238,9 @@ config:
             enable: false
           superscript:
             enable: false
-        footnote: true
+        footnote:
+          enable: true
+          enableAutoIDPrefix: false
         linkify: true
         linkifyProtocol: https
         passthrough:
--- a/hugolib/page__meta.go
+++ b/hugolib/page__meta.go
@@ -794,11 +794,9 @@ func (p *pageMeta) newContentConverter(ps *pageState, markup string) (converter.
 		return converter.NopConverter, fmt.Errorf("no content renderer found for markup %q, page: %s", markup, ps.getPageInfoForError())
 	}
 
-	var id string
 	var filename string
 	var path string
 	if p.f != nil {
-		id = p.f.UniqueID()
 		filename = p.f.Filename()
 		path = p.f.Path()
 	} else {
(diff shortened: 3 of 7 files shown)


jmooring:
I was initially concerned that the footnote links would change if the page path changed, as the document ID is derived from it. However, I realized this is already the case whenever you insert a new footnote before an existing one—the numbering shifts, and the links change. Given that content changes are much more frequent than path changes, I believe this is an acceptable trade-off.

And how often does one link to a footnote in another document? Never, or close to it.

bep:
>And how often does one link to a footnote in another document? Never, or close to it.

I have never seen it.

markup/goldmark/convert.go
```diff
@@ -262,6 +276,7 @@ func (c *goldmarkConverter) Parse(ctx converter.RenderContext) (converter.Result
 		reader,
 		parser.WithContext(pctx),
 	)
+	doc.OwnerDocument().AddMeta("documentID", c.ctx.DocumentID)
```
bep, line 272: I have scanned the code, and it looks like this is the first time we use the value of `DocumentID` (at least in a user facing API).

Currently, this value is derived from the content's backing file (`f.UniqueID()`), blank if not. For now, let's skip the discussion about that method, but this means that 

1. Pages without a content file will not have an ID.
2. Pages created via content adapters will have the same file ID (I assume)

I suggest that we where the `converter.DocumentContext` is created creates a new ID which is the XxHashed version of `page.Path`.

jmooring:
> Pages created via content adapters will have the same file ID (I assume)

I have verified that this is true and will add some test cases.

jmooring:
> creates a new ID

I have done this, and everything works fine, but having two IDs in the `converter.DocumentContext` struct isn't satisfying.  Can we leave the `f.UniqueID` method as is, and simply change how `DocumentID` is set? For example, change this:

```go
var id string
var filename string
var path string
if p.f != nil {
    id = p.f.UniqueID()
    filename = p.f.Filename()
    path = p.f.Path()
} else {
    path = p.Path()
}

cpp, err := cp.New(
    converter.DocumentContext{
	    Document:       doc,
	    DocumentLookup: documentLookup,
	    DocumentID:     id,
	    DocumentName:   path,
	    Filename:       filename,
    },
)
```

to this

```go
var filename string
var path string
if p.f != nil {
    filename = p.f.Filename()
    path = p.f.Path()
} else {
   path = p.Path()
}

cpp, err := cp.New(
    converter.DocumentContext{
	    Document:       doc,
	    DocumentLookup: documentLookup,
	    DocumentID:     hashing.XxHashFromStringHexEncoded(p.Path()),
	    DocumentName:   path,
	    Filename:       filename,
    },
)
```

As you noted, `DocumentID` doesn't appear to be used anywhere, so it seems like changing its derivation should not be a problem.

jmooring:
@bep Please see my previous comment when you have a moment.

bep:
>to this

Yes, that was how I imagined it in tthe first place, so please do.

markup/goldmark/convert.go
```diff
@@ -176,8 +176,19 @@ func newMarkdown(pcfg converter.ProviderConfig) goldmark.Markdown {
...
 	}
 
-	if cfg.Extensions.Footnote {
-		extensions = append(extensions, extension.Footnote)
+	if cfg.Extensions.Footnote.Enable {
+		if cfg.Extensions.Footnote.EnableAutoIDPrefix {
+			extensions = append(extensions, extension.NewFootnote(extension.WithFootnoteIDPrefixFunction(func(n ast.Node) []byte {
+				documentID, ok := n.OwnerDocument().Meta()["documentID"].(string)
+				if !ok {
+					panic("BUG: unable to cast documentID metadata to string")
```
bep: This is nit-picking, but  "should not happen" situations do not need friendly error messages, so I would prefer if you just dropped the `ok` and did a direct type case.

markup/goldmark/convert.go
```diff
@@ -176,8 +176,19 @@ func newMarkdown(pcfg converter.ProviderConfig) goldmark.Markdown {
...
+				}
+
+				return []byte("h-" + documentID + "-")
```
bep: Why do we need/want these prefix{/suffixes?

markup/goldmark/convert.go
jmooring: I will change.

markup/goldmark/convert.go
jmooring: The `h-` prefix is to make sure that element `id` attributes do not start with a digit. I know, it is unlikely that anyone would try to target these with CSS (where the selector may not begin with a digit), but it seems like a best practice to me. I can drop the hyphen if you'd like.

The `-` suffix was just symmetry with the other hyphen. 

I'm happy to do whatever you'd like here, but I think at a minimum we should do:

```go
return []byte("h" + documentID)
```

bep: Thanks for the explanation, I was just curious. This makes sense.

PR #13983, opened 2025-09-13
Add List function for struct field names and map keys

### LINKED ISSUE #9148: Add debug.List

UrielOfir:
Introduce a List function that retrieves field names and method names from structs or keys from maps using reflection. This addition enhances the debugging capabilities by providing a way to inspect data structures more effectively. Include tests to validate the functionality across various data types.

Resolve #9148

### DIFF
--- a/tpl/debug/debug.go
+++ b/tpl/debug/debug.go
@@ -16,6 +16,7 @@ package debug
 
 import (
 	"encoding/json"
+	"reflect"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


siddlal:
Hello!
Is this issues still relevant?

bep:
Thanks for this, and I'm sorry for the delayed response. There are now 4 PRs implementing this. I have looked at all of them, and they mostly look fine on the basic level, but there are some issues there I'm guessing it's hard to handle if you're not very familiar with Hugo's code base. Instead of me picking one of the PR and try to get that over the finish line, I'm assigning the relevant issue to myself.

PR #13984, opened 2025-09-13
skip livereload.js injection if no tags found

AndrewChubatiuk:
building hugo site with root index.html that serves react app with react-router-dom that contains menu, header, footer and breadcrump and other pages that are build inside`internal` folder that contain only parts of html bodies that are expected to be rendered inside root react app body and livereload.js script is not expected to be injected inside these parts. added condition that ignores injection of livereload.js is no relevant tags found

### DIFF
--- a/testscripts/commands/server.txt
+++ b/testscripts/commands/server.txt
@@ -24,6 +24,7 @@ myenv = "theproduction"
 -- config/development/params.toml --
 myenv = "thedevelopment"
 -- layouts/index.html --
+<!DOCTYPE html>
 <body>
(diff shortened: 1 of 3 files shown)


bep:
Sorry for the late feedback, but there are test failures.

AndrewChubatiuk:
Okay, will update tests soon, didn't expect this case to be covered

AndrewChubatiuk:
updated tests

bep:
@AndrewChubatiuk It would be good if you could also run the test locally; `go test ./...`  from the root.

AndrewChubatiuk:
fixed that test

PR #13985, opened 2025-09-14
fix(snap): Add `desktop` plug to `snapcraft.yaml`

maxkapur:
Currently, when Hugo is installed as a snap, `hugo serve --openBrowser` fails with the following permissions error:

```
WARN  Failed to open browser: fork/exec /usr/bin/xdg-open: permission denied
```

The solution is to add the `desktop` plug in `snap/snapcraft.yaml`, as done here.

I tested this fix by building/running the snap locally on an Ubuntu Desktop 25.04 virtual machine. I don't think there is any automated testing for the build/functionality of the snap but would be happy to add a test if you can point me in the right direction.

References:

- https://snapcraft.io/docs/desktop-interfaces
- https://askubuntu.com/a/1229462

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -89,6 +89,7 @@ apps:
     plugs:
       - home
       - network-bind
+      - desktop
       - removable-media
       - etc-gitconfig
       - gitconfig


maxkapur:
I think that CI failure is unrelated to this PR and has to do with GitHub changing the name of Windows CI runners from `windows-2025` to `windows-2025-x86` (specifying the platform), breaking the version of the `setup-ruby` action pinned in the GH actions workflow.

Compare the [Windows versions file](https://github.com/ruby/setup-ruby/blob/a6e6f86333f0a2523ece813039b8b4be04560854/windows-versions.json) from the pinned version to the [current version](https://github.com/ruby/setup-ruby/blob/master/windows-versions.json).

Upgrading the version of `setup-ruby` might fix that (I can do a separate PR if you want) but I will hold off in case you recognize the issue as something else.

jmooring:
Go ahead and submit a separate PR to use setup-ruby v1.257.0. 

There may be a good reason we are pinned to the older version, but testing the latest in a PR won't hurt.

PR #13989, opened 2025-09-17
commands: Map --minify CLI flag to the correct configuration key

### LINKED ISSUE #13988: Map --minify CLI flag to the correct (non-deprecated) config key
We're currently mapping `--minify` to the deprecated `minifyOutput` key in the config root. Should map to `minify.minifyOutput` instead.

Reference: <https://discourse.gohugo.io/t/deprecation-message-in-hugo-0-150-0/55929>

jmooring:
Closes #13988

### DIFF
--- a/testscripts/commands/hugo__minify_issue13988.txt
+++ b/testscripts/commands/hugo__minify_issue13988.txt
@@ -1,0 +1,13 @@
+hugo --minify --logLevel=info
+
+! stderr 'deprecated'
+grep '<p>one</p><p>two</p>'  public/index.html
+
+-- hugo.toml --
+disableKinds = ['page','rss','section','sitemap','taxonomy','term']
+-- content/_index.md --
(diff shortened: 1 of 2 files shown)


testscripts/commands/hugo__minify_issue13988.txt
```diff
@@ -0,0 +1,13 @@
+hugo --minify --logLevel=info
+
+! stderr 'deprecated: site config minifyOutput was deprecated in Hugo v0\.150\.0'
```
bep: This all looks good, but I have learned a lesson before about these "!" assertions that it is best to keep the text you're checking as short as possible, in this case `deprecated` would be enough , and that protect against any future spelling corrections.

jmooring: Good idea.

Issue #14002, opened 2025-09-25
plain text markup


davidejones:
The current markup / content formats are

Content format|Media type|Identifier|File extensions
:--|:--|:--|:--
Markdown|`text/markdown`|`markdown`|`markdown`,`md`, `mdown`
HTML|`text/html`|`html`|`htm`, `html`
Emacs Org Mode|`text/org`|`org`|`org`
AsciiDoc|`text/asciidoc`|`asciidoc`|`ad`, `adoc`, `asciidoc`
Pandoc|`text/pandoc`|`pandoc`|`pandoc`, `pdc`
reStructuredText|`text/rst`|`rst`|`rst`

I'm proposing an additional type of plaintext that is picked up by the extension txt or can be used with `markup: txt`

Content format|Media type|Identifier|File extensions
:--|:--|:--|:--
Plain Text |`text/plain`|`txt`|`txt`

My main use case is for outputting preprocessed html that was created by another tool that has no need for further parsing/understanding in hugo. 

For example it creates lots files like this

```
---
title: Foo Page
markup: txt
---

<div><strong>html generated</strong><div>more</div> by another tool here</div>
```

The intention is that this should be treated as static html that is just raw output and avoiding the build time impact of html parsing the files as part of hugo/goldmark etc.

jmooring:
If I understand this correctly, you have HTML files with front matter. What's the benefit of another content format, other than your existing files have a `txt` suffix?

davidejones:
> If I understand this correctly, you have HTML files with front matter. What's the benefit of another content format, other than your existing files have a `txt` suffix?

Its mostly for performance, if it ends up running through any of the existing markup parsing its taking time to process the body unnecessarily and at scale this can take quite some time. 

I'm thinking of this as a passthrough or plain text markup to avoid a lot of that overhead. So what you put in it is what you get out too but with the usual hugo benefits frontmatter etc.

bep:
>Its mostly for performance, if it ends up running through any of the existing markup parsing its taking time to process the body unnecessarily and at scale this can take quite some time.

There's no. markup processing with HTML files with front matter.

jmooring:
With the HTML content format, we look for a manual summary break (`<!--more-->`) and shortcode calls. That doesn't seem like a lot of overhead to me.

bep:
>With the HTML content format, we look for a manual summary break () and shortcode calls. That doesn't seem like a lot of overhead to me.

[...]

PR #14003, opened 2025-09-25
plain text markup

davidejones:
Proposing as a solution to https://github.com/gohugoio/hugo/issues/14002

### DIFF
--- a/markup/plaintext/convert_test.go
+++ b/markup/plaintext/convert_test.go
@@ -1,0 +1,37 @@
+// Copyright 2025 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


markup/plaintext/convert_test.go
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package plaintext
+
+import (
```
bep, line 16: I think it would be helpful if you could also create an integration test (there are many examples with the name integration in them) that demonstrates this in a real site.

davidejones:
I don't think i have a relevant use case for this anymore so i'm closing this out.

PR #14014, opened 2025-09-29
modules: Improve error checking

### LINKED ISSUE #14010: modules: Improve error checking and messaging

<pre>
$ hugo version
0.151.0
</pre>

Yes.

### TLDR description

When using a config like this:
```
$ cat hugo.yml 
baseURL: 'https://example.org/'
languageCode: 'en-us'
title: 'My New Hugo Site'

module:
  imports:
    - path: github.com/adityatelange/hugo-PaperMod
      version: v8.0
```

`hugo serve` will crash saying:
```
Error: command error: failed to load modules: failed to download module github.com/adityatelange/hugo-PaperMod@v8.0: failed to execute 'go [mod download -json github.com/adityatelange/hugo-PaperMod@v8.0]': failed to execute binary "go" with args [mod download -json github.com/adityatelange/hugo-PaperMod@v8.0]:  *errors.errorString
```

### Steps to reproduce

1. `mkdir example`
2. `docker run -it --rm -p 1313:1313 -v .:/project  ghcr.io/gohugoio/hugo:v0.150.1 new site example`
3. `cd example`
4. `mv hugo.toml hugo.yml`, then edit the file and change it into yaml, just because I prefer working with yaml.
5. `docker run -it --rm -p 1313:1313 -v .:/project  ghcr.io/gohugoio/hugo:v0.150.1 mod init example`
6. Add an import to the config with a `version` key, so that the whole file looks like this:
```
baseURL: 'https://example.org/'
languageCode: 'en-us'
title: 'My New Hugo Site'

module:
  imports:
    - path: github.com/adityatelange/hugo-PaperMod
      version: v8.0
```
7. `docker run -it --rm -p 1313:1313 -v .:/project  ghcr.io/gohugoio/hugo:v0.150.1 serve --bind 0.0.0.0`

I expect step 7 to download/install the module and serve the website, but instead I get this error:
```
Error: command error: failed to load modules: failed to download module github.com/adityatelange/hugo-PaperMod@v8.0: failed to execute 'go [mod download -json github.com/adityatelange/hugo-PaperMod@v8.0]': failed to execute binary "go" with args [mod download -json github.com/adityatelange/hugo-PaperMod@v8.0]:  *errors.errorString
```

8. Just to narrow down the scope of the problem, comment out the `version` key in the yaml config and run the `serve` command again, and it works as expected.

And just to confirm that the `v8.0` tag of this module does exist, it's here: https://github.com/adityatelange/hugo-PaperMod/releases/tag/v8.0

### A short note about the context of why I'm pinning this version

I build my site by way of GitHub Actions, so I can't rely on installing modules locally and expecting modules to remain at the same version on my local hard drive. If I build my site using v8.0 of a given module today in September 2025, I want to be able to walk away from the blog and come back to it 3 years later and push another commit and be sure that I'm still using the same dependencies/versions in 2028 as I used in 2025. I don't want a breaking change in my theme to break my site unless I have explicitly updated the version pin for the theme.

### Links

Related issue: https://github.com/gohugoio/hugo/issues/13964
Related commit: https://github.com/gohugoio/hugo/commit/747cf4ad657b6eaeffcd0c7b0a10a38980c7b54c (presumably)
Related author: @bep 

Thanks for all your efforts on this great project.

jmooring:
1. Validate combination of module path and version
2. Capture and display error messages from JSON sent to stdout
3. ~~Remove invalid semver from mod_vendor__versions.txt testscript~~ (Need to investigate: `module.Check` reports that `<=v6.7.42` is "not a semantic version". That is true, but it is a valid _semantic version comparison_ per the [Go docs](https://go.dev/ref/mod#version-queries).)

Closes #14010

### DIFF
--- a/common/collections/stack.go
+++ b/common/collections/stack.go
@@ -13,9 +13,13 @@
 
 package collections
 
-import "slices"
+import (
+	"iter"
+	"slices"
+	"sync"
 
-import "sync"
+	"github.com/gohugoio/hugo/common/hiter"
+)
 
 // Stack is a simple LIFO stack that is safe for concurrent use.
 type Stack[T any] struct {
@@ -58,6 +62,11 @@ func (s *Stack[T]) Len() int {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	return len(s.items)
+}
+
+// All returns all items in the stack, from bottom to top.
+func (s *Stack[T]) All() iter.Seq2[int, T] {
+	return hiter.Lock2(slices.All(s.items), s.mu.RLock, s.mu.RUnlock)
 }
 
 func (s *Stack[T]) Drain() []T {
--- a/common/hiter/iter.go
+++ b/common/hiter/iter.go
@@ -38,3 +38,29 @@ func Concat2[K, V any](seqs ...iter.Seq2[K, V]) iter.Seq2[K, V] {
 		}
 	}
 }
+
+// Lock returns an iterator that locks before iterating and unlocks after.
+func Lock[V any](seq iter.Seq[V], lock, unlock func()) iter.Seq[V] {
+	return func(yield func(V) bool) {
+		lock()
+		defer unlock()
+		for e := range seq {
+			if !yield(e) {
+				return
+			}
+		}
+	}
+}
+
+// Lock2 returns an iterator that locks before iterating and unlocks after.
+func Lock2[K, V any](seq iter.Seq2[K, V], lock, unlock func()) iter.Seq2[K, V] {
+	return func(yield func(K, V) bool) {
+		lock()
+		defer unlock()
+		for k, v := range seq {
+			if !yield(k, v) {
+				return
+			}
+		}
+	}
+}
--- a/hugolib/content_map_page_assembler.go
+++ b/hugolib/content_map_page_assembler.go
@@ -14,14 +14,18 @@
 package hugolib
 
 import (
+	"cmp"
 	"context"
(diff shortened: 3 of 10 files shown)


modules/collect.go
```diff
@@ -371,6 +371,14 @@ func (c *collector) addAndRecurse(owner *moduleAdapter) error {
 			continue
 		}
 
+		// Validate the combination of module path and version.
+		if moduleImport.Version != "" {
+			err := module.Check(moduleImport.Path, moduleImport.Version)
```
bep: I understand this i a draft, and I'm not totally sure what the above check does, but for `version` when `path` points to a non-Go-module, a path that does not end with /v2 may point at a v2.0.0 version. I tested this earlier today with:

https://github.com/bep/hugomodnogomod

jmooring: Yeah, still looking into this. The intent is to prevent invalid semvers and invalid path/semver pairs such as:

```text
path = "github.com/user/project/v3"
version = 'v42.6.7'
```

> path points to a non-Go-module

Thanks. I will probably use `https://github.com/bep/hugomodnogomod` in an integration test unless this is a temporary repo.

bep: > unless this is a temporary repo.

I will let it live, I will probably have a use for it myself some day.

jmooring: Note to self: there isn't a simple and reliable way to verify that the combination of a module path and  _version query_ is semantically correct before attempting to get, list, etc.

PR #14041, opened 2025-10-11
Reimplement schema partial in JSON-LD to pass spec validation

### LINKED ISSUE #7431: Update, document, or remove internal Schema.org template
<pre>
$ hugo version
Hugo Static Site Generator v0.73.0-428907CC/extended darwin/amd64 BuildDate: 2020-06-23T16:37:47Z
</pre>

The current schema template does not define itemscope and itemtype, which are required for it to be valid. If you submit a page with the current meta tags to [W3.org Validator](https://validator.w3.org/) you will see errors:

`The itemprop attribute was specified, but the element is not a property of any item.`

If you run the HTML through [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool) no data will be parsed because it requires the above mentioned properties.

The wordCount meta entry in the template is only for the Article schema type, but if you set the type to Article in Google's tool, it will tell you it also requires "author", "headline", "publisher" which are not included in the template. The additional required fields that are included in the template, "datePublished" and "image", aren't required in the template.

In its current form, the template does not provide valid Schema.org structures. For users that are unfamiliar with Schema data structures it is adding HTML that won't validate and additional data that will not actually be parsed.

This template is mentioned, but not documented so users are not aware of the above issues. I feel documenting the template does little because the user still has to add several pieces of data to Params or frontmatter for anything meaningful to be added. The type of page is not as generic as OpenGraph (4 choices vs dozens in Schema)(ie "Article" is not always what I would consider "IsPage==True" to represent.)

nisbet-hubbard:
The current microdata-style markup does not apply to any particular item and, as a result, fails to pass W3C’s Nu HTML Checker.

This PR reimplements the same semantics in JSON-LD, a self-contained format that is also recommended by Google Search Central.

Closes #7431

### DIFF
--- a/tpl/tplimpl/embedded/templates/_partials/schema.html
+++ b/tpl/tplimpl/embedded/templates/_partials/schema.html
@@ -1,56 +1,69 @@
-{{- with or .Title site.Title | plainify }}
-  <meta itemprop="name" content="{{ . }}">
-{{- end }}
+<script type="application/ld+json">
+{
+  "@context": "https://schema.org",
+  "@type": "{{ with .Params.schemaType }}{{ . }}{{ else }}WebPage{{ end }}",
 
-{{- with or .Description .Summary site.Params.description | plainify | htmlUnescape }}
-  <meta itemprop="description" content="{{ trim . "\n\r\t " }}">
-{{- end }}
+  {{- $title := or .Title site.Title | plainify }}
+  "name": "{{ $title }}",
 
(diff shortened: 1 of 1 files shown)


jmooring:
https://github.com/gohugoio/hugo/issues/7431 is currently an unaccepted proposal. Although we obviously need to do something (what we have is broken), we haven't decided  _what_ to do or _how_ to do it. Closing this for now. Please continue discussion at https://github.com/gohugoio/hugo/issues/7431.

Issue #14044, opened 2025-10-12
Automatic summary may generate invalid HTML with blockquotes and lists


imomaliev:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.151.0+extended+withdeploy darwin/arm64 BuildDate=2025-10-02T13:30:36Z VendorInfo=brew
</pre>

### Does this issue reproduce with the latest release?

Yes

## Issue

Using a blockquote that extends past the `.Summary` cutoff causes the content below `{{ .Summary }}` to be treated as part of the blockquote until the end of the parent tag. [Reproduction repo](https://github.com/imomaliev/hugo-blockquote-summary-leak)

<img width="848" height="652" alt="Image" src="https://github.com/user-attachments/assets/212f747c-af3c-4af5-89fe-78fe9dc04ebd" />

Hugo's `.Summary` may lead to `blockquote` end tag be below cutoff point

## Explanation

In most modern browsers this html code will be parsed and rendered without errors

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<meta name="viewport" content="width=device-width">
</head>
<body>
  <main>
    <blockquote>
      <p>1</p>
  </main>
</body>
</html>
```

Leading to shown in browser as a source

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<meta name="viewport" content="width=device-width">
</head>
<body>
  <main>
    <blockquote>
      <p>1</p>
      <!-- This blockquote end tag was automatically inserted -->
    </blockquote>
  </main>
</body>
</html>
```

This is due to [these parsing rules](html.spec.whatwg.org/multipage/parsing.html#parsing-main-inbody#:~:text=An%20end%20tag%20whose%20tag%20name%20is%20one%20of%3A%20%22address%22%2C%20%22article%22%2C%20%22aside%22%2C%20%22blockquote%22)

> An end tag whose tag name is one of: "address", "article", "aside", "blockquote", "button", "center", "details", "dialog", "dir", "div", "dl", "fieldset", "figcaption", "figure", "footer", "header", "hgroup", "listing", "main", "menu", "nav", "ol", "pre", "search", "section", "select", "summary", "ul"
> 
>     If the stack of open elements does not have an element in scope that is an HTML element with the same tag name as that of the token, then this is a parse error; ignore the token.
> 
>     Otherwise, run these steps:
> 
>         Generate implied end tags.
> 
[...]

PR #14045, opened 2025-10-12
create/skeletons: Wrap section and home lists with section tags

imomaliev:
Read more https://github.com/gohugoio/hugo/issues/14044

### DIFF
--- a/create/skeletons/theme/layouts/home.html
+++ b/create/skeletons/theme/layouts/home.html
@@ -1,7 +1,9 @@
 {{ define "main" }}
   {{ .Content }}
   {{ range site.RegularPages }}
-    <h2><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></h2>
-    {{ .Summary }}
+    <section>
+      <h2><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></h2>
+      {{ .Summary }}
+    </section>
   {{ end }}
 {{ end }}
--- a/create/skeletons/theme/layouts/section.html
+++ b/create/skeletons/theme/layouts/section.html
@@ -2,7 +2,9 @@
   <h1>{{ .Title }}</h1>
   {{ .Content }}
   {{ range .Pages }}
-    <h2><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></h2>
-    {{ .Summary }}
+    <section>
+      <h2><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></h2>
+      {{ .Summary }}
+    </section>
   {{ end }}
 {{ end }}


jmooring:
Neither `article` nor `section` are semantically correct in this context. See:

- <https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/article>
- <https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section>

Use a `div` element instead.

imomaliev:
@jmooring Hi, thanks for the reply!

> Neither article nor section are semantically correct in this context.
> Use a div element instead.

I might be misunderstanding something, so can you please help me understand this better? Why are `article` and `section` not correct in this context?

[MDN docs for `article` state](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/article#:~:text=A%20given%20document)

> A given document can have multiple articles in it; for example, on a blog that shows the text of each article one after another as the reader scrolls, each post would be contained in an `<article>` element, possibly with one or more `<section>`s within.

and [for `section`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section#:~:text=If%20the%20contents%20of%20the%20element)

> If the contents of the element represent a standalone, atomic unit of content that makes sense syndicated as a standalone piece (e.g., a blog post or blog comment, or a newspaper article), the [`<article>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/article) element would be a better choice.

The `home.html` and `section.html` are clearly used for listing articles with "summary" of their text. In my understanding this exactly the case outlined in docs and should be a better fit than `div`.

jmooring:
Using an `article` element is incorrect because a summary is not a _self-contained_ composition.

Go ahead and use `section` as long as you wrap both the heading and the summary.

imomaliev:
> Using an article element is incorrect because a summary is not a self-contained composition.
> Go ahead and use section as long as you wrap both the heading and the summary.

But I am not wrapping `.Summary` by itself, I am wrapping it together with `h2` for `.Page.Title`. I think the title was incorrect and this is why there is some missunderstaning.

I've updated title to `Use article tag to wrap pages in collection rendering when using .Summary to avoid issues with tag leakage`. Can you please review the code and tell me if I should it change it to `section` or `div`?

jmooring:
> Go ahead and use `section` as long as you wrap both the heading and the summary.

Or to put it another way, don't use an `article` element.

imomaliev:
Replaced `article` with `section` in code and updated the title

jmooring:
Final step... please review the [commit message guidelines](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines). It should be something like:

```text
create/skeletons: Wrap section and home lists with section tags
````

imomaliev:
@jmooring I'm so sorry, missed that part when reading `CONTRIBUTING.md` yesterday. Thanks for your patience)

jmooring:
No problem, looks good, thank you for your contribution.

imomaliev:
Wanted to leave this research for the context, not for continuing the discussion

---

After re-reading [the spec](https://html.spec.whatwg.org/#article-or-section) and multiple articles [1](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Structuring_documents#Enter_HTML5_structural_elements), [2](https://www.smashingmagazine.com/2020/01/html5-article-section/),[3](https://html5doctor.com/the-article-element/)

> The [article](https://html.spec.whatwg.org/#the-article-element) element [represents](https://html.spec.whatwg.org/#represents) a complete, or self-contained, composition in a document, page, application, or site and that is, in principle, independently distributable or reusable, e.g. in syndication. This could be a forum post, a magazine or newspaper article, a blog entry, a user-submitted comment, an interactive widget or gadget, or any other independent item of content.

> When used specifically with content to be redistributed in syndication, the [article](https://html.spec.whatwg.org/#the-article-element) element is similar in purpose to the entry element in Atom. [[ATOM]](https://html.spec.whatwg.org/#refsATOM)

> A [section](https://html.spec.whatwg.org/#the-section-element) forms part of something else. An [article](https://html.spec.whatwg.org/#the-article-element) is its own thing. But how does one know which is which? Mostly the real answer is "it depends on author intent".

I still think `article` tag usage here could be considered semantically correct (when wrapping together with heading). If we are talking in the context of syndication, `title + summary` in collections pages like `home.html` and `section.html` may be considered _self-contained composition_ or "independent item of content." Because one could argue that it is possible to distribute only article's intro (title + summary) as separate content. As well as in the context of using [screen readers](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/HTML#structure_page_sections_logically), this may be preferable to `section` because we are listing "repeating content units".

jmooring:
I have approved this PR as-is. I think we're done here.

imomaliev:
> I have approved this PR as-is. I think we're done here

Yes, of course. As I've stated in the previous comment, I only intended to add some information after additional research, not to continue with the discussion. You are the "author" in the context of my previous comment, so your intent is what matters))

PR #14063, opened 2025-10-18
tpl/lang: Warn when using reserved translation keys

### LINKED ISSUE #14061: Warning messages when using a reserved i18n key
When I was working on a template, I noticed that my new translations weren't coming through. After some searching around for what I did wrong, I went to the docs and noticed that there are several reserved keys, (`description` in my case).

To be fair, it's on me that I didn't consult the documentation sooner. And I really appreciate the attention to detail that it _can_ be found in the documentation. But it might be nice if there was a warning about this in the CLI output. Perhaps only when `printI18nWarnings` is set.

Would there be interest for this? If so, I can try to see if I can create a PR for this. I imagine it would save people a bit of time, and create awareness that there are reserved keys.

Dobefu:
When using a translation key that is reserved in the `go-i18n` package, Hugo will now log an error instead of silently failing.

Fixes #14061

**Please note:** 
- I've looked into adding a test case, but there does not appear to be any tests for the `Translate` method. If wanted, I can add some. But perhaps it won't be needed for a minor change like this. Up to the reviewer(s) 🙂.
- Since translation keys may be multiple levels deep, I did not use `SplitN` with a count of 1.
- I'm intentionally not bounds checking, since an empty string will create an empty slice and won't trigger a runtime exception.

### DIFF
--- a/tpl/lang/lang.go
+++ b/tpl/lang/lang.go
@@ -30,6 +30,20 @@ import (
 	"github.com/spf13/cast"
 )
 
+var reservedTranslationIDs = map[string]string{
+	"id":          "",
+	"description": "",
+	"hash":        "",
+	"leftdelim":   "",
+	"rightdelim":  "",
+	"zero":        "",
+	"one":         "",
+	"two":         "",
+	"few":         "",
+	"many":        "",
+	"other":       "",
+}
+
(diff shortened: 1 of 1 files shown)


bep:
Thanks, but this should now be covered in 559a029de7e856dd919e659e1a1386c2c3a58696

PR #14094, opened 2025-10-24
markup/asciidocext: Improve Asciidoctor integration

### LINKED ISSUE #9202: asciidocext: The output destination directory is url escaped and the image is not displayed.

<pre>
$ hugo version 
Hugo Static Site Generator v0.80.0-792EF0F4/extended linux/amd64 BuildDate: 2020-12-31T13:46:18Z
</pre>

Yes

* PlantumlをHugoで表示する.adoc

```adoc
= PlantumlをHugoで表示する

[source,plantuml]
----
@startuml
A1 - B1
A1 -- C2
A1 -- D2
C2 - E2
@enduml
----

[plantuml]
----
@startuml
A1 - B1
A1 -- C2
A1 -- D2
C2 - E2
@enduml
----

```

```config.yml
markup:
  asciidocext:
    extensions:
      - asciidoctor-html5s
      - asciidoctor-diagram
    workingFolderCurrent: true
    trace: true
    attributes:
      my-base-url: 'https://localhost/'
      my-attribute-name: my value
```

* `hugo --destination output`

```tree output
~
├── posts
│   ├── Plantuml%E3%82%92Hugo%E3%81%A7%E8%A1%A8%E7%A4%BA%E3%81%99%E3%82%8B
│   │   └── diag-44963d5c8fb0ea5d64d7c880e7bfa7fe.png
│   ├── PlantumlをHugoで表示する
│   │   └── index.html
│   ├── index.html
│   ├── index.xml
│   └── page
│       └── 1
│           └── index.html
~
```

### LINKED ISSUE #10183: markup/asciidocext: Incorrect diagram path due to spaces and casing
I have a small Hugo site using AsciiDoc ([sources](https://github.com/madhead/wiki)). I want to add [diagrams](https://docs.asciidoctor.org/diagram-extension/latest) to it.

This page helps a lot to set them up: https://gohugo.io/content-management/formats. What I have to do to get the diagrams working, is just adding a few lines to the `config.yml`:

```diff
 markup:
   tableOfContents:
     startLevel: 2
+  asciidocExt:
+    extensions:
+      - asciidoctor-diagram
+    workingFolderCurrent: true
```

After that the diagrams mostly work, but I've noticed an issue.

Take a look at [this page](https://github.com/madhead/wiki/blob/master/content/wiki/3D%20printing/_index.adoc): `content/wiki/3D printing/_index.adoc`. It's path on the filesystem is not slug-friendly: it contains spaces and upper case characters.

Hugo renders this page into a file named `public/wiki/3d-printing/index.html`. Basically, the path is slugified.

However, when it calls `asciidoctor-diagram`, the `outdir` is different and corresponds to the original `*.adoc` file path:

```
DEBUG 2022/08/19 02:26:46 Render page 🗜️ 3D printing to "/wiki/3d-printing/index.html"
INFO 2022/08/19 02:26:46 Rendering wiki/3D printing/_index.adoc  using asciidoctor args [-r asciidoctor-diagram --base-dir /home/madhead/Projects/wiki/content/wiki/3D printing -a outdir=/home/madhead/Projects/wiki/public/wiki/3D printing --no-header-footer --verbose -] ...
```

Note this:

- `base-dir`: `/home/madhead/Projects/wiki/content/wiki/3D printing`
- `outdir`: `/home/madhead/Projects/wiki/public/wiki/3D printing`

As a result, I have two directories now:

image

What's even worse, is that the resulting HTML page is incorrect: it refers the image as if it were in the same directory:

```html
<img src="diag-8ac9b322bf3e2e1e7b0df17f6f64d0a6.png" alt="Diagram" width="540" height="280"/>
```

As you might guess, the diagram is not shown on the page.

This issue exists only when the path of page source (the `*.adoc` file) is different from the resulting slug. If the page path is the same as it's own slugified path, then everything works correctly.

Is it possible to tune this behavior? I'd prefer the diagrams be rendered into slugified directories.

Reading the code didn't help me a lot, looks like here https://github.com/gohugoio/hugo/blob/f7e00c039ff3cea5f991b05c1e325666004cf129/markup/asciidocext/convert.go#L161 the path is pretty rigidly tied to the `filepath.Dir(ctx.DocumentName)` or `filepath.Dir(ctx.Filename)`.

### LINKED ISSUE #10473: markup/asciidocext: Incorrect diagram path in multilingual project

<pre>
$ hugo version
hugo v0.107.0-2221b5b30a285d01220a26a82305906ad3291880+extended linux/amd64 BuildDate=2022-11-24T13:59:45Z VendorInfo=gohugoio
</pre>

Yes.

### Details

When using Hugo to render asciidoc articles containing diagrams by default it puts diagram images to the folder without language prefix even for multilingual projects.

#### Scenario 1 (single language) - OK

E.g. I have a project with a single language and an article `content/post/post-3.ru.adoc` with a diagram which will be rendered to:
```
$ tree public/post/post-3/
public/post/post-3/
├── index.html
└── sequence-diagram-example.svg
```

It works just fine, `post-3` will reference its diagram from `public/post/post-3` directory.

#### Scenario 2 (project with language directories) - FAIL

Now I have a project with several languages - each of them in its own directory, `config.toml`:
```
defaultContentLanguage = 'ru'
defaultContentLanguageInSubdir = true
```
`post-3` will be put to (and tries to reference diagram from here):
```
$ tree public/ru/post/post-3/
public/ru/post/post-3/
└── index.html
```
But diagram still located in non-language directory:
```
$ tree public/post/post-3/
public/post/post-3/
└── sequence-diagram-example.svg
```
#### Workaround

For now I'm able to bypass that issue by manually setting image location in each article. From above example:
```
:imagesoutdir: public/ru/post/post-3
```

### LINKED ISSUE #14160: markup/asciidocext: Implement diagram caching
The  [v3.1.0](https://github.com/asciidoctor/asciidoctor-diagram/releases/tag/v3.1.0) release of the asciidoctor-diagram extension introduced the ability to cache generated image files in addition to the existing metadata files.

To make this work in a Hugo context, we need to:

1. Set the  `diagram-cache-images-option` document attribute to `true`
2. Set the `diagram-cachedir` document attribute to a path outside of the `publishDir`

Implementation details:

1. The `diagram-cache-images-option` attribute will be hardcoded to `true`. Caching metadata files without caching image files defeats the purpose of caching when building a site. To disable caching entirely (both image and metadata files), users should set the `diagram-nocache-option` document attribute to `true`.
1. The the `diagram-cachedir` attribute will be set to the compiled value of the `caches.misc.dir` setting in the site config, normalized for the current OS.
1. Users will be warned if they attempt to override either  `diagram-cache-images-option` or `diagram-cachedir` in the site config.

jmooring:
Fixes an issue where improper attribute derivation from the page's relative permalink caused failures with `outdir`, `imagesoutdir`, and `imagesdir` when `markup.asciidocext.workingFolderCurrent` is enabled. The updated logic now correctly handles:

- Multi-byte characters
- Multilingual multi-host sites
- Site builds from a subdirectory
- Pages using ugly URLs

Supports diagram caching as implemented in v3.1.0 of the asciidoctor-diagram extension:

- Enables caching by default
- Sets default cache location to the compiled value of `caches.misc.dir`

Reduces duration of integration tests by:

- Generating GoAT diagrams instead of Ditaa diagrams
- Taking advantage of asciidoctor-diagram caching

Closes #9202
Closes #10183
Closes #10473
Closes #14160

### DIFF
--- a/markup/asciidocext/asciidoc_integration_test.go
+++ b/markup/asciidocext/asciidoc_integration_test.go
@@ -1,0 +1,1155 @@
+// Copyright 2025 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package asciidocext_test
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/gohugoio/hugo/htesting"
+	"github.com/gohugoio/hugo/hugolib"
+	"github.com/gohugoio/hugo/markup/asciidocext"
+)
+
+type contentFile struct {
+	// The path to the content file, relative to the project directory.
+	path string
+	// The title of the content file.
+	title string
+}
+
+type publishedContentFile struct {
+	// The path to the published content file, relative to the project directory.
+	path string
+	// The expected string in the content file used for content test assertions.
+	match string
+}
+
+type testCase struct {
+	name                           string
+	baseURL                        string
+	defaultContentLanguageInSubdir bool
+	langEnBaseURL                  string
+	langDeBaseURL                  string
+	langDeDisabled                 bool
+	uglyURLsS1                     bool
+	validate                       func(b *hugolib.IntegrationTestBuilder)
+}
+
+// Issue 9202, Issue 10183, Issue 10473
+func TestAsciiDocDiagrams(t *testing.T) {
+	if !htesting.IsRealCI() {
+		t.Skip()
+	}
+	if ok, err := asciidocext.Supports(); !ok {
+		t.Skip(err)
+	}
+	if ok, err := asciidocext.SupportsGoATDiagrams(); !ok {
+		t.Skip(err)
+	}
+
+	t.Cleanup(func() {
+		resetDefaultAsciiDocExtConfig()
+	})
+
+	diagramCacheDir := t.TempDir() // we want this to persist for all tests
+
+	files := createContentFiles() + contentAdapter + layouts
+
+	testCases := []testCase{
+		// Test 1 - Monolingual, uglyURLs = false for s1, without subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 1", "https://example.org/", false, "", "", true, false, validatePublishedSite_1},
+		// Test 2 - Monolingual, uglyURLs = false for s1, with subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 2", "https://example.org/subdir/", false, "", "", true, false, validatePublishedSite_2},
+		// // Test 3 - Monolingual, uglyURLs = true for s1, without subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 3", "https://example.org/", false, "", "", true, true, validatePublishedSite_3},
+		// // Test 4 - Monolingual, uglyURLs = true for s1, with subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 4", "https://example.org/subdir/", false, "", "", true, true, validatePublishedSite_4},
+		// // Test 5 - Multilingual, single-host, uglyURLs = false for s1, without subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 5", "https://example.org/", false, "", "", false, false, validatePublishedSite_5},
+		// Test 6 - Multilingual, single-host, uglyURLs = false for s1, with subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 6", "https://example.org/subdir/", false, "", "", false, false, validatePublishedSite_6},
+		// Test 7 - Multilingual, single-host, uglyURLs = true for s1, without subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 7", "https://example.org/", false, "", "", false, true, validatePublishedSite_7},
+		// Test 8 - Multilingual, single-host, uglyURLs = true for s1, with subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 8", "https://example.org/subdir/", false, "", "", false, true, validatePublishedSite_8},
+		// Test 9 - Multilingual, single-host, uglyURLs = false for s1, without subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 9", "https://example.org/", true, "", "", false, false, validatePublishedSite_9},
+		// Test 10 - Multilingual, single-host, uglyURLs = false for s1, with subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 10", "https://example.org/subdir/", true, "", "", false, false, validatePublishedSite_10},
+		// Test 11 - Multilingual, single-host, uglyURLs = true for s1, without subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 11", "https://example.org/", true, "", "", false, true, validatePublishedSite_11},
+		// Test 12 - Multilingual, single-host, uglyURLs = true for s1, with subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 12", "https://example.org/subdir/", true, "", "", false, true, validatePublishedSite_12},
+		// Test 13 - Multilingual, multi-host, uglyURLs = false, without subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 13", "", false, `baseURL = "https://.en.example.org/"`, `baseURL = "https://.de.example.org/"`, false, false, validatePublishedSite_13},
+		// Test 14 - Multilingual, multi-host, uglyURLs = false for s1, with subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 14", "", false, `baseURL = "https://.en.example.org/subdir/"`, `baseURL = "https://.de.example.org/subdir/"`, false, false, validatePublishedSite_14},
+		// Test 15 - Multilingual, multi-host, uglyURLs = true for s1, without subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 15", "", false, `baseURL = "https://.en.example.org/"`, `baseURL = "https://.de.example.org/"`, false, true, validatePublishedSite_15},
+		// Test 16 - Multilingual, multi-host, uglyURLs = true for s1, with subdir in base URL, defaultContentLanguageInSubdir = false
+		{"Test 16", "", false, `baseURL = "https://.en.example.org/subdir/"`, `baseURL = "https://.de.example.org/subdir/"`, false, true, validatePublishedSite_16},
+		// Test 17 - Multilingual, multi-host, uglyURLs = false for s1, without subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 17", "", true, `baseURL = "https://.en.example.org/"`, `baseURL = "https://.de.example.org/"`, false, false, validatePublishedSite_17},
+		// Test 18 - Multilingual, multi-host, uglyURLs = false for s1, with subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 18", "", true, `baseURL = "https://.en.example.org/subdir/"`, `baseURL = "https://.de.example.org/subdir/"`, false, false, validatePublishedSite_18},
+		// Test 19 - Multilingual, multi-host, uglyURLs = true for s1, without subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 19", "", true, `baseURL = "https://.en.example.org/"`, `baseURL = "https://.de.example.org/"`, false, true, validatePublishedSite_19},
+		// Test 20 - Multilingual, multi-host, uglyURLs = true for s1, with subdir in base URL, defaultContentLanguageInSubdir = true
+		{"Test 20", "", true, `baseURL = "https://.en.example.org/subdir/"`, `baseURL = "https://.de.example.org/subdir/"`, false, true, validatePublishedSite_20},
+	}
+
+	for _, tc := range testCases {
+		// Asciidoctor is really, really slow on Windows. Only run a few tests.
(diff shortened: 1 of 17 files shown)


jmooring:
I'm not thrilled about the duration of the integration test (it's the second largest contributor to overall test time), but I don't want to pare it down either.

markup/asciidocext/asciidoc_integration_test.go
```diff
@@ -0,0 +1,1326 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package asciidocext converts AsciiDoc to HTML using Asciidoctor
+// external binary. The `asciidoc` module is reserved for a future golang
+// implementation.
```
bep: Nitpick, but I suspect the above should be removed.

markup/asciidocext/asciidoc_integration_test.go
```diff
@@ -0,0 +1,1326 @@
...
+	f = files + replacer.Replace(configFileWithPlaceholders)
+	b = hugolib.Test(t, f)
+	validatePublishedSite_20(b, c)
+}
+
+// osFileExists reports whether a file with the given path exists on the OS FS.
+// We need to use os.Stat for existence checks on Asciidoctor-generated
+// diagrams. They are written directly to the OS file system, making them
+// invisible to the afero FS used by b.AssertFileExists.
+func osFileExists(path string) bool {
```
bep: Looking at this re. the tests above, I think the tests could be more real if you used OS fs in the tests.

```
b := hugolib.Test(t, files, hugolib.TestOptWithOSFs)
```

With that, this helper method could also be removed.

Edit in: As you do the Chdir above with a tempdir, you need to use `hugolib.TestOptWithConfig` and set both NeedsOsFS=true and WorkingDir.

And with that I also suggest to split the test run so you get 1 temp dir for each hugolib.Test -- so you don't get test results intermingled.

jmooring: > don't get test results intermingled

Great catch. I'm embarrassed that I didn't see that, particularly since I clear the public directory every time to avoid this when testing locally.

Incidentally, rebasing this against master only required one minor change, which was (at least to me) a pleasant surprise.

Before:

```go
relPath = filepath.Join(a.Cfg.Conf.Language().Lang, relPath)
```

After:

```go
language, ok := a.Cfg.Conf.Language().(*langs.Language)
if !ok {
	return []string{}, fmt.Errorf("expected *langs.Language, got %T", a.Cfg.Conf.Language())
}
relPath = filepath.Join(language.Lang, relPath)
```

bep: Yea, I introduced a package cycle so the `allconfig` package couldn't import the `langs` package anymore ...

As to this PR, I'm a little confused about it's state re. my comment above. To me it looks like it's still using a hybrid of in memory and OS fs.

jmooring: > I'm a little confused about it's state re. my comment above

I've refactored to split the test into separate runs, but I still need to figure out how/where to place the Asciidoctor diagram cache. Without access to the cache this integration test is stupid slow.

EDIT: The Asciidoctor diagram cache is not a cache for Asciidoctor diagrams. Instead it contains a checksum file for each image. In a Hugo context that means that the diagrams are regenerated if you clear the public directory. Yuck. <https://github.com/asciidoctor/asciidoctor-diagram/issues/496>

EDIT2: Changed content to use Ditaa diagrams instead of PlantUML. These diagrams take less time to generate. Testing locally takes about 60 seconds, but this test will still be slow on GitHub. There's movement on the asciidoctor-diagram issue above, which should have a pretty big impact on test time.

markup/asciidocext/convert.go
```diff
@@ -29,21 +29,38 @@ type provider struct{}
...
-		return true
+		return true, nil
+
+	}
+
+	return hasAsciiDoc, err
+}
+
+// SupportsDitaaDiagrams reports whether the AsciiDoc converter can render
+// Dita diagrams. Only used in tests.
```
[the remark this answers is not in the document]
jmooring: Fixed

jmooring:
This is ready for review.

Please note that the integration test is currently slow when executed in a GitHub Action:

- Linux: 60 s
- Windows: 120 s

This performance issue is expected to be ~resolved~ improved once [asciidoctor/asciidoctor-diagram#497](https://github.com/asciidoctor/asciidoctor-diagram/pull/497) is merged and a new release is rolled out by the maintainers. Local testing shows the duration decreases by approximately 50% with this fix.

There may be other ways to address the performance, such as using a diagram type that doesn't require JRE[^1] on the backend, provided there's a diagram generation exec that's available for all platforms. 

[^1]: Both PlantUML and Ditaa diagrams require JRE.

bep:
It times in at 64.572s on my MacBook M1 with 32 GB of ram ...

So, running all tests with `CI_LOCAL=true` set (we skip some other slow tests when not on CI):

This branch:

```
ok      github.com/gohugoio/hugo/markup/asciidocext     117.271s
go test -count 1 ./...  864.43s user 179.40s system 697% cpu 2:29.65 total
```

On master:

```
go test -count 1 ./...  881.50s user 155.33s system 656% cpu 2:37.89 total
```

Which, in relative terms, isn't disastrous (but not that great, either).

Go's test cache helps, of course, but I run all the tests pretty regulary, and I would hate for that feedback loop to get any slower because of a feature I ... don't use.

Could we, at least until the cache feature above is released, add a `if !htesting.IsRealCI() { t.Skip() }`?

jmooring:
> Could we, at least until the cache feature above is released, add a if !htesting.IsRealCI() { t.Skip() }?

Yes, I was thinking the same thing. I'll do that later today.

I evaluated other supported diagram types with backends that are cross-platform, easy to install, and well-maintained.

While our current method (Ditaa via JRE) is faster than those alternatives, I discovered that using the GoAT CLI can reduce the test duration by approximately 50%. Currently, this requires a minor workaround (renaming the executable).

To formally implement this performance improvement without the trickery, I have [created a PR](https://github.com/asciidoctor/asciidoctor-diagram/pull/500) in the asciidoctor-diagram repository to add formal support for the GoAT tool. I hope this will be merged and included in the next release, which is expected to feature caching.

jmooring:
The CI test has errored out 3 times. The last run threw `fatal error: concurrent map writes`.

bep:
@jmooring the "concurrent map writes" is my mistake. I will fix that tomorrow, see #14140.

bep:
I think the practical way forward to get this merged (as it solves a real issue) is to pull up "some" of the most important test cases to the top and do a 

```go
const slowThreshold = 5
if i > slowThreshold && os.GetEnv("SOME_VARIABLE" )) == "" {
   break
}
```

... or something.

And then revisit this later when the cache issue is resolved.

jmooring:
> most important test cases

Rebased on master. Integration test will only run when `IsRealCI` is true, and then only for three of the sub-tests.

```go
if !slices.Contains([]string{"Test 4", "Test 8", "Test 16"}, tc.name) {
    continue
}
```

Our CI testing... 15s on Linux and 30s on Windows. Caching, plus GoAT instead of Ditaa diagrams, should make this much faster. Spinning up JRE for each diagram isn't cheap. GoAT support will be in the next `asciidoctor-diagram` release (PR  merged yesterday).

jmooring:
@bep Let's hold off on this for a day or two. There's a new [upstream release](https://github.com/asciidoctor/asciidoctor-diagram/releases/tag/v3.1.0) that implements caching and GoAT diagrams.

jmooring:
Perhaps this is a hangover from the outage(s) a few hours ago, but integration tests fail every time:

> failed to execute binary \"npm\" with args [install]: npm warn tar TAR_ENTRY_ERROR ENOSPC: no space left on device

bep:
>Confidence Score: 15%

There's only a 15% chance you're not human, according to OpenAI. Running on every PR force push doesn't make sense, though ... I'll see if I can fix that.

jmooring:
> There's only a 15% chance you're not human

Yeah, well, I'd argue the number is higher than that, but I appreciate its charitable assessment.

jmooring:
@bep I could use another set of eyes on this. I cannot figure out why the CI runs continue to fail. The new integration test passes:

```text
ok  	github.com/gohugoio/hugo/markup/asciidocext	53.224s
```

But the job dies later in the process.

bep:
@jmooring I will have a look. We have had "space issues" before, which is the reason we currently do not run on Darwin. But the "free-disk-space" should work, and it's strange that it works in other PRs.

bep:
I have tried to narrow it down in #14190 ...  This  I don't understand, it must be some issue on the GitHub side of the fence. I need to look at this with a fresh head tomorrow.

bep:
OK, #14196 should fix it, will merge once it goes green.

jmooring:
It's looks like there are still some space issues or something on Linux.

bep:
OK, now this issue has spread to #14197 -- which I guess also gives me a hint about what the real issue is: Test binary sizes now crossing a breaking threshold. I will see if I can find a fix.

jmooring:
This is ready for final review.

Integration test details:

- Generates GoAT diagrams instead of Ditaa diagrams
- Enables the caching mechanism introduced in v3.0.1 of the asciidoctor-diagram extension
- Runs only when IsRealCI is true
- Runs all subtests on Linux
- Runs 3 of the subtests on Windows

When testing locally on Linux with Ditaa diagrams, the benefits of diagram caching are obvious. When using GoAT diagrams, the benefit is less noticeable because generation is very fast. On Windows, the benefits of diagram caching for simple diagrams are negligible due to the overall slower performance of the Asciidoctor calls.

Current CI test duration:

- Linux: ~50s (runs all 20 subtests)
- Windows: ~30s (runs 3 of 20 subtests)

markup/asciidocext/internal/converter.go
```diff
@@ -105,42 +166,63 @@ func (a *AsciidocConverter) ParseArgs(ctx converter.DocumentContext) []string {
...
-		destinationDir := a.Cfg.Conf.BaseConfig().PublishDir
+		page, ok := ctx.Document.(pageSubset)
+		if !ok {
+			return nil, fmt.Errorf("expected pageSubset, got %T", ctx.Document)
+		}
 
-		if destinationDir == "" {
-			a.Cfg.Logger.Errorln("markup.asciidocext.workingFolderCurrent requires hugo command option --destination to be set")
+		// Derive the outdir document attribute from the relative permalink.
+		relPath := strings.TrimPrefix(page.RelPermalink(), a.Cfg.Conf.BaseURL().BasePathNoTrailingSlash)
```
bep, line 176: This usually works fine until it ... doesn't. But following `relPath` downwards, I think it helps to strictly treat it as a `path` (Unix style slashes) until you create the a filename from it (e.g. via `filepath.Join` that will convert the slashes for you. That is, use `path.Ext` etc. and not `filepath.Ext`.

PR #14142, opened 2025-11-15
static: Preserve .gitignore and .gitattributes in --cleanDestinationDir

### LINKED ISSUE #14097: Please make --cleanDestinationDir spare more than just the .git directory
@jmooring asked me to create this proposal as a result of the discussion in that thread: https://discourse.gohugo.io/t/please-make-taxonomy-bycount-produce-a-stable-ordering/56149

Currently, when Hugo cleans the destination directory, it uses a filter defined in [commands/hugobuilder.go](https://github.com/gohugoio/hugo/blob/b95f7d2552b45b3698a6d191591add4e74236380/commands/hugobuilder.go#L465-L467)  as follows:

```golang
    		syncer.DeleteFilter = func(f fsync.FileInfo) bool {
    			return f.IsDir() && strings.HasPrefix(f.Name(), ".")
    		}
```

So, it refrains from deleting directories prefixed with a dot.

This means that it will spare my `.git` directory, but it won't spare my `.gitignore` file.  So, my `.gitignore` file gets deleted on every single hugo build, which is a bummer.

This is a proposal to extend the filter a bit.

@jmooring suggests that the filter should additionally spare the `.gitignore` and `.gitattributes` files only.

I think it would be better to simply get rid of the `f.isDir() && ` part so that the filter spares anything prefixed with a dot.

Ahamed1846:
This PR improves the behavior of the --cleanDestinationDir feature when syncing static files.

Currently, Hugo deletes all files in the destination directory except for dot-directories when `CleanDestinationDir` is enabled. This causes important files like `.gitignore` and `.gitattributes` to be removed on each build, which is undesirable.

This change updates the DeleteFilter to also preserve `.gitignore` and `.gitattributes` files while keeping the existing behavior for dot-directories intact.

Fixes #14097

### DIFF
--- a/commands/hugobuilder.go
+++ b/commands/hugobuilder.go
@@ -464,7 +464,15 @@ func (c *hugoBuilder) copyStaticTo(sourceFs *filesystems.SourceFilesystem) (uint
 		infol.Logf("removing all files from destination that don't exist in static dirs")
 
 		syncer.DeleteFilter = func(f fsync.FileInfo) bool {
-			return f.IsDir() && strings.HasPrefix(f.Name(), ".")
+			name := f.Name()
+
+			// Keep .gitignore and .gitattributes anywhere
+			if name == ".gitignore" || name == ".gitattributes" {
+				return true
+			}
+
+			// Keep Hugo's original dot-directory behavior
+			return f.IsDir() && strings.HasPrefix(name, ".")
 		}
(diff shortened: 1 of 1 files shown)


bep:
The test failure is my mistake... I will update this branch and re-run the tests soon.

PR #14146, opened 2025-11-16
Dockerfile cleanup: Improve caching, reduce image size, and simplify ENV usage

Dev-GA-Arjun:
This PR cleans up the Dockerfile to make the image smaller and improve build caching.
Changes included:

Added --no-cache to apk add to avoid extra layers

Combined multiple ENV statements into one block

Removed unused/duplicate ARG TARGETPLATFORM

Cleaned up the downloaded tar after extraction

Reordered a few lines for clarity and consistency

These are all minor improvements but help reduce the final image size and keep the Dockerfile easier to maintain.


bep:
Thanks for this. This is most likely an improvement, but I'll have to pass on this for now, reasoning:

1) I don't have time to test this.
2) I see some rather arbitrary removal of comments that doesn't make much sense.

PR #14154, opened 2025-11-16
tpl/collections: Improve collections.D

### LINKED ISSUE #14143: Proposal: Make collections.D handle n > hi gracefully + add error handling

<pre>
$ hugo version 0.152.0

</pre>

yes. 

(**N.B: Proposal generated with support of Claude AI**)

`collections.D` currently returns nil when requesting more items than available (`n > hi`). This proposal suggests capping `n` to `hi` instead, and improving error handling for invalid inputs.

This issue was discussed in detail on the Hugo Discourse from this comment https://discourse.gohugo.io/t/configure-related-content-fallback/56243/8. But you can refer to previous comments for context.

**Current Behavior**

When `collections.D(seed, n, hi)` is called with `n > hi`, it returns nil:

```
if key.n <= 0 || key.hi <= 0 || key.n > key.hi {
    return nil
}
```
Example:

```
{{ $pages := slice "page1" "page2" "page3" }}  {{/* 3 pages */}}
{{ collections.D 42 6 ($pages | len) }}        {{/* Want 6, have 3 */}}
{{/* Returns: nil (empty) */}}
```

**Problem**

The primary documented use case for `collections.D` is selecting random pages:

```
{{ range collections.D seed $n ($pages | len) }}
  {{ with (index $pages .) }}
    <li><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></li>
  {{ end }}
{{ end }}
```

In real-world scenarios with `.Related` content, the number of available related pages varies per post:

* Post A might have 10 related pages
* Post B might have only 3 related pages

With a configurable `$n = 6`, Post B returns nothing instead of showing its 3 available related pages.

Current workaround required:

```
{{- $actualN := cond (lt ($pages | len) $n) ($pages | len) $n }}
{{ range collections.D seed $actualN ($pages | len) }}
```

This defeats the purpose of having a single configurable parameter.

**Expected Behavior**

When `n > hi`, `collections.D` should return all available items (indices `[0, 1, 2, ..., hi-1]`), similar to how `collections.Shuffle` behaves:

```
{{ $pages := slice "page1" "page2" "page3" }}
{{ ($pages | shuffle) | first 6 }}
{{/* Returns: all 3 pages */}}
```

This is the standard behavior users expect from random selection functions.

**Proposed Solution**

See https://discourse.gohugo.io/t/configure-related-content-fallback/56243/11.

"This sounds like a reasonable enhancement:

```
if key.n > key.hi {
    key.n = key.hi  
}
```

And while the body’s open we should consider throwing errors when `key.n <= 0` or when `key.hi <= 0`."

jmooring:
- When n is greater than hi, return sorted slice of length hi in the half-open interval [0, hi).
- Throw errors instead of returning nil

Closes #14143

### DIFF
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -537,27 +537,62 @@ type dKey struct {
 	hi   int
 }
 
-// D returns a slice of n unique random numbers in the range [0, hi) using the provded seed,
-// using  J. S. Vitter's Method D for sequential random sampling, from Vitter, J.S.
-// - An Efficient Algorithm for Sequential Random Sampling - ACM Trans. Math. Software 11 (1985), 37-57.
-// See  https://getkerf.wordpress.com/2016/03/30/the-best-algorithm-no-one-knows-about/
-func (ns *Namespace) D(seed, n, hi int) []int {
-	key := dKey{seed: cast.ToUint64(seed), n: n, hi: hi}
-	if key.n <= 0 || key.hi <= 0 || key.n > key.hi {
-		return nil
+// D returns a sorted slice of unique random integers in the half-open interval
+// [0, hi) using the provided seed value. The number of elements in the
+// resulting slice is n or hi, whichever is less.
+//
+// If n <= hi, it returns a sorted random sample of size n using J. S. Vitter’s
+// Method D for sequential random sampling.
+//
+// If n > hi, it returns the full, sorted range [0, hi) of size hi.
+//
+// If n == 0 or hi == 0, it returns an empty slice.
+//
+// Reference:
+//
+//	J. S. Vitter, "An efficient algorithm for sequential random sampling," ACM Trans. Math. Softw., vol. 11, no. 1, pp. 37–57, 1985.
+//	See also: https://getkerf.wordpress.com/2016/03/30/the-best-algorithm-no-one-knows-about/
+func (ns *Namespace) D(seed, n, hi any) ([]int, error) {
+	seedInt, err := cast.ToInt64E(seed)
+	if err != nil || seedInt < 0 {
+		return nil, fmt.Errorf("the seed value (%v) must be a non-negative integer", seed)
 	}
-	if key.n > maxSeqSize {
-		panic(errSeqSizeExceedsLimit)
+
+	nInt, err := cast.ToIntE(n)
+	if err != nil || nInt < 0 || nInt > maxSeqSize {
+		return nil, fmt.Errorf("the number of requested values (%v) must be a non-negative integer <= %d", n, maxSeqSize)
 	}
-	v, _ := ns.dCache.GetOrCreate(key, func() ([]int, error) {
+
+	hiInt, err := cast.ToIntE(hi)
+	if err != nil || hiInt < 0 || hiInt > maxSeqSize {
+		return nil, fmt.Errorf("the maximum requested value (%v) must be a non-negative integer <= %d", hi, maxSeqSize)
+	}
+
+	if nInt == 0 || hiInt == 0 {
+		return []int{}, nil
+	}
+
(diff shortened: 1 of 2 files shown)


jmooring:
Attributions:
- Initial code review by Gemini
- Conversion to table test assisted by Gemini

tpl/collections/collections_test.go
```diff
@@ -811,7 +883,7 @@ func BenchmarkD2(b *testing.B) {
 		name := fmt.Sprintf("n=%d,max=%d", n, max)
 		b.Run(name, func(b *testing.B) {
 			for b.Loop() {
-				ns.D(seed, n, max)
+				_, _ = ns.D(uint64(seed), uint64(n), uint64(max))
```
bep: Not much practical difference, but it looks odd, so I think we might as well just change the signature of runBenchmark ...

tpl/collections/collections.go
```diff
@@ -533,31 +533,55 @@ func (ns *Namespace) Slice(args ...any) any {
...
+//	J. S. Vitter, "An efficient algorithm for sequential random sampling," ACM Trans. Math. Softw., vol. 11, no. 1, pp. 37–57, 1985.
+//	See also: https://getkerf.wordpress.com/2016/03/30/the-best-algorithm-no-one-knows-about/
+func (ns *Namespace) D(seed, n, hi uint64) ([]int, error) {
+	if n <= 0 {
+		return nil, fmt.Errorf("the number of requested values (%d) must be greater than zero", n)
 	}
-	if key.n > maxSeqSize {
-		panic(errSeqSizeExceedsLimit)
+	if hi <= 0 {
+		return nil, fmt.Errorf("the maximum requested value (%d) must be greater than zero", hi)
```
bep: I don't think we want to make 0 an error for `n` or `hi`. If that was discussed in the original issue, I must have misread it ... Ref the "common use case" in the docs, I think it's not uncommon (or at least not an error case) to run that on some empty collection. I would expect both of these to return an empty slice.

```handlebars
  {{ $p := site.RegularPages }}
  {{ range collections.D time.Now.YearDay 5 ($p | len) }}
    {{ with (index $p .) }}
      <li><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></li>
    {{ end }}
  {{ end }}
```

jmooring:
I'm fine returning nil when n or hi is zero, and will make the changes above.

But using `uint64` for each arg produces unhelpful error messages:

> executing "all.html" at <-1>: expected unsigned integer; found -1

> executing "all.html" at <1.234>: expected unsigned integer; found 1.234

I am going do this instead, and cast (with error checking) as needed:

```text
func (ns *Namespace) D(seed, n, hi any) ([]int, error) 
```

tpl/collections/collections.go
```diff
@@ -537,27 +537,62 @@ type dKey struct {
...
+//
+// If n > hi, it returns the full, sorted range [0, hi) of size hi.
+//
+// If n == 0 or hi == 0, it returns an empty slice.
+//
+// Reference:
+//
+//	J. S. Vitter, "An efficient algorithm for sequential random sampling," ACM Trans. Math. Softw., vol. 11, no. 1, pp. 37–57, 1985.
+//	See also: https://getkerf.wordpress.com/2016/03/30/the-best-algorithm-no-one-knows-about/
+func (ns *Namespace) D(seed, n, hi any) ([]int, error) {
```
bep, line 555: I suggest we make `any` => `uint64`. We now do proper type conversion of  numeric argsuments to methods, so ints, floats etc. should  just work (strings doesn't work, but that sounds like an odd thing to support).

jmooring:
> I suggest we make any => uint64

If we do that, and the user provides a negative number for n or hi, we get unhelpful messages such as:

> executing "all.html" at <-1>: expected unsigned integer; found -1

Issue #13589, opened 2025-04-11
Find index of a page within a collection of pages


c-f-h:
If I have a collection of pages, say all pages in the current category or another taxonomy, I would like to find the numerical index of a specific page (usually, the current page).

This would be useful for outputting something like "this post is k out of n in the series on..."

I guess I could write a linear search using templates, but it seems really clunky and inefficient. Maybe there should be an .IndexOf method on the Pages object?

Example of intended use:

```
{{- $pages := where site.RegularPages "Params.categories" "intersect" .Params.categories -}}
{{- $currentIndex := indexOf $pages . -}}
```

Shiwang0-0:
Hi! 
I implemented an initial Pages.IndexOf method
It performs an O(n) linear search using Hugo's existing Page.Eq semantics and returns -1 when the page isn't found.

This enables the intended use case:
```
{{ $pages := where site.RegularPages "Params.categories" "intersect" .Params.categories }}
{{ $pages = $pages.ByDate }}
{{ $currentIndex := $pages.IndexOf . }}
```

Would this API/approach be acceptable? If so, I'd be happy to open a PR.

bep:
@Shiwang0-0 there's already a `searchPage` func in there that I suspect should be fast enough? If not, we may consider switching that with yours. There should be decent benchmark coverage.

As to this proposal, my only concern here is that `IndexOf` would need to return `-1` for "not found", so we cannot do `{{ with $pages.IndexOf . }}{{ end }}`, but I guess there's no way around that.

Shiwang0-0:
@bep I read through the `searchPage` implementation and I agree that it seems better fit here. It already has  linear search for <1000 pages and binary search for the rest.
The tests and benchmarking for the same is already done so reusing makes more sense

regarding the -1 return and the 0 index with `{{ with $pages.IndexOf . }}`
this seems like a reasonable workaround
```
{{ $index := $pages.IndexOf . }}

{{ if ge $index 0 }}
    This post is {{ add $index 1 }} out of {{ len $pages }}.
{{ end }}
```

does this sound good ?

bep:
>does this sound good ?

Yes.

jmooring:
I think we can close this per https://github.com/gohugoio/hugo/pull/15227.

Also, I am having a difficult time coming up with a real-world example for the docs. When would I use `IndexOf` instead of doing something like this?

```text
{{ range $k, $p := site.RegularPages }}
[...]

PR #14156, opened 2025-11-17
tpl :Add `indexOf` template function for page position lookup

MohammedTharikS:
Adds a new `indexOf` template function that returns the position of a
page within a `Pages` collection. This exposes the existing
`Pages.IndexOf()` method to templates, making it easier for users to
determine page order in lists.


jmooring:
The related issue (https://github.com/gohugoio/hugo/issues/13589) is still a proposal. It has not been accepted.

Per the [Contributing](https://github.com/gohugoio/hugo/blob/master/README.md#contributing) section of this project's README file:

> **Do not submit a pull request until the project lead accepts the proposal.**

Closing.

PR #14158, opened 2025-11-17
transform: Support external files in OpenAPI (#8067)

amarnath-cdr:
This PR updates transform.Open APIDoc From JSON to use the resource handler, allowing it to load OpenAPI documents from remote URLs as well as local files, resolving #8067.


bep:
I suspect the issue creator meant to add external file support in https://gohugo.io/functions/openapi3/unmarshal/

Anyhow, we're not adding another OpenAPI library.

PR #14162, opened 2025-11-17
config: add strict validation for top-level config keys

surjithks138:
This change adds strict validation to the configuration loader by introducing
a new `validateTopLevelConfigKeys` function. After unmarshalling and key
renaming, the config map is checked against an allowed list of known
top-level Hugo configuration keys.

If an unknown key is found, Hugo now returns a clear error, preventing
silent misconfigurations and helping users detect typos or deprecated fields.

This improves config reliability and aligns behavior with stricter validation
expected in modern Hugo setups.


bep:
Thanks for this,

Issue #14165, opened 2025-11-17
cache/filecache: Unable to resolve the :resourceDir token


jmooring:
While working on https://github.com/gohugoio/hugo/pull/14094, I encountered an issue where the _compiled_ value of `caches.misc.dir` does not correctly resolve the `:resourceDir` token.

The `:cacheDir` token works as expected when used in the **uncompiled** value, but `:resourceDir` currently resolves to an empty string.

Many years ago, the `:resourcedir` token was resolved to a specific path. However, in 2018, this behavior was changed to resolve the token to an empty string.

Here's the diff:

https://github.com/gohugoio/hugo/commit/f9b4eb4#diff-1001eddce35a486c3212a0e83856f4d13ed76f50834bc6aee008416e2d61ca9aL184

jmooring:
See <https://github.com/gohugoio/hugo/pull/14167#issuecomment-3546834472>.

xndvaz:
After reading #14167 and the newer #15038, I don't think this is a simple placeholder-resolution bug.

My read is that `:resourceDir` currently has special semantics: it marks the cache as backed by the layered `ResourcesCache` rather than behaving like `:cacheDir`. That matches the maintainer note in #14167 and the historical change in f9b4eb4 to support generated resources from themes.

Before another implementation PR, can we confirm the intended contract?

1. Should `:resourceDir` continue to mean "use the layered ResourcesCache" rather than "expand to the configured resourceDir path"?
2. Should values like `:resourceDir/foo` be rejected with a config error, as mentioned in #14167?
3. Or should Hugo support arbitrary subpaths under `:resourceDir`, even when multiple cache entries use it?

A docs-only PR or regression test that only pins today's behavior seems premature if this is really an unresolved design/API decision. Likewise, an implementation that makes `:resourceDir` expand like `:cacheDir` may accidentally break the layered resources behavior unless that is the intended direction.

PR #14167, opened 2025-11-18
cache/filecache: Fix resolution of :resourceDir token

jmooring:
This is a first pass at resolving #14165, but three existing tests fail. This isn't as simple as I had hoped.

I clearly don't understand the relationship between `resourceDir` and `resourcesGenDir`.


bep:
Quickly scanning the code now:

*  The `:resourceDir` token is a special case, and using it means using the layered  `p.BaseFs.ResourcesCache` (project, theme1, theme2 ...), so allow themes to bundle generated CSS caches.
* The limitation here is that it's not possible to say e.g. `:resourceDir/foo`, and you should get an error for that, which you don't.
* Fixing it so `:resourceDir/foo` works looks tricky, especially considering that you could have `:resourceDir` used in multiple entries.

PR #14172, opened 2025-11-18
Sort colors

bep:


bep:
@jmooring this seem to work, but: Custom Github repo secrets (Open AI key) isn't visible to PR created from fork, which obviously is the common case ... I would be surprised if there are no way around this, will check tomorrow, any hints welcomed.

bep:
Never mind, I found the answer using ... AI.

PR #14205, opened 2025-11-21
Add support for languageCode in i18n translation lookup

### LINKED ISSUE #14204: feat(i18n): prefer `languageCode` translation tables before falling back to key
Hugo currently loads i18n translation tables based only on the language key in the `languages` map (`Lang`). This causes a mismatch when a site sets a more specific `languageCode` but retains a short language key.

Example:

```yaml
languages:
  pt:
    languageCode: pt-BR
```

```text
i18n/
├── en.toml # Default language
├── [No pt.toml file exists here or in themes etc.]
└── pt-br.toml
```

When rendering `{{ T "some-key" }}` for the `pt` site, Hugo falls back to `en.toml` because it looks only for `pt.toml`, ignoring the declared `languageCode: pt-BR` even though a matching `pt-br.toml` file exists.

## Proposed behavior

If `languages.<LANG_KEY>.languageCode` is set:

1. Try loading translations using `languageCode`.
2. If no matching file exists, ~fall back to~ follow current behavior.

## Benefits

The proposed behavior maximizes compatibility with existing behavior, while maximizing utility to site authors. Also it:

- Makes translation lookup consistent with the language declared for the site.
- Supports region-specific translations (`pt-BR`, `pt-PT`, etc.) without needing multiple language keys.
- Improves compatibility with themes/modules that ship region-coded translation files (e.g., `pt-br.toml`, `zh-hans.toml`).

## Related

- #7982
- Downstream context: https://github.com/open-telemetry/opentelemetry.io/issues/8481

## Minimal site for testing

- [lang-code-feature-test-site](https://drive.google.com/file/d/187DO_rV2_wtHyKpXI2Vs_ZbPjJ97NktW/view)

chalin:
- Fixes #14204
- Makes minimal (IMHO) changes, and preserves existing behavior when possible
- Adds test cases:
  - `TestI18nLanguageCode`: Uses `languageCode` for translation lookup when set
  - `TestI18nLanguageCodePreference`: Prefers `languageCode` file when both `Lang.toml` and `languageCode.toml` exist
  - `TestI18nLanguageCodeEqualsLang`: Behavior unchanged when `languageCode` equals `Lang` (case-insensitive)
  - `TestI18nLanguageCodeFallback`: Falls back to default language when `languageCode` file doesn't exist
  - `TestI18nLanguageCodeMultipleLanguages`: Works independently for multiple languages

## Minimal site for testing

- [lang-code-feature-test-site](https://drive.google.com/file/d/187DO_rV2_wtHyKpXI2Vs_ZbPjJ97NktW/view)

### Screenshots (using the test site)

Before:

> Before

After:

> After

/cc @vitorvasc @svrnm

### DIFF
--- a/langs/i18n/i18n_test.go
+++ b/langs/i18n/i18n_test.go
@@ -501,6 +501,168 @@ func TestI18nTranslate(t *testing.T) {
 	}
 }
 
+// TestI18nLanguageCode: when languageCode is set, use it for translation file lookup.
+func TestI18nLanguageCode(t *testing.T) {
+	c := qt.New(t)
+
+	afs := afero.NewMemMapFs()
+
+	// Create only pt-br.toml (no pt.toml)
+	err := afero.WriteFile(afs, filepath.Join("i18n", "pt-br.toml"), []byte("hello = \"Olá\""), 0o755)
+	c.Assert(err, qt.IsNil)
+
+	// Set up config with pt language and languageCode pt-BR
+	cfg := config.New()
+	cfg.Set("defaultContentLanguage", "pt")
+	cfg.Set("languages", map[string]any{
+		"pt": map[string]any{
+			"languageCode": "pt-BR",
+		},
+	})
+
+	d, _ := prepareDeps(afs, cfg)
+
+	ctx := context.Background()
+	actual := d.Translate(ctx, "hello", nil)
+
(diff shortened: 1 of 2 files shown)


bep:
#14204 needs to get out of the "proposal" state before we can start implementing this.

PR #14214, opened 2025-11-25
tpl/urls: Add PathEscape and PathUnescape functions

### LINKED ISSUE #14209: tpl/urls: Add urls.PathUnescape function
This would call Go's [`url.PathUnescape` ](https://pkg.go.dev/net/url#PathUnescape) function.

Reference: <https://discourse.gohugo.io/t/converting-a-linktitle-or-a-title-to-the-default-slug/56279>

jmooring:
Closes #14209

Creation of table tests assisted by Gemini... sort of.

### DIFF
--- a/tpl/urls/urls_test.go
+++ b/tpl/urls/urls_test.go
@@ -15,6 +15,7 @@ package urls
 
 import (
 	"net/url"
+	"regexp"
 	"testing"
 
 	"github.com/gohugoio/hugo/config/testconfig"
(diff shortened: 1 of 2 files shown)


tpl/urls/urls_test.go
```diff
@@ -105,3 +106,88 @@ func TestJoinPath(t *testing.T) {
...
+		{"map", map[string]string{}, "", true, "unable to cast"},
+		{"struct", tstNoStringer{}, "", true, "unable to cast"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := ns.PathEscape(tt.input)
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("PathEscape() error = %v, wantErr %v", err, tt.wantErr)
```
bep: It would great if Gemini could follow Hugo's test styleguide, which is I would say in this case means:

* Always use quicktest for assertions. This means using `c.Run` and not `t.Run`
* This follows the last item, but: Never use `t.Error` (unless inside a goroutine). Every test error is (in my head) fatal.

tpl/urls/urls_test.go
```diff
@@ -105,3 +106,88 @@ func TestJoinPath(t *testing.T) {
...
+		{"map", map[string]string{}, "", true, "unable to cast"},
+		{"struct", tstNoStringer{}, "", true, "unable to cast"},
+		{"malformed hex", "bad%g0escape", "", true, "invalid URL escape"},
+		{"incomplete hex", "trailing%", "", true, "invalid URL escape"},
+		{"single hex digit", "trail%1", "", true, "invalid URL escape"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := ns.PathUnescape(tt.input) // Use PathUnescape here
```
bep: What's the purpose of the above comment?

bep (changes requested):

tpl/urls/urls_test.go
jmooring: Thanks, will change.

tpl/urls/urls_test.go
jmooring: Gemini cruft. Will remove.

PR #14217, opened 2025-11-26
Support languageCode translation lookup when lang translation is missing

### LINKED ISSUE #14204: feat(i18n): prefer `languageCode` translation tables before falling back to key

chalin:
- Fixes #14204 
- Specifically, this implements the behavior described by @bep in [the comment](https://github.com/gohugoio/hugo/issues/14204#issuecomment-3574722684):

  > This setup should give me (the `nn` language user) `i18n/nn.yml` because both exists:
  > 
  > ```
  > -- hugo.toml --
  > [languages.nn]
  > languageCode = "no-NN"
  > -- i18n/nn.yml --
  > -- i18n/no-nn.yml -- 
  > ```
  > 
  > This setup should give me (the `nn` language user) ` i18n/no-NN.yml` (the fallback):
  > 
  > ```
  > -- hugo.toml --
  > [languages.nn]
  > languageCode = "no-NN"
  > -- i18n/no-NN.yml -- 
  > ```
- Adds integration tests, in particular for each of the two cases described above:
  - `TestI18nPrefersLangTranslationBeforeLanguageCode`
  - `TestI18nUsesLanguageCodeWhenLangFileMissing`
- Adds unit tests for the new functionality related to the presence or not of a translation file. See design note below
- Supersedes #14205

---

### Design note

We can't just test `tp.t.Func(lang)` to determine if `lang` has an associated translation file because a translation function is always created for the default language. Because of this I needed to add a bit of extra functionality to determine whether the default lang has an associated translation file or not.

To achieve this I create a new `bundle` type to contain `*i18n.Bundle` and two extra fields.

I've added some unit tests to confirm that "a translation function is always created for the default language", and to exercise the `bundle` type functionality related to knowing whether the default lang has an associated translation file or not.

### DIFF
--- a/langs/i18n/translationProvider.go
+++ b/langs/i18n/translationProvider.go
@@ -35,6 +35,20 @@ import (
 	"github.com/gohugoio/hugo/source"
 )
 
+type bundle struct {
+	i18n                          *i18n.Bundle
+	defaultLang                   string // normalized
+	defaultLangHasTranslationFile bool
+}
+
+func newBundle(defaultLangTag language.Tag) *bundle {
+	return &bundle{
+		i18n:                          i18n.NewBundle(defaultLangTag),
+		defaultLang:                   defaultLangTag.String(),
+		defaultLangHasTranslationFile: false,
+	}
+}
+
 // TranslationProvider provides translation handling, i.e. loading
 // of bundles etc.
 type TranslationProvider struct {
@@ -52,12 +66,12 @@ func (tp *TranslationProvider) NewResource(dst *deps.Deps) error {
 	if err != nil {
 		defaultLangTag = language.English
 	}
-	bundle := i18n.NewBundle(defaultLangTag)
+	bundle := newBundle(defaultLangTag)
 
-	bundle.RegisterUnmarshalFunc("toml", toml.Unmarshal)
-	bundle.RegisterUnmarshalFunc("yaml", metadecoders.UnmarshalYaml)
-	bundle.RegisterUnmarshalFunc("yml", metadecoders.UnmarshalYaml)
-	bundle.RegisterUnmarshalFunc("json", json.Unmarshal)
+	bundle.i18n.RegisterUnmarshalFunc("toml", toml.Unmarshal)
+	bundle.i18n.RegisterUnmarshalFunc("yaml", metadecoders.UnmarshalYaml)
+	bundle.i18n.RegisterUnmarshalFunc("yml", metadecoders.UnmarshalYaml)
+	bundle.i18n.RegisterUnmarshalFunc("json", json.Unmarshal)
 
 	w := hugofs.NewWalkway(
 		hugofs.WalkwayConfig{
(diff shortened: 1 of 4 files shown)


langs/i18n/translationProvider.go
```diff
@@ -128,10 +145,27 @@ func addTranslationFile(bundle *i18n.Bundle, r *source.File) error {
...
+	dst.Translate = tp.getTranslateFunc(dst)
 	return nil
 }
 
+// getTranslateFunc returns the translation function for the language in Deps
+func (tp *TranslationProvider) getTranslateFunc(dst *deps.Deps) func(ctx context.Context, translationID string, templateData any) string {
+	lang := dst.Conf.Language().(*langs.Language)
+	langKey := lang.Lang
+
+	_, ok := tp.t.translateFuncs[langKey]
```
bep, line 157: I will look at the other parts of this later, but this looks the opposite of what we agreed on (ref https://github.com/gohugoio/hugo/issues/14204#issuecomment-3577766774)

chalin: I was puzzled by your pseudo code (and probably should have confirmed) when I looked at it more closely because it doesn't support the scenarios you named in https://github.com/gohugoio/hugo/issues/14204#issuecomment-3574722684. Is that because you changed your mind relative to those two scenarios?

chalin: The pseudocode matches the original PR #14205. Is that the behavior you want?

bep:
@chalin lets complete this in #14225 -- I have moved and adjusted your integration tests there and and added you as a "co contributor". There was too many moving parts in this PR that I didn't fully understand.

chalin:
That works for me, thanks for moving this forward @bep.

Issue #14112, opened 2025-11-05
Panic on new content creation


danshearer:
### What version of Hugo are you using (`hugo version`)?

<pre>
hugo v0.152.2+extended+withdeploy linux/amd64 BuildDate=unknown
</pre>

### Does this issue reproduce with the latest release?
Yes

### Conditions to reproduce

A file with the same name exists only with different case, for example Floop.md or fLoOp.md . There is no problem if floop,md exists with exact case, when hugo correctly reports that the file exists and aborts.

<pre>
hugo new content/articles floop.md
</pre>

### Panic

[user@machine ]$ hugo new content/articles/floop.md
panic: [BUG] no Page found for "/home/user/site/example.org/content/articles/floop.md"

goroutine 1 [running]:
github.com/gohugoio/hugo/create.(*contentBuilder).applyArcheType(0xc000bf6e40, {0xc000a91a00, 0x3b}, {0x5636a68403c0, 0xc00038fec0})
        github.com/gohugoio/hugo/create/content.go:275 +0x346
github.com/gohugoio/hugo/create.(*contentBuilder).buildFile(0xc000bf6e40)
        github.com/gohugoio/hugo/create/content.go:244 +0x1a8
github.com/gohugoio/hugo/create.NewContent.func1()
        github.com/gohugoio/hugo/create/content.go:105 +0x216
github.com/gohugoio/hugo/create.NewContent(0xc000b1cb00, {0x0, 0x0}, {0x7ffc7c90a108, 0x19}, 0x0)
        github.com/gohugoio/hugo/create/content.go:108 +0x5a9
github.com/gohugoio/hugo/commands.newNewCommand.func1({0x0?, 0x0?}, 0x0?, 0xc0008c0840, {0xc000a51320, 0x0?, 0x0?})
        github.com/gohugoio/hugo/commands/new.go:62 +0x18e
github.com/gohugoio/hugo/commands.(*simpleCommand).Run(0x0?, {0x5636a6829f58?, 0x5636a866d2e0?}, 0x0?, {0xc000a51320?, 0xc0007459f8?, 0x0?})
        github.com/gohugoio/hugo/commands/commandeer.go:649 +0x3c
github.com/bep/simplecobra.(*Commandeer).compile.func1(0xc000899a00?, {0xc000a51320?, 0x4?, 0x5636a570c246?})
        github.com/bep/simplecobra@v0.6.1/simplecobra.go:113 +0x4c
[...]

PR #14256, opened 2025-12-11
create: Prevent panic when content file exists with different case

tharunn0:
Fixes https://github.com/gohugoio/hugo/issues/14112 
– prevents the panic in hugo new when a content file exists with different case only.

Adds a small case-insensitive check in CreateContentPlaceHolder and returns a clear error instead of panicking.
--force behaviour is intentionally unchanged (still panics on pure case-conflicts).

mage -v check and mage hugoRace pass.


tharunn0:
Hi @bep  👋
Just checking in on this PR when you have time.

Since it addresses a panic when content files differ by case, I wanted to ask if there’s anything I should adjust or improve to help move it forward.

Happy to make changes or add more tests if needed.

jmooring:
https://github.com/gohugoio/hugo/issues/14112 was fixed in v0.159.0 with https://github.com/gohugoio/hugo/pull/14647.

PR #14332, opened 2025-12-31
snap: Update Node 22 => Node 24

jmooring:


jmooring:
This results in GLIBC errors when trying to run hugo because we're building on core22... Node 22 provided the necessary libs. We may have to go to core24, but that has other issues.

PR #14375, opened 2026-01-13
fix: chroma bump requires test updates

### LINKED ISSUE #14374: Tests updates for Chroma v2.22.0

I'm on [`7165e0fc7e0e5db285eacd5bc833865e25a62247`](https://github.com/gohugoio/hugo/commit/7165e0fc7e0e5db285eacd5bc833865e25a62247).

The issue comes from a change I authored in chroma [`31592d5`](https://github.com/alecthomas/chroma/commit/31592d5b6e3ec9af5d7789015070352ceedf04cf).

It's intended to fix a rendering issue when a line of code exceeds a certain length. Details may be found on issue [#1101](https://github.com/alecthomas/chroma/issues/1101) of chroma.

I'd have a commit ready to use to close this issue, I've opened it just to open the PR. 
I've adapted the hugo tests to adapt accordingly. Chroma highlights include a new CSS property `-webkit-text-size-adjust: none;`

Hope that helps,
Valentin

jacqueminv:
FIXES #14374

### DIFF
--- a/tpl/tplimpl/shortcodes_integration_test.go
+++ b/tpl/tplimpl/shortcodes_integration_test.go
@@ -229,12 +229,12 @@ An inline {{< highlight go "hl_inline=true" >}}fmt.Println("Value of i:", i)Hell
 
 	b := hugolib.Test(t, files)
 
-	b.AssertFileContent("public/p1/index.html", "576ee13be18ddba2")
+	b.AssertFileContent("public/p1/index.html", "4f7e70e3bc7702fc")
 	b.AssertFileContent("public/p2/index.html", "3f97c98e654f23a2")
-	b.AssertFileContent("public/p3/index.html", "7634b47df1859f58")
-	b.AssertFileContent("public/p4/index.html", "385a15e400df4e39")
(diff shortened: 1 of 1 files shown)


bep:
We will handle this in #14368 (or similar issue; the test adjustments needs to be in the same PR).

/cc @jmooring I'm not convinced that the hash based assertions used in these tests are the right approach here, I suspect it's better to assert on the output and make it less fragile by doing some whitespace normalisation. For one, it's not possible for me to look at the diff in this PR and tell if it's doing the right thing.

jmooring:
> I'm not convinced that the hash based assertions used in these tests are the right approach

I'm not convinced either. I used this approach out of expediency and making the tests easier to read, thinking that the hash changes would be infrequent. And yes, when they occur, it takes some additional digging (i.e., looking at the logs for the GitHub test action).

Whether the hash changes have been frequent or infrequent is somewhat subjective, but to-date I feel they have been infrequent. I would say:

1. If the frequency of the hash changes increases, let's do something about it.
2. If you feel that the hash changes are currently too frequent, create an issue and assign it to me.

PR #14386, opened 2026-01-16
hugolib: Fix relative alias generation

### LINKED ISSUE #14381: Alias redirects to wrong page / output format
Here’s a small repro: [hugo-alias-and-print-test2.zip](https://drive.google.com/file/d/1sqVI-T4Q27CpcD0vhJutdWUJXqJ6V3vw/view?usp=sharing). It’s features are:

* Defines a PRINT output format, declared as `rel: alternate`, and used for sections in addition to HTML
* Two languages, `en` and `fr`
* `fr` pages have `en` pages as fallback
* `en/about/_index.md` declares`aliases: [info]` in the front matter

Building the site and inspecting the alias file:

```console
$ hugo
Start building sites … 
hugo v0.154.5-a6f99cca223a29cad1d4cdaa6a1a90508ac1da71+extended darwin/arm64 BuildDate=2026-01-11T20:53:23Z VendorInfo=gohugoio


                  │ EN │ FR 
──────────────────┼────┼────
 Pages            │ 10 │  9 
 Paginator pages  │  0 │  0 
 Non-page files   │  0 │  0 
 Static files     │  0 │  0 
 Processed images │  0 │  0 
 Aliases          │  3 │  2 
 Cleaned          │  0 │  0 

Total in 21 ms
$  cat public/info/index.html
```

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>https://example.org/_print/about/</title>
    <link rel="canonical" href="https://example.org/about/">
    <meta charset="utf-8">
    <meta http-equiv="refresh" content="0; url=https://example.org/_print/about/">
  </head>
</html>
```

Notice how the alias file has the right canonical URL, but it refreshes to the print version of the page rather than the canonical version of the page.

**Expected output**: the alias file refresh target URL should be https://example.org/about/ not https://example.org/_print/about/.

---

Reference: <https://discourse.gohugo.io/t/how-is-the-canonical-page-determined/56549>

jmooring:
Closes #14381

This turned out to be a trivial change.

Including the expected site structure in the test comments was helpful during test development, but let me know if you would prefer to remove them.

### DIFF
--- a/hugolib/alias_test.go
+++ b/hugolib/alias_test.go
@@ -191,3 +191,194 @@ aliases:
 
 	b.AssertPublishDir("n/index.html", "yes/index.html", "no/index.html", "yes/index.html")
 }
+
+// Issue 14381
+func TestIssue14381(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+baseURL      = 'https://example.org'
+disableKinds = ['home', 'rss', 'sitemap', 'taxonomy', 'term']
+
+[outputFormats.print]
+  isHTML        = IS_HTML
+  mediaType     = 'text/html'
+  path          = 'print'
+
+[outputs]
+  page    = ['html', 'print']
+  section = ['html', 'print']
+-- content/foo/s1/_index.md --
+---
+title: s1
+aliases: [s1-alias]
+---
+-- content/foo/s1/p1.md --
+---
+title: p1
+aliases: [p1-alias]
+---
+-- content/foo/s2/_index.md --
+---
+title: s2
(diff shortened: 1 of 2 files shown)


jmooring:
It looks like this was merged while still in draft. I had kept it in draft status because I'm not yet confident in the multilingual behavior, and we need better test coverage for those cases. I’ll work on a follow-up PR today to address these gaps.

bep:
@jmooring OK, I apologise ... It looked finished, so I assumed that you forgot to undraft it. I'll try to behave in the future ...

jmooring:
No problem.

PR #14387, opened 2026-01-17
fix: typos in comments

NAM-MAN:
Fix typos in comments:
- `is is` → `is` (x3)
- `a a` → `a`
- `to to` → `to` (x3)

### DIFF
--- a/config/allconfig/docshelper.go
+++ b/config/allconfig/docshelper.go
@@ -19,7 +19,7 @@ import (
 	"github.com/gohugoio/hugo/docshelper"
 )
 
-// This is is just some helpers used to create some JSON used in the Hugo docs.
+// This is just some helpers used to create some JSON used in the Hugo docs.
 func init() {
 	docsProvider := func() docshelper.DocProvider {
 		cfg := config.New()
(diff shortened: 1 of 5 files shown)


common/collections/append.go
```diff
@@ -82,7 +82,7 @@ func Append(to any, from ...any) (any, error) {
 					return from[0], nil
 				}
 
-				// If we get []string []string, we append the from slice to to
+				// If we get []string []string, we append the from slice to
```
bep: This isn't entirely correct.

transform/chain.go
```diff
@@ -73,7 +73,7 @@ func (ft fromToBuffer) To() io.Writer {
 }
 
 // Apply passes the given from io.Reader through the transformation chain.
-// The result is written to to.
+// The result is written to.
```
bep: "to to" sounds odd, but it is correct.

NAM-MAN:
Thanks @bep for catching that! You're absolutely right - I missed that `to` is a variable/parameter name in those cases, so "to to" is actually correct (preposition + variable name).

I've reverted the changes to `append.go` and `chain.go`, keeping only the genuine typo fixes:
- `is is` → `is` (x3)
- `a a` → `a`  
- `to to` → `to` (only in `pagelexer_shortcode.go` where it's not a variable name)

Sorry for the noise! 🙇

PR #14395, opened 2026-01-19
hugolib: Fix multilingual alias generation

### LINKED ISSUE #14388: hugolib: Fix multilingual single-host alias generation
While PR #14386 corrected alias publication paths for monolingual and multilingual multihost configurations, multilingual single-host sites still exhibit incorrect alias publication paths under certain conditions. 

1.  When an alias path in front matter is absolute, the alias is incorrectly published relative to the `publishDir`. It _should_ be published relative to the _site root_.
  
    content path|front matter alias|expected publication path|actual publication path
    :--|:--|:--|:--
    `content/fr/foo/s1/p1.md`|`/p1-alias`|`public/fr/my-alias/index.html`|`public/my-alias/index.html`

    That means, for example, that aliases for French content are clobbered by aliases for German content. Yes, this is a breaking change, but what we have is already broken.

2. When an output format has a non-empty `path` parameter, the `path` directory is incorrectly created in the root of the `publishDir`. It _should_ be published in the _site root_. Consider an output format where the `path` is set to `print`:

    content path|front matter alias|expected publication path|actual publication path
    :--|:--|:--|:--
    `content/fr/foo/s1/p1.md`|`p1-alias`|`public/fr/print/foo/s1/p1-alias/index.html`|`public/print/fr/foo/s1/p1-alias/index.html`

> [!note]
> As of v0.153.0, the _site root_ may be composed of language, role, and/or version. For example, `public/guest/v1.0.0/fr`.

jmooring:
Closes #14388

AI disclosure:

Agent|Task|Level of coaching required
:--|:--|:--
Gemini|Creation of the `paths.InjectSegment` function|High
Gemini|Creation of the `TestInjectSegment` table test|Low
Gemini|Improving comment clarity and accuracy|High

### DIFF
--- a/hugolib/alias_test.go
+++ b/hugolib/alias_test.go
@@ -232,10 +232,10 @@ aliases: [/p2-alias]
 -- layouts/all.html --
 {{ .Title }}
 `
+	// ------------------------------------------------------------------------
+	// Test 1: Create aliases for the html output format only
+	// ------------------------------------------------------------------------
 
-	// Test 1
-	// Expected site structure:
-	//
 	// public/
 	// ├── foo/
 	// │  ├── s1/
@@ -270,6 +270,7 @@ aliases: [/p2-alias]
 	f := strings.ReplaceAll(files, "IS_HTML", "false")
 	b := Test(t, f)
 
+	// output format: html
 	b.AssertFileContent("public/foo/s1-alias/index.html",
 		`<title>https://example.org/foo/s1/</title>`,
 		`<link rel="canonical" href="https://example.org/foo/s1/">`,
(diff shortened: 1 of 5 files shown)


hugolib/alias_test.go
```diff
@@ -382,3 +388,267 @@ aliases: [/p2-alias]
 		`<meta http-equiv="refresh" content="0; url=https://example.org/print/foo/s2/p2/">`,
 	)
 }
+
+// Issue 14388
+func TestIssue14388(t *testing.T) {
```
bep, line 393: I'm adding this as a ... comment for later;

So, I'm a fan of having the issue number in the test name (that way, if it fail in e.g. a CI build, I can go to the issue and get some sense of what's going on); but I also like having a descriptive prefix so I understand what it's generally fixing when I read the source (like now), e.g. `TestMultilingualSingleHostAliasIssue14388`.

PR #14420, opened 2026-01-23
tpl: Fix partial decorator panic when partial returns falsy

### LINKED ISSUE #14419: tpl: Panic "partial decorator ID mismatch" when partial returns falsy value
## Description

Since Hugo v0.154.0, using nested `{{ with partial/partialCached ... }}` calls where one partial returns a falsy value (`false`, `nil`, `""`) causes a panic:

```
partial decorator ID mismatch
```

## Reproduction

**Minimal reproduction repository:** https://github.com/simonheimlicher/hugo-partial-decorator-mismatch

```bash
git clone https://github.com/simonheimlicher/hugo-partial-decorator-mismatch
cd hugo-partial-decorator-mismatch
hugo
```

### Trigger pattern

The bug requires nested partials where one returns falsy:

```html
{{/* outer.html */}}
{{ with partialCached "middle.html" . .RelPermalink }}
  {{ $result = . }}
{{ end }}

{{/* middle.html */}}
{{ with partialCached "inner-falsy.html" . "key" }}
  {{/* This block is skipped when inner returns false */}}
{{ end }}

{{/* inner-falsy.html */}}
{{ return false }}
```

### Actual behavior (v0.154.x)

```
ERROR error calling partialCached: execute of template failed:
  error calling _popPartialDecorator: partial decorator ID mismatch
```

### Expected behavior

Site builds successfully (as it did in v0.153.x).

## Root cause

The bug is in `tpl/tplimpl/templatetransform.go` in the `handleWithPartial` function.

When `{{ with partial "foo" }}` is transformed for partial decorators:

1. `_pushPartialDecorator("HASH")` executes unconditionally as part of the `with` condition
2. If the partial returns a falsy value, the `with` block is skipped
3. `_popPartialDecorator("HASH")` is INSIDE the `with` block, so it never executes
4. The decorator ID stack has an orphan entry
5. The next partial call tries to pop a different ID → panic

## Affected versions

- v0.154.0 through v0.154.5 (current)

## Fix

The fix adds a pop call to the else branch of the `with` statement, ensuring the decorator stack is always balanced regardless of whether the partial returns truthy or falsy.

simonheimlicher:
## Summary

Fixes #14419

When `{{ with partial "foo" }}` returns a falsy value (`false`, `nil`, `""`), the `with` block is skipped but `_popPartialDecorator` was only called inside the `with` block. This left an orphan entry on the decorator stack, causing "partial decorator ID mismatch" panic on subsequent partial calls.

**Reproduction:** https://github.com/simonheimlicher/hugo-partial-decorator-mismatch

### Root cause

In `tpl/tplimpl/templatetransform.go`, the `handleWithPartial` function transforms `{{ with partial ... }}` to include decorator push/pop calls. The push happens unconditionally (as part of the condition evaluation), but the pop was only inside the `with` block's List (truthy branch). When the partial returns falsy, the block is skipped and the pop never executes.

### Fix

Add `_popPartialDecorator` call to the else branch (`ElseList`) of the `with` statement, ensuring the decorator stack is always balanced regardless of whether the partial returns truthy or falsy.

## Test plan

- [x] New test `TestDecoratorPartialFalsyReturn` reproduces the exact bug pattern
- [x] Test passes with fix, panics without it
- [x] All existing decorator tests pass
- [x] Verified with reproduction site

## AI assistance disclosure

This PR was written primarily by Claude Code with Opus 4.5. All generated code has been reviewed and understood by the contributor.

### DIFF
--- a/tpl/templates/decorator_falsy_test.go
+++ b/tpl/templates/decorator_falsy_test.go
@@ -1,0 +1,63 @@
+// Copyright 2025 The Hugo Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package templates_test
+
+import (
+	"testing"
+
+	"github.com/gohugoio/hugo/hugolib"
+)
+
+// TestDecoratorPartialFalsyReturn tests that partial decorators work correctly
+// when a partial returns a falsy value (false, nil, ""). This was a bug in
+// v0.154.0-v0.154.5 where the decorator stack would become unbalanced.
+// See https://github.com/gohugoio/hugo/issues/14419
+func TestDecoratorPartialFalsyReturn(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+disableKinds = ["section", "taxonomy", "term", "sitemap", "RSS"]
+-- content/p1.md --
+---
+title: "Page 1"
+---
+-- layouts/_partials/a.html --
+{{ $result := dict }}
+{{ with partialCached "b.html" . .RelPermalink }}
+  {{ $result = . }}
+{{ end }}
+{{ return $result }}
+-- layouts/_partials/b.html --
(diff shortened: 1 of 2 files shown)


bep:
I suggest you rebase this against the master. I have done some "spac savings" work in the CI in th master branch.

simonheimlicher:
@bep It turned out that when running `mage -v check` locally on macOS 15 on a MacBook Pro M2, there is a segmentation fault. I have opened a PR to address the root cause #14421. 
However, right now, GH Actions seem to fail due to chocolatey: [[NuGet] Response status code does not indicate success: 504 (Gateway Time-out)](https://github.com/gohugoio/hugo/actions/runs/21303045060/job/61325017126?pr=14421)

bep:
@simonheimlicher the other branch is merged.

PR #14421, opened 2026-01-23
resources: Fix race condition in test helper

simonheimlicher:
## Summary

Fix race condition in `fetchResourceForSpec` test helper that causes segfault when `TestImageTransformConcurrent` runs with `imageProcWorkers > 1`.

- The function was reusing a single `ReadSeekCloser` across multiple goroutines
- Exposed by commit e569dd59a ("Misc webp performance work") which increased `imageProcWorkers` from 1 to 2
- Fix: create a new reader on each call to the open function

## Test plan

- [x] `go test -race -run TestImageTransformConcurrent ./resources` passes
- [x] `mage check` passes locally

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,7 +122,6 @@ jobs:
         name: Test
         run: |
           mage -v test
-          go clean -i -testcache
         env:
           HUGO_BUILD_TAGS: extended,withdeploy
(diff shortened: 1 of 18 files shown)


bep:
I will merge once it go green ... Thanks.

simonheimlicher:
Thanks, I will now rebase #14420 onto master; fingers crossed that this one goes green as well this time.

PR #14495, opened 2026-02-09
markup/highlight: Allow lineNos to be true, false, "inline", or "table"

### LINKED ISSUE #13481: markup/highlight: Allow lineNos to be true, false, "inline", or "table"
Fenced code blocks, the highlight shortcode, and the `transform.Highlight` function allow you to set `lineNos` to `true`, `false`, `inline`, or `table`.

But you can't use the last two in your site configuration. 

It would be helpful if the options were consistent everywhere we use and document them:

Image

jmooring:
Closes #13481

### DIFF
--- a/markup/highlight/highlight_integration_test.go
+++ b/markup/highlight/highlight_integration_test.go
@@ -14,8 +14,10 @@
 package highlight_test
 
 import (
+	"strings"
 	"testing"
 
+	qt "github.com/frankban/quicktest"
 	"github.com/gohugoio/hugo/hugolib"
 )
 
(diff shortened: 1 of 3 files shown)


markup/highlight/highlight_integration_test.go
```diff
@@ -129,3 +131,75 @@ xəx := 0
 		 <div class="highlight no-prose"><pre
 	`)
 }
+
+func TestHighlightLineNos(t *testing.T) {
+	t.Parallel()
+	c := qt.New(t)
```
bep: Nip pick, but `c` here is not needed. I don't think it hurts, but it confused me.

PR #14505, opened 2026-02-12
all: Change site to project where appropriate

### LINKED ISSUE #14504: all: Change "site" to "project" where contextually appropriate
This issue is about using the right words in the right places. A Hugo _project_ consists of one or more _sites_.

This should primarily focus on user-facing strings, but there is at least one obvious internal usage: the functions related to creating a new project skeleton.

jmooring:
This primarily focuses on:

- The cmd package
  - [x] Use `hugo new project` instead of `hugo new site`
  - [x] Alias the `site` command to the `project` command (without deprecation)
- The skeletons package
  - [x] Move the project skeleton files from `create/skeletons/site` to `create/skeletons/project`
  - [x] Change the names of related functions

There are many other internal uses of the word "site". The vast majority of those are accurate. There may be a few  internal instances of the word "site" that are inaccurate, but we can handle those over time if needed. This PR is intended to be a surgical change addressing the immediate, user-facing needs.

Closes #14504

### DIFF
--- a/commands/commandeer.go
+++ b/commands/commandeer.go
@@ -520,8 +520,8 @@ func (r *rootCommand) initRootCommand(subCommandName string, cd *simplecobra.Com
 		commandName = subCommandName
 	}
 	cmd.Use = fmt.Sprintf("%s [flags]", commandName)
-	cmd.Short = "Build your site"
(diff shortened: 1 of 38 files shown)


jmooring:
Rebased.

PR #14512, opened 2026-02-13
hugolib: Add hugo.Sites and .Site.IsDefault(), modify .Site.Sites

### LINKED ISSUE #14479: Consider deprecating .Site.Sites in favor of hugo.Sites
Conceptually, I think `hugo.Sites` makes more sense than `.Site.Sites`. I'm pretty sure we bolted `Sites` onto `Site` long before we had a `hugo` function namespace.

I created this issue while working on the documentation. Historically we have (at times) used the terms _site_ and _project_ interchangeably, which was fine for a monolingual (single site) project. But going forward we need to get better about using these terms consistently: a _project_ consists of one or more _sites_.

And `.Site.Sites` isn't fitting into my mental model.

---

See related: https://github.com/gohugoio/hugo/issues/14481

### LINKED ISSUE #14481: Add Site.IsDefault and make .Sites return sites in all dimensions
We have `IsDefault` methods on `Language`, `Role`, and `Version`. It would be convenient if we could do the same for `Site`. Use case:

Consider this template (`layouts/home.redirects`) to create a `_redirects` file:

```text
{{- if not (hugo.Store.Get "has_printed_redirects") -}}
  {{- range .Sites -}}
    {{- range $p := .Pages -}}
      {{- range .Aliases -}}
        {{- if findRE `\s` . -}}
          {{- errorf "One of the front matter aliases in %q contains whitespace" $p.String -}}
        {{- end -}}
        {{- printf "%s %s 301\n" . $p.RelPermalink -}}
      {{- end -}}
    {{- end -}}
  {{- end -}}
  {{- hugo.Store.Set "has_printed_redirects" true -}}
{{- end -}}
```

To make sure we do this only once, we've wrapped in a conditional Store check. It would be slightly more convenient to do this:

```text
{{- if .Site.IsDefault -}}
  {{- range .Sites -}}
    ...
  {{- end -}}
{{- end -}}
```

There may be a better way to do this; I'm open to suggestions.

---

See related: https://github.com/gohugoio/hugo/issues/14479

jmooring:
Changes:

- Add `hugo.Sites` to return all sites for all dimensions
- Modify `.Site.Sites` to return all sites for all dimensions
- Deprecate `.Site.Sites` in favor of hugo.Sites
- Add `.Site.IsDefault()` to report whether the current site is the
  default site across all dimensions
- Consolidate tests

Closes #14479
Closes #14481


common/hugo/hugo.go
```diff
@@ -174,8 +184,13 @@ type ConfigProvider interface {
...
 }
 
+// SitesProvider provides access to all sites.
+type SitesProvider interface {
+	Sites() any
+}
+
 // NewInfo creates a new Hugo Info object.
-func NewInfo(conf ConfigProvider, deps []*Dependency) HugoInfo {
+func NewInfo(conf ConfigProvider, deps []*Dependency, sitesProvider ...SitesProvider) HugoInfo {
```
bep: I agree it looks tempting to just add a vararg options, but I suggest replacing `conf` with `opts HugoInfoOptions` or something that has fields for both ...

hugolib/hugo_sites.go
```diff
@@ -104,6 +104,7 @@ type HugoSites struct {
 
 	printUnusedTemplatesInit sync.Once
 	printPathWarningsInit    sync.Once
+	deprecateSiteSitesOnce   sync.Once
```
bep: There's an obvious naming standard established above ...

hugolib/hugo_sites.go
```diff
@@ -122,6 +123,24 @@ type HugoSites struct {
 	buildCounter atomic.Uint64
 }
 
+// hugoSitesSitesProvider is a wrapper that implements hugo.SitesProvider.
+// This avoids naming conflict with HugoSites.Sites field.
+type hugoSitesSitesProvider struct {
+	h *HugoSites
+}
+
+func (sp hugoSitesSitesProvider) Sites() any {
```
bep, line 133: I think this can be simplified to:

```
return slices.Collect(h.allSites(nil))
```

common/hugo/hugo.go
bep: OK; I didn't notice the draft status, I'll stop reading now, but I'll publish these comments, as I think they're useful.

jmooring:
Related: <https://github.com/gohugoio/hugo/issues/14521>

hugolib/site.go
```diff
@@ -627,15 +634,28 @@ func (s *Site) LanguageCode() string {
...
+// Returns all sites for all dimensions.
+// Deprecated: Use hugo.Sites instead.
 func (s *Site) Sites() page.Sites {
-	sites := make(page.Sites, len(s.h.Sites))
-	for i, s := range s.h.Sites {
-		sites[i] = s.Site()
+	s.h.printSiteSitesDeprecationInit.Do(func() {
+		hugo.Deprecate(".Site.Sites", "Use hugo.Sites instead.", "v0.156.0")
+	})
+	var sites page.Sites
```
bep, line 643: Why not `slices.Collect(s.h.allSites(nil))` .... or something.

PR #14522, opened 2026-02-15
Fix for badly parsed formats.FromFilename

mikekasprzak:
While working on a fix for #10449, I stumbled across a bug with the output formats "FromFilename" method.

Example:

* `myfile.amp.html` is detected as an AMP file :heavy_check_mark: 
* `my.file.has.dots.amp.html` is not detected as an AMP file :x: 
* `myfile.amp.html.is.wrong` is detected as an AMP file :x:

The "file format detection based on file extension" code split the filename at `.`'s, and mistakenly assumes the extension will always be in the same position in the array.

This PR fixes that.


output/outputFormat_test.go
```diff
@@ -122,6 +122,12 @@ func TestGetFormatByFilename(t *testing.T) {
 	f, found := formats.FromFilename("my.amp.html")
 	c.Assert(found, qt.Equals, true)
 	c.Assert(f, qt.Equals, AMPFormat)
+	f, found = formats.FromFilename("my.other.amp.html")
+	c.Assert(found, qt.Equals, true)
+	c.Assert(f, qt.Equals, AMPFormat)
+	f, found = formats.FromFilename("my.amp.html.invalid_file_extension")
```
bep, line 128: This is not a valid case. The case above is good, this isn't.

mikekasprzak: How is this not a valid case? One of the bugs this fixes was exactly this: when the filename contains extra dots, it will fail to correctly detect the file format.

EDIT: Unless you mean the '.invalid_file_extension' case is bad. That said this was also a bug that this code fixes.

EDIT2: Ah, I just noticed that the `if outFormat != "" {` block was wrong. It should fallback to `GetByName(ext)` if getting by outFormat fails.

PR #14581, opened 2026-03-01
docs: fix deprecated Wikipedia link for allowlist in GetRemote

stuckvgn:
The `[allowlist]` reference link in GetRemote.md points to `https://en.wikipedia.org/wiki/Whitelist`, which Wikipedia has deprecated and renamed to Allowlist. This PR updates the URL to `https://en.wikipedia.org/wiki/Allowlist` to match the current article location.

*This PR was identified and generated with AI assistance (Claude Code).*

### DIFF
--- a/docs/content/en/functions/resources/GetRemote.md
+++ b/docs/content/en/functions/resources/GetRemote.md
@@ -234,7 +234,7 @@ Note that the entry above is:
 - An array of [regular expressions](g)
 
 [`try`]: /functions/go-template/try
-[allowlist]: https://en.wikipedia.org/wiki/Whitelist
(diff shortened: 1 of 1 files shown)


bep:
The source for the Hugo documentation lives in its own repository, so we would appreciate if you could open this PR directly in `hugoDocs`:  https://github.com/gohugoio/hugoDocs/pulls

We have a copy in this repository as a Git sub tree, and do allow documentation changes that's related to code contributions, but we try to avoid general documentation changes, as that will lead to merge conflicts later. Also, adding the changes to the `hugoDocs` will getting them faster out on the web.

PR #14598, opened 2026-03-06
tpl/strings: Add strings.ReplacePairs function

### LINKED ISSUE #14594: Add strings.NewReplacer
https://pkg.go.dev/strings#NewReplacer

jmooring:
Closes #14594

I recognize that the type checking (`types.TypeToString`) is more strict than in other `strings` functions, but for this particular function using something like `cast.ToStringSliceE` just doesn't feel right. So we're strict (`string` or any of the `template` types) on both the input and the replacement pairs.

The signature is:

```text
strings.ReplacePairs OLD NEW [OLD NEW ...] INPUT
```

So you can do this:

```go-html-template
{{ "aabbcc" | strings.ReplacePairs "a" "x" "b" "y" "c" "z" }} → xxyyzz
```

Or this:

```go-html-template
{{ $pairs := slice
  "a" "x"
  "b" "y"
  "c" "z"
}}
{{ "aabbcc" | strings.ReplacePairs $pairs }} → xxyyzz
```

docs: <https://deploy-preview-3424--gohugoio.netlify.app/functions/strings/replacepairs/>

### DIFF
--- a/tpl/strings/init.go
+++ b/tpl/strings/init.go
@@ -145,6 +145,20 @@ func init() {
 			},
 		)
 
+		ns.AddMethodMapping(ctx.ReplacePairs,
+			nil,
+			[][2]string{
+				{
+					`{{ "aab" | strings.ReplacePairs "a" "b" "b" "c" }}`,
+					`bbc`,
+				},
+				{
+					`{{ "aab" | strings.ReplacePairs (slice "a" "b" "b" "c") }}`,
+					`bbc`,
+				},
+			},
+		)
+
 		ns.AddMethodMapping(ctx.ReplaceRE,
 			[]string{"replaceRE"},
 			[][2]string{
--- a/tpl/strings/strings.go
+++ b/tpl/strings/strings.go
@@ -23,6 +23,8 @@ import (
 	"unicode"
 	"unicode/utf8"
 
+	"github.com/gohugoio/hugo/common/hmaps"
+	"github.com/gohugoio/hugo/common/hreflect"
 	"github.com/gohugoio/hugo/common/text"
 	"github.com/gohugoio/hugo/deps"
 	"github.com/gohugoio/hugo/helpers"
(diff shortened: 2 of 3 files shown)


tpl/strings/strings.go
```diff
@@ -34,14 +36,25 @@ import (
 
 // New returns a new instance of the strings-namespaced template functions.
 func New(d *deps.Deps) *Namespace {
-	return &Namespace{deps: d}
+	if d.MemCache == nil {
+		panic("must provide MemCache")
+	}
+	return &Namespace{
+		deps: d,
+		cacheReplacer: dynacache.GetOrCreatePartition[string, *strings.Replacer](
```
bep: I would prefer we use the simpler cache e.g. used in `collections.D` for this. And only use this cache when either/or: 1. The cache can get big/take up much memory (e.g. unmarshal), 2. It needs eviction matching during rebuilds.

tpl/strings/init.go
```diff
@@ -145,6 +145,20 @@ func init() {
...
 
+		ns.AddMethodMapping(ctx.ReplacePairs,
+			nil,
+			[][2]string{
+				{
+					`{{ strings.ReplacePairs "a" "b" "b" "c" "aab" }}`,
+					`bbc`,
+				},
+				{
+					`{{ strings.ReplacePairs (slice "a" "b" "b" "c") "aab" }}`,
```
bep: It would be great if the examples followed the pipe construct used in the docs.

tpl/strings/strings.go
```diff
@@ -251,6 +264,66 @@ func (ns *Namespace) Replace(s, old, new any, limit ...any) (string, error) {
...
+		case []string:
+			p = v
+		case []any:
+			p = make([]string, len(v))
+			for i, elem := range v {
+				str, ok := types.TypeToString(elem)
+				if !ok {
+					return "", fmt.Errorf("replacement pairs must be strings, got %T", elem)
+				}
+				p[i] = str
```
bep: Missing a default case here.

tpl/strings/strings.go
```diff
@@ -251,6 +264,66 @@ func (ns *Namespace) Replace(s, old, new any, limit ...any) (string, error) {
...
 
+// ReplacePairs returns a copy of a string with multiple replacements performed
+// in a single pass. The last argument is the source string. Preceding arguments
+// are old/new string pairs, either as a slice or as individual arguments.
+func (ns *Namespace) ReplacePairs(args ...any) (string, error) {
+	if len(args) < 2 {
+		return "", fmt.Errorf("requires at least 2 arguments")
+	}
+
+	ss, ok := types.TypeToString(args[len(args)-1])
```
bep: Having this func behave differently compared to Replace in this area is surprising -- they should be interchangeable. The most surprising part is types that implements the `fmt.Stringer` interface -- and especially for types where we implement that method later, e.g. when we changed `Summary` from string to a struct (I think).

tpl/strings/strings_test.go
```diff
@@ -50,7 +54,7 @@ func TestChomp(t *testing.T) {
 		result, err := ns.Chomp(test.s)
 
 		if b, ok := test.expect.(bool); ok && !b {
-			c.Assert(err, qt.Not(qt.IsNil))
+			c.Assert(err, qt.IsNotNil)
```
bep: Having totally unrelated changes in a PR that should be reviewed by others isn't great. I may do these drive-by edits myself, but not when I hand it over to others to review.

PR #14603, opened 2026-03-08
testing: Modernize quicktest assertions

jmooring:
- Replace qt.Not(qt.IsNil) with qt.IsNotNil
- Replace manual len() comparisons with qt.HasLen

### DIFF
--- a/cache/dynacache/dynacache_test.go
+++ b/cache/dynacache/dynacache_test.go
@@ -58,10 +58,10 @@ func TestCache(t *testing.T) {
 
 	opts := OptionsPartition{Weight: 30}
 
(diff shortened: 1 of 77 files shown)


bep:
I'm closing this. I have had a similar PR before, and my take then was:

* `qt.HasLen` almost always a bad choice, because the output you get on error is very noice (printing of big slices). The important part is the number. If that fails, you need to do some manual digging.
*  `qt.Not(qt.IsNil)` with `qt.IsNotNil` is a cosmetic change; I agree it's nicer to look at, so I often make that changes on _tests I'm working on_, but it's not worth its own mass commit.

jmooring:
Fair enough.

PR #14616, opened 2026-03-13
hugolib: Fix server no watch

### LINKED ISSUE #14615: Site fails to load with hugo server when --watch=false
This is consistently reproducible with v0.153.0 and later, even with the simplest of projects.

After running `hugo server --watch=false`, open the site in your browser. Instead of seeing the home page, you will see this:

Image

Reference: <https://discourse.gohugo.io/t/error-when-loading-site-after-successful-build/56854/>

jmooring:
Closes #14615

This was just a missed condition check. The error message is "Build called multiple times when not in watch or server mode" but we were not checking for server mode.

### DIFF
--- a/hugolib/hugo_sites_build.go
+++ b/hugolib/hugo_sites_build.go
@@ -61,7 +61,7 @@ import (
 // Build builds all sites. If filesystem events are provided,
 // this is considered to be a potential partial rebuild.
 func (h *HugoSites) Build(config BuildCfg, events ...fsnotify.Event) error {
-	if h.isRebuild() && !h.Conf.Watching() {
+	if h.isRebuild() && !h.Conf.Watching() && !h.Conf.Running() {
 		return errors.New("Build called multiple times when not in watch or server mode (typically with hugolib.Test(t, files).Build(); Build() is already called once by Test)")
 	}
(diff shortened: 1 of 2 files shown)


hugolib/hugo_sites_build.go
```diff
@@ -61,7 +61,7 @@ import (
 // Build builds all sites. If filesystem events are provided,
 // this is considered to be a potential partial rebuild.
 func (h *HugoSites) Build(config BuildCfg, events ...fsnotify.Event) error {
-	if h.isRebuild() && !h.Conf.Watching() {
+	if h.isRebuild() && !h.Conf.Watching() && !h.Conf.Running() {
```
bep, line 64: So, it was me remembering wrong here, but I'm pretty sure h.Conf.Running() returns `true` in both cases here (meaning either watch=true or server is running), which would be nice to get right to avoid future confusion.

jmooring:
I'm struggling with this. In an integration test, using TestOptWatching() implies watching only, while TestOptRunning() implies both. That doesn't seem to match what happens when you do this from the command line.

bep:
OK, it Friday afternoon at 16:25, the sun shines outside, the tests for this passes, I'll merge this and leave this confusion for another date.

PR #14659, opened 2026-03-23
markup/rst: Add short syntax highlighting with fallback

### LINKED ISSUE #5349: reStructuredText uses long token names for syntax highlighting

xndvaz:
This PR fixes #5349 by passing `--syntax-highlight=short` to `rst2html`, so generated token classes use the short Pygments form.

Why this approach:
- Keeps the change scoped to `markup/rst` with minimal behavioral surface.
- Adds a compatibility fallback: if output is not parseable with `--syntax-highlight=short`, Hugo retries once without that flag.
- Uses a dedicated warning ID (`warning-rst-syntax-highlight-fallback`) to make fallback logging deduplicable and easier to track.

Tests:
- Argument assembly (Unix/Windows path shapes).
- Normal path with short syntax highlighting enabled.
- Compatibility fallback path when `--syntax-highlight=short` is not usable.
- Existing RST conversion test remains covered.

Validation:
- `go test ./markup/rst/...`
- `go test -race ./markup/rst/...`
- `./check.sh ./markup/rst/...`
- `./check.sh ./markup/...`
- `./check.sh ./...` (final rerun passed)

AI assistance disclosure:
- This PR was developed with AI assistance. Scope was limited to this bug fix, tests, and validation commands; all changes were reviewed before submission.


xndvaz:
Quick note on commit verification history: the earlier commit `b7e1b28` (shown as Unverified) was superseded by a force-push, and the current commit on this PR is `889275e` with a valid Verified SSH signature.

I updated this to keep the final PR state signed and verifiable.

xndvaz:
Follow-up note: this PR now includes a small test-only commit (`c6000c08f`) to fix current CI on `master` after deprecation removal of `_build`.

Specifically, `TestCascadeBuildOptionsTaxonomies` in `hugolib/cascade_test.go` used `cascade._build`, which now errors. The fixture was updated to `cascade.build` with no runtime behavior change for Hugo itself.

xndvaz:
Hi @bep, gentle FUP on this PR.
It is now up to date with `master` and all checks are green.  
Happy to adjust anything if needed. Thanks for your time!

markup/rst/convert.go
```diff
@@ -75,23 +80,20 @@ func (c *rstConverter) getRstContent(src []byte, ctx converter.DocumentContext)
...
-		result, err = internal.ExternallyRenderContent(c.cfg, ctx, src, binaryName, args)
-	}
+	result, err = c.renderContent(ctx, src, binaryName, binaryPath, true)
 
 	if err != nil {
 		return nil, err
 	}
 
+	if shouldRetryWithoutShortSyntaxHighlight(result) {
+		logger.Warnidf(constants.WarnRstSyntaxHighlightFallback, "%s did not return a parseable body with %s; retrying without it", binaryName, rst2ShortSyntaxHighlightArg)
```
bep: Is this something that can happen? Do you know of any realistic case?

xndvaz: Yes, this can happen with older docutils/rst2html versions that don’t support `--syntax-highlight=short`.

In that case, `rst2html` exits non-zero and writes an “unrecognized arguments” error, and we may end up without a parseable `<body>...</body>` output. The fallback retries once without `--syntax-highlight=short`, so rendering still succeeds.

I also updated `TestGetRstContentFallbackWithoutShortSyntaxHighlight` to simulate that realistic failure mode (stderr + exit code 2), instead of a silent empty-output case.

bep: Yea, OK. If Gemini is correct that flag was added in the 0.13 release (2016-12-20), almost 10 years ago. We could possibly detect that error and throw an error telling people that they need to upgrade.

xndvaz: Makes sense. I usually try to maximize backward compatibility, which is why I initially kept the fallback.

That said, I agree requiring an upgrade is the better trade-off here. I’ve updated this to detect unsupported `--syntax-highlight=short` and return a clear error:

`rst2html does not support --syntax-highlight=short; please upgrade Docutils (0.13+ required)`.

markup/rst/convert.go
```diff
@@ -75,23 +82,28 @@ func (c *rstConverter) getRstContent(src []byte, ctx converter.DocumentContext)
...
-		args := []string{"--leave-comments", "--initial-header-level=2"}
-		result, err = internal.ExternallyRenderContent(c.cfg, ctx, src, binaryName, args)
-	}
+	result, err = c.renderContent(ctx, src, binaryName, binaryPath, true)
 
 	if err != nil {
 		return nil, err
 	}
 
+	if shouldRetryWithoutShortSyntaxHighlight(result) {
```
bep: This entire bock looks both superfluous and too complex (including the warn statement).

I would expect `renderContent` above to return an non-nil err when passed an unknown flag.

If I do 

```
 rst2html --adfasdfasdf                         ✘ 2 main ✱
Usage
=====
  rst2html [options] [<source> [<destination>]]

rst2html: error: no such option: --adfasdfasdf
```


So, in `TestGetRstContentFallbackWithoutShortSyntaxHighlight`, which currently does not test anything real, I would rather see just make `rst2ShortSyntaxHighlightArg` into a var and temporarily change that to something unknown during test and verify that you get a sensible error return for the unknown flag.

xndvaz: Good call. I simplified this.

I removed the fallback/retry block (including the warn + extra detection helpers), made `rst2ShortSyntaxHighlightArg` a package var, and updated the test to temporarily set it to `--adfasdfasdf` and assert we return a non-nil error.

So this is now fail-fast: if the output has no parseable `<body>...</body>`, we return a clear error that the flag may be unsupported.

Qodo-Free-For-OSS:
Hi, The new body parsing treats missing `<body>`/`</body>` markers as a hard error, whereas the previous code returned a best-effort slice of the output. This can break environments where rst2html emits different body markers/formatting and will now fail rendering instead of returning HTML.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Retry or relax body parse

> ### Issue description
> `getRstContent` now errors if the rendered HTML doesn’t contain exact `<body>\n` and `\n</body>` markers. Older/variant rst2html outputs may differ slightly, causing unnecessary failures.
> 
> ### Issue Context
> If you add the intended fallback retry (short → long), the parse check should be used to trigger fallback rather than immediately failing. After fallback, you may still want to be lenient and return best-effort output as before.
> 
> ### Fix Focus Areas
> - markup/rst/convert.go[85-92]
> 
> ### What to change
> - Use the parse failure as the trigger for the fallback attempt.
> - After the fallback attempt, consider restoring prior best-effort behavior (e.g., default to full output or safe slicing) rather than hard erroring, unless a hard error is explicitly desired across all rst2html variants.

---
*Found by [Qodo](https://qodo.ai) code review*

bep:
@xndvaz I suggest that you start fresh with a new branch and a new PR. I don't have time to navigate this PR over the finish lime.

To fix this issue, we need to add `--syntax-highlight=short` and one or two **integration tests**. That should be a much smaller diff than this one.

xndvaz:
Opened a fresh minimal replacement PR as suggested: #14841.

This keeps the history here while giving the fix a clean review thread.

PR #14660, opened 2026-03-24
hugolib: Add regression test for render error content file path

### LINKED ISSUE #14607: Error 'render of "page" failed' does not mention content file
When upgrading to a new Hugo version, I get an error during build:

> render: failed to render pages: render of "page" failed: "/data/hugo-base/layouts/plugins/single.html:35:77": execute of template failed: template: plugins/single.html:35:77: executing "main" at <(index $.Site.Taxonomies.capabilities (anchorize .)).Page>: error calling Page: WeightedPages is empty

The problem is caused by a content file, but Hugo does not tell me which one. This really hinders debugging when you have 300+ content pages.

Please mention the name of the content file that is being processed.

(This is a split off from https://github.com/gohugoio/hugo/issues/12693#issuecomment-2263412558)

<pre>
$ hugo version
hugo v0.131.0+extended linux/amd64 BuildDate=2024-09-03T01:15:37Z VendorInfo=debian:0.131.0-1
</pre>

Yes.
Error message source code location is https://github.com/gohugoio/hugo/blob/13a95b9c0167c832fe2babf961ac5cad22e3a1e4/hugolib/site.go#L1735

majiayu000:
Fixes #14607

Adds `TestRenderErrorIncludesContentFile` to verify that template render errors include the content file path (e.g. `content/posts/p1.md`) rather than just the page kind.

### DIFF
--- a/hugolib/hugo_sites_build_errors_test.go
+++ b/hugolib/hugo_sites_build_errors_test.go
@@ -604,6 +604,29 @@ Home.
 	b.Assert(err.Error(), qt.Contains, `can't evaluate field ThisDoesNotExist`)
 }
 
+// https://github.com/gohugoio/hugo/issues/14607
+func TestRenderErrorIncludesContentFile(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+-- content/posts/p1.md --
+---
+title: "My Post"
+---
+
+Some content.
+-- layouts/page.html --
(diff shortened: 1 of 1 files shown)


cweiske:
@majiayu000 is an AI bot, and the PR does nothing sensible. It only adds a new test without modifying hugo's behavior.
Go away @majiayu000 and die.

PR #14665, opened 2026-03-24
Fix build with Go 1.26 on Fedora 44 Beta

flyn-org:
Fix a handfull of issues related to printf-style functions and format string verb mismatches.

### DIFF
--- a/commands/hugobuilder.go
+++ b/commands/hugobuilder.go
@@ -195,12 +195,12 @@ func (c *hugoBuilder) initMemProfile() {
 
 	f, err := os.Create(c.r.memprofile)
 	if err != nil {
-		c.r.logger.Errorf("could not create memory profile: ", err)
+		c.r.logger.Errorf("could not create memory profile: %s", err)
 	}
 	defer f.Close()
 	runtime.GC() // get up-to-date statistics
 	if err := pprof.WriteHeapProfile(f); err != nil {
-		c.r.logger.Errorf("could not write memory profile: ", err)
+		c.r.logger.Errorf("could not write memory profile: %s", err)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


bep:
While these fixes look fine, I would like to understand why this an issue on Fedora and not e.g. in the GitHub actions Linux runners? Where do these errors surface? Building/testing/go vet?

flyn-org:
Here is an error that occurs when I build the Fedora hugo package without this patch:
```
PASS
ok  	github.com/gohugoio/hugo/codegen	0.005s
github.com/gohugoio/hugo/commands
# github.com/gohugoio/hugo/commands
./hugobuilder.go:198:58: (github.com/gohugoio/hugo/common/loggers.Logger).Errorf call has arguments but no formatting directives
./hugobuilder.go:203:57: (github.com/gohugoio/hugo/common/loggers.Logger).Errorf call has arguments but no formatting directives
./hugobuilder.go:370:6: (github.com/gohugoio/hugo/common/loggers.Logger).Errorln call has possible Printf formatting directive %s
./hugobuilder.go:390:6: (github.com/gohugoio/hugo/common/loggers.Logger).Errorln call has possible Printf formatting directive %s
FAIL	github.com/gohugoio/hugo/commands [build failed]
```
Aside from that, I am not yet sure if there are any options involved that would be different than yours. I had initially thought this was simply because of the update to golang 1.26.1. That was the case with golang 1.24 and commit 4b0c194f.

bep:
OK, I don't understand this, but I guess they make sense.

jmooring:
Why are the `%q` => `%v` changes necessary?

flyn-org:
> Why are the `%q` => `%v` changes necessary?

I found the arguments that correspond to those format-string verbs to be slices, not integers.

Qodo-Free-For-OSS:
Hi, initMemProfile logs os.Create failure but continues and defers f.Close() even when f is nil, causing a panic when the deferred Close runs.

**Severity:** action required | **Category:** reliability

**How to fix:** Return on create failure

> ### Issue description
> `initMemProfile()` continues after `os.Create` error and defers `f.Close()` even when `f` is nil, which can panic.
> 
> ### Issue Context
> `initProfiling()` calls `initMemProfile()` in its stop/cleanup function, so this can crash during shutdown/cleanup when the memprofile file cannot be created.
> 
> ### Fix Focus Areas
> - commands/hugobuilder.go[191-205]
> - commands/hugobuilder.go[252-290]
> 
> ### Suggested change
> After logging the create error, return early (or guard `defer f.Close()` with `if f != nil`). Also consider returning after `WriteHeapProfile` failure if desired.

---

PR #14678, opened 2026-03-27
fix: prevent TOC rendering panics in Goldmark by adding recovery

QingGo:
The original issue is in https://github.com/gohugoio/hugo/issues/14677. I added panic recovery blocks in tocTransformer.Transform around the t.r.Render call to catch these out-of-range errors and fall back to rendering the node's raw text. And added a minimum test case to reproduce the bug.

### DIFF
--- a/hugolib/toc_panic_test.go
+++ b/hugolib/toc_panic_test.go
@@ -1,0 +1,31 @@
+package hugolib
+
+import (
+	"testing"
+)
+
+// TestTOCPanicMinimal reproduces the index out of range panic in the TOC transformer
+func TestTOCPanicMinimal(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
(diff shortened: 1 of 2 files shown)


bep:
Thanks, but recovering a panic won't make the problem go away (or: it hides it under the carpet). I think we need to do better.

QingGo:
> Thanks, but recovering a panic won't make the problem go away (or: it hides it under the carpet). I think we need to do better.

Hi @bep, fully agree with your judgement. My initial panic recovery fix was just a temporary band-aid to avoid build crashes, not a proper root-cause resolution.

The fix in #14742 is a much cleaner, more idiomatic solution that properly addresses the missing renderer registration for passthrough nodes in the TOC pipeline, which is the real root of the panic we identified.

Thanks for your guidance and for landing the proper fix. I'll close this PR now.

PR #14686, opened 2026-03-28
tpl/debug: add List function

### LINKED ISSUE #9148: Add debug.List

ogulcanaydogan:
Add `debug.List` that uses reflection to return a sorted string slice of:
- **Structs/pointers**: exported field names + method names
- **Maps**: keys (cast to string)

Non-recursive, one level deep.

```
{{ debug.List . }}       → ["BundleType","Content","Data","Date",...]
{{ debug.List $myMap }}  → ["key1","key2","key3"]
```

> This PR was written with assistance from Claude Code.

Fixes #9148

### DIFF
--- a/tpl/debug/debug.go
+++ b/tpl/debug/debug.go
@@ -16,6 +16,7 @@ package debug
 
 import (
 	"encoding/json"
+	"reflect"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


Qodo-Free-For-OSS:
Hi, debug.List dereferences pointers using v.Elem() without checking v.IsNil(), which panics when val is a typed nil pointer (e.g., a nil *T stored in an interface). This can crash template rendering/site builds when debug.List is called on optional pointer values.

**Severity:** action required | **Category:** reliability

**How to fix:** Guard Elem with IsNil

> ### Issue description
> `debug.List` panics when passed a typed-nil pointer because it calls `reflect.Value.Elem()` without checking `IsNil()`.
> 
> ### Issue Context
> This can occur in templates when a value is `(*T)(nil)` inside an `any`/interface, which is not `nil` at the interface level.
> 
> ### Fix Focus Areas
> - tpl/debug/debug.go[126-159]
> 
> ### Suggested change
> Update the pointer-unwrapping loop to handle nil pointers safely, e.g.:
> - While `v.Kind() == reflect.Pointer`:
>   - if `v.IsNil()` return `nil` (or an empty slice)
>   - else `v = v.Elem()`
> Optionally also unwrap interfaces similarly if you want `debug.List` to work when `val` is an interface wrapping a pointer/map/struct.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

bep:
Thanks for this, and I'm sorry for the delayed response. There are now 4 PRs implementing this. I have looked at all of them, and they mostly look fine on the basic level, but there are some issues there I'm guessing it's hard to handle if you're not very familiar with Hugo's code base. Instead of me picking one of the PR and try to get that over the finish line, I'm assigning the relevant issue to myself.

Issue #14576, opened 2026-02-28
Issue in rendering site when using multilingual website as subdir and uglyurl


mdugas:
### What version of Hugo are you using (`hugo version`)?
Tested from 0.116.0 until 0.157.0, confirmed appeared in version 0.154.4;

### Does this issue reproduce with the latest release?
yes

### Description

When using uglyurl with defaultContentLanguage and defaultContentLanguageInSubdir, the page render engine create the redirect to the set default language in `index.html` as well as `<defaultLanguage>/index.html` while other languages render properly. 

According to a test i've done, this started as of version 0.154.4. See this [pastebin](https://pastebin.com/f5bnLwqK) link for test results; you can see as of 0.154.4 (test starting at line 2457) that `index.html` and `en/index.html` are both the same md5 value vs `fr/index.html` has a different one. 

I've ran this batch of test with en as default language, but re-running using fr as the default language produce the same result where for that run `fr/index.html` is the same hash as `/index.html` while `/en/index.html` is rendered correctly.

### Relevant `hugo.yml` config file used in testing
``` yaml
baseURL: 'https://example.org/'
title: 'My New Hugo Project'
theme: testtheme

defaultContentLanguage: en
defaultContentLanguageInSubdir: true

languages:
  en:
    languageName: English
    weight: 1
  fr:
    languageName: Français
    weight: 2

uglyurls: true
disableKinds: ["RSS", "term", "taxonomy", "sitemap"]
```

### Content of `/index.html` and `/en/index.html`
``` html
<!doctype html>
<html lang="en">
  <head>
    <title>https://example.org/en/index.html</title>
    <link rel="canonical" href="https://example.org/en/index.html">
    <meta charset="utf-8" />
    <meta http-equiv="refresh" content="0; url=https://example.org/en/index.html">
  </head>
</html>
```

### Content of `/fr/index.html`
```
<!DOCTYPE html>
<html lang="fr" dir="ltr">
<head>
	<meta name="generator" content="Hugo 0.154.4">
  <meta charset="utf-8">
<meta name="viewport" content="width=device-width">
<title>My New Hugo Project</title>

      <link rel="stylesheet" href="/css/main.min.73be4799b2091d293b47b52ca2642fe82aa2b1c9c050e8d1b389a80939a30dd4.css" integrity="sha256-c75HmbIJHSk7R7UsomQv6CqiscnAUOjRs4moCTmjDdQ=" crossorigin="anonymous">


[...]

Issue #13346, opened 2025-02-04
Can't set relative aliases for home page


BrianLeishman:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.143.0-DEV-13b208e2f70fd0bfad8a4ae33a30afb5fd4b7477+extended linux/amd64 BuildDate=2025-01-31T08:29:10Z
</pre>

### Does this issue reproduce with the latest release?
Yes

---

Related issue about multi language sites and relative aliases here: https://github.com/gohugoio/hugo/issues/5757

I have a multi language site where I am trying to set the alias of a home page in the front matter like so:

```md
---
title: Welcome to our cool website!
layout: single

aliases:
  - welcome
---

# My cool website

```

And while this works page bundle pages correctly (it's relative to the "parent" directory of the `(_)index` file), this seems to be causing this bug on the root/site home page, throwing this error:

> render: alias "../welcome" traverses outside the website root directory

BrianLeishman:
Just reworked that test to match our setup:

https://github.com/BrianLeishman/hugo-testing/tree/hugo-github-issue-13346

I'm using the content mounts to handle "default" files for the other language, but I feel like the `defaultContentLanguageInSubdir` is more likely the culprit, as we need that off

jmooring:
I can reproduce the behavior by setting `defaultContentLanguageInSubdir = false`. 

For now, set your aliases like this:

```text
aliases = ['/en/welcome']
aliases = ['/de/willkommen']
```

jmooring:
This issue is not specific to multilingual sites. You cannot a create an alias for the home page with a relative path.

bep:
>this is a bug or an ennhacement.

Smells like a bug to me, but I haven't thought about aliasing the home page being useful.

BrianLeishman:
Use case to us is that this is a new website replacing a very old version, huge site overhaul. But the old site had a bunch of landing pages that didn't make sense anymore, and our home page is kind of the starting point to browsing our site in the new user experience.

So all those old links now route to home.

jmooring:
Blocked by https://github.com/gohugoio/hugo/issues/14576.

PR #14689, opened 2026-03-29
fix: multilingual URL rendering with uglyURLs

xingzihai:
This PR fixes multilingual URL rendering when uglyURLs is enabled.\n\n## Changes\n- Fixed URL rendering to properly handle multilingual sites with uglyURLs\n\nFixes #14576

### DIFF
--- a/hugolib/site_render.go
+++ b/hugolib/site_render.go
@@ -366,7 +366,19 @@ func (s *Site) renderDefaultSiteRedirect(home *pageState) error {
 		//    /,/guest/ =  /guest/v1.0.0/
 		parts := strings.Split(strings.Trim(homeLink, "/"), "/")
 
-		for i := 0; i < len(parts)-1; i++ {
+		// With uglyURLs enabled for home pages, homeLink ends with an index file
+		// (e.g., "/en/index.html"). We need to skip creating redirects for the
+		// directory that contains this index file, as the redirect path would
+		// resolve to the same file (via targetPathAlias), overwriting the home page.
+		// See https://github.com/gohugoio/hugo/issues/14576
+		lastPartIsIndex := len(parts) > 0 && strings.HasPrefix(parts[len(parts)-1], of.BaseName+".")
+
+		skipCount := 0
+		if lastPartIsIndex {
+			skipCount = 1
+		}
+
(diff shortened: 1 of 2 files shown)


jmooring:
This PR does not address this comment:
<https://github.com/gohugoio/hugo/issues/14576#issuecomment-4016949005>

If we decide on a partial fix to close #14576, we must create a new issue to cover the multihost scenario.

I also think that we may want to address #13346 concurrently.

Qodo-Free-For-OSS:
Hi, renderDefaultSiteRedirect skips creating a redirect for the directory containing the home output whenever the homeLink filename starts with of.BaseName + ".". Alias publishing always writes redirects to index.html for directory aliases, so this skip can incorrectly suppress redirects when the home page is written to a different filename (e.g., index.b via custom HTML suffix), where no overwrite/collision would occur.

**Severity:** remediation recommended | **Category:** correctness

**How to fix:** Skip only on real collision

> ### Issue description
> `renderDefaultSiteRedirect` currently decides whether to skip generating the alias for the directory containing the home output by checking if the last path segment of `homeLink` looks like an index file (`HasPrefix(last, of.BaseName + ".")`). This is only a proxy for the real problem (alias target overwriting the home page) and can be wrong when the home page’s output filename is not `index.html`.
> 
> ### Issue Context
> - Alias publishing resolves directory aliases to `.../index.html` via `targetPathAlias`, independent of output format suffix/baseName.
> - Home output filenames can vary with custom media type suffixes / output format configuration.
> 
> ### Fix Focus Areas
> - hugolib/site_render.go[369-383]
> - hugolib/alias.go[123-187]
> 
> ### Suggested approach
> Instead of checking the *link* string, compute whether the would-be alias publish path would resolve to the same target file as the home page:
> 1. Compute the home page target filename (e.g., `home.pageOutput.targetPaths().TargetFilename`).
> 2. Compute the alias target filename for the directory alias you are about to add (use the same logic as publishing, e.g., call/replicate `targetPathAlias` on the alias path).
> 3. Only skip the last directory alias when `aliasTargetFilename == homeTargetFilename`.
> This keeps the current fix for `/en/index.html` overwrites while not disabling redirects unnecessarily for custom suffix/baseName setups.

---

PR #14690, opened 2026-03-29
hugolib: Fix error position for nested shortcode extraction failures

### LINKED ISSUE #11302: The "failed to extract" error refers to outer shortcode instead of inner shortcode

<pre>
$ hugo version
hugo v0.111.1-39a4a3cf676533859217805c36181c7a3dfa321c darwin/arm64 BuildDate=2023-03-02T10:08:26Z VendorInfo=gohugoio
</pre>

yes

When evaluating this file:

```md
+++
Type = "note"
Date = "2020-03-30T01:19:59-05:00"

+++

{{< figure >}}
{{< img src="lighting-in-pantry.jpg" alt="a pantry with led lights shining bright light out of it" >}}
{{% figcaption %}}
Snazzy 💡
{{% /figcaption %}}
{{< /figure >}}
```

i get this error:

```
Start building sites … 
hugo v0.111.1-39a4a3cf676533859217805c36181c7a3dfa321c darwin/arm64 BuildDate=2023-03-02T10:08:26Z VendorInfo=gohugoio
Error: Error building site: "/Users/calebjasik/Git/jasik-xyz/content/2020-03-30-0119/index.md:7:1": failed to extract shortcode: unclosed shortcode "figure"
Total in 88 ms
```

Before v0.111.0, this compiled fine, but I'm not sure the "is closed" detector is effective if there is additional shortcodes in the `.Inner` of the shortcode?

xingzihai:
Fixes #11302

## Problem
When nested shortcode extraction fails, the error position was not correctly reported, making debugging difficult.

## Solution
Fixed the error position reporting to correctly identify the location of the failing shortcode in nested scenarios.

## Testing
- Added tests for nested shortcode error position reporting
- Verified error messages now show correct line numbers

### DIFF
--- a/hugolib/shortcode.go
+++ b/hugolib/shortcode.go
@@ -612,6 +612,7 @@ Loop:
 			if cnt > 0 {
 				// nested shortcode; append it to inner content
 				pt.Backup()
+				nestedPos := currItem.Pos()
 				nested, err := s.extractShortcode(nestedOrdinal, nextLevel, source, pt)
 				nestedOrdinal++
 				if nested != nil && nested.name != "" {
@@ -621,7 +622,14 @@ Loop:
 				if err == nil {
 					sc.inner = append(sc.inner, nested)
 				} else {
-					return sc, err
+					// Wrap the error with the nested shortcode's position so that
+					// the error points to the correct shortcode (the inner one)
+					// rather than the outer shortcode.
+					if fe, ok := err.(herrors.FileError); ok {
+						return sc, fe
+					}
+					pos := posFromInput("", source, nestedPos)
+					return sc, herrors.NewFileErrorFromPos(err, pos)
 				}
 
 			} else {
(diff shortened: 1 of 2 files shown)


hugolib/shortcode.go
```diff
@@ -612,6 +612,7 @@ Loop:
 			if cnt > 0 {
 				// nested shortcode; append it to inner content
 				pt.Backup()
+				nestedPos := currItem.Pos()
```
bep, line 615: Can you move this assignment down to the error branch below. I'm pretty sure that's not the case here, but we have some places where `Pos()` is relatively expensive to calculate, and moving it avoids me (and future mes) to have to look.

Other than that, this all looks good.

bep:
Can you also sign the CLA? If that doesn't somehow work, you need to make sure the email used in the commit matches one in your GitHub account.

Issue #14711, opened 2026-04-03
Default else block of with causes stray quotes to be injected in <script> tags


asdofindia:
Hi, I was debugging an issue with geekblog theme https://github.com/thegeeklab/hugo-geekblog/issues/827 and came across what's potentially an issue with hugo itself? The following bug report is generated by Gemini, but after I made it discover the exact issue in [this conversation](https://gemini.google.com/share/1d1e6a066346). I've verified the content of this bug report and I stand by it except of course I do not know if this is intended behavior. I'll create a PR with a suggested fix in case this is unintended behavior.

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.159.2 linux/amd64 BuildDate=unknown
</pre>

### Does this issue reproduce with the latest release?

Yes. Reproduced with `CGO_ENABLED=0 go install github.com/gohugoio/hugo@latest`.

### Description

When using `{{ with partial "name" . }}` inside a `<script>` tag (such as when building JSON-LD schema), if the partial evaluates to an empty string, Hugo renders a literal pair of double-quotes (`""`) instead of rendering nothing.

This happens because Hugo's internal AST transformer (`tplimpl`) secretly injects an `else` block into `with partial` statements, which collides with Go's `html/template` context-aware JavaScript escaper.

#### The Bug Mechanism

In `tplimpl/template.go`, the function `handleWithPartial` intercepts `with partial` calls to manage the partial decorator stack. It automatically appends a fallback `else` branch that calls `_popPartialDecorator`:

```go
// From tplimpl/template.go -> handleWithPartial()
// When the partial returns a falsy value, the with block is skipped,
// but we still need to pop the decorator ID from the stack.
// Add a pop call to the else branch.
popElse := popPartialDecoratorElse.CopyList()
// ...
if withNode.ElseList == nil {
    withNode.ElseList = popElse // <-- INJECTED ELSE BLOCK
}
```

The function that this `else` block executes is defined in the `init()` function, and it returns a primitive Go `string`:

```go
"_popPartialDecorator": func(string) string { return "" }
```

**The Collision:**

1.  We call `{{ with partial "empty" . }}` inside a `<script>` tag.
2.  The partial returns `""`. The `with` condition evaluates as falsy.
3.  The engine falls back to the injected `else` block, executing `_popPartialDecorator`, which yields the primitive string `""`.
[...]

PR #14713, opened 2026-04-03
tpl: prevent falsy with partial rendering quotes in script context

asdofindia:
This commit was assisted by Google Gemini. [Conversation link](https://gemini.google.com/share/cd34cb3959d6).

More details in #14711

I do know that the test looks right and the fix works with the test. I'm not sure about whether there are effects farther in the codebase by switching this to a template.JS string

### DIFF
--- a/tpl/templates/templates.go
+++ b/tpl/templates/templates.go
@@ -17,6 +17,7 @@ package templates
 import (
 	"context"
 	"fmt"
+	"html/template"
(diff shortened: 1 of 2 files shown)


asdofindia:
Oh sorry. I was coding and didn't see this was fixed in https://github.com/gohugoio/hugo/pull/14712/

Issue #12993, opened 2024-10-29
Some URL-safe special characters cannot be used in Page path


trwnh:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.136.5+extended linux/amd64 BuildDate=unknown
</pre>

### Does this issue reproduce with the latest release?

yes

### Steps to reproduce

1. Add a page (e.g. using a content adapter) whose `path` includes an exclamation mark (e.g. `yu-gi-oh!`)

### Expected behavior

The page is created with the url including the exclamation mark (e.g. `/tags/yu-gi-oh!`)

### Actual behavior

The exclamation mark seems to be sanitized out (e.g. `/tags/yu-gi-oh`, creating a path conflict with an already existing page)

### Additional information

https://gohugo.io/methods/page/path/ describes the following:

> To determine the logical path for pages backed by a file, Hugo starts with the file path, relative to the content directory, and then:
>
> - Strips the file extension
> - Strips the language identifier
> - Converts the result to lower case
> - Replaces spaces with hyphens
>
> The value returned by the Path method on a Page object is independent of content format, language, and URL modifiers such as the slug and url front matter fields.

Nowhere in these 4 steps does it say anything about removing URL-safe characters entirely. This issue seems to occur for some URL-safe characters, but not all. These are the ones that work:

- `content/foo-bar.md` => `/foo-bar/` (correct).
- `content/foo_bar.md` => `/foo_bar/` (correct).
- `content/foo.bar.md` => `/foo.bar/` (correct).
- `content/foo+bar.md` => `/foo+bar/` (correct).
- `content/foo@bar.md` => `/foo@bar/` (correct).
- `content/foo~bar.md` => `/foo~bar/` (correct).

And these are the ones that don't work:

- `content/foo$bar.md` => `/foobar/` (incorrect). `/foo$bar/` leads to a Page Not Found.
- `content/foo!bar.md` => `/foobar/` (incorrect). `/foo!bar/` leads to a Page Not Found.
- `content/foo*bar.md` => `/foobar/` (incorrect). `/foo*bar/` leads to a Page Not Found.
- `content/foo'bar.md` => `/foobar/` (incorrect). `/foo'bar/` leads to a Page Not Found.
- `content/foo(bar.md` => `/foobar/` (incorrect). `/foo(bar/` leads to a Page Not Found.
- `content/foo)bar.md` => `/foobar/` (incorrect). `/foo)bar/` leads to a Page Not Found.
[...]

PR #14723, opened 2026-04-06
fix(paths): allow '=' character in path validation

tomzx:
# What

- Allow the `=` character in Hugo page paths and URLs.

# Why

- `=` is a valid URL path character per RFC 3986 (sub-delimiters set).
- Hugo was silently dropping `=` from content file names/slugs, causing path conflicts and broken URLs (e.g. `foo=bar.md` would produce `/foobar/` instead of `/foo=bar/`).

# References

- https://github.com/gohugoio/hugo/issues/12993

### DIFF
--- a/common/paths/path.go
+++ b/common/paths/path.go
@@ -329,7 +329,7 @@ func isAllowedPathCharacter(s string, i int, r rune) bool {
 	}
 	// Check for the most likely first (faster).
 	isAllowed := unicode.IsLetter(r) || unicode.IsDigit(r)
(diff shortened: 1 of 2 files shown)


bep:
The issue you refer to, #12993, is a general issue about this topic. Also, it doesn't mention `=`. I'm not merging this issue to make one person happy and potentially lots of other people angry because we broke they're URLs. This has been stable Hugo behaviour for a decade or so, so if we're changing it, we're doing all changes at once, based on some consensus in an issue.

Issue #14325, opened 2025-12-30
:contentbasename and :slugcontentbasename broken for taxonomies since v0.153


hacdias:
v0.153.0 has changed the behavior of the tokens `:contentbasename` and  `:slugcontentbasename` as resolution of #14104. Sorry that I missed the ping on that issue, otherwise we could've caught this before.

The new behavior breaks my website, and it seems to be against what's documented and what has been discussed and agreed upon before on issue #11722. It was known that it'd be different than `:filename` and give the directory name for bundles with `_index.md` pages. It's also documented on https://gohugo.io/methods/page/file/#contentbasename (linked from https://gohugo.io/configuration/permalinks/#contentbasename):

> **ContentBaseName**
> (string) If the page is a branch or leaf bundle, the name of the containing directory, else the TranslationBaseName.

For testing, you can just use my website. Here's the relevant taxonomy configuration: https://github.com/hacdias/hacdias.com/blob/9e91c45e7dda87979f3e0e1e7c62a32056d1a4f8/config.toml#L17-L21

All "categories" build to "/" now and all "tags" build to "/tags":

```
WARN  Duplicate target paths: /feed.json (3), /feed.xml (3), /index.html (3), /page/10/index.html (2), /page/11/index.html (2), /page/12/index.html (2), /page/13/index.html (2), /page/14/index.html (2), /page/15/index.html (2), /page/16/index.html (2), /page/17/index.html (2), /page/18/index.html (2), /page/19/index.html (2), /page/2/index.html (2), /page/20/index.html (2), /page/21/index.html (2), /page/22/index.html (2), /page/23/index.html (2), /page/24/index.html (2), /page/3/index.html (2), /page/4/index.html (2), /page/5/index.html (2), /page/6/index.html (2), /page/7/index.html (2), /page/8/index.html (2), /page/9/index.html (2), /styles.css (2), /tags/feed.json (27), /tags/feed.xml (27), /tags/index.html (28), page/1/index.html (2), tags/page/1/index.html (26)
```

So I'm not really sure why the change, and the only alternative seems to manually set a slug for all of these taxonomies, which is time consuming.

I'd argue that #14126 should be reverted since it doesn't match the documentation and the decision in #11722.

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.153.4+extended+withdeploy darwin/arm64 BuildDate=2025-12-28T21:34:16Z VendorInfo=Homebrew
</pre>

[...]

PR #14126, opened 2025-11-10
resources/page: Fix slugorcontentbasename for section pages

dvdksn:
## Summary

This PR fixes the `slugorcontentbasename` permalink token creating an extra subdirectory for section pages (`_index.md` files).

## Problem

When using `slugorcontentbasename` in permalinks for section pages, an extra subdirectory was created. For example:

```yaml
permalinks:
  section:
    foo: /:sections[1:]/:slugorcontentbasename/
```

With content at `content/foo/bar/_index.md`, this would generate `/bar/bar/index.html` instead of the expected `/bar/index.html`.

The deprecated `slugorfilename` token worked correctly by returning an empty string for `_index.md` files.

## Solution

Added special handling in `pageToPermalinkContentBaseName` to return an empty string for section pages (`_index.md` files), matching the behavior of `pageToPermalinkFilename`.

## Testing

- Added integration test `TestIssue14104` to verify the fix
- All existing tests pass
- Tested with the reproduction repository from the issue

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Fixes #14104

Issue #14104, opened 2025-10-30
:slugorcontentbasename creates subdirectory for section page


dvdksn:
I've been using the `slugorfilename` permalink token to strip out the first section, like so:

```
permalinks:
  page:
    foo: /:sections[1:]/:slugorfilename/
  section:
    foo: /:sections[1:]/:slugorfilename/
```

After getting warnings that this is deprecated, and I should use `slugorcontentbasename` instead, I tried it. But this seems to generate an additional subdirectory for section pages where the `slugorfilename` did not:

```
permalinks:
  page:
    foo: /:sections[1:]/:slugorcontentbasename/
  section:
    foo: /:sections[1:]/:slugorcontentbasename/
```

I.e., given content pages like this:

```console
$ tree content/
content/
└── foo
    ├── _index.md
    └── bar
        ├── _index.md
        └── somepage.md

3 directories, 3 files
```

The html output for `content/bar/_index.md` is generated at `public/bar/bar/html`:

```console
$ tree public/
public/
├── bar
│   ├── bar
│   │   └── index.html
│   └── somepage
│       └── index.html
└── foo
    └── index.html

5 directories, 3 files
```

See repro: https://github.com/dvdksn/contentbasename

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
v0.152.2
</pre>

### Does this issue reproduce with the latest release?
yes

jmooring:
Re-opening because we reverted #14126.

jmooring:
Failing integration test:

`resources/page/permalinks_integration_test.go`

```go
// Issue 14104
func TestIssue14104(t *testing.T) {
	t.Parallel()

	files := `
-- hugo.toml --
disableKinds = ['rss','sitemap','taxonomy','term']

[permalinks.page]
s1 = "/:sections[1:]/:slugorcontentbasename/"

[permalinks.section]
s1 = "/:sections[1:]/:slugorcontentbasename/"

-- content/s1/s2/_index.md --
---
title: s2
---
-- content/s1/s2/p1.md --
---
title: p1
---
-- layouts/all.html --
{{ .Title }}|
`

	b := hugolib.Test(t, files)

	b.AssertFileExists("public/index.html", true)
	b.AssertFileExists("public/s2/index.html", true) // fail
	b.AssertFileExists("public/s2/p1/index.html", true)

	b.AssertFileExists("public/s1/index.html", false) // fail
[...]

PR #14767, opened 2026-04-17
resources/page: Add passing test for Issue #14325

jmooring:
This increases test coverage by adding an integration test that covers the configuration and site structure as described in #14325. This is designed to prevent regressions similar to the breakage seen in #14126 during the fix for #14104.

I used Gemini, sort of, to refine the test.

### DIFF
--- a/resources/page/permalinks_integration_test.go
+++ b/resources/page/permalinks_integration_test.go
@@ -14,6 +14,7 @@
 package page_test
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


resources/page/permalinks_integration_test.go
```diff
@@ -435,3 +436,82 @@ title: aBc
 	b = hugolib.Test(t, files)
 	b.AssertFileExists("public/aBc/index.html", true)
 }
+
+func TestIssue14325(t *testing.T) {
```
bep: I used to write test with names like that, but then looking at them 2 year later, it gets annoying to have to look the issue to know what it's about. So now I prefer func names on the form "TestEmptyTitleCreatesHavoceIssue1234" so I at least get some gist about it.

jmooring: Sorry, I forgot that you mentioned that before. Will fix.

bep: I'm not sure I mentioned it before, but I still think it's a good idae The test itself looks solid and a good idea. though.

PR #14794, opened 2026-04-22
langs: Use Language.Locale as primary localization key

### LINKED ISSUE #9109: Localization: allow to match locale with an additional parameter instead of the language key.
Hugo supports localize time and dates since v0.87.0, but I'm stuck by specifying the language key that not defined in https://github.com/gohugoio/locales.

For example:

```toml
[zh_hans] # localization works as expected.
  languageName = "zh_hans"

[zh-cn]
  languageName = "zh-cn"
  locale = "zh_hans" # additional parameter to match the locale
 ```

- https://discourse.gohugo.io/t/is-it-possible-to-make-language-key-to-be-differenced-with-locale-name/35386

jmooring:
Localization now uses `Language.Locale `as the golocales lookup key, falling back to `Language.Lang`, then `defaultContentLanguage`, then `en`.

Close #9109

### DIFF
--- a/langs/language.go
+++ b/langs/language.go
@@ -79,11 +79,13 @@ func (l *Language) IsDefault() bool {
 
 // NewLanguage creates a new language.
 func NewLanguage(lang, defaultContentLanguage, timeZone string, languageConfig LanguageConfig, logger loggers.Logger) (*Language, error) {
-	translator := golocales.New(lang)
-	if translator == nil {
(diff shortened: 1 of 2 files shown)


jmooring:
This is a follow-on to https://github.com/gohugoio/hugo/pull/14225, where we now use `Language.Locale` as the primary key for string translations.

PR #14841, opened 2026-05-04
markup/rst: Make syntax highlighting configurable

### LINKED ISSUE #5349: reStructuredText uses long token names for syntax highlighting

xndvaz:
Fixes #5349
Supersedes #14659

## Summary
- add `markup.rst.syntaxHighlight` with Docutils' CLI default (`long`)
- pass `--syntax-highlight` only when configured as `short` or `none`
- document the RST markup configuration and default
- cover the real `rst2html` render path without requiring Pygments

## Notes
- Default behavior remains Docutils' `long` syntax classes.
- The integration test allow-list now matches the command Hugo actually executes: `rst2html`/`rst2html.py` on Unix-like systems, `python`/`python.exe` on Windows.

## Tests
- `go test ./hugolib -run TestRSTSyntaxHighlightConfigIssue5349 -count=1`
- `./check.sh ./markup/rst/...`
- `./check.sh ./markup/markup_config/...`
- `./check.sh ./hugolib/...`

AI assistance disclosure:
- Developed with AI assistance; I reviewed the final diff, scope, and validation before submission.


Qodo-Free-For-OSS:
Hi, The new integration test uses `security.exec.allow = ['rst2html']`, but allow-list entries are regex patterns; an unanchored pattern can match unintended command names containing `rst2html`.

**Severity:** remediation recommended | **Category:** security

**How to fix:** Anchor allow-list pattern

> ### Issue description
> The test config’s `security.exec.allow = ['rst2html']` is interpreted as a regexp, which can match more than the intended binary name.
> 
> ### Issue Context
> `Whitelist.Accept` compiles the patterns as regexps and uses `MatchString(name)`.
> 
> ### Fix Focus Areas
> - hugolib/rst_integration_test.go[51-57]
> 
> ### Fix outline
> - Change `allow = ['rst2html']` to `allow = ['^rst2html$']` (or `'^rst2html(\\.py)?$'` if you want to allow both explicitly).

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*Found by [Qodo](https://qodo.ai) code review*

xndvaz:
Addressed in 22984f4: updated test config to use anchored allow-list regex (^rst2html$).

hugolib/rst_integration_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func TestRSTSyntaxHighlightShortIssue5349(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("needs a POSIX shell script")
+	}
+
+	c := qt.New(t)
+	dir := t.TempDir()
+	rst2html := filepath.Join(dir, "rst2html")
```
bep: I'm having trouble understanding this, but to me it does not make sense to have an integration test that "simulates" rst. We make sure that this works with the REAL rst as installed as part of the CI.

xndvaz:
Agreed, thanks @bep. The previous test used a fake rst2html only to verify the added flag, but I agree that this is better covered with the real executable.

Updated the test to use the real rst2html path installed in CI and added Pygments to the CI setup so the syntax-highlight output is actually exercised.

xndvaz:
Follow-up CI note: the current failure is unrelated to this RST PR.

It is the existing TestNPMGlobalInstalls failure already present on master, caused by an exact Tailwind CSS byte-size assertion changing from 4557 to 4479. I opened #14860 separately to fix that CI fragility, and its checks are green.

Keeping this PR scoped to the RST change.

.github/workflows/test.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         run: go install github.com/magefile/mage@v1.15.0
       - name: Install docutils
         run: |
-          pip install docutils
+          pip install docutils Pygments
```
bep, line 66: I will have to get back to this later, but if this patch requires `pygments` (I suspect it's lowercase?), I think we need to reconsider doing this as the new default (as many may not want it) ... We could add a config struct for rst similar to what we have for the other markup engines with a sensible option with a comment that you would need to install `pygments` for it to work. I'm not totally sure about this, though.

xndvaz: Good point, thanks.

I changed this to avoid making `--syntax-highlight=short` the new default. The PR now adds an explicit `markup.rst.syntaxHighlight` setting instead, with Docutils' default `long` preserved.

The CI setup also no longer installs `pygments`; the integration test uses `syntaxHighlight = "none"` so it verifies the Hugo/rst2html configuration path without adding a new CI dependency. The docs mention that Docutils needs Pygments available when users want highlighted code blocks.

So the short class names are now opt-in via:

```toml
[markup.rst]
syntaxHighlight = "short"

Comment (jmooring) [reply .github/workflows/test.yml:66]: @bep Highlighted code classes require the Pygments library. This is an existing requirement (not related to this PR) if you want to perform syntax highlighting, but since `pip install docutils` doesn’t include it by default (unlike the `python3-docutils` apt package), our CI currently lacks it. Without Pygments in CI, we can't run affirmative integration tests for this PR.

File: hugolib/rst_integration_test.go [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+	if !rst.Supports() {
+		t.Skip("rst not installed")
+	}
+
+	files := `
+-- hugo.toml --
+baseURL = "https://example.org"
+disableKinds = ["home", "section", "taxonomy", "term", "rss", "sitemap"]
+[security.exec]
+allow = ['^python(\.exe)?$', '^rst2html(\.py)?$']
```
jmooring: Why include python? It's not called.

xndvaz: Good catch. I tightened the test allow-list.

On Unix-like systems Hugo executes `rst2html`/`rst2html.py` directly, so the test now only allows that there. On Windows the RST renderer invokes `python` with the script path, so the test allows `python` only on Windows.

The updated CI is green on both Ubuntu and Windows.

jmooring:
Although the configuration key `syntaxHighlight` matches the command line flag, that name doesn't tell me what it does when looking at a Hugo config file. To be consistent with other highlight options, I suggest this instead:

```toml
[markup.rst.highlight]
classNaming = 'long'
```

With this description:

> The naming convention for CSS classes generated during syntax highlighting, one of `long`, `short`, or `none`. Default is `long`.

xndvaz:
@bep were you able to verify this PR? Do you think it needs any adjustments?

jmooring:
See my previous comments. This should not be merged.

xndvaz:
Thanks @jmooring, I followed your recommendation and changed this to use:

```toml
[markup.rst.highlight]
classNaming = "long"
```

The docs, default config, and tests now use that shape, while keeping the default Docutils behavior unchanged and without adding a Pygments dependency.

jmooring:
As I mentioned [earlier](https://github.com/gohugoio/hugo/pull/14841#discussion_r3243929840), I think we should be testing that this change actually does what it is supposed to do. But to do that we need Pygments installed during the CI setup. But that's bep's call.

xndvaz:
Thanks, agreed. I added Pygments back to the CI setup as a test-only dependency and changed the RST integration test to exercise the affirmative `classNaming = "short"` path.

The test now verifies that Docutils emits the short Pygments class name (`k`) and not the long one (`keyword`). It still also covers `classNaming = "none"`, so the no-Pygments/no-highlighting path remains explicit.

If @bep would rather avoid installing Pygments in CI, I can adjust, but this now tests the behavior this PR is intended to expose.

bep:
I have looked at this, but this needs to wait:

* I understand that people who use RST would want this.
* But this is a little too many building blocks to add a flag to a feature that I don't use (but have to maintain).
* Most things looks fine, but I see some obvious things that needs to be improved before merging this, and since I suspect I'm mostly talking to Claude or similar here, I need to decide if that is something I would like to do (I find it considerably easier to motivate myself to review Claude code that's initiated by myself compared to others').

xndvaz:
Thanks for taking the time to look, @bep.

I understand the maintenance concern, especially for a feature you don't use yourself and with the extra moving parts around Docutils/Pygments/CI.

Just to clarify the process: I started with one approach, then adjusted the configuration shape and the test coverage based on @jmooring's feedback. I do use AI assistance, but I'm not trying to outsource the conversation or the responsibility for the patch. I'm reading the feedback, reviewing the code and trying to learn the project's expectations so I can make the best contribution I can.

If you are willing, could you point me to the obvious things you think should be improved? I'm happy to simplify the PR, reduce the number of moving parts or take a different direction if that makes this easier to maintain.

bep:
@xndvaz I will get back to this, but my general concern is this: Me doing code reviews for human contribs has an educational effect (as in: they will eventually do better etc.). With AI contributions., that is not the case. I could have Claude fix #5349 in much less time than it would take me to review this PR. And that is just a part of the big picture.

xndvaz:
I understand your point, although I do see value in fixing a real user-facing gap like this even if the implementation was AI-assisted.

That said, I also understand that review and maintenance cost is your call as a maintainer, and I don't wanna push extra burden onto you. Thanks for taking the time to explain your position.

PR #14843, opened 2026-05-05
markup/highlight: fix HighlightCodeBlock.Inner for code blocks without language

### LINKED ISSUE #14820: `transform.HighlightCodeBlock` returns `.Inner` with wrapping tags for code blocks without lang
documentation states that `.Inner`

> Returns highlighted code without any wrapping elements, allowing you to create your own wrapper.

but that is not the case for codeblocks without any `.Type`.

reproducer:
`````shell
; hugo new project meow
# a bunch of output
; cd meow
; echo '{{ .Content }}' > layouts/home.html
; mkdir layouts/_markup
; cat >layouts/_markup/render-codeblock.html <<'EOF'
{{ $result := transform.HighlightCodeBlock . }}
<pre><code>{{ $result.Inner }}</code></pre>
EOF
; cat >content/_index.md <<'EOF'
```
meow
```
EOF
; hugo build
; cat public/index.html

<pre><code><pre tabindex="0"><code>meow</code></pre></code></pre>

`````

<pre>
$ hugo version
hugo v0.160.1+extended+withdeploy linux/amd64 BuildDate=unknown VendorInfo=nixpkgs
</pre>

yes

mitre88:
When a code block has no language specified and no lexer can be found, the preWrapper was being created and used to write content, but its low/high positions were not being returned. This caused .Inner() to return the full highlighted output including wrapping tags.

Now when lexer is nil, we track the preWrapper and return its low/high positions so that .Inner() correctly extracts just the code content.

Fixes #14820

### DIFF
--- a/markup/highlight/highlight.go
+++ b/markup/highlight/highlight.go
@@ -179,13 +179,17 @@ func highlight(fw hugio.FlexiWriter, code, lang string, attributes []attributes.
 	w := &byteCountFlexiWriter{delegate: fw}
 
 	if lexer == nil {
+		var pw *preWrapper
 		if cfg.Hl_inline {
 			fmt.Fprintf(w, "<code%s>%s</code>", inlineCodeAttrs(lang), gohtml.EscapeString(code))
 		} else {
-			preWrapper := getPreWrapper(lang, w)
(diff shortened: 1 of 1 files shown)


bep:
Thanks, but this is already fixed in #14822

PR #14861, opened 2026-05-10
parser: Drop empty sub maps from `hugo config` output

### LINKED ISSUE #14855: In "hugo config", don't print empty sub maps
E.g.

```toml
[permalinks]
```

```toml
[[permalinks]]
  [permalinks.target]
    path = '/{photo,photo/**}'

    [permalinks.target.sites]

```

I would expect/prefer this:

```toml
[[permalinks]]
  [permalinks.target]
    path = '/{photo,photo/**}'
```

ChrisJr404:
Took a stab at this. Happy to drop the PR if you've already got something underway.

`ReplacingJSONMarshaller.MarshalJSON` walks the unmarshalled map and deletes zero-valued leaves when `OmitEmpty` is true, but it skips the truthiness check for any value that is itself a map. So once recursion empties a child map, the now-empty parent map sticks around in the output. With the issue's input, that surfaces as:

```toml
[[permalinks]]
  [permalinks.target]
    path = '/{photo,photo/**}'

    [permalinks.target.sites]

```

`sitesmatrix.Sites.IsZero` is defined and would return true here, but at this point in the pipeline everything is already a `map[string]any`, so the struct-level `IsZero` is gone.

The change after recursing into a child map, drop the key if the child is now empty. Slices of maps are deliberately left alone since removing slice entries would shift indices and the existing behavior never did that.

Verified locally with the issue's `hugo.toml`. Output for `permalinks`, `privacy`, etc. is now collapsed to just the populated fields, in `--format toml`, `--format yaml`, and `--format json`. `--printZero` is unchanged (that path doesn't enter the `OmitEmpty` block).

Added a regression test in `parser/lowercase_camel_json_test.go` covering:
- a nested chain of maps that should fully prune away
- a `target.sites.{matrix,complements}` shape with empty slice fields
- a slice-of-maps where the entry has an empty sub map but a populated sibling key (verifies the slice path still strips the sub map but keeps the entry)

Fixes #14855

### DIFF
--- a/parser/lowercase_camel_json.go
+++ b/parser/lowercase_camel_json.go
@@ -116,6 +116,9 @@ func (c ReplacingJSONMarshaller) MarshalJSON() ([]byte, error) {
 					switch vv := v.(type) {
 					case map[string]any:
 						removeZeroVAlues(vv)
+						if len(vv) == 0 {
+							delete(m, k)
+						}
 					case []any:
 						for _, vvv := range vv {
 							if m, ok := vvv.(map[string]any); ok {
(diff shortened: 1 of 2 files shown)


bep:

Issue #13492, opened 2025-03-13
panic: deferred execution with id "__hdeferred/...=" not found


germanftorres:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.145.0+extended+withdeploy linux/amd64 BuildDate=2025-02-26T15:41:25Z VendorInfo=brew
</pre>

### Does this issue reproduce with the latest release?
Yes

Trying to integrate TailwindCSS v4 using the [docs](https://gohugo.io/functions/css/tailwindcss/), I experience a `hugo server` crash when editing some part of a template and trying to reload another page. This has been already described in #12963. Using `--disableFastRender` resolves the issue, but there is no livereload functionality either. 

```
>HUGO_MODULE_WORKSPACE=hugo.work hugo server --config hugo.toml,hugo.dev.toml

Watching for changes in /home/german/dev_work/etheria/hugo-modules/{adroller,imgproxy,seo-tools,sky-theme}
Watching for config changes in /home/german/dev_work/etheria/hugo-modules/sky-theme/hugo.toml, /home/german/dev_work/etheria/hugo-modules/sky-theme/hugo.dev.toml, /home/german/dev_work/etheria/hugo-modules/sky-theme/go.mod, /home/german/dev_work/etheria/hugo-modules/sky-theme/hugo.work
Start building sites … 
hugo v0.145.0+extended+withdeploy linux/amd64 BuildDate=2025-02-26T15:41:25Z VendorInfo=brew


                   |  ES   
-------------------+-------
  Pages            |  218  
  Paginator pages  |    0  
  Non-page files   |    5  
  Static files     | 1398  
  Processed images |    0  
  Aliases          |    0  
  Cleaned          |    0  

Built in 860 ms
Environment: "development"
Serving pages from disk
Running in Fast Render Mode. For full rebuilds on change: hugo server --disableFastRender
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1) 
Press Ctrl+C to stop

Change detected, rebuilding site (#1).
2025-03-13 16:14:25.518 +0100
Template changed /posts/next-ad.html
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
Total in 634 ms
panic: deferred execution with id "__hdeferred/8e575cf704df60ac_5be98fa57b3a47362e72ab81d8b5b9f1__d=" not found

goroutine 20445 [running]:
[...]

Issue #13737, opened 2025-05-23
hugo server - Panics and crashes - some random - live reload of js modules (not sure trigger)


naturedamends:
This issue includes the following potential issues
- Panic when file does not exist
- Random panic when using template.defer (may be fixed with by not using a partial)
- Live reload of javascript assets fails with (template.defer or js batch (notsure))
- Can't find the resource
- hugo.toml references a non existing module. hugo server stops rendering

> the following are, `hugo server -M`

I deleted runner2.js.  Should it not handle this panic and keep the server alive for when any errors are resolved.

```Asset changed /js/runner2.js
panic: open /Users/other/repos/gists/hugo-modules/themes/sd-basic/assets/js/runner2.js: no such file or directory

goroutine 681 [running]:
github.com/gohugoio/hugo/internal/js/esbuild.(*batcher).doBuild.func4({{0xc00098d6b0, 0x16}, {0x101b56ffc, 0x10}, {0x0, 0x0}, {0x0, 0x0}, 0x0})
	/root/project/hugo/internal/js/esbuild/batch.go:578 +0x108
github.com/gohugoio/hugo/internal/js/esbuild.createBuildPlugins.func2.3({{0xc00098d6b0, 0x16}, {0x101b56ffc, 0x10}, {0x0, 0x0}, {0x0, 0x0}, 0x0})
	/root/project/hugo/internal/js/esbuild/resolve.go:261 +0x119
github.com/evanw/esbuild/pkg/api.(*pluginImpl).onLoad.func1({{0x0, 0x0}, {{0xc00098d6b0, 0x16}, {0x101b56ffc, 0x10}, {0x0, 0x0}, {{0x0, 0x0}}, ...}})
	/root/project/gomodcache/github.com/evanw/esbuild@v0.25.3/pkg/api/api_impl.go:1996 +0x190
github.com/evanw/esbuild/internal/bundler.runOnLoadPlugins({0xc0007c53f0?, 0xfa0fc7baf79bca60?, 0xb097d09a804dd294?}, {0x102053400?, 0xc001600ca0}, 0xc0015ea4e0, {0xc00034a460, 0xc0016034a0, 0xc0016034b8, 0xc001600c80, ...}, ...)
	/root/project/gomodcache/github.com/evanw/esbuild@v0.25.3/internal/bundler/bundler.go:1125 +0xcb5
[...]

PR #14887, opened 2026-05-14
hugolib: Return error for missing deferred execution

xndvaz:
## Summary

- Return a render error instead of panicking when a published file contains a deferred template placeholder with no matching execution in the current build state.
- Include the output filename in the error to make stale placeholders easier to diagnose.
- Return an error instead of panicking if the deferred template path is missing.

This is a narrow follow-up hardening for one panic path mentioned in #13737. The issue is already closed by the js.Batch missing-import fix, so this intentionally uses See rather than Fixes.

See #13737

## Tests

- go test ./hugolib -run TestDeferredTemplateMissingExecutionReturnsErrorIssue13737 -count=20
- ./check.sh ./hugolib/...
- ./check.sh

## AI Assistance

I used AI assistance to help investigate the issue context, implement the change, and prepare tests. I reviewed the final diff, scope, and test coverage before publishing.

### DIFF
--- a/hugolib/hugo_sites_build.go
+++ b/hugolib/hugo_sites_build.go
@@ -543,7 +543,7 @@ func (s *Site) executeDeferredTemplates(de *deps.DeferredExecutions) error {
 			if err := func() error {
 				deferred, found := de.Executions.Get(id)
 				if !found {
-					panic(fmt.Sprintf("deferred execution with id %q not found", id))
+					return fmt.Errorf("deferred execution with id %q not found in %q", id, filename)
 				}
 				deferred.Mu.Lock()
 				defer deferred.Mu.Unlock()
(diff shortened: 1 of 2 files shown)


bep:
I don't think this is right. This is an error situation that 1. Should never happen (is a bug) and 2. It's not obvious what the user should do about it. So we panic to make it as loud and detailed as possible.

bep:
Ignore my last comment, there is one known case where this missing deferred execution can happen.

hugolib/hugo_sites_build.go
```diff
@@ -543,7 +543,7 @@ func (s *Site) executeDeferredTemplates(de *deps.DeferredExecutions) error {
 			if err := func() error {
 				deferred, found := de.Executions.Get(id)
 				if !found {
-					panic(fmt.Sprintf("deferred execution with id %q not found", id))
+					return fmt.Errorf("deferred execution with id %q not found in %q", id, filename)
```
bep, line 546: Can you adjust this error message to also include something ala:

>templates.Defer can only be used in a main template (typically baseof.html), see https://gohugo.io/functions/templates/defer/

bep:
Never mind, I change my mind again: The proper fix for this is to make it so 

>panic: deferred execution with id "__hdeferred/1e4c77c62ffd7372_11__d=" not found

Never happens.

As in: Somehow fix #13492

PR #14904, opened 2026-05-15
common/hmaps: Merge slice-valued module config into site config

### LINKED ISSUE #13869: Cascade array settings not merged from imported module

hugo v0.148.1-98ba786f2f5dca0866f47ab79f394370bcb77d2f+extended linux/amd64 BuildDate=2025-07-11T12:56:21Z VendorInfo=gohugoio

Yes

Importing a module with cascade array settings does not merge the settings with the main config, even if the main config says so via '_merge=deep' (if I understand it correctly, technically '_merge' should not be necessary here, as there are no other cascade settings, but I could be wrong).

I've created a showcase repo here: https://github.com/doompadee/hugo-cascade

I would expect that ''hugo config' reveals the imported cascade settings. There are none at all when `_merge` is missing. With `merge = deep` there are only the section identifiers present, but no settings values.

See also: #12899

Reference: <https://discourse.gohugo.io/t/unable-to-merge-cascade-config-from-theme-module/55409>

jmooring:
When a module provides a config key whose value is a slice (e.g. cascade or permalinks), and the site config declares the same key as a map with only a merge strategy marker (_merge = 'deep'), the types do not match and Params.merge silently dropped the module's value, leaving the site with no effective cascade or permalink config from the module.

Fix Params.merge so that when the destination value is an empty Params (IsZero — only the _merge key is present) and the source value is a non-Params type, the source value is used provided the user-declared merge strategy is not 'none'. This honours the explicit _merge directive regardless of the surrounding shallow-merge context.

Closes #13869

### DIFF
--- a/common/hmaps/params.go
+++ b/common/hmaps/params.go
@@ -120,6 +120,14 @@ func (p Params) merge(ps ParamsMergeStrategy, pp Params) {
 			if vvv, ok := vv.(Params); ok {
 				if pv, ok := v.(Params); ok {
 					vvv.merge(ms, pv)
+				} else if vvv.IsZero() {
+					// The existing value is an empty Params (just merge metadata)
+					// and the incoming value is a non-Params type (e.g. a slice).
+					// If the user has set an explicit non-none merge strategy,
+					// honor it by using the incoming value.
+					if s, found := vvv.GetMergeStrategy(); found && s != ParamsMergeStrategyNone {
+						p[k] = v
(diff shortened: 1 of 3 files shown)


jmooring:
The changes in https://github.com/gohugoio/hugo/pull/14883 reminded me of https://github.com/gohugoio/hugo/issues/13869, addressed by this PR... they're sort of related.

PR #14905, opened 2026-05-15
hugolib: Preserve nested leaf bundle resources on rebuild

### LINKED ISSUE #13961: erroneously building resources
Assume the following folder structure:

```
/content/_index.md
/content/a/index.md
/content/a/b/index.md
```

When building this, the **expected output** would be:

```
/index.html
/a/index.html
```

`/content/a/b/index.md` is not a valid content file, as `/content/a/index.md` already represents a leaf. This is also the behaviour, that can be observed when building the site with `hugo` (or `hugo server`).

However, if the development server (`hugo server`) is running and **a change to `/content/a/b/index.md` is made**, this will trigger a partial rebuild, that **will create** `/a/b/index.html`. This is due to hugo determining the path type based on the file name rather than the full path of the file.

Rendering `/content/a/b/index.md` can lead to variety of consequences (e.g. ending up anywhere in the output, if `url` is set) and cause a lot of confusion.

<pre>
$ hugo version
hugo v0.147.9+extended+withdeploy linux/amd64 BuildDate=unknown VendorInfo=nixpkgs
</pre>

I tested it only with the version above.

xndvaz:
## Summary

- keep server partial rebuilds rooted at the owning leaf bundle for nested changes
- preserve the full-build behavior where content below a leaf bundle is treated as bundled resources
- add a regression test for editing and adding nested content below a leaf bundle

## Details

The full content walk already has the directory context needed to classify files below a leaf bundle as resources. A server partial rebuild could start below that leaf bundle, so a changed `content/a/b/index.md` was parsed as a new leaf bundle and rendered as `/a/b/`.

This keeps the partial walk rooted at the existing leaf bundle owner for nested changes, letting the existing component filesystem resource classification apply consistently.

Fixes #13961

## Tests

- `go test ./hugolib -run TestRebuildEditNestedContentInLeafBundleIssue13961 -count=1`
- `./check.sh ./hugolib`
- `git diff --check`

I also attempted `./check.sh` locally, but repo-wide staticcheck compilation stopped with `no space left on device` before tests ran. CI should provide the full validation.

## AI Assistance

AI assistance was used to inspect the rebuild path, draft the patch, and run local validation. I reviewed the changes and test coverage before submitting.

### DIFF
--- a/hugolib/pages_capture.go
+++ b/hugolib/pages_capture.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -152,7 +153,9 @@ func (c *pagesCollector) Collect() (collectErr error) {
 		}
 
 		for _, id := range c.ids {
-			if id.p.IsLeafBundle() {
+			if owner := c.leafBundleOwner(id.p); owner != nil {
+				collectErr = c.collectDirBelowLeafBundle(owner, id)
+			} else if id.p.IsLeafBundle() {
 				collectErr = c.collectDir(
 					id.p,
 					false,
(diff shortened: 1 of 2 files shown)


bep:
I'm closing this and assigning it to myself. This issue cannot be vibe coded. It requires a deep knowledge about the existing setup. Also, there are some non-obvious choices to be made here that isn't in the original issue.

PR #14916, opened 2026-05-19
markup/converter/hooks: Allow overriding Type in code block context

### LINKED ISSUE #11872: Allow to specify language for transform.HighlightCodeBlock
As discussed [here](https://discourse.gohugo.io/t/change-language-in-highlightcodeblock/47724), it would be useful if one could overwrite the language in `transform.HighlightCodeBlock` - similar how it is possible to do so in `transform.Highlight`. At the moment `transform.HighlightCodeBlock` uses `.Type` which can't be changed.

One suggestion is to allow to overwrite the language via the options. (Although this could create some "interesting" scenarios if this option is then used in `transform.Highlight` or in a code block attribute in Markdown.)

jmooring:
Closes #11872

---

The purpose of this draft PR is to help evaluate the proposal described in this [comment](https://github.com/gohugoio/hugo/issues/11872#issuecomment-4472928140). No immediate code review expected; just sharing as a proof of concept for whenever we triage the issue.

Note that the intended use case was not possible prior to the recently merged fix in https://github.com/gohugoio/hugo/pull/14911.

### DIFF
--- a/markup/converter/hooks/hooks.go
+++ b/markup/converter/hooks/hooks.go
@@ -70,6 +70,9 @@ type CodeblockContext interface {
 	// The type of code block. This will be the programming language, e.g. bash, when doing code highlighting.
 	Type() string
 
+	// WithType returns a copy of the context with Type overridden.
+	WithType(lang string) CodeblockContext
+
 	// The text between the code fences.
 	Inner() string
 }
--- a/markup/goldmark/codeblocks/codeblocks_integration_test.go
+++ b/markup/goldmark/codeblocks/codeblocks_integration_test.go
@@ -165,6 +165,50 @@ plain <code> & text
 	)
 }
 
+// Issue 11872
+func TestCodeblockWithType11872(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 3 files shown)


bep:
I don't think we use the `WithType` construct in other user facing APIs, and I think we want to think hard before doing it. I suspect it's harder to document/explain to people, and maybe less flexible than the stupid alternative.


We already have a options map to "override the default highlighting options":

```
transform.Highlight CODE LANG [OPTIONS]
transform.HighlightCodeBlock CONTEXT [OPTIONS]
```

I don't see why we cannot make `type` into a "highlighting options" and consolidate the above into:


```
transform.Highlight CODE [OPTIONS]
transform.HighlightCodeBlock CONTEXT [OPTIONS]
```

We could even simplify the above to:

```
transform.Highlight OPTIONS
transform.HighlightCodeBlock CONTEXT [OPTIONS]
```

And make `code` into an `option`. In the last line, the definition of `Options` would be:

"override the default highlighting options and the options passed in from Markdown".

PR #14931, opened 2026-05-23
tpl/collections: Include template path in IsSet unsupported type warning

### LINKED ISSUE #11794: Warning about "IsSet with unsupported type" is not helpful
Issue #8667 still applies in latest

```
❯ hugo
Start building sites …
hugo v0.121.1-00b46fed8e47f7bb0a85d7cfc2d9f1356379b740+extended darwin/amd64 BuildDate=2023-12-08T08:47:45Z VendorInfo=brew

WARN  calling IsSet with unsupported type "invalid" (<nil>) will always return false.

```
The fix in #8668 is valid , even if I'd do it this way:

```
diff --git a/tpl/collections/collections.go b/tpl/collections/collections.go
--- a/tpl/collections/collections.go	(revision 6f13430d4a3b0d8b196f13958fbfb6478be1f3aa)
+++ b/tpl/collections/collections.go	(date 1702330109068)
@@ -396,7 +396,7 @@
 			return av.MapIndex(kv).IsValid(), nil
 		}
 	default:
-		ns.deps.Log.Warnf("calling IsSet with unsupported type %q (%T) will always return false.\n", av.Kind(), c)
+		ns.deps.Log.Warnf("calling IsSet with unsupported type %q (%T) [Key: %q] will always return false.\n", av.Kind(), c, kv)
 	}
 
 	return false, nil

```
This allows to see that a key 'name' is invalid

> ❯ /Users/hrupp/golang/bin/hugo
> Start building sites …
> hugo v0.122.0-DEV-6f13430d4a3b0d8b196f13958fbfb6478be1f3aa darwin/amd64 BuildDate=2023-12-08T09:01:24Z
> 
> WARN  calling IsSet with unsupported type "invalid" (<nil>) [Key: "name"] will always return false.

which then allows me to look through my files to see that this is the culprit

                    {{if isset .Site.Params.author "name"}}

which then allows me to see that indeed in hugo.toml, there is no section `[params.author]`.
A better fix would be to also show the place of the issue, but with the above, the message can at least be used for debugging.

truffle-dev:
Fixes #11794

## Summary

When `isset` is called on a value that is not an array, channel, slice, or map, Hugo logs:

```
WARN  calling IsSet with unsupported type "string" (string) will always return false.
```

On any site with more than one template there's no way to find the offending call site. The warning is also emitted via `Log.Warnf`, so it has no ID and can't be silenced via `ignoreLogs`.

This PR threads `context.Context` through `IsSet`, looks up the executing template via `tpl.Context.CurrentTemplate.Get`, and surfaces the filename in the warning. It also switches `Log.Warnf` to `Log.Warnidf` with a new constant `WarnIsSetUnsupportedType` so the warning is groupable and suppressible.

After the change:

```
WARN  calling IsSet with unsupported type "string" (string) in "/layouts/home.html" will always return false.
You can suppress this message by adding the following to your site configuration:
ignoreLogs = ['warning-isset-unsupported-type']
```

## Changes

- `tpl/collections/collections.go` — `IsSet` accepts `ctx context.Context` as its first arg, matching the pattern used by `Delimit` and other ctx-aware namespace functions. The default branch resolves the current template and appends `in "<path>"` to the warning, then routes through `Log.Warnidf`.
- `common/constants/constants.go` — adds `WarnIsSetUnsupportedType`.
- `tpl/collections/collections_test.go` — passes `context.Background()` to the existing `TestIsSet` cases.
- `tpl/collections/collections_integration_test.go` — adds a regression test asserting the template path appears in the warning.
- `docs/data/docs.yaml` — adds `ctx` to `IsSet`'s `Args` list, matching the documented shape for other ctx-aware namespace functions.

## Tests

- `go test ./tpl/collections/... -run TestIsSet -v` — unit and integration tests pass.
- `go test ./tpl/collections/...` — full package green.
- `go vet ./tpl/collections/...` — clean.
- `gofmt -l` on touched paths — clean.
- `go build ./...` — full tree builds.

## AI Assistance

I used AI assistance to investigate the issue, design the fix, and prepare tests. I reviewed the final diff, kept the scope narrow per CONTRIBUTING.md, and ran the checks listed above before opening.

### DIFF
--- a/common/constants/constants.go
+++ b/common/constants/constants.go
@@ -20,6 +20,7 @@ const (
 	WarnGoldmarkRawHTML          = "warning-goldmark-raw-html"
 	WarnPartialSuperfluousPrefix = "warning-partial-superfluous-prefix"
 	WarnHomePageIsLeafBundle     = "warning-home-page-is-leaf-bundle"
+	WarnIsSetUnsupportedType     = "warning-isset-unsupported-type"
 )
 
 // Field/method names with special meaning.
--- a/docs/data/docs.yaml
+++ b/docs/data/docs.yaml
@@ -2376,6 +2376,7 @@ tpl:
         - isSet
         - isset
         Args:
+        - ctx
         - c
         - key
         Description: |-
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -20,17 +20,20 @@ import (
 	"errors"
 	"fmt"
 	"math/rand/v2"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"time"
 
 	"github.com/gohugoio/hugo/common/collections"
+	"github.com/gohugoio/hugo/common/constants"
 	"github.com/gohugoio/hugo/common/hmaps"
 	"github.com/gohugoio/hugo/common/hreflect"
 	"github.com/gohugoio/hugo/common/hstore"
 	"github.com/gohugoio/hugo/common/types"
 	"github.com/gohugoio/hugo/deps"
 	"github.com/gohugoio/hugo/langs"
+	"github.com/gohugoio/hugo/tpl"
 	"github.com/gohugoio/hugo/tpl/compare"
 	"github.com/spf13/cast"
 )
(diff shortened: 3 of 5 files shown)


truffle-dev:
Hi @bep — small focused fix against #11794 (open since 2023), ~13 days idle. Happy to rebase or adjust the warning message if anything looks off. No rush, just keeping it on your radar.

truffle-dev:
Closing this for now — it's sat about a month without traction, and I know #11794 is low-priority, so I'd rather not leave a stale PR in your review queue. The fix still stands if it's ever useful; happy to reopen anytime. Thanks @bep.

PR #14938, opened 2026-05-25
commands: Improve chtimes permission hint

### LINKED ISSUE #7302: Error when building to a folder not owned by the Hugo process

```shell
$ hugo version
Hugo Static Site Generator v0.71.0-06150C87/extended linux/amd64 BuildDate: 2020-05-18T16:15:29Z
```

Yes it does.

The error occures when the Hugo process have write access to the destination folder but does not own it.

To reproduce, create a `public` folder owned by root in your Hugo site project (requires sudo access):

```shell
sudo sh -c "mkdir public && chmod a+w public"
```

Run Hugo:

```
hugo
```

Log:

```
Building sites … WARN 2020/05/21 14:09:25 found no layout file for "HTML" for kind "section": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
Total in 111 ms
Error: Error copying static files: chtimes [..]/public/: operation not permitted
```

xndvaz:
Fixes #7302

## Summary
- add a targeted hint for static sync permission failures caused by `chtimes`
- apply the same hint in both the build path and `hugo server` static sync logging
- add focused tests for the `chtimes` permission case and non-matching errors

## Root cause
When the publishDir is writable but not owned by the Hugo process, the OS can reject `chtimes` during static sync. Hugo surfaced the raw error, but did not point users to the existing `--noTimes` / `noTimes = true` workaround.

## Validation
- `PATH="/Users/xndvaz/go/bin:$PATH" ./check.sh ./commands/...`
- `PATH="/Users/xndvaz/go/bin:$PATH" ./check.sh`

## AI assistance
This PR was written with Codex assistance. I reviewed the generated changes and validation results.


commands/errors_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"syscall"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+)
+
+func TestWrapStaticSyncError(t *testing.T) {
+	c := qt.New(t)
+
+	err := wrapStaticSyncError(&os.PathError{Op: "chtimes", Path: "public", Err: syscall.EPERM})
```
bep: In my head there are 2 types of tests:

1. Unit tests (tests e.g. a function in isolation)
2. Integration tests (tests more or less the entire chain with as a real a setup as practical)

This test falls between these two and I question its value. A test for a bug fix needs to 1. Demonstrate that we have fixed the bug and 2. Prevent future regressions of that same bug. The error created here is a fake one, which means that I don't know if it matches reality.

I'm assuming that testing the real scenario (e.g. with a testscript) would not be practical, so I suggest just removing the test. I have no problem seeing what the wrapper func does, and I assume it will work, and if not if will not make it worse.

xndvaz: Removed the synthetic test as suggested in `110d69d45`.

I re-ran `PATH="/Users/xndvaz/go/bin:$PATH" ./check.sh ./commands/...` and `PATH="/Users/xndvaz/go/bin:$PATH" ./check.sh` locally before pushing. The PR checks are re-running now.

bep:
@xndvaz have you manually tested this patch and can confirm that this is works as described?

jmooring:
> have you manually tested this patch and can confirm that this is works as described?

This does what it says it should do, but I don't think it improves the user experience in relation to the conditions described in #7302.

```sh
$ sudo sh -c "mkdir public && chmod a+w public"

$ hugo
Start building sites … 
hugo v0.163.0-DEV-889ce2c161636ff9e04db6f507fae5bbb273a92a+withdeploy linux/amd64 BuildDate=2026-06-01T18:48:36Z VendorInfo=jmooring

Total in 11 ms
ERROR error copying static files: chtimes /home/jmooring/code/hugo-testing/public: operation not permitted; if the publishDir is writable but not owned by the Hugo process, try --noTimes or set noTimes = true

$ hugo --noTimes  # same error when combined with --noChmod
Start building sites … 
hugo v0.163.0-DEV-889ce2c161636ff9e04db6f507fae5bbb273a92a+withdeploy linux/amd64 BuildDate=2026-06-01T18:48:36Z VendorInfo=jmooring

Total in 10 ms
ERROR error copying static files: stat /home/jmooring/code/hugo-testing/public/favicon.ico: permission denied
```

So you fix one error, and then encounter another. It seems to me we should just state something like "you have insufficient permissions to write files to the publishdDir".

xndvaz:
Closing this as well, in line with #14841.

This is a much smaller change, but I understand the maintenance/review cost concern around AI-assisted contributions and don't want to add more burden here.

Thanks for the review and feedback.

Issue #12192, opened 2024-03-01
404 page treated as paginated


earthboundkid:
Building my site with Hugo 0.123.7 results in a /404/ directory being created with 99 entries, as though it were paginated.

See #12174.

### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
hugo v0.123.7-312735366b20d64bd61bff8627f593749f86c964+extended darwin/arm64 BuildDate=2024-03-01T16:16:06Z VendorInfo=brew
</pre>

### Does this issue reproduce with the latest release?

Yes.

jmooring:
I have no idea how to reproduce this. Can you please provide an example?

earthboundkid:
Clone https://github.com/spotlightpa/poor-richard. Run yarn. Run `hugo`. Look at the /404/ directory.

jmooring:
1\) [`layouts/_default/baseof.html`](https://github.com/spotlightpa/poor-richard/blob/master/layouts/_default/baseof.html#L12) calls [`layouts/partials/helper/paginator.html`](https://github.com/spotlightpa/poor-richard/blob/master/layouts/partials/helper/paginator.html).

2\) [`layouts/partials/helper/paginator.html`](https://github.com/spotlightpa/poor-richard/blob/master/layouts/partials/helper/paginator.html#L7) paginates if `Page.IsNode` is `true`.

3\) `Page.IsNode` is `true` for page kind `404`. It has been that way since v0.54.0 (maybe earlier). Whether the return value makes sense is debatable; see #12162.

So, the partial says, "Paginate this thing,"  which obviously doesn't make any sense for a 404 page. We'll get to why things changed in a minute, but you should exclude page kind `404` from pagination by changing this:

```text
{{ if .IsNode }}
```

To this:

```text
{{ if and .IsNode (ne .Kind "404") }}
```

4\) This should have been happening with v0.122.0 and earlier. Why wasn't it?

- In both versions the `.Pages` collection passed into the 404 template seems to be equivalent to `Site.Pages`. I don't understand why the 404 template needs a `.Pages` collection, but that's another topic. The fact that there _is_ a `Pages` collection means the the `Paginator` method can act upon something.
- In v0.122.0 and earlier, calling the `Paginator` method within a 404 template didn't do anything. Now it does.

It looks like we fixed (either directly or indirectly) the `Paginator` method in v0.123.0 when called from the 404 template, and that change exposed the conditional logic problem in the partial.

[...]

Issue #11574, opened 2023-10-18
Deprecate Page.IsNode


jmooring:
Why?
- It's meaning is unclear
- It's superfluous (use `if not .IsPage` instead)
- Not sure anyone actually uses it

onedrawingperday:
Not in favour of this proposal as I am using it in my templating.

bep:
This needs to at least wait (and in any case, we can limit the change to the documentation to start with).

jmooring:
Updated documentation.

jmooring:
I'm reopening this because we recently introduced the concept of logical paths, and the easiest way to describe a logical path is the path between two nodes on a node tree, either relative to each other or absolute from the root of the tree.

In a data structure, the term "node" is never limited to a parent.
<https://en.wikipedia.org/wiki/Tree_(data_structure)>

But our existing definition of node is limited to page kinds `home`, `section`, `taxonomy`, `term`, `404`, and maybe some others, _but the definition excludes page kind `page`_.

I would like to deprecate `.Page.IsNode` recommending the use of `not .IsPage` instead. `.IsPage` returns `false` for page kinds  `home`, `section`, `taxonomy`, `term`, `404`, and maybe some others.

This is not a technical issue... `.IsNode` works fine. This is about the language of Hugo and the opportunity to clarify concepts in documentation.

jmooring:
@bep Please comment when you have a moment. Thanks.

jmooring:
I'd like to revisit this.

Deprecating `Page.IsNode` in favor of `Page.IsList` would be a good change, not for technical reasons, but for conceptual clarity.

Currently, we define a "node" as a specific set of page kinds (`home`, `section`, `taxonomy`, `term`). However, in a data tree, a node is any item in the tree, whether it has descendants or not.

This broader, more standard definition of "node" would allow us to more accurately describe a page's logical path as the path between any two nodes, either relative to each other or absolute from the root. This is not currently possible with our existing [definition](https://gohugo.io/quick-reference/glossary/#node).

jmooring:
@bep I'd like to either close this issue or accept the proposal.

bep:
This issue got under my radar, I apologise for that.

To sum up; currently we have:

* IsPage: Kind == page
* IsNode: !IsPage()

The first one is well defined, the latter is murky.

In Hugo's content tree, every piece of content is defined as a node that is either a `branch` (can have children) or a `leaf` node.

[...]

PR #14956, opened 2026-05-27
hugolib: Make IsNode false for 404 pages

### LINKED ISSUE #12162: .IsNode is true for 404.html
As discussed [here](https://discourse.gohugo.io/t/404-html-reports-isnode-as-true/48487), `.IsNode` should be `false` for the `404.html` template.
 
<pre>
$ hugo version
hugo v0.123.3-a75a659f6fc0cb3a52b2b2ba666a81f79a459376+extended linux/amd64 BuildDate=2024-02-23T17:09:20Z VendorInfo=gohugoio
</pre>

Yes

xndvaz:
## Summary
- make `.IsNode` return `false` for `kind=404`
- add an integration test covering the `404.html` template context

## Context
The public `IsNode` contract describes node pages as `home`, `section`, `taxonomy`, and `term`.
The 404 template was returning `true` because `pageMeta.IsNode` treated every non-`page` kind as a node.

## Root cause
`pageMeta.IsNode` used `!IsPage()` instead of checking the actual branch kinds.

## Testing
- `go test ./hugolib -run 'Test404$|Test404IsNotNode$|Test404EditTemplate$|Test404Panic14283$'
- `PATH="/Users/xndvaz/go/bin:$PATH" ./check.sh`

## Related
- Fixes #12162
- Related #12192
- Related #11574

## AI assistance
This PR was written with Codex assistance, then reviewed and validated locally.

### DIFF
--- a/hugolib/404_test.go
+++ b/hugolib/404_test.go
@@ -63,6 +63,29 @@ Data: 1|
 `)
 }
 
+// See issue 12162.
+func Test404IsNotNode(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+baseURL = "http://example.com/"
+disableKinds = ["rss", "sitemap", "taxonomy", "term"]
+-- layouts/404.html --
+Kind: {{ .Kind }}|
+IsNode: {{ .IsNode }}|
+IsPage: {{ .IsPage }}|
+`
+
+	b := Test(t, files)
+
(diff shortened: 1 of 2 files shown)


bep:
There are several questions (e.g. #11574) that needs to be answered (which I guess we should have done a long time ago) before we can implement something.

Issue #14991, opened 2026-06-04
resources/jsconfig: Remove deprecated baseUrl setting


jmooring:
When working on the docs site I saw this problem notice in VS Code:

> Option 'baseUrl' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
  Visit https://aka.ms/ts6 for migration information.

The linked page states:

> This feature was designed for use in conjunction with AMD module loaders in the browser, and is not recommended in any other context. As of TypeScript 4.1, baseUrl is no longer required to be set when using [paths](https://www.typescriptlang.org/tsconfig/#paths).

And we're setting `paths` so I think we can remove the `baseURL` setting in `jsconfig.json`.

PR #14996, opened 2026-06-04
resources/jsconfig: Remove deprecated baseUrl setting

gugacyber:
This PR removes the deprecated `baseUrl` setting from `resources/jsconfig` as requested in #14991. 

All tests passed locally.


bep:
The build fails (malformed source code).

PR #15022, opened 2026-06-09
herrors: standardize feature not available error message

### LINKED ISSUE #13003: Standardize error message thrown when a feature is not available
I'd like to standardize the error message thrown when a particular feature is not available in the current version. This change would make this section of the documentation a bit simpler:

<https://gohugo.io/troubleshooting/faq/#an-error-message-indicates-that-a-feature-is-not-available-why>

I will handle this if this proposal is accepted.

cchakibb:
Fixes #13003

Standardizes the error message thrown when a feature is not available
in the current Hugo edition.

The previous message referenced "your current Hugo version" which was
inaccurate — the issue is edition-specific (standard vs extended), not
version-specific. The message also included a goo.gl shortlink whose
anchor has become stale since the FAQ was updated.

The new message aligns with what the documentation already displays:

    this feature is not available in this edition of Hugo

### DIFF
--- a/common/herrors/errors.go
+++ b/common/herrors/errors.go
@@ -84,7 +84,7 @@ func IsFeatureNotAvailableError(err error) bool {
 //
 // We will, at least to begin with, make some Hugo features (SCSS with libsass) optional,
 // and this error is used to signal those situations.
-var ErrFeatureNotAvailable = &FeatureNotAvailableError{Cause: errors.New("this feature is not available in your current Hugo version, see https://goo.gl/YMrWcn for more information")}
+var ErrFeatureNotAvailable = &FeatureNotAvailableError{Cause: errors.New("this feature is not available in this edition of Hugo")}
 
 // FeatureNotAvailableError is an error type used to signal that a feature is not available.
 type FeatureNotAvailableError struct {


jmooring:
Issue #13003 is unclear about what "standardize" means, but it was primarily about "this version/edition doesn't support" vs "you must install" -- those are two very different things. For example, when transpiling to CSS with Dart Sass and Dart Sass is not installed, we throw the "this feature is not available in your current Hugo version" message, which is inaccurate. The _feature_ is available, but the _executable_ is not.

> ERROR error building site: render: [en v1.0.0 guest] failed to render pages: render of "/home/jmooring/code/hugo-feature-test/content/tests/dart-sass.md" failed: "/home/jmooring/code/hugo-feature-test/layouts/tests/dart-sass/page.html:23:18": execute of template failed: template: tests/dart-sass/page.html:23:18: executing "main" at <.Content>: error calling Content: TOCSS-DART: failed to transform "/temp/sass" (application/octet-stream). You need to install Dart Sass, see https://gohugo.io//functions/css/sass/#dart-sass: this feature is not available in your current Hugo version, see https://goo.gl/YMrWcn for more information

In any event, changing the text of one of the related messages doesn't address the issue.

cchakibb:
Hi @jmooring , thanks for the detailed explanation! Makes total sense. I misunderstood the scope and thought it was just a string update, but I see now that it requires a deeper architectural distinction between missing executables and edition limitations. I'll leave this to you and look for another issue to get familiar with the codebase. 
Cheers!

PR #15029, opened 2026-06-11
commands: Add --skipUnchanged flag for mtime-based render optimization

### LINKED ISSUE #15026: commands: Add --skipUnchanged flag for mtime-based render optimization
When running `hugo` (static generation), every page is always re-rendered regardless of whether the source content has changed. The fast render optimization that skips unchanged pages only applies to `hugo server` mode.

This proposes a `--skipUnchanged` flag that, during a plain `hugo` build, skips rendering a page if its output file is newer than its source content file (and all config files). This speeds up incremental builds where only a few content files have changed.

Pages without a source file (home, taxonomies, sections) are always rendered. The flag has no effect in server mode, which already has its own incremental rebuild logic.

patrikhson:
Fixes #15026

Adds a `--skipUnchanged` flag to the `hugo` build command. When set, pages whose output file is newer than both the source content file and all config files are skipped entirely, speeding up incremental builds where only a few content files have changed.

## How it works

- Compares the mtime of `content/posts/my-post.md` against `public/posts/my-post/index.html`
- Also checks all config files (hugo.toml, theme configs, module configs), since every rendered page depends on site config
- Pages without a source file (home, taxonomy, section list) are always rendered — there is no single file to compare against
- Has no effect in `hugo server` mode, which already has watch-based incremental rebuilds

## New files

- `hugofs/mtime_syncer.go` — replaces `spf13/fsync` for static file copying with an mtime+size-based syncer
- `hugofs/mtime_syncer_test.go` — unit tests
- `hugolib/skip_unchanged_test.go` — integration tests

## AI disclosure

This PR was written primarily by Claude Code.

### DIFF
--- a/commands/commandeer.go
+++ b/commands/commandeer.go
@@ -134,6 +134,7 @@ type rootCommand struct {
 	poll            string
 	forceSyncStatic bool
 	panicOnWarning  bool
+	skipUnchanged   bool
 
 	// Profile flags (for debugging of performance problems)
 	cpuprofile   string
@@ -392,6 +393,7 @@ func (r *rootCommand) Name() string {
 
 func (r *rootCommand) Run(ctx context.Context, cd *simplecobra.Commandeer, args []string) error {
 	b := newHugoBuilder(r, nil)
+	b.skipUnchanged = r.skipUnchanged
 
 	if !r.buildWatch {
 		defer b.postBuild("Total", time.Now())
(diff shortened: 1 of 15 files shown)


patrikhson:
I noticed the original patch was unintentionally changing behavior for all users: the new ultra fast `MtimeSyncer` had replaced `spf13/fsync` unconditionally for static file sync. The difference matters because `fsync` compares file contents byte-for-byte (checking sizes first, then reading both files if sizes match), while `MtimeSyncer` checks size and mtime. If one want to use the traditional code, just do not add --skipUnchanged argument, and the new feature will not be used.

The patch has been updated. Static file sync now uses `spf13/fsync` (the original behavior) by default. `MtimeSyncer` is only activated when `--skipUnchanged` is passed, consistent with the rest of the feature.

bep:
Thanks for this,

patrikhson:
The PR references issue [#15026](https://github.com/gohugoio/hugo/issues/15026) at the top of the description with _**Fixes #15026**_. Does that not satisfy the requirement, or should it be linked differently? Maybe the problem here is that I did both suggest an enhancement AND post a patch basically at the same time?

patrikhson:
Thank you for the feedback — it has been very helpful in understanding the full scope of what a correct implementation would require.

I am closing this PR. The current approach of comparing only content file and config file mtimes is too naive, as correctly pointed out.

However, while looking into this further I found that Hugo already builds a complete dependency graph during rendering — the `identity` package tracks exactly which templates, partials, shortcodes, and data files each page depends on. This is what powers the fast render mode in `hugo server`. The difference is that in server mode this graph lives in memory across rebuilds, while a static `hugo` build throws it away when it finishes.

A correct implementation would persist that graph to disk after a full build, together with the mtime of every input file. On the next build, stat those files, find what changed, walk the graph to identify affected pages, and only re-render those. Hugo already has most of the pieces — the missing one is persisting the graph between builds.

I will keep the feature request open in case there is interest in pursuing this properly.

PR #15031, opened 2026-06-11
metrics: format durations with consistent decimal places in --templateMetrics output

### LINKED ISSUE #15027: metrics: --templateMetrics duration columns have inconsistent formatting
The `--templateMetrics` output table prints duration values (sum/avg/max) using `time.Duration.String()`, which varies both units and decimal places depending on the magnitude. This makes the table hard to scan when rows have durations of different magnitudes.

**Proposed fix:** Use a fixed-decimal formatter (two decimal places, unit suffix padded to consistent width) so decimal points align across all rows in the table.

patrikhson:
Fixes #15027

The `--templateMetrics` table prints duration values (sum/avg/max columns) using `time.Duration.String()`, which trims trailing zeros and produces anywhere from 0 to 9 decimal places depending on the exact value. This makes columns hard to scan.

Before:
```
  cumulative      average      maximum  count  template
  ----------     --------     --------  -----  --------
          2s   1.23456789s       500ms      4  _default/single.html
       1.5ms     750.123µs        1.2µs    42  partials/head.html
```

After:
```
  cumulative      average      maximum  count  template
  ----------     --------     --------  -----  --------
       2.00s        1.23s     500.00ms      4  _default/single.html
       1.50ms     750.12µs       1.20µs    42  partials/head.html
```

This PR adds a `formatDuration` helper that always produces exactly two decimal places.

## AI disclosure

This PR was written primarily by Claude Code.


jmooring:
Closing. The proposal in https://github.com/gohugoio/hugo/issues/15027 has not been accepted.

<https://github.com/gohugoio/hugo?tab=readme-ov-file#contributing>

<https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#code-contribution>

> A non-trivial code change without an issue reference with one of the labels `bug` or `enhancement` will not be merged.

<https://gohugo.io/contribute/development/#introduction>

And by the way, this PR affects 20 files, changing far more than is required. I suggest you examine your local Git workflow.

PR #15037, opened 2026-06-14
tpl: Refine error for return in non-partial templates

### LINKED ISSUE #11708: Refine error for return in non-partial templates
layouts/home.html

```text
{{ return "hugo" }}
```

The error message is:

> execute of template failed: template: all.html:2:5: executing "all.html" at <return>: wrong number of args for return: want 0 got 1

wondr-wclabs:
Fixes #11708.

When a user used the `return` function in a non-partial template (such as a standard layout), the Go `text/template` engine threw an unhelpful error. This intercepts the `return` node during AST transformation for non-partial templates and outputs a clear error instead. Added a unit test.

### DIFF
--- a/tpl/partials/partials_integration_test.go
+++ b/tpl/partials/partials_integration_test.go
@@ -335,3 +335,20 @@ BAR
 
 	b.AssertFileContent("public/index.html", "OO:BAR")
 }
+
+// See Issue #11708
+func TestReturnInNonPartialError(t *testing.T) {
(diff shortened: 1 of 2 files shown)


bep:
There's a test failure.

jmooring:
Superseded by https://github.com/gohugoio/hugo/pull/15215. New message is:
> execute of template failed: return with a value is only supported in partials

PR #15058, opened 2026-06-23
hugolib: Throw error when Render cannot find template

### LINKED ISSUE #15052: hugolib: Throw error when Render method is unable to find specified template
Currently, when you invoke `.Page.Render` and pass it the name of a non-existent template, the method returns an empty string without warning or error. This isn't anything new; Hugo has behaved this way since 0.54.0, and probably earlier.

There may be a good reason for the existing behavior, but it was surprising to me.

Relevant code:

https://github.com/gohugoio/hugo/blob/d15baf53a91372843c45eef7eb5b87c25a4b6bf1/hugolib/page__per_output.go#L107-L126

Integration test:

```go
// See issue 15052.
func TestRenderViewNotFound(t *testing.T) {
	t.Parallel()

	files := `
-- hugo.toml --
-- layouts/home.html --
{{ .Render "view_foo" }}
{{ .Render "view_bar" }}
-- layouts/view_foo.html --
foo
`

	b.Assert(err, qt.IsNotNil) // fails (no error is thrown)
}
```

Reference: <https://discourse.gohugo.io/t/silent-error-with-render-function-in-v0-163-2-extended-withdeploy/57290/7>

bsyilmaz:
Previously, invoking `.Page.Render` with a non-existent template name would silently return an empty string. This behavior masks developer errors and makes debugging template failures difficult.

This commit updates the `Render` method to explicitly return a formatted error containing the name of the missing template when it cannot be resolved, adhering to Go's idiomatic error-handling patterns.

Fixes #15052

### DIFF
--- a/hugolib/page__per_output.go
+++ b/hugolib/page__per_output.go
@@ -114,7 +114,7 @@ func (pco *pageContentOutput) Render(ctx context.Context, layout ...string) (tem
 	}
 
 	if !found {
-		return "", nil
+		return "", fmt.Errorf("template %q not found", layout)
 	}
 
 	// Make sure to send the *pageState and not the *pageContentOutput to the template.
--- a/hugolib/page_test.go
+++ b/hugolib/page_test.go
@@ -2055,3 +2055,22 @@ disableKinds = ["taxonomy", "term"]
 		}
 	}
 }
+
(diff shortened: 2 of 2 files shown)


jmooring:
See #15055.

PR #15069, opened 2026-06-27
transform: Fix invalid YAML anchors for repeated empty arrays

### LINKED ISSUE #14596: Invalid YAML Anchor Syntax in transform.Remarshal
# Hugo Bug Report: Invalid YAML Anchor Syntax in transform.Remarshal

## Summary

`transform.Remarshal "yaml"` generates invalid YAML syntax when marshalling data structures with repeated empty arrays. It creates self-referencing anchors (`options: &options *options`) which violates YAML specification and causes parse errors.

## Affected Versions

- **Broken**: Hugo v0.155.0, v0.155.1, v0.155.2, v0.155.3, v0.156.0, v0.157.0
- **Works**: Hugo v0.147.7 and earlier

The bug appears to have been introduced between v0.147.7 and v0.155.0.

## Impact

Any Hugo site using `transform.Remarshal "yaml"` to generate YAML files will produce **invalid YAML** that cannot be parsed by standard YAML parsers (yq, Python PyYAML, Go yaml.v3, etc.).

## Description

When Hugo's `transform.Remarshal "yaml"` encounters repeated empty arrays in a data structure, it attempts to optimize by creating YAML anchors. However, it generates syntactically invalid YAML:

**Invalid output (Hugo v0.155.0+):**
```yaml
slug:
  options: &options *options  # ❌ INVALID: Can't define and reference on same line
  pattern: /^[a-z0-9]+$/
```

**Valid output (Hugo v0.147.7):**
```yaml
slug:
  options: []  # ✓ VALID: Simple empty array
  pattern: /^[a-z0-9]+$/
```

**Correct anchor syntax would be:**
```yaml
slug:
  options: &options []  # Define anchor with value
  pattern: /^[a-z0-9]+$/

title:
  options: *options     # Reference anchor elsewhere
  pattern: /^.+$/
```

### 1. Create Hugo site

```bash
hugo new site yaml-bug-test
cd yaml-bug-test
```

### 2. Create go.mod

```go
module example.com/yaml-bug-test

go 1.18
```

### 3. Create config.yml

```yaml
title: YAML Bug Test

outputFormats:
  TestSchema:
    baseName: test-schema
    mediaType: text/yaml
    isPlainText: true

outputs:
  home:
    - TestSchema

params:
  testData:
    field1:
      options: []
      pattern: /^test1$/
    field2:
      options: []
      pattern: /^test2$/
    field3:
      options: []
      pattern: /^test3$/
```

### 4. Create layouts/index.testschema.yaml

```go
{{ .Site.Params.testData | transform.Remarshal "yaml" -}}
```

### 5. Build and verify

```bash
hugo --gc

# Check for invalid anchors
grep "options:.*\*" public/test-schema.yaml

# Attempt to parse (will fail with v0.155.0+)
yq eval public/test-schema.yaml
```

**Expected output (Hugo v0.147.7):**
```yaml
field1:
  options: []
  pattern: /^test1$/
field2:
  options: []
  pattern: /^test2$/
field3:
  options: []
  pattern: /^test3$/
```

**Actual output (Hugo v0.155.0+):**
```yaml
field1:
  options: &options []
  pattern: /^test1$/
field2:
  options: *options *options    # ❌ INVALID SYNTAX
  pattern: /^test2$/
field3:
  options: *options *options    # ❌ INVALID SYNTAX
  pattern: /^test3$/
```

## Error When Parsing

```bash
$ yq eval public/test-schema.yaml
Error: bad file 'public/test-schema.yaml': yaml: line X: could not find expected ':'
```

## Root Cause Analysis

The bug appears to be in Hugo's YAML marshaller when it attempts to create anchors for repeated values. The logic incorrectly generates:

```yaml
key: &anchor *anchor
```

Instead of the correct two-step process:
1. Define anchor: `key: &anchor value`
2. Reference anchor elsewhere: `key: *anchor`

### Temporary Fix 1: Downgrade Hugo
Use Hugo v0.147.7 or earlier:
```yaml
# In CI/CD workflows
hugo-version: "0.147.7"
```

### Temporary Fix 2: Use JSON + yq
Avoid Hugo's YAML marshaller entirely:

**Change template from:**
```go
{{ $data | transform.Remarshal "yaml" -}}
```

**To:**
```go
{{ $data | jsonify -}}
```

Then convert JSON to YAML externally:
```bash
hugo --gc
yq eval -P public/schema.json > public/schema.yaml
```

## Environment

- **OS**: macOS (ARM64), Linux (AMD64) - affects all platforms
- **Hugo Installation**: Binary, Docker (hugomods/hugo), all installation methods
- **Go Version**: Not relevant (Hugo binary issue)

## Additional Context

This bug breaks any workflow that:
1. Uses `transform.Remarshal "yaml"` to generate YAML configuration files
2. Has data structures with repeated empty arrays or slices
3. Needs to process the generated YAML with external tools (yq, Ansible, Kubernetes, etc.)

The bug appears to be in Hugo's underlying YAML library or its usage of it. The self-referencing anchor syntax `&anchor *anchor` has never been valid YAML according to the [YAML 1.2 specification](https://yaml.org/spec/1.2/spec.html#id2785586).

## Reproduction Repository

[Link to minimal reproduction repo if you create one]

## Related Issues

[Search results from Hugo issues if any similar issues exist]

---

**This bug makes `transform.Remarshal "yaml"` unusable for production YAML generation in Hugo v0.155.0+**

amitmishra11:
This PR was written primarily by Claude Code; I reviewed the change and verified it against the existing tests before submitting.

## Problem

`transform.Remarshal "yaml"` emits invalid, self-referencing YAML anchors when the input contains more than one empty array, e.g.:

```yaml
field1:
  options: &options []
field2:
  options: &options *options    # invalid: defines and dereferences the same anchor at once
```

Root cause: the YAML decoder (`goccy/go-yaml`) returns the exact same shared slice instance for every empty array literal in the source document. The smart-anchor-aware encoder option (`yaml.WithSmartAnchor()`, used to guard against billion-laughs-style expansion attacks) then treats these as one value referenced multiple times, and its anchor handling for empty sequences produces invalid output.

## Fix

In `applyMarshalTypes` (`tpl/transform/remarshal.go`), give each empty array its own backing allocation before marshaling, so the encoder no longer sees them as the same shared value. Non-empty arrays are left untouched, so genuinely shared/anchored structures still get smart-anchor deduplication.

## Testing

- Added a regression test reproducing the exact issue scenario (three sibling keys each holding an independent empty array), asserting the output round-trips and never contains an anchor-define-and-dereference-in-one-node pattern.
- Ran the full `tpl/transform` package test suite, including `TestRemarshaBillionLaughs`, to confirm the existing DoS-expansion protection is unaffected by this change.
- Ran `gofmt` and `go vet` on the changed files; both clean. I was unable to run `staticcheck` in my environment due to a local Go toolchain version mismatch unrelated to this change, so that check should be verified by CI.

Fixes #14596

### DIFF
--- a/tpl/transform/remarshal.go
+++ b/tpl/transform/remarshal.go
@@ -69,6 +69,18 @@ func applyMarshalTypes(m map[string]any) {
 		switch t := v.(type) {
 		case map[string]any:
 			applyMarshalTypes(t)
+		case []any:
+			if len(t) == 0 {
+				// The YAML decoder returns the same shared instance for
+				// every empty array literal, which makes the smart-anchor
+				// encoder treat them as one value referenced many times
+				// and emit an invalid self-referencing anchor. Give each
+				// one its own backing array (len 0 alone shares a pointer)
+				// to break the false sharing without touching real
+				// (non-empty) anchors, which still need dedup to guard
+				// against billion-laughs-style expansion.
+				m[k] = make([]any, 0, 1)
+			}
 		case float64:
 			i := int64(t)
 			if t == float64(i) {
--- a/tpl/transform/remarshal_test.go
+++ b/tpl/transform/remarshal_test.go
@@ -14,6 +14,7 @@
 package transform_test
 
 import (
+	"regexp"
 	"testing"
 
 	"github.com/gohugoio/hugo/htesting"
@@ -197,6 +198,36 @@ a = "b"
 
 		_, err = ns.Remarshal("json", "asdf")
 		c.Assert(err, qt.Not(qt.IsNil))
+	})
+
+	// Issue 14596
+	c.Run("Repeated empty arrays do not produce invalid YAML anchors", func(c *qt.C) {
+		input := `
+field1:
+  options: []
+  pattern: x1
+field2:
+  options: []
+  pattern: x2
+field3:
+  options: []
+  pattern: x3
+`
+
+		converted, err := ns.Remarshal("yaml", input)
+		c.Assert(err, qt.IsNil)
+
+		// The output must be valid YAML, in particular it must not contain
+		// a node that both defines and dereferences an anchor on the same
+		// line (e.g. "&options *options").
+		roundtripped, err := ns.Remarshal("json", converted)
+		c.Assert(err, qt.IsNil)
+		c.Assert(roundtripped, qt.Contains, `"field1"`)
+		c.Assert(roundtripped, qt.Contains, `"field2"`)
+		c.Assert(roundtripped, qt.Contains, `"field3"`)
+
+		// A node must never both define and dereference an anchor at once.
+		c.Assert(regexp.MustCompile(`&\S+\s+\*\S+`).MatchString(converted), qt.IsFalse)
 	})
 }
 


jmooring:
This should be fixed upstream, which is why the issue has the "upstream" label. There is already an upstream issue and corresponding pull request.

- <https://github.com/goccy/go-yaml/issues/853>
- <https://github.com/goccy/go-yaml/pull/854>

amitmishra11:
Thanks for pointing that out, and sorry for the noise. I should have checked the issue's existing labels and comments before implementing a fix, not just searched for competing PRs. I missed that this already had the Upstream label and an existing comment from bep explaining it's not something you intend to work around in Hugo itself.

Since the real fix is already in progress upstream (goccy/go-yaml#854), I'll close this PR.

PR #15082, opened 2026-07-03
Add markup.rst.syntaxHighlight option

### LINKED ISSUE #5349: reStructuredText uses long token names for syntax highlighting

bep:
Fixes #5349

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,7 +63,7 @@ jobs:
         run: go install github.com/gohugoio/gotmplfmt@latest
       - name: Install docutils
         run: |
-          pip install docutils
+          pip install docutils Pygments
           rst2html --version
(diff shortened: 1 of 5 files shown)


jmooring (commented):
This works as expected, but the purpose of this isn't clear:

```toml
[markup.rst]
  syntaxHighlight = 'short'
```

Perhaps something like this instead:

```toml
[markup.rst.highlight]
  classNaming = 'long'
```

But I don't feel strongly about it.

bep:
>This works as expected, but the purpose of this isn't clear:

Well,

1. It matches the upstream flag (`--syntaxHighlight`)
2. The flat config also sets an expectation for the future: There will not be many RST options.

jmooring:
Works for me. We're going to document this anyway.

PR #15090, opened 2026-07-05
hugolib: Support subdirectory paths in .Render

### LINKED ISSUE #15056: Support subdirectory paths in .Render
Allow `.Render "subdir/myview"` to look up a content view template named `myview` starting from `layouts/{page-path}/subdir/`, walking up the tree as usual. Note that `subdir` is not part of the tree.

This construct worked in earlier versions; we removed it in v0.146.0 with implementation of the new template system.

The purpose of such a construct is to organize content views in a dedicated directory.

Reference: <https://discourse.gohugo.io/t/silent-error-with-render-function-in-v0-163-2-extended-withdeploy/57290/9>

jmooring:
A slash in the `.Render` argument scopes the lookup to a subdirectory. The subdirectory portion may itself contain multiple path segments. For example, `.Render "a/b/c"` resolves a template named `c` inside the directory path `a/b`:
    
    layouts/{page-path}/a/b/c.html
    layouts/{section}/a/b/c.html
    layouts/a/b/c.html
    
This walks up the page path, matching the subdirectory at each level, consistent with how subdirectory shortcodes are resolved.
    
    Closes #15056

### DIFF
--- a/hugolib/page.go
+++ b/hugolib/page.go
@@ -631,7 +631,41 @@ func (ps *pageState) resolveTemplate(layouts ...string) (*tplimpl.TemplInfo, boo
 	dir, d := ps.GetInternalTemplateBasePathAndDescriptor()
 
 	if len(layouts) > 0 {
-		d.LayoutFromUser = layouts[0]
+		layout := layouts[0]
+		if i := strings.LastIndex(layout, "/"); i != -1 {
+			if i == len(layout)-1 {
+				return nil, false, nil
+			}
+			subdir := layout[:i]
+			d.LayoutFromUser = layout[i+1:]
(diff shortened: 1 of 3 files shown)


bep:
I have some fundamental comments to this PR:

* It leaks a lot of template logic out of the `templatestore`
* By doing multiple `LookupPagesLayoutAtPath` it's not obvious to me that the template specificity rules (distance, output format, language ...) are preserved; it may be correct (I don't have those rules in my head), but it's not obvious looking at the tests.

I have pulled the tests into #15091 which I think is a better implementation (but that may be in the eye of the beholder).

jmooring:
Understood. No problem. I'm glad the tests were useful. Thanks for moving this forward.

PR #15102, opened 2026-07-07
tpl/collections: Fix shuffle panic on arrays and pointers to slices

upuddu:
`shuffle`'s type switch accepts arrays, slices and strings (and pointers to those), but the result was built with `reflect.MakeSlice(reflect.TypeOf(l), ...)`, which panics with "reflect.MakeSlice of non-slice type" for anything but a plain slice value - so `{{ shuffle "abc" }}`, an array, or a pointer to a slice crash the build.

Build the result from the indirected value's type so a pointer to a slice works, collect arrays into a slice of their element type, and drop strings from the accepted kinds so they return the regular "can't iterate over" error instead of panicking. Behavior for a plain slice (including named slice types) is unchanged.

### DIFF
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -505,13 +505,21 @@ func (ns *Namespace) Shuffle(l any) (any, error) {
 	}
 
 	switch lv.Kind() {
-	case reflect.Array, reflect.Slice, reflect.String:
+	case reflect.Array, reflect.Slice:
 		// okay
 	default:
 		return nil, errors.New("can't iterate over " + reflect.ValueOf(l).Type().String())
 	}
 
-	shuffled := reflect.MakeSlice(reflect.TypeOf(l), lv.Len(), lv.Len())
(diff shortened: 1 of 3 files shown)


bep:
This needs an integration test (use `hugolib.Test`) that demonstrates a real failure scenario.

upuddu:
Added an integration test (`TestShuffleNonSlice`) using `hugolib.Test`. Without the fix, `{{ shuffle "abc" }}` fails with `reflect.MakeSlice of non-slice type`; with it you get the normal `can't iterate over string` error, and a plain slice still shuffles as before. Thanks for the review!

Awesome195:
Plug-ins JS

bep:
Thanks, but I'm closing this for now. The integration test cover 1 out of 3 of the cases patched. The two others seems to be made up examples that cannot be reproduced in a Hugo site.

I also notices that this PR doesn't reference an issue, which I guess explains this.

PR #15119, opened 2026-07-15
feat: add ignoreUnusedTemplates config option

### LINKED ISSUE #15110: Add an allowlist for intentionally unexecuted templates when using `printUnusedTemplates`
Hugo's `printUnusedTemplates` option is very useful for keeping a site's template tree clean. I would like to request a small companion mechanism to mark some templates as intentionally allowed to remain unexecuted during successful builds.

### Current behavior

`printUnusedTemplates` reports templates whose execution count is zero during the build. This is a good default behavior and I do not think it should change.

However, some templates are intentionally only used on error paths or other rare conditional paths. For example, a site may have a shared partial used to fail fast with consistent diagnostics:

```go-html-template
{{ partial "fail.html" (dict
  "message" "missing required parameter"
  "location" "..."
  "fix" "..."
) }}
```

During a successful production build, this partial is not executed, so `printUnusedTemplates` reports it as unused even though it is intentionally part of the template system.

A site can work around this by calling the partial once with a no-op argument, but that feels like a workaround: it changes template execution only to satisfy the diagnostic check.

## Proposal

Add a config option to exclude specific templates from `printUnusedTemplates` warnings, for example:

```toml
ignoreUnusedTemplates = [
  "_partials/fail.html",
]
```

Glob patterns would also be useful, but exact template names alone would already solve the core problem:

```toml
ignoreUnusedTemplates = [
  "_partials/fail.html",
  "_partials/debug/*.html",
]
```

The important part is that this would not change the meaning of `printUnusedTemplates`: templates would still be considered "used" only when actually executed. The option would only suppress warnings for templates the project explicitly declares as intentionally retained.

## Use cases

This would help with templates that are valid and intentional but may not execute in a successful build:

- shared error/fail-fast partials
- debug-only partials
- optional theme extension points
- templates used only by certain content states, languages, output formats, or build environments
- rare shortcode/render-hook variants that a theme wants to keep available

## Alternatives considered

Changing printUnusedTemplates to count static references as usage would make the warning less precise, and could hide templates that are referenced but never actually reached.

A template-local marker such as a comment directive could also work, but a config allowlist seems simpler and keeps the diagnostic policy in site configuration.

1solomonwakhungu:
## Summary

Adds a new `ignoreUnusedTemplates` config option that suppresses unused template warnings for templates that are intentionally retained but may not execute during a successful build.

## Problem

Hugo's `printUnusedTemplates` option reports templates whose execution count is zero during a build. Some templates are intentionally only used on error paths or rare conditional paths (e.g., shared fail-fast partials, debug-only templates, optional theme extension points). These templates are valid and intentional but appear as "unused" in the warnings.

## Solution

Add an `ignoreUnusedTemplates` config option that accepts a list of glob patterns. Templates matching any pattern are excluded from `printUnusedTemplates` warnings.

The matching uses Go's `path.Match` for glob support. Both the pattern and template path are normalized by trimming leading slashes for consistent matching.

Also updates the `printUnusedTemplates` CLI flag description from "print warnings on unused templates" to "print warnings on potentially unused templates".

## Changes

- `config/allconfig/allconfig.go`: Added `IgnoreUnusedTemplates` field to `RootConfig`, added to `cloneForLang`
- `hugolib/hugo_sites_build.go`: Added `isIgnoredUnusedTemplate` function, integrated into `printUnusedTemplatesOnce`
- `commands/commandeer.go`: Updated CLI flag description
- `hugolib/is_ignored_unused_template_test.go` (new): 7 unit tests for glob matching
- `tpl/tplimpl/templatestore_integration_test.go`: Integration test verifying ignore list doesn't affect detection

## Test Plan

- `go build ./...` compiles
- `go test ./hugolib/... -run TestIsIgnoredUnusedTemplate -v` passes (7 cases)
- `go test ./tpl/tplimpl/... -run TestPrintUnusedTemplatesIgnore -v` passes
- `go test ./tpl/tplimpl/... -run TestPrintUnusedTemplates -v` passes (no regression)

Fixes #15110

### DIFF
--- a/commands/commandeer.go
+++ b/commands/commandeer.go
@@ -628,7 +628,7 @@ func applyLocalFlagsBuild(cmd *cobra.Command, r *rootCommand) {
 	cmd.Flags().BoolP("noChmod", "", false, "don't sync permission mode of files")
 	cmd.Flags().BoolP("printI18nWarnings", "", false, "print missing translations")
 	cmd.Flags().BoolP("printPathWarnings", "", false, "print warnings on duplicate target paths etc.")
-	cmd.Flags().BoolP("printUnusedTemplates", "", false, "print warnings on unused templates.")
+	cmd.Flags().BoolP("printUnusedTemplates", "", false, "print warnings on potentially unused templates.")
 	cmd.Flags().StringVarP(&r.cpuprofile, "profile-cpu", "", "", "write cpu profile to `file`")
 	cmd.Flags().StringVarP(&r.memprofile, "profile-mem", "", "", "write memory profile to `file`")
 	cmd.Flags().BoolVarP(&r.printm, "printMemoryUsage", "", false, "print memory usage to screen at intervals")
--- a/config/allconfig/allconfig.go
+++ b/config/allconfig/allconfig.go
@@ -247,6 +247,7 @@ func (c Config) cloneForLang() *Config {
 	x.MainSections = copyStringSlice(x.MainSections)
 	x.IgnoreLogs = copyStringSlice(x.IgnoreLogs)
 	x.IgnoreFiles = copyStringSlice(x.IgnoreFiles)
+	x.IgnoreUnusedTemplates = copyStringSlice(x.IgnoreUnusedTemplates)
 	x.Theme = copyStringSlice(x.Theme)
 
 	// Collapse all static dirs to one.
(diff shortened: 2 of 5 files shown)


1solomonwakhungu:
Thanks! Amended the commit with `--signoff` so the Signed-off-by line is now included. The CLA should recheck automatically.

bep:

1solomonwakhungu:
Thanks for the pointer. This PR implements the feature requested in #15110, which is currently labeled "Proposal". Would it be appropriate to add the "enhancement" label to that issue, or should I open a separate issue with the "enhancement" label specifically for this change? Happy to adjust either way.

PR #15166, opened 2026-08-05
hugolib: Preserve inline manual summary formatting

### LINKED ISSUE #11440: Manual summary breaks Markdown formatting
Manual summary breaks Markdown formatting by [inserting](https://github.com/gohugoio/hugo/blob/master/hugolib/page.go#L762C29-L762C29) [extra newlines](https://github.com/gohugoio/hugo/blob/master/hugolib/page__content.go#L27). This causes the Summary and rest of content to be placed in separate `<p>` elements even if they should not have been. This treats the `<!--more-->` separator differently from any other `<!-- -->` comment and does not appear to be documented.

<pre>
$ hugo version
hugo v0.118.2-da7983ac4b94d97d776d7c2405040de97e95c03d+extended darwin/amd64 BuildDate=2023-08-31T11:23:51Z VendorInfo=brew
</pre>

Yes

ANSHUL-REAL:
Inline `<!--more-->` dividers were always replaced with a blank-line-wrapped marker, which forced Markdown to split a paragraph at the summary boundary.

This keeps the existing behavior for standalone dividers while preserving inline dividers. The summary extractor now closes the summary markup and reopens it for the remaining content, so `.Summary`, `.ContentWithoutSummary`, and `.Content` all remain valid.

Tests:

- `go test ./resources/page -run 'TestExtractSummaryFromHTMLWithDivider|TestExpandDivider' -count=1`
- `go test ./hugolib -run 'TestSummaryManualSplit(PreservesParagraph)?$' -count=1`

Fixes #11440

### DIFF
--- a/hugolib/page__content.go
+++ b/hugolib/page__content.go
@@ -14,6 +14,7 @@
 package hugolib
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -55,8 +56,8 @@ const (
 )
 
 var (
-	internalSummaryDividerPreString = "\n\n" + internalSummaryDividerBase + "\n\n"
-	internalSummaryDividerPre       = []byte(internalSummaryDividerPreString)
+	internalSummaryDivider    = []byte(internalSummaryDividerBase)
+	internalSummaryDividerPre = []byte("\n\n" + internalSummaryDividerBase + "\n\n")
 )
 
 type pageContentReplacement struct {
@@ -326,7 +327,11 @@ Loop:
 
 			// The content may be rendered by Goldmark or similar,
 			// and we need to track the summary.
-			pi.AddReplacement(internalSummaryDividerPre, it)
+			divider := internalSummaryDividerPre
+			if !bytes.ContainsRune(it.Val(source), '\n') {
+				divider = internalSummaryDivider
+			}
+			pi.AddReplacement(divider, it)
 
 		// Handle shortcode
 		case it.IsLeftShortcodeDelim():
--- a/hugolib/page__content.go
+++ b/hugolib/page__content.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strings"
 	"sync/atomic"
+	"unicode"
 	"unicode/utf8"
 
 	maps0 "maps"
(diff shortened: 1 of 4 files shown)


bep:
@ANSHUL-REAL there are test failures.

ANSHUL-REAL:
Thanks for pointing that out. I've fixed the failing test and pushed the update.

ANSHUL-REAL:
Tracked down the remaining failure. The inline divider check was also treating `aaa <!--more--> bbb` as inline, which collapsed both sides into a single paragraph. I updated it so inline handling only applies when the divider directly touches non-whitespace text on both sides; spaced and shortcode-adjacent dividers keep the existing paragraph boundary. The failing pages-from-data test and the affected summary tests now pass.

bep:
Thanks for this. On the surface this looks good, my instinct is that this puts too much (fragile?) parsing logic into the HTML layer. I asked Claude code to review it, and he found several issues (the conclusion below). This needs to be 

---

## PR 15166 review — inline `<!--more-->` handling

  **Verdict:** not mergeable as-is. Six confirmed correctness regressions (all reproduced end-to-end), one of which crashes whole-site builds.

  ### Correctness (confirmed)

  - **Build panic** — `page_markup.go:386`: when `closingTagsAtEnd` doesn't match, `endMarkup` stays zero and `s[divider.High:0]` panics. `# Head<!--more-->rest`
  fails the entire build; fine on master.
  - **Non-`<p>` blocks broken** — the extractor only understands `<p>`/`<div>`. A divider in a heading yields `<h2 id="headhugomore42rest">Head</p>` and an
  **empty** `ContentWithoutSummary` (rest of page silently dropped).
  - **Scan crosses block boundaries** — the backward start-tag scan has no stop condition, so a divider in a list adopts an unrelated earlier paragraph's `<p>`,
  producing invalid HTML in both Summary and remainder.
  - **Orphan closing tags** — `abc<!--more-->def **bold**` renders Summary as `<p>abc</strong></p>` (text dropped, its closing tags kept).
  - **Spurious re-opened inline tags** — start-tag expansion swallows any tag run, so `*em* text<!--more-->rest` leaks an unclosed `<em>` that italicizes the rest
  of the section.
  - **CommonMark flanking broken** — the bare `HUGOMORE42` placeholder glues onto adjacent text, so `_summary_<!--more-->content` renders literal underscores
  instead of `<em>`.

  ### Design/cleanup

  - Shortcode adjacency is sniffed via raw `}}`/`{{` bytes instead of parser tokens — misfires on literal text containing those sequences.
  - The new backward scan duplicates (and has already drifted from) the existing one a few lines above; it's also quadratic (full-prefix regexp per `>`).
  - New test lacks the `// See issue 11440.` doc comment per AGENTS.md.

  ### Root cause in one line

  The bare divider is inserted into the token stream with no awareness of markdown context, and the HTML extractor downstream assumes it always lands inside a
  well-formed `<p>`/`<div>` — every failure above is a variant of that assumption breaking. A real fix likely needs to work at the parser/renderer level rather
  than post-hoc HTML string surgery.

PR #15205, opened 2026-08-14
hugolib: Keep nested index.md as a resource on rebuild

### LINKED ISSUE #13961: erroneously building resources

arronKler:
A full content walk classifies files below a leaf bundle as resources. A server rebuild can start at the nested directory, so `content/a/b/index.md` was parsed as a new leaf from its filename and published as `/a/b/`.

Look up the owning leaf in the page tree and keep those nested bundle files as resources on rebuild, including when they are inserted.

Fixes #13961

This PR was written with AI assistance; I reviewed the rebuild classification and the regression test.

### DIFF
--- a/hugolib/content_map.go
+++ b/hugolib/content_map.go
@@ -300,6 +300,15 @@ func (m *pageMap) AddFi(fi hugofs.FileMetaInfo, buildConfig *BuildCfg) (pageSour
 		}
 
 	default:
+		// Partial rebuilds may start below a leaf; keep those files as resources.
+		if m.leafBundleOwner(pi) != nil {
+			paths.ModifyPathBundleTypeResource(pi)
+			if err := insertResource(fi); err != nil {
+				addErr = err
+			}
+			return
+		}
+
 		m.s.Log.Trace(logg.StringFunc(
 			func() string {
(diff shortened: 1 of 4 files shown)


arronKler:
I have read the CLA Document and I hereby sign the CLA

PR #15231, opened 2026-08-21
resources: Fix permalinks for multihost resources

### LINKED ISSUE #12560: Resource Permalink returns incorrect base URL with multihost site
This isn't anything new. I am able to reproduce with v0.54.0 and later.

```go
func TestFoo(t *testing.T) {
	t.Parallel()

	files := `
-- hugo.toml --
[languages.en]
baseURL = 'https://en.example.org/'
weight = 1
[languages.de]
baseURL = 'https://de.example.org/'
weight = 2
-- layouts/index.html --
{{ with resources.Get "main.css" }}{{ .Permalink }}{{ end }}
-- assets/main.css --
body {color: red;}
`

	b := hugolib.Test(t, files)

	b.AssertFileContent("public/en/main.css", "body {color: red;}")
	b.AssertFileContent("public/en/index.html", "https://en.example.org/main.css")
	b.AssertFileContent("public/de/main.css", "body {color: red;}")
	b.AssertFileContent("public/de/index.html", "https://de.example.org/main.css") // fails
}
```

The last assertion fails; the file is identical to public/en/index.html, which means it's pointing to the wrong site.

Reference: <https://discourse.gohugo.io/t/multihost-trouble/50103>

The work around is to use `RelPermalink` instead:

```text
{{ with resources.Get "main.css" }}
  {{ .RelPermalink }}
{{ end }}
```

Soundcreates:
Fixes #12560.

## Summary

Shared resource caches are used by each language site in a multihost build, but their keys did not include the site's base URL. As a result, a resource or transformed resource cached while rendering one language could return an absolute permalink for another language's host.

Namespace the per-site resource cache wrapper by base URL, including local, remote, resource-list, and transformation caches. Keep the underlying cache partitions shared while preventing cross-host resource objects from being reused.

Added a regression assertion to the existing multihost test covering a fingerprinted asset.

## Testing

- `go test ./hugolib -run '^TestMultihost$' -count=1`
- `go test ./resources -run '^(TestNewResource|TestTransformCached|TestTransform)$' -count=1`
- `git diff --check`

The broader resource test suite also contains existing OS-filesystem tests that cannot run in this environment because writes under `/private/Users` are denied.

## AI assistance disclosure

This PR was written with substantial assistance from OpenAI Codex. The implementation, test change, and reported test results were manually reviewed.


bep:
While this may be give the desired end result, but I'm pretty sure it blows the value of the resource cache out with the bath water. Consider the following example:

```
{{ $css := resources.Get "main.css" | css.TailwindCSS | css.PostCSS }}
```

The above is relatively slow, say 1 second. The resource cache prevents the above from executing more than once, so it's liveable.

With the fix in this PR, and given you have a project with 10 languages, 2 roles and 5 versions, the above would run 100 times, which is ... 100 seconds.

I don't have the details of #12560 in my head, but:

* `$css.Permalink` above is correct even in a multihost site, but it may point to a different host, which is unfortunate (e.g. for CORS reasons).
* But we do copy the resources to each host, so you can safely `$css.RelPermalink` and it would work fine.

I'm not saying we shouldn't fix #12560, but it needs to be fixed in a simpler way that does not hurt the common setup.

Soundcreates:
Addressed the cache concern.

The transformation cache is shared again, so expensive transforms such as `css.TailwindCSS | css.PostCSS` are not rerun for every language/role/version. Cached resource adapters are rebound to the active site spec before use, and permalink generation uses that active spec while preserving the shared transformed resource.

Validation:
- `go test ./resources ./hugolib -run \^(TestMultihost|TestNewResource|TestTransformCached|TestTransform)$' -count=1` passes.\n- `./check.sh ./resources/...` completed staticcheck, but the package test phase hits unrelated OS-filesystem failures in this environment (`mkdir /private/Users: permission denied`).

PR #15232, opened 2026-08-21
metrics: Improve template metrics duration formatting

Shiwang0-0:
Improves the formatting of template metrics durations.

- Formats durations with two decimal places.
- Selects the appropriate unit independently for each duration (`s`, `ms`, `µs`, or `ns`).
- Adds spacing between the duration value and unit so the unit suffixes are aligned.
- Updates the duration-normalization regex in `TestIncludeCacheHints` to account for the new spacing.

See #15027

### DIFF
--- a/metrics/metrics_test.go
+++ b/metrics/metrics_test.go
@@ -17,6 +17,7 @@ import (
 	"html/template"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/gohugoio/hugo/resources/page"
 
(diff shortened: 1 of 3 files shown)


metrics/metrics_test.go
```diff
@@ -66,3 +67,22 @@ func BenchmarkHowSimilar(b *testing.B) {
...
+		{4*time.Second + 342*time.Millisecond, "4.34  s"}, // additional spacing between value and unit
+		{170*time.Millisecond + 289*time.Microsecond, "170.29 ms"},
+		{16*time.Microsecond + 90*time.Nanosecond, "16.09 µs"},
+		{147 * time.Nanosecond, "147.00 ns"},
+	}
+
+	for _, tt := range tests {
+		got := formatDuration(tt.duration)
+		if got != tt.want {
+			t.Errorf("formatDuration(%v) = %q, want %q", tt.duration, got, tt.want)
```
bep: This is a nit pick, all other tests in Hugo has assertions on the form `c.Assert(got, qt.Equals, want)`. Other than that, this PR looks great.

PR #15238, opened 2026-08-22
related: Add tokenize option for index values

### LINKED ISSUE #7515: Related content: Partial-text indexing
When specifying the indices for related content, it would be useful to specify that some of them were split by word, so that their text could be matched partially. This would allow the title to be used for more in-depth indexing, rather than requiring the user to specify tags / keywords.

This would likely increase memory usage and decrease performance, so should definitely be opt-in!

I suspect common words (a, an, the etc) might want to be ignored? Although doing that in a sane, i18n-friendly manner might be a challenge!

### LINKED ISSUE #15199: related: Add option to tokenize headings
See <https://discourse.gohugo.io/t/related-posts-based-on-heading/57467>.

I suggest using the name `tokenize`, documented as follows:

> (`bool`) Whether to split multi-word values into individual keywords for indexing and matching. Currently applies to heading titles in indices of type `fragments`; non-heading fragments such as definition terms continue to use exact-identifier matching regardless of this setting. This increases index size in proportion to the average number of words per heading; use `cardinalityThreshold` to prune common words and reduce build time. Default is `false`.

jmooring:
Closes #7515
Closes #15199

### DIFF
--- a/related/inverted_index.go
+++ b/related/inverted_index.go
@@ -18,12 +18,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"html"
 	"maps"
 	"math"
 	"sort"
 	"strings"
 	"sync"
 	"time"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/gohugoio/hugo/common/collections"
 	"github.com/gohugoio/hugo/common/hmaps"
@@ -122,9 +125,15 @@ type IndexConfig struct {
 	// used in more than 50% of the documents in the index.
 	CardinalityThreshold int
 
-	// Will lower case all string values in and queries tothis index.
+	// Will lower case all string values in and queries to this index.
 	// May get better accurate results, but at a slight performance cost.
 	ToLower bool
+
+	// Whether to split multi-word values into individual keywords.
+	Tokenize bool
+
+	// Minimum token length in Unicode code points when tokenizing.
+	MinTokenLength int
 }
 
 // Document is the interface an indexable document in Hugo must fulfill.
@@ -220,8 +229,24 @@ func (idx *InvertedIndex) Add(ctx context.Context, docs ...Document) error {
 
 			if config.Type == TypeFragments {
 				if fp, ok := doc.(FragmentProvider); ok {
-					for _, fragment := range fp.Fragments(ctx).Identifiers {
+					frags := fp.Fragments(ctx)
+					seenKW := make(map[Keyword]bool)
+					for _, fragment := range frags.Identifiers {
 						added = true
+						if config.Tokenize {
+							if h, found := frags.HeadingsMap[fragment]; found {
+								for _, word := range splitWords(h.Title) {
+									if !config.shouldSkipToken(word) {
+										kw := config.stringToKeyword(word)
+										if !seenKW[kw] {
+											seenKW[kw] = true
+											setm[kw] = append(setm[kw], doc)
+										}
+									}
+								}
+								continue
+							}
+						}
 						setm[FragmentKeyword(fragment)] = append(setm[FragmentKeyword(fragment)], doc)
 					}
 				}
@@ -377,7 +402,23 @@ func (idx *InvertedIndex) Search(ctx context.Context, opts SearchOpts) ([]Docume
 			}
 			if opts.Document != nil {
 				if fp, ok := opts.Document.(FragmentProvider); ok {
-					for _, fragment := range fp.Fragments(ctx).Identifiers {
+					frags := fp.Fragments(ctx)
+					seenKW := make(map[Keyword]bool)
+					for _, fragment := range frags.Identifiers {
+						if cfg.Tokenize {
+							if h, found := frags.HeadingsMap[fragment]; found {
+								for _, word := range splitWords(h.Title) {
+									if !cfg.shouldSkipToken(word) {
+										kw := cfg.stringToKeyword(word)
+										if !seenKW[kw] {
+											seenKW[kw] = true
+											keywords = append(keywords, kw)
+										}
+									}
+								}
+								continue
+							}
+						}
 						keywords = append(keywords, FragmentKeyword(fragment))
 					}
 				}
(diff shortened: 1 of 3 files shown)


bep:
@jmooring I know this is a draft, but I have been thinking a little on the scope of this option, as to limit to headings only:

1. The setting would be default false, so it would not break anything.
2. Allowing it for all would provide the most flexibility, and I suspect there will be people (probably including me) that would be a little miffed that this option wasn't available for `title`.
3. There may be footguns in here, but that's ultimately the user's fault.

So, if it's not too much hassle, I suggest we just allow this for fields.

jmooring:
Revised documentation:

> `tokenize`
>
> (`bool`) Whether to tokenize string values by splitting them into individual words before indexing, matching, and filtering. For indices of type `fragments`, non-heading fragments such as definition terms are indexed by their exact identifier regardless of this setting. Tokenizing increases index size in proportion to the number of words per value; consider setting `cardinalityThreshold` to remove high-frequency words from the index. Default is `false`.

jmooring:
Run benchmarks:

```text
git checkout HEAD~1
go test -run='^$' -bench=BenchmarkRelatedNewIndex -benchmem -count=20 ./related/... > /tmp/before.txt
go test -run='^$' -bench=BenchmarkRelatedMatchesIn -benchmem -count=20 ./related/... >> /tmp/before.txt
go test -run='^$' -bench=BenchmarkRelatedSite -benchmem -count=20 ./related/... >> /tmp/before.txt

git checkout -
go test -run='^$' -bench=BenchmarkRelatedNewIndex -benchmem -count=20 ./related/... > /tmp/after.txt
go test -run='^$' -bench=BenchmarkRelatedMatchesIn -benchmem -count=20 ./related/... >> /tmp/after.txt
go test -run='^$' -bench=BenchmarkRelatedSite -benchmem -count=20 ./related/... >> /tmp/after.txt

benchstat /tmp/before.txt /tmp/after.txt
```

Result:

```text
goos: linux
goarch: amd64
pkg: github.com/gohugoio/hugo/related
cpu: 13th Gen Intel(R) Core(TM) i7-1355U
                          │ /tmp/before.txt │           /tmp/after.txt           │
                          │     sec/op      │   sec/op     vs base               │
RelatedNewIndex/singles-8       22.30µ ± 1%   24.17µ ± 6%  +8.38% (p=0.000 n=20)
RelatedNewIndex/all-8           21.57µ ± 2%   22.33µ ± 1%  +3.54% (p=0.000 n=20)
RelatedMatchesIn-8              38.10µ ± 3%   37.86µ ± 4%       ~ (p=0.461 n=20)
RelatedSite-8                   23.19m ± 6%   23.39m ± 4%       ~ (p=0.602 n=20)
geomean                         143.6µ        147.9µ       +2.98%

                          │ /tmp/before.txt │           /tmp/after.txt            │
                          │      B/op       │     B/op      vs base               │
RelatedNewIndex/singles-8      20.39Ki ± 0%   19.77Ki ± 0%  -3.07% (p=0.000 n=20)
RelatedNewIndex/all-8          22.14Ki ± 0%   21.52Ki ± 0%  -2.82% (p=0.000 n=20)
RelatedMatchesIn-8             12.06Ki ± 2%   12.09Ki ± 2%  +0.28% (p=0.021 n=20)
RelatedSite-8                  37.94Mi ± 0%   37.94Mi ± 0%       ~ (p=0.142 n=20)
geomean                        120.6Ki        118.9Ki       -1.41%

                          │ /tmp/before.txt │           /tmp/after.txt           │
                          │    allocs/op    │  allocs/op   vs base               │
RelatedNewIndex/singles-8        198.0 ± 0%    200.0 ± 0%  +1.01% (p=0.000 n=20)
RelatedNewIndex/all-8            199.0 ± 0%    201.0 ± 0%  +1.01% (p=0.000 n=20)
RelatedMatchesIn-8               7.000 ± 0%    7.000 ± 0%       ~ (p=1.000 n=20)
RelatedSite-8                   166.2k ± 0%   166.2k ± 0%       ~ (p=0.482 n=20)
geomean                          462.7         465.0       +0.50%
```

jmooring:
Here's a practical example that I'm not thrilled about:

```text
git clone --single-branch -b hugo-github-issue-15238 https://github.com/jmooring/hugo-testing hugo-github-issue-15238
cd hugo-github-issue-15238
hugo server
```

This is the relevant portion of the project configuration:

```toml
[related]
  includeNewer = true
  threshold = 80
  [[related.indices]]
    toLower = true
    name = 'title'
    tokenize = true
    weight = 100
    cardinalityThreshold = 60
```

Even with `cardinalityThreshold = 60`, common stop words like "in" and "the" are not filtered from the `title` index and continue to skew the related content results. Possible approaches to address this:

1. Implement a per-index, language-agnostic minimum token length, with a config structure like:

  ```toml
  [[related.indices]]
    toLower = true
    name    = "title"
    weight  = 100
    [related.indices.tokenize]
      enable        = true
      minimumLength = 4
  ```
  
2. Implement a per-language stop word list in the project configuration... yuck.
  
3. Import an existing stop word package such as [`toadharvard/stopwords-iso`](https://github.com/toadharvard/stopwords-iso) or [`bbalet/stopwords`](https://github.com/bbalet/stopwords), **neither of which are actively maintained**.
  
4. Roll our own stop word filter, embedding [`stopwords-iso`](https://github.com/stopwords-iso/stopwords-iso). This is expensive.

So that leaves option 1 and 4 (unless you can think of something else). I'd be inclined to go with Option 1 for now.

jmooring:
If we go with Option 1 above, perhaps a flat config is better:

```toml
[[related.indices]]
  toLower = true
  name    = "title"
  weight  = 100
  tokenize      = true
  minTokenLength = 4
```

When both `minTokenLength` and `cardinalityThreshold` are set, `minTokenLength` runs first, so short tokens never enter the index. Then `cardinalityThreshold` removes tokens that appear too frequently across the content. The two complement each other: `minTokenLength` handles short function words regardless of how often they appear, while `cardinalityThreshold` handles longer but overused words like "tutorial" or "introduction".

I ran a quick test of a `minTokenLength` implementation and am pleased with the result.

jmooring:
I'm happy with how this behaves. New options:

> `tokenize`
>
> Whether to tokenize string values by splitting them into individual words before indexing, matching, and filtering. For indices of type `fragments`, non-heading fragments such as definition terms are indexed by their exact identifier regardless of this setting. Tokenizing increases index size in proportion to the number of words per value; consider setting `minTokenLength` to exclude short tokens and `cardinalityThreshold` to remove high-frequency words from the index. Default is `false`.

> `minTokenLength`
>
> When `tokenize` is `true`, the minimum token length in Unicode code points. Tokens shorter than this value are not indexed. Use this to exclude short function words such as "a", "in", and "the" regardless of how often they appear. `minTokenLength` is applied before `cardinalityThreshold`, so excluded tokens never enter the index. Default is `0` (no minimum).

Take it for a spin:

```text
git clone --single-branch -b hugo-github-issue-15238 https://github.com/jmooring/hugo-testing hugo-github-issue-15238
cd hugo-github-issue-15238
hugo server
```

related/inverted_index.go
```diff
@@ -220,8 +229,20 @@ func (idx *InvertedIndex) Add(ctx context.Context, docs ...Document) error {
...
-					for _, fragment := range fp.Fragments(ctx).Identifiers {
+					frags := fp.Fragments(ctx)
+					for _, fragment := range frags.Identifiers {
 						added = true
+						if config.Tokenize {
+							if h, found := frags.HeadingsMap[fragment]; found {
+								for _, word := range splitWords(h.Title) {
+									if !config.shouldSkipToken(word) {
+										kw := config.stringToKeyword(word)
+										setm[kw] = append(setm[kw], doc)
```
bep: I think we need to do some deduplication here on both the indexing and query side. 

Config: a `fragments` index with `tokenize = true` and a `keywords` index with weight 10. 

  doc1:

  ```md
  ## I love apples
  ```

  doc2:

  ```md
  ## I love apples, apples and more apples
  ## I love apples, apples and more apples
  ## I love apples, apples and more apples
  ```

  doc3:

  ```md
  ## I love apples, apples and more apples
  ## I love apples, apples and more apples
  ## I love apples, apples and more apples
  ```

If we search for related to `doc2`:

  * **doc3**: 9 × 9 = 81, weight 810
  * **doc1**: 9 × 1 = 9, weight 90

So, doc2 and doc3's importance gets bloated from repetition. This gets especially visible in multiindex settings (e.g. `fragments` and `keywords`) where matches can win (and gets pulled to the top of the sort order) purely by repetition.

related/inverted_index.go
```diff
@@ -618,6 +698,28 @@ type Keyword interface {
 	String() string
 }
 
+// splitWords splits s into normalized words, stripping HTML entities and
+// leading/trailing punctuation from each word.
+func splitWords(title string) []string {
+	// Fast path: no whitespace means at most one token.
+	if strings.IndexFunc(title, unicode.IsSpace) == -1 {
```
bep: Considering the html.UnescapeString below we should also check for `&` here to catch `Apples&nbsp;and&nbsp;oranges!`.

related/inverted_index.go
```diff
@@ -618,6 +698,28 @@ type Keyword interface {
 	String() string
 }
 
+// splitWords splits s into normalized words, stripping HTML entities and
+// leading/trailing punctuation from each word.
+func splitWords(title string) []string {
```
bep, line 712: I think `splitWords` needs a good old fashioned unit test. I looks correct/sensible, but theres a lot of trim magic going on.

bep:
@jmooring the code and the options looks good; I will take it for a spin tomorrow.

PR #15239, opened 2026-08-23
fix(hstore): hold read lock for the full GetSortedMapValues operation

### LINKED ISSUE #15237: GetSortedMapValues reads the map after releasing the read lock, so a concurrent SetInMap aborts the build

<pre>
$ hugo version
hugo v0.164.0-ce2470e7012b5ab5fc4e10ebe4027e9f8d9e00dc+extended windows/amd64 BuildDate=2026-07-06T16:39:30Z VendorInfo=gohugoio
</pre>

Yes. `common/hstore/scratch.go` is byte-identical at `v0.164.0`, at `v0.165.0` and on `master`, and the line numbers cited below are the same in all three.

---

`Scratch.GetSortedMapValues` holds the read lock only long enough to fetch the map header out of `c.values`. It releases the lock and then iterates and indexes that map:

```go
// common/hstore/scratch.go @ v0.165.0
func (c *Scratch) GetSortedMapValues(key string) any {
	c.mu.RLock()

	if c.values[key] == nil {
		c.mu.RUnlock()
		return nil
	}

	unsortedMap := c.values[key].(map[string]any)
	c.mu.RUnlock()                                // L141: lock released here
	var keys []string
	for mapKey := range unsortedMap {             // L143: map iteration, unlocked
		keys = append(keys, mapKey)
	}

	sort.Strings(keys)

	sortedArray := make([]any, len(unsortedMap))  // L149: map read, unlocked
	for i, mapKey := range keys {
		sortedArray[i] = unsortedMap[mapKey]      // L151: map read, unlocked
	}

	return sortedArray
}
```

`SetInMap` writes into that same map while holding the write lock. A mutex protects a map only when both sides hold it during the access, so `GetSortedMapValues` running concurrently with `SetInMap` is an unsynchronized read against a write. The Go runtime treats that as fatal and aborts the process; it is not a race a program can be built to tolerate.

Every other method on this type reads or writes `c.values` under the lock. This one is the outlier: it takes the lock, then does its actual work after releasing it.

This is reachable from ordinary templates. `hugo.Store`, `site.Store`, `PAGE.Store` and `SHORTCODE.Store` all return `*hstore.Scratch`, `hugo.Store` is one instance shared across the whole build, and `renderPages` in `hugolib/site_render.go` renders pages concurrently over `config.GetNumWorkerMultiplier()` workers. So a site that collects into a map with `SetInMap` from a partial called per page, and enumerates it with `GetSortedMapValues` from another template, can abort mid-build. `SetInMap` and `GetSortedMapValues` are documented together as a pair on both `hugo.Store` and `site.Store` — the two build-shared scopes — and no Store documentation page mentions concurrency at all, so nothing warns a template author away from that shape.

When it does abort, the build stops with a runtime panic rather than a Hugo error, no finished site is produced, and the panic names no template. The same commit will usually build cleanly on the next run, so it reads as flaky CI rather than as a defect.

The sibling `Values()` carries a comment warning that using it on a shared `Scratch` "will lead to concurrency issues". `GetSortedMapValues` carries no such warning, which is part of why it reads as the safe way to get a map back out of a store.

I have not built a Hugo-site reproduction, because at the site level this is nondeterministic by nature: it needs a shared store written during concurrent page rendering and read while that is still in flight, and when it fires the panic points at no template. The Go-level reproduction below pins the same two calls deterministically.

### Reproduction

This uses only the exported API of `common/hstore`; nothing touches the map directly.

`go.mod`:

```
module hstorerepro

go 1.26.0

require github.com/gohugoio/hugo v0.164.0
```

Run `go mod tidy` after creating it; that adds nine indirect requirements and the `go.sum` entries.

`repro_test.go`:

```go
package hstorerepro

import (
	"sync"
	"testing"

	"github.com/gohugoio/hugo/common/hstore"
)

// One goroutine enumerates the map through the store's own accessor while
// another writes to it through the store's own mutator. Both use only the
// exported API; neither ever touches the map directly.
func TestGetSortedMapValuesRacesSetInMap(t *testing.T) {
	s := hstore.NewScratch()
	s.SetInMap("k", "seed", "seed")

	var wg sync.WaitGroup
	wg.Add(2)

	go func() {
		defer wg.Done()
		for i := 0; i < 100000; i++ {
			s.SetInMap("k", string(rune('a'+i%26)), i)
		}
	}()

	go func() {
		defer wg.Done()
		for i := 0; i < 100000; i++ {
			_ = s.GetSortedMapValues("k")
		}
	}()

	wg.Wait()
}
```

`go test -count=1 .` fails on every run here, with go1.26.2 on windows/amd64. It lands on either unlocked access. Both transcripts below are complete single runs, with only the module-cache and temporary-directory prefixes shortened to `...`:

```
fatal error: concurrent map read and map write

goroutine 10 [running]:
internal/runtime/maps.fatal({0x7ff6230b07fc?, 0x30?})
	C:/Program Files/Go/src/runtime/panic.go:1181 +0x18
github.com/gohugoio/hugo/common/hstore.(*Scratch).GetSortedMapValues(0x4ce82e38820, {0x7ff6230bd980, 0x1})
	.../hugo@v0.164.0/common/hstore/scratch.go:151 +0x30d
hstorerepro.TestGetSortedMapValuesRacesSetInMap.func2()
	.../hstore-repro/repro_test.go:30 +0x65
created by hstorerepro.TestGetSortedMapValuesRacesSetInMap in goroutine 8
	.../hstore-repro/repro_test.go:27 +0x135
FAIL	hstorerepro	0.595s
FAIL
```

```
fatal error: concurrent map iteration and map write

goroutine 10 [running]:
internal/runtime/maps.fatal({0x7ff72f4523b8?, 0x7ff72f264301?})
	C:/Program Files/Go/src/runtime/panic.go:1181 +0x18
internal/runtime/maps.(*Iter).Next(0xf18943adf18?)
	C:/Program Files/Go/src/internal/runtime/maps/table.go:819 +0x86
github.com/gohugoio/hugo/common/hstore.(*Scratch).GetSortedMapValues(0xf18942ba820, {0x7ff72f45d980, 0x1})
	.../hugo@v0.164.0/common/hstore/scratch.go:143 +0x17f
hstorerepro.TestGetSortedMapValuesRacesSetInMap.func2()
	.../hstore-repro/repro_test.go:30 +0x65
created by hstorerepro.TestGetSortedMapValuesRacesSetInMap in goroutine 8
	.../hstore-repro/repro_test.go:27 +0x135
FAIL	hstorerepro	0.578s
FAIL
```

Line 143 is the `range` over the map and line 151 is the index into it, so both of the accesses that can abort were reached. The `len(unsortedMap)` at line 149 is a third unlocked read; it does not trigger the runtime check.

### Suggested fix

Hold the read lock across the whole function, so the iteration and the indexing happen under it:

```go
func (c *Scratch) GetSortedMapValues(key string) any {
	c.mu.RLock()
	defer c.mu.RUnlock()

	if c.values[key] == nil {
		return nil
	}
	// ... unchanged
}
```

This adds no nested locking and no re-entrancy, and the function already returns a freshly allocated slice. `sort.Strings` then runs on a local slice while the read lock is held, which blocks writers for that span; if that is unwanted, copying the map into the slice under the lock and sorting after releasing it gives the same guarantee without it.

This is not a request to make `Scratch` concurrency-safe in general — it already is, by design and by every other method on the type. It is one function that acquires the right lock and then does its work outside it.

A short comment on the method saying the read is safe against a concurrent `SetInMap` would help too: the asymmetry with `Values()`, which does warn, is what makes this one look safe.

Note that `Get` returns the stored value itself, so when that value is a map the caller holds the live map and any read of it has the same exposure. I have kept this report to `GetSortedMapValues` because that one is internal to the type and fixable without changing any signature.

volcanicll:
Fixes #15237

## Summary

`Scratch.GetSortedMapValues` in `common/hstore/scratch.go` released the `RWMutex` after extracting the map header, then iterated and indexed the map without holding the lock. A concurrent `SetInMap` writing into the same map caused a fatal Go runtime abort (unsynchronized map read against a write).

Every other method on this type holds the mutex during its full operation. This was the only outlier.

## Fix

Replace the early `c.mu.RUnlock()` calls with `defer c.mu.RUnlock()` so the read lock covers all map accesses:

```diff
 func (c *Scratch) GetSortedMapValues(key string) any {
 	c.mu.RLock()
+	defer c.mu.RUnlock()

 	if c.values[key] == nil {
-		c.mu.RUnlock()
 		return nil
 	}

 	unsortedMap := c.values[key].(map[string]any)
-	c.mu.RUnlock()
```

No behavior change for single-threaded callers.

## Test

Added `TestScratchGetSortedMapValuesConcurrentWithSetInMap`: 100 goroutines calling `SetInMap` and 100 calling `GetSortedMapValues` concurrently on the same Scratch.

**On the old code**, `go test -race -count=1` immediately reports:
```
WARNING: DATA RACE
Write at 0x... by goroutine ...
--- FAIL: TestScratchGetSortedMapValuesConcurrentWithSetInMap
    race detected during execution of test
FAIL
```

**On the fixed code**: `go test -race -count=5 ./common/hstore/` passes all 5 rounds. All existing tests also pass.

### DIFF
--- a/common/hstore/scratch.go
+++ b/common/hstore/scratch.go
@@ -131,14 +131,13 @@ func (c *Scratch) DeleteInMap(key string, mapKey string) string {
 // GetSortedMapValues returns a sorted map previously filled with SetInMap.
 func (c *Scratch) GetSortedMapValues(key string) any {
 	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	if c.values[key] == nil {
-		c.mu.RUnlock()
 		return nil
 	}
 
 	unsortedMap := c.values[key].(map[string]any)
-	c.mu.RUnlock()
 	var keys []string
 	for mapKey := range unsortedMap {
 		keys = append(keys, mapKey)
--- a/common/hstore/scratch_test.go
+++ b/common/hstore/scratch_test.go
@@ -14,6 +14,7 @@
 package hstore
 
 import (
+	"fmt"
 	"sync"
 	"testing"
 
@@ -214,4 +215,26 @@ func BenchmarkScratchGet(b *testing.B) {
 	for b.Loop() {
 		scratch.Get("A")
 	}
+}
+
+func TestScratchGetSortedMapValuesConcurrentWithSetInMap(t *testing.T) {
+	c := qt.New(t)
+
+	s := NewScratch()
+	var wg sync.WaitGroup
+	for i := 0; i < 100; i++ {
+		wg.Add(2)
+		go func(i int) {
+			defer wg.Done()
+			s.SetInMap("m", fmt.Sprintf("k%d", i), i)
+		}(i)
+		go func() {
+			defer wg.Done()
+			s.GetSortedMapValues("m")
+		}()
+	}
+	wg.Wait()
+
+	val := s.GetSortedMapValues("m").([]any)
+	c.Assert(val, qt.HasLen, 100)
 }


bep:

volcanicll:
CLA is signed — the `license/cla` check now passes, and the commit email (`72688335+volcanicll@users.noreply.github.com`) is linked to my GitHub account. The PR is ready for review. Thanks!

bep:
I may have missed out on who was first with this fix, and I apologise for that as I got 3 PRs for the same issue. This was fixed in 166d3ee1ce9414090247a20b4710cd57a737de05

PR #15240, opened 2026-08-23
tpl/tplimpl: support inline partials in content adapters

### LINKED ISSUE #14480: Add support for inline partials in content adapters
After discussion in https://discourse.gohugo.io/t/using-inline-partials-in-content-adapters/56693 I'd like to propose the following feature (text copied from that post):

I’m experimenting with Content Adapters, specifically I’d like to be able to use Go templates and pass these as the content to the `.AddPage`method. I came up with a simple example based on a combination of the examples in https://gohugo.io/content-management/content-adapters/#addpage and https://gohugo.io/functions/go-template/define/#article as follows, which I added to `content/test/_content.gotmpl`:

```
{{ define "_partials/inline/foo.html" }}
  {{ printf "The _Hunchback of Notre Dame_ was written by Victor Hugo in %v." .answer }}
{{ end }}

{{ $content := dict
  "mediaType" "text/markdown"
  "value" (partial "inline/foo.html" (dict "answer" 1831))
}}

{{ $page := dict
  "content" $content
  "kind" "section"
  "path" "../test"
  "title" "The Hunchback of Notre Dame"
}}

{{ .AddPage $page }}
```

Unfortunately, this results in the following error message:

```
error building site: 
process: readAndProcessContent: "/content/test/_content.gotmpl:7:11": 
template: /content/test/_content.gotmpl:7:11: 
executing "/content/test/_content.gotmpl" at <partial "inline/foo.html" (dict "answer" 1831)>: 
error calling partial: partial "inline/foo.html" not found
```

According to @bep this is currently not supported, but could possibly be added.

zanarellidev:
Fixes #14480

Extract inline partials parsed in `_content.gotmpl` into the template store so calls to `partial "inline/..."` inside content adapters resolve and execute properly.

### DIFF
--- a/hugolib/pagesfromdata/pagesfromgotmpl_integration_test.go
+++ b/hugolib/pagesfromdata/pagesfromgotmpl_integration_test.go
@@ -961,3 +961,36 @@ templateMetrics = true
 	b.Assert(got, qt.Contains, "/docs/_content.gotmpl")
 	b.Assert(got, qt.Not(qt.Contains), "content/docs/_content.gotmpl")
 }
+
+// See issue 14480.
+func TestContentAdapterInlinePartial(t *testing.T) {
+	t.Parallel()
+
+	files := `
+-- hugo.toml --
+disableKinds = ["taxonomy", "term", "sitemap", "robotsTXT"]
+-- content/test/_content.gotmpl --
+{{ define "_partials/inline/foo.html" }}
+The answer is {{ .answer }}.
+{{ end }}
+
+{{ $content := dict
+  "mediaType" "text/markdown"
+  "value" (partial "inline/foo.html" (dict "answer" 1831))
+}}
+
+{{ $page := dict
+  "content" $content
+  "kind" "page"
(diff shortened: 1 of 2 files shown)


bep:
See my comment in https://github.com/gohugoio/hugo/issues/14480#issuecomment-3864299733

The above is probably not a simple thing to do, but I don't think we can add inline partials to content adapters if it leaks into the regular templates' namespace -- that would be confusing for the end user. I will try to think a little about how to solve this; Ideally I think we should make them scoped to the current content adapter template only,

zanarellidev:
@bep That makes total sense. Scoping inline templates strictly to the executing adapter (or `resources.ExecuteAsTemplate` invocation context) via an isolated/child template namespace rather than registering into the global template store avoids polluting the page rendering namespace.

I can look into encapsulating the inline partial lookup within the adapter's execution context so they remain purely local. If you already have a preferred architectural pattern for sub-template namespaces in Hugo, let me know and I can adapt this PR accordingly!

PR #15241, opened 2026-08-24
resources: Publish copies of page resources

### LINKED ISSUE #10584: resources.Copy silently fails with page resources
This works:

```
{{ with resources.Get "a.jpg" }}
  {{ with . | resources.Copy "b.jpg" }}
  {{ end }}
{{ end }}
```

This silently fails:

```
{{ with .Resources.Get "a.jpg" }}
  {{ with . | resources.Copy "b.jpg" }}
  {{ end }}
{{ end }}
```

Reference:
<https://discourse.gohugo.io/t/copying-a-resource-in-page-bundles/42077/2>

See also: https://github.com/gohugoio/hugo/issues/12270

---

Here's a trick to get this to work... call the `Publish` method. Seems that `.RelPermalink` doesn't publish the resource.

```text
{{ with .Resources.Get "a.jpg" }}
  {{ with . | resources.Copy "b.jpg" }}
    {{ .Publish }}
  {{ end }}
{{ end }}
```

4RH1T3CT0R7:
`resources.Copy` inherits the publish mode of its source, and two kinds of source never carry `publishOnce`: a page's own resources, which the build publishes eagerly, and transformed resources, which write to their destination on init. Copies of either got none, so the file never got written even though `.RelPermalink` returned a link to it.

A copy has its own target path and nothing else writes it, so it now always publishes lazily.

Fixes #10584

This PR was written primarily by Claude Code. The change has been reviewed manually; the new test fails on master with the reported symptom.

### DIFF
--- a/resources/transform.go
+++ b/resources/transform.go
@@ -238,16 +238,15 @@ func (r *resourceAdapter) GetDependencyManager() identity.Manager {
 func (r resourceAdapter) cloneTo(targetPath string) resource.Resource {
 	newtTarget := r.target.cloneTo(targetPath)
 	r.sourceTarget = newtTarget.(transformableResource)
-	newInner := &resourceAdapterInner{
-		ctx:    r.ctx,
-		spec:   r.spec,
-		Staler: r.Staler,
-		target: newtTarget.(transformableResource),
+	// The copy has its own target path and is never part of an eager publish set
+	// (e.g. a page's own resources), so it must always publish lazily.
+	r.resourceAdapterInner = &resourceAdapterInner{
+		ctx:         r.ctx,
+		spec:        r.spec,
+		publishOnce: &publishOnce{},
+		target:      newtTarget.(transformableResource),
+		Staler:      r.Staler,
 	}
-	if r.resourceAdapterInner.publishOnce != nil {
-		newInner.publishOnce = &publishOnce{}
-	}
-	r.resourceAdapterInner = newInner
 	return &r
 }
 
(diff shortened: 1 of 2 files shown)


4RH1T3CT0R7:
@bep would you have a moment for this one? The added test fails on master with the reported symptom and passes with the change.

jmooring:
The desired behavior when copying page resources has not been discussed.

In <https://gohugo.io/functions/resources/copy/> we state:

> The TARGETPATH is relative to the [site root](http://localhost:1313/quick-reference/glossary/#site-root).

That makes sense for global and remote resources, but in my view it does **not** make sense for page resources. I would expect to be able to publish these relative to the published page.

We could use a forward slash at the beginning of the `TARGETPATH` to indicate a site-relative (not server-relative) copy, else page-relative. The forward slash would be ignored/irrelevant for global and remote resources.

```text
{{ with .Resources.Get "b.txt" }}
  Original: {{ .RelPermalink }}
  {{ with . | resources.Copy "foo/b-copy-1.txt" }}
    Relative to page: {{ .RelPermalink }}
  {{ end }}
  {{ with . | resources.Copy "/foo/b-copy-2.txt" }}
    Relative to site root: {{ .RelPermalink }}
  {{ end }}
{{ end }}
```

This needs approval from the project lead. Please do not update the PR until there's a decision.

jmooring:
The desired behavior needs some discussion and refinement. See https://github.com/gohugoio/hugo/issues/10584#issuecomment-5717907922.

PR #15248, opened 2026-08-28
resources/images: Parse hex colors as non-premultiplied NRGBA

### LINKED ISSUE #12536: images: Opacity problem with Resize + Padding with PNG output

```console
$ hugo version
hugo v0.126.1+extended linux/amd64
```

Yes.

### Report

The [`images.Padding`](https://gohugo.io/functions/images/padding/) function currently ignores the transparency setting when padding a PNG image. While the documentation does state '*conversion to WEBP is required to support transparency*' (under [Other recipes](https://gohugo.io/functions/images/padding/#other-recipes)), I can't see a reason why PNG shouldn't be supported, given that it has an alpha channel. For the moment I have resorted to first converting the image to WEBP, then applying transparent padding, and then converting it back to PNG, but this does not feel like an elegant solution.

hamodywe:
Fixes #12536 (root-cause analysis in [this comment](https://github.com/gohugoio/hugo/issues/12536#issuecomment-5447487023)).

## What was wrong

Hex notation is not alpha-premultiplied, but `hexStringToColorGo` parsed it into a `color.RGBA` — which **is** premultiplied by definition. Every transparent color (e.g. `#00f7`) therefore became an invalid premultiplied color (component > alpha). Whether the corruption was visible depended on the destination image type the filter pipeline picks, which follows the source image type: a JPEG source draws onto `*image.NRGBA`, where the invalid color is clamped back to the intended value by accident; an opaque-PNG source (e.g. the intermediate written by a previous `.Resize "x200 png"`) decodes to an alpha-premultiplied `*image.RGBA`, the invalid pixels survive, and the PNG encoder un-premultiplies them into a different color. That is precisely the difference between the working and broken constructs in the issue.

## The change

- `hexStringToColorGo` returns the `color.NRGBA` the hex string denotes.
- `ColorGoToHexString` un-premultiplies via `color.NRGBAModel` for symmetry (opaque colors unchanged).
- `Color` normalizes to the same canonical non-premultiplied form whether it came from a hex string or from image pixels, so equal colors compare equal.
- `AddColorToPalette` compares color values instead of interface values, so an equal color is still found across color models.

## Verification

Built two binaries and ran a minimal site using the exact templates from the issue thread, sampling the padding pixel of the emitted PNGs:

| Construct | before | after |
|---|---|---|
| `images.Process "resize x200 png"` + `Padding` (one pass) | `rgba(0 0 255 / 119)` ✓ | `rgba(0 0 255 / 119)` ✓ (unchanged) |
| `.Resize "x200 png"` then `Filter` + `Padding` (chained) | `rgba(0 0 36 / 119)` ✗ | `rgba(0 0 255 / 119)` ✓ |

- New `TestPaddingTransparentColor` covers both destination paths through `ImageProcessor.Filter` with an encode/decode round-trip; with the fix reverted, exactly the RGBA-source case fails (the YCbCr case documents the previously-working path).
- `hexStringToColorGo` tests extended with transparent colors.
- `go test ./resources/... ./tpl/images/...` green, including the golden image tests (all golden colors are opaque, where NRGBA and RGBA bytes are identical).

## Note for review

The output bytes change for transparent filter colors while image cache keys do not, so an affected site needs its `resources/_gen` cache cleared to pick up the fix. If you'd rather force that globally I can bump `mainImageVersionNumber` in this PR — left out deliberately since it invalidates every cached image for everyone.

This also lets the `images.Padding` docs drop the "conversion to WEBP is required to support transparency" workaround note once released.

*Disclosure per the contribution guidelines: this PR was written with AI assistance (Claude Code); the analysis and both before/after builds were verified locally as described above.*

### DIFF
--- a/resources/images/color.go
+++ b/resources/images/color.go
@@ -69,9 +69,13 @@ func (c Color) Key() string {
 }
 
 func (c *Color) init() error {
-	c.hex = ColorGoToHexString(c.color)
-	r, g, b, _ := c.color.RGBA()
-	c.luminance = 0.2126*c.toSRGB(uint8(r)) + 0.7152*c.toSRGB(uint8(g)) + 0.0722*c.toSRGB(uint8(b))
+	// Normalize to the canonical non-premultiplied form so that equal colors
+	// compare equal no matter where they came from (a hex string or the
+	// pixels of an image).
+	nrgba := color.NRGBAModel.Convert(c.color).(color.NRGBA)
+	c.color = nrgba
+	c.hex = ColorGoToHexString(nrgba)
+	c.luminance = 0.2126*c.toSRGB(nrgba.R) + 0.7152*c.toSRGB(nrgba.G) + 0.0722*c.toSRGB(nrgba.B)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


bep:

PR #15278, opened 2026-09-02
Fix: prevent render as <pre> code block

djibe:
Shortcode is broken as code indentation leads to a render as codeblock.


bep:
Can you add a test that demonstrates this broken behaviour?

jmooring:
We include an example of using the vimeo shortcode in our documentation.

This is the Markdown:
https://github.com/gohugoio/hugoDocs/blob/master/content/en/shortcodes/vimeo.md?plain=1#L28

And this is the result:
https://gohugo.io/shortcodes/vimeo/#example

djibe:
Then it must come from my code.
Sorry to have distrubed you.
It won't happen again.

Thanks for Hugo.

PR #15287, opened 2026-09-03
markup/pandoc: Support pandoc 3.11 and later

### LINKED ISSUE #15271: markup/pandoc: Support pandoc 3.11 and later

<pre>
$ hugo version
hugo v0.165.0+extended+withdeploy linux/amd64 BuildDate=unknown
</pre>

Yes.

### The issue

When building my website with `hugo`, I get the following warning for content that use the Pandoc markup:

```bash
[WARNING] Deprecated: --mathjax. Use --math-method=mathjax[:URL] instead.
```

Pandoc on my system (Arch) is on the latest version (`3.11`). In this version [release notes][1], one can read:

> Add --math-method option. This replaces (now deprecated but still functional) options --mathml, --mathjax, --gladtex, --katex, --webtex. The plain style can now be specified explicitly. In defaults files, html-math-method is now math-method (though html-math-method will still work).

This is not quite an *issue*, as everything works fine.

[1]: https://pandoc.org/releases.html#pandoc-3.11-2026-08-28

jmooring:
Closes #15271
Closes #15280

### DIFF
--- a/markup/pandoc/convert.go
+++ b/markup/pandoc/convert.go
@@ -15,7 +15,10 @@
 package pandoc
 
 import (
+	"bytes"
 	"fmt"
+	"io"
+	"sync"
 
 	"github.com/gohugoio/hugo/common/hexec"
 	"github.com/gohugoio/hugo/htesting"
@@ -42,6 +45,9 @@ func (p provider) New(cfg converter.ProviderConfig) (converter.Provider, error) 
 type pandocConverter struct {
 	ctx converter.DocumentContext
 	cfg converter.ProviderConfig
+
+	mathMethodSupportedOnce sync.Once
+	mathMethodSupported     bool
 }
 
 func (c *pandocConverter) Convert(ctx converter.RenderContext) (converter.ResultRender, error) {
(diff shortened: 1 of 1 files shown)


jmooring:
Tested with 3.11 locally. CI will continue to test with whatever version it installs.

markup/pandoc/convert.go
```diff
@@ -75,6 +82,31 @@ func getPandocBinaryName() string {
...
+
+var (
+	mathMethodSupportedOnce sync.Once
+	mathMethodSupported     bool
+)
+
+// pandocSupportsMathMethod reports whether the installed pandoc binary
+// supports the --math-method flag.
+func pandocSupportsMathMethod(cfg converter.ProviderConfig) bool {
+	mathMethodSupportedOnce.Do(func() {
```
bep: We don't want to use a global once that depends on the `cfg`. There's a `pandocConverter` struct that you can put these fields in (that converter will be recreated if the config changes). Admittingly rare, but there may/will be people in the future complaining that "i changed my sec config while the server was running to allow pandoc, but nothing happened ..." or seomething,

PR #15298, opened 2026-09-06
create: Print new content path

### LINKED ISSUE #15233: "hugo new" should be more quiet
When one types `hugo new foo/bar.md` it produces output that looks like

```
Content "/Users/eax/svn/blog/testblog/content/cases/bar.md" created
```

which is helpful but not quickly automatable to open. In particular it prevents 
```
vim -- $(hugo new foo/bar.md)
```

from open the file.

I suggest dropping the words "Content" and "created" along with the quotes so that one can use the above shortcut.

I'm open to requiring a flag for this but don't think there is enough value in the non-scripted case to warrant adding a flag.

Soundcreates:
Fixes #15233.

`hugo new` now prints the created content path without the surrounding label and quotes, so the output can be passed directly to another command, for example `vim -- $(hugo new foo/bar.md)`.

Added a regression test covering the log output.

AI assistance disclosure: This PR was prepared primarily with AI assistance; the change was manually reviewed and the focused Hugo checks passed.

### DIFF
--- a/create/content.go
+++ b/create/content.go
@@ -252,7 +252,7 @@ func (b *contentBuilder) buildFile() (string, error) {
 		return "", err
 	}
 
-	b.h.Log.Printf("Content %q created", contentPlaceholderAbsFilename)
+	b.h.Log.Printf("%s", contentPlaceholderAbsFilename)
 
 	return contentPlaceholderAbsFilename, nil
 }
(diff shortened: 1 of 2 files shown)


jjrauch32-lab (approved):

jmooring:
https://github.com/gohugoio/hugo/issues/15233 is a proposal. It has not been accepted by the project lead.

See <https://github.com/gohugoio/hugo#contributing>, specifically:

PR #15306, opened 2026-09-07
security: Add security.http.proxyFromEnvironment config option

### LINKED ISSUE #15302: Add config option security.http.proxyFromEnvironment = false
This is a security hardening issue.

In d6e6f9e500eebdeae8e28de830fff2e3bfc7d534 (unreleased) we added a dial-time URL validation.

If proxied (e.g. `HTTPS_PROXY` set on users machine or in CI) the above validation will only see the proxy address. There's no workaround for that, but I suggest by adding a new `security.http.proxyFromEnvironment` bool config flag (default disabled).

Thanks to @Reload3d for reporting this issue.

swapnil-goat:
Add a bool config option `security.http.proxyFromEnvironment`, defaulting to `false`, to disable inheriting `HTTP_PROXY`/`HTTPS_PROXY` from the environment on remote fetches unless explicitly enabled.

Closes #15302

### DIFF
--- a/config/security/securityConfig.go
+++ b/config/security/securityConfig.go
@@ -133,6 +133,10 @@ type HTTP struct {
 
 	// Media types where the Content-Type in the response is used instead of resolving from the file content.
 	MediaTypes Whitelist `json:"mediaTypes"`
+
+	// ProxyFromEnvironment specifies whether to use HTTP/HTTPS proxy settings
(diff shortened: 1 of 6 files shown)


bep:
Any issue assigned to me must be assumed there's a fix from me incoming, but thanks.

Issue #14079, opened 2025-10-22
Can't add value from `Params` to a Date using AddDate


mwegnr:
### What version of Hugo are you using (`hugo version`)?

<pre>
$ hugo version
0.152.0
</pre>

### Does this issue reproduce with the latest release?
Yes, it first occured with the latest release

### Description
Previously, integer-Params defined in the Front-Matter could be added directly to a date with `AddDate 0 0 .Params.int_param`. This is broken in 0.152.0 since the param is now parsed as `uint64` and `AddDate` expects an `int`

### Example Code
Front-Matter:
```
---
title: Homepage
test_param_int: 4
---
```

`list.html`:
```
  {{ $test_date := time.Now }}
  Date + 2: {{ $test_date.AddDate 0 0 2 | time.Format ":date_medium" }}<br> {{/*  <- works  */}}
  Date + Param.test_param_int: {{ $test_date.AddDate 0 0 (.Params.test_param_int) | time.Format ":date_medium" }}<br> {{/*  <- breaks  */}}
```

Error message:
```
ERROR Rebuild failed: render: failed to render pages: render of "/srv/hugo/content/_index.md" failed: "/srv/hugo/layouts/list.html:8:65": execute of template failed at <.Params.test_param_int>: wrong type for value; expected int; got uint64
```

### Example Hugo Page
https://github.com/mwegnr/hugo-uint64-int-param

mwegnr:
If I cast the value to an int with `cast.ToInt`, it also works

```
Date + Param.test_param_int: {{ $test_date.AddDate 0 0 (cast.ToInt .Params.test_param_int) | time.Format ":date_medium" }}<br>
```

Not sure if this is intended

bep:
>Not sure if this is intended

Well,

* [github.com/goccy/go-yaml](https://github.com/goccy/go-yaml] is the only maintained YAML 1.1 Go library.
* That library uses `uint64` for unsigned integers.
* `.AddDate` is a method that lives outside of Hugo.

We had no choice but to upgrade to the new library, we had to do that at some point anyhow. I knew there would be some side effects. 

I will leave this open, but don't expect a quick fix. But we should find a general solution to these subtle integer issues. That Go does not have a varint type is, I think, Go's biggest design flaw.

jmooring:
I recognize this is an unexpected and undesirable behavioral change, but also note that the construct above has _never_ worked if the front matter is TOML (3 -> int64) or JSON (3 -> float64).

[...]

PR #15310, opened 2026-09-08
common/hreflect: Fix the off-by-one in the float to int overflow guards

youdie006:
`convertToIntIfPossible` rejects an out-of-range float64 with `f > float64(math.MaxInt64)` (`common/hreflect/convert.go:208`). `float64(math.MaxInt64)` rounds up to exactly 2^63, which int64 cannot hold, so `f == 2^63` passes the guard, `math.Trunc` leaves it unchanged, and `typ.OverflowInt(int64(f))` is handed an already-wrapped value and also passes.

`convertToUintIfPossible` has the same problem at `:148` against `float64(math.MaxUint64)`, which rounds up to 2^64.

Measured on `master`:

```
float64 input              to type    ok?      result
2^63 (float64(MaxInt64))   int64      true     -9223372036854775808
largest float64 < 2^63     int64      true     9223372036854774784
-2^63 (float64(MinInt64))  int64      true     -9223372036854775808
2^64 (float64(MaxUint64))  uint64     true     9223372036854775808
```

So a positive value comes back as the most negative int64, and 2^64 comes back as exactly half of what was asked for — from a function whose doc comment at `:95` says *"This conversion is lossless."*

### Why this is reachable

`ConvertIfPossible` exists so template calls can pass data-file numbers to Go methods with concrete integer parameters (`:96` cites #14079). `validateType` calls it at `tpl/internal/go_templates/texttemplate/hugo_template.go:553`, and a large integer in a JSON or YAML data file decodes to float64.

Executing a template that passes such a value to a method taking an `int`, on `master`:

```
{{ (.Limit .Big) }}   error: slice bounds out of range [:-9223372036854775808]
```

The negative value reaches the method and it panics on the slice bound. With this change the same input reports the ordinary type error instead:

```
{{ (.Limit .Big) }}   error: wrong type for value; expected int; got float64
```

which is already what one float step above 2^63 does today.

### The change

`>` becomes `>=` on the two upper bounds. Nothing else moves.

The lower bound is deliberately left alone: unlike the upper one, `float64(math.MinInt64)` is exactly -2^63 and *is* representable, so it stays in range. A test row pins that so the two bounds do not get made symmetric later.

The uint branch of the same function already gets this right by comparing exactly — `val.Uint() > uint64(math.MaxInt64)` at `:198` — rather than in float64.

### Testing

Rows added to the existing `TestConvertIfPossible` table: the two rejected boundaries, the largest float64 that is a valid int64 and uint64, and `float64(math.MinInt64)` as accepted.

- The new rejection rows fail on `master`; the accepted rows pass on both trees.
- Mutation-checked, all four caught: reverting either upper bound to `>`, over-rejecting the exact negative bound, and over-rejecting the largest valid int64.
- `go test ./common/... ./tpl/...` passes, as do `./hugolib/...`, `./resources/...` and `./markup/...` including `TestYAMLAddDateIssue14079`.
- `go build ./...` and `GOOS=windows go build ./common/...` clean; `gofmt` via `./check.sh` clean; `go vet ./common/... ./tpl/...` clean.

## AI Assistance Notice

This PR was written primarily by Claude Code. I have reviewed the change and can explain it; the tables above are verbatim from running the conversion and the template engine against `master`.

### DIFF
--- a/common/hreflect/convert.go
+++ b/common/hreflect/convert.go
@@ -145,7 +145,7 @@ func convertToUintIfPossible(val reflect.Value, typ reflect.Type) (reflect.Value
 	}
 	if IsFloat(val.Kind()) {
 		f := val.Float()
-		if f < 0 || f > float64(math.MaxUint64) {
+		if f < 0 || f >= float64(math.MaxUint64) {
 			return reflect.Value{}, false
 		}
 		if f != math.Trunc(f) {
@@ -205,7 +205,7 @@ func convertToIntIfPossible(val reflect.Value, typ reflect.Type) (reflect.Value,
 	}
 	if IsFloat(val.Kind()) {
 		f := val.Float()
-		if f < float64(math.MinInt64) || f > float64(math.MaxInt64) {
+		if f < float64(math.MinInt64) || f >= float64(math.MaxInt64) {
 			return reflect.Value{}, false
 		}
 		if f != math.Trunc(f) {
--- a/common/hreflect/convert_test.go
+++ b/common/hreflect/convert_test.go
@@ -110,6 +110,29 @@ func TestConvertIfPossible(t *testing.T) {
 			ok:    false, // overflow
 		},
 		{
+			// float64(math.MaxInt64) rounds up to 2^63, which is out of range.
+			name:  "float64(math.MaxInt64) to int64",
+			value: float64(math.MaxInt64),
+			typ:   int64(0),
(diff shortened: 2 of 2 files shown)


bep:
Thanks for this: One general comment: You (and others) need to ask your AI agent to stop writing these long novels in PR descriptions/commit message. This is a relative simple issue that can be explained in one sentence.

PR #15326, opened 2026-09-10
tpl/compare: Make eq compare numbers by value

### LINKED ISSUE #15322: tpl/compare: eq, le and ge disagree when comparing an int with an equal float

<pre>
$ hugo version
</pre>

I built `master` at `ae07063` and reproduced it there. I did not run a tagged release binary, so I have not confirmed it against one directly, though the code path is unchanged in `tpl/compare/compare.go`.

`eq`, `ne`, `lt`, `le`, `gt` and `ge` agreeing with each other. Whatever the rule for comparing an `int` against a `float`, exactly one of `lt`/`eq`/`gt` should hold for a given pair, `le` should mean `lt or eq`, and `ge` should mean `gt or eq`.

### What did you see instead?

For `1` against `1.0`, the six functions contradict each other. `le` and `ge` are both true while `lt`, `gt` and `eq` are all false, so the value is reported as neither less than, nor greater than, nor equal to itself.

`layouts/index.html`:

```go-html-template
eq  1 1.0 = {{ eq 1 1.0 }}
ne  1 1.0 = {{ ne 1 1.0 }}
lt  1 1.0 = {{ lt 1 1.0 }}
le  1 1.0 = {{ le 1 1.0 }}
gt  1 1.0 = {{ gt 1 1.0 }}
ge  1 1.0 = {{ ge 1 1.0 }}
```

Rendered output:

```
eq  1 1.0 = false
ne  1 1.0 = true
lt  1 1.0 = false
le  1 1.0 = true
gt  1 1.0 = false
ge  1 1.0 = true
```

`eq` normalises an `int` to `int64` and a `float` to `float64` and then compares them as `any`, so they are never equal. `lt`/`le`/`gt`/`ge` go through `compareGet`, which converts both to `float64`, so they compare numerically. The two halves of the API use different rules.

Same result for `0` vs `0.0`, `-1` vs `-1.0`, `int64(3)` vs `float64(3)` and `uint(4)` vs `4.0`. Mixed values where the numbers genuinely differ (`1` vs `2.0`) are consistent, so this only shows up when the two are numerically equal.

This is adjacent to #3721, which asked for `eq` to be more lenient and was closed by the stale bot in 2018 without a decision. This report is narrower: whichever convention you prefer, the current combination is not self-consistent.

I have not sent a patch because the fix is a semantics choice, either making `eq` compare numerically or making `le`/`ge` strict about mixed types, and that is your call. Happy to send a PR once you say which.

dylanpulver:
Closes #15322.

Per your note that making `eq` compare numerically would generate the least noise.

`Eq` normalizes ints to `int64` and floats to `float64`, then compares with `reflect.DeepEqual`, which reports them unequal. `Lt`, `Le`, `Gt` and `Ge` go through `compareGet`, which converts both to `float64`. So for `1` and `1.0`, `le` and `ge` were both true while `lt`, `eq` and `gt` were all false: the value was neither less than, greater than, nor equal to itself.

This adds a `numbersEqual` fallback that runs only when `DeepEqual` says no and both sides are numeric.

I did not reuse `compareGet`, though it would have been shorter. It also compares arrays, slices, maps and channels **by length**, so routing `eq` through it would make `eq` of two same-length slices return true.

Also unchanged: a number and a numeric string. `eq 1 "1"` is still false. Worth flagging that it has the same shape as the reported bug (`eq=false lt=false gt=false le=true ge=true`), but equating a number with a numeric string is a much larger semantic change than making two numbers compare by value, so I left it alone. Happy to open a separate issue for it if you want it tracked.

The new test pins the six operators against each other for mixed numeric kinds, so they cannot drift apart again without a failure.

`go test ./tpl/... ./hugolib/...` passes.

This PR was written with AI assistance (Claude Code).

### DIFF
--- a/tpl/compare/compare.go
+++ b/tpl/compare/compare.go
@@ -150,8 +150,48 @@ func (n *Namespace) Eq(first any, others ...any) bool {
 		if reflect.DeepEqual(normFirst, other) {
 			return true
 		}
+
+		// Numbers of different kinds normalize to different Go types, so
+		// reflect.DeepEqual reports 1 and 1.0 as unequal. Lt, Le, Gt and Ge
+		// compare such pairs numerically via compareGet, so eq did not agree
+		// with them: le and ge were both true while lt, eq and gt were all
+		// false. See #15322.
+		if numbersEqual(normFirst, other) {
+			return true
+		}
 	}
 
+	return false
+}
+
+// numbersEqual reports whether a and b are both numbers of equal value. Both
+// arguments must already be normalized to int64, uint64 or float64. Mixed
+// int/float comparisons go through float64, matching compareGet, so that eq
+// agrees with the ordering operators for every pair they both accept.
+func numbersEqual(a, b any) bool {
+	switch av := a.(type) {
+	case int64:
+		switch bv := b.(type) {
+		case uint64:
+			return av >= 0 && uint64(av) == bv
+		case float64:
+			return float64(av) == bv
+		}
+	case uint64:
+		switch bv := b.(type) {
+		case int64:
+			return bv >= 0 && av == uint64(bv)
+		case float64:
+			return float64(av) == bv
+		}
+	case float64:
+		switch bv := b.(type) {
+		case int64:
+			return av == float64(bv)
+		case uint64:
+			return av == float64(bv)
+		}
+	}
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


bep:
Fixed in 4d628bbdc3441f3c44b1f6f4e6ad04f4df4728bf

PR #15328, opened 2026-09-11
tpl/collections: Fix union of slices with different numeric element types

hktitof:
## Summary
union returns an empty slice with no error whenever the two arguments have different slice types and neither has an interface element type, so `union (slice 1) (slice 1.0)` gives `[]` while `intersect` with the same inputs gives `[2]`

the early return for different element types drops everything even when both element types are numeric, and the code right below already converts numbers across types with convertNumber, intersect also matches numbers across types, so the two siblings disagree on the same input

this narrows the early return to non numeric mismatches only, numeric lists now unify through the existing conversion path, values that cant convert lossless are skipped, string mismatches keep the old empty result so the tested behavior from the `[]A ∪ []B` cases is unchanged

### Steps to reproduce

`layouts/index.html`:

```go-html-template
{{ printf "%T" (slice 1) }}
{{ printf "%T" (slice 1.0) }}
union int,int    -> {{ union (slice 1) (slice 2) }}
union int,float  -> {{ union (slice 1) (slice 1.0) }}
union int,string -> {{ union (slice 1) (slice "a") }}
union lists      -> {{ union (slice 1 2) (slice 2.0 3.0) }}
intersect lists  -> {{ intersect (slice 1 2) (slice 2.0 3.0) }}
```

before, on master at ae07063:

```
union int,int    -> [1 2]
union int,float  -> []
union int,string -> []
union lists      -> []
intersect lists  -> [2]
```

after this change:

```
union int,int    -> [1 2]
union int,float  -> [1]
union int,string -> []
union lists      -> [1 2 3]
intersect lists  -> [2]
```

tests added to TestUnion pin the numeric unify cases, the full tpl/collections suite passes, see issue 15323

## Assistance Disclosure
AI used: yes, for drafting test coverage and description formatting

### DIFF
--- a/tpl/collections/collections.go
+++ b/tpl/collections/collections.go
@@ -657,7 +657,12 @@ func (ns *Namespace) Union(l1, l2 any) (any, error) {
 			if l1v.Type() != l2v.Type() &&
 				l1v.Type().Elem().Kind() != reflect.Interface &&
 				l2v.Type().Elem().Kind() != reflect.Interface {
-				return ins.r.Interface(), nil
+				// Numeric element types unify through the conversion below,
+				// mirroring how intersect matches numbers across types.
+				if !hreflect.IsNumber(l1v.Type().Elem().Kind()) ||
+					!hreflect.IsNumber(l2v.Type().Elem().Kind()) {
+					return ins.r.Interface(), nil
+				}
 			}
 
 			var (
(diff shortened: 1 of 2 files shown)


tpl/collections/collections.go
```diff
@@ -698,7 +703,7 @@ func (ns *Namespace) Union(l1, l2 any) (any, error) {
 				case hreflect.IsNumber(kind):
 					var err error
 					l2vv, err = convertNumber(l2vv, typ)
-					if err == nil {
+					if err == nil && l2vv.Type().AssignableTo(l1v.Type().Elem()) {
```
bep, line 707: The motivation for the change above is unclear. Can you either add a test that demonstrates its value or remove it.

bep (changes requested):

tpl/collections/collections.go
hktitof: thanks for the direction, the guard is load-bearing so i kept it and added a test case for it: with an empty []int the result prototype comes from l2, so union of an empty []int and []float64{1.0} panics without the guard (reflect.Set: value of type float64 is not assignable to type int), with the guard it returns []int{} cleanly. full collections suite is green on the amended commit

bep: OK, but can you remove the hyphen so the comment makes sense.

hktitof: done, removed the hyphen from the comment @bep, thanks

PR #15335, opened 2026-09-14
tpl/collections: Fix union of slices with different element types

### LINKED ISSUE #15323: tpl/collections: union silently returns an empty list when the two slice types differ

<pre>
$ hugo version
</pre>

I built `master` at `ae07063` and reproduced it there. I did not run a tagged release binary, so I have not confirmed it against one directly, though `Union` is unchanged in `tpl/collections/collections.go`.

`union` either returning the union of its two arguments, or returning an error. It already errors for the other argument shapes it cannot handle, for example `union does not support slices or arrays of uncomparable types` and `can't iterate over ...`.

### What did you see instead?

`union` returns an empty slice, with no error, whenever the two arguments have different slice types and neither has an interface element type.

`layouts/index.html`:

```go-html-template
{{ printf "%T" (slice 1) }}
{{ printf "%T" (slice 1.0) }}
union int,int    -> {{ union (slice 1) (slice 2) }}
union int,float  -> {{ union (slice 1) (slice 1.0) }}
union int,string -> {{ union (slice 1) (slice "a") }}
union lists      -> {{ union (slice 1 2) (slice 2.0 3.0) }}
intersect lists  -> {{ intersect (slice 1 2) (slice 2.0 3.0) }}
```

Output:

```
[]int
[]float64
union int,int    -> [1 2]
union int,float  -> []
union int,string -> []
union lists      -> []
intersect lists  -> [2]
```

`intersect` handles the same two mixed-type lists and returns `[2]`, so the sibling functions disagree about whether these arguments are usable at all.

The cause is this early return in `Union`:

```go
if l1v.Type() != l2v.Type() &&
	l1v.Type().Elem().Kind() != reflect.Interface &&
	l2v.Type().Elem().Kind() != reflect.Interface {
	return ins.r.Interface(), nil
}
```

`ins.r` is the empty slice made just above, so a caller gets an empty list rather than an error. It is reachable from ordinary templates because `slice 1` is `[]int` and `slice 1.0` is `[]float64`, and it also runs before the uncomparable-types check, so `union (slice (slice 1)) (slice 2)` returns `[]` instead of the error that check would raise.

Adding one element of another type to the first argument avoids it, since the slice then becomes `[]interface {}`: `union (slice 1 "x") (slice 1.0)` returns `[1 x]`.

### Stack traceback or browser console output

None, no error is raised. That is the problem.

---

This is a different mechanism from #3411, which was closed in 2017 and is now locked with a note asking that related bugs be opened as new issues.

Happy to send a PR. Returning an error here would be a small change; making `union` accept mixed types the way `intersect` does is bigger, so I would rather you pick the direction first.

yuefdev:
union returned an empty slice when the two lists had different element types, e.g. `union (slice 1 2) (slice 2.0 3.0)`. Values from the second list are now converted to the element type of the first list and skipped if the conversion is not possible, which is how []T ∪ []any already behaved.

This also fixes panics when appending values that were not assignable to the element type of the first list, e.g. `[]int{} ∪ []any{1}`.

Fixes #15323


bep:
Fixed in ffbcdba9e9768b6a1bd0580eb56ed26c486f2cce

PR #15344, opened 2026-09-14
tpl/compare: add require template function

### LINKED ISSUE #15290: Feature request : add a `require` method for values that can be nil
Right now when a value can be `nil` we have the `default` method which produce a default value if the value piped is `nil` : 
```
{{ $value := ... | default <default-val> }}
```
I would like to have something similar which produce an error instead when the value is `nil`.
```
{{ $value := ... | require <error message> }}
```

Currently, the only way to do this is to check the value using if conditions : 
```
{{ $value := ... }}
{{ if not $value }}
    {{ errorf ... }}
{{ end }}
```

That can become tedious if we need to ensure many values must not be `nil`

Doing it with my proposed `require` function means it can be all be done in one line, making the partial and template code cleaner.

Jia-ben00:
## Description

Adds a `require` template function that returns an error when the piped value is nil or falsy, serving as the error-returning counterpart of `default`.

### Usage

```go-html-template
{{ $img := resources.Get "images/hero.jpg" | require "hero image is required" }}
```

If `resources.Get` returns nil (e.g. the file does not exist), the template build fails with the message "hero image is required" instead of silently continuing with a nil value.

### Design decisions

- **Semantics**: `require` checks for nil and all falsy values (0, "", empty slices/maps, zero time), consistent with how `default` determines whether a value is "set".
- **Boolean `false`**: Unlike `default` (which treats `false` as set), `require` treats `false` as not set and returns the error. This was discussed in #15290 — the consensus was that rejecting `false` makes more sense for a "required value" check.
- **Variadic second argument**: Like `default`, the piped value is variadic to handle the case where a missing key in `index` produces no piped argument.

### Testing

Added `TestRequireFunc` covering:
- Set values (true, non-empty string, non-zero numbers, non-empty slices/maps/arrays, non-zero time, structs) returned without error
- Unset values (false, "", 0, empty slices/maps/arrays, nil, zero time) return the configured error
- Variadic no-argument case (missing key in pipeline)

Closes #15290


bep:
Thanks for this,

PR #15349, opened 2026-09-15
fix(inflect): preserve opening quote spacing in humanize

### LINKED ISSUE #15126: humanize introduces space after „

163.2

Probably

```
humanize "Painting „Title“"
```
results in
`Painting „ title“`

thus a space after the first quote.

zzz-yu:
Fixes #15126

The inflect Humanize wrapper currently delegates to flect twice. For text containing the low double quotation mark (U+201E), flect inserts a separator space after the opening quote, so `humanize "Painting „Title“"` returns `Painting „ title“`.

Keep the opening quote attached to the following word after the existing normalization. Add a regression case covering the reported input while leaving the existing humanize, ordinal, pluralize, and singularize behavior unchanged.


jmooring:
This needs to be fixed in the upstream library. That's why the issue has an "upstream" label. See issue and pull request references in https://github.com/gohugoio/hugo/issues/15126.

PR #15350, opened 2026-09-16
fix(tpl/compare): make Eq consistent with Lt/Le/Gt/Ge for numeric comparisons

### LINKED ISSUE #15322: tpl/compare: eq, le and ge disagree when comparing an int with an equal float

OMD-123:
## Summary

This PR makes the `Eq` function use the same centralized `compareGetWithCollator` logic as `Lt`, `Le`, `Gt`, and `Ge`, ensuring consistent numeric comparison behavior across all comparison operators.

## Problem

The previous `Eq` implementation used a custom `normalize` function that converted integers to `float64` but had edge cases where it didn't match the behavior of the other comparison operators (which use `compareGetWithCollator`).

## Solution

Replace the custom normalization in `Eq` with a direct call to `compareGetWithCollator`, so all operators (`Eq`, `Ne`, `Lt`, `Le`, `Gt`, `Ge`) use the exact same numeric comparison logic.

## Changes

- `Eq` now uses `compareGetWithCollator` for numeric comparisons
- Removed the custom `normalize` function from `Eq`
- All numeric types (int, uint, float) are now compared consistently across all operators
- Edge cases like `int64(1<<53 + 1)` vs `float64(1<<53)` now behave identically in `Eq` and `Gt`/`Lt`

## Testing

All existing tests in `compare_test.go` pass with this change, including:
- Mixed int/float comparisons (e.g., `1 == 1.0`, `int64(3) == float64(3)`)
- Large number precision cases
- Uint/int boundary comparisons
- Custom Eqer implementations
- String, time, and version comparisons

Fixes #15322


bep:
> The previous Eq implementation used a custom normalize function that converted integers to float64 but had edge cases where it didn't match the behavior of the other comparison operators (which use compareGetWithCollator).

There's no test cases in this PR. If there are edge cases that the existing implementation does not cover, please add some test cases that demonstrates them.

PR #15359, opened 2026-09-19
tpl/collections: Compare numbers exactly in set operations

### LINKED ISSUE #15358: tpl/collections: intersect/union/uniq collapse or drop distinct big integers beyond the float64 exact range

<pre>
master (5698de94), built from source
</pre>

Yes, 0.166.0 shows the same class of wrong result (see below)

### Steps to reproduce

1. render this in a layout

```
{{ intersect (slice 9007199254740993 9007199254740992) (slice 9007199254740992 9007199254740993) }}
{{ intersect (slice 9007199254740993) (slice 9007199254740993) }}
{{ in (slice 9007199254740993) 9007199254740992 }}
```

2. Expected: the first intersect returns both values bcs they are in both lists, the second returns the value, in returns false bcs the two numbers are distinct (2^53+1 vs 2^53)
3. Actual on master:

```
intersectBoth: [9007199254740992]
intersectSame: []
inDiff: true
```

9007199254740993 silently disappears from the intersect result even though it is in both lists, a value does not match itself, and two distinct numbers compare equal

on 0.166.0 the same template prints intersectBoth: [9007199254740993] and inDiff: true, so the collision is there too, just a different arm of it

the set operations key numbers through float64, so int64 values beyond 2^53 (where float64 stops being exact) collapse or get dropped depending on which path takes them, the tpl/compare consolidation (4d628bbd) already made eq and where compare numbers exactly, the set operations in tpl/collections still go through the old float64 paths

hktitof:
> **Context:** follow-up to the numeric comparison consolidation 4d628bbd, which made eq and where compare numbers exactly but left the set operations on the float64 paths

### What does this PR change?

the set operations in tpl/collections still went through float64 to compare or key numbers, so int64 values beyond 2^53 (where float64 stops being exact) collapse or get dropped

- intersect dropped a value that is present in both lists: `intersect (slice 9007199254740993 9007199254740992) (slice 9007199254740992 9007199254740993)` returned `[9007199254740992]` instead of both
- a value did not match itself: `intersect (slice 9007199254740993) (slice 9007199254740993)` returned `[]`
- the same numbers answered differently depending on the path: `in` still says false for the two distinct numbers, while the intersect pair compare went through a float64 conversion that refuses these values, so they never matched, not even themselves

intersect now compares value pairs with hreflect.CompareNumbers, the exact comparison this consolidation introduced, and normalize keys numbers by their exact value so equal ones still unify across int, uint and float types in in, intersect, union, uniq, symdiff and complement

### Steps to reproduce

1. render this in a layout (or run `go test ./tpl/collections/ -run TestSetOperationsExactBigNumbers` on master)

```
{{ intersect (slice 9007199254740993 9007199254740992) (slice 9007199254740992 9007199254740993) }}
{{ intersect (slice 9007199254740993) (slice 9007199254740993) }}
{{ in (slice 9007199254740993) 9007199254740992 }}
```

2. Expected: the first intersect returns both values bcs they are in both lists, the second returns the value, in returns false bcs the two numbers are distinct (2^53+1 vs 2^53)
3. Actual (raw output on untouched master 5698de94, from the integration test in this pr):

```
intersectBoth: [9007199254740992]
intersectSame: []
inDiff: false
```

on 0.166.0 the same template prints `intersectBoth: [9007199254740993]` and `inDiff: true`, so the collision is there too

### Tests

- TestSetOperationsExactBigNumbers (integration, template level): pins the full matrix, distinct values stay distinct and equal values still match across int and float
- TestSetOperationsUseExactNumericComparison (unit): the same contract on the namespace functions

`./check.sh` passes (gofmt, staticcheck, tests)

Fixes #15358


bep:

Please create an issue with label proposal first.

hktitof:
thanks for the pointer @bep, the pr already fixes #15358, the issue is open but has no labels yet since i can't add those myself, so i'll leave it for the proposal triage and revisit once it carries a label

PR #15361, opened 2026-09-19
parser/metadecoders: Strip BOM from data passed to UnmarshalToMap

### LINKED ISSUE #15360: parser/metadecoders: UnmarshalToMap does not strip a leading UTF-8 BOM
follow up to #15355, that fix strips a leading UTF-8 BOM in Unmarshal, but the map decoder used for config files and front matter (UnmarshalToMap) still passes the BOM on to the format decoders

a hugo.toml that starts with a BOM (files edited in Windows editors do this a lot) fails the build on current master:

```
failed to load config: "/hugo.toml:1:1": unmarshal failed: toml: invalid character at start of key: U+00EF 'ï'
```

and a BOM prefixed config.yaml does not error at all, it silently keeps its first key as "\ufefftitle" so the setting never applies

a config.json with a BOM fails the same way as the toml case

the one line TrimPrefix that Unmarshal got covers this too, i have it ready with a regression test in config/allconfig, pr coming right up

hktitof:
> **Context:** follow-up to ec578f0c, which strips a leading UTF-8 BOM in Unmarshal but left the map decoder on the untrimmed path

### What does this PR change?

UnmarshalToMap is the decoder behind config files and front matter, it did not strip a leading UTF-8 BOM, so a config file saved with a BOM (Windows editors do this) breaks at config load

- a BOM prefixed hugo.toml fails the build on master ec578f0c:

```
failed to load config: "/hugo.toml:1:1": unmarshal failed: toml: invalid character at start of key: U+00EF 'ï'
```

- a BOM prefixed config.yaml does not even error, yaml keeps the BOM glued to the first key so the setting silently never applies (the key comes out as "\ufefftitle")

this adds the same one line TrimPrefix that Unmarshal got, so both decoders strip the BOM before parsing

### Steps to reproduce

1. save a hugo.toml that starts with the UTF-8 BOM bytes ef bb bf and run hugo (or run `go test ./config/allconfig/ -run TestSiteConfigWithBOM` on master)
2. Expected: the site builds and the config applies
3. Actual (raw output on untouched master ec578f0c, from the integration test in this pr):

```
failed to load config: "/hugo.toml:1:1": unmarshal failed: toml: invalid character at start of key: U+00EF 'ï'
```

### Tests

- TestSiteConfigWithBOM (integration, config/allconfig): a BOM prefixed hugo.toml builds and the title reaches the rendered site

`./check.sh` passes (gofmt, staticcheck, tests)

Fixes #15360


bep:
Thanks for this,

hktitof:
thanks for the explanation @bep, makes sense, a fix closer to the file system sounds like the right layer, happy to see this closed

Issue #14937, opened 2026-05-24
Add build.cleanDestinationDir config option


ilovelinux:
## Usecase

I'm using Codeberg Pages to host my website, you can find the repostitory at https://codeberg.org/ilovelinux/pages

I have two branches:
- `main`: hugo project
- `pages`: hugo build (output of `public/`)

Since I'm using Codeberg Pages with a custom domain, I also need a `.domains` file[^codebergDomains] to be pushed in the `pages` branch.

### When I

- Have files in the `static/` directory.
- Switch to `main` branch.
- Create the `public/` folder using a git worktree of the `pages` branch.
- Run `hugo build --cleanDestinationDir`.

### I need

- To remove outdated content
- To retain:
  - `.git` **file** created by `git worktree`[^gitWorktreeFile].
  - `.domains` **file** used by Codeberg Pages in order to allow custom domains[^codebergDomains].
  - `_redirects` **file** for redirects (however I'm not using it yet)[^codebergRedirects].

### Instead

- The files I need to be retained are deleted.

## Proposal

Add a configuration field to list relative file paths to `public/` that should be retained when using `--cleanDestinationDir`.

#### Example
```yaml
 filesToRetain:
  - .git
  - .domains
  - _redirects
  - foo/bar/.mysecretfile
```

Files retained:
- `public/.git`
- `public/.domains`
- `public/_redirects`
- `public/foo/bar/.mysecretfile`

> [!NOTE]
> I'm not suggesting to use `filesToRetain` as configuration field, but I couldn't think of a better name.

## Notes

> [!IMPORTANT]
> `--cleanDestinationDir` does nothing when when the `static/` directory doesn't exist since its value is used inside the `copyStaticTo` function.
> https://github.com/gohugoio/hugo/blob/c4bbc2805c7092deafabae01e71ff2dda37bd877/commands/hugobuilder.go#L481

### Workaround

> [!WARNING]
> This approach is error-prone.

I implemented a workaround in my [Makefile](https://codeberg.org/ilovelinux/pages/src/commit/072c6dc048096c98c3422e824cb1d3bbc0b60736/Makefile#L9-L11) that:
- Repairs the worktree folder, as suggested in https://github.com/gohugoio/hugo/pull/6261#issuecomment-1692321365.
- Restores the `.domains` file.

### Related issues & PR
- https://github.com/gohugoio/hugo/issues/3202 was about `.git` **folder** being deleted.
  Fixed thanks to:
[...]

PR #15362, opened 2026-09-20
commands: Handle missing staticDir in cleanDestinationDir

### LINKED ISSUE #8433: commands: Make cleanDestinationDir work even when staticDir doesn't exist

jmooring:
Closes #8433


testscripts/commands/hugo__clean_destination_dir.txt
```diff
@@ -0,0 +1,38 @@
+# See issue 8433.
+
+mkdir public
+cp temp.txt public/public.txt
```
bep: This test is missing the "static file in subdirectory" case.

jmooring: I've added the test, but realized that we don't have any existing tests for the delete filter either. I'll add that too.

jmooring: > we don't have any existing tests

Good exercise. This surfaced a pre-existing bug where the contents of stale directories were deleted without consulting the delete filter.

jmooring:
While this addresses #8433 and fixes a pre-existing bug, I'm beginning to think that we should also implement https://github.com/gohugoio/hugo/issues/14937 (at some point, using glob patterns) with defaults matching the existing behavior.

testscripts/commands/hugo__clean_destination_dir.txt
```diff
@@ -0,0 +1,230 @@
+# --- Test --cleanDestinationDir without static directory. ---
+
+# Setup:
```
bep: This test script is incredibly hard to read/verify.

I suggest that we instead

* Create a set of golden test dirs using txtar and move them into public for each step
* Then do `tree public` and compare `stdout` to a golden output
* Use p.UpdateScripts = true to get the golden output right

bep:
Looks good. I'm going to test/complete this in #15375 -- I'm a little extra wary about changes that is about deleting stuff on people's computers.

jmooring:
Sounds good.

> I'm a little extra wary about changes that is about deleting stuff on people's computers

Same here.

PR #15365, opened 2026-09-21
tpl/collections: Compare numbers exactly in set operations

### LINKED ISSUE #15358: tpl/collections: intersect/union/uniq collapse or drop distinct big integers beyond the float64 exact range

hexbinoct:
Fixes #15358.

The issue says the set operations key numbers through float64 and collapse values past 2^53. The collapse is not quite what happens: `hreflect.ToFloat64E` refuses a lossy conversion rather than performing it. The refusal is what breaks things, in two separate places.

**`intersect` treats a refused conversion as "not equal",** so a value outside float64's exact range fails to match itself:

```
{{ intersect (slice 9007199254740993) (slice 9007199254740993) }}
→ []

{{ intersect (slice 9007199254740993 9007199254740992) (slice 9007199254740992 9007199254740993) }}
→ [9007199254740992]
```

**`normalize` returns a key whose Go type depends on whether that conversion succeeded:** float64 when it converts, the raw value when it does not. An `int64` and an `int` of the same large value then land under different map keys, so `in` and `uniq` stop agreeing with `eq`:

```toml
[params]
big = 9007199254740993
bigs = [9007199254740993, 9007199254740992]
```

```
{{ in site.Params.bigs 9007199254740993 }}           → false
{{ uniq (slice site.Params.big 9007199254740993) }}  → [9007199254740993 9007199254740993]
{{ eq site.Params.big 9007199254740993 }}            → true
```

`intersect` now uses `hreflect.CompareNumbers`, the helper 4d628bbd added. `normalize` gets an exact key: integral values key as int64 (uint64 when too large), everything else as float64, so 1 and 1.0 still share a key.

One side effect worth your call: `intersect` no longer matches a nil element against the number 0. On master `{{ intersect (slice 0) (slice nil) }}` returns `[0]`, because `ToFloat64E` maps a nil interface to 0. I added a test row for the new behaviour and can revert that part if you would rather keep it.

Tests are rows in `TestIn`, `TestIntersect` and `TestUniq` plus one integration test. Each one fails on master and passes with the change. `./check.sh ./tpl/...` is clean, `go test ./...` and `staticcheck ./...` pass.

This PR was written primarily by Claude Code, using Opus 5. I reviewed the change and ran everything above.


bep:
Please check if there are other PRs solving the same issue before creating new ones.

